PR #928, opened 2019-02-07
Add a proper list of addresses type for Kademlia

### LINKED ISSUE #924: Add a limit to the number of multiaddresses in Kademlia
Also add a TTL.

tomaka:
Fix #924 

Replaces the `SmallVec<[Multiaddr; 8]>` with a proper `Addresses` struct containing the addresses of a node.

Also changes the logic of Kademlia: either we store only the addresses we're connected to, or only addresses that we're not connected to, but not both.

Also adds a TTL of one hour for addresses. I don't know in practice if that will work though. Since there's no TTL in the protocol, two nodes can just query each other indefinitely and refresh a stale entry.


Wiezzel:
@tomaka
>Also changes the logic of Kademlia: either we store only the addresses we're connected to, or only addresses that we're not connected to, but not both.

What is the purpose of this change?

Also, was it intended that node addresses are no longer added to known addresses when `FindNodeRes` is received? This results in node addresses not being propagated in the network.

tomaka:
> Also, was it intended that node addresses are no longer added to known addresses when `FindNodeRes` is received? This results in node addresses not being propagated in the network.

Yes indeed. It was a misunderstanding of how Kademlia is supposed to behave.
When you receive a Kademlia query, you are supposed to only report about addresses you have been connected to, or that the node self-reports. Indirect propagation of addresses shouldn't happen. Otherwise, since there is no time-to-live mechanism, addresses continue being propagated forever between nodes and never disappear.
However this means that the Kademlia initialization process (where we query 256 random IDs in order to fill our k-buckets) should try to connect to the discovered nodes, and that isn't done.

I originally intended to continue fixing Kademlia in #942, but my progress is blocked by a weird stack overflow in tokio when trying to test #942 
I'll try to extract changes that I have made or intend to make in #942 in separate PRs, in order to finally have a well-behaving well-tested Kademlia.

Wiezzel:
@tomaka What is the priority of fixing Kademlia? Could it be expected anytime soon?

Issue #866, opened 2019-01-17
Finish moving everything to edition 2018


tomaka:

tomaka:
All the `extern crate` statements (except the reexports in the facade) should also be removed before closing this.

tomaka:
Done in #929

PR #929, opened 2019-02-07
Address edition-2018 idioms.

romanb:
Completes https://github.com/libp2p/rust-libp2p/issues/866. This is basically a walk-through with `cargo fix --edition-idioms` augmented by some manual labour. The remaining uses of `extern crate` are as follows:

```
src/lib.rs|313| /// # extern crate libp2p;
transports/tcp/src/lib.rs|30| //! extern crate libp2p_tcp;
transports/websocket/src/lib.rs|55| //! extern crate libp2p_core;
transports/websocket/src/lib.rs|56| //! extern crate libp2p_tcp;
transports/websocket/src/lib.rs|57| //! extern crate libp2p_websocket;
transports/websocket/src/lib.rs|73| extern crate stdweb;
transports/uds/src/lib.rs|36| //! extern crate libp2p_uds;
misc/core-derive/src/lib.rs|23| extern crate proc_macro;
protocols/secio/src/lib.rs|83| extern crate stdweb;
```

That is, it is still used in the public API documentation, for `proc_macro` where it is required and for the `stdweb` macros, due to dependencies on other macros (e.g. `_js_impl`) which would otherwise have to be explicitly imported as well.

Is it desirable to remove the usage of `extern crate` in the public API docs as well already? If so, it could be added to this PR.

### DIFF
--- a/core/src/either.rs
+++ b/core/src/either.rs
@@ -34,7 +34,7 @@ where
     A: fmt::Display,
     B: fmt::Display
 {
-    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         match self {
             EitherError::A(a) => a.fmt(f),
             EitherError::B(b) => b.fmt(f)
--- a/core/src/keys_proto.rs
+++ b/core/src/keys_proto.rs
@@ -103,7 +103,7 @@ impl ::protobuf::Message for PublicKey {
         true
     }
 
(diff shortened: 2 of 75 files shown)


romanb:
The remaining CI failure seems to be due to the fact that the `test-wasm` job still uses a docker image with Rust `1.31` whereas the "uniform-paths" feature has been stabilised in `1.32` which the other CI jobs are using. I've confirmed that build works on that same docker image when upgrading the stable channel.

gnunicorn:
@tomaka any idea how we can update that circle ci image?

tomaka:
https://github.com/libp2p/rust-libp2p/pull/932 updates the image.

PR #934, opened 2019-02-10
Fix compiling for WASM with websockets disabled

tomaka:
Compiling for WASM with WebSockets disabled leads to a compilation error at the moment.

### DIFF
--- a/core/src/transport/dummy.rs
+++ b/core/src/transport/dummy.rs
@@ -23,6 +23,8 @@ use multiaddr::Multiaddr;
 use std::{fmt, io, marker::PhantomData};
 
 /// Implementation of `Transport` that doesn't support any multiaddr.
+///
+/// Useful for testing purposes, or as a fallback implementation when no protocol is available.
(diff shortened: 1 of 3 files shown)


core/src/transport/dummy.rs
```diff
@@ -0,0 +1,108 @@
...
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+// FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+// DEALINGS IN THE SOFTWARE.
+
+use crate::transport::{Transport, TransportError};
+use multiaddr::Multiaddr;
+use std::{fmt, io, marker::PhantomData};
+
+/// Implementation of `Transport` that doesn't support any multiaddr.
```
gnunicorn: Maybe we want to add one sentence about the appropriate use cases?

PR #935, opened 2019-02-10
Fix infinite loop in read_one

tomaka:


tomaka:
Oh, that's a duplicate of #910

PR #942, opened 2019-02-14
Add parity-kademlia

tomaka:
Adds a new alternative `Kademlia` implementation for Substrate/Polkadot.

The difference with libp2p's Kademlia are:

- The protocol is more simple, as we have less messages.
- Each node belongs to a specific namespace. When we connect, we ask the remote what its namespace is. There is no support for multiple namespaces at the moment. While the concept of a namespace can be applied to several use cases, in Substrate/Polkadot we would have one namespace per chain.
- PeerIds are prefixed with the namespace they belong to before being inserted in the k-buckets. The "find peer" query requires passing a namespace in addition to the `PeerId`, and we will obtain results that belong to the same namespace or to a nearby namespace.
- One can also perform a query to find a list of providers for a specific Ristretto public key.
- Each node that owns a Ristretto key can insert its identity in the DHT at the key corresponding to the Ristretto public key. The identity of the node includes a Schnorr signature so that records cannot be spoofed.


tomaka:
Integrates #943

dvc94ch:
@tomaka: why don't u use ipfs as the substrate blockchain backing storage? I think in a couple of weeks, the rust ipfs node should work well enough that I could start prototyping this...

tomaka:
IPFS is not adapted to blockchains.
I'm not sure what you have in mind, but providing blocks as if they were IPFS hashes means that every single node would have to broadcast through the entire network that it has a certain block. It would also prevent retrieving only partial information about a block.

dvc94ch:
Interesting, I've not looked into how substrate handles this, but I thought that IPFS intends to scale to the scale of the current Internet.

dvc94ch:
Do you have a better suggestion for what paritytech finds valuable? I don't know much of anything but I learn quickly. I have a backup job for mid of April that seems like a sure thing, but I've thought that before and it fell through due to me not liking my future boss and declining.

tomaka:
Closing as stale and too many conflicts anyway.

PR #944, opened 2019-02-15
Decode multiaddresses from visit_seq

tomaka:
Apparently `serde_cbor` returns back multiaddresses as "sequences".

### DIFF
--- a/misc/multiaddr/src/lib.rs
+++ b/misc/multiaddr/src/lib.rs
@@ -48,13 +48,23 @@ impl<'de> Deserialize<'de> for Multiaddr {
     where
         D: Deserializer<'de>,
     {
-        struct Visitor;
+        struct Visitor { is_human_readable: bool };
 
         impl<'de> de::Visitor<'de> for Visitor {
             type Value = Multiaddr;
 
             fn expecting(&self, formatter: &mut fmt::Formatter<'_>) -> fmt::Result {
                 formatter.write_str("multiaddress")
+            }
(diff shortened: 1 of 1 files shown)


tomaka:
How would we know that a serde codec (like JSON, CBOR, etc.) doesn't turn a `Vec<u8>` into a sequence of `u8`?
I don't think there's any guideline about what the possible matches are between serialized types and deserialized types.

twittner:
> How would we know that a serde codec (like JSON, CBOR, etc.) doesn't turn a `Vec<u8>` into a sequence of `u8`?

The proposed change was to encode a `[u8]`, not a `Vec<u8>`.

> I don't think there's any guideline about what the possible matches are between serialized types and deserialized types.

The [serde data model](https://serde.rs/data-model.html) distinguishes "byte arrays" from sequence types such as `Vec<_>`. It is my understanding that a hypothetical codec which maps byte arrays to sequences would not be conforming to this model.

twittner:
> The [serde data model](https://serde.rs/data-model.html) distinguishes "byte arrays" from sequence types such as `Vec<_>`. It is my understanding that a hypothetical codec which maps byte arrays to sequences would not be conforming to this model.

I stand corrected. https://docs.rs/serde/1.0.87/serde/trait.Serializer.html#tymethod.serialize_bytes states that forwarding to `serialize_seq` is an option. So I think we should to both, add support for deserialising sequences and encoding into byte arrays.

PR #951, opened 2019-02-18
Rewrite the MemoryTransport to be similar to the TcpConfig

### LINKED ISSUE #913: Rewrite the MemoryTransport to be more similar to the TcpConfig
The problem with the memory transport right now is that it's too limited. We can't test discovery with it, for example.
Instead we should, in my opinion, allow addresses like `/memory/<port>`, and have some sort of static global hub similar to localhost TCP/IP connections.

Considering the usage of `MemoryTransport` everywhere, I would suggest creating a different struct and deprecating the old one, instead of simply rewriting the code.

tomaka:
Makes it possible to use the `MemoryTransport` as a drop-in replacement for `TcpConfig`, except that the addresses are of the form `/memory/N` (where `N` is a u64), instead of `/ip4/.../tcp/...`.

Close #913

### DIFF
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -13,6 +13,7 @@ categories = ["network-programming", "asynchronous"]
 bs58 = "0.2.0"
 bytes = "0.4"
 fnv = "1.0"
+lazy_static = "1.2"
 log = "0.4"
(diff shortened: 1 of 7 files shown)


tomaka:
One potential weird decision is that when you dial a listener, the address that the listener sees for you is the same as the address of the listener.
This mimics the behaviour of `/unix/...` and `/udp/...` (if `udp` was implemented).

PR #957, opened 2019-02-19
Fix simultaneous dialing test

tomaka:
It turns out that the `raw_swarm_simultaneous_connect` test had an issue: we were only applying the muxer as outbound. This was causing the test to erroneously succeed.

It turns out that fixing it was quite complicated. The underlying issue is that when we have two connections to the same remote, and one of them has a higher priority, immediately closing the one with the lower priority is a bad idea, because the remote might not know that the one with the higher priority has been open yet. There's no quick and easy fix to that, as the end of the handshake of the higher-prio connection might get lost or take a shorter path than the closing of the lower-prio connection. The only way to reliably do that is to wait for some data from the remote on the higher priority connection, which confirms that it has acknowledged it.

The changes are as follow:

- Closing a task or interrupting a reach attempt now returns respectively a `ClosedTask` or a `InterruptedReachAttempt`. These object represent a connection that's still open with the remote, but that is detached from the swarm. The swarm won't receive events from them.
- Although not required for this fix, the `TaskClosed` event now returns a `ClosedTask`, for consistency.
- You can now have a task or a connection to a peer take over respectively a `ClosedTask` or a `InterruptedReachAttempt`. "Taking over" means that the closed task or reach attempt will be closed only when we receive an acknowledgement on the task/connection that is taking over.
- Adds a `StreamMuxer::is_remote_acknowledged() -> bool` method. It should return true if the remote has sent some bytes on the network.
- Communication from the outside to the inside of a task is now done through a `ExtToInMessage` enum.
- The simultaneous dialing test now uses `tokio_timer::Delay` to start the dialing. Right now we're using randomness, but that doesn't make sense as we always wait for both nodes to start dialing in the first iteration of the loop.

I had to remote some tests because they were manually constructive private structs, which is a bad idea.

### DIFF
--- a/core/tests/raw_swarm_simult.rs
+++ b/core/tests/raw_swarm_simult.rs
@@ -20,9 +20,10 @@
 
 use futures::{future, prelude::*};
 use libp2p_core::nodes::raw_swarm::{RawSwarm, RawSwarmEvent, IncomingError};
-use libp2p_core::{Transport, upgrade, upgrade::OutboundUpgradeExt};
+use libp2p_core::{Transport, upgrade, upgrade::OutboundUpgradeExt, upgrade::InboundUpgradeExt};
 use libp2p_core::protocols_handler::{ProtocolsHandler, KeepAlive, ProtocolsHandlerEvent, ProtocolsHandlerUpgrErr};
-use std::io;
+use std::{io, time::Duration, time::Instant};
+use tokio_timer::Delay;
 
 // TODO: replace with DummyProtocolsHandler after https://github.com/servo/rust-smallvec/issues/139 ?
 struct TestHandler<TSubstream>(std::marker::PhantomData<TSubstream>, bool);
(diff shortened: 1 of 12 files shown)


tomaka:
This is supposed to fix https://github.com/paritytech/substrate/issues/1777, but apparently it doesn't.

core/tests/raw_swarm_simult.rs
```diff
@@ -209,11 +233,12 @@ fn raw_swarm_simultaneous_connect() {
                         }
                     }
 
-                    if swarm1_step == 3 && swarm2_step == 3 {
+                    // TODO: make sure that >= 5 is correct
+                    if swarm1_step + swarm2_step >= 5 {
```
tomaka: If a node dials, then receives an incoming connection, it will switch to connected and be only at step 2. Then there's no further event. Hence the ` >= 5`, so either must be `3` and the other can be `2`. However I think I should make this a bit stronger before we merge.

tomaka:
I think we should have a `DialError` generated when we cancel an outgoing dial on purpose.
This would also make the test more precise.

tomaka:
I'm going to merge this because we need version 0.4 soon, and open an issue for the `DialError`.

Issue #233, opened 2018-06-04
rust-ipfs? 


adamski:
I'm looking around for an IPFS implementation I can use in a cross-platform desktop/mobile application. I've been playing with https://github.com/Agorise/c-ipfs, but it seems to be incomplete and unstable from my testing. 

I don't necessary need a full IPFS implementation, but would ideally like to be able to interact with nodes powered by go-ipfs. My use case is "offline first" application user data storage, and I'd like to replace my current dependency on Couchbase server infastructure with IPFS. 

In order to use this library I would need to create a C++ wrapper for it, but it seems that things are moving and changing fast here so I wonder if its better to wait until the API has settled?

whyrusleeping:
@adamski With a working libp2p, you can simply connect to any existing ipfs node and start speaking the protocols you want. You would have to reimplement some of those, but the ipfs protocols are actually quite simple. Bitswap is just a single protobuf that you send to request blocks and receive (the same protobuf) to receive those blocks. You could probably implement cpp-ipfs using rust-libp2p pretty easily ;)

adamski:
Great, thats very encouraging, thanks.

adamski:
Coming back to this - are there any examples of using libp2p to speak to IPFS? Are there any docs that explain some of the terms here, e.g. "protobuf"?

adamski:
@tomaka are you able to help with this at all? Any helpful links to similar JS code you can point me to?

whyrusleeping:
> are there any examples of using libp2p to speak to IPFS? Are there any docs that explain some of the terms here, e.g. "protobuf"?

Other than ipfs? (ipfs uses libp2p to talk to other ipfs nodes)

What bits do you want to do? If you just want to be able to fetch data from ipfs nodes, you'll want to implement bitswap. The go implementation of that is here: https://github.com/ipfs/go-bitswap

The raw protobuf message is defined here: https://github.com/ipfs/go-bitswap/blob/master/message/pb/message.proto

adamski:
Yes, I just want to fetch and provide data to other IPFS nodes. Thanks for the links!

adamski:
Does anyone here have a rough idea of days required to implement this? I'm in the process of applying for a research collaboration grant with a local university where implementing rust-ipfs will be the main focus. Thanks.

tomaka:
[...]

Issue #967, opened 2019-02-21
Support /dnsaddr in multiaddr (and maybe /dns as well?)


tomaka:
We don't support multiaddresses with the `/dnsaddr` component.

The `/dns` component is marked as "reserved" [here](https://github.com/multiformats/multiaddr/blob/f3bb08e27b81523d1f1f0371eb1800395b98922d/protocols.csv#L8), so it's unclear whether it should be implemented or not.

This issue also covers adding support for that in `libp2p-dns`, or alternatively opening a different issue for adding support in `libp2p-dns`.

raulk:
go-multiaddr-dns currently handles dns4, dns6, and dnsaddr resolutions. My guess is that dns was reserved as a safeguard to avoid future appropriations that could lead to confusion.

lgierth:
At the time we weren't certain about the needed scope of `/dns`, so we marked it reserved and went with `/dns4` and `/dns6` instead. `/dnsaddr` is an additional mechanism apart from regular A/AAAA record resolution.

ackintosh:
> Support /dnsaddr

I would like to tackle to this issue.  ✨

PR #968, opened 2019-02-21
Trying to talk to legacy js & go IPFS mDNS implementations

gnunicorn:
as reported in #233 

discover over mDNS:
- [x] (latest) jsipfs
- [x] (latest) ipfs-go

---

_It is currently not in the scope, that these can discover us (as we are unsure their messages apply to the standard). However, they still might ;) ._


misc/mdns/src/service.rs
```diff
@@ -398,44 +454,66 @@ impl<'a> MdnsResponse<'a> {
...
+                let mut iter = name.splitn(2, |c| c == '.');
+                if let Ok(peer_id) = match iter.next() {
+                    Some(n) => n.parse::<PeerId>(),
+                    None => return None,
+                } {
+                    if let RData::SRV(srv) = record.data {
+                        // FIXME: assuming ipv4
+                        let addr = match format!(
+                            // FIXME: dns4 is probably incorrect
+                            "/dns4/{}/tcp/{}/p2p/{}",
```
gnunicorn: two issues here:
 - js-libp2p-mdns sends over `os.hostname()` - which isn't necessarily accessible over LAN
 - `dns4` is probably _wrong_ for this use case, `dnsaddrs` might be more appropriate but it is unclear how these are supposed to interact from the spec and we can't parse it yet -- see #967 --

Dialing of this addrs appears to fail anyway.

dvc94ch: `js-libp2p-mdns` also sends an A record for the hostname.

gnunicorn: Yeah, I noticed that, but it was a lot more pain to get into the code base, so it took a lot longer - but it's in there now: https://github.com/libp2p/rust-libp2p/pull/968/files#diff-030eefcb7ad763f8dd6732ebef838e0eR490

gnunicorn:
@dvc94ch this now discovers both legacy ipfs-go and ipfs-js implementations and returns proper multiaddrs for them (`example client1` shows `dialing XYZ`, here, too). It requires a small patch on the ipfs-side and I am not able they ever actually connected... Maybe you want to try it, this is the change required...

```patch
diff --git i/src/p2p/behaviour.rs w/src/p2p/behaviour.rs
index 2a47c45..d9e972e 100644
--- i/src/p2p/behaviour.rs
+++ w/src/p2p/behaviour.rs
@@ -90,7 +90,7 @@ impl<TSubstream: AsyncRead + AsyncWrite, TSwarmTypes: SwarmTypes> Behaviour<TSub
     pub fn new(options: SwarmOptions<TSwarmTypes>, repo: TSwarmTypes::TRepo) -> Self {
         info!("Local peer id: {}", options.peer_id.to_base58());
 
-        let mdns = Mdns::new().expect("Failed to create mDNS service");
+        let mdns = Mdns::new_with_legacy().expect("Failed to create mDNS service");
 
         let mut kademlia = Kademlia::new(options.peer_id.to_owned());
         for (addr, peer_id) in &options.bootstrap {
```

misc/mdns/src/service.rs
```diff
@@ -455,24 +537,73 @@ impl<'a> fmt::Debug for MdnsResponse<'a> {
 }
 
 /// A peer discovered by the service.
-pub struct MdnsPeer<'a> {
```
gnunicorn: I don't think these changes are still needed.

dvc94ch:
Mmh, I can't get it to work. The jsipfs node worked fine when it discovered the rust-ipfs node.

Working scenario:
js-ipfs discovers rust-ipfs
js-ipfs connects to rust-ipfs
rust-ipfs send-wantlist to js-ipfs
js-ipfs send-block to rust-ipfs

Scenario that doesn't work
rust-ipfs discovers js-ipfs
rust-ipfs connects to js-ipfs
nothing happens

gnunicorn:
> js-ipfs discovers rust-ipfs

This works? Curious, because we aren't actually sending any js-ipfs-style discovery-responses.

> rust-ipfs connects to js-ipfs
> nothing happens
> 
> js-ipfs never receives a libp2p connect event.

What is the code you are using? And how are you determining whether the js-ipfs ever received the event?

dvc94ch:
> This works? Curious, because we aren't actually sending any js-ipfs-style discovery-responses.

Sorry for being unclear, it works iff u patch it to work [0].

> What is the code you are using? And how are you determining whether the js-ipfs ever received the event?
```diff
diff --git a/src/core/components/libp2p.js b/src/core/components/libp2p.js
index c45b65f..5b4e286 100644
--- a/src/core/components/libp2p.js
+++ b/src/core/components/libp2p.js
@@ -40,7 +40,10 @@ module.exports = function libp2p (self, config) {
     }
   })
 
-  libp2p.on('peer:connect', peerInfo => peerBook.put(peerInfo))
+  libp2p.on('peer:connect', (peerInfo) => {
+    console.log(peerInfo)
+    return peerBook.put(peerInfo)
+  })
 
   return libp2p
 }
```

- [0] https://github.com/dvc94ch/js-libp2p-mdns

gnunicorn:
your latest master (with my branch) doesn't work for me at all. the background thread dies with:

```bash
$ cargo +nightly run --example client1 
    Finished dev [unoptimized + debuginfo] target(s) in 0.72s
     Running `target/debug/examples/client1`
[2019-02-26T09:52:11Z INFO  ipfs::p2p::behaviour] Local peer id: QmS1CGvbfvZ6iWGddsvCpGiADSD4kQe7MpKdCaZKrsoFkr
[2019-02-26T09:52:12Z INFO  ipfs::p2p] Listening on "/ip4/127.0.0.1/tcp/40199"
thread 'tokio-runtime-worker-0' panicked at 'not yet implemented: async-await-preview currently only supports futures 0.1. Use the compatibility layer of futures 0.3 instead, if you want to use futures 0.3.', /home/ben/.cargo/registry/src/github.com-1ecc6299db9ec823/tokio-async-await-0.1.6/src/compat/backward.rs:76:5
note: Run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
[2019-02-26T09:52:12Z INFO  ipfs::p2p::behaviour] Providing block QmRQKDyEycNmMAWUVXUPdWJXLuWwPwLJCXavpSkUZoX3HR
[2019-02-26T09:52:12Z INFO  ipfs::p2p::behaviour] Want block QmVNrZhKw9JwYa4YPEZVccQxfgQJq993yP78QEN28927vq
[2019-02-26T09:52:12Z INFO  ipfs::p2p::behaviour] kad: Could not find provider for QmVNrZhKw9JwYa4YPEZVccQxfgQJq993yP78QEN28927vq
``` 

but also under my previous version, I didn't see any network traffic happening between the nodes after the rust ipfs telling me "queuing dialing".

dvc94ch:
Sorry about that, I haven't pushed the change yet.

```diff
git diff src/future.rs
diff --git a/src/future.rs b/src/future.rs
index 270275a..858bafb 100644
--- a/src/future.rs
+++ b/src/future.rs
@@ -32,6 +32,8 @@ impl<TBlockStore: BlockStore> Future for BlockFuture<TBlockStore> {
             Poll::Ready(Ok(None)) => {
                 let future = self.block_store.get(self.cid.clone());
                 self.get_mut().future = future;
+                tokio::prelude::task::current().notify();
+                //waker.wake();
                 Poll::Pending
             },
             Poll::Ready(Err(err)) => {
```

dvc94ch:
> but also under my previous version, I didn't see any network traffic happening between the nodes after the rust ipfs telling me "queuing dialing".

The poll function then returns the `DialPeer` event that was queued and the `Swarm` should do the connecting. I still haven't figured the problem out yet, but I appreciate your time and input on this.

gnunicorn:
@dvc94ch FYI, you can also [find me on riot](https://matrix.to/#/@ben:matrix.parity.io)

tomaka:
Relevant: https://github.com/libp2p/specs/pull/80#issuecomment-481748611

PR #971, opened 2019-02-25
Add libp2p-bluetooth

tomaka:
Status: working prototype, but needs a lot of cleanup, possibly extracting in a different library.

Adds a Bluetooth transport in the form of the `BluetoothConfig`, and a network behaviour named `BluetoothDiscovery`.

When listening, the transport will listen to an RFCOMM port, make the local device discoverable by outside devices, and register the service to the local SDP (Service Discovery Protocol) server managed by the Bluetooth daemon of the system.

This SDP registeration registers the libp2p service with the following service class ID: `d8263f85-cdca-4ac8-8fee-81c64221d6d5` (this is an arbitrary UUID that I generated on my machine). This service contains a custom attribute whose ID is `0x0300` and whose value is the base58 of the local peer ID.

The `BluetoothDiscovery` type will scan nearby Bluetooth devices for any that supports this libp2p service, and will store a view of this them. The associated multiaddresses are of the form `/bluetooth/xx:xx:xx:xx:xx:xx/l2cap/3/rfcomm/x`.

This multiaddress format can be dialed by the transport, which will establish a connection through RFCOMM.


tomaka:
We are unfortunately depending on libdbus. I can't find any pure Rust alternative to it.

tomaka:
> I can't find any pure Rust alternative to it.

Update: there are now at least two pure Rust dbus libraries:
- https://gitlab.freedesktop.org/zeenix/zbus
- https://github.com/KillingSpark/rustbus

thomaseizinger:
I've spent a bit of time looking through the Rust Bluetooth ecosystem and at least for targeting Linux only, this looks very promising: https://github.com/bluez/bluer

thomaseizinger:
> I've spent a bit of time looking through the Rust Bluetooth ecosystem and at least for targeting Linux only, this looks very promising: [bluez/bluer](https://github.com/bluez/bluer)

This does depend on `dbus` which depends on `libdbus` though.

umgefahren:
Is this transport abandoned?

Since bluetooth crates are discussed here, I would like to through in this one: https://crates.io/crates/btleplug

thomaseizinger:
> Is this transport abandoned?

I think it'd be really cool but yeah, someone needs to drive (and eventually spec) this.

> Since bluetooth crates are discussed here, I would like to through in this one: https://crates.io/crates/btleplug

This one is host-only so I am not sure if it is sufficient.

thomaseizinger:
Closing as stale.

Issue #903, opened 2019-01-30
Somehow unify noise::Keypair and secio::KeyPair


tomaka:
It feels stupid to have duplicates of all the cryptographic functions.
However things are a bit blurry in terms of API.

burdges:
As I mentioned elsewhere, we should hide SECIO behind some scary sounding feature flag so that nobody uses it going forward.  

We would likely care about roughly this issue whenever we add a post-quantum ephemeral key exchange, probably Kyber:  https://pq-crystals.org/kyber/  

Also, if we ever add TLS 1.3 integration for all these web-like transports, but I've kinda changed my mind about TLS 1.3 making sense because It'd primarily serve to better obfuscate polkadot traffic and we'd do better with an actual pluggable transports scheme.

romanb:
Closed by https://github.com/libp2p/rust-libp2p/pull/972

Issue #807, opened 2018-12-27
Expand identity keypairs with more functionalities


tomaka:
- Add the possibility to generate secp256k1 and RSA keys.
- Add the possibility to export the generated private key, with hopefully an API that's not too easy to misuse.
- Maybe add some utility functions to load/save a private key to file, with the file having the right permissions (400).
- As a more general thing, try to improve the usability of this struct.

tomaka:
> Maybe add some utility functions to load/save a private key to file, with the file having the right permissions (400).

What I mean by that is that the content of this file could be moved to libp2p:
https://github.com/paritytech/substrate/blob/ecfc6f14846bbe3778573d2ea1b9613090fcd91c/core/network-libp2p/src/secret.rs

tomaka:
We should also consider moving this type to the core, or in a general `libp2p-crypto` library.

burdges:
I'm ambivalent about RSA but sure why not.  :)

We should not really encourage using secp256k1 at the transport layer, primarily because secp256k1 implementations have issues with side channels.  I've even heard folks who really know secp256k1 claim only one good implementation exists.  Could secp256k1 have some feature gate with a warning?

romanb:
An update in the light of https://github.com/libp2p/rust-libp2p/pull/972 where `SecioKeyPair` has been removed:

> Add the possibility to generate secp256k1 and RSA keys.

Support for generating Secp256k1 keys is there. RSA keys cannot easily be generated since ring does not have support for doing that yet (https://github.com/briansmith/ring/issues/219). I don't think that is actually important though.

> Add the possibility to export the generated private key, with hopefully an API that's not too easy to misuse.

Both ed25519 and secp256k1 identity keypairs provide access to the secret key bytes, e.g. for storage. These are also the keypairs that can be generated through code, so right now there is a sensible symmetry: RSA keypairs must be created externally and cannot be exported / stored from code, whereas ed25519 and secp256k1 keypairs can be generated in code and also exported for storage.

> Maybe add some utility functions to load/save a private key to file, with the file having the right permissions (400).

I think that is best left to a separate libp2p crate (e.g. `libp2p-keystore` or similar) or even an external crate. In any case it is not something that needs to be in `libp2p-core` and hence should probably not be there IMHO.

> As a more general thing, try to improve the usability of this struct.

I don't know what that entails in the context of the new `libp2p_core::identity::Keypair`.

[...]

PR #972, opened 2019-02-25
Consolidate keypairs in core.

romanb:
This PR is a proposal for some consolidation w.r.t. keypairs used in libp2p, including some small enhancements. It is primarily a refactoring with small improvements along the way in terms of addressing small problems encountered and adding more tests and documentation.

## Context

The context of this PR is mainly https://github.com/libp2p/rust-libp2p/issues/903 and [this comment](https://github.com/libp2p/rust-libp2p/issues/807#issuecomment-457533582).

## Motivation

The concept of nodes having a cryptographic identity is central to libp2p, hence the motivation
to introduce the concept of an identity keypair in `libp2p_core`, instead of only having the
concept of a public key there, as it currently stands.

## Changes

Since the PR contains some noise as a result of some module and function name
changes that make it seem more unwieldy than it actually is, I provide summaries of the main
changes below, with links to the most relevant parts of the changeset.

### `core`

The new module `identity` defines the concept of an identity keypair alongside the already
existing public key, which moved to this module. The submodules define the supported types
of identity keypairs, currently `ed25519`, `rsa` and `secp256k1`. The `identity::Keypair`
and `identity::PublicKey` provide generic functionality for creating and validating signatures.
Overall the `identity::Keypair` is a reincarnation of the former `SecioKeyPair`.

While migrating code from `secio` into `core`, I ran across [this TODO](https://github.com/libp2p/rust-libp2p/blob/master/protocols/secio/src/handshake.rs#L451-L453). As far as I could tell, the underlying problem is that the [openssl rsa](https://www.openssl.org/docs/man1.0.2/man1/openssl-rsa.html) command reads and emits [X.509 SubjectPublicKeyInfo](https://tools.ietf.org/html/rfc5280#section-4.1) structures in DER by default, whereas `ring` works with a DER-encoded [PKCS#1 RSAPublicKey](https://tools.ietf.org/html/rfc3447#appendix-A.1.1). The former is preceded by an `AlgorithmIdentifier` which I presume is stripped off by the current code to "fix" the problem. I didn't really see why it is necessary to extract the public key out of the PKCS#8 format into DER outside of libp2p and then feed it in separately. Instead, the public key is now extracted from the `RsaKeyPair`, where it already is in the desired encoding (PKCS#1 RSAPublicKey in DER). Additionally, public RSA keys now support explicit encoding/decoding to/from X.509 SubjectPublicKeyInfo structures in DER, as is required for libp2p interoperability. So these changes are backward-compatible and just simplify importing RSA keys as well as add stricter parsing of public keys.

Main changes:

  * [core::identity](https://github.com/libp2p/rust-libp2p/compare/master...romanb:core-identity?expand=1#diff-b1992da482356f9ac08e6f9b31cbb7f9)
  * [core::identity::rsa](https://github.com/libp2p/rust-libp2p/compare/master...romanb:core-identity?expand=1#diff-60ad6ab176a38443ed68e45450abfc36)
  * [core::identity::ed25519](https://github.com/libp2p/rust-libp2p/compare/master...romanb:core-identity?expand=1#diff-6fc7fa740301b374ba592d98db7b4e77)
  * [core::identity::secp256k1](https://github.com/libp2p/rust-libp2p/compare/master...romanb:core-identity?expand=1#diff-e18b2655c2a7bfb50e6eef104bd2b4a9)

### `secio`

The `secio` protocol has been simplified as a result of the keypair consolidation in `core`. 

Main changes:

  * [secio::lib](https://github.com/libp2p/rust-libp2p/compare/master...romanb:core-identity?expand=1#diff-63e26be49df276a1d6dd2ddb0db991cc)
  * [secio::handshake](https://github.com/libp2p/rust-libp2p/compare/master...romanb:core-identity?expand=1#diff-76f12ee8bbd77353bb44b67257c036bf)

### `noise`

Since the Ed25519 keypairs have moved to core in this PR, that leaves only the Curve25519 keypairs for X25519 in `libp2p-noise`, as well as the optional conversion of Ed25519 keys to X25519 keys, which I completed by adding private key (and thus keypair) conversion, including tests for conformity with libsodium. The implementation has been slightly simplified by reusing the x25519 function and types from `x25519-dalek`.

The custom `snow::types::Dh` implementation is equivalent to the one provided by
snow's current `DefaultResolver`, however, I assumed that disabling the default features of `snow`
was very much deliberate, possibly to avoid pulling in some of its dependencies, and possibly to
avoid being surprised by changing defaults, so I left that in place, only slightly simplified.
Otherwise, if these are not concerns, we could delegate for both the choice of `Dh` and `Rng` to the
default resolver (the current default `Dh` in snow also uses `x25519-dalek`, the default `Rng` is
the `OsRng`).

Main changes:

  * [noise::keys](https://github.com/libp2p/rust-libp2p/compare/master...romanb:core-identity?expand=1#diff-6a3ddc4bb8d09566deb9a2c1ebe10796)
  * [noise::util](https://github.com/libp2p/rust-libp2p/compare/master...romanb:core-identity?expand=1#diff-18d56171885b6b73362717c0a78d04e0)


protocols/noise/src/keys.rs
```diff
@@ -18,156 +18,185 @@
...
-    bytes: [u8; 32],
-    _marker: std::marker::PhantomData<T>
+use curve25519_dalek::edwards::CompressedEdwardsY;
+use libp2p_core::identity::ed25519;
+use ring::digest::{SHA512, digest};
+use x25519_dalek as x25519;
+
+/// X25519 static keypair.
+pub struct StaticKeypair {
+    secret: x25519::StaticSecret,
```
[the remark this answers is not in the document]
burdges: We might abstract Noise over the curve, perhaps as part of trying to get https://noiseexplorer.com/ to generate Rust, but I'd maybe first ask Trevor and/or search the [Noise mailing list archives](https://moderncrypto.org/mail-archive/noise/) for any discussion about this. 

Aside from Cruve448, I'd think the target curves might be FourQ because it's possibly faster and only very slightly less secure, and the Ristretto group for when you want complex zk protocols integrated with your key exchange.  I'd avoid secp256k1 here myself.

protocols/noise/src/keys.rs
```diff
@@ -18,156 +18,185 @@
...
+impl StaticPublicKey {
+    /// Construct a static X25519 public key from an Ed25519 public key.
+    ///
+    /// See also [`StaticKeypair::from_ed25519`].
     ///
-    /// Performs scalar multiplication with the given public key.
-    pub fn ecdh(&self, pk: &PublicKey<Curve25519>) -> [u8; 32] {
-        self.x25519(&MontgomeryPoint(pk.bytes)).0
+    /// [`StaticKeypair::from_ed25519`]: struct.StaticKeypair.html#method.from_ed25519
+    pub fn from_ed25519(pk: &ed25519::PublicKey) -> Option<StaticPublicKey> {
```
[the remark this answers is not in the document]
burdges: An ed25519 deserialization can fail when the point lies on the twist.  Actually x25519 too but x25519 does not consider this an error.

In this case, `ed25519::PublicKey` seemingly wraps an `ed25519_dalek::PublicKey`, which previously wrapped just a `[u8; 32]`.  In that state, there is no reason to believe decompression occurred previously, so this could error if it's the first decompression.  

I think `ed25519_dalek::PublicKey` now does decompression earlier, but not sure the version being used here, or if they might change their mind again in future.

burdges:
In the long run, SECIO should be removed or more likely hidden behind some scary sounding feature flag.

burdges:
If I understand, identity key pairs exist solely to certify the non-ephemeral keys actually used by Noise, yes?  There are already signature extensions to Noise which using might someday simplify another implementation, although currently most never get implemented.  I'm not familiar with them but some do now appear and many should be discussed at https://moderncrypto.org/mail-archive/noise/

Related:  https://github.com/tendermint/tendermint/issues/3315

protocols/noise/src/keys.rs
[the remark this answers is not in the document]
twittner: Curve 448 should be uncontroversial as it is already part of [the spec](http://noiseprotocol.org/noise.html#the-448-dh-functions).

protocols/noise/src/keys.rs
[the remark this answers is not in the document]
twittner: Yes, I had assumed that the Edwards key would be uncompressed because it is `encode`d first and then decompressed again and the conversion to Montgomery should be infallible if done directly. But I have not looked into `core::identity::ed25519` yet.

protocols/noise/src/keys.rs
```diff
@@ -18,156 +18,185 @@
...
+    ///
+    /// [Noise: Static Key Reuse](http://www.noiseprotocol.org/noise.html#security-considerations)
+    /// [Ed25519 to Curve25519](https://libsodium.gitbook.io/doc/advanced/ed25519-curve25519)
+    pub fn from_ed25519(ed: &ed25519::Keypair) -> StaticKeypair {
+        // An Ed25519 public key is derived off the left half of the SHA512 of the
+        // secret scalar, hence a matching conversion of the secret key must do
+        // the same to yield a Curve25519 keypair with the same public key.
+        let ed25519_sk = ed.secret_bytes();
+        let mut curve25519_sk: [u8; 32] = [0; 32];
+        let hash = digest(&SHA512, ed25519_sk);
```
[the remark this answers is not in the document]
romanb: There is a comment just a few lines above. See also [this code](https://docs.rs/ed25519-dalek/1.0.0-pre.1/src/ed25519_dalek/public.rs.html#53-67).

protocols/noise/src/keys.rs
```diff
@@ -18,156 +18,185 @@
...
 
-impl SecretKey {
-    /// Compute the public key from this secret key.
-    ///
-    /// Performs scalar multiplication with Curve25519's base point.
-    pub fn public(&self) -> PublicKey<Curve25519> {
-        PublicKey::new(self.x25519(&X25519_BASEPOINT).0)
-    }
+/// Static X25519 public key.
+pub struct StaticPublicKey(pub x25519::PublicKey);
```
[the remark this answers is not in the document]
romanb: Only because the module generally does not intend to hide / encapsulate the use of `x25519-dalek`, but only add a bit of functionality on top. It is nevertheless certainly not really necessary to have the field `pub`. This will probably change anyway in the course of the necessary changes to also support other curves.

protocols/noise/src/keys.rs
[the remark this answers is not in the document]
romanb: I will ty to come up with some changes to keep the door open for other curves in the future.

protocols/noise/src/util.rs
```diff
@@ -80,49 +86,33 @@ impl rand::CryptoRng for Rng {}
...
 impl snow::types::Random for Rng {}
 
-/// X25519 Diffie-Hellman key agreement.
-struct X25519 {
-    keypair: Keypair<Curve25519>
-}
+/// Short-lived container for a static or ephemeral X25519 keypair
+/// used during DH computations by `snow`.
+#[derive(Default)]
+struct Dh25519 { sk: [u8; 32], pk:  [u8; 32] }
```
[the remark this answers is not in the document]
romanb: Good point, thanks.

core/src/identity.rs
```diff
@@ -0,0 +1,195 @@
...
+pub mod ed25519;
+#[cfg(not(any(target_os = "emscripten", target_os = "unknown")))]
+pub mod rsa;
+#[cfg(feature = "secp256k1")]
+pub mod secp256k1;
+
+pub mod error;
+
+use self::error::*;
+// use log::debug;
```
tomaka:
```suggestion
```

core/src/identity.rs
```diff
@@ -0,0 +1,195 @@
...
+pub enum Keypair {
+    Ed25519(ed25519::Keypair),
+    #[cfg(not(any(target_os = "emscripten", target_os = "unknown")))]
+    Rsa(rsa::Keypair),
+    #[cfg(feature = "secp256k1")]
+    Secp256k1(secp256k1::Keypair)
+}
+
+impl Keypair {
+    pub fn new_ed25519() -> Keypair {
```
tomaka:
```suggestion
    /// Generates a new ED25519 key pair.
    pub fn generate_ed25519() -> Keypair {
```

core/src/identity.rs
```diff
@@ -0,0 +1,195 @@
...
+        Keypair::Ed25519(ed25519::Keypair::generate())
+    }
+
+    #[cfg(feature = "secp256k1")]
+    pub fn new_secp256k1() -> Keypair {
```
tomaka: Needs doccomment.

core/src/identity.rs
```diff
@@ -0,0 +1,195 @@
...
+        Keypair::Secp256k1(secp256k1::Keypair::generate())
+    }
+
+    #[cfg(not(any(target_os = "emscripten", target_os = "unknown")))]
+    pub fn rsa_from_pkcs8(pkcs8_der: &[u8]) -> Result<Keypair, DecodingError> {
```
tomaka: Needs doccomment.

core/src/identity.rs
```diff
@@ -0,0 +1,195 @@
...
+            Rsa(pk) => pk.verify(msg, sig),
+            #[cfg(feature = "secp256k1")]
+            Secp256k1(pk) => pk.verify(msg, sig)
+        }
+    }
+
+    /// Encode the public key according to the [libp2p-specs].
+    ///
+    /// [libp2p-specs]: https://github.com/libp2p/specs/blob/feat/peer-ids/peer-ids/peer-ids.md
+    pub fn encode(self) -> Result<Vec<u8>, EncodingError> {
```
tomaka: Should be named `to_protobuf_encoding()` for consistency with https://docs.rs/libp2p-core/0.4.0/libp2p_core/enum.PublicKey.html#method.into_protobuf_encoding

I also don't think it's worth mentioning that we do that according to the specs. This is implied for everything we do.

core/src/identity.rs
```diff
@@ -0,0 +1,195 @@
...
+
+        public_key
+            .write_to_bytes()
+            .map_err(|e| EncodingError::new("Protobuf", e))
+    }
+
+    /// Decodes the public key according to the [libp2p-specs].
+    ///
+    /// [libp2p-specs]: https://github.com/libp2p/specs/blob/feat/peer-ids/peer-ids/peer-ids.md
+    pub fn decode(bytes: &[u8]) -> Result<PublicKey, DecodingError> {
```
tomaka: `from_protobuf_encoding`

core/src/identity/error.rs
```diff
@@ -0,0 +1,97 @@
+
```
tomaka: Missing license.

core/src/identity/error.rs
```diff
@@ -0,0 +1,97 @@
...
+}
+
+impl From<String> for EncodingError {
+    fn from(s: String) -> EncodingError {
+        EncodingError { msg: s, source: None }
+    }
+}
+
+impl Error for EncodingError {
+    fn cause(&self) -> Option<&dyn Error> {
```
tomaka: I think `cause()` is deprecated in favour of `source`.

core/src/identity/error.rs
```diff
@@ -0,0 +1,97 @@
...
+}
+
+impl fmt::Display for DecodingError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "Key decoding error: {}", self.msg)
+    }
+}
+
+impl Error for DecodingError {
+    fn cause(&self) -> Option<&dyn Error> {
```
tomaka: Same.

core/src/identity/error.rs
```diff
@@ -0,0 +1,97 @@
...
+}
+
+impl From<String> for SigningError {
+    fn from(s: String) -> SigningError {
+        SigningError { msg: s, source: None }
+    }
+}
+
+impl Error for SigningError {
+    fn cause(&self) -> Option<&dyn Error> {
```
tomaka: Same.

core/src/identity/rsa.rs
```diff
@@ -0,0 +1,254 @@
...
+    /// Create a `Keypair` from a DER-encoded private key in PKCS#8 PrivateKeyInfo
+    /// format (i.e. unencrypted).
+    pub fn from_pkcs8(pk8: &[u8]) -> Result<Keypair, DecodingError> {
+        let kp = RsaKeyPair::from_pkcs8(Input::from(&pk8[..]))
+            .map_err(|e| DecodingError::new("RSA PKCS#8", e))?;
+        Ok(Keypair(Arc::new(kp)))
+    }
+
+    /// Obtain the public key from the RSA keypair.
+    pub fn public(&self) -> PublicKey {
```
tomaka: `to_public_key()` instead maybe?

misc/peer-id-generator/src/main.rs
```diff
@@ -67,13 +66,11 @@ fn main() {
...
         thread::spawn(move || loop {
-            let private_key: [u8; 32] = rand::random();
-            let generated = SecioKeyPair::secp256k1_raw_key(private_key).unwrap();
-            let peer_id: PeerId = generated.to_public_key().into_peer_id();
+            let generated = identity::Keypair::new_secp256k1();
+            let peer_id = generated.public().into_peer_id();
             let base58 = peer_id.to_base58();
             if base58[2..].starts_with(&prefix) {
                 println!("Found {:?}", peer_id);
-                println!("=> Private key = {:?}", private_key);
```
tomaka: The while point of this binary is that it prints the private keys that it generates.

protocols/noise/src/keys.rs
```diff
@@ -18,156 +18,185 @@
...
     }
-}
 
-impl<T> Into<(SecretKey, PublicKey<T>)> for Keypair<T> {
-    fn into(self) -> (SecretKey, PublicKey<T>) {
-        (self.secret, self.public)
+    // Bindings to libsodium's ed25519 to curve25519 key conversions, to check that
+    // they agree with the conversions performed in this module.
+
+    extern "C" {
```
tomaka: I know it's a test, but by principle I'm not really fan of introducing any sort of unsafe call at all.

core/src/identity.rs
```diff
@@ -0,0 +1,195 @@
...
+pub enum PublicKey {
+    Ed25519(ed25519::PublicKey),
+    #[cfg(not(any(target_os = "emscripten", target_os = "unknown")))]
+    Rsa(rsa::PublicKey),
+    #[cfg(feature = "secp256k1")]
+    Secp256k1(secp256k1::PublicKey)
+}
+
+impl PublicKey {
+    pub fn verify(&self, msg: &[u8], sig: &[u8]) -> bool {
```
tomaka: Maybe `Result<(), ()>` is more appropriate?

core/src/identity.rs
```diff
@@ -0,0 +1,195 @@
...
+        match self {
+            Ed25519(ref pair) => Ok(pair.sign(msg)),
+            #[cfg(not(any(target_os = "emscripten", target_os = "unknown")))]
+            Rsa(ref pair) => pair.sign(msg),
+            #[cfg(feature = "secp256k1")]
+            Secp256k1(ref pair) => Ok(pair.sign(msg)),
+        }
+    }
+
+    pub fn public(&self) -> PublicKey {
```
tomaka: `to_public_key`?

tomaka:
A merge or rebase on master should fix the WASM CI showing this weird error.

protocols/noise/src/keys.rs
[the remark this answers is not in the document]
romanb: Indeed [PublicKey::from_bytes](https://docs.rs/ed25519-dalek/1.0.0-pre.1/src/ed25519_dalek/public.rs.html#128-144) already does the decompression, and that is also used in [Keypair::from_bytes](https://docs.rs/ed25519-dalek/1.0.0-pre.1/src/ed25519_dalek/ed25519.rs.html#200-211), so decompression should actually be infallible at this point.

romanb:
> If I understand, identity key pairs exist solely to certify the non-ephemeral keys actually used by Noise, yes?

If you are referring to the authenticity of the static public keys, then potentially yes, just not only for Noise. My understanding is that in general the concept of a node in libp2p having a long-term keypair whose public key is hashed into the "peer ID" is in itself independent of specific protocols but can be used to authenticate the remote in such protocols, if its public identity key is already known (and in some way verified or trusted). Hence the ability to use an ed25519 identity keypair of a node as a static curve25519 keypair in the Noise protocol, but it may also use a different one, whose authenticity then needs to be verified separately, of course.

protocols/noise/src/keys.rs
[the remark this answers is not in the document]
romanb: I create a new PR [here](https://github.com/romanb/rust-libp2p/pull/1) against the branch of this PR that shows the approach I came up with so far to generalise over the choice of DH.

protocols/noise/src/keys.rs
romanb: What do you suggest? To remove the test?

core/src/identity.rs
romanb: In what sense more appropriate?

romanb: Do you mean the implied semantics of `Ok` and `Err` are superior to `true` and `false`? If you think that is preferable, I don't mind.

tomaka: > Do you mean the implied semantics of `Ok` and `Err` are superior to `true` and `false`?

That's what I mean, yes. I don't have a strong opinion though.

protocols/noise/src/keys.rs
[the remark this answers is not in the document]
burdges: In general, you cannot necessarily rely on that always being the case because some libraries might do this differently.

romanb:
I think I have addressed all smaller comments with two exceptions:

  * The renaming of `public()` -> `to_public_key()` across all keypair types.
  * Changing the result type from `bool` to `Result<(),()>` for signature verification functions.

I would only do these if someone feels strongly about them. I also added some more doc-comments and memory zeroing of secrets using the `zeroize` crate (suggested by @twittner) in the new `core::identity` modules where it seemed applicable.

For the necessary changes on the Noise protocol for abstracting over the choice of DH keys, I've created a [separate PR](https://github.com/romanb/rust-libp2p/pull/1) for which I would appreciate feedback. It could either be merged into the branch for this PR or as a follow-up PR against `rust-libp2p` directly, after this one has been merged, whatever is preferred.

twittner:
> I've created a [separate PR](https://github.com/romanb/rust-libp2p/pull/1) for which I would appreciate feedback.

I intend to review this soon-ish.

core/src/identity/error.rs
```diff
@@ -0,0 +1,117 @@
...
+/// An error during signing of a message.
+#[derive(Debug)]
+pub struct SigningError {
+    msg: String,
+    source: Option<Box<dyn Error + Send + Sync>>
+}
+
+/// An error during encoding of key material.
+impl SigningError {
+    pub fn new(msg: &str, source: impl Error + Send + Sync + 'static) -> SigningError {
```
tomaka:
```suggestion
    pub(crate) fn new(msg: &str, source: impl Error + Send + Sync + 'static) -> SigningError {
```

core/src/identity/error.rs
```diff
@@ -0,0 +1,117 @@
...
+
+#[derive(Debug)]
+pub struct EncodingError {
+    msg: String,
+    source: Option<Box<dyn Error + Send + Sync>>
+}
+
+/// An error during encoding of key material.
+impl EncodingError {
+    pub fn new(msg: &str, source: impl Error + Send + Sync + 'static) -> EncodingError {
```
tomaka:
```suggestion
    pub(crate) fn new(msg: &str, source: impl Error + Send + Sync + 'static) -> EncodingError {
```

core/src/identity/error.rs
```diff
@@ -0,0 +1,117 @@
...
+
+/// An error during decoding of key material.
+#[derive(Debug)]
+pub struct DecodingError {
+    msg: String,
+    source: Option<Box<dyn Error + Send + Sync>>
+}
+
+impl DecodingError {
+    pub fn new(msg: &str, source: impl Error + Send + Sync + 'static) -> DecodingError {
```
tomaka:
```suggestion
    pub(crate) fn new(msg: &str, source: impl Error + Send + Sync + 'static) -> DecodingError {
```

core/src/identity/rsa.rs
tomaka: :arrow_up:

core/src/identity.rs
```diff
@@ -0,0 +1,208 @@
...
+            #[cfg(not(any(target_os = "emscripten", target_os = "unknown")))]
+            Rsa(pk) => pk.verify(msg, sig),
+            #[cfg(feature = "secp256k1")]
+            Secp256k1(pk) => pk.verify(msg, sig)
+        }
+    }
+
+    /// Encode the public key into a protobuf structure for storage or
+    /// exchange with other nodes.
+    pub fn into_protobuf_encoding(self) -> Result<Vec<u8>, EncodingError> {
```
tomaka: This shouldn't return an error.

core/src/identity.rs
```diff
@@ -0,0 +1,208 @@
...
+            }
+            #[cfg(not(feature = "secp256k1"))]
+            keys_proto::KeyType::Secp256k1 => {
+                log::debug!("support for secp256k1 was disabled at compile-time");
+                Err("Unsupported".to_string().into())
+            },
+        }
+    }
+
+    pub fn into_peer_id(self) -> PeerId {
```
tomaka: Doc-comment.

tomaka:
Also, before merging I need to ask if we are absolutely certain that this is backwards compatible.
I don't want to introduce even more issues into Substrate. We already have more than enough.

romanb:
> I'd also love if everything had a doc-comment, because I'd like to do #976 eventually, and this pushes us further away from it.

I'm not sure what the scope of "everything" is in the context of this PR, but I've at least fully covered the newly introduced modules now, along with addressing the rest of your latest comments. Feel free to point out anything else that is missing.

> Also, before merging I need to ask if we are absolutely certain that this is backwards compatible on the network.

That is imperative, of course, but I take every approving review as confirmation on that matter. What is your verdict? Besides that, do you have any specific measures or additional tests in mind to increase confidence on the network-level backward-compatibility?

The changes to the library APIs are obviously not backward-compatible but I'm already preparing a branch for the necessary changes to substrate in the context of also supporting Ed25519 (i.e. https://github.com/paritytech/substrate/issues/1494).

> I don't want to introduce even more issues into Substrate. We already have more than enough.

That seems to imply that anything merged to master here is picked up by substrate shortly after, but surely if there are important networking issues being worked out in substrate, any major dependency upgrades are out of the question anyway and substrate would work off a stable release branch of libp2p that only receives patches until the situation stabilizes?

tomaka:
> The changes to the library APIs are obviously not backward-compatible but I'm already preparing a branch for the necessary changes to substrate in the context of also supporting Ed25519 (i.e. paritytech/substrate#1494).

Ah nice! If the Substrate branch seems to work, that is a confirmation that the changes are backwards compatible.

> That seems to imply that anything merged to master here is picked up by substrate shortly after, but surely if there are important networking issues being worked out in substrate, any major dependency upgrades are out of the question anyway and substrate would work off a stable release branch of libp2p that only receives patches until the situation stabilizes?

There are certainly quite a few issues in Substrate that comes from libp2p. It's quite important IMO that we can fix things in libp2p and then update Substrate without introducing new issues.

tomaka:
If the Substrate ed25519 branch works, then let's merge this once the network in Substrate is generally working ok.

tomaka:
After looking more deeply into it, it doesn't seem that any of the issues in Substrate are caused by libp2p. We can merge this PR as soon as you have a PR for Substrate that can connect to an existing testnet.

tomaka:
Confirmed working on substrate.

Issue #23, opened 2017-11-01
Add an example of rust-libp2p running in the browser


tomaka:
With asmjs and/or wasm.

Is that an objective as well? It should be easy to achieve.

diasdavid:
You can run libp2p in the browser today by using the JavaScript implementation https://github.com/libp2p/js-libp2p

You can find an example here https://github.com/libp2p/js-libp2p/tree/master/examples/libp2p-in-the-browser

tomaka:
It is now possible to open a websocket listener from the browser by compiling for emscripten.
[The gist here](https://gist.github.com/tomaka/25e0f01c427a1bce0e0ddb10d278ead9) shows how to drive a future to completion using the `stdweb` crate.

Things still to be done:

- Get crypto to work. The `ring` library doesn't compile for emscripten.
- Add continuous integration. Unfortunately travis is not the most appropriate solution to compile for emscripten, and even if we manage to compile the listener testing it would require a javascript interpreter.

tomaka:
> Add continuous integration. Unfortunately travis is not the most appropriate solution to compile for emscripten, and even if we manage to compile the listener testing it would require a javascript interpreter.

This is done.

> Get crypto to work. The ring library doesn't compile for emscripten.

This is not going to happen anytime soon.
I think the best solution is to tweak the `paritytech/ring` repo to use the `WebCrypto` API when compiling for emscripten.

gnunicorn:
[Futures supports stdweb (since version 0.2)](https://github.com/koute/stdweb/pull/206), which in turn supports `wasm32-unknown-unknown`-targets now, too. So, if we can upgrade futures to `0.2`

Regarding
> I think the best solution is to tweak the paritytech/ring repo to use the WebCrypto API when compiling for emscripten.

I very much agree with [brian smith, that getting WebCrypto into Ring will be a challenge](https://github.com/briansmith/ring/issues/657#issuecomment-396416821), already because ring is synchronous, while WebCrypto returns promises.

I think the better approach would be to refactor libp2p to use `WebCrypto` and `std_web` for `wasm`-targets, dropping ring altogether in that case. However, that means making a bunch of things for async than they are right now, I suppose. 

---

Edit: that might also turn out to be non-trivial, as `WebCrypto` doesn't yet support all algorithms, that we need e.g. [secp256k1](https://github.com/w3c/webcrypto/issues/82). So I guess for some curves we might be more vulnerable within the wasm-environment :thinking: .

[...]

PR #975, opened 2019-02-26
Fall back if no tokio executor available

tomaka:
If no tokio executor is available, we fall back to executing tasks directly.

This is mostly for #23

### DIFF
--- a/core/src/nodes/handled_node_tasks.rs
+++ b/core/src/nodes/handled_node_tasks.rs
@@ -35,7 +35,7 @@ use std::{
     fmt,
     mem
 };
-use tokio_executor;
+use tokio_executor::Executor;
 
 mod tests;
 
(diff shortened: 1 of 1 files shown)


core/src/nodes/handled_node_tasks.rs
```diff
@@ -316,7 +320,24 @@ impl<TInEvent, TOutEvent, TIntoHandler, TReachErr, THandlerErr, TUserData, TPeer
     // TODO: look into merging with `poll()`
     fn poll_inner(&mut self) -> Async<(InToExtMessage<TOutEvent, TIntoHandler, TReachErr, THandlerErr, TPeerId>, TaskId)> {
         for to_spawn in self.to_spawn.drain() {
-            tokio_executor::spawn(to_spawn);
+            // We try to use the default executor, but fall back to polling the task manually if
+            // no executor is available. This makes it possible to use the core in environments
+            // outside of tokio.
+            let mut executor = tokio_executor::DefaultExecutor::current();
```
twittner: How about using the `futures::future::Executor` trait instead as an additional parameter type when constructing `HandledNodesTasks`? You could provide an implementation that loops over a set of futures or else users could plug in their own alternatives, including using a tokio executor.

twittner: Although it would be difficult to only use the small `Executor` interface. It can not accommodate the manual looping you are doing here I guess.

tomaka: We discussed this together, and concluded that it's a change orthogonal to this PR.

PR #884, opened 2019-01-22
Simplify the implementation of Ping

tomaka:
Fix #828

PR #979, opened 2019-02-27
Ping wasn't working anymore

tomaka:
#884 simplified ping too much by removing the system that was actually periodically pinging nodes.
This reintroduces a handler that wraps around the `OneShotHandler` and adds back the mechanism.

### DIFF
--- a/protocols/ping/src/handler.rs
+++ b/protocols/ping/src/handler.rs
@@ -47,7 +47,7 @@ where
     TSubstream: AsyncRead + AsyncWrite,
 {
     /// Builds a new `PingHandler`.
-    pub fn new() -> PingHandler<TSubstream> {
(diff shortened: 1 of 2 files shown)


protocols/ping/src/handler.rs
```diff
@@ -0,0 +1,135 @@
...
+    /// When the next ping triggers.
+    next_ping: Delay,
+}
+
+impl<TSubstream> PingHandler<TSubstream>
+where
+    TSubstream: AsyncRead + AsyncWrite,
+{
+    /// Builds a new `PingHandler`.
+    pub fn new() -> PingHandler<TSubstream> {
```
twittner:
```suggestion
    pub fn new() -> Self {
```

PR #980, opened 2019-02-27
Rewrite the WebCrypto ECDH using wasm-bindgen

tomaka:
It was untested and not working anyway. We don't lose anything by rewriting it.
This PR is tested and works against the Rust secio implementation.

cc #23


tomaka:
Note that while it is not great to assume that we are in a browser environment when compiling for WASM, it is intended that we eventually merge `impl_ring.rs` and `impl_webcrypto.rs` into one code that uses a library that successfully compiles for WASM.

protocols/secio/src/exchange/impl_webcrypto.rs
```diff
@@ -22,115 +22,163 @@
...
+            // `tmp_view`. This is guaranteed by the fact that we clone this array right below.
+            let tmp_view = unsafe { js_sys::Uint8Array::view(other_public_key) };
+            js_sys::Uint8Array::new(tmp_view.as_ref())
+        };
 
-    let send = move |out: Reference| {
-        let _ = tx.take()
-            .expect("JavaScript promise has been resolved twice")       // TODO: prove
-            .send(out);
+        // Note: contrary to what one might think, we shouldn't add the "deriveBits" usage.
```
twittner: Could you explain that bit?

tomaka: The `importKey` function takes a list of "usages": https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/importKey
For me it was obvious that one should pass the "deriveBits" usage, but if I remember correctly in practice both Firefox and Chrome produce an error if I do that.

protocols/secio/src/exchange/impl_webcrypto.rs
```diff
@@ -22,115 +22,163 @@
...
+use send_wrapper::SendWrapper;
+use std::io;
+use wasm_bindgen::prelude::*;
+
+/// Opaque private key type. Contains the private key and the `SubtleCrypto` object.
+pub type AgreementPrivateKey = SendSyncHack<(JsValue, web_sys::SubtleCrypto)>;
+
+/// We use a `SendWrapper` from the `send_wrapper` crate around our JS data type. JavaScript data
+/// types are not `Send`/`Sync`, but since WASM is single-threaded we know that we're only ever
+/// going to access them from the same thread.
```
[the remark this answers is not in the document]
tomaka: Well, I don't know how to solve that and suggest opening an issue instead of blocking this PR. It is not a safety problem, only a panicking problem.

protocols/secio/src/exchange/impl_webcrypto.rs
```diff
@@ -22,115 +22,163 @@
...
-    let other_public_key = TypedArray::from(other_public_key).buffer();
-
-    let out_size = out_size as u32;
+    let (private_key, crypto) = key.0.take();
+
+    // We start by importing the remote's public key into the WebCrypto world.
+    let import_promise = {
+        let other_public_key = {
+            // This unsafe is here because the lifetime of `other_public_key` must not outlive the
+            // `tmp_view`. This is guaranteed by the fact that we clone this array right below.
```
tomaka:
```suggestion
            // `tmp_view`. This is guaranteed by the fact that we clone this array right below.
            // See also https://github.com/rustwasm/wasm-bindgen/issues/1303
```

PR #987, opened 2019-02-28
Use bounded channels in transport

### LINKED ISSUE #973: Replace unbounded channels with bounded ones
We have two usages of `mpsc::UnboudedSender`/`mpsc::UnboundedReceiver`. In order to avoid possible issues with the buffer growing forever, it would be nice to replace them with bounded channels (ie. `mpsc::Sender`/`mpsc::Receiver`).

mattrutherford:
Partially addresses issues #973. 

- [x] Replace unbounded channels in `transport::memory`
- [ ] ~~Replace unbounded channels in `nodes::handled_node_tasks`~~

To replace unbounded channels in `handled_node_tasks` would have implications that warrant further consideration, recommend to revisit at a later time.

### DIFF
--- a/core/src/transport/memory.rs
+++ b/core/src/transport/memory.rs
@@ -29,19 +29,50 @@ use rw_stream_sink::RwStreamSink;
 use std::{collections::hash_map::Entry, error, fmt, io, num::NonZeroU64};
 
 lazy_static! {
-    static ref HUB: Mutex<FnvHashMap<NonZeroU64, mpsc::UnboundedSender<Channel<Bytes>>>> = Mutex::new(FnvHashMap::default());
+    static ref HUB: Mutex<FnvHashMap<NonZeroU64, mpsc::Sender<Channel<Bytes>>>> = Mutex::new(FnvHashMap::default());
 }
+
+static CHAN_BUF_SIZE: usize = 1000;
 
 /// Transport that supports `/memory/N` multiaddresses.
 #[derive(Debug, Copy, Clone, Default)]
 pub struct MemoryTransport;
+
+pub struct DialFuture {
+    sender: mpsc::Sender<Channel<Bytes>>,
+    channel_to_send: Option<Channel<Bytes>>,
+    channel_to_return: Option<Channel<Bytes>>,
+}
+
+impl Future for DialFuture {
+    type Item = Channel<Bytes>;
+    type Error = MemoryTransportError;
+
+    fn poll(&mut self) -> Poll<Self::Item, Self::Error> {
+        if let Some(c) = self.channel_to_send.take() {
+            match self.sender.start_send(c) {
+                Err(_) => return Err(MemoryTransportError::Unreachable),
+                Ok(AsyncSink::NotReady(t)) => {
+                    self.channel_to_send = Some(t);
+                    return Ok(Async::NotReady)
+                },
+                _ => (),
+            }
+        }
+        match self.sender.poll_complete() {
+            Err(_) => Err(MemoryTransportError::Unreachable),
+            Ok(Async::NotReady) => Ok(Async::NotReady),
+            Ok(Async::Ready(_)) => Ok(Async::Ready(self.channel_to_return.take().unwrap())),
+        }
+    }
+}
 
 impl Transport for MemoryTransport {
     type Output = Channel<Bytes>;
     type Error = MemoryTransportError;
     type Listener = Listener;
     type ListenerUpgrade = FutureResult<Self::Output, Self::Error>;
-    type Dial = FutureResult<Self::Output, Self::Error>;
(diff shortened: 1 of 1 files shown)


core/src/transport/memory.rs
```diff
@@ -29,19 +29,50 @@ use rw_stream_sink::RwStreamSink;
 use std::{collections::hash_map::Entry, error, fmt, io, num::NonZeroU64};
 
 lazy_static! {
-    static ref HUB: Mutex<FnvHashMap<NonZeroU64, mpsc::UnboundedSender<Channel<Bytes>>>> = Mutex::new(FnvHashMap::default());
+    static ref HUB: Mutex<FnvHashMap<NonZeroU64, mpsc::Sender<Channel<Bytes>>>> = Mutex::new(FnvHashMap::default());
 }
 
+static CHAN_BUF_SIZE: usize = 1000;
```
tomaka: This doesn't need to be a constant. Inlining the value in the source code is enough. Also, 1000 is not a great value.

core/src/transport/memory.rs
```diff
@@ -68,7 +99,7 @@ impl Transport for MemoryTransport {
 
         let actual_addr = Protocol::Memory(port.get()).into();
 
-        let (tx, rx) = mpsc::unbounded();
+        let (tx, rx) = mpsc::channel(CHAN_BUF_SIZE);
```
tomaka: I think `2` is a good value. It's the number of maximum dialing connections that haven't been processed by the listener yet.
```suggestion
        let (tx, rx) = mpsc::channel(2);
```

core/src/transport/memory.rs
```diff
@@ -94,20 +125,18 @@ impl Transport for MemoryTransport {
...
-            let (a_tx, a_rx) = mpsc::unbounded();
-            let (b_tx, b_rx) = mpsc::unbounded();
-            let a = RwStreamSink::new(Chan { incoming: a_rx, outgoing: b_tx });
-            let b = RwStreamSink::new(Chan { incoming: b_rx, outgoing: a_tx });
-            if tx.unbounded_send(b).is_err() {
-                return Err(TransportError::Other(MemoryTransportError::Unreachable));
-            }
-            a
+        if let Some(sender) = hub.get(&port) {
+            let (a_tx, a_rx) = mpsc::channel(CHAN_BUF_SIZE);
```
tomaka: I'd say a value of `4096` for these two channels is good. It's the number of bytes that are pending.

core/src/transport/memory.rs
```diff
@@ -29,19 +29,50 @@ use rw_stream_sink::RwStreamSink;
...
+            match self.sender.start_send(c) {
+                Err(_) => return Err(MemoryTransportError::Unreachable),
+                Ok(AsyncSink::NotReady(t)) => {
+                    self.channel_to_send = Some(t);
+                    return Ok(Async::NotReady)
+                },
+                _ => (),
+            }
+        }
+        match self.sender.poll_complete() {
```
tomaka:
```suggestion
        match self.sender.close() {
```

`closed` automatically implies `poll_complete`.

core/src/transport/memory.rs
```diff
@@ -29,19 +29,50 @@ use rw_stream_sink::RwStreamSink;
...
+            Err(_) => Err(MemoryTransportError::Unreachable),
+            Ok(Async::NotReady) => Ok(Async::NotReady),
+            Ok(Async::Ready(_)) => Ok(Async::Ready(self.channel_to_return.take().unwrap())),
```
tomaka: We generally forbid `unwrap()` outside of tests in favour of a call to `expect()` that contains a message or a proof.
```suggestion
            Ok(Async::Ready(_)) => Ok(Async::Ready(self.channel_to_return.take()
                .expect("Future has been polled again after it is finished"))),
```

core/src/transport/memory.rs
```diff
@@ -29,19 +29,50 @@ use rw_stream_sink::RwStreamSink;
...
+    static ref HUB: Mutex<FnvHashMap<NonZeroU64, mpsc::Sender<Channel<Bytes>>>> = Mutex::new(FnvHashMap::default());
 }
 
+static CHAN_BUF_SIZE: usize = 1000;
+
 /// Transport that supports `/memory/N` multiaddresses.
 #[derive(Debug, Copy, Clone, Default)]
 pub struct MemoryTransport;
 
+pub struct DialFuture {
```
tomaka: Needs a small doc-comment. Nothing big, just *something*.
```suggestion
/// Connection to a `MemoryTransport` currently being opened.
pub struct DialFuture {
```

mattrutherford:
Waiting for https://github.com/libp2p/rust-libp2p/pull/992 to get merged

PR #989, opened 2019-03-01
Fix error being ignored in RwSinkStream

tomaka:
Right now, `RwSinkStream::read` tries to pull as many items as possible from the underlying stream.
Unfortunately, if an error happens while we have already read some data, the current implementation returns `Ok(num_bytes_read)` instead of returning the error.

This PR changes the code `RwSinkStream` to:
- Be way more simple, and thus with way fewer possible sources of issues.
- Stop reading after the current item is done, instead of pulling the following one. It will still skip over empty items so as to not accidentally return `Ok(0)` when there are more items remaining.

### DIFF
--- a/misc/rw-stream-sink/src/lib.rs
+++ b/misc/rw-stream-sink/src/lib.rs
@@ -63,61 +63,26 @@ where
     S::Item: IntoBuf,
 {
     fn read(&mut self, buf: &mut [u8]) -> Result<usize, IoError> {
-        let mut written = 0;
-
-        loop {
-            let need_new_item = if let Some(ref i) = self.current_item {
-                !i.has_remaining()
-            } else {
-                true
-            };
-
-            if need_new_item {
-                loop {
-                    self.current_item = match self.inner.poll() {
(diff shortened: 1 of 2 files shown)


misc/rw-stream-sink/src/lib.rs
```diff
@@ -63,61 +63,26 @@ where
...
+                Async::Ready(Some(i)) => i.into_buf(),
+                Async::Ready(None) => return Ok(0),     // EOF
+                Async::NotReady => return Err(IoErrorKind::WouldBlock.into()),
+            });
+        };
+
+        // Copy it!
+        debug_assert!(item_to_copy.has_remaining());
+        let to_copy = cmp::min(buf.len(), item_to_copy.remaining());
+        item_to_copy.by_ref().take(to_copy).copy_to_slice(&mut buf[..to_copy]);
```
twittner:
```suggestion
        item_to_copy.take(to_copy).copy_to_slice(&mut buf[..to_copy]);
```

PR #996, opened 2019-03-11
Rewrite the Kademlia k-buckets to be more explicit

tomaka:
Right now the API of k-buckets is pretty much magic. There are tons of corner cases that aren't covered.
This PR makes the API clearer so that the behaviour behaves properly.

Work-in-progress because this needs a review pass from me, and because more tests are needed.

Also we probably want to make time management from the k-buckets part of the API, for better testability and for WASM.

### DIFF
--- a/protocols/kad/src/kbucket.rs
+++ b/protocols/kad/src/kbucket.rs
@@ -29,6 +29,7 @@ use arrayvec::ArrayVec;
 use bigint::U512;
 use libp2p_core::PeerId;
 use multihash::Multihash;
+use std::num::NonZeroUsize;
 use std::slice::IterMut as SliceIterMut;
 use std::time::{Duration, Instant};
 use std::vec::IntoIter as VecIntoIter;
@@ -96,9 +97,7 @@ pub trait KBucketsPeerId<TOther = Self>: PartialEq<TOther> {
 
     /// Returns then number of bits that are necessary to store the distance between peer IDs.
     /// Used for pre-allocations.
-    ///
-    /// > **Note**: Returning 0 would lead to a panic.
-    fn max_distance() -> usize;
+    fn max_distance() -> NonZeroUsize;
 }
 
 impl KBucketsPeerId for PeerId {
@@ -106,7 +105,7 @@ impl KBucketsPeerId for PeerId {
         <Multihash as KBucketsPeerId<Multihash>>::distance_with(self.as_ref(), other.as_ref())
     }
 
-    fn max_distance() -> usize {
+    fn max_distance() -> NonZeroUsize {
         <Multihash as KBucketsPeerId>::max_distance()
     }
 }
@@ -116,7 +115,7 @@ impl KBucketsPeerId<PeerId> for Multihash {
         <Multihash as KBucketsPeerId<Multihash>>::distance_with(self, other.as_ref())
     }
 
-    fn max_distance() -> usize {
+    fn max_distance() -> NonZeroUsize {
         <PeerId as KBucketsPeerId>::max_distance()
     }
 }
@@ -131,8 +130,8 @@ impl KBucketsPeerId for Multihash {
         512 - xor.leading_zeros()
     }
 
-    fn max_distance() -> usize {
-        512
+    fn max_distance() -> NonZeroUsize {
+        NonZeroUsize::new(512).expect("512 is not zero; QED")
     }
 }
 
@@ -145,8 +144,10 @@ where
         A::distance_with(&self.0, &other.0) + B::distance_with(&self.1, &other.1)
     }
 
-    fn max_distance() -> usize {
-        <A as KBucketsPeerId<A>>::max_distance() + <B as KBucketsPeerId<B>>::max_distance()
+    fn max_distance() -> NonZeroUsize {
+        let n = <A as KBucketsPeerId<A>>::max_distance().get()
+            .saturating_add(<B as KBucketsPeerId<B>>::max_distance().get());
+        NonZeroUsize::new(n).expect("Saturating-add of two non-zeros can't be zero; QED")
     }
 }
 
@@ -158,7 +159,7 @@ where
         T::distance_with(*self, *other)
     }
 
-    fn max_distance() -> usize {
+    fn max_distance() -> NonZeroUsize {
         <T as KBucketsPeerId>::max_distance()
     }
 }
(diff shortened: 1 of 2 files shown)


tomaka:
Forgot to mention that this was based on top of #994. Only the last commit is relevant to this PR.

tomaka:
Ready for review!

I also added a `KBucketAdded` event that notifies when we insert a new value in a bucket.

protocols/kad/src/kbucket.rs
```diff
@@ -161,254 +176,502 @@ where
...
+        // Update the pending node state.
+        // TODO: must be reported to the user somehow, in a non-annoying API
+        if let Some(pending) = self.tables[bucket_num].pending_node.take() {
+            if pending.replace < Instant::now() {
+                let table = &mut self.tables[bucket_num];
+                let first_connected_pos = &mut table.first_connected_pos;
+                // If all the nodes in the bucket are connected, then there shouldn't be any
+                // pending node.
+                debug_assert!(*first_connected_pos >= 1);
+                let _ = table.nodes.remove(0);
```
twittner:
```suggestion
                table.nodes.remove(0);
```

protocols/kad/src/kbucket.rs
```diff
@@ -161,254 +176,502 @@ where
...
+
+impl<'a, TPeerId, TVal> EntryNotInKbucket<'a, TPeerId, TVal>
+where
+    TPeerId: KBucketsPeerId + Clone,
+{
+    /// Inserts the node as connected, if possible.
+    pub fn insert_connected(self, value: TVal) -> InsertOutcome<TPeerId> {
+        let table = {
+            let num = self.parent.bucket_num(&self.peer_id)
+                .expect("we can only build a EntryInKbucketDisc if we know of a bucket; QED");
```
twittner:
```suggestion
                .expect("we can only build a EntryNotInKbucket if we know of a bucket; QED");
```

protocols/kad/src/kbucket.rs
```diff
@@ -161,254 +176,502 @@ where
...
-                intermediate.insert(pos, self.my_id.clone());
-            }
+    /// Inserts the node as disconnected, if possible.
+    ///
+    /// > **Note**: This function will never return `Pending`. If the bucket is full, we simply
+    /// >           do nothing.
+    pub fn insert_disconnected(self, value: TVal) -> InsertOutcome<TPeerId> {
+        let table = {
+            let num = self.parent.bucket_num(&self.peer_id)
+                .expect("we can only build a EntryInKbucketDisc if we know of a bucket; QED");
```
twittner:
```suggestion
                .expect("we can only build a EntryNotInKbucket if we know of a bucket; QED");
```

protocols/kad/src/kbucket.rs
```diff
@@ -161,254 +176,502 @@ where
...
-        table.flush(self.unresponsive_timeout);
+impl<'a, TPeerId, TVal> EntryInKbucketDiscPending<'a, TPeerId, TVal>
+where
+    TPeerId: KBucketsPeerId + Clone,
+{
+    /// Returns the value associated to the entry in the bucket.
+    pub fn value(&mut self) -> &mut TVal {
+        let table = {
+            let num = self.parent.bucket_num(&self.peer_id)
+                .expect("we can only build a EntryInKbucketPending if we know of a bucket; QED");
```
twittner:
```suggestion
                .expect("we can only build a EntryInKbucketDiscPending if we know of a bucket; QED");
```

protocols/kad/src/kbucket.rs
```diff
@@ -161,254 +176,502 @@ where
...
-                let elem = table.nodes.remove(pos);
-                table.first_connected_pos -= 1;
-                table.nodes.insert(table.first_connected_pos, elem);
-            }
-            table.latest_update = Instant::now();
-            Update::Updated
+        assert!(table.pending_node.as_ref().map(|n| &n.node.id) == Some(self.peer_id));
+        &mut table.pending_node
+            .as_mut()
+            .expect("we can only build a EntryInKbucketPending if the node is pending; QED")
```
twittner:
```suggestion
            .expect("we can only build a EntryInKbucketDiscPending if the node is pending; QED")
```

protocols/kad/src/kbucket.rs
```diff
@@ -161,254 +176,502 @@ where
...
-            // to the oldest node.
-            let pending_node = Node {
-                id: id.clone(),
-                value: Default::default(),
+    /// Reports that we are now connected to the given node.
+    pub fn set_connected(self) -> EntryInKbucketConnPending<'a, TPeerId, TVal> {
+        {
+            let table = {
+                let num = self.parent.bucket_num(&self.peer_id)
+                    .expect("we can only build a EntryInKbucketConnPending if we know of a bucket; QED");
```
twittner:
```suggestion
                    .expect("we can only build a EntryInKbucketDiscPending if we know of a bucket; QED");
```

protocols/kad/src/kbucket.rs
```diff
@@ -161,254 +176,502 @@ where
...
+                &mut self.parent.tables[num]
             };
-            table.pending_node = Some((pending_node, Instant::now()));
-            Update::Pending(&table.nodes[0].id)
 
-        } else {
-            debug_assert!(table.first_connected_pos == 0 || table.pending_node.is_some());
-            Update::Discarded
+            let mut pending = table.pending_node.as_mut()
+                .expect("we can only build a EntryInKbucketConnPending if there's a pending node; QED");
```
twittner:
```suggestion
                .expect("we can only build a EntryInKbucketDiscPending if there's a pending node; QED");
```

protocols/kad/src/kbucket.rs
```diff
@@ -161,254 +176,502 @@ where
...
+    ///
+    /// Pending nodes are ignored.
+    pub fn find_closest(&mut self, id: &impl KBucketsPeerId<TPeerId>) -> VecIntoIter<TPeerId> {
+        // TODO: optimize
+        let mut out = Vec::new();
+        for table in self.tables.iter_mut() {
+            for node in table.nodes.iter() {
+                out.push(node.id.clone());
+            }
+            // TODO: that's kind of wrong
```
twittner: Could you be more specific?

tomaka: We need the same logic here as in the `entry` method. The code here works but doesn't have the same behaviour, which is a bit stupid.

In practice though, we need a way to report when a pending node replaces a valid node (there's another TODO for that), meaning that we need to rework this anyway.

twittner: Maybe this should be mentioned in the comment?

tomaka: Done

tomaka:
We discussed this a bit in private.
Whereas I don't trust the code that it's bug free, I have high confidence that this PR is a huge improvement over the current situation.
There's an `unimplemented!` in `behaviour.rs` which I intend to fix before we publish 0.6.

PR #970, opened 2019-02-25
Experiment with compiling for WASM

tomaka:
Quick and dirty work-in-progress PR about trying to add a `wasm-test` crate in `misc` that compiles to WASM.

Not ready and not working.

Issue #558, opened 2018-10-11
UPnP in the TcpConfig


tomaka:
Add `attempt_upnp: bool` in the `TcpConfig`, and implement it!

tomaka:
Some references:
- http://www.upnp.org/specs/gw/UPnP-gw-WANIPConnection-v2-Service.pdf
- http://www.upnp-hacks.org/igd.html
- https://github.com/sbstp/rust-igd

One issue is that `listen_on` is only allowed to produce one listened address, while in practice if we traverse a NAT we are actually listening on two addresses: the one on the LAN and the on on the WAN.
Maybe `listen_on` should return an `Iterator<Item = Multiaddr>` instead of a `Multiaddr`.

twittner:
How about [PCP](https://en.wikipedia.org/wiki/Port_Control_Protocol) instead of IGD?

tomaka:
Whatever works!

dvdplm:
@twittner I couldn't find any rust lib that implements PCP – do you know of any?

Isan-Rivkin:
@dvdplm 
*Maybe* this one [pcp_mmv-rs](https://github.com/saurvs/pcp_mmv-rs)
Or you could maybe use FFI 
- [pcp](https://github.com/libpcp/pcp)
- [repcpc](https://github.com/alfredh/repcpc)

twittner:
@dvdplm: No, I do not. FYI, I have started a small UPNP-IGD implementation at https://github.com/paritytech/upnp-igdp but it is currently using a fork of tokio and I am reluctant to test it, given the security implications of opening port mappings available to the outside world.

dvdplm:
> fork of tokio

Curious: why did you have to fork it?

tomaka:
> Curious: why did you have to fork it?

https://github.com/tokio-rs/tokio/pull/710

tomaka:
I'm kind of conflicted whether this is a good "getting-started" issue.

The problem we encountered is that opening on the router is not instantaneous.
When one calls `listen_on` on `TcpConfig`, it needs to open ports. However the `listen_on` method is expected to immediately return the listened address, which depends on opening ports.

This also raises the question of making `listen_on` report known external addresses, for example the IP addresses of the local network interfaces.

thomaseizinger:
Closed as stale.

PR #1004, opened 2019-03-17
Add libp2p-js-transport

tomaka:
Extracted from #970 

Add a `libp2p-js-transport` crate that uses `wasm-bindgen`. It allows passing a `JsValue` referencing a JavaScript object that implements [`interface-transport`](https://github.com/libp2p/interface-transport), and makes it possible to use it from Rust as any other transport.

Once that is merged, we should be able to remove the JavaScript code from `libp2p-websocket`.

This is a WIP because there is one main issue in the API: our Rust transport, when listening, expects the transport to immediately yield the new listened address. The Js transport, however, is asynchronous. I don't know how to fix that except by changing either our of their API. However we probably have to change our API for #558 and this would probably be the way to go.


tomaka:
> This is a WIP because there is one main issue in the API: our Rust transport, when listening, expects the transport to immediately yield the new listened address. The Js transport, however, is asynchronous. I don't know how to fix that except by changing either our of their API. However we probably have to change our API for #558 and this would probably be the way to go.

Correction: it seems that they call the callback immediately. However this can still be a list of addresses instead of a single one.

tomaka:
I seem to be getting errors with:
> pull-stream called the reader callback multiple times in a row

I don't really know what that implies.

Issue #46, opened 2017-11-24
Add a transport capability system


tomaka:
A transport protocol like UDP doesn't guarantee reliability or ordering, but some upper layer protocols may require that the transport supports these characteristics.

Therefore it wouldn't be a bad idea to add methods to the `Transport` trait so that upper layers can check for these capabilities.

Vurich:
The only problem I can see with this is it makes writing new `Transport`s a PITA, since you have to manually write which capabilities the type supports. This is true whether we do it in the type system or at runtime.

tomaka:
I can see only two capabilities:

- Reliability ; are all packets guaranteed to arrive to destination?
- Ordering ; are packets arriving in order?
- Congestion control ; can we get flooded with data? (not sure whether this capability is useful to know though)

Examples: TCP and UTP would provide all three. UDP would provide none. Multiplex would remove the congestion control capability.

tomaka:
One thing to note is that the `Transport` trait is currently not adapted to UDP.
The socket type that the trait implementation must provide must implement `AsyncRead + AsyncWrite`, which considers that data acts as a contiguous stream. In other words, it is not adapted to UDP.

A fix would be to replace the `AsyncRead + AsyncWrite` requirements with `Stream + Sink`. The TCP implementation would be free to split the packets of the `Sink` into multiple actual packets. However doing this properly would require support for HKTs in Rust, because we don't know the type of the items to use for the `Stream` and `Sink`.

Vurich:
Maybe we can have `StreamTransport`, `PacketTransport` and `ReliablePacketTransport`/`OrderedPacketTransport` (the latter two being marker traits extending `PacketTransport`). We can `impl PacketTransport for T where T: StreamTransport` treating the whole stream as one huge packet, so most implementers just need to implement `StreamTransport` and don't need to care about reliability/order.

Vurich:
How about this design:

```rust
pub trait StreamTransport: Transport where Self::Output: AsyncRead + AsyncWrite {}
pub trait PacketTransport: Transport where Self::Output: Stream {}
pub trait ReliablePacketTransport: PacketTransport where Self::Output: Stream {} // Need to reiterate bounds
pub trait OrderedPacketTransport: PacketTransport where Self::Output: Stream {} // Need to reiterate bounds
```

Then you just implement the correct markers for you

tomaka:
[...]

PR #1005, opened 2019-03-18
Documentation updates

romanb:
Documentation updates:
  
  1. libp2p: Update the top-level module documentation, already including  intra-rustdoc links, removing outdated documentation (e.g. for topologies), updating examples and polishing the text.

  2. libp2p-core: Update the transport documentation to clarify that a `Transport` is really an abstraction only for connection-oriented communication.

Regarding the second point, I have read through some prior related discussion on that matter in https://github.com/libp2p/rust-libp2p/issues/46. Here is my reasoning for the direction I've taken:

  * The existing documentation and code mentions "connections" in many places.
  * The concept of "dialing" makes no sense for a connection-less transport.
  * Pretty much all upgrades work on connection-oriented transports.
  * I really think it is not a good idea to try to fit connection-oriented and connection-less transports under the same abstraction.

The intra-rustdoc links do no harm on stable but already render nicely on nightly. The only situation where the links don't work yet are for cross-crate inlined documentation, but that is still going to happen (See https://github.com/rust-lang/rust/issues/43466).


core/src/transport/mod.rs
```diff
@@ -18,13 +18,12 @@
...
 
-//! Handles entering a connection with a peer.
+//! Connection-oriented communication channels.
 //!
-//! The main element of this module is the `Transport` trait. It is implemented on objects that
-//! allow dialing and listening.
-//!
-//! The rest of the module holds combinators that allow tweaking an implementation of `Transport`,
-//! combine multiple transports together, or combine a transport with an upgrade.
+//! The main entity of this module is the [`Transport`] trait, which provides an
```
tomaka: Does this automatically create a link to `Transport`? I know that such a feature was planned, but I don't know whether it is stable.

core/src/transport/mod.rs
```diff
@@ -47,73 +46,85 @@ pub use self::choice::OrTransport;
...
+/// whose connection setup incorporates all earlier upgrades followed by the new upgrade,
+/// i.e. the order of the upgrades is significant.
 ///
 /// > **Note**: The methods of this trait use `self` and not `&self` or `&mut self`. In other
 /// >           words, listening or dialing consumes the transport object. This has been designed
 /// >           so that you would implement this trait on `&Foo` or `&mut Foo` instead of directly
 /// >           on `Foo`.
 pub trait Transport {
-    /// The raw connection to a peer.
+    /// A data stream (i.e. connection) created by the transport.
```
tomaka: This is not necessarily a data stream.
Once you add layers on top of a `Transport`, it can produce something else. Most notably, the output of the transport that you pass to the `Swarm` must be `(PeerId, impl StreamMuxer)`.

core/src/transport/mod.rs
```diff
@@ -47,73 +46,85 @@ pub use self::choice::OrTransport;
 pub use self::memory::MemoryTransport;
 pub use self::upgrade::Upgrade;
 
-/// A transport is an object that can be used to produce connections by listening or dialing a
-/// peer.
+/// A transport is a connection-oriented communication channel between two peers,
```
tomaka: The transport itself is not a channel, but allows creating channels. I think this should be worded a bit differently.

core/src/transport/mod.rs
```diff
@@ -47,73 +46,85 @@ pub use self::choice::OrTransport;
...
     type Error: error::Error;
 
-    /// The listener produces incoming connections.
+    /// A stream of incoming connections.
     ///
     /// An item should be produced whenever a connection is received at the lowest level of the
-    /// transport stack. The item is a `Future` that is signalled once some pre-processing has
-    /// taken place, and that connection has been upgraded to the wanted protocols.
+    /// transport stack. The item is a [`Future`] that resolves to a connection once all
+    /// protocol upgrades have been applied.
```
tomaka: The protocol upgrades are applied on the `ListenerUpgrade`. May need better wording.

core/src/transport/mod.rs
```diff
@@ -47,73 +46,85 @@ pub use self::choice::OrTransport;
...
-    /// A future which indicates that we are currently dialing to a peer.
+    /// A pending, outbound connection, obtained through dialing on the transport.
     type Dial: Future<Item = Self::Output, Error = Self::Error>;
 
-    /// Listen on the given multiaddr. Returns a stream of incoming connections, plus a modified
-    /// version of the `Multiaddr`. This new `Multiaddr` is the one that that should be advertised
-    /// to other nodes, instead of the one passed as parameter.
+    /// Listens on the given [`Multiaddr`], producing a stream of pending, inbound connections.
     ///
-    /// > **Note**: The reason why we need to change the `Multiaddr` on success is to handle
```
tomaka: This style of putting `> ` in front of notes is what the documentation of the standard library does, so I followed it.

src/lib.rs
```diff
@@ -26,108 +26,115 @@
...
 //! ## Multiaddr
 //!
-//! A `Multiaddr` is a way to reach a node. Examples:
+//! A [`Multiaddr`] is a self-describing network address and protocol stack
+//! that is used to establish connections to peers. Some examples:
 //!
 //! * `/ip4/80.123.90.4/tcp/5432`
 //! * `/ip6/[::1]/udp/10560/quic`
-//! * `/unix//path/to/socket`
+//! * `/unix/path/to/socket`
```
tomaka: This extra `/` is actually on purpose. As far as I know, the specs of multiaddr don't specify a way to express these kind of addresses, but I remember seeing this as an example somewhere.

src/lib.rs
```diff
@@ -26,108 +26,115 @@
...
+//!
+//!   1. Creating a *base* implementation of [`Transport`] that is upgraded with
+//!      all the desired protocols, such as for transport security and multiplexing.
+//!   2. Creating a struct that implements the [`NetworkBehaviour`] trait and combines all the
+//!      desired network behaviours.
+//!   3. Instantiating a [`Swarm`] with the transport and network behaviour from the previous
+//!      steps.
 //!
-//! The most simple way to use libp2p consists in the following steps:
+//! The swarm instance can then be polled, e.g. with the [tokio] library, in order to
```
tomaka: Note that how we poll the swarm depends on the transport being used. If you use a `TcpConfig`, then you **must** use tokio, because the `TcpConfig` wraps around a `tokio_tcp::TcpStream`.

core/src/transport/mod.rs
romanb: Good point.

core/src/transport/mod.rs
romanb: True, I will try to reword it.

core/src/transport/mod.rs
romanb: Ok.

core/src/transport/mod.rs
romanb: Yes it does, but right now only when running `cargo doc` on nightly. It does not need a feature flag though. On stable these will just render as `[Foo]` instead of `Foo`, i.e. with brackets, but that seems harmless for the benefit of being forward-compatible. We could even generate the API docs using nightly before the RFC is stabilized, i.e. build the code with stable and only the docs with nightly to get the nicer API docs earlier.

romanb: FYI this type of link in particular is referred to as [Implied Shortcut Reference Links](https://github.com/rust-lang/rfcs/blob/master/text/1946-intra-rustdoc-links.md#implied-shortcut-reference-links).

src/lib.rs
romanb: Oh, good to know, thanks.

core/src/transport/mod.rs
romanb: New suggestion in https://github.com/libp2p/rust-libp2p/pull/1005/commits/3216bae891144c5118fa90de9423ab3b878ceb6b.

core/src/transport/mod.rs
romanb: New suggestion in https://github.com/libp2p/rust-libp2p/pull/1005/commits/3216bae891144c5118fa90de9423ab3b878ceb6b.

core/src/transport/mod.rs
romanb: New suggestion in https://github.com/libp2p/rust-libp2p/commit/3216bae891144c5118fa90de9423ab3b878ceb6b.

src/lib.rs
romanb: Good point, I will try to make that clearer.

romanb: Actually, I'm not sure I understand. Apart from being the obvious choice, why is it strictly necessary to use `tokio` to poll a swarm if the transport is based on a `tokio_tcp::TcpStream`?

tomaka: When you use a tokio runtime to poll, every time it polls a future tokio will define a thread-local-storage value containing a "reactor" (https://docs.rs/tokio-reactor). The reactor in this thread-local-storage is then picked up by the `tokio_tcp::TcpStream` to run itself. It is essentially an implicit parameter.

If you don't use a tokio runtime, then the `tokio_tcp::TcpStream` will produce an error because it can't find the reactor.

romanb: Got it, thanks. I will update the documentation to make that clear then before merging.

PR #1010, opened 2019-03-18
Add a Transport wrapper for nodes-hammering prevention

### LINKED ISSUE #1007: Add a Transport wrapper for nodes-hammering-prevention
Add a transport wrapper that keeps in memory addresses that the underlying transport fails to reach. When such an address is tried again, it adds a delay in the `Dial` future before actually transmitting the connection request to the underlying transport.

Can be added to the root `libp2p` crate, similar to the `BandwidthLogging` type.

tomaka:
Close #1007 

Adds a `HammerPrevention` type that memorizes addresses and adds a back-off delay when connecting to them.


src/hammer_prevent.rs
```diff
@@ -0,0 +1,243 @@
...
+                        },
+                        Err(err) => return Err(HammerPreventionErr::Timer(err)),
+                    }
+                }
+
+                HammerPreventionDialInner::Dialing(mut dial) => {
+                    match dial.poll() {
+                        Ok(Async::Ready(output)) => {
+                            // On success, remove the address from the list.
+                            let _ = self.addresses.lock().remove(&self.local_addr);
```
twittner:
```suggestion
                            self.addresses.lock().remove(&self.local_addr);
```

twittner (commented):
Reminds me of the circuit breaker pattern. The name "HammerPrevention" takes some getting used to. What is the advantage of adding a delay instead of returning an error immediately until some time has passed?

tomaka:
> What is the advantage of adding a delay instead of returning an error immediately until some time has passed?

We still want to try dialing the node again. If we were to return an error immediately, then it becomes the responsibility of the core to add this delay.

twittner:
In the circuit breaker pattern errors would only be returned for some (exponentially growing) time, after which the request is allowed to go through again. Cf. https://docs.microsoft.com/en-us/azure/architecture/patterns/circuit-breaker

tomaka:
I see. But we would still need a `Delay` somewhere, otherwise we will just have a spin loop where we error and try dialing again immediately.

twittner:
Yes, that is true.

twittner (approved):
LGTM (except for the name).

tomaka:
I started revisiting this.
There's a fundamental issue in the code (that was already there at the time): if we enforce a delay but then the `Transport` returns `MultiaddrNotSupported`, then we don't really know what to do with it.

We could do a dummy call to `self.inner.dial` ahead of time to check that the multiaddress is supported, but in this situation the wrapper is explicitly trying to prevent calls to `dial` from happening too often. This would totally defeat the point.

mxinden (commented):
Definitely in favor of the name change to something along the lines of *circuit breaking*.

thomaseizinger:
Closing as stale.

PR #1011, opened 2019-03-19
use upstream multiaddr crate

### LINKED ISSUE #758: Switch to upstream multiaddr crate

montekki:
Attempt to fix #758 and use upstream `multiaddr`. The `memory` addrs are obviously broken here since they are not a part of `multiaddr`. Probably there could be some extension wrapper implemented for upstream multiaddr to make things work in case we will continue extending the upstream with other protocol families


tomaka:
The intent of #758 is that we should port the changes that we have locally in the upstream repo.

PR #1013, opened 2019-03-19
Split address reach error and node reach error

tomaka:
The current API of `NetworkBehaviour` has some pretty big flaws:
- If we try to dial a node but we don't know any address for it, we ignore the problem and nothing happens (nothing is reported back).
- On dial failure, there's no way to know if the swarm is still trying to connect to a node through a different address or if it's no longer trying.
- Similarly, a dial failure kind of implies that we're not connected to the node, while it is possible, in the case of simulatenous dialing, that we report a dial failure *after* the connection.

This PR fixes that by splitting `inject_dial_failure` in two: `inject_dial_failure` only takes a `PeerId` and is called when we are no longer trying to dial, while `inject_addr_reach_failure` is called when we fail to reach a specific address. If this was the last address of a node that we failed to reach, then both methods are called.

### DIFF
--- a/core/src/swarm.rs
+++ b/core/src/swarm.rs
@@ -182,7 +182,9 @@ where TBehaviour: NetworkBehaviour,
         match me.raw_swarm.peer(peer_id.clone()) {
             raw_swarm::Peer::NotConnected(peer) => {
                 let handler = me.behaviour.new_handler().into_node_handler_builder();
-                let _ = peer.connect_iter(addrs, handler);
+                if peer.connect_iter(addrs, handler).is_err() {
+                    me.behaviour.inject_dial_failure(&peer_id);
+                }
             },
             raw_swarm::Peer::PendingConnect(mut peer) => {
                 peer.append_multiaddr_attempts(addrs)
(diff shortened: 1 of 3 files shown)


tomaka:
I went with `inject_dial_failure` because each `Dial` message you generate can generate a corresponding  event, either a `inject_dial_failure` or a `inject_connected`.

As for the order, it should be defined. To the opposite order (first a node reach error then an address reach error) doesn't really make sense, since reaching a node requires 0 to N address reach attempts.

core/src/swarm.rs
```diff
@@ -362,8 +367,17 @@ pub trait NetworkBehaviour {
         event: <<Self::ProtocolsHandler as IntoProtocolsHandler>::Handler as ProtocolsHandler>::OutEvent
     );
 
-    /// Indicates to the behaviour that we tried to reach a node, but failed.
-    fn inject_dial_failure(&mut self, _peer_id: Option<&PeerId>, _addr: &Multiaddr, _error: &dyn error::Error) {
+    /// Indicates to the behaviour that we tried to reach an address, but failed.
+    ///
+    /// If we were trying to reach a specific node, its ID is passed as parameter. If this is the
+    /// last address to attempt for the given node, then `inject_dial_failure` must be called
+    /// afterwards.
```
romanb: I suppose this documentation is for implementors of the trait, not for libp2p developers, so I'd rather write:

> [..], then `inject_dial_failure` is called afterwards.

tomaka:
> I went with inject_dial_failure because each Dial message you generate can generate a corresponding event, either a inject_dial_failure or a inject_connected.

What's your opinion on that?

romanb:
> > I went with inject_dial_failure because each Dial message you generate can generate a corresponding event, either a inject_dial_failure or a inject_connected.
> 
> What's your opinion on that?

Sounds ok to me.

PR #1014, opened 2019-03-20
Replace sha1 and tiny-keccak with RustCrypto crates

newpavlov:
You already use two hash crates from RustCrypto and it's a bit strange to use `sha1` and `tiny-keccak` instead of `sha-1` and `sha3`. Also migrating to them simplifies code a bit.

### DIFF
--- a/misc/multihash/src/lib.rs
+++ b/misc/multihash/src/lib.rs
@@ -10,44 +10,24 @@ mod hashes;
 
 use sha2::Digest;
 use std::fmt::Write;
-use tiny_keccak::Keccak;
 use unsigned_varint::{decode, encode};
 
 pub use self::errors::{DecodeError, DecodeOwnedError, EncodeError};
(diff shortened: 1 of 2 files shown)


misc/multihash/src/lib.rs
```diff
@@ -10,44 +10,24 @@ mod hashes;
...
-        let mut kec = Keccak::$constructor();
-        kec.update($input);
-        kec.finalize($output);
-    }};
-    (blake2, $algorithm:ident, $input:expr, $output:expr) => {{
-        let mut hasher = blake2::$algorithm::default();
-        hasher.input($input);
-        $output.copy_from_slice(hasher.result().as_ref());
-    }};
+// Helper function for encoding input into output using given `Digest`
```
tomaka:
```suggestion
/// Helper function for encoding input into output using given `Digest`
```

PR #1015, opened 2019-03-20
Allow configuring the maximum message length

### LINKED ISSUE #991: Hidden Message  Length Settings
hi,  just start playing with libp2p here. It seems there are some limit on the message length? 

I tried to code following the chat example and there seems a limit around 1974 bytes. (simply modify the chat example to send a lot of bytes can reproduce the issue.)

Looked into yamux or mplex or tcp settings for buffer size and seems no help changing values there. 

Is there anything i am missing here?

Thanks!

tomaka:
Fix #991 

Instead of a `PeerId`, `Floodsub::new` now accepts a `FloodsubConfig` struct that contains the maximum allowed size for a message.

This fix is however not ideal, because libp2p will automatically combine multiple messages into one without accounting for that limit. In other words, if the limit is 2048 and you send two messages of 1025 bytes at once, you may get an error even though it would be feasible to dispatch them. After this PR is merged, an issue should be opened about this.

### DIFF
--- a/protocols/floodsub/src/layer.rs
+++ b/protocols/floodsub/src/layer.rs
@@ -18,7 +18,7 @@
 // FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
 // DEALINGS IN THE SOFTWARE.
 
-use crate::protocol::{FloodsubConfig, FloodsubMessage, FloodsubRpc, FloodsubSubscription, FloodsubSubscriptionAction};
+use crate::protocol::{FloodsubListen, FloodsubMessage, FloodsubSend, FloodsubRpc, FloodsubSubscription, FloodsubSubscriptionAction};
 use crate::topic::{Topic, TopicHash};
 use cuckoofilter::CuckooFilter;
 use fnv::FnvHashSet;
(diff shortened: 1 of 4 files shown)


protocols/floodsub/src/layer.rs
```diff
@@ -60,12 +60,21 @@ pub struct Floodsub<TSubstream> {
     marker: PhantomData<TSubstream>,
 }
 
+/// Configuration for the floodsub behaviour.
+#[derive(Debug, Clone)]
+pub struct FloodsubConfig {
+    /// Identifier of the local node.
+    pub local_peer_id: PeerId,
+    /// Maximum number of bytes allowed in an RPC message.
+    pub max_message_size: usize,
```
twittner: Should these fields really be public? This could easily break clients in the future.

romanb:
> [..] because libp2p will automatically combine multiple messages into one without accounting for that limit.

Out of curiosity: Where in libp2p does that happen (and is it specific to pubsub)?

> [..] In other words, if the limit is 2048 and you send two messages of 1025 bytes at once, you may get an error even though it would be feasible to dispatch them. After this PR is merged, an issue should be opened about this.

Does it really make sense to add this configuration option if it doesn't work properly, i.e. shouldn't this problem be addressed in the context of (or before) adding this option? Can a user (developer) control when messages get "flushed" and are thus potentially subject to being combined? In the latter case, introducing the option seems fine if the behavior w.r.t. combining messages is documented.

AgeManning:
> Out of curiosity: Where in libp2p does that happen (and is it specific to floodsub)?

I'm also curious about this. I thought `read_one()` and `write_one()` would send length-prefixed message per stream. If two messages were sent of 1024 sizes, wouldn't two streams be created, length pre-fixed by 1024. When read (by `read_one()`) only the 1024 bytes would be read for each message?

tomaka:
Yes, it's specific to floodsub. Floodsub tries to combine as many "pubsub message" into one "network message" as possible. If the size of the network message goes over the limit, then writing will produce an error but floodsub isn't able to recover it.

romanb:
I started working on some `floodsub` updates, as there have been quite a few changes also to the specs in the past months, and this PR will then be superseded. I will close this PR once the new PR for the floodsub updates is open.

romanb:
This PR is probably superseded by https://github.com/libp2p/rust-libp2p/pull/1720#issuecomment-720376851.

mxinden:
With Gossipsub 1.0 and 1.1 I am closing here. Feel free to comment in case you would like to continue this effort.

PR #1019, opened 2019-03-21
Improve Debug and Display for PeerId

tomaka:


core/src/peer_id.rs
```diff
@@ -35,7 +35,15 @@ pub struct PeerId {
...
-        write!(f, "PeerId({})", self.to_base58())
+        f.debug_tuple("PeerId")
+            .field(&self.to_base58())
+            .finish()
+    }
+}
+
+impl fmt::Display for PeerId {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        fmt::Display::fmt(&self.to_base58(), f)
```
twittner:
```suggestion
        self.to_base58().fmt(f)
```

tomaka: Why is that better?

twittner: It looks less noisy to me and is shorter.

tomaka: But if we imported `Debug` and `Display` directly instead of through `fmt`, we'd get a compilation error.

twittner: Sure, but we did not.

PR #1024, opened 2019-03-24
Add a Toggle NetworkBehaviour

tomaka:
Allows having a disabled `NetworkBehaviour`.

This PR also adds implementations of `ProtocolsHandler` and `IntoProtocolsHandler` that can be disabled, but they remain "private" (as in, they can't be constructed).
This is because they accept events but will panic if they receive one, which can be a very unintuitive behaviour.


tomaka:
I've got a branch ready for mDNS support in Substrate after this PR is merged.

Issue #794, opened 2018-12-18
Figure out the design of nat_traversal and external addresses


tomaka:
For https://github.com/paritytech/substrate/issues/1193
cc #544 

The following problems are unresolved:

- Should `nat_traversal()` be implemented by `Transport`s, or maybe on `Multiaddress` instead?
- Since reported IPs can be malicious, should we add a TTL to external addresses? Use a more elaborate system than just storing them in a Vec?
- Does reporting a bad IP address have any consequence?

tomaka:
Also:

- In protocols like Kademlia and Identify, where nodes self-report their external addresses, do we need a system against malicious nodes that report wrong addresses or a large number of addresses?

twittner:
> Should `nat_traversal()` be implemented by `Transport`s, or maybe on `Multiaddress` instead?

It would make sense to me to not name this method `nat_traversal` but rather `merge_matching_prefix` and to try implementing it on `Multiaddr`.

As for real NAT traversal we would need more than reporting externally observed addresses. Many NATs perform address and port filtering which makes the address reported by some peer useless for other peers. This limitation may already produce many invalid addresses without peers acting maliciously. Some scoring algorithm could help filtering out those addresses.

> Does reporting a bad IP address have any consequence?

At least it would harm connectivity, i.e. it would take longer to establish connections to peers. Otherwise it depends on how a node keeps and uses reported addresses.

> Since reported IPs can be malicious, should we add a TTL to external addresses?

It would help to make explicit what kind of malicious behaviour we want to protect against. Is it only the reporting of invalid or unreachable addresses? Are we concerned about DDOS attacks or what other kinds of bad behaviour?

dvdplm:
My two cents on this is that it's wise to keep `Multiaddress` as dumb as we can and limited in scope to validate addresses only at the "grammar" level, keeping logic to determine what is a "good" or "bad" address elsewhere.
Real NAT traversal, like @twittner says, is sort of a black box for external observers: there's no way to tell which transforms were performed (or why) from the outside. It seems to me that the Transport is the layer that has the best chance of making educate guesses as to what a given behaviour means (and it will guess wrong quite often). I would imagine that it's going to be mainly the application layer that implements whatever logic it needs to decide when an address is "malicious" or not.

tomaka:
> It would make sense to me to not name this method nat_traversal but rather merge_matching_prefix and to try implementing it on Multiaddr.

[...]

PR #498, opened 2018-09-19
Refactor multiaddr crate.

twittner:
- ~~~Use `Bytes` instead of `Vec` as the internal representation of `Multiaddr`. This can help with frequent cloning or `to_bytes` calls.~~~

- Remove `AddrComponent`. Instead `Protocol` directly contains its associated data.

- Various smaller changes around conversions to Multiaddr from other types, e.g. socket addresses.

- Expand tests to include property tests which test encoding/decoding identity.

PR #1028, opened 2019-03-28
Return a sequence of `Multiaddr`s in `listen_on`.

twittner:
This is part of the proposed changes to the `Transport` trait as outlined in [[1]].

To allow transport implementations to listen on multiple addresses we now report all those addresses as a sequence of multi-addresses instead of supporting only a single one.

Instead of making `Transport::listen_on` return a generic iterator a concrete type `MultiaddrSeq` is introduced to support non-emptiness.

[1]: https://github.com/libp2p/rust-libp2p/issues/794#issuecomment-475950601

### DIFF
--- a/core/src/nodes/listeners.rs
+++ b/core/src/nodes/listeners.rs
@@ -20,7 +20,7 @@
 
 //! Manage listening on multiple multiaddresses at once.
 
-use crate::{Multiaddr, Transport, transport::TransportError};
+use crate::{Multiaddr, MultiaddrSeq, Transport, transport::TransportError};
 use futures::prelude::*;
 use std::{collections::VecDeque, fmt};
 use void::Void;
@@ -55,11 +55,11 @@ use void::Void;
 /// // The `listeners` will now generate events when polled.
 /// let future = listeners.for_each(move |event| {
 ///     match event {
-///         ListenersEvent::Closed { listen_addr, listener, result } => {
-///             println!("Listener {} has been closed: {:?}", listen_addr, result);
+///         ListenersEvent::Closed { listen_addrs, listener, result } => {
+///             println!("Listener {} has been closed: {:?}", listen_addrs, result);
 ///         },
-///         ListenersEvent::Incoming { upgrade, listen_addr, .. } => {
-///             println!("A connection has arrived on {}", listen_addr);
+///         ListenersEvent::Incoming { upgrade, listen_addrs, .. } => {
+///             println!("A connection has arrived on {}", listen_addrs);
 ///             // We don't do anything with the newly-opened connection, but in a real-life
 ///             // program you probably want to use it!
 ///             drop(upgrade);
@@ -90,8 +90,8 @@ where
 {
     /// The object that actually listens.
     listener: TTrans::Listener,
-    /// Address it is listening on.
-    address: Multiaddr,
+    /// Addresses it is listening on.
+    addresses: MultiaddrSeq
 }
 
 /// Event that can happen on the `ListenersStream`.
@@ -104,7 +104,7 @@ where
         /// The produced upgrade.
         upgrade: TTrans::ListenerUpgrade,
         /// Address of the listener which received the connection.
-        listen_addr: Multiaddr,
+        listen_addrs: MultiaddrSeq,
         /// Address used to send back data to the incoming client.
         send_back_addr: Multiaddr,
     },
@@ -112,7 +112,7 @@ where
     /// A listener has closed, either gracefully or with an error.
     Closed {
         /// Address of the listener which closed.
-        listen_addr: Multiaddr,
+        listen_addrs: MultiaddrSeq,
         /// The listener that closed.
         listener: TTrans::Listener,
         /// The error that happened. `Ok` if gracefully closed.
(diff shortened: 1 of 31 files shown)


tomaka:
Why use a `MultiaddrSeq` instead of an iterator?

twittner:
> Why use a `MultiaddrSeq` instead of an iterator?

From https://github.com/libp2p/rust-libp2p/pull/1028#issue-265400168:

> [...] to support non-emptiness.

By using a concrete type we get more specific semantics and a simpler implementation. Using a `Iterator + Clone` type does not buy us much here.

core/src/nodes/listeners.rs
```diff
@@ -104,15 +104,15 @@ where
         /// The produced upgrade.
         upgrade: TTrans::ListenerUpgrade,
         /// Address of the listener which received the connection.
-        listen_addr: Multiaddr,
+        listen_addrs: MultiaddrSeq,
```
tomaka: Well, that's clearly wrong. A connection only arrives on a single multiaddress. That means more changes are needed.

tomaka:
> to support non-emptiness.

Do we want to not support emptiness?

twittner:
Currently, `Transport::listen_on` returns a pair of `(Listener, Multiaddr)`. Note that this is `Multiaddr`, not `Option<Multiaddr>`. The proposal in [[1]] extends the `Multiaddr` to *"list of multiaddresses"* which I understood to mean *"one or more"* given that we currently require an address to be returned. If we want to support listeners without an (immediate) listen address why returning a list/iterator at all? The proposed stream event API would have to include for each upgrade the listener address anyway as we would otherwise not know which of the potentially many addresses the listener used when creating this event. So instead of the proposed API in [[1]], I would suggest we use the one below which only returns a `Listener` from `Transport::listen_on`. The address information is embedded in `ListenerEvent`. A listener may start without any address, but always has to say which listener address was used when creating an upgrade event. It may also later transition to not having a known address by generating `AddressExpired` events for all its currently known addresses, but eventually has to come back with one when the next upgrade is created.

```rust
pub trait Transport {
    type Output;
    type Error: error::Error;

    type Listener: Stream<Item = ListenerEvent<Self::ListenerUpgrade>, Error = Self::Error>;
    type ListenerUpgrade: Future<Item = Self::Output, Error = Self::Error>;
    type Dial: Future<Item = Self::Output, Error = Self::Error>;

    fn listen_on(self, addr: Multiaddr) -> Result<Self::Listener, TransportError<Self::Error>>;

    fn dial(self, addr: Multiaddr) -> Result<Self::Dial, TransportError<Self::Error>>;
}

enum ListenerEvent<T> {
    Upgrade {
        upgrade: T,
        listen_addr: Multiaddr,
        remote_addr: Multiaddr
    }
    AddressExpired(Multiaddr)
}
```

This may also be a good time to revisit the proposal in [[2]] which got rejected at the time, to use `Bytes` instead of `Vec` for the internal representation of `Multiaddr`.

[1]: https://github.com/libp2p/rust-libp2p/issues/794#issuecomment-475950601
[2]: https://github.com/libp2p/rust-libp2p/pull/498

tomaka:
> If we want to support listeners without an (immediate) listen address why returning a list/iterator at all?

The very pragmatic reason is that a lot tests create a transport or a swarm, make it listen on something, then dial the returned address.

I'm not sure which decision to take here. In most situations you know your listened addresses immediately, which is why it makes sense to return them immediately and not in the future.

tomaka:
> I would suggest we use the one below

We still need to know the addresses that we listen on before we receive a connection, otherwise we can't advertise this address to other nodes.

The typical process in Substrate is: we start listening, then we dial a bootstrap, then we indicate to this bootstrap node which addresses we're listening on, and this address is then advertised to the rest of the network.
If we don't have an event "StartedToListenOnAnAddress", then we're blocked and we will never receive any incoming connection.

twittner:
I have a branch (https://github.com/libp2p/rust-libp2p/compare/master...twittner:listener-event-wip) which contains a work in progress of the interface outlined [above](https://github.com/libp2p/rust-libp2p/pull/1028#issuecomment-478486546), except that `ListenerEvent` is defined as:

```rust
/// Event produced by [`Transport::Listener`]s
#[derive(Clone, Debug, PartialEq)]
pub enum ListenerEvent<T> {
    /// The transport is listening on a new additional [`Multiaddr`].
    NewAddress(Multiaddr),
    /// An upgrade, consisting of the upgrade future, the listener address and the remote address.
    Upgrade {
        /// The upgrade.
        upgrade: T,
        /// The listening address which produced this upgrade.
        listen_addr: Multiaddr,
        /// The remote address which produced this upgrade.
        remote_addr: Multiaddr
    },
    /// A [`Multiaddr`] is no longer used for listening.
    AddressExpired(Multiaddr)
}
```

I do not think we should also have `Transport::listen_on` return an iterator over multi-addresses only because tests are written in that way currently. We should rather be prepared to deal with transports that return no immediate listen address. In fact supporting those transports has quite an impact on the API:

1. We need a separate stable identifier to refer to individual listeners as the set of listen addresses is volatile and potentially empty.
2. `Swarm::listen_on` can no longer return a `Multiaddr`. Listen addresses become available only when polling the swarm.

Most tests can be adjusted with little effort, however I am unsure how to make listen addresses available to users of `Swarm`. `ListenersStream` and `RawSwarm` produce events so it is easy enough to poll those, but `Swarm` calls the various behaviour methods and only produces behaviour events when polled. It would seem to me that `Swarm::poll` needs to be changed to return a `SwarmEvent` instead of only behaviour events if we want to make the listen addresses available to code outside of behaviour implementations. What is your opinion?

tomaka:
> Most tests can be adjusted with little effort,

That was my main concern, but if they can be updated then that's good.

> It would seem to me that Swarm::poll needs to be changed to return a SwarmEvent instead of only behaviour events

That's obviously not great, as the `Swarm` at the moment is more or less just a wrapper around the `NetworkBehaviour` in terms of API.

twittner:
I am closing this PR and point to https://github.com/libp2p/rust-libp2p/pull/1032 which has been rebased to not be a continuation of this PR, but to implement the basic interface described above.

PR #668, opened 2018-11-20
Use libsecp256k1 for secio handshakes

dvdplm:
Experiment using `libsecp256k1` to get rid of plenty of `unsafe` and gain `no_std` capability.

Related to https://github.com/libp2p/rust-libp2p/issues/627

PR #1029, opened 2019-03-29
Replace `secp256k1` crate with `libsecp256k1`.

twittner:
Unfortunately we could not implement `AsRef<[u8]>` for `SecretKey` as the crate does not provide a means to do so.

Supersedes https://github.com/libp2p/rust-libp2p/pull/668 which became stale after https://github.com/libp2p/rust-libp2p/pull/972.


core/src/identity/secp256k1.rs
```diff
@@ -74,28 +68,30 @@ impl From<Keypair> for SecretKey {
...
 
     /// Create a secret key from a byte slice, zeroing the slice on success.
     /// If the bytes do not constitute a valid Secp256k1 secret key, an
     /// error is returned.
     pub fn from_bytes(mut sk: impl AsMut<[u8]>) -> Result<SecretKey, DecodingError> {
         let sk_bytes = sk.as_mut();
-        let secret = secp::key::SecretKey::from_slice(&*sk_bytes)
-            .map_err(|e| DecodingError::new("Secp256k1 secret key", e))?;
+        let secret = secp256k1::SecretKey::parse_slice(&*sk_bytes)
+            .map_err(|_| DecodingError::new("failed to parse secp256k1 secret key"))?;
```
romanb: I just realized that there is a missing `sk_bytes.zeroize();` below this line, i.e. the function currently does not do what the documentation says. Since this is an oversight unrelated to this PR, I created a separate, tiny PR for that: https://github.com/libp2p/rust-libp2p/pull/1040.

core/src/identity/secp256k1.rs
```diff
@@ -122,36 +118,38 @@ impl SecretKey {
     /// ECDSA signature, as defined in [RFC3278].
     ///
     /// [RFC3278]: https://tools.ietf.org/html/rfc3278#section-8.2
-    pub fn sign(&self, msg: &[u8]) -> Vec<u8> {
-        let m = Message::from_slice(Sha256::digest(&msg).as_ref())
-            .expect("digest output length doesn't match secp256k1 input length");
-        SECP.sign(&m, &self.0).serialize_der()
+    pub fn sign(&self, msg: &[u8]) -> Result<Vec<u8>, SigningError> {
+        let m = Message::parse_slice(Sha256::digest(msg).as_ref())
+            .map_err(|_| SigningError::new("failed to parse secp256k1 digest"))?;
```
romanb: Why did you choose not to preserve the source of the error in this and a few other places involving parsing?

romanb:
Is eventual `no_std` support the only motivation for this switch or are there others? I'm just wondering. Superficially it seems like `libsecp256k1` has seen a lot less use / exposure. The change set looks good to me, but because I don't feel like I'm in the position to say whether switching to this library is desirable, I'd rather leave the approval of this PR to someone else.

tomaka:
> Is eventual no_std support the only motivation for this switch or are there others?

Yes, that's the only reason!

> Superficially it seems like libsecp256k1 has seen a lot less use / exposure.

It's true, which is why, to me, switching to a different library is not a huge deal.
I think it's ok to consider secp256k1 as "legacy", and thus it's especially annoying when compiling libp2p fails because of an issue with secp256k1.

core/src/identity/secp256k1.rs
twittner: Mostly because the `source` is required to satisfy `Error + Send + Sync + 'static` and those errors do not.

core/src/identity/secp256k1.rs
```diff
@@ -74,28 +68,30 @@ impl From<Keypair> for SecretKey {
...
+pub struct SecretKey(secp256k1::SecretKey);
 
 impl SecretKey {
     /// Generate a new Secp256k1 secret key.
     pub fn generate() -> SecretKey {
-        SecretKey(secp::key::SecretKey::new(&mut secp::rand::thread_rng()))
+        let mut r = rand::thread_rng();
+        let mut b = [0; secp256k1::util::SECRET_KEY_SIZE];
+        // This is how it is done in `secp256k1::SecretKey::random` which
+        // we do not use here because it uses `rand::Rng` from rand-0.4.
```
dvdplm: I had a stab at upgrading to rand 0.6 but the tests use [secp256k1-test](https://crates.io/crates/secp256k1-test) to compare results to the C lib and that crate is stuck on `rand` 0.4.

tomaka:
The CI should now (hopefully) pass. We should see it happening after rebasing.

twittner:
CI failed again. I think the version of llvm in `tomaka/cargo-web` is too old.

```
> llc --version
LLVM (http://llvm.org/):
  LLVM version 3.8.1
  
  Optimized build.
  Built Jun  2 2017 (13:11:29).
  Default target: x86_64-pc-linux-gnu
  Host CPU: skylake

  Registered Targets:
    aarch64    - AArch64 (little endian)
    aarch64_be - AArch64 (big endian)
    amdgcn     - AMD GCN GPUs
    arm        - ARM
    arm64      - ARM64 (little endian)
    armeb      - ARM (big endian)
    bpf        - BPF (host endian)
    bpfeb      - BPF (big endian)
    bpfel      - BPF (little endian)
    cpp        - C++ backend
    hexagon    - Hexagon
    mips       - Mips
    mips64     - Mips64 [experimental]
    mips64el   - Mips64el [experimental]
    mipsel     - Mipsel
    msp430     - MSP430 [experimental]
    nvptx      - NVIDIA PTX 32-bit
    nvptx64    - NVIDIA PTX 64-bit
    ppc32      - PowerPC 32
    ppc64      - PowerPC 64
    ppc64le    - PowerPC 64 LE
    r600       - AMD GPUs HD2XXX-HD6XXX
    sparc      - Sparc
    sparcel    - Sparc LE
    sparcv9    - Sparc V9
    systemz    - SystemZ
    thumb      - Thumb
    thumbeb    - Thumb (big endian)
    x86        - 32-bit X86: Pentium-Pro and above
    x86-64     - 64-bit X86: EM64T and AMD64
    xcore      - XCore
```

The webassembly target became stable only in [version 8.0.0](https://releases.llvm.org/8.0.0/docs/ReleaseNotes.html#changes-to-the-webassembly-target).

```
> llc --version
LLVM (http://llvm.org/):
  LLVM version 8.0.0
  Optimized build.
  Default target: x86_64-pc-linux-gnu
  Host CPU: skylake

  Registered Targets:
    aarch64    - AArch64 (little endian)
    aarch64_be - AArch64 (big endian)
    amdgcn     - AMD GCN GPUs
    arm        - ARM
    arm64      - ARM64 (little endian)
    armeb      - ARM (big endian)
    bpf        - BPF (host endian)
    bpfeb      - BPF (big endian)
    bpfel      - BPF (little endian)
    hexagon    - Hexagon
    lanai      - Lanai
    mips       - MIPS (32-bit big endian)
    mips64     - MIPS (64-bit big endian)
    mips64el   - MIPS (64-bit little endian)
    mipsel     - MIPS (32-bit little endian)
    msp430     - MSP430 [experimental]
    nvptx      - NVIDIA PTX 32-bit
    nvptx64    - NVIDIA PTX 64-bit
    ppc32      - PowerPC 32
    ppc64      - PowerPC 64
    ppc64le    - PowerPC 64 LE
    r600       - AMD GPUs HD2XXX-HD6XXX
    sparc      - Sparc
    sparcel    - Sparc LE
    sparcv9    - Sparc V9
    systemz    - SystemZ
    thumb      - Thumb
    thumbeb    - Thumb (big endian)
    wasm32     - WebAssembly 32-bit
    wasm64     - WebAssembly 64-bit
    x86        - 32-bit X86: Pentium-Pro and above
    x86-64     - 64-bit X86: EM64T and AMD64
    xcore      - XCore
```

tomaka:
For what it's worth, Debian stretch (which the image is based on) only has LLVM 6.0 available in its default package repositories.

tomaka:
I pushed a commit that enables the `no_cc` feature for `clear_on_drop` when compiling for WASM.

tomaka:
Well, github doesn't let me request a review from @twittner.

twittner:
> I pushed a commit that enables the `no_cc` feature for `clear_on_drop` when compiling for WASM.

Because of how feature resolution in cargo works today, this will use the `no_cc` feature on all targets. See this (slightly edited) output:

```
> cargo clean -p clear_on_drop && cargo check --target=wasm32-unknown-unknown -v 2>&1| rg "clear_on_drop.*no_cc"
     Running `rustc --crate-name build_script_build clear_on_drop-0.2.3/build.rs --color never --crate-type bin --emit=dep-info,link -C debuginfo=2 --cfg 'feature="no_cc"'
     Running `rustc --crate-name clear_on_drop clear_on_drop-0.2.3/src/lib.rs --color never --crate-type lib --emit=dep-info,link -C debuginfo=2 --cfg 'feature="no_cc"'
     Running `rustc --crate-name clear_on_drop clear_on_drop-0.2.3/src/lib.rs --color never --crate-type lib --emit=dep-info,metadata -C debuginfo=2 --cfg 'feature="no_cc"'

> cargo clean -p clear_on_drop && cargo check -v 2>&1| rg "clear_on_drop.*no_cc"
     Running `rustc --crate-name build_script_build clear_on_drop-0.2.3/build.rs --color never --crate-type bin --emit=dep-info,link -C debuginfo=2 --cfg 'feature="no_cc"'
     Running `rustc --crate-name clear_on_drop clear_on_drop-0.2.3/src/lib.rs --color never --crate-type lib --emit=dep-info,link -C debuginfo=2 --cfg 'feature="no_cc"'
     Running `rustc --crate-name clear_on_drop clear_on_drop-0.2.3/src/lib.rs --color never --crate-type lib --emit=dep-info,metadata -C debuginfo=2 --cfg 'feature="no_cc"'
```

tomaka:
Is it really that bad? We're talking about a hypothetical issue (zeroing being optimized out by the compiler) that could only be exploited in case of undefined behaviour (which shouldn't happen if we don't use unsafe) about a protocol that's on the road of being deprecated (secp256k1).

twittner:
> We're talking about a hypothetical issue (zeroing being optimized out by the compiler) that could only be exploited in case of undefined behaviour (which shouldn't happen if we don't use unsafe) about a protocol that's on the road of being deprecated (secp256k1).

It is not only about undefined behaviour within *our* process, but also undefined behaviour in other processes which, when combined with OS-level exploits, might allow an attacker to read our process' memory.

> Is it really that bad?

Given that we want to phase out secp256k1 support I guess we can accept the risk. If so, we should *always* use the `no_cc` feature because otherwise looking at the Cargo.toml would give the wrong impression that it is only effective for wasm/emscripten targets.

tomaka:
So what's your opinion? I'd be in favour of merging.

twittner:
I just remembered, that `clear_on_drop` is also used by `ed25519-dalek`, `curve25519-dalek` and `x25519-dalek` so this would not just affect `secp256k1`. Given that we only enable the `no_cc` feature to work around our CI problems I would prefer to change our build image to use an up-to-date LLVM version and not use `no_cc`.

tomaka:
You're right. I took the time to fix the image to use clang 8. The tests are still failing because DockerHub hasn't rebuilt the image yet, and I will restart the build once that's done.

tomaka:
I've realized that since we now have clang 8 available, we can simply compile the C secp256k1 library that this PR removes, and have the same result.

In other words, this PR no longer brings anything in terms of WASM compatibility.

twittner:
Feel free to close it.

tomaka:
I'm still getting weird errors about lax_der_parsing.o being in the wrong format. Let's just merge this.

PR #1032, opened 2019-03-29
 Have `Transport::Listener`s produce `ListenerEvent`s.

twittner:
In order to support transports which do not immediately know their listen address or which dynamically add or remove addresses to the set of addresses they are listening on, this PR returns just a `Listener` from `Transport::listen_on` instead of a pair of listener and multi-address. The `Transport::Listener` is a `Stream` whose items are [`ListenerEvent`](https://github.com/libp2p/rust-libp2p/pull/1032/files#diff-b073c23a91a1a310f31ac5265f6140d5R261)s, which are either upgrades or address information.

A few points worth mentioning:

1. Previously `Multiaddr`esses were also used as identifiers for listeners in the swarm. Given that listeners are no longer assumed to have a stable address, we have to generate stable [`Id`](https://github.com/libp2p/rust-libp2p/pull/1032/files#diff-b35d96f723d9e1ad4e01cbd9840b2754R133)s to refer to them.
2. `Transport` combinators such as `and_then` or `map` made a `ConnectedPoint` enum available to the function. Since we no longer assume immediate addresses those have been replaced by `Endpoint` to allow functions to differentiate between dialer and listener mode.
3. `Swarm::poll` now produces [`SwarmEvent`](https://github.com/libp2p/rust-libp2p/pull/1032/files#diff-cbf45a300b302d16ccfd568f50ccaea5R226)s in order to make address events available to code outside of `NetworkBehviour` implementations. (If there is a way to avoid this I would be happy to know.)

Previous discussions can be found in https://github.com/libp2p/rust-libp2p/pull/1028 and https://github.com/libp2p/rust-libp2p/issues/794#issuecomment-475950601.

Assuming that this PR gets merged eventually, I would like to point out again, that we should reconsider changing the internal representation of `Multiaddr` from `Vec<u8>` to `Bytes`, as previously proposed in https://github.com/libp2p/rust-libp2p/pull/498, since we clone `Multiaddr` values much more often now.

**Edit**: #1041 has been opened which changes `Multiaddr` to use `Bytes` instead of `Vec<u8>`.


core/src/nodes/raw_swarm.rs
```diff
@@ -152,7 +168,9 @@ where
     /// This can include, for example, an error during the handshake of the encryption layer, or
     /// the connection unexpectedly closed.
     IncomingConnectionError {
-        /// Address of the listener which received the connection.
+        /// ID of the listener which received the connection.
+        listener_id: Id,
```
tomaka: Having a `listener_id` is only necessary if we want to allow closing a listener, which isn't possible at the moment.
On one hand it's nice to have the tools to allow that later, on the other hand we might never add this feature or add it in a different way (eg. allow closing individual multiaddresses or something).
IMO we should just not have a `listener_id`, and we can always add it if we want to allow cancelling a listener entirely.

core/src/transport/memory.rs
```diff
@@ -173,23 +175,34 @@ impl error::Error for MemoryTransportError {}
 pub struct Listener {
     /// Port we're listening on.
     port: NonZeroU64,
+    /// The address we are listening on.
+    addr: Multiaddr,
     /// Receives incoming connections.
     receiver: mpsc::Receiver<Channel<Bytes>>,
+    tell_listen_addr: bool
```
tomaka: Needs a doc-comment. It's not obvious what this means.

core/tests/raw_swarm_dial_error.rs
```diff
@@ -181,8 +188,16 @@ fn dial_self() {
         RawSwarm::new(transport, local_public_key.into())
     };
 
-    let listen = swarm.listen_on("/ip4/127.0.0.1/tcp/0".parse().unwrap()).unwrap();
-    swarm.dial(listen.clone(), TestHandler::default().into_node_handler_builder()).unwrap();
+    let listener = swarm.listen_on("/ip4/127.0.0.1/tcp/0".parse().unwrap()).unwrap();
+
+    let address =
+        if let Async::Ready(RawSwarmEvent::NewListenerAddress { listen_addr, .. }) = swarm.poll() {
```
tomaka: Well, that's a bit dangerous as we're not inside a tokio context.

core/tests/raw_swarm_dial_error.rs
```diff
@@ -194,27 +209,34 @@ fn dial_self() {
...
                 }) => {
-                    assert_eq!(multiaddr, listen);
+                    assert_eq!(multiaddr, address);
                     assert!(!got_dial_err);
                     got_dial_err = true;
                     if got_inc_err {
                         return Ok(Async::Ready(()));
                     }
                 },
+                Async::Ready(RawSwarmEvent::NewListenerAddress { listener_id, listen_addr }) => {
```
tomaka: Can this really happen? I'd put `unreachable!()` here instead.

transports/tcp/src/lib.rs
```diff
@@ -277,20 +273,22 @@ impl Future for TcpDialFut {
 /// Stream that listens on an TCP/IP address.
 pub struct TcpListenStream {
     inner: Result<SleepOnError<Incoming>, Option<io::Error>>,
+    listen_addr: Multiaddr,
     /// Original configuration.
     config: TcpConfig,
+    tell_new_addr: bool
```
tomaka: Same remark.

core/src/transport/and_then.rs
```diff
@@ -91,31 +78,31 @@ where
...
+    fun: TMap
+}
 
 impl<TListener, TMap, TTransOut, TMapOut, TListUpgr, TTransErr> Stream for AndThenStream<TListener, TMap>
 where
-    TListener: Stream<Item = (TListUpgr, Multiaddr), Error = TTransErr>,
+    TListener: Stream<Item = ListenerEvent<TListUpgr>, Error = TTransErr>,
     TListUpgr: Future<Item = TTransOut, Error = TTransErr>,
-    TMap: FnOnce(TTransOut, ConnectedPoint) -> TMapOut + Clone,
+    TMap: FnOnce(TTransOut, Endpoint) -> TMapOut + Clone,
```
tomaka: What's the rational for changing `ConnectedPoint` to `Endpoint`?

core/src/nodes/raw_swarm.rs
twittner: The listener ID was introduced because previously the listener address was also used for identification purposes, e.g. in tests. It removed the ID now. Without being able to refer to individual listeners, some tests now made no longer sense and had to be removed as well.

core/src/transport/and_then.rs
twittner: The replacement of `ConnectedPoint` by `Endpoint` was a side-effect of some previous iteration of this PR which did not make the listen address available in upgrades. This got eventually corrected, but I forgot to bring back the `ConnectedPoint`.

core/src/nodes/listeners.rs
```diff
@@ -90,15 +93,25 @@ where
 {
     /// The object that actually listens.
     listener: TTrans::Listener,
-    /// Address it is listening on.
-    address: Multiaddr,
+    /// Addresses it is listening on.
+    addresses: HashSet<Multiaddr>
```
tomaka: What about `SmallVec<[Multiaddr; 2]>`, as this is the number of addresses that I'd expect (127.0.0.1 and the interface IP)?

tomaka:
Generally :+1: but I'm really not sure about the change that makes `Swarm` return a `SwarmEvent`.
The `Swarm` is supposed to be a high-level easy-to-use thing, and the user generally really doesn't really care about the actual address it's listening on.

As far as I know, the listened address is only used in kademlia tests, identify tests, and in Substrate to print in the logs.

An alternative idea could be to we add methods on the `NetworkBehaviour` to signal listening address changes (with a default empty implementation), and add an implementation of `NetworkBehaviour` that generates these changes as events.

For kademlia and identify tests, another idea could be to start polling the swarm, call `Swarm::listened_addrs` regularly, and dial the address only when an address appears.

core/src/nodes/listeners.rs
twittner: Can do. However the number of addresses is clearly dependent on the number of interfaces (and >2 even on my machine). Still small enough I guess so that going from O(1) to O(n) for inserts/lookups does not hurt.

twittner:
As requested I have removed the `SwarmEvent`. Instead tests use randomly generated port numbers (collisions may occur especially in the identify test which only uses the ephemeral TCP port range).

To me it seems the swarm API should change to make the dynamic set of listen addresses accessible outside of `NetworkBehaviour` implementations, be it for tests or for other use-cases. Having to have pre-knowledge of addresses looks wrong to me. If `SwarmEvent` is deemed unacceptable we should come up with something else.

tomaka:
> To me it seems the swarm API should change to make the dynamic set of listen addresses accessible outside of NetworkBehaviour implementations

The point of the `Swarm` is to be the battery behind the `NetworkBehaviour`, while the user is only ever supposed to use the inherent methods of the `NetworkBehaviour` and the events sent by it.

twittner:
> The point of the `Swarm` is to be the battery behind the `NetworkBehaviour`, while the user is only ever supposed to use the inherent methods of the `NetworkBehaviour` and the events sent by it.

The `Swarm` is also a `Stream` of events which is another part of its public API users of libp2p interact with. Arguably integration tests are API users too and it would make this API more useful to those tests if they would be informed about listener address changes.

tomaka:
For me it's an all-or-nothing thing. If we have a `SwarmEvent`, it should not just contain a "new address" variant, but also "connected to a node", "started dialing", "disconnected", "failed to dial", and so on.

As I've already explained a few times, while I'm quite happy with the `RawSwarm` and the content of `core::nodes` in general, I'm really not sure that `Swarm` and `NetworkBehaviour` are the right approach.

PR #1035, opened 2019-03-29
Split the Swarm from the NetworkBehaviour

tomaka:
The `swarm.rs` module is getting more and more complex as we add features.
This PR splits `swarm.rs` in two: `Swarm`/`SwarmBuilder`/etc. in one sub-module, and `NetworkBehaviour`/`NetworkBehaviourAction`/etc. in another.
Doesn't change any logic.


tomaka:
> I wonder though why not keep Swarm etc. in the parent module swarm.rs instead of moving it to swarm/swarm.rs? Just curious.

I don't have a strong opinion on that, but I like that the private fields of `Swarm` are not accessible to the other modules.

PR #1041, opened 2019-04-03
Change `Multiaddr` representation to `Bytes`.

twittner:
If #1032 gets merged we should optimise cloning of `Multiaddr` values since they are part of every incoming connection upgrade and transports will clone them a lot. To make this more efficient, this PR changes the `Multiaddr` representation from `Vec<u8>` to `Bytes` which uses reference counting internally.

Taking the feedback from #498 into account, the `Bytes` type is not visible in the API of `Multiaddr` but an implementation detail. Here is an outline of the changes:

1. Rename `Multiaddr::from_bytes` to `Multiaddr::try_from_vec`. This should emphasise the fact that this is a fallible operation. If `TryFrom` becomes available in rust stable we might prefer an implementation of this trait for `Vec<u8>` instead of this inherent method.
2. Rename `Multiaddr::to_bytes` to `Multiaddr::to_vec`.
3. Remove `Multiaddr::into_bytes`. Since `Multiaddr` is reference counted internally, one always has to make a copy, so calling `Multiaddr::to_vec` is the honest thing to do.
4. While cloning becomes much more efficient with this change, `Multiaddr::append` takes a hit because we have to create an intermediate `Vec<u8>`. However with the assumption that modification is less frequent than cloning it is a price worth paying.


tomaka:
The change is good to me, but I'd prefer to keep the old methods and deprecate them instead of removing them straight. This way we don't have to do yet again another major version bump.

tomaka:
Also, `try_from_vec` should be `from_slice` instead in my opinion.
At the moment it is "from vec" in order to avoid an allocation, but here we have the allocation anyway.

twittner:
> The change is good to me, but I'd prefer to keep the old methods and deprecate them instead of removing them straight. This way we don't have to do yet again another major version bump.

Agreed. I will do so.

> Also, `try_from_vec` should be `from_slice` or `try_from_slice` instead in my opinion.
> At the moment it is "from vec" in order to avoid an allocation, but here we have the allocation anyway.

I was considering it but in fact there is no allocation involved when turning `Vec<u8>` into `Bytes` which is why I kept it this way (cf. https://docs.rs/bytes/0.4.12/src/bytes/bytes.rs.html#1799-1815).

PR #1044, opened 2019-04-04
Expand wildcard IP addresses in TCP transport.

twittner:
Wildcard IP addresses (e.g. 0.0.0.0) are used to listen on all host interfaces. To report those addresses such that clients know about them and can actually make use of them we use the `get_if_addrs` crate and maintain a collection of addresses. We report the whole expansion at the
very beginning of the listener stream with `ListenerEvent::NewAddress` events and add new addresses should they come to our attention.

What remains to be done is to potentially allow users to filter IP addresses, for example the local loopback one, and to detect expired addresses not only if a new address is discovered.

### DIFF
--- a/transports/tcp/src/lib.rs
+++ b/transports/tcp/src/lib.rs
@@ -391,7 +391,7 @@ impl Stream for TcpListenStream {
                 Ok(addr) => addr,
                 Err(err) => {
                     error!("Failed to get peer address: {:?}", err);
-                    return Err(io::Error::new(io::ErrorKind::Other, err))
+                    return Err(err)
                 }
             };
 
@@ -431,7 +431,7 @@ impl Stream for TcpListenStream {
                 }
                 Err(err) => {
                     error!("Failed to get local address of incoming socket: {:?}", err);
-                    return Err(io::Error::new(io::ErrorKind::Other, err))
+                    return Err(err)
                 }
             };
 
--- a/transports/tcp/src/lib.rs
+++ b/transports/tcp/src/lib.rs
@@ -356,7 +356,7 @@ pub struct TcpListenStream {
     /// Stream of incoming sockets.
     inner: Result<SleepOnError<Incoming>, Option<io::Error>>,
     /// The port which we use as our listen port in listener event addresses.
-    port:  u16,
+    port: u16,
     /// The set of known addresses.
     addrs: Addresses,
     /// Temporary buffer of listener events.
@@ -390,10 +390,8 @@ impl Stream for TcpListenStream {
             let sock_addr = match sock.peer_addr() {
                 Ok(addr) => addr,
                 Err(err) => {
-                    // If we can't get the address of the newly-opened socket, there's
-                    // nothing we can do except ignore this connection attempt.
                     error!("Failed to get peer address: {:?}", err);
-                    continue
+                    return Err(io::Error::new(io::ErrorKind::Other, err))
                 }
             };
 
@@ -433,18 +431,15 @@ impl Stream for TcpListenStream {
                 }
                 Err(err) => {
                     error!("Failed to get local address of incoming socket: {:?}", err);
-                    continue
+                    return Err(io::Error::new(io::ErrorKind::Other, err))
                 }
             };
 
             let remote_addr = match sock_addr.to_multiaddr() {
                 Ok(addr) => addr,
                 Err(err) => {
-                    // Socket addresses are supposed to be valid `Multiaddr`s.
-                    // If we fail to derive one from the peer address we can onlu
-                    // ignore this connection.
                     error!("Failed to create multiaddr from peer address: {:?}", err);
-                    continue
+                    return Err(io::Error::new(io::ErrorKind::Other, err))
                 }
             };
 
--- a/transports/tcp/src/lib.rs
+++ b/transports/tcp/src/lib.rs
@@ -38,14 +38,24 @@
 //! The `TcpConfig` structs implements the `Transport` trait of the `swarm` library. See the
 //! documentation of `swarm` and of libp2p in general to learn how to use the `Transport` trait.
 
-use futures::{future, future::FutureResult, prelude::*, Async, Poll};
+use futures::{
+    future::{self, FutureResult},
+    prelude::*,
+    stream::{self, Chain, IterOk}
+};
+use get_if_addrs::get_if_addrs;
 use libp2p_core::{Transport, transport::{ListenerEvent, TransportError}};
 use log::{debug, error};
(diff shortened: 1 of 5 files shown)


transports/tcp/src/lib.rs
```diff
@@ -123,60 +128,77 @@ impl TcpConfig {
...
+            if socket_addr.ip().is_unspecified() {
+                let addrs = host_addresses(port).map_err(TransportError::Other)?;
+                debug!("Listening on {:?}", addrs.values());
+                let new_addrs = addrs.values()
+                    .cloned()
+                    .map(ListenerEvent::NewAddress)
+                    .collect::<Vec<_>>();
+                (Either::Right(addrs), Either::Right(stream::iter_ok(new_addrs)))
+            } else {
+                let ma = local_addr.to_multiaddr().expect("A socket address is a valid Multiaddr.");
```
tomaka: I think you need to append the port here.

transports/tcp/src/lib.rs
```diff
@@ -270,73 +312,140 @@ impl Future for TcpDialFut {
...
+    }
+}
+
 /// Stream that listens on an TCP/IP address.
 pub struct TcpListenStream {
+    /// Stream of incoming sockets.
     inner: Result<SleepOnError<Incoming>, Option<io::Error>>,
-    listen_addr: Multiaddr,
+    /// The port which we use as our listen port in listener event addresses.
+    port:  u16,
```
tomaka:
```suggestion
    port: u16,
```

transports/tcp/src/lib.rs
```diff
@@ -270,73 +312,140 @@ impl Future for TcpDialFut {
...
+                    }
+                }
+                Err(err) => {
+                    error!("Failed to get local address of incoming socket: {:?}", err);
+                    continue
+                }
+            };
+
+            let remote_addr = match sock_addr.to_multiaddr() {
+                Ok(addr) => addr,
```
tomaka: Similarly, I think you have to push the port here.

transports/tcp/src/lib.rs
```diff
@@ -270,73 +312,140 @@ impl Future for TcpDialFut {
...
+                    // If we can't get the address of the newly-opened socket, there's
+                    // nothing we can do except ignore this connection attempt.
+                    error!("Failed to get peer address: {:?}", err);
+                    continue
+                }
+            };
+
+            let listen_addr = match sock.local_addr() {
+                Ok(addr) => match self.addrs.as_mut() {
+                    Either::Left(ma) => ma.clone(),
```
tomaka: I'm not a fan of using `Left` and `Right` here instead of more explicit variants.

transports/tcp/src/lib.rs
```diff
@@ -270,73 +312,140 @@ impl Future for TcpDialFut {
...
+                Ok(Async::Ready(Some(sock))) => sock,
+                Ok(Async::Ready(None)) => return Ok(Async::Ready(None)),
+                Ok(Async::NotReady) => return Ok(Async::NotReady),
+                Err(()) => unreachable!("sleep_on_error never produces an error")
+            };
+
+            let sock_addr = match sock.peer_addr() {
+                Ok(addr) => addr,
+                Err(err) => {
+                    // If we can't get the address of the newly-opened socket, there's
```
tomaka: Actually, we should return `future::err` here.

tomaka:
I'd suggest to not use `to_multiaddr` at all, in order to avoid generating multiaddresses where the port is not specified.

transports/tcp/src/lib.rs
twittner: Sorry, I did not expect your review so soon ;-). I had just force pushed to this branch.

transports/tcp/src/lib.rs
twittner: `local_addr` is a `SocketAddr` which includes the port number, which is either the one provided in the original multi-address or, if a wildcard port number was used, the one that `bind` selected.

transports/tcp/src/lib.rs
twittner: It is currently implemented this way on `master` but I agree that returning an error is more appropriate.

twittner:
> I'd suggest to not use `to_multiaddr` at all, in order to avoid generating multiaddresses where the port is not specified, and instead have a function that requires passing the `SocketAddr` and port as parameters.

All uses of `to_multiaddr` are with `SocketAddr` values that contain port numbers AFAICT.

tomaka:
> All uses of to_multiaddr are with SocketAddr values that contain port numbers AFAICT.

Ugh, I didn't realize that the `ToMultiaddr` trait assumed that `SocketAddr` was TCP/IP. This is wrong.

twittner:
> Ugh, I didn't realize that the `ToMultiaddr` trait assumed that `SocketAddr` was TCP. This is wrong.

Yes, but I think fixing `multiaddr` is outside the scope of this PR.

tomaka:
> Yes, but I think fixing multiaddr is outside the scope of this PR.

Assuming that `to_multiaddr` gets removed for `SocketAddr`, I'd argue that we should not use it.
It's fine if this PR still uses it though, but we delay the problem.

transports/tcp/src/lib.rs
twittner: When you wrote `future::err` you meant [`Err(io::Error ...)`](https://github.com/libp2p/rust-libp2p/pull/1044/commits/da55db23b8f6b18f7f78751f8fef48bb55fb057b), right?

tomaka: I was thinking of returning `Ok(future::err(...))`, similar to lines 450-454.

twittner: This would require a change to `ListenerEvent` since we have no value for `remote_addr`.

twittner: I also think that an error from `TcpStream::peer_addr` and `TcpListener::local_addr` within this loop would be serious problem. Why should it fail at all? On POSIX systems for example, `peer_addr` maps to libc's `getpeername` and `local_addr` to `getsockname`. Here is the union of errors those functions return:
```
EBADF    The argument sockfd is not a valid file descriptor.
EFAULT   The addr argument points to memory not in a valid part of the process address space.
EINVAL   addrlen is invalid (e.g., is negative).
ENOBUFS  Insufficient resources were available in the system to perform the operation.
ENOTCONN The socket is not connected.
ENOTSOCK The file descriptor sockfd does not refer to a socket.
```
None of this should happen, but if it does I think we are in trouble and a `return Err(...)` is warranted.

Issue #1030, opened 2019-03-29
Make it possible to pass metadata from the transport to the protocols handler


tomaka:
At the moment, the `SecioOutput` contains the ephemeral key used during negotiation, but since the `Swarm` requires the transport output to be `(PeerId, impl StreamMuxer)`, we have no way to pass it around.

This is required for https://github.com/paritytech/substrate/issues/271

tomaka:
We don't need it anymore in Substrate, so I'm going to lower the priority tag.

thomaseizinger:
Closing in favor of https://github.com/libp2p/rust-libp2p/issues/2946.

PR #1045, opened 2019-04-04
Generalize TPeerId into TConnInfo

tomaka:
This is the first part of #1030 

At the moment, the structs in the `nodes` module have a `TPeerId` parameter that notably requires `Eq + Hash`.

This PR turns `TPeerId` into `TConnInfo`, and introduces a new trait that `TConnInfo` must implement:

```rust
/// Information about a connection.
pub trait ConnectionInfo {
    /// Identity of the node we are connected to.
    type PeerId: Eq + Hash;

    /// Returns the identity of the node we are connected to on this connection.
    fn peer_id(&self) -> &Self::PeerId;
}

impl ConnectionInfo for PeerId {
    type PeerId = PeerId;

    fn peer_id(&self) -> &PeerId {
        self
    }
}
```

By default, a `TConnInfo` is a `PeerId`. But after this PR, it can consist of anything that contains a `PeerId`.


core/src/nodes/collection.rs
```diff
@@ -143,65 +143,85 @@ where TOutEvent: fmt::Debug,
...
-    #[inline]
-    pub fn peer_id(&self) -> &TPeerId {
-        &self.peer_id
+    /// Returns the information of the connection.
+    pub fn connection_info(&self) -> &TConnInfo {
+        self.conn_info.as_ref().expect("conn_info is always Some when the object is alive; QED")
+    }
+
+    /// Returns the identity of the node we connected to.
+    pub fn peer_id(&self) -> &TPeerId
```
twittner: How about removing `peer_id()`?

tomaka: I had it for convenience, as a "why not have it" item, but I'll remove it.

tomaka: It is there for convenience and is used a lot in `raw_swarm`. Can I keep it? :pray:

twittner: Of course you can keep it :smile:. It was just a thought.

PR #1046, opened 2019-04-04
Add ExpandedSwarm

tomaka:
Turns `Swarm` into a `type Swarm<...> = ExpandedSwarm<...>` where `ExpandedSwarm` has a lot more template parameters.

Unfortunately Rust still has a lot of limitations when it comes to bounds. I got blocked in Substrate because of some combination of requirements that wasn't possible to satisfy in a struct declaration. This PR makes it possible to use the swarm in ways that weren't possible by removing the mandatory `where TBehaviour: NetworkBehaviour` requirement from `Swarm`.


tomaka:
> better off without the NetworkBehaviour trait and instead have users consume events from the swarm stream and providing them with a command channel for the swarm?

From my understanding that basically means using the `RawSwarm` directly.
The `NetworkBehaviour` trait is mostly for composability, so that we can provide behaviours and let the user combine behaviours together in an easy way.

Issue #950, opened 2019-02-18
The ping behaviour shouldn't use the OneShotHandler


tomaka:
Ping's job is to maintain connections alive, which is too "meta" to be handled by the `OneShotHandler`.
In the process, tests should be written for ping.

romanb:
What does "too meta" mean in this context?

tomaka:
The role of the `OneShotHandler` is to handle a basic protocol where we send messages and optionally receive responses.
However the role of "ping" is also to close the connection if no response is received in a given time. This is too different from a `OneShotHandler` that they shouldn't share their code.

romanb:
Closed by https://github.com/libp2p/rust-libp2p/pull/1049.

Issue #864, opened 2019-01-17
Add timeout to `ProtocolsHandlerEvent::OutboundSubstreamRequest`


tomaka:
Right now the substream opening timeout is set for each connection by the node handler. Instead it is the `ProtocolsHandler` that should handle this.

At the same time, making `connection_keep_alive()` return true should be done by the `ProtocolsHandler` as well, and not by the `NodeHandler`.

tomaka:
The second point hasn't been done in order to reduce the difficulty of implementing `ProtocolsHandler`. However after #833 this should no longer be a valid reason.

romanb:
Closed by https://github.com/libp2p/rust-libp2p/pull/1049.

Issue #842, opened 2019-01-10
Added a test requiring a ping time inferior to a certain threshold


tomaka:
I see ping times that tend to be quite high, and that's a bit worrisome.
I think we should add a test where two nodes connect with a `Ping` behaviour, and ensure that the ping time is inferior to, say, 15ms.

romanb:
Closed by https://github.com/libp2p/rust-libp2p/pull/1049.

Issue #828, opened 2019-01-08
Adjust Ping to be a one-channel-one-shot protocol


tomaka:
Right now the code of `ping` still supports sending multiple pings on the same substream.
This complicates the code quite a bit.

PR #1049, opened 2019-04-09
libp2p-ping improvements.

romanb:
This PR suggests some improvements to `libp2p-ping` relating to (and thus possibly closing) the following issues:

  * #950: This PR removes use of the `OneShotHandler`, but still sends each ping over a new substream as before and as seems to be intentional since #828.

  * #842: This PR adds an integration test that exercises the ping behaviour through a `Swarm`, requiring the RTT to be below a desired threshold. This requires disabling Nagle's algorithm as it can interact badly with delayed ACKs on otherwise low-traffic connections, which has been observed to be the case in the context of the new ping example and the integration test. Ping times observed locally with Nagle's algorithm enabled ~100ms and disabled ~5-10ms (debug) resp. < 1ms (`--release`). While disabling Nagle's algorithm in production deployments is debatable, it should be uncontroversial for development environments to maybe avoid some surprises / head-scratching over network delays, so this PR sets `nodelay(true)` on the `TcpConfig` in `CommonTransport` which underlies the development transports.

  * #864: This PR moves control of the inbound and outbound (sub)stream protocol upgrade
    timeouts from the `NodeHandlerWrapperBuilder` to the `ProtocolsHandler`. That may also alleviate the need for a custom timeout on an `OutboundSubstreamRequest` (#864) as a `ProtocolsHandler` is now free to adjust these timeouts over time. The `NodeHandlerWrapperBuilder` does not seem to play an important role, judging by how easy it was to remove the timeout configuration from it, leaving it so bare that it could possibly be removed, but that may just be ignorance on my side due to lack of familiarity (and possibly history) of that code.

These changes affect the public API of `libp2p-ping` (for the better, I hope). Furthermore, in terms of behaviour, this implementation avoids pinging on the connection in  both directions by resetting the ping timer on every successfully answered ping, i.e. the peer with the lower ping interval "drives" the ping protocol to keep the connection alive. This is best illustrated through some output of the new ping example:

Node 1:
```
Local peer id: PeerId("QmYjrdGG9XNn38N5nabUbw11Pv2Zd57p8sLdycLzP4WwmE")
Listening on "/ip4/0.0.0.0/tcp/36481" (ping interval: 19s)
PingEvent { peer: PeerId("QmbvR17VPFgSxc9UXCEqhVjpg7KEFBkVyXSxQ8dWxz526N"), result: Success(Pong) }
PingEvent { peer: PeerId("QmbvR17VPFgSxc9UXCEqhVjpg7KEFBkVyXSxQ8dWxz526N"), result: Success(Ping { rtt: 107.178µs }) }
PingEvent { peer: PeerId("QmbvR17VPFgSxc9UXCEqhVjpg7KEFBkVyXSxQ8dWxz526N"), result: Success(Pong) }
PingEvent { peer: PeerId("QmbvR17VPFgSxc9UXCEqhVjpg7KEFBkVyXSxQ8dWxz526N"), result: Success(Pong) }
PingEvent { peer: PeerId("QmbvR17VPFgSxc9UXCEqhVjpg7KEFBkVyXSxQ8dWxz526N"), result: Success(Pong) }
...
``` 

Node 2:
```
Local peer id: PeerId("QmbvR17VPFgSxc9UXCEqhVjpg7KEFBkVyXSxQ8dWxz526N")
Listening on "/ip4/0.0.0.0/tcp/43495" (ping interval: 13s)
Dialed "/ip4/127.0.0.1/tcp/36481"
PingEvent { peer: PeerId("QmYjrdGG9XNn38N5nabUbw11Pv2Zd57p8sLdycLzP4WwmE"), result: Success(Ping { rtt: 130.099µs }) }
PingEvent { peer: PeerId("QmYjrdGG9XNn38N5nabUbw11Pv2Zd57p8sLdycLzP4WwmE"), result: Success(Pong) }
PingEvent { peer: PeerId("QmYjrdGG9XNn38N5nabUbw11Pv2Zd57p8sLdycLzP4WwmE"), result: Success(Ping { rtt: 573.946µs }) }
PingEvent { peer: PeerId("QmYjrdGG9XNn38N5nabUbw11Pv2Zd57p8sLdycLzP4WwmE"), result: Success(Ping { rtt: 212.955µs }) }
PingEvent { peer: PeerId("QmYjrdGG9XNn38N5nabUbw11Pv2Zd57p8sLdycLzP4WwmE"), result: Success(Ping { rtt: 601.327µs }) }
...
```

Please let me know if that is undesirable and always pinging in both directions is preferred. It could also be made configurable.

Other changes:

  * Documentation improvements along the way.
  * A new, basic ping example (which was used to capture the output shown above).


tomaka:
> The NodeHandlerWrapperBuilder does not seem to play an important role, judging by how easy it was to remove the timeout configuration from it, leaving it so bare that it could possibly be removed, but that may just be ignorance on my side due to lack of familiarity (and possibly history) of that code.

To explain this:
- When we start dialing or when we receive an incoming connection, we create a background task and use an object that implements [the `IntoNodeHandler` trait](https://docs.rs/libp2p-core/0.6.0/libp2p_core/nodes/handled_node_tasks/trait.IntoNodeHandler.html).
- When the dialing succeeds, when or the incoming connection has been successfully upgraded, we call the `into_handler()` method on this object. This yields an object that implements `NodeHandler` that *actually* handles the connection.
- The `ProtocolsHandler` trait has its equivalent with [`IntoProtocolsHandler`](https://docs.rs/libp2p-core/0.6.0/libp2p_core/protocols_handler/trait.IntoProtocolsHandler.html).
- The `NodeHandlerWrapper` wraps around a `ProtocolsHandler` and implements `NodeHandler`, while the `NodeHandlerWrapperBuilder` wraps around an `IntoProtocolsHandler` and implements `IntoNodeHandler`.
- It is the `Swarm` that does the bridging between the node* and protocols* worlds, as shown here: https://github.com/libp2p/rust-libp2p/blob/98b251740377d78f06fa7fedee64138e052fb9cc/core/src/swarm/swarm.rs#L57
- This is all a bit confusing because the `IntoNodeHandler` and `IntoProtocolsHandler` traits are also automatically implemented on objects that implement respectively `NodeHandler` and `ProtocolsHandler`. This means that the user can directly pass a handler, instead of passing a builder.

tomaka:
> this implementation avoids pinging on the connection in both directions by resetting the ping timer on every successfully answered ping, i.e. the peer with the lower ping interval "drives" the ping protocol to keep the connection alive.

I think it is still desirable to ping on both sides in order for both sides to know the ping time of the remote.

core/src/protocols_handler/mod.rs
```diff
@@ -108,22 +108,35 @@ pub trait ProtocolsHandler {
     /// and will be passed back in `inject_substream` or `inject_outbound_closed`.
     type OutboundOpenInfo;
 
-    /// Produces a `ConnectionUpgrade` for the protocol or protocols to accept when listening.
+    /// The timeout for protocol negotiation on inbound substreams.
+    fn inbound_timeout(&self) -> Duration {
+        Duration::from_secs(10)
```
tomaka: What about changing `listen_protocol()` to return a tuple of `(Self::InboundProtocol, Duration)` instead?

core/src/protocols_handler/mod.rs
```diff
@@ -108,22 +108,35 @@ pub trait ProtocolsHandler {
...
+    }
+
+    /// The timeout for protocol negotiation on outbound substreams.
+    fn outbound_timeout(&self) -> Duration {
```
tomaka: Similarly, what about adding a `timeout` field to `ProtocolsHandlerEvent::OutboundSubstreamRequest` instead?

core/src/protocols_handler/mod.rs
```diff
@@ -134,28 +147,41 @@ pub trait ProtocolsHandler {
...
     ///
-    /// On the other hand, the return value is only an indication and doesn't mean that the user
-    /// will not close the connection.
+    /// Returning [`KeepAlive::Until`] indicates that the connection may be closed
+    /// and this handler destroyed after the specified `Instant`.
     ///
-    /// When multiple `ProtocolsHandler` are combined together, the largest `KeepAlive` should be
-    /// used.
+    /// Returning [`KeepAlive::Forever`] indicates that the connection should
+    /// always be kept alive. The connection will only be closed if
```
tomaka: Should "always" be kept alive is a bit misleading in my opinion:
- We may allow the user to forcibly close connections in the future.
- When `connection_keep_alive()` is called again, it can return a value other than `Forever`.

examples/ping.rs
```diff
@@ -0,0 +1,95 @@
...
+    let cfg = PingConfig::new().with_interval(int);
+    let behaviour = Ping::new(cfg);
+
+    // Create a Swarm to manage connections established with
+    // the transport and supporting the ping behaviour.
+    let mut swarm = Swarm::new(transport, behaviour, peer_id);
+
+    // Listen on all interfaces and a random, OS-assigned port.
+    let listen_addr = Swarm::listen_on(&mut swarm, "/ip4/0.0.0.0/tcp/0".parse().unwrap()).unwrap();
+    println!("Listening on {:?} (ping interval: {:?})", listen_addr, int);
```
tomaka: Just as a note, this will conflict with #1032

protocols/ping/src/handler.rs
```diff
@@ -21,107 +21,207 @@
...
+pub struct PingConfig {
+    /// The timeout of an outbound ping.
+    timeout: Duration,
+    /// The duration between the last successful outbound or inbound ping
+    /// and the next outbound ping.
+    interval: Duration,
+    /// The maximum number of failed outbound pings before the associated
+    /// connection is deemed unhealthy, indicating to the `Swarm` that it
+    /// should be closed.
+    max_timeouts: NonZeroU32
```
tomaka: What's the reason for having a value different from `1` here?
You mention that inbound pings can keep a connection alive even if outbound pings fail, but I don't really see how that could happen unless the remote is malicious/buggy/overloaded.

protocols/ping/src/handler.rs
```diff
@@ -21,107 +21,207 @@
...
-        self.inner.inject_fully_negotiated_inbound(protocol)
+    fn listen_protocol(&self) -> protocol::Ping {
+        protocol::Ping
     }
 
-    fn inject_fully_negotiated_outbound(&mut self, duration: Duration, info: Self::OutboundOpenInfo) {
-        self.inner.inject_fully_negotiated_outbound(duration, info)
+    fn inject_fully_negotiated_inbound(&mut self, _: ()) {
+        // A ping from a remote peer has been answered.
+        self.last_result = Some(Ok(PingSuccess::Pong));
```
tomaka: If we receive both an inbound and an outbound ping simultaneously, one of them will be ignored because they both write over `last_result`.

At the moment I think that we call `poll()` after every single method call, but I can imagine a future where we call multiple methods other than `poll` in a row, then only `poll` at the end.

Alternatively, we could also enforce that `poll()` gets called after every single method call, and write a doc-comment about this in the `ProtocolsHandler` trait definition.

protocols/ping/src/protocol.rs
```diff
@@ -75,143 +94,126 @@ where
 
     #[inline]
     fn upgrade_outbound(self, socket: Negotiated<TSocket>, _: Self::Info) -> Self::Future {
-        let payload: [u8; 32] = EntropyRng::default().sample(Standard);
-        debug!("Preparing for ping with payload {:?}", payload);
+        let payload = thread_rng().sample(distributions::Standard);
```
tomaka: Nit: I liked the fact that `[u8; 32]` was explicitly written here, as it's an important part of the protocol.

protocols/ping/src/protocol.rs
```diff
@@ -75,143 +94,126 @@ where
...
+                    PingDialerState::Shutdown {
+                        inner: nio::shutdown(socket),
+                        rtt,
                     }
                 },
-                PingDialerInner::Shutdown { ref mut inner, ping_time } => {
-                    let _ = try_ready!(inner.poll());
-                    return Ok(Async::Ready(ping_time));
+                PingDialerState::Shutdown { ref mut inner, rtt } => {
+                    try_ready!(inner.poll());
```
tomaka: Does that not produce a warning? The `let _ = ` are usually here to silence warnings about unused results.

romanb:
> To explain this:
>
> [..]

Thanks for this elaboration, which I think matches my understanding so far. Do we agree that it is the `ProtocolsHandler` that should be in control of the inbound and outbound timeouts for substream upgrades? As you have shown, the `Swarm` fixes the `THandler` type parameter of the `RawSwarm` to `NodeHandlerWrapperBuilder`. But as this PR shows, that was never done anywhere, i.e. `with_in_negotiation_timeout` and `with_out_negotiation_timeout` were never used. 

I figured that the type `NodeHandlerWrapperBuilder` is a part of the public `ProtocolsHandler` API in the form of `ProtocolsHandler::into_node_handler_builder`. So I first thought the timeouts can be adjusted by overriding the default implementation of this method, which can be done as follows:
```rust
    fn into_node_handler_builder(self) -> NodeHandlerWrapperBuilder<Self> {
        <Self as IntoProtocolsHandler>::into_node_handler_builder(self)
            .with_in_negotiation_timeout(Duration::from_secs(42))
            .with_out_negotiation_timeout(Duration::from_secs(42))
    }
```
but that would not seem advisable as it would result in diverging implementations between `<MyProtocolsHandler as IntoProtocolsHandler>::into_node_handler_builder` and `<MyProtocolsHandler as ProtocolsHandler>::into_node_handler_builder`, so maybe there should be a big warning about overriding this method. That kind of API seems a bit subtle / fragile and I'm left wondering what is or was the idea about where  `with_in_negotiation_timeout` / `with_out_negotiation_timeout` are meant to be used?

Since `Swarm` fixes the choice of the `IntoNodeHandler` implementation to `NodeHandlerWrapperBuilder` which in turn fixes the choice of the `NodeHandler` implementation to `NodeHandlerWrapper`, I'm also wondering where this flexibility pays off? Is the intention to allow users of `libp2p_core` to use a `RawSwarm` directly, ignoring `Swarm`, and constructing their own Swarm-like API on-top, which seems a bit far-fetched? Or is it particularly useful for testing?

> This means that the user can directly pass a handler, instead of passing a builder.

Which part of the API are you referring to here exactly, i.e. where would the user of `libp2p` directly pass a handler instead of a builder?

Thanks for your patience!

romanb:
> > this implementation avoids pinging on the connection in both directions by resetting the ping timer on every successfully answered ping, i.e. the peer with the lower ping interval "drives" the ping protocol to keep the connection alive.
> 
> I think it is still desirable to ping on both sides in order for both sides to know the ping time of the remote.

Should it not even be configurable? When using the ping protocol only as an application-layer keep-alive functionality, without interest in the round-trip times, it seems wasteful to ping in both directions, does it not?

examples/ping.rs
romanb: Thanks, I'm aware!

core/src/protocols_handler/mod.rs
romanb: Good point, that should be reworded.

protocols/ping/src/handler.rs
romanb: > What's the reason for allowing values different from `1` here? (EDIT: by that I mean: what's the reason for having this feature at all?)

I think basing a connection keep-alive on the timeout of a single request is a bad idea and leads to unnecessary disconnects / connection turnover. There can always be the occasional unfortunate request that gets routed along a bad path. Don't you think so?

> You mention that inbound pings can keep a connection alive even if outbound pings fail, but I don't really see how that could happen unless the remote is malicious/buggy/overloaded.

That is simply because, as I mentioned, when a ping is received that currently also resets the local node's own ping timer and extends the keep-alive of the connection. I.e. every inbound ping or answer to an outbound ping is taken as indication that the connection is still "alive". Only if there is a time window where pings don't get through in either direction is the connection deemed "dead".

protocols/ping/src/handler.rs
romanb: I am aware of the possible overriding of the `last_result` and considered queuing all results but I currently don't think there is actually any harm in only keeping the  `last_result`, do you? I mean, it is clear that the precision of the ping-based keep-alive managed by this handler is determined by how frequently it is being polled and if multiple inbound/outbound substream outputs are "injected" between polling, it just means that this precision suffers. Do you think that a) it would be documented more clearly that only the result of the last inbound or outbound substream upgrade is remembered between invocations of `poll` or b) that all results should be queued? I don't have a strong opinion on the matter, the current implementation just seemed simpler and I didn't see much harm in overriding the occasional intermediate result.

protocols/ping/src/protocol.rs
romanb: Ok, though `[u8; 32]` can be seen in quite a few places in the type signatures in this module already! :)

protocols/ping/src/protocol.rs
romanb: It does not produce a warning in this case.

core/src/protocols_handler/mod.rs
romanb: Thanks for the suggestion, I haven't though of that. I can try going down that route, though it does mean a breaking API change for `ProtocolsHandler`, if we're ok with that.

core/src/protocols_handler/mod.rs
romanb: I can try going down that route, together with the change to `listen_protocol` that you suggested, though it does mean a breaking API change for `ProtocolsHandler`, if we're ok with that.

protocols/ping/src/handler.rs
tomaka: > There can always be the occasional unfortunate request that gets routed along a bad path. Don't you think so?

All the libp2p transports must be reliable, so if the ping doesn't go through it will be sent again.

tomaka:
> Do we agree that it is the ProtocolsHandler that should be in control of the inbound and outbound timeouts for substream upgrades?

Yes!

> As you have shown, the Swarm fixes the THandler type parameter of the RawSwarm to NodeHandlerWrapperBuilder. But as this PR shows, with_in_negotiation_timeout and with_out_negotiation_timeout were never used.

True. This was to leave the door open for people who want to use the `RawSwarm` instead of the `Swarm`. If possible, the `Swarm` shouldn't rely on anything private.

> That kind of API seems a bit subtle / fragile and I'm left wondering what is or was the idea about where with_in_negotiation_timeout / with_out_negotiation_timeout are meant to be used?

It's totally fine to remove them.

> Is the intention to allow users of libp2p_core to use a RawSwarm directly, ignoring Swarm, and constructing their own Swarm-like API on-top, which seems a bit far-fetched?

Yes! Mostly because the `Swarm` is quite opinionated while the `RawSwarm` isn't. The `Swarm` forces you to do things that you may not want to do, such as garbage collecting connections after a few seconds.

I've recently realized that it wasn't really possible to build your own `Swarm` because one cannot build the `PollParameters` object that must be passed to `NetworkBehaviour`. This should be fixed eventually.

tomaka:
> Which part of the API are you referring to here exactly, i.e. where would the user of libp2p directly pass a handler instead of a builder?

The `type ProtocolsHandler;` associated type in `NetworkBehaviour` normally has to implement `IntoProtocolsHandler`.
But since the trait is automatically implemented on anything that implements `ProtocolsHandler`, you can simply specify a struct that implements `ProtocolsHandler` instead.

> Should it not even be configurable? When using the ping protocol only as an application-layer keep-alive functionality, without interest in the round-trip times, it seems wasteful to ping in both directions, does it not?

Oh sure, it can be configurable! In Substrate we would want pings in both directions, but I can totally imagine applications that don't need that.

protocols/ping/src/handler.rs
tomaka: I'm also unsure of what to do.

protocols/ping/src/handler.rs
romanb: > > There can always be the occasional unfortunate request that gets routed along a bad path. Don't you think so?
> 
> This is true for ICMP pings, but here all libp2p transports must be reliable, so if the ping doesn't go through it will automatically be sent again.

Hm, the semantics of "reliable" are necessarily not well-defined for the transports though and there can be other reasons in the sending or receiving application for an occasional transient error during processing of a request, which I think should not immediately cause a connection to be closed. Thus I still think this configuration option is useful and should not default to `1`. Does anyone else reading this have an opinion on the matter one way or the other?

romanb:
> > Is the intention to allow users of libp2p_core to use a RawSwarm directly, ignoring Swarm, and constructing their own Swarm-like API on-top, which seems a bit far-fetched?
> 
> Yes! Mostly because the `Swarm` is quite opinionated while the `RawSwarm` isn't. The `Swarm` forces you to do things that you may not want to do, such as garbage collecting connections after a few seconds.
> 
> I've recently realized that it wasn't really possible to build your own `Swarm` because one cannot build the `PollParameters` object that must be passed to `NetworkBehaviour`. This should be fixed eventually.

I see. Thanks for clarifying that.

protocols/ping/src/handler.rs
tomaka: >  the semantics of "reliable" are necessarily not well-defined for the transports though

It's probably not well documented, but implementations of `StreamMuxer` must be reliable (either the connection is well-working or it's completely dead, there's no in between), and a transport must be reliable if we build a muxer on top of it.

> there can be other reasons in the sending or receiving application for an occasional transient error during processing of a request

I'd argue that no, there shouldn't be any reason for an occasional transient error!

romanb: > > the semantics of "reliable" are necessarily not well-defined for the transports though
> 
> It's probably not well documented, but implementations of `StreamMuxer` must be reliable (either the connection is well-working or it's completely dead, there's no in between), and a transport must be reliable if we build a muxer on top of it.

I meant something else though, namely that the definition of "reliable" is never the same between two different transports, e.g. it has a very specific meaning for TCP in terms of maximum timeouts for retransmissions before giving up, and many other configuration options. I think the common denominator for the abstract concept of an underlying reliable transport in the context of a higher-level protocol as in `libp2p-ping` can only be that it has some form of error-recovery e.g. through retransmission and/or error-detecting or error-correcting codes and delivers messages in-order, if they are delivered at all.

> > there can be other reasons in the sending or receiving application for an occasional transient error during processing of a request
> 
> I'd argue that no, there shouldn't be any reason for an occasional transient error!

That would be great, but does not match my perception of reality, where even assuming a flawless transport and network, there are a lot of other pieces of software and hardware involved on both the sender's and receiver's side, all subject to transient failures of varying frequency. All I'm saying is that I don't think it is advisable to judge a network connection (or really anything) as broken based on a single observation of failure, given that a) they are not cheap to recreate and b) it can cause significant work for application protocols to pick up from where they left off on a new connection.

I'm not sure its worth so much discussion though, given that the current approach certainly allows the configuration that you prefer. Are you strongly opposed to allowing this to be configurable at all?

core/src/protocols_handler/mod.rs
tomaka: Breaking changes are ok as long as they are not too large and we properly document in the CHANGELOG. Also, we're already breaking the `Transport` trait in this release.

protocols/ping/src/handler.rs
tomaka: > All I'm saying is that I don't think it is advisable to judge a network connection (or really anything) as broken based on a single observation of failure

I'm pretty sure that TCP guarantees that no packet gets lost or modified.

Additionally, the state of the stream cipher in the symmetric encryption follows the data that we send on the connection. If any byte we send is lost the connection become unrecoverable, even if we wanted to recover from it. The hmac also guarantees that if any byte is modified, an error is  triggered and the whole connection gets killed as well.

> I'm not sure its worth so much discussion though, given that the current approach certainly allows the configuration that you prefer. Are you strongly opposed to allowing this to be configurable at all?

I'm not opposed to having this configurable. I think the default should be `1` though.

romanb:
@tomaka I addressed your comments in two stages:

  1. https://github.com/libp2p/rust-libp2p/pull/1049/commits/a127afd6d75ab0f83d3c9f022fa686f3e01cc501 : Adds the `PingPolicy` to the config, defaulting to always pinging and adjusts the default config for `max_timeouts: 1`. Regarding [this discussion](https://github.com/libp2p/rust-libp2p/pull/1049/files#r273670143), it is actually indeed possible that `inject_fully_negotiated_inbound` and `inject_fully_negotiated_outbound` are both called before `poll()`, which is easy to witness when the peers ping each other pretty much simultaneously. So to ensure all events are delivered, all results are now buffered between `poll()`s, which guarantees that all events are eventually delivered.

  2. https://github.com/libp2p/rust-libp2p/pull/1049/commits/67bb2eb17eb2c61f90b6268f4899467b607107ca : Removes `inbound_timeout` / `outbound_timeout` from `ProtocolsHandler` with replacements as suggested. Note that I chose to introduce a `ListenProtocol` instead of just returning a tuple to future-proof the API a bit. Let me know if you'd rather really just see a plain tuple be used.

If these changes are satisfactory, I can make a final merge from / rebase against the current master.

core/src/protocols_handler/mod.rs
```diff
@@ -246,6 +236,69 @@ pub trait ProtocolsHandler {
...
+        ListenProtocol {
+            upgrade: self.upgrade.clone(),
+            timeout: self.timeout,
+            _marker: std::marker::PhantomData
+        }
+    }
+}
+
+impl<TUpgrade, TSubstream> ListenProtocol<TUpgrade, TSubstream> {
+    pub fn new(upgrade: TUpgrade) -> ListenProtocol<TUpgrade, TSubstream> {
```
tomaka: Nit: I think we should have an `impl From<TUpgrade> for ListenProtocol` that does the same thing as `new`.

romanb:
@tomaka I actually realised that `ListenProtocol` is needlessly specific and generalised it to `SubstreamProtocol`, reusing it both for `listen_protocol()` and `OutboundSubstreamRequest`. It makes for a more symmetric API and better code reuse. See https://github.com/libp2p/rust-libp2p/pull/1049/commits/2ae05f8ceb7436261493643559eedae38b6c9b3b. Let me know what you think of that.

PR #1050, opened 2019-04-10
Patch reading/writing frame lengths in libp2p-noise.

romanb:
This is a patch extracted from https://github.com/libp2p/rust-libp2p/pull/1027 concerning the reading and writing of frame lengths, since the fate of that PR it still unclear.

### DIFF
--- a/protocols/noise/src/io.rs
+++ b/protocols/noise/src/io.rs
@@ -282,7 +282,8 @@ impl<T: io::Write> io::Write for NoiseOutput<T> {
                             self.write_state = WriteState::Eof;
                             return Err(io::ErrorKind::WriteZero.into())
                         }
-                        Ok(true) => ()                    }
+                        Ok(true) => ()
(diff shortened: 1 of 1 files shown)


protocols/noise/src/io.rs
```diff
@@ -249,13 +268,21 @@ impl<T: io::Write> io::Write for NoiseOutput<T> {
...
+                                self.write_state = WriteState::WriteLen{ len, buf, off };
+                            }
+                            return Err(e)
+                        }
+                        Ok(false) => {
+                            trace!("write: eof");
+                            self.write_state = WriteState::Eof;
+                            return Err(io::ErrorKind::WriteZero.into())
+                        }
+                        Ok(true) => ()                    }
```
twittner: The curly brace should go on the next line I think.

romanb: Oops!

PR #1052, opened 2019-04-11
Remove `Transport::nat_traversal` and refactor `multiaddr`.

twittner:
Context: #794.

The functionality is available through `Multiaddr::replace`. What we currently call "nat_traversal" is merley a replacement of an IP address prefix in a `Multiaddr`, hence it can be done directly on `Multiaddr` values instead of having to go through a `Transport`.

In addition this PR consolidates changes made to `Multiaddr` in previous commits which resulted in lots of deprecations. It adds some more (see below for the complete list of API changes) and removes all deprecated functionality, requiring a minor version bump.

Here are the changes to `multiaddr` compared to the currently published version:

1.  Removed `into_bytes` (use `to_vec` instead).
2.  Renamed `to_bytes` to `to_vec`.
3.  Removed `from_bytes` (use the `TryFrom` impl instead).
4.  Added `with_capacity`.
5.  Added `len`.
6.  Removed `as_slice` (use `AsRef` impl instead).
7.  Removed `encapsulate` (use `push` or `with` instead).
8.  Removed `decapsulate` (use `pop` instead).
9.  Renamed `append` to `push`.
10. Added `with`.
11. Added `replace`.
12. Removed `ToMultiaddr` trait (use `TryFrom` instead).


tomaka:
Given that the `TryFrom` and `TryInto` traits have just been stabilized, another option is to deprecate `ToMultiaddr` entirely and implement `TryFrom` instead.

twittner:
It is a good idea to use `TryFrom` once rust 1.34.0 has been released later today, but deprecating `ToMultiaddr` also has implications for `Multiaddr::encapsulate` and `Multiaddr::decapsulate` because of their generic type bound. Honestly, I think we should rather go for multiaddr 0.3 instead of massaging all these deprecations. I plan to migrate `Transport::nat_traversal` to `Multiaddr` and it would be okay to postpone merging this PR until then. All the multiaddr changes together deserve a 0.3 release IMO.

twittner:
The images `tomaka/cargo-web` and `tomaka/rust-mingw-docker` need to be updated to rust 1.34.0.

tomaka:
I'm not sure how I feel about the NAT traversal thingy.
The new code definitely doesn't look very robust. For example if you pass `/ip4/1.2.3.4/tcp/5/p2p-circuit/ip4/6.7.8.9/tcp/10` and `/ip4/11.12.13.14/tcp/15`, you get `/ip4/11.12.13.14/tcp/5/p2p-circuit/ip4/11.12.13.14/tcp/10` (the IP is replaced twice).

I think that:
1) It would be better to have that logic in a freestanding function. I understand why we wouldn't put it in the `multiaddr` crate, but at least have it available outside of `RawSwarm` and `Mdns`. We probably want to write lots of tests for this function.
2) The function logic should be: replace only the first element the address if it is `/ip4` or `/ip6`, not *any* location.

twittner:
> The new code definitely doesn't look very robust. For example if you pass `/ip4/1.2.3.4/tcp/5/p2p-circuit/ip4/6.7.8.9/tcp/10` and `/ip4/11.12.13.14/tcp/15`, you get `/ip4/11.12.13.14/tcp/5/p2p-circuit/ip4/11.12.13.14/tcp/10` (the IP is replaced twice).

No, [`Multiaddr::replace`](https://github.com/libp2p/rust-libp2p/pull/1052/files#diff-75dfd2141539fcdccd76db7a161a7b9dR141) requires a positional parameter and in [`raw_swarm.rs`](https://github.com/libp2p/rust-libp2p/pull/1052/files#diff-0c5110fdcca8841806bc220275bed002R776) only the first address protocol is considered so your example would be translated to `/ip4/11.12.13.14/tcp/5/p2p-circuit/ip4/6.7.8.9/tcp/10`.

> 1. It would be better to have that logic in a freestanding function. I understand why we wouldn't put it in the `multiaddr` crate, but at least have it available outside of `RawSwarm` and `Mdns`. Also make `RawSwarm` and `Mdns` call the same function. We probably want to write lots of tests for this function.

I do not mind that. The focus was to remove `Transport::nat_traversal` and I did not want to touch the swarm parts too much. Just tell me where you would like to see that function and I move it there.

> 2. The function logic should be: replace only the first element the address if it is `/ip4` or `/ip6`, not _any_ location.

As mentioned above, this is already the case.

tomaka:
> Just tell me where you would like to see that function and I move it there.

I think that a freestanding function in its module but re-exported from the root of `libp2p-core` is fine.

core/src/translation.rs
```diff
@@ -0,0 +1,43 @@
...
+    original.replace(0, move |proto| match proto {
+        Protocol::Ip4(_) | Protocol::Ip6(_) => match observed.iter().next() {
+            x@Some(Protocol::Ip4(_)) => x,
+            x@Some(Protocol::Ip6(_)) => x,
+            _ => None
+        }
+        _ => None
+    })
+}
+
```
tomaka:
```suggestion
// TODO: add tests
```

tomaka:
cc'ing #794 to keep track of things

core/src/translation.rs
twittner: Note that some tests already exist in multiaddr: https://github.com/libp2p/rust-libp2p/pull/1052/files#diff-fe826bf311232628164a9542628146adR306, but the more the merrier.

Issue #1003, opened 2019-03-16
Add a method to Kademlia to get content of k-buckets


tomaka:
Useful for general diagnostic.

tomaka:
While this has been done in #1016, it would be nice to have a better API for `Kademlia` that allows querying the state of a particular peer, and in general being able to obtain any information we need about the buckets from the external world.

romanb:
It would be good to update this issue in light of the current Kademlia API (or close it). I personally don't really know what's missing.

tomaka:
> I personally don't really know what's missing.

The idea behind this issue was to be able to print out the content of the k-buckets for debugging purposes.

I must admit that I don't remember what's missing apart from `kbuckets_entries`.

PR #1056, opened 2019-04-12
Add a method to Kademlia to get detailed information about peers

elferdo:
Addresses #1003 

This PR adds `peer_state()` and `peer_states()` to `Kademlia` and allows to query their state within the table. I guess issue #1003 suggests also being able to query values stored in Kademlia but if that is necessary, I would address that in a separate PR.


tomaka:
Looks good, but the code duplication because of `&self` vs `&mut self` is definitely not great.
I'd be in favour of having just one API and requiring `&mut self` everywhere, instead of having this duplication.

elferdo:
I can change it, but I feel that having a query interface that takes `&mut` communicates a misleading intention. In general I prefer assuming complexity in the library rather than shifting an unnecessary requirement to the user. If we assume duplication now, we can still refactor later and users will not be affected. If we offer an interface that takes `&mut` now (which is a relatively strong requirement) and manage to remove it down the road, we will force our users to change their code.

tomaka:
Another argument in favour of `&mut` is that the `peer_state_query` method you added doesn't have the logic that lazily updates the k-bucket in case of a timeout.
Because of that, the getters can return a state that is not "the correct one", for a certain definition of "correct". I have mixed feelings about that, and I also don't know any clear path forward.

I don't think `&mut` is a strong requirement. Having a non-mut `Swarm` is virtually useless, and therefore I don't see any situation where the user wouldn't have a `&mut` reference of the behaviour.

elferdo:
> Another argument in favour of `&mut` is that the `peer_state_query` method you added doesn't have the logic that lazily updates the k-bucket in case of a timeout.

Should it? I would expect the new query methods to just return a current snapshot of the table.

> Because of that, the getters can return a state that is not "the correct one", for a certain definition of "correct". Calling `entry` and calling `peer_state_query` (which to me are supposed to be equivalent) can return two different states. I have mixed feelings about that, but I also don't know any clear path forward.

I agree that, in terms of information returned both should return the same status. My understanding of `entry` is that, for an entry not already in the DHT, the user has the possibility of inserting it straight away. For the new method I understood that we just want to know what is in the DHT. I think I'm missing the part with the timeout, will check the code again.

> I don't think `&mut` is a strong requirement. Having a non-mut `Swarm` is virtually useless, and therefore I don't see any situation where the user wouldn't have a `&mut` reference of the behaviour.

I think this argument couples the design of the interface with its use. Agreed, if kad is only used in `Swarm`, then there may not be any other possibility. If for some reason we would like to reuse kad somewhere else, then the mutability requirement would be already there.

tomaka:
> Should it? I would expect the new query methods to just return a current snapshot of the table.

That's the issue I don't know how to properly solve in terms of the API.

The situation is: we tell the k-bucket "if node A hasn't responded in 20 seconds, then replace it with B". Then if we immediately call `entry(A)`, it will still be there. If we wait 21 seconds and call `entry(A)`, it will no longer be there.

That's a bad way to do it, and in my opinion it should be changed to have some explicit `update` method. But the details of that are not straight-forward I think.

elferdo:
That looks like something more complex than the `&mut`/code duplication discussion. Are we happy if I remove duplication for this PR and leave the timeout issue for another one?

tomaka:
> Are we happy if I remove duplication for this PR and leave the timeout issue for another one?

Yes!

elferdo:
I'm trying to replace `kbuckets::PeerState` with the original `kbuckets::Entry` and I'm stuck trying to find a way to iterate over nodes and generate a collection of `Entry`s that hold a `&mut` to the parent. It makes sense, because if that were possible we would end up with many objects holding mutable references to the `KBucketsTable`. Please correct me if I'm wrong, and if you have an idea how to achieve this let me know.

Alternatively, what we can do to avoid code redundancy is make the `Entry` structures generic on the type of `parent`, and let it be mutable or inmutable depending on context:

```rust
/// Represents an entry in a k-bucket.                                                                                                                                                                                                                                                    
pub struct EntryInKbucketConn<'a, TParent, TPeerId> {                                                                                                                                                                                                                                     
    //parent: &'a mut KBucketsTable<TPeerId, TVal>,                                                                                                                                                                                                                                   
    parent: TParent,                                                                                                                                                                                                                                                                      
    peer_id: &'a TPeerId,                                                                                                                                                                                                                                                                 
}                                                                                                                                                                                                                                                                                         
```    
This is relatively straightforward, but then type signatures look cumbersome:

```rust
/// Returns an iterator to all the peer IDs in the bucket, including the pending nodes.                                                                                                                                                                                               
pub fn entries<'a>(&'a self) -> impl 'a + Iterator<Item = Entry<'a, &'a KBucketsTable<TPeerId, TVal>, TPeerId>>{                                                                                                                                                                      
```

tomaka:
> I'm stuck trying to find a way to iterate over nodes and generate a collection of Entrys that hold a &mut to the parent.

Ah yeah, I didn't anticipate this.
Considering that entries can modify the table, it would be a bad idea to allow this. Calling a method on one entry could modify the state of another entry, and that would be quite bad.

I guess that this is a good motivation for restoring the code duplication.
Alternatively, we could make `entries()` return a list of `PeerId`s, and the user has to access them one by one if they want more information.

protocols/kad/src/kbucket.rs
```diff
@@ -293,11 +293,18 @@ where
     }
 
     /// Returns an iterator to all the peer IDs in the bucket, without the pending nodes.
-    pub fn entries_not_pending(&self) -> impl Iterator<Item = (&TPeerId, &TVal)> {
+    pub fn entries_not_pending(&self) -> impl Iterator<Item = &TPeerId> {
```
tomaka: Why change this? It is a strictly less good version.

protocols/kad/src/behaviour.rs
```diff
@@ -268,10 +300,27 @@ impl<TSubstream> Kademlia<TSubstream> {
     }
 
     /// Returns an iterator to all the peer IDs in the bucket, without the pending nodes.
-    pub fn kbuckets_entries(&self) -> impl Iterator<Item = &PeerId> {
-        self.kbuckets.entries_not_pending().map(|(kad_hash, _)| kad_hash.peer_id())
+    pub fn kbuckets_entries_not_pending<'a>(&'a self) -> impl 'a + Iterator<Item = &PeerId> {
```
tomaka: I don't think we need both a version with pending nodes and non-pending nodes.
Pending nodes aren't part of the k-buckets in my opinion, so we should either have a method that considers them not-part-of-buckets, or rename `kbuckets_entries_not_pending` to `kbuckets_entries` and `kbuckets_entries` to `kbuckets_entries_pending`.

protocols/kad/src/kbucket.rs
elferdo: Well, I changed this in one of the intermediate steps, and it turns out that the only function consuming it was actually discarding the value, so it didn't look harmful. But yes, as it is right now, it is a poorer version.

protocols/kad/src/behaviour.rs
elferdo: The renaming scheme sounds easiest to me right now.

tomaka: Well, option #1 consists in removing a method.

romanb:
This is very much out of date but I believe most if not all of what was done here is already provided by the refactored APIs (in particular since https://github.com/libp2p/rust-libp2p/pull/1117 which included substantial changes). I suggest to close this PR and update the associated ticket with details about what is still missing in the context of the current API.

Issue #786, opened 2018-12-17
Swarm has no way to refuse incoming connections


tomaka:
If we want to ban peers, we should be able to refuse incoming connections by their peer ID.

This is a design issue, and thoughts need to be put in for how to fix it.

montekki:
Is there anything else to be done here?

PR #1065, opened 2019-04-16
swarm: ban connections based on peerid

montekki:
Tries to solve #786


examples/chat.rs
```diff
@@ -118,6 +118,13 @@ fn main() {
         libp2p::Swarm::new(transport, behaviour, local_peer_id)
     };
 
+    if let Some(to_ban) = std::env::args().nth(1) {
```
tomaka: I suppose you put this to test if it worked, but I don't think it should be part of this example.

core/src/swarm/swarm.rs
```diff
@@ -210,6 +213,10 @@ where TBehaviour: NetworkBehaviour<ProtocolsHandler = THandler>,
             me.external_addrs.push(addr);
         }
     }
+
+    pub fn ban_peer_id(&mut self, peer_id: PeerId) {
+        self.banned_peers.insert(peer_id);
```
tomaka: Should probably disconnect any existing connection as well.

core/src/swarm/swarm.rs
```diff
@@ -73,6 +74,8 @@ where
     /// List of multiaddresses we're listening on, after account for external IP addresses and
     /// similar mechanisms.
     external_addrs: SmallVec<[Multiaddr; 8]>,
+
+    banned_peers: HashSet<PeerId>,
```
tomaka:
```suggestion
    /// List of nodes for which we deny any incoming connection.
    banned_peers: HashSet<PeerId>,
```

core/src/swarm/swarm.rs
```diff
@@ -210,6 +213,10 @@ where TBehaviour: NetworkBehaviour<ProtocolsHandler = THandler>,
             me.external_addrs.push(addr);
         }
     }
+
+    pub fn ban_peer_id(&mut self, peer_id: PeerId) {
```
tomaka: Needs a doc-comment as well.

core/src/swarm/swarm.rs
```diff
@@ -258,7 +265,11 @@ where TBehaviour: NetworkBehaviour<ProtocolsHandler = THandler>,
                     self.behaviour.inject_node_event(conn_info.peer_id().clone(), event);
                 },
                 Async::Ready(RawSwarmEvent::Connected { conn_info, endpoint }) => {
-                    self.behaviour.inject_connected(conn_info.peer_id().clone(), endpoint);
+                    if self.banned_peers.contains(conn_info.peer_id()) {
+                        self.raw_swarm.peer(conn_info.peer_id().clone()).into_connected().unwrap().close();
```
tomaka: We try to do like Substrate/Polkadot and use `expect` with a reason.
```suggestion
                        self.raw_swarm.peer(conn_info.peer_id().clone())
                            .into_connected()
                            .expect("the RawSwarm just notified us that we were connected; QED")
                            .close();
```

tomaka:
I think that it could also be a good idea to ignore any `DialPeer` action that we receive containing a banned node.

core/src/swarm/swarm.rs
```diff
@@ -210,6 +214,17 @@ where TBehaviour: NetworkBehaviour<ProtocolsHandler = THandler>,
             me.external_addrs.push(addr);
         }
     }
+
+    /// Bans a peer by it's peer ID
```
tomaka:
```suggestion
    /// Bans a peer by its peer ID.
    ///
    /// Any incoming connection and any dialing attempt will immediately be rejected.
    /// This function has no effect is the peer is already banned.
```

core/src/swarm/swarm.rs
```diff
@@ -210,6 +214,17 @@ where TBehaviour: NetworkBehaviour<ProtocolsHandler = THandler>,
...
+    pub fn ban_peer_id(&mut self, peer_id: PeerId) {
+        self.banned_peers.insert(peer_id.clone());
+        self.raw_swarm.peer(peer_id).into_connected().map(|c| c.close());
+    }
+
+    /// Removes a peer from ban by it's peer ID
```
tomaka:
```suggestion
    /// Unbans a peer.
```

core/src/swarm/swarm.rs
```diff
@@ -258,7 +273,11 @@ where TBehaviour: NetworkBehaviour<ProtocolsHandler = THandler>,
                     self.behaviour.inject_node_event(conn_info.peer_id().clone(), event);
                 },
                 Async::Ready(RawSwarmEvent::Connected { conn_info, endpoint }) => {
-                    self.behaviour.inject_connected(conn_info.peer_id().clone(), endpoint);
+                    if self.banned_peers.contains(conn_info.peer_id()) {
+                        self.raw_swarm.peer(conn_info.peer_id().clone()).into_connected().map(|c| c.close());
```
tomaka: I actually preferred the `expect` here. If `into_connected` returns `None` then there's a very serious bug and it's preferable to panic rather than silently continue in an invalid state.

PR #1067, opened 2019-04-17
Remove libp2p-ping keep-alive functionality.

romanb:
This seemed to be a misunderstanding: `libp2p-ping` is not supposed to keep connections alive, only to add another condition (failed pings) for terminating them. This is corrected here along with further tests.

---

**Old description**: This was an oversight: The current connection timeout is always short of one `interval`, because the "countdown" begins with the last received or sent pong (depending on the policy). In effect, the current default config has a remaining connection timeout of 5 seconds (20 - 15) from the point when a ping is sent, instead of the expected 20 seconds of the configured `timeout`.

Instead, the "countdown" of the connection timeout should always begin with the next scheduled ping. That also makes all configurations valid and easier to understand, avoiding pitfalls.

The important properties of the ping handler are now tested to hold for all configurations, in particular:

  * The next ping must be scheduled no earlier than the ping interval and no later than the connection timeout.

  * The "countdown" for the connection timeout starts on the next ping, i.e. the full connection timeout remains at the instant when the next ping is sent. ~~


tomaka:
Oh wait, I didn't realize that in the previous PR.
`connection_keep_alive()` should always return `Now`, otherwise the connection will never close.
`connection_keep_alive()` is essentially a mechanism for garbage-collecting connections that we don't need. Pinging is not a motivation for keeping a connection alive.

romanb:
> Oh wait, I didn't realize that in the previous PR.
> `connection_keep_alive()` should always return `Now`, otherwise the connection will never close.

Returning `Now` would mean that the connection would be closed immediately regardless, you cannot possibly mean that? The old implementation relied on the `OneShotHandler` always returning `Forever`. I cannot confirm that the connection is never closed (I'm referring to [here](https://github.com/libp2p/rust-libp2p/blob/master/core/src/protocols_handler/node_handler.rs#L260-L264) and [here](https://github.com/libp2p/rust-libp2p/blob/master/core/src/protocols_handler/node_handler.rs#L286-L300)) and it wouldn't even make sense, because it would mean returning `KeepAlive::Until` with an instant in the past has no effect, which is surely not the case, I've seen it resulting in the connection being closed.

> `connection_keep_alive()` is essentially a mechanism for garbage-collecting connections that we don't need. Pinging is not a motivation for keeping a connection alive.

I'm not sure what you're saying, but however you call it, it is about closing connections that are deemed no longer useful for some reason. Application-layer connection keep-alive and measuring the RTT are the two use-cases for pinging that I understood from the old implementation. Previously it was simply the case that when a ping timed out, returning an error from `poll()` was (ab)used to have the connection be closed, in combination with letting the `OneShotHandler` always return `KeepAlive::Forever`, i.e. a hard-coded variant of the new default configuration with `max_timeouts=1`. This was simply made configurable and the keep-alive based on the existing `connection_keep_alive()` functionality, which seemed much more appropriate (to me anyway).

Can you elaborate more on the issues you see with the existing implementation? In any case, this PR is a patch for what is already in master, so I think it should go in regardless of how things are changed afterwards.

tomaka:
> Returning Now would mean that the connection would be closed immediately regardless, you cannot possibly mean that?

I do mean that!

> The old implementation relied on the OneShotHandler always returning Forever.

The `OneShotHandler` returns `Forever` if we are currently opening an outbound substream, or `Until(some_inactivity_timeout)` if we are not.
Part of the reasons why `Ping` had to switch away from the `OneShotHandler` is this behaviour.

>  Application-layer connection keep-alive and measuring the RTT are the two use-cases for pinging that I understood from the old implementation.

The primary use-case of `Ping` is to *check* whether a connection is still alive, not *force* it to stay alive.

At least in the Unix world, the timeout of a TCP connection is around an hour. That means, we need one hour without any packet being transferred on that connection for the kernel to close due it inactivity. The objective of `Ping` is to detect zombie connections earlier than that.
Consequently, a connection that is only active because of `Ping` is not useful.

romanb:
> The primary use-case of `Ping` is to _check_ whether a connection is still alive, not _force_ it to stay alive.

Ok, I think I see the misunderstanding now. Let me try to phrase how the ping protocol should behave: By always returning `KeepAlive::now` from `connection_keep_alive`, it should constantly tell the Swarm (or rather the `NodeHandler`): "The connection can be closed if nothing else is going on." while at the same time only adding an *additional* condition for aborting the connection, namely one (or possibly configurably many) ping timeouts. Correct?

I really did think one purpose was to keep the connection alive for future traffic even if it is idle, not just to add an additional condition for aborting it, based on a misunderstanding of how `KeepAlive` and the `OneShotHandler` work.

If I now have the right understanding, I can repurpose this PR to make the necessary changes.

tomaka:
> Correct?

Yes!

Return an error in `poll()` is a guarantee that the connection will be closed, while returning `Now` in `connection_keep_alive()` isn't.

romanb:
Ok, the PR has been updated accordingly. In this context, I did not see a use for the `PingPolicy` any longer, despite what was discussed in an earlier PR, so I removed it. Let me know if you still see a use for it. I hope this is now the desired implementation.

core/src/protocols_handler/mod.rs
```diff
@@ -150,9 +150,10 @@ pub trait ProtocolsHandler {
     ///
     /// This method is called by the `Swarm` after each invocation of
     /// [`ProtocolsHandler::poll`] to determine if the connection and the associated
-    /// `ProtocolsHandler`s should be kept alive and if so, for how long.
+    /// `ProtocolsHandler`s should be kept alive as far as this handler is concernced
```
tomaka:
```suggestion
    /// `ProtocolsHandler`s should be kept alive as far as this handler is concerned
```

core/src/protocols_handler/mod.rs
```diff
@@ -150,9 +150,10 @@ pub trait ProtocolsHandler {
...
+    /// and if so, for how long.
     ///
-    /// Returning [`KeepAlive::Now`] indicates that the connection should be
+    /// Returning [`KeepAlive::Now`] indicates that the connection can be
```
tomaka: "can vs should" a bit debatable. The value returned here is authoritative from the point of view of the swarm/nodehandler, so from this aspect it should be "should" or even "will".
However I see your point: since handlers are merged together with other handlers by the user, each individual handler doesn't know if returning `Now` is actually going to close the connection.

romanb:
@tomaka Note that as it currently is, at least on a fast enough machine like mine, this actually still doesn't work as desired because `Delay` only guarantees a precision of 1 millisecond, which effectively sends the swarm / node handler into a busy loop, never closing the connection, because less than one millisecond passed between `Delay::new(Instant::now())` and the polling of the delay later, and the delay gets recreated every time. Only after https://github.com/libp2p/rust-libp2p/pull/1067/commits/1da47349670d0ce080ad859bbd2644bdc6627cd7 can I actually see in the ping example that the connection is immediately closed after the first ping.

romanb:
I created https://github.com/libp2p/rust-libp2p/pull/1072 which this PR is now considered blocked on (or some other proper fix for the underlying problem).

PR #1068, opened 2019-04-18
Fix `multiaddr::util::BytesWriter`.

twittner:
The current implementation does not properly reserve enough space for `Write::write_all` to succeed. The property test has been improved to trigger that issue.

### DIFF
--- a/misc/multiaddr/src/util.rs
+++ b/misc/multiaddr/src/util.rs
@@ -1,4 +1,4 @@
-use bytes::{BufMut, BytesMut};
+use bytes::BytesMut;
 
 /// An [`io::Write`] impl for [`BytesMut`].
 ///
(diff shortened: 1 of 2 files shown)


misc/multiaddr/src/util.rs
```diff
@@ -9,25 +9,8 @@ pub(crate) struct BytesWriter(pub(crate) BytesMut);
...
-                Ok(0) => return Err(std::io::ErrorKind::WriteZero.into()),
-                Ok(n) => buf = &buf[n ..],
-                Err(e) => if e.kind() != std::io::ErrorKind::Interrupted {
-                    return Err(e)
-                }
-            }
-        }
-        Ok(())
+        self.0.extend_from_slice(src);
+        Ok(src.len())
```
romanb: Does the `BytesWriter` type now actually still serve a purpose?

romanb: Oh, I guess because `BytesMut` does not implement this trait.

PR #1071, opened 2019-04-18
Switch to wasm-timer

### LINKED ISSUE #900: Abstract over tokio-timer in order to compile for the browser
I don't know exactly *how*, but we should not use `tokio-timer` when compiling for WASM/Emscripten, as this will not work.

tomaka:
Close #900

The `wasm-timer` crate abstracts over `tokio_timer` and `std::time::Instant` and make them work in the browser.

We replace all usages of `tokio_timer` and `Instant` with this crate.

This might be controversial, as this change is quite invasive.
However, once #1029, #1070, and this PR are all merged, it should be possible to run a WASM example that uses the master branch.


tomaka:
While the idea of WASI seems cool, it doesn't really look ready to me yet (example: https://github.com/rust-lang/rust/blob/bfb443eb1de484fde141fa9090a9f4291cbe60a5/src/libstd/sys/wasi/net.rs#L14-L106).
We additionally would need some support from browsers to actually implement WASI, and that doesn't seem to be in the process of happening.

PR #1072, opened 2019-04-18
Fix connection & handler shutdown when using `KeepAlive::Now`.

romanb:
This is supposed to address a bug revealed in the context of https://github.com/libp2p/rust-libp2p/pull/1067, namely that `Delay::new(Instant::now())` is never immediately ready (not even with a multi-threaded runtime and additional artificial delay between Delay creation and polling - I'm not very familiar with the internals of `tokio-timer` yet and appreciate any illumination someone can provide). 

The result is that `KeepAlive::Now` currently has no effect, since the delay is re-created on every execution of `poll()` in the `NodeHandlerWrapper`. It can also send the node handler into a busy-loop, since every newly created Delay with `Instant::now()` will trigger a task wakeup, which creates a new `Delay` with `Instant::now()`, and so forth.

Evidently, two tests were even depending on `KeepAlive::Now` not working, namely:

  * `raw_swarm_simultaneous_connect`: That was fixed here by changing the `TestHandler` to return `KeepAlive::Forever` instead of `KeepAlive::Now`.

  * `periodic_id_works`: The `PeriodicIdHandler` returns `KeepAlive::Now` as soon as it got the first `Identified` event, which, if `KeepAlive::Now` works, can result in the connection being closed immediately afterwards. Hence the other peer will not get the `Identified` event its waiting for. Therefore this test cannot wait for both peers to receive that event. The test was changed to be successful as long as either peer receives a correct event.

This is a good indication that fixing this problem has a high likelihood to break existing applications.

This PR addresses the problem by removing the use of `Delay::new(Instant::now())` for "immediate" connection shutdown entirely. An important assumption is thereby that as long as the node handler has non-empty `negotiating_in` and `negotiating_out`, the handler is not dependent on such a Delay for task wakeup. At least the tests are happy with that.

PS: I'd really like to rename `KeepAlive::Now` to `KeepAlive::No` and `KeepAlive::Forever` to `KeepAlive::Yes`, as I find the current naming quite irritating but maybe this is just me. I see the enum constructors as representing possible answers to the question "Should I keep the connection alive?" asked by the node handler to the protocols handler and `Now` is a very weird answer.


tomaka:
> This is a good indication that fixing this problem has a high likelihood to break existing applications.

I think that's fine. We might reveal bugs, but it's for the greater good.

> PS: I'd really like to rename KeepAlive::Now to KeepAlive::No and KeepAlive::Forever to KeepAlive::Yes, as I find the current naming quite irritating but maybe this is just me. I see the enum constructors as representing possible answers to the question "Should I keep the connection alive?" asked by the node handler to the protocols handler and Now is a very weird answer.

I agree with that.

PR #1073, opened 2019-04-19
&mut self -> &mut Self

tomaka:
cc #1065


tomaka:
> What motivates basing the Swarm API on associated functions in this way?

`Swarm` has a `Deref`s to the `NetworkBehaviour`, and it is bad form to have at the same time regular methods and a deref (see `Arc` for example).

Normally the `Swarm`'s inherent methods aren't supposed to be called (eg. you're not supposed to dial, as the network behaviour does that already; initially it was the same for listening), which is why I did that.

This is of course a debatable decision, but for now I think it's more important to be consistent.

Issue #739, opened 2018-12-06
Many warnings about this


daogangtang:
like this.
```
warning: use of deprecated item 'tokio_io::codec::length_delimited': Moved to tokio-codec
  --> protocols/secio/src/handshake.rs:46:5
   |
46 | use tokio_io::codec::length_delimited;
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: #[warn(deprecated)] on by default
```

dvdplm:
Yes, it's a bit frustrating. The issue is [here](https://github.com/tokio-rs/tokio/issues/680) and a PR to fix it [here](https://github.com/tokio-rs/tokio/pull/685).

tomaka:
This has been fixed through #1328

PR #1075, opened 2019-04-20
secio: fix deprecated length_delimited warnings

montekki:
the current  `secio` code uses the `tokio_io::codec::length_delimited::Framed` that is being deprecated due to a move to `tokio::codec`, hence a ton of build warnings. However, the `Encoder` implementation for new `LengthDelmitedCodec` now has `Item = Bytes` instead of `Item = BytesMut` as in the older version which changes things a bit for fullcodec so `EncoderMiddleware`'s underlying `Sink: Item` now changes to `Bytes`.

This change probably has to be done at some point anyway, however I am not sure about bringing the whole `tokio` dependency in. That's needed because `LengthDelimitedCodec` lives there and not in `tokio-codec`.

### DIFF
--- a/protocols/secio/Cargo.toml
+++ b/protocols/secio/Cargo.toml
@@ -25,6 +25,7 @@ ctr = "0.3"
 lazy_static = "1.2.0"
 rw-stream-sink = { version = "0.1.1", path = "../../misc/rw-stream-sink" }
 tokio-io = "0.1.0"
+tokio = "0.1.1"
 sha2 = "0.8.0"
 hmac = "0.7.0"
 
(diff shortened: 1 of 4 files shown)


romanb:
I think the reason why these deprecation warnings have not been addressed yet, is that they are blocked by https://github.com/rust-lang/cargo/issues/4866. I.e. pulling in the entirety of `tokio` will not allow compiling for `wasm32` and trying to use only the `codec` feature doesn't work correctly due to that issue. See also https://github.com/tokio-rs/tokio/issues/680#issuecomment-453288927 and https://github.com/libp2p/rust-libp2p/issues/739.

montekki:
Oh, I wasn't aware of those, that's unfortunate. Should this be closed in this case?

romanb:
I think its fine to leave it open, but I labeled it as blocked.

PR #1079, opened 2019-04-22
Add a OneSubstreamMuxer

### LINKED ISSUE #488: Add a DummyMuxer that provides only one substream
The new swarm requires an implementation of `StreamMuxer` to be passed to it.
We should give users the possibility to not use a muxer by providing a `DummyMuxer` that wraps around a stream and provides one substream (the stream itself).

The constructor of `DummyMuxer` needs to be passed an `Endpoint` that determines whether the one substream is inbound or outbound.

tomaka:
Close #488


core/src/muxing/one_substream.rs
```diff
@@ -0,0 +1,137 @@
...
+use crate::{Endpoint, muxing::StreamMuxer};
+use futures::prelude::*;
+use parking_lot::Mutex;
+use std::{io, sync::atomic::{AtomicBool, Ordering}};
+use tokio_io::{AsyncRead, AsyncWrite};
+
+/// Implementation of `StreamMuxer` that allows only one substream on top of a connection,
+/// yielding the connection itself.
+///
+/// Nothing happens on the wire.
```
romanb: That seems a bit misleading. Surely it only means that there is no substream protocol used on the wire?

tomaka: Yes. I don't know how to phrase that.

romanb: How about `No substream protocol used on the wire.`?

tomaka: I changed it in the latest commit. Let me know.

romanb: I think you forgot to check-in the new file?

tomaka: Oops

PR #1083, opened 2019-04-23
muxing: adds an error type to streammuxer

### LINKED ISSUE #894: Add a custom Error type in StreamMuxer
So that we can have a more strongly-typed error type.
This is however not trivial because we're implementing the `Read` and `Write` traits on substreams, which forces us to use a `std::io::Error` for reading and writing.

This issue is about figuring out the details of what to change, if anything.

Related to #845

montekki:
Tries to add an `Error` associated type to `StreamMuxer` trait. Implementations like `AsyncRead` place an `Into<io::Error>` on this type everywhere.

Fixes #894


core/src/muxing.rs
```diff
@@ -83,6 +83,9 @@ pub trait StreamMuxer {
     /// Future that will be resolved when the outgoing substream is open.
     type OutboundSubstream;
 
+    /// Error type of the muxer
+    type Error;
```
tomaka: By curiosity, did you try adding a `Into<io::Error>` bound here? This might make it possible to remove lots of requirements from elsewhere.

examples/chat.rs
```diff
@@ -66,7 +66,7 @@ fn main() {
     let local_peer_id = PeerId::from(local_key.public());
     println!("Local peer id: {:?}", local_peer_id);
 
-    // Set up a an encrypted DNS-enabled TCP Transport over the Mplex and Yamux protocols
+    // jet up a an encrypted DNS-enabled TCP Transport over the Mplex and Yamux protocols
```
tomaka:
```suggestion
    // Set up a an encrypted DNS-enabled TCP Transport over the Mplex and Yamux protocols
```

core/src/muxing.rs
montekki: actually didn't think of it, thanks

PR #1085, opened 2019-04-24
Pass the ConnectedPoint to into_handler()

### LINKED ISSUE #1054: Pass the ConnectedPoint to into_handler()
In addition to passing the connection info, we should also pass the `ConnectedPoint` when turning a handler builder into a proper handler.

tomaka:
Fix #1054 

This works by having the `RawSwarm` pass `(TConnInfo, ConnectedPoint)` instead of `TConnInfo` to the lower layers, so that lower layers don't know who we are connected to (as should be the case).

There's one design issue remaining: we create a dummy protocols handler in `Swarm` in order to determine which protocols we support. In this PR, I pass a dummy `ConnectedPoint` when doing so. This is obviously a bad solution that should be fixed by not creating a dummy protocols handler.

### DIFF
--- a/core/src/nodes/raw_swarm.rs
+++ b/core/src/nodes/raw_swarm.rs
@@ -61,7 +61,7 @@ where
     listeners: ListenersStream<TTrans>,
 
     /// The nodes currently active.
-    active_nodes: CollectionStream<TInEvent, TOutEvent, THandler, InternalReachErr<TTrans::Error, TConnInfo>, THandlerErr, (), TConnInfo, TPeerId>,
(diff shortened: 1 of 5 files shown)


tomaka:
Ready for review now that #1099 is in.

tomaka:
> Have you considered introducing a struct, e.g. along the lines of ConnectionContext<TConnInfo> { info: TConnInfo, point: ConnectedPoint }?

I agree with that. I admit I used a tuple because of laziness and because the code of `RawSwarm` needs some reorganization anyway to make it more simple.

tomaka:
> Have you considered introducing a struct, e.g. along the lines of ConnectionContext<TConnInfo> { info: TConnInfo, point: ConnectedPoint }?

This PR only adds tuples internally in the `RawSwarm`, so I guess it's fine and can be changed later.
Will merge.

PR #1086, opened 2019-04-24
Change `PingHandler::connection_keep_alive`.

twittner:
Instead of `KeepAlive::No` we ask to keep the connection alive for at least one second. The reason for this change is that `tokio_timer::Delay` only guarantees millisecond precision and while we try to make the `OutboundSubstreamRequest` as soon as possible, the handler may in fact return `NotReady` before 1 millisecond has passed. This in turn will make the `NodeHandler` close the connection immediately and the pinging never even begins.

### DIFF
--- a/protocols/ping/src/handler.rs
+++ b/protocols/ping/src/handler.rs
@@ -138,6 +138,8 @@ impl Error for PingFailure {
 pub struct PingHandler<TSubstream> {
     /// Configuration options.
     config: PingConfig,
+    /// Beginning of time for this `PingHandler`.
+    epoch: Instant,
     /// The timer for when to send the next ping.
     next_ping: Delay,
     /// The pending results from inbound or outbound pings, ready
@@ -151,9 +153,11 @@ pub struct PingHandler<TSubstream> {
 impl<TSubstream> PingHandler<TSubstream> {
     /// Builds a new `PingHandler` with the given configuration.
     pub fn new(config: PingConfig) -> Self {
+        let now = Instant::now();
         PingHandler {
             config,
-            next_ping: Delay::new(Instant::now()),
+            epoch: now,
+            next_ping: Delay::new(now),
             pending_results: VecDeque::with_capacity(2),
             failures: 0,
             _marker: std::marker::PhantomData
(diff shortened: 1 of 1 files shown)


tomaka:
I don't understand the change.

> The reason for this change is that tokio_timer::Delay only guarantees millisecond precision

Which `Delay` are you referring to? The one in `PingHandler`?

twittner:
> Which `Delay` are you referring to? The one in `PingHandler`?

Yes, in the initial next_ping: `Delay::new(Instant::now())`. But I just realised that the actual problem is a race between the pending pong and the time it takes for the node handler to query `PingHandler::connection_keep_alive`.

twittner:
This needs more thinking.

tomaka:
The `Ping` behaviour isn't supposed to keep connections alive. Its role is to kill connections that aren't responsive, but not to keep connections alive.

twittner:
So, the actual problem is the result of this sequence of events:

1. `PingHandler` request a new outbound substream.
2. The NodeHandler returns a `NodeHandlerEvent::OutboundSubstreamRequest`
3. `NodeHandler::poll` is invoked again, *before* `NodeHandler::inject_substream` is called which in turn
4. calls `PingHandler::poll` again which is `NotReady`.
5. Finally `PingHandler::connection_keep_alive` is asked, which returns `KeepAlive::No` and since the `NodeHandler`'s `negotiating_in` and `negotiating_out` collections are empty, `NodeHandlerWrapperError::UselessTimeout` is returned, terminating the connection.

twittner:
> The `Ping` behaviour isn't supposed to keep connections alive. Its role is to kill connections that aren't responsive, but not to keep connections alive.

Yes, I know. The PR was premature and a mere workaround. I have not fully understood the actual problem then.

romanb:
@tomaka The problem that currently exists is that the ping integration test is a bit unstable (i.e. can hang) for two reasons:

  1. If `next_ping.poll()` returns `NotReady` on the first call. This can currently happen due to initializing `next_ping` with `Delay::new(Instant::now())`. That can be fixed by creating the delay with an instant in the past.

  2. The second problem can occur as a result of the race mentioned by @twittner above.

In other words, while it may be technically acceptable / correct that the connection may be closed at any such time, it is thus currently not possible to write stable tests, such as the current ping integration test, for protocols using `KeepAlive::No`, e.g. the ping integration test relies on at least one ping being sent.

romanb:
@tomaka Is it fully intentional that the `queued_dial_upgrades` are not considered for the [shutdown decision](https://github.com/libp2p/rust-libp2p/blob/master/core/src/protocols_handler/node_handler.rs#L310), i.e. such that they must be empty? Doing so would solve problem `2.`.

tomaka:
Oh, if it's for tests I understand it. Otherwise the fact that a connection can be closed at any point is "working as intended" to me.

> Is it fully intentional that the queued_dial_upgrades are not considered for the shutdown decision, i.e. such that they must be empty? Doing so would solve problem 2..

No that's not intentional. However it would also be more correct, in my opinion, to remove this mechanism entirely, in other words to not block the shutdown on having substreams being negotiated (as is the case right now).
Instead the test could also be fixed by having some sort of "forced keep-alive" behaviour, or by having a wrapper around `NetworkBehaviour` that keeps connections alive.

PR #1093, opened 2019-04-28
Allow a path for WS multiaddresses

tomaka:
Adds the possibility to add a URL path to multiaddress that contain `ws` or `wss`.

I'm doing this as a draft because I don't really know what to do with the string representation of these. At the moment, outputting such an address works, but parsing it back will fail.


twittner:
How about adding a generic `Path` protocol (family) which length-prefixes its value? Paths are assumed to be UTF-8 encoded and the length denotes the total length in bytes (base 10):

```
Path   <- 'path:' Length
Length <- [0-9]+
```

For example:

- `/ip4/127.0.0.1/tcp/8000/ws/path:5/hello`
- `/ip4/127.0.0.1/tcp/8000/ws/path:23/this/is/some/path/value/p2p/QmcgpsyWgH8Y8ajJz1Cu72KnS5uo2Aa2LpzU7kinSupNKC`
- `/path:45/C:\Documents and Settings\All Users\Documents`
- `/path:0`

This way paths can be combined with other protocols. Possible drawbacks are that the end of the path may be difficult to spot with longer paths containing slashes and parsing becomes slightly more complicated.

tomaka:
There's already some discussion open about that: https://github.com/multiformats/multiaddr/issues/87

To me the syntax is extremely arbitrary and not important at all. What I need is a decision, and one that all implementers of libp2p agrees with.

twittner:
> There's already some discussion open about that

I was not aware.

> To me the syntax is extremely arbitrary and not important at all.

My suggestion was not about syntax per se, but about avoiding the need for ad-hoc extensions like `ws_path` and `wss_path`. If there is already some discussion how to solve this, fine. Please disregard the suggestion.

tomaka:
Ready for review.
I went for `/ip4/1.2.3.4/tcp/5/x-parity-ws/%2ffoo%2fbar` as a translation of `ws//1.2.3.4:5/foo/bar`.

tomaka:
I don't want a solution that's generic. This PR doesn't try to solve the question of having a nice human-readable string, because every single human on this planet has different opinion on what is nice-looking or not. I just want something that works and is straight-forward to parse.

PR #1095, opened 2019-04-30
Replace PeerId with Multihash for interface consistency

elferdo:
In `Kademlia`, two similar methods: `add_providing` and `remove_providing` take different types. There is no need for `add_providing` to take a `PeerId`, it can take a `&Multihash` just like `remove_providing`.

### DIFF
--- a/protocols/kad/src/behaviour.rs
+++ b/protocols/kad/src/behaviour.rs
@@ -293,7 +293,7 @@ impl<TSubstream> Kademlia<TSubstream> {
     ///
     /// The actual meaning of *providing* the value of a key is not defined, and is specific to
     /// the value whose key is the hash.
-    pub fn add_providing(&mut self, key: &Multihash) {
+    pub fn add_providing(&mut self, key: Multihash) {
         self.providing_keys.insert(key.clone());
         let providers = self.values_providers.entry(key.clone()).or_insert_with(Default::default);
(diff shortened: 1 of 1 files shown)


protocols/kad/src/behaviour.rs
```diff
@@ -293,9 +293,9 @@ impl<TSubstream> Kademlia<TSubstream> {
     ///
     /// The actual meaning of *providing* the value of a key is not defined, and is specific to
     /// the value whose key is the hash.
-    pub fn add_providing(&mut self, key: PeerId) {
-        self.providing_keys.insert(key.clone().into());
-        let providers = self.values_providers.entry(key.into()).or_insert_with(Default::default);
+    pub fn add_providing(&mut self, key: &Multihash) {
```
tomaka: What about `Multihash` instead of `&Multihash`, so that we remove the `clone`?
```suggestion
    pub fn add_providing(&mut self, key: Multihash) {
```

protocols/kad/src/behaviour.rs
```diff
@@ -293,9 +293,9 @@ impl<TSubstream> Kademlia<TSubstream> {
...
+        self.providing_keys.insert(key.clone());
+        let providers = self.values_providers.entry(key.clone()).or_insert_with(Default::default);
```
```suggestion
        let providers = self.values_providers.entry(key).or_insert_with(Default::default);
```

PR #1096, opened 2019-05-02
Add ranked address collection.

twittner:
Keep external addresses in a ranked collection and ensure iteration is performed in order from highest to lowest rank. The implementation assumes that iteration dominates over modification and optimises for the former.

For context see issue #794.

### DIFF
--- a/core/src/swarm.rs
+++ b/core/src/swarm.rs
@@ -42,6 +42,7 @@
 
 mod behaviour;
 mod swarm;
+mod registry;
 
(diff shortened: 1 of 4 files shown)


tomaka:
Looks good. I would add a piece of documentation to the `ReportObservedAddr` item saying that care should be taken to emit it as a regular rate per node.
What I mean is that if a node was allowed to push an observed address to us, it could still be a problem.

Issue #1053, opened 2019-04-11
Kademlia tests sometimes fail


twittner:
The kademlia tests sometimes fail, [for example](https://circleci.com/gh/libp2p/rust-libp2p/12310):

```
running 13 tests
thread 'tokio-runtime-worker-0' panicked at 'assertion failed: `(left == right)`
  left: `0`,
 right: `2`', protocols/kad/src/behaviour/test.rs:141:29
note: Run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
test behaviour::test::direct_query ... FAILED
test behaviour::test::basic_find_node ... ok
test behaviour::test::unresponsive_not_returned_direct ... ok
test behaviour::test::indirect_query ... ok
test kbucket::tests::basic_closest ... ok
test kbucket::tests::distance_correct_order ... ok
test behaviour::test::unresponsive_not_returned_indirect ... ok
test kbucket::tests::self_distance_zero ... ok
test kbucket::tests::update_local_id_fails ... ok
test query::tests::continue_second_result ... ok
test query::tests::start_by_sending_rpc_to_known_peers ... ok
test query::tests::timeout_works ... ok
test kbucket::tests::full_kbucket ... ok

failures:

---- behaviour::test::direct_query stdout ----
thread 'behaviour::test::direct_query' panicked at 'called `Result::unwrap()` on an `Err` value: Canceled', src/libcore/result.rs:997:5


failures:
    behaviour::test::direct_query

test result: FAILED. 12 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out
```

[Another one](https://circleci.com/gh/libp2p/rust-libp2p/12324):

```
running 13 tests
test behaviour::test::basic_find_node ... ok
test behaviour::test::direct_query ... ok
test behaviour::test::unresponsive_not_returned_direct ... ok
thread 'tokio-runtime-worker-1' panicked at 'assertion failed: `(left == right)`
  left: `0`,
 right: `3`', protocols/kad/src/behaviour/test.rs:190:29
note: Run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
test behaviour::test::indirect_query ... FAILED
test kbucket::tests::basic_closest ... ok
test kbucket::tests::distance_correct_order ... ok
test behaviour::test::unresponsive_not_returned_indirect ... ok
test kbucket::tests::self_distance_zero ... ok
test kbucket::tests::update_local_id_fails ... ok
test query::tests::continue_second_result ... ok
[...]

Issue #341, opened 2018-07-19
Kademlia tries to connect to the local node


tomaka:
A Kademlia query can learn about the local node through the process, and request that we connect to the local node. This is obviously not great.

However the fix is maybe not as straight forward as `if requested_peer == local { continue }`, because opening a Kademlia query with the local node essentially means that we need locally-available information.

This needs some thoughts.

tomaka:
Some update. Now that Kademlia has been correctly reworked, this is no longer an issue.
However we should still add checks to filter out our local node from results coming from other nodes.
In particular, we don't want to emit `Discovered` events and add elements in `untrusted_addresses` about our local node.

romanb:
Closed by https://github.com/libp2p/rust-libp2p/pull/1097.

PR #1097, opened 2019-05-02
Fix self-dialing in Kademlia.

romanb:
Addresses https://github.com/libp2p/rust-libp2p/issues/341 which is the cause for one of the observations made in https://github.com/libp2p/rust-libp2p/issues/1053. However, the latter is not considered to be fully addressed by these changes and needs further investigation.

Currently, whenever a search for a key yields a response containing the initiating peer as one of the closest peers known to the remote, the local node attempts to dial itself. That attempt is [ignored by the Swarm](https://github.com/libp2p/rust-libp2p/blob/master/core/src/swarm/swarm.rs#L185), but the Kademlia behaviour now believes it still has an RPC ongoing which will eventually time out. That timeout delays successful completion of the query. Hence, any query where a remote responds with the ID of the local node takes at least as long as the `rpc_timeout` (currently 8 seconds) to complete.

This problem is fixed here by ensuring that Kademlia never tries to dial the local node. Furthermore, `Discovered` events are no longer emitted for the local node and it is no longer inserted into the `untrusted_addresses` from discovery, as described in #341.

This commit also includes a change to the condition for freezing / terminating a Kademlia query upon receiving a response. Specifically, the condition is tightened such that it only applies if in addition to `parallelism` consecutive responses that failed to yield a peer closer to the target, the last response must also either not have reported any new peer or the number of collected peers has already reached the number of desired results. In effect, a Kademlia query now tries harder to actually return `k` closest peers, i.e. as long as `k == max_results` have not been reached and new nodes are discovered, they are queried even if the distance to the target seems to increase, since these are still considered to be among the `k` closest nodes to the target. If this change is undesirable, note that the tests can only run in a stable fashion with at most 4 nodes, since the node IDs are randomly distributed and thus it often occurs that the iterative lookups move further away from the target, hitting the `no_closer_in_a_row >= parallelism` condition and resulting in the query returning fewer results than the tests expect (this is how I ran into it while adding more variability to the tests).

The tests have been refactored for less duplication and a few additional expectations added.

### DIFF
--- a/protocols/kad/src/behaviour.rs
+++ b/protocols/kad/src/behaviour.rs
@@ -350,6 +350,7 @@ impl<TSubstream> Kademlia<TSubstream> {
         );
     }
 
+    /// Processes discovered peers from a query.
     fn discovered<'a, I>(&'a mut self, query_id: &QueryId, source: &PeerId, peers: I)
     where
         I: Iterator<Item=&'a KadPeer> + Clone
--- a/protocols/kad/src/behaviour.rs
+++ b/protocols/kad/src/behaviour.rs
@@ -250,7 +250,7 @@ impl<TSubstream> Kademlia<TSubstream> {
         let parallelism = 3;
 
         Kademlia {
-            kbuckets: KBucketsTable::new(KadHash::from(local_peer_id), Duration::from_secs(60)),   // TODO: constant
+            kbuckets: KBucketsTable::new(local_peer_id.into(), Duration::from_secs(60)),   // TODO: constant
             queued_events: SmallVec::new(),
             active_queries: Default::default(),
             connected_peers: Default::default(),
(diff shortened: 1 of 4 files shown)


protocols/kad/src/behaviour.rs
```diff
@@ -349,6 +349,32 @@ impl<TSubstream> Kademlia<TSubstream> {
             })
         );
     }
+
+    fn discovered<'a, I>(&'a mut self, query_id: &QueryId, source: &PeerId, peers: I)
```
tomaka: Needs a small doc-comment.

tomaka:
```suggestion
    /// Called every time an RPC query returns some node identities.
    fn discovered<'a, I>(&'a mut self, query_id: &QueryId, source: &PeerId, peers: I)
```

PR #1101, opened 2019-05-05
Some Kademlia code cleanup.

romanb:
Some suggestions for cleanup in `libp2p-kad` without any (intended) change of behaviour, including removal of outdated documentation. I intend to do some more work on the Kademlia implementation and will eventually re-add up-to-date documentation.


protocols/kad/src/query.rs
```diff
@@ -295,67 +291,47 @@ where
         let mut active_counter = 0;
 
         // While iterating over peers, count the number of queries in a row (from closer to further
-        // away from target) that are in the succeeded in state.
-        // Contains `None` if the chain is broken.
-        let mut succeeded_counter = Some(0);
+        // away from target) that are in the succeeded state.
+        let mut succeeded_counter = 0;
```
tomaka: The idea behind this `Option` is that we want to end the query only in the three nearest have succeeded. If for example element 0, 1 and 3 have succeeded but element 2 is in progress, then we should continue the query.
Looking at the old code, though, I think this wasn't working because nothing sets it to `None`.

tomaka:
I think there's a code comment that mentions that, but the code in `upgrade.rs` is intended to be rewritten to be more similar to its equivalent in, say, floodsub.

protocols/kad/src/query.rs
romanb: > The idea behind this `Option` is that we want to end the query only in the three nearest entries have succeeded. If for example elements 0, 1 and 3 have succeeded but element 2 is in progress, then we should continue the query.

Why "three"? The success counter is only ever compared to `max_results`, i.e. the `k` parameter. I assume you mean that this is supposed to implement the following from the paper:

> The lookup terminates when the initiator has queried and gotten responses from the `k` closest nodes it has seen.

and since `closer_peers` in the query can currently grow beyond `k` (I'm not sure yet that's correct or really necessary) the query may terminate earlier than it should right now?

> Looking at the old code, though, I think this wasn't working because I can't see anything that ever sets it to `None`.

Correct, which is why I removed it. I can rework (and actually test) the termination condition later but I would like this PR to remain a small cleanup that does not change behavior. Feel free to create tickets for the changes you'd like to see getting done (the more descriptive, the better)!

romanb:
> I think there's a code comment that mentions that, but the code in upgrade.rs is intended to be rewritten to be more similar to its equivalent in, say, floodsub.

I'm not sure what you mean. There is no `upgrade.rs` in `libp2p-kad`. Do you mean `protocol.rs` and are you perhaps referring to [this TODO](https://github.com/libp2p/rust-libp2p/blob/master/protocols/kad/src/protocol.rs#L137-L139)? If you could create a ticket for that, too, elaborating some more on the desired similarity to the `floodsub` implementation, that would be great. As I said above, I'd like this PR to just remain a small cleanup.

protocols/kad/src/query.rs
tomaka: All you said is correct.

tomaka:
> I'm not sure what you mean. There is no upgrade.rs in libp2p-kad. Do you mean protocol.rs and are you perhaps referring to this TODO?

Indeed!

romanb:
Are there any remaining objections to this PR?

PR #1106, opened 2019-05-06
Integration of Noise Protocol Implementations Generated by Noise Explorer

georgio:
This is still a work in progress. We are attempting to fix a bug possibly located in protocols/noise/src/io.rs

cc: @SymbolicSoft @kaepora @tomaka  @burdges @romanb


protocols/noise/src/lib.rs
```diff
@@ -48,195 +48,275 @@
...
+        if let Ok(prologue) = Message::from_bytes(b"Noise_IX_25519_ChaChaPoly_Blake2s") {
+            let session = NoiseSession::ix(noisesession_ix::NoiseSession::init_session(
+                false,
+                prologue,
+                self.0,
+                PublicKey::empty(),
+            ));
+            return rt1::NoiseInboundFuture::new(socket, session);
+        }
+        panic!("Should not panic");
```
tomaka: I think this should be removed, and you should have `Message::from_bytes(...).unwrap()` instead.

protocols/noise/src/lib.rs
```diff
@@ -48,195 +48,275 @@
...
+        if let Ok(prologue) = Message::from_bytes(b"Noise_IX_25519_ChaChaPoly_Blake2s") {
+            let session = NoiseSession::ix(noisesession_ix::NoiseSession::init_session(
+                true,
+                prologue,
+                self.0,
+                PublicKey::empty(),
+            ));
+            return rt1::NoiseOutboundFuture::new(socket, session);
+        }
+        panic!("Should not panic");
```
tomaka: Ditto

protocols/noise/src/lib.rs
```diff
@@ -48,195 +48,275 @@
...
+        if let Ok(prologue) = Message::from_bytes(b"Noise_XX_25519_ChaChaPoly_Blake2s") {
+            let session = NoiseSession::xx(noisesession_xx::NoiseSession::init_session(
+                false,
+                prologue,
+                self.0,
+                PublicKey::empty(),
+            ));
+            return rt15::NoiseInboundFuture::new(socket, session);
+        }
+        panic!("Should not panic");
```
tomaka: Ditto

protocols/noise/src/lib.rs
```diff
@@ -48,195 +48,275 @@
...
+        if let Ok(prologue) = Message::from_bytes(b"Noise_XX_25519_ChaChaPoly_Blake2s") {
+            let session = NoiseSession::xx(noisesession_xx::NoiseSession::init_session(
+                true,
+                prologue,
+                self.0,
+                PublicKey::empty(),
+            ));
+            return rt15::NoiseOutboundFuture::new(socket, session);
+        }
+        panic!("Should not panic");
```
tomaka: Ditto

protocols/noise/tests/smoke.rs
```diff
@@ -30,65 +34,70 @@ use tokio::{self, io};
...
-        let client_keypair = Keypair::<X25519>::new();
-        let client_transport = TcpConfig::new()
-            .and_then(move |output, endpoint| {
-                if endpoint.is_dialer() {
-                    Either::A(apply_outbound(output, NoiseConfig::ik_dialer(client_keypair, server_public)))
-                } else {
-                    Either::B(apply_inbound(output, NoiseConfig::xx(client_keypair)))
-                }
-            });
+        let server_keypair = Keypair::new();
```
tomaka: Nit: while I don't have a strong opinion on style, it's always a bit annoying when a large portion of the diff consists in insubstantial style changes.

georgio:
@tomaka Thank you for your review. This commit fixes the issues you have mentioned above. https://github.com/libp2p/rust-libp2p/pull/1106/commits/4c6b05aacfbc4fe1f077ad01795ecfa813bd2812.

I am currently working on a solution to avoid returning empty remote static keys as mentioned here https://github.com/libp2p/rust-libp2p/pull/1106#discussion_r281559348.

The bug that we discussed the other day is still around. As you previously mentioned, this is most likely something very minor in disguise. I would very much appreciate us going efforts to solve this. Maybe @romanb could give us a hand?

Please let us know if there's also anything else you would like us to take a look at before finalizing the work on this.

romanb:
This effort seems to have stalled, so I'm closing this for now. If someone is interested in reviving this, the main obstacle that needs to be overcome with the generated code is the libp2p-spec-compliant integration of handshake payloads carrying public keys and signatures, together with an adequate integration with `libp2p-core`.

burdges:
Yup.  I never judged the code production ready.  It required users understand how noise manipulates its buffers, so quite a nasty implementation detail exposed.  RustCrypto AEAD traits solved a similar problem, but they rejected my proposal to solve it in way usable for handshakes.  shrug..

PR #1107, opened 2019-05-07
Add implementations of prepare_uninitialized_buffer and read_buf where relevant

### LINKED ISSUE #1080: Implement prepare_uninitialized_buffer and read_buf in our AsyncRead implementations
For optimisation purposes. Quite a lot of CPU seems to be wasted in superfluous memsets.

tomaka:
Fixes #1080 

Most of the CPU usage is in `memcpy` (EDIT: and `memset`), so this should hopefully give a speed boost to the library.
Also adds `prepare_uninitialized_buffer` to the `StreamMuxer` trait.

### DIFF
--- a/muxers/yamux/src/lib.rs
+++ b/muxers/yamux/src/lib.rs
@@ -80,6 +80,10 @@ where
     fn destroy_outbound(&self, _: Self::OutboundSubstream) {
     }
 
+    unsafe fn prepare_uninitialized_buffer(&self, _: &mut [u8]) -> bool {
+        false
+    }
+
     fn read_substream(&self, sub: &mut Self::Substream, buf: &mut [u8]) -> Poll<usize, IoError> {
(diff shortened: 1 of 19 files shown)


muxers/yamux/src/lib.rs
```diff
@@ -80,6 +80,10 @@ where
     fn destroy_outbound(&self, _: Self::OutboundSubstream) {
     }
 
+    unsafe fn prepare_uninitialized_buffer(&self, _: &mut [u8]) -> bool {
+        false
```
tomaka: @twittner Please confirm that `yamux` doesn't read data from the buffer.

romanb: As far as I can tell, it doesn't.

twittner: Confirmed.

Issue #1087, opened 2019-04-24
Have a sharded Kademlia protocol


tomaka:
(note: this issue is about designing a protocol, not a bugfix).

There is no way, from Kademlia itself, to know if a node is IPFS, Polkadot, Edgeware, or something else.
This results is Substrate/Polkadot trying a lot of wrong nodes, then disconnecting from them.

Even if, say, Polkadot's network is totally separate from Edgeware's and IPFS's networks, all the Polkadot parachains would be part of a single network, and we will eventually need a way to discover the nodes/collators which are part of a specific parachain, for validators to be able to connect to them.

In #942 (which is more or less abandoned), I went with a "namespaces" system. Each node belongs to a namespace, this namespace is part of the identity of the node. I'm not sure if that's a good solution.

Ethereum 2.0 obviously has a similar issue with Sharding. Since I opened #942, they created discovery v5 (https://github.com/fjl/p2p-drafts), which I haven't read so far.

tomaka:
Relevant, as we could also use this "get_providers" system if it is improved: https://github.com/libp2p/specs/issues/163

burdges:
A priori, I'd think a namespace might be another hash function mapping from keys to storage nodes.  In principle, any node or key could select as many namespaces as desired.  And the default namespace itself might be optional.  If too few nodes participate in a namespace then look ups within that namespace become slow, due to the first lookup failing, or fail outright, if the stored or queried key does not enable the default namespace or reduces replicas there.  We've other corner cases when namespaces have few nodes too.  Anything like this requires more thought.

tomaka:
What I was thinking with namespaces is that a node is stored in the DHT at index `<namespace>|<node_id>` instead of just `<node_id>`.

For example, you'd store something like `dot25b109cf56...` in the DHT if your namespace is `dot` (modulo hashing the namespace as well). All nodes from all chains would participate in the discovery of all other chains.

burdges:
I'd think any performance improvement would require that storage nodes must opt into namespaces they like though, right?

burdges:
As an aside, there is sometimes a need to validate record placed into the DHT, which means some record type and validation logic too.

burdges:
We do not really have good answers here right now.  We currently think: 

Polkadot should _not_ run on the same network as IPFS, etc.  We see no reason to hurt performance just to be on the same network.  We also want the freedom to adopt different or nicer crypto then the support, including maybe add a post-quantum handshake, and maybe break other things.

[...]

PR #1108, opened 2019-05-08
Kademlia: Correct XOR metric.

romanb:
**Note**: Sits on top of https://github.com/libp2p/rust-libp2p/pull/1101.

This PR addresses what I believe to be two issues with the current implementation and use of Kademlia's XOR metric:

  1. The distance is currently equated with the bucket index, i.e. `distance(a,b) - 1` is the index of the bucket into which either peer is put by the other. The result is a metric that is not unidirectional, as defined in the Kademlia paper and, as far as I could see, as it is implemented in e.g. libp2p-go and libp2p-js (see e.g. [here](https://github.com/libp2p/go-libp2p-kbucket/blob/master/keyspace/xor.go#L21-L44)), which is to interpret the result of the XOR as-is as an integer representing the distance. The bucket index is of course still determined by the distance.

  2. The current `KBucketsPeerId` trait and its instances compute distances between types with differing bit lengths as well as between types that re-hash all inputs (i.e. `KadHash`) and "plain" `PeerId`s or `Multihash`es, or even just between `PeerId`s or `Multihash`es. Such distances are incorrect as per the libp2p spec requirement that all distances be computed over the XOR of the SHA256 hashes of the inputs, be it peer IDs, multihashes, etc. The computed distances can even fall outside of the image of the metric used for the `KBucketsTable` and thus fall in the range of a non-existent bucket. In the latter case, such distances are not currently used as a bucket index - they can only occur temporarily in the context of comparing distances for the purpose of sorting peers - but that still seems very undesirable.

These issues are addressed in this PR as follows:

  * Unidirectionality of the XOR metric is restored by keeping the "full" integer representation of the bitwise XOR. The result is an XOR metric as defined in the paper and as computed by other libp2p implementations. This should also open the door to avoiding the  "full table scan" when searching for the keys closest to a given key - the ideal order in which to visit the buckets can be computed with the help of the distance bit string, which is something I'd like to try afterwards.

  * As a simplification and to make it easy to "do the right thing", the XOR metric is only defined on an opaque `kbucket::Key` type, partially derived from the current `KadHash`. `KadHash` and `KBucketsPeerId` are removed.

These changes do not affect the distribution of keys into buckets, i.e. the same peer ID (as a former `KadHash`) still falls into the same bucket, but restoring unidirectionality means e.g. that lookups for the closest peers for a specific key from different nodes will now converge towards the same set of closest peers, as described in the paper.

Tests for the properties of the XOR metric have been added and some existing tests tweaked.


tomaka:
> Such distances are incorrect as per the libp2p spec requirement that all distances be computed over the XOR of the SHA256 hashes of the inputs, be it peer IDs, multihashes, etc.

I haven't reviewed yet, but to answer on that:

I had some "kademlia-with-namespaces" prototype in progress, where a prefix (for example containing the name of the Polkadot parachain a node belongs to) would be added at the front of the hash of the identity.

The idea would then to perform iterative queries for `name-of-the-chain|random-id`, and the results would contain (at least partially) nodes that belong to that same chain.
In order for this to work, we obviously shouldn't hash the prefix because calculating the distance.

See https://github.com/libp2p/rust-libp2p/issues/1087

However in practice I'm not sure whether we are going to continue in this direction.

tomaka:
Could you merge/rebase?

romanb:
Yes, I will do that by the end of the day.

burdges:
I'm not sure I understand the concerns around unidirectionality..

protocols/kad/src/kbucket.rs
```diff
@@ -308,10 +273,10 @@ where
         BucketsIter(self.tables.iter_mut(), self.unresponsive_timeout)
     }
 
-    /// Finds the nodes closest to `id`, ordered by distance.
+    /// Finds the keys closest to `key`, ordered by distance.
     ///
     /// Pending nodes are ignored.
-    pub fn find_closest(&mut self, id: &impl KBucketsPeerId<TPeerId>) -> VecIntoIter<TPeerId> {
+    pub fn find_closest<T>(&mut self, key: &Key<T>) -> VecIntoIter<Key<TPeerId>> {
```
tomaka: I was curious what the code of this method would look like now. If I understand correctly, it will now be possible to write an optimized version of this (but it's not part of this PR)?

tomaka:
I'd like to leave the door open for not hashing the key in the k-bucket, in case we want to implement the namespacing thing (#1087). If you think that this can easily be added back in the future, then +1 for this.

protocols/kad/src/kbucket.rs
tomaka: Oh nevermind, I was mixing with #1117

romanb: Right, that is in https://github.com/libp2p/rust-libp2p/pull/1117 which I'm still polishing up, but the part that deals with this optimisation in particular is stable and [can be found here](https://github.com/libp2p/rust-libp2p/blob/80cc7eaa78a62e291842e149f52c47659260eb42/protocols/kad/src/kbucket.rs#L259-L447). The actual `Iterator` impl for the (generic) `ClosestIter` is at the end of that highlighted code fragment - it just draws from two underlying iterators. The actual ordering of the bucket indices is provided by the (internal) `ClosestBucketsIter`.

FatemeShirazi:
> I'm not sure I understand the concerns around unidirectionality.. @FatemeShirazi

@romanb is right about the unidirectionality problem, interpreting the result of the XOR would eliminate unidirectionality and it is addressed correctly as far as I understand.

romanb:
> I'd like to leave the door open for not hashing the key in the k-bucket, in case we want to implement the namespacing thing (#1087). If you think that this can easily be added back in the future, then +1 for this.

@tomaka I don't think I actually understand (yet) what you have in mind or what is being discussed in #1087, but technically you could of course permit any 32 byte array as a key. However, when not hashing, it is worth pointing out that the proof that operations take logarithmic time in Kademlia rests on the assumption of uniformly random distributed keys.

PR #1113, opened 2019-05-10
Add parity_multiaddr::from_websockets_url

tomaka:
Allows parsing `ws://127.0.0.1:5000` into `/ip4/127.0.0.1/tcp/5000/ws`.


tomaka:
Relevant: #1093

tomaka:
> Why not a more general URL to multi-address conversion which also supports HTTP and HTTPS?

I've thought about this and wasn't sure if it's a good idea.

We may want to add support for new multiaddr protocols later, and when that happens we would have to decide whether we change the behaviour of the function, or we don't change it and be in a weird situation where the function accepts some protocols and not others.

However in practice after thinking a bit more I don't see a reason why changing the behaviour of the function is a bad thing.

misc/multiaddr/Cargo.toml
```diff
@@ -17,6 +17,7 @@ data-encoding = "2.1"
 multihash = { package = "parity-multihash", version = "0.1.0", path = "../multihash" }
 serde = "1.0.70"
 unsigned-varint = "0.2"
+urlparse = "0.7.3"
```
twittner: Why did you go for this crate instead of [url](https://crates.io/crates/url) which is also used in Servo?

misc/multiaddr/src/from_url.rs
```diff
@@ -0,0 +1,145 @@
...
+/// Attempts to parse an URL into a multiaddress.
+///
+/// # Example
+///
+/// ```
+/// let addr = parity_multiaddr::from_url("ws://127.0.0.1:8080/").unwrap();
+/// assert_eq!(addr, "/ip4/127.0.0.1/tcp/8080/ws".parse().unwrap());
+/// ```
+///
+pub fn from_url(url: &str) -> std::result::Result<Multiaddr, FromUrlErr> {
```
twittner: Assuming we would use the `url` crate, how about a `TryFrom` impl for its `Url` type?

tomaka:
I'm a bit reluctant to expose the `url` crate in our API.

twittner:
> I'm a bit reluctant to expose the `url` crate in our API.

Why? We already expose `bytes` and `serde`. The `url` crate is well established and has almost twice as many reverse dependencies as `bytes`.

tomaka:
I'm reluctant to expose *any* crate in our API. We're exposing `bytes` because we don't have the choice, and for `serde` I was reluctant as well.

twittner:
> I'm reluctant to expose *any* crate in our API. We're exposing `bytes` because we don't have the choice [...]

If you are concerned, the `TryFrom<Bytes>` impl only serves as an optimisation so that clients can create a `Multiaddr` directly from a `Bytes` value without the need for a `Vec<u8>`. But AFAICT we do not need it otherwise.

misc/multiaddr/src/from_url.rs
```diff
@@ -0,0 +1,154 @@
...
+    let url = url::Url::parse(url).map_err(|_| FromUrlErr::BadUrl)?;
+
+    let (protocol, default_port) = match url.scheme() {
+        "ws" => (Protocol::Ws, 80),
+        "wss" => (Protocol::Wss, 443),
+        "http" => (Protocol::Http, 80),
+        "https" => (Protocol::Https, 443),
+        _ => return Err(FromUrlErr::WrongScheme)
```
twittner: Maybe we should expand the doc comment a bit and mention which schemes are currently supported and also which URL parts are ignored. In fact I think we should error on URLs which contain information we can not preserve in a `Multiaddr`, such as user name, password, query string or fragment. Also I think `UnsupportedScheme` (or just `Unsupported` if we would error on parts we can not handle) would be a better name than `WrongScheme`. Additionally, a function `from_url_lossy` could be offered which ignores the bits we do not care about (essentially what `from_url` does right now).

tomaka:
I've also added support for `unix:/foo/bar` in the last commit.

PR #1114, opened 2019-05-11
Fix small todo in handled tasks

tomaka:


core/src/nodes/handled_node_tasks.rs
```diff
@@ -335,9 +335,12 @@ impl<TInEvent, TOutEvent, TIntoHandler, TReachErr, THandlerErr, TUserData, TConn
...
             let mut task = self.local_spawns.swap_remove(n);
             match task.poll() {
-                Ok(Async::Ready(())) => (),
+                Ok(Async::Ready(())) => {},
                 Ok(Async::NotReady) => self.local_spawns.push(task),
-                Err(_err) => ()     // TODO: log this?
+                // It would normally be desirable to either report or log when a background task
+                // errors. However the default tokio executor doesn't do anything in case of error,
+                // and therefore we mimic this behaviour by also not doing anything.
+                Err(_err) => {}
```
romanb: The error type is `()` anyway, isn't it, so the match could be more specifically `Err(())` and there isn't really anything useful to log or report, is there?

PR #1116, opened 2019-05-13
Add an OptionalTransport type

tomaka:
Basically a glorified `Option<T>`.

The reason for this, is that in a WASM context you want to pass an external transport, but in a non-WASM context you want not to.

Thanks to an `OptionalTransport<wasm_ext::ExtTransport>`, you can share the same code between WASM and non-WASM.

### DIFF
--- a/core/src/transport/optional.rs
+++ b/core/src/transport/optional.rs
@@ -1,0 +1,72 @@
+// Copyright 2017 Parity Technologies (UK) Ltd.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a
+// copy of this software and associated documentation files (the "Software"),
+// to deal in the Software without restriction, including without limitation
+// the rights to use, copy, modify, merge, publish, distribute, sublicense,
+// and/or sell copies of the Software, and to permit persons to whom the
+// Software is furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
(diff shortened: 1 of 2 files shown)


core/src/transport/optional.rs
```diff
@@ -0,0 +1,77 @@
...
+/// `None`), then any attempt to dial or listen will return `MultiaddrNotSupported`. If it is
+/// enabled (read: contains `Some`), then dialing and listening will be handled by the inner
+/// transport.
+#[derive(Debug, Copy, Clone)]
+pub struct OptionalTransport<T>(Option<T>);
+
+impl<T> OptionalTransport<T> {
+    /// Builds an `OptionalTransport` with the given transport in an enabled
+    /// state.
+    pub fn new(inner: T) -> OptionalTransport<T> {
```
twittner:
```suggestion
    pub fn some(inner: T) -> OptionalTransport<T> {
```

core/src/transport/optional.rs
```diff
@@ -0,0 +1,77 @@
...
+        OptionalTransport(Some(inner))
+    }
+
+    /// Builds a disabled `OptionalTransport`.
+    pub fn empty() -> OptionalTransport<T> {
```
twittner:
```suggestion
    pub fn none() -> OptionalTransport<T> {
```

Issue #1115, opened 2019-05-13
Allow custom protocol Id for all protocols


AgeManning:
It has been suggested that applications use their own protocol id's as a fundamental segregation of networks between different applications.

Recently, a go application using libp2p failed as it became interconnected with ipfs nodes because somehow the nodes got on the discovery protocol and the two applications where using the same protocol id. 

I suggest a PR (which i'm happy to write) that allows each protocol's id to be customized. This, however would not be backwards compatible (as far as I can tell). I'd likely add the option to the behaviour struct upon initialization.

AgeManning:
In particular, I'm referring to this:
https://github.com/ethereum/eth2.0-specs/pull/935#discussion_r280413163

tomaka:
I'm not against making that change, but I thought that libp2p was all about interoperability. If suddenly every network has their own protocol in order to prevent them from talking to each other, I'm confused about the point of having protocols that are part of the libp2p specs.

AgeManning:
Yeah i see your point. I guess in my mind the interoperability part is accross implementations but not accross applications/networks. 

I think giving the option to users to segregate their network from other applications is strictly better than not doing so, as they can always use the defaults to interop with other nodes. 

Curious about @raulk thoughts, as he initially suggested this.

raulk:
I believe it’s a legitimate use case to use libp2p as the networking library of an independent network. Some applications may require stronger guarantees that are not enforceable in hybrid networks. We don’t support overlay network (logical segregation) on top of a single physical network yet, so segregating in the manner suggested here should create a similar effect. On a different note, I think it’s fine to standardise the behaviour of libp2p protocols without obliging users to run them on a Singleton swarm/network.

tomaka:
In practice, only Kademlia requires customization, right?
Ping and identify across networks doesn't hurt in any way as far as I can tell. Floodsub/gossipsub have topics whose names are network-specific.
Only Kademlia could be harmful in the sense that it would discover nodes that are irrelevant.

raulk:
Yeah, Kademlia would require customisation on the grounds that it stores state, and you might not want to leak state across networks. But IMO the ultimate segregation comes by customising the Multistream protocol ID. The go implementation allows this and several downstreams like OB1 use it.

AgeManning:
If go allows this segregation, perhaps we add this into gossip and floodsub as well. 

[...]

PR #1118, opened 2019-05-14
Allow changing the Kademlia protocol name

tomaka:
cc #1115 
cc @romanb

### DIFF
--- a/protocols/kad/src/behaviour.rs
+++ b/protocols/kad/src/behaviour.rs
@@ -30,7 +30,7 @@ use libp2p_core::swarm::{ConnectedPoint, NetworkBehaviour, NetworkBehaviourActio
 use libp2p_core::{protocols_handler::ProtocolsHandler, Multiaddr, PeerId};
 use multihash::Multihash;
 use smallvec::SmallVec;
-use std::{error, marker::PhantomData, num::NonZeroUsize, time::Duration};
+use std::{borrow::Cow, error, marker::PhantomData, num::NonZeroUsize, time::Duration};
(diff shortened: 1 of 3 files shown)


tomaka:
I pushed another commit that actually exposes this functionality in the `Kademlia` network behaviour.

protocols/kad/src/behaviour.rs
```diff
@@ -189,6 +192,17 @@ impl<TSubstream> Kademlia<TSubstream> {
         Self::new_inner(local_peer_id)
     }
 
+    /// Same as `new`, but allows configuring a different protocol name.
+    ///
+    /// Nodes are only able to talk to nodes that support protocols with the same name. By using
+    /// a different name, you can force a specific set of nodes to talk to each other and not
+    /// others.
```
romanb: How about:
```
/// The same as `new`, but using a custom protocol name.
///
/// Kademlia nodes only communicate with other nodes using the same protocol name. Using a custom name therefore
/// allows to segregate the DHT from others, if that is desired.
```

protocols/kad/src/behaviour.rs
```diff
@@ -41,6 +41,9 @@ pub struct Kademlia<TSubstream> {
     /// Storage for the nodes. Contains the known multiaddresses for this node.
     kbuckets: KBucketsTable<KadHash, Addresses>,
 
+    /// If `Some`, we overwrite the Kademlia protocol name with this one.
+    protocol_name_overwrite: Option<Cow<'static, [u8]>>,
```
```suggestion
    protocol_name_override: Option<Cow<'static, [u8]>>,
```

PR #1121, opened 2019-05-14
Add the possibility to assume that multistream-select will succeed

### LINKED ISSUE #659: Add the possibility to assume that multistream-select will succeed
If the connection upgrade only accepts one protocol, we can optimize the `multistream-select` process by assuming that it works, and closing the connection/substream altogether if later the remote returns `"na"`.

I would suggest modifying `dialer_select_proto` so that if the `size_hint()` of the protocols list returns a maximum of 1, immediately succeed after sending the data without waiting for the remote to confirm, and wraps around the returned stream.

The behaviour should also be precisely documented.

tomaka:
Fix #659

Does unfortunately not work and is quite hacky.

One unfortunate consequence (which is not a hack), is that we require the AsyncRead trait on write-only substreams, because we need to read whether the remote sent a negotiation success message. This is a small breaking change.

### DIFF
--- a/misc/multistream-select/src/dialer_select.rs
+++ b/misc/multistream-select/src/dialer_select.rs
@@ -21,7 +21,7 @@
 //! Contains the `dialer_select_proto` code, which allows selecting a protocol thanks to
 //! `multistream-select` for the dialer.
 
-use futures::{future::Either, prelude::*, stream::StreamFuture};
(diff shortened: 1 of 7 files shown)


romanb:
Closing as a result of https://github.com/libp2p/rust-libp2p/pull/1212.

Issue #604, opened 2018-11-02
Figure out the caching mechanism in circleci


tomaka:
The caching commands we're using for CircleCI right now are wrong. Caches are immutable, and therefore writing over an existing cache doesn't work.

Similarly we should cache the Docker image we build in the `test` command.

kirushik:
Can we use `Cargo.toml` as a beacon for changes in dependencies?
Then it would be something like

```yaml
- restore_cache:
    keys:
        - v1-{{ checksum "Cargo.toml" }}
```

tomaka:
@TriplEight Could I ask you to do the rest of the CircleCI file in addition to #1122?

TriplEight:
Ok, will do.

romanb:
CircleCI seems to no longer be used.

PR #1122, opened 2019-05-14
new dockerfile

TriplEight:
- uses a new builder image 
- with `sccache`.


TriplEight:
Example of of running with cache: https://circleci.com/gh/libp2p/rust-libp2p/13225#build-parameters/containers/0

Like in default GitLab cache behaviour, CircleCi downloads, archives, uploads and unarchives the cache, and it takes considerable time. Probably, storing `sccache`s cache will not make sense for this project. It really makes the payload run faster, but handling cache can potentially take more time.

Also, I've noticed that caching doesn't work in other jobs here being set up not correctly.

tomaka:
> Also, I've noticed that caching doesn't work in other jobs here being set up not correctly.

https://github.com/libp2p/rust-libp2p/issues/604

TriplEight:
do you need any changes from this PR?

PR #1124, opened 2019-05-15
Add some diagnostics for the same address being reported despite not in list

tomaka:


transports/tcp/src/lib.rs
```diff
@@ -422,7 +422,7 @@ impl Stream for TcpListenStream {
 
             match apply_config(&self.config, &sock) {
                 Ok(()) => {
-                    debug!("Incoming connection from {}", remote_addr);
+                    debug!("Incoming connection from {} on {}", remote_addr, listen_addr);
```
twittner:
```suggestion
                    trace!("Incoming connection from {} on {}", remote_addr, listen_addr);
```

Requires importing `log::trace`.

core/src/nodes/listeners.rs
```diff
@@ -188,7 +188,9 @@ where
                     if remaining == 0 { break }
                 }
                 Ok(Async::Ready(Some(ListenerEvent::Upgrade { upgrade, listen_addr, remote_addr }))) => {
-                    debug_assert!(listener.addresses.contains(&listen_addr));
+                    debug_assert!(listener.addresses.contains(&listen_addr),
+                        "Transport reported connection from {} not in the list: {:?}",
```
twittner:
```suggestion
                        "Transport reported listen address {} not in the list: {:?}",
```

PR #1127, opened 2019-05-15
Fix ED25519 signature validation

appaquet:
The ED25519 verify method wasn't checking if the actual verification was successful, but was only checking if the signature decoding was successful. 

The RSA and Secp256k1 keys logic looks valid.


tomaka:
Hmm, probably worth notifying the cargo audit people.

cc @kirushik

core/src/identity/ed25519.rs
```diff
@@ -100,7 +100,7 @@ pub struct PublicKey(ed25519::PublicKey);
 impl PublicKey {
     /// Verify the Ed25519 signature on a message using the public key.
     pub fn verify(&self, msg: &[u8], sig: &[u8]) -> bool {
```
tarcieri: You might want to consider changing this method to return a `Result<(), ()>` so as to strategically avoid this bugclass in the future

PR #1132, opened 2019-05-17
Map all 127.0.0.0/8 addresses to 127.0.0.1.

twittner:
Since every local socket address in the 127.0.0.0/8 space is looped back to 127.0.0.1/32 we should only have to report the later as the listen address. For other addresses we still attempt to discover host addresses when we encounter an unknown local address. We now also check that after
the host addresses have been reset that the address is now found, otherwise we produce an error.

### DIFF
--- a/transports/tcp/src/lib.rs
+++ b/transports/tcp/src/lib.rs
@@ -55,7 +55,7 @@ use std::{
     fmt,
     io::{self, Read, Write},
     iter::{self, FromIterator},
-    net::{IpAddr, SocketAddr},
+    net::{IpAddr, Ipv4Addr, SocketAddr},
     time::Duration,
     vec::IntoIter
 };
@@ -334,6 +334,8 @@ enum Addresses {
     Many(HashMap<IpAddr, Multiaddr>)
 }
 
+type Buffer = VecDeque<ListenerEvent<FutureResult<TcpTransStream, io::Error>>>;
+
 /// Stream that listens on an TCP/IP address.
 pub struct TcpListenStream {
     /// Stream of incoming sockets.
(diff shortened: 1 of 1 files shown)


elferdo:
Is this really so? While 127.0.0.1/32 is the most commonly used loopback address, nothing prevents a user from adding more within the 127.x.x.x range, as far as I know.

twittner:
Interesting, thanks @elferdo! I wonder what the correct approach is. Do you think the following would work?

1. Check for exact IP address equality and report the first match as listen address.
2. If there was no exact match in 1, check if any known listen address contains the given local IP address. This would utilise the netmask. If multiple listen addresses match report the first one.
3. If neither 1 nor 2 found an address, check for any new host addresses. If the checks in 1 and 2 then still do not find a matching listen address, report an error, otherwise report the match as listen address.

tomaka:
To give some context this PR: we sometimes receive an incoming connection whose "local address" (ie. the address we're supposedly listening on) is `127.94.0.1`. And since we aren't aware that we were listening on this address, a panic is triggered.

elferdo:
How is the list of known local addresses prefilled? The usual way to start a tcp/ip socket is to bind it to 0.0.0.0 and listen on that, which basically means listen on all interfaces. Do we query for the list of interfaces somehow?

> we sometimes receive an incoming connection whose "local address" (ie. the address we're supposedly listening on) is 127.94.0.1

It's interesting if it only happens through the loopback device. Is it a network of docker nodes? Or somehow connected over VPN?. How about, for example a laptop connected to a wifi network and over cable at the same time? Would that fail too?

twittner:
> Do we query for the list of interfaces somehow?

Yes, using https://crates.io/crates/get_if_addrs.

twittner:
> It's interesting if it only happens through the loopback device.

That is because on many systems by default you can send to any address in 127.0.0.0/8 and it will be looped back to 127.0.0.1/32.

> How about, for example a laptop connected to a wifi network and over cable at the same time? Would that fail too?

The assumption is that the local address of the accepted socket matches one of the interface's addresses. It is not entirely clear to me if for interfaces other than `lo` we might get addresses that do not match an interface's address but to be on the safe side I suggested [above](https://github.com/libp2p/rust-libp2p/pull/1132#issuecomment-493452862) to also consider the netmask if no exact match has been found.

elferdo:
> > It's interesting if it only happens through the loopback device.
> 
> That is because on many systems by default you can send to any address in 127.0.0.0/8 and it will be looped back to 127.0.0.1/32.

That's pretty weird to me. But if that's a fact, then it's a fact.

> The assumption is that the local address of the accepted socket matches one of the interface's addresses. It is not entirely clear to me if for interfaces other than `lo` we might get addresses that do not match an interface's address but to be on the safe side I suggested [above](https://github.com/libp2p/rust-libp2p/pull/1132#issuecomment-493452862) to also consider the netmask if no exact match has been found.

That's a reasonable assumption. I can only think of DHCP giving out a lease with a new IP address, or a mobile device roaming around, I don't know how mobile phones handle IPs. But otherwise I would expect the connected socket to match one of the known addresses.

twittner:
I updated the implementation to use the strategy outlined [above](https://github.com/libp2p/rust-libp2p/pull/1132#issuecomment-493452862).

transports/tcp/src/lib.rs
```diff
@@ -343,11 +390,67 @@ pub struct TcpListenStream {
...
+// If not found, check for host address changes.
+// This is a function rather than a method due to borrowing issues.
+fn map_addr(addr: &SocketAddr, addrs: &mut Addresses, pending: &mut Buffer, port: u16)
+    -> Result<Multiaddr, io::Error>
+{
+    match addrs {
+        Addresses::One(ref ma) => Ok(ma.clone()),
+        Addresses::Many(ref mut addrs) => {
+            // Check for exact match:
+            if let Some((_, _, ma)) = addrs.iter().find(|(i, ..)| i == &addr.ip()) {
```
tomaka: I didn't even know you could use `..` in tuples.

transports/tcp/src/lib.rs
```diff
@@ -242,15 +244,58 @@ fn sockaddr_to_multiaddr(ip: IpAddr, port: u16) -> Multiaddr {
...
+                IpNet::V6(Ipv6Net::new(ip6.ip, plen).expect("prefix len has been checked"))
+            }
+        };
+        addrs.push((ip, ipn, ma))
     }
     Ok(addrs)
 }
 
+// Derive the prefix len of the given netmask, i.e. the number of leading 1s in its octets.
+fn prefix_len(netmask: &IpAddr) -> u8 {
```
tomaka: I would be much more confident in this function if it had a test.

twittner:
Instead of trying to figure out the listen address maybe we should just report the local address and change `ListenerEvent::Upgrade` to:

```rust
Upgrade {
        /// The upgrade.
        upgrade: T,
        /// The local address which produced this upgrade.
        local_addr: Multiaddr,
        /// The remote address which produced this upgrade.
        remote_addr: Multiaddr
    }
```

PR #1139, opened 2019-05-22
libp2p-kad: Remove connected peer earlier in inject_disconnected.

romanb:
The removal of a disconnected peer from `self.connected_peers` in the Kademlia behaviour has been moved to the end of `inject_disconnected` in an earlier PR. However, there is a debug assertion in `connection_updated` that asserts that the peer is no longer present, which is currently called before the removal of the peer. I don't actually think the order is important in this context, i.e. the assertion does not seem particularly important to me, but this PR restores the previous order of operations for now.


romanb:
Actually, I don't think the debug assertion should trigger since `inject_disconnected` sets the status to `NodeStatus::Disconnected`. It may instead be triggered by `inject_connected`. I will take a closer look at the diffs.

romanb:
Superseded by https://github.com/libp2p/rust-libp2p/pull/1140.

Issue #146, opened 2018-03-15
Implement Kademlia record store


tomaka:
It is currently unimplemented.

tomaka:
Record storage is about storing records that other peers want to store on our machine, and then loading them back when requested. More concretely, it's about implementing the `ADD_VALUE` and `GET_VALUE` messages. Right now when we receive such a message we immediately close the Kademlia substream.

The challenge in doing is mostly implementing the signatures system that makes it possible to prove that the record wasn't altered.

tomaka:
Considering that the user may want to store records on the disk, or cache records in a certain way, it is in my opinion preferable to not store the records directly in the `Kademlia` struct, but to let users entirely handle that.
In other words, when a record is requested from the network, an event would be emitted (for example, `KademliaOut::RecordRequest`) and the user would have to answer this request by providing the data.

romanb:
> Considering that the user may want to store records on the disk, or cache records in a certain way, it is in my opinion preferable to not store the records directly in the `Kademlia` struct, but to let users entirely handle that.
> In other words, when a record is requested from the network, an event would be emitted (for example, `KademliaOut::FetchRecord`) and the user would have to answer this request by providing the data.
> A similar event (eg. `KademliaOut::AddRecord`) would of course be emitted when the network wants us to store a record.

I'm not so sure. Implementing record storage and retrieval in such a way that persistence of records has a negligible probability of failure in the context of node failures, nodes joining and leaving seems non-trivial, involving a combination of regularly scheduled operations like bucket refreshes, record re-replication and record re-publication in somewhat carefully chosen intervals. And all of that seems central to a Kademlia DHT implementation. It does of course seem desirable to abstract over the concrete choice of storage (hashmap, files, dbms, ...) through a small interface, combined with configurable intervals for all the necessary regular record gossip. I do think that `libp2p-kad` should ship with an in-memory storage option that can configured or swapped out for something else according to a defined interface (probably containing at least functions along the lines of `get`, `put`, `select` and `validate`, whereby the latter two come from the libp2p spec section regarding entry validation, which I'm not sure we want or need to have?).

romanb:
Closing this as a result of https://github.com/libp2p/rust-libp2p/pull/1189. I think any specific missing pieces or bugs should be filed in new, separate issues.

PR #1144, opened 2019-05-23
Kademlia Records

montekki:


tomaka:
A few remarks:

- As explained in #146, I think that it should be the responsibility of the user to store records.
- The Kademlia iterative query should stop whenever it finds a record. However this is vulnerable to misbehaviour (a node can return a bad record and stop the iterative query by itself). We need some sort of way to ask the user whether a record is satisfactory.

However I think it is fine as a first draft.

montekki:
@tomaka 
* ok
* It doesn't look so hard to implement the quorum logic from the specs actually

About works: There are some tests and there is also a more complicated test (not pushed yet) based off the chain topology in `query_iter` test. It works but I am not really satisfied with how, i.e. in the topology of a chain the value is now only distributed to the "next" neighbor that we are connected to, that I am trying to find out what to do with now.

I would argue for at least
* implementing the logic of `PUT_VALUE` in a way that data is distributed to some decent amount of nodes
* implementing the logic of `GET_VALUE` with quorum-ing
* probably implementing validation

before merging this

protocols/kad/src/behaviour.rs
```diff
@@ -283,6 +315,18 @@ impl<TSubstream> Kademlia<TSubstream> {
         self.start_query(QueryInfoInner::GetProviders { target, pending_results: Vec::new() });
     }
 
+    /// Starts an iterative `GET_VALUE` request.
+    pub fn get_data(&mut self, key: Multihash) {
```
```suggestion
    pub fn get_value(&mut self, key: Multihash) {
```

protocols/kad/src/behaviour.rs
```diff
@@ -283,6 +315,18 @@ impl<TSubstream> Kademlia<TSubstream> {
...
+        self.start_query(QueryInfoInner::GetValue { key, results: vec![] });
+    }
+
+    /// Starts an iterative `PUT_VALUE` request
+    pub fn put_data(&mut self, key: Multihash, data: &[u8]) {
```
```suggestion
    pub fn put_value(&mut self, key: &Multihash, value: Vec<u8>) {
```

romanb: How about taking `data` as a `Vec<u8>`, i.e. let the function take ownership? I think that may avoid a clone if the caller of the function no longer needs the value afterwards, and taking ownership of the values seems to be common for "insert" operations on data structures. Similarly, taking the key by reference gives a bit more leeway in the implementation itself, in case it may not always need the key by-value.

protocols/kad/src/behaviour.rs
```diff
@@ -283,6 +315,18 @@ impl<TSubstream> Kademlia<TSubstream> {
...
+        // TODO: Probably we shouldn't store the value ourselves
+        self.records.insert(key.clone(), data.to_vec());
```
romanb: Is it intentional that the record is inserted here and again later when the `PutValue` query completes [here](https://github.com/libp2p/rust-libp2p/pull/1144/files#diff-88d683446d7fa7abbd7a347fb7d7c540R850)?

protocols/kad/src/behaviour.rs
```diff
@@ -623,6 +667,42 @@ where
...
+            } => {
+                if let Some(query) = self.active_queries.get_mut(&user_data) {
+                    if let QueryInfoInner::GetValue {
+                        key: _,
+                        results
+                    } = &mut query.target_mut().inner {
+                        if let Some(result) = result {
+                            results.push(result);
+                        }
+                    }}
```
romanb: I think it would be good if the initial implementation already stops the query as soon as a value is returned.

protocols/kad/src/behaviour.rs
```diff
@@ -753,6 +833,33 @@ where
...
+
+                        let event = KademliaOut::GetValueRes {
+                            result: result.clone(),
+                            closer_peers: closer_peers.collect(),
+                        };
+
+                        break Async::Ready(NetworkBehaviourAction::GenerateEvent(event));
+                    },
+                    QueryInfoInner::PutValue { key, value } => {
+                        self.records.insert(key.clone(), value.clone());
```
romanb: See the other comment about this seemingly duplicate insertion.

protocols/kad/src/behaviour.rs
```diff
@@ -802,6 +909,14 @@ pub enum KademliaOut {
         /// List of peers ordered from closest to furthest away.
         closer_peers: Vec<PeerId>,
     },
+
+    /// Result of a `GET_VALUE` query
+    GetValueRes {
```
```suggestion
    GetValueResult {
```
Just for naming consistency.

protocols/kad/src/behaviour.rs
```diff
@@ -802,6 +909,14 @@ pub enum KademliaOut {
...
+        /// The result that we have probably received
+        result: Option<(Multihash, Vec<u8>)>,
```
romanb: Is the key really needed in the result, i.e. a tuple needed? Is the key even included in the response? The libp2p-specs are a bit vague, but it only mentions that the response has `record` (in the protobuf) set if the value is found.

protocols/kad/src/behaviour/test.rs
```diff
@@ -223,3 +224,79 @@ fn unresponsive_not_returned_indirect() {
...
+
+    let target_key = multihash::encode(Hash::SHA2256, &vec![1,2,3]).unwrap();
+    swarms[0].get_data(target_key.clone());
+    swarms[1].put_data(target_key.clone(), &vec![4,5,6]);
+
+    Runtime::new().unwrap().block_on(
+        future::poll_fn(move || -> Result<_, io::Error> {
+            for swarm in &mut swarms {
+                loop {
+                    match swarm.poll().unwrap() {
```
romanb: Is there not a potential race here where the `get_value` request may be processed by the other swarm before the `put_value`? In other words, isn't it necessary to first wait for the `put` to complete before issuing the `get`?

protocols/kad/src/handler.rs
```diff
@@ -182,6 +182,35 @@ pub enum KademliaHandlerEvent<TUserData> {
...
+        key: Multihash,
+
+        /// Identifier of the request. Needs to be passed back when answering.
+        request_id: KademliaRequestId,
+    },
+
+    /// Response to a `KademliaHandlerIn::GetValue`.
+    GetValueRes {
+        /// The result is present if the key has been found
+        result: Option<(Multihash, Vec<u8>)>,
```
romanb: See the other comment about whether this needs to be a tuple.

protocols/kad/src/handler.rs
```diff
@@ -771,5 +877,19 @@ fn process_kad_response<TUserData>(
...
+        KadResponseMsg::GetValue {
+            result,
+            closer_peers,
+        } => KademliaHandlerEvent::GetValueRes {
+            result,
+            closer_peers,
+            user_data,
+        },
+        KadResponseMsg::PutValue { key, value } => {
+            KademliaHandlerEvent::PutValue {
```
romanb: Shouldn't this be a `KademliaHandlerEvent::PutValueRes` and I would think without any contents? Again the libp2p-spec isn't really clear to me, but it doesn't say anything about whether there are any fields present in a `PUT_VALUE` response (or indeed that there is any response at all).

protocols/kad/src/protocol.rs
```diff
@@ -294,6 +308,22 @@ pub enum KadResponseMsg {
...
+        closer_peers: Vec<KadPeer>,
+    },
+
+    /// Response to a `PutValue`.
+    PutValue {
+        /// The key of the record.
+        key: Multihash,
+        /// The value of the record.
+        value: Vec<u8>,
+    },
```
romanb: See the other comment about whether there is actually supposed to be a response to a `PUT_VALUE` and if so whether it carries any data.

protocols/kad/src/protocol.rs
```diff
@@ -361,6 +413,36 @@ fn resp_msg_to_proto(kad_msg: KadResponseMsg) -> proto::Message {
...
+            result.map(|(key, value)| {
+                let mut record = proto::Record::new();
+                record.set_key(key.into_bytes());
+                record.set_value(value);
+                msg.set_record(record);
+            });
+
+            msg
+        }
+        KadResponseMsg::PutValue {key, value} =>  {
```
romanb: See the other comments about `PUT_VALUE` responses. Is this code currently even executed? I don't see a code path that actually _sends_ such a response?

protocols/kad/src/protocol.rs
```diff
@@ -414,8 +502,24 @@ fn proto_to_resp_msg(mut message: proto::Message) -> Result<KadResponseMsg, io::
...
+                    let key = Multihash::from_bytes(record.take_key()).map_err(invalid_data)?;
+                    Some((key, record.take_value().into()))
+                }
+                false => None,
+            };
+
+            let closer_peers = message
+                .mut_closerPeers()
+                .iter_mut()
+                .filter_map(|peer| KadPeer::try_from(peer).ok())
```
romanb: It may be preferable to fail here if serialisation of any `KadPeer` fails, instead of silently dropping entries that cannot be parsed, which may lead to confusion.

protocols/kad/src/protocol.rs
```diff
@@ -446,8 +550,15 @@ fn proto_to_resp_msg(mut message: proto::Message) -> Result<KadResponseMsg, io::
             })
         }
 
-        proto::Message_MessageType::PUT_VALUE =>
-            Err(invalid_data("received an unexpected PUT_VALUE message")),
+        proto::Message_MessageType::PUT_VALUE => {
```
romanb: See the other comments about `PUT_VALUE` responses.

protocols/kad/src/behaviour.rs
montekki: Here we put the data on the side of the putter

protocols/kad/src/behaviour.rs
montekki: Here we put the data on the side of the peer that received `PutValue`, as stated in the `TODO` the first one is probably not necessary

protocols/kad/src/protocol.rs
montekki: No, this is a leftover garbage that needs to go since `PUT_VALUE` doesn't have a response (judging from the docs at least), does it?

protocols/kad/src/handler.rs
montekki: We should check other implementations, but, yes, I also doubt that there should be any response.

protocols/kad/src/behaviour/test.rs
montekki: That is true, but since the `PUT_VALUE` will probably have no response to it, I am puzzled with possible ways to actually wait for it.

romanb:
> I would argue for at least
> 
>     * implementing the logic of `PUT_VALUE` in a way that data is distributed to some decent amount of nodes

But that is already happening, as far as I can tell. What you implemented is the standard iterative `PUT_VALUE` that stores the record on the (up to) `k` closest nodes to the key. Of course this is generally already assuming a properly bootstrapped / initialised routing table. For persistence, the record would  need to be replicated and regularly republished, of course, but that is maybe beyond the scope of this PR. I do think, however, that there should be at least either already an expiration on the records, or the size of the record store (current hashmap) should be capped, which may anyway be necessary eventually.

> 
>     * implementing the logic of `GET_VALUE` with quorum-ing

I think it would be fine to just stop on the first result for now, which is the "standard" Kademlia. The quorum scheme with the `Q` parameter seems to be an extension of libp2p that could be added later.

>     * probably implementing validation

What would you want to do to validate the records? Validation seems to be something that is just offered as an API to user code, i.e. such that user code can do whatever validation it likes. Do you mean you would already want to provide an interface for validation?

protocols/kad/src/behaviour.rs
montekki: My thoughts on this: we most likely need to know to what key request we received this response, since, `Multihash` might(?) be different hashing algorithms or even something like `/pk/{bytes}` (?). In any case, my intuition is that we need the key here, as it happens with `GetProvidersResult`.

montekki:
> What would you want to do to validate the records? Validation seems to be something that is just offered as an API to user code, i.e. such that user code can do whatever validation it likes. Do you mean you would already want to provide an interface for validation?

The `Record` has some fields for signatures being passed around doesn't it? We probably should check them and when we are asked for some value, we should send out the record with the signatures

romanb:
> > What would you want to do to validate the records? Validation seems to be something that is just offered as an API to user code, i.e. such that user code can do whatever validation it likes. Do you mean you would already want to provide an interface for validation?
> 
> The `Record` has some fields for signatures being passed around doesn't it?

Not any longer, as far as I know. See the [latest protobuf]( https://github.com/libp2p/specs/blob/8b89dc2521b48bf6edab7c93e8129156a7f5f02c/kad-dht/README.md#rpc-messages)

protocols/kad/src/behaviour.rs
montekki: I will bring back taking `key` by value since that's what the other methods do

tomaka:
> The Kademlia iterative query should stop whenever it finds a record. However this is vulnerable to misbehaviour (a node can return a bad record and stop the iterative query by itself). We need some sort of way to ask the user whether a record is satisfactory.

I just realized the "stupidity" of that remark.

In the regular Kademlia protocol, the key of a record is its hash, which eliminates the question of knowing whether a record is valid.

In the situation of Polkadot, a record would be valid if it contains a signature using the key of the record as a public key.

In order to solve this question, I think that we can simply introduce a trait that checks the validity in a self-contained way (ie. a pure function).
We could also add a template parameter to Kademlia representing the type of a record, and require a `TryFrom<UnverifiedRecord>` trait implementation.

protocols/kad/src/handler.rs
montekki: Actually, [there is a response](https://github.com/libp2p/go-libp2p-kad-dht/blob/master/handlers.go#L148)  for any message but the `handleAddProvider` which returns a `nil, nil`

protocols/kad/src/behaviour.rs
```diff
@@ -170,6 +191,16 @@ impl QueryInfo {
                 key: unimplemented!(), // TODO: target.clone(),
                 user_data,
             },
+            QueryInfoInner::GetValue { key, .. } => {
+                KademliaHandlerIn::GetValue {
+                    key: key.clone(),
+                    user_data,
+                }
+            },
+            QueryInfoInner::PutValue { key, value } => KademliaHandlerIn::PutValue {
```
```suggestion
            QueryInfoInner::PutValue { key, .. } => KademliaHandlerIn::FindNodeReq {
```

I think this should actually map onto `KademliaHandlerIn::FindNodeReq`. A query is always an iterative lookup for the closest nodes to some key. An iterative `PUT_VALUE` is basically just an iterative `FIND_NODE` followed by a primitive `STORE` (or in libp2p context `PUT_VALUE`) to the `k` closest nodes found, [which happens here](https://github.com/libp2p/rust-libp2p/pull/1144/files#diff-88d683446d7fa7abbd7a347fb7d7c540R854) once the query has finished.

protocols/kad/src/behaviour.rs
romanb: Are you sure about that? As far as I'm aware only the sender creates a query and drives it to completion, so this code still runs on the side of the sender just like [this earlier insertion](https://github.com/libp2p/rust-libp2p/pull/1144/files#diff-88d683446d7fa7abbd7a347fb7d7c540R326), whereas the receiver adds it [here](https://github.com/libp2p/rust-libp2p/pull/1144/files#diff-88d683446d7fa7abbd7a347fb7d7c540R703). Am I really mistaken?

protocols/kad/src/behaviour.rs
```diff
@@ -283,6 +315,18 @@ impl<TSubstream> Kademlia<TSubstream> {
         self.start_query(QueryInfoInner::GetProviders { target, pending_results: Vec::new() });
     }
 
+    /// Starts an iterative `GET_VALUE` request.
+    pub fn get_value(&mut self, key: Multihash) {
+        self.start_query(QueryInfoInner::GetValue { key, results: vec![] });
```
romanb: It would be good if `get_value` always first checks its local storage and if `Q=1` (which we can start with and may even have as the default), then we can short-circuit and immediately queue a `KademliaOut::GetValueRes`, without issuing a query, provided that we don't constrain ourselves to always return closest peers in the result, but only when the value was not found, as I suggest in another comment.

protocols/kad/src/behaviour.rs
```diff
@@ -802,6 +909,14 @@ pub enum KademliaOut {
         /// List of peers ordered from closest to furthest away.
         closer_peers: Vec<PeerId>,
     },
+
+    /// Result of a `GET_VALUE` query
+    GetValueResult {
+        /// The result that we have probably received
+        result: Option<(Multihash, Vec<u8>)>,
+        /// List of peers ordered from closes to furthest from the key
+        closer_peers: Vec<PeerId>,
```
romanb: I think a `GetValueResult` should actually not contain `closer_peers` if a value is found, i.e. it should either contain the value, or the closest peers. This allows to short-circuit `get_value` if the key is found in the local storage (and `Q=1`).

protocols/kad/src/behaviour.rs
romanb: The ability to short-circuit when first looking into `self.records` also provides an argument for taking `key` by reference here, because it may then actually not need to be cloned (only if the key is not found in `self.records` or `Q > 1`.

protocols/kad/src/behaviour/test.rs
romanb: Even if there is no response to `PUT_VALUE` there should still be a `KademliaOut::PutValueResult` when the query is complete. When that event is emitted, it must be guaranteed that the local node has stored the value and at that point a `get_value` from another node for that key must always succeed. Currently the test may rely on the fact that `put_value` immediately puts the record into local storage, even before starting the query, but it is better not to make that assumption here.

protocols/kad/src/protocol.rs
romanb: Interestingly, if I'm reading it correctly, the go implementation [seems to suggest](https://github.com/libp2p/go-libp2p-kad-dht/blob/cf9bd9b649fb7df4a5632357a17d86d5ccc28649/dht.go#L161-L177) that there is indeed a response and that it even contains both `key` and `value` again, which they check whether it matches what they sent.

romanb: Similarly [for javascript](https://github.com/libp2p/js-libp2p-kad-dht/blob/8d6106b9c9dcf742f1b1985e464c2aa1955464b6/src/private.js#L242-L251). I cannot say that I currently understand why it is done that way.

montekki: Yes, In the other comment I've pasted a link to handlers in the `handlers.go`, they send out responses for everything but `AddProvider`

protocols/kad/src/protocol.rs
montekki: but this is done in all other message types.

protocols/kad/src/behaviour.rs
montekki: Oh, yes you are right

protocols/kad/src/behaviour.rs
montekki: It looks like it's checked [in the go implementation](https://github.com/libp2p/go-libp2p-kad-dht/blob/cf9bd9b649fb7df4a5632357a17d86d5ccc28649/dht.go#L200)

protocols/kad/src/behaviour.rs
```diff
@@ -283,6 +316,26 @@ impl<TSubstream> Kademlia<TSubstream> {
...
+                    closer_peers: None,
+                }
+            ));
+        }
+        self.start_query(QueryInfoInner::GetValue { key: key.clone(), results: vec![] });
+    }
+
+    /// Starts an iterative `PUT_VALUE` request
+    pub fn put_value(&mut self, key: Multihash, value: Vec<u8>) {
+        // TODO: Probably we shouldn't store the value ourselves
```
romanb: Right, this really shouldn't be here. The local node should only store the value if it is itself among the `k` closest to the key and it can only know that once the query is complete.

protocols/kad/src/behaviour.rs
```diff
@@ -623,6 +705,66 @@ where
...
+                    },
+                });
+            }
+            KademliaHandlerEvent::PutValueRes {
+                key: _,
+                user_data,
+            } => {
+                if let Some(query) = self.active_queries.get_mut(&user_data) {
+                    // A hacky way to end a query
+                    query.inject_rpc_result(&source, vec![source.clone()]);
```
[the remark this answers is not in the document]
montekki: Yes, I have re-read the specs and now I see that you're right about put_value being a two-step process with finding the closest peers first, whereas I believed that we just need to send values to 20 closest peers we already know of.
I will try to follow the steps you described now.

tomaka:
I think that the `RecordStore` trait should be asynchronous, at least when retrieving a value, but also potentially when storing one.

But again, this PR is supposed to be a first prototype that gets improved later.

protocols/kad/src/behaviour/test.rs
```diff
@@ -223,3 +240,96 @@ fn unresponsive_not_returned_indirect() {
...
+                }
+            }
+
+            Ok(Async::NotReady)
+        }))
+        .unwrap()
+}
+
+#[test]
+fn put_value() {
```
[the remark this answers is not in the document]
montekki: Done

protocols/kad/src/behaviour.rs
```diff
@@ -452,11 +563,17 @@ impl<TSubstream> Kademlia<TSubstream> {
...
+    fn has_record(&self, key: &Multihash) -> bool {
+        self.records.get(key).is_some()
+    }
 }
 
-impl<TSubstream> NetworkBehaviour for Kademlia<TSubstream>
+impl<TSubstream, TRecordStorage> NetworkBehaviour for Kademlia<TSubstream, TRecordStorage>
 where
     TSubstream: AsyncRead + AsyncWrite,
+    TRecordStorage: RecordStore + Default,
```
romanb: Why the `Default` constraint here?

protocols/kad/src/behaviour.rs
```diff
@@ -712,6 +906,30 @@ where
                 }
             }
 
+            let mut finished_write = None;
+
+            for (&query_id, write) in self.active_writes.iter_mut() {
```
romanb: Why `iter_mut`? How about:

```rust
let finished_write = self.active_writes.iter()
    .find_map(|(&query_id, write)|
        if write.done() {
            Some(query_id)
        } else {
            None
        })
```

protocols/kad/src/behaviour.rs
```diff
@@ -753,6 +971,34 @@ where
                             self.queued_events.push(event);
                         }
                     },
+                    QueryInfoInner::GetValue { key: _, results } => {
+                        let result = match results.first() {
```
```suggestion
                        // TODO: Return all results. The query should finish as soon as enough results have been collected.
                        let result = match results.into_iter().next() {
```
With `into_iter().next()` there is no need to clone the record one line below, as it is moved.

protocols/kad/src/behaviour/test.rs
```diff
@@ -223,3 +224,105 @@ fn unresponsive_not_returned_indirect() {
...
+    let target_key = multihash::encode(Hash::SHA2256, &vec![1,2,3]).unwrap();
+    swarms[0].get_value(&target_key);
+
+    Runtime::new().unwrap().block_on(
+        future::poll_fn(move || -> Result<_, io::Error> {
+            for swarm in &mut swarms {
+                loop {
+                    match swarm.poll().unwrap() {
+                        Async::Ready(Some(KademliaOut::GetValueResult(result))) => {
+                            assert_eq!(result, GetValueResult::NotFound{ closest_peers: vec![]});
```
romanb: `closest_peers` should not be empty here.

protocols/kad/src/protocol.rs
```diff
@@ -371,11 +455,17 @@ fn proto_to_req_msg(mut message: proto::Message) -> Result<KadRequestMsg, io::Er
     match message.get_field_type() {
         proto::Message_MessageType::PING => Ok(KadRequestMsg::Ping),
 
-        proto::Message_MessageType::PUT_VALUE =>
-            Err(invalid_data("Unsupported: PUT_VALUE message.")),
+        proto::Message_MessageType::PUT_VALUE => {
+            let record = message.mut_record();
+            let key = Multihash::from_bytes(record.take_key()).map_err(invalid_data)?;
+            Ok(KadRequestMsg::PutValue { key, value: record.take_value().to_vec() })
```
romanb: Does `take_value()` not already return a `Vec<u8>`? I.e. why `to_vec()`?

protocols/kad/src/protocol.rs
```diff
@@ -414,8 +504,24 @@ fn proto_to_resp_msg(mut message: proto::Message) -> Result<KadResponseMsg, io::
...
         proto::Message_MessageType::PING => Ok(KadResponseMsg::Pong),
 
-        proto::Message_MessageType::GET_VALUE =>
-            Err(invalid_data("Unsupported: GET_VALUE message")),
+        proto::Message_MessageType::GET_VALUE => {
+            let result = match message.has_record() {
+                true => {
+                    let mut record = message.take_record();
+                    let key = Multihash::from_bytes(record.take_key()).map_err(invalid_data)?;
+                    Some(Record::new(key, record.take_value().into()))
```
romanb: Why `.into()`?

protocols/kad/src/behaviour.rs
montekki: The trait bound from the [this impl block for Kademlia](https://github.com/montekki/rust-libp2p/blob/fs-kademlia-session-key-lookups/protocols/kad/src/behaviour.rs#L215) brings it here since we use methods from it.

protocols/kad/src/behaviour.rs
```diff
@@ -623,6 +721,93 @@ where
...
+                        .into_target_and_closest_peers();
+
+                    match query_info.inner {
+                        QueryInfoInner::GetValue { key: _, results, .. } => {
+                            let result = GetValueResult::Found { results };
+                            let event = KademliaOut::GetValueResult(result);
+
+                            self.queued_events.push(NetworkBehaviourAction::GenerateEvent(event));
+                        }
+                        _ => (),
```
romanb: Should be a `panic!` or `unreachable!`?

protocols/kad/src/behaviour.rs
```diff
@@ -761,6 +1008,24 @@ where
     }
 }
 
+/// The result of a `GET_VALUE` query.
+#[derive(Debug, Clone, PartialEq)]
+pub enum GetValueResult {
+    /// The results received from peers.
+    Found { results: Vec<Record> },
```
romanb: Since you didn't like my suggestion for splitting the results into head/tail, shouldn't it at least be documented that `results` is never empty here? It is otherwise not clear to the user at all whether `results` can be empty and if so, what to do about it.

protocols/kad/src/lib.rs
```diff
@@ -24,14 +24,17 @@
 //       be useful later for record store
 #![allow(dead_code)]
 
-pub use self::behaviour::{Kademlia, KademliaOut};
+pub use self::behaviour::{Kademlia, GetValueResult, KademliaOut};
```
romanb:
```suggestion
pub use self::behaviour::{Kademlia, KademliaOut, GetValueResult, PutValueResult};
```

protocols/kad/src/protocol.rs
```diff
@@ -326,6 +357,24 @@ fn req_msg_to_proto(kad_msg: KadRequestMsg) -> proto::Message {
             msg.mut_providerPeers().push(provider_peer.into());
             msg
         }
+        KadRequestMsg::GetValue { key } => {
+            let mut msg = proto::Message::new();
+            msg.set_field_type(proto::Message_MessageType::GET_VALUE);
+            msg.set_clusterLevelRaw(10);
+            msg.set_key(key.as_bytes().to_vec());
```
romanb:
```suggestion
            msg.set_key(key.into_bytes());
```

protocols/kad/src/protocol.rs
```diff
@@ -361,6 +410,41 @@ fn resp_msg_to_proto(kad_msg: KadResponseMsg) -> proto::Message {
...
+            msg.set_field_type(proto::Message_MessageType::GET_VALUE);
+            msg.set_clusterLevelRaw(9);
+            for peer in closer_peers {
+                msg.mut_closerPeers().push(peer.into());
+            }
+
+            if let Some(Record{ key, value }) = result {
+                let mut record = proto::Record::new();
+                record.set_key(key.into_bytes());
+                record.set_value(value.to_vec());
```
romanb:
```suggestion
                record.set_value(value);
```

protocols/kad/src/protocol.rs
```diff
@@ -446,8 +551,18 @@ fn proto_to_resp_msg(mut message: proto::Message) -> Result<KadResponseMsg, io::
...
+        proto::Message_MessageType::PUT_VALUE => {
+            let key = Multihash::from_bytes(message.take_key()).map_err(invalid_data)?;
+            if !message.has_record() {
+                return Err(invalid_data("received PUT_VALUE message with no record"));
+            }
+
+            let mut record = message.take_record();
+            Ok(KadResponseMsg::PutValue {
+                key,
+                value: record.take_value().into(),
```
romanb:
```suggestion
                value: record.take_value(),
```

protocols/kad/src/record.rs
```diff
@@ -0,0 +1,107 @@
...
+pub struct Record {
+    /// Key of the record.
+    pub key: Multihash,
+    /// Value of the record.
+    pub value: Vec<u8>,
+}
+
+impl Record {
+    /// Creates a new `Record` with a given key and value.
+    pub fn new(key: Multihash, value: Vec<u8>) -> Self { Record { key, value } }
```
romanb: This function is no longer needed, since the Record constructor is public.

protocols/kad/src/record.rs
```diff
@@ -0,0 +1,107 @@
...
+impl Default for MemoryRecordStorage {
+    fn default() -> Self {
+        MemoryRecordStorage::new(Self::MAX_RECORDS, Self::MAX_RECORD_SIZE)
+    }
+}
+
+impl RecordStore for MemoryRecordStorage {
+    fn get(&self, k: &Multihash) -> Option<Cow<Record>> {
+        match self.records.get(k) {
+            Some(rec) => Some(Cow::Owned(rec.to_owned())),
```
romanb:
```suggestion
            Some(rec) => Some(Cow::Borrowed(rec)),
```
The whole point of using `Cow` is that an implementation of `RecordStore` can use `Borrowed` if it keeps the record in-memory.

protocols/kad/src/behaviour.rs
```diff
@@ -283,6 +333,43 @@ impl<TSubstream> Kademlia<TSubstream> {
...
+        if let Some(record) = self.records.get(key) {
+            results.push(record.into_owned());
+            if num_results == 1 {
+                self.queued_events.push(NetworkBehaviourAction::GenerateEvent(
+                    KademliaOut::GetValueResult(
+                        GetValueResult::Found { results }
+                )));
+                return;
+            } else {
+                num_results -= 1;
```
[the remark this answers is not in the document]
montekki: I don't think this is right: we were requested for 10 results, we found one of them in our local storage, we need to ask the network for 9 results.

protocols/kad/src/behaviour.rs
```diff
@@ -23,20 +23,22 @@ use crate::handler::{KademliaHandler, KademliaHandlerEvent, KademliaHandlerIn};
...
-use std::{borrow::Cow, error, marker::PhantomData, time::Duration};
+use std::{borrow::Cow, error, marker::PhantomData, num::NonZeroU8, time::Duration};
 use tokio_io::{AsyncRead, AsyncWrite};
 use wasm_timer::{Instant, Interval};
 
 mod test;
 
 /// Network behaviour that handles Kademlia.
-pub struct Kademlia<TSubstream> {
+pub struct Kademlia<TSubstream, TRecordStorage:RecordStore = MemoryRecordStorage> {
```
```suggestion
pub struct Kademlia<TSubstream, TRecordStorage: RecordStore = MemoryRecordStorage> {
```

protocols/kad/src/behaviour.rs
```diff
@@ -283,6 +333,45 @@ impl<TSubstream> Kademlia<TSubstream> {
         self.start_query(QueryInfoInner::GetProviders { target, pending_results: Vec::new() });
     }
 
+    /// Starts an iterative `GET_VALUE` request.
+    ///
+    /// Returns a number of results that is in the inerval [1, 20],
```
```suggestion
    /// Returns a number of results that is in the interval [1, 20],
```

protocols/kad/src/behaviour.rs
tomaka: It should be removed from both impl blocks, and added back only where relevant.
```suggestion
    TRecordStorage: RecordStore,
```

protocols/kad/src/behaviour.rs
```diff
@@ -170,35 +199,54 @@ impl QueryInfo {
...
+                user_data,
+            }
         }
     }
 }
 
-impl<TSubstream> Kademlia<TSubstream> {
+impl<TSubstream, TRecordStorage> Kademlia<TSubstream, TRecordStorage>
+where
+    TRecordStorage: Default + RecordStore
```
```suggestion
    TRecordStorage: RecordStore
```

protocols/kad/src/behaviour.rs
```diff
@@ -170,35 +199,54 @@ impl QueryInfo {
...
+{
     /// Creates a `Kademlia`.
     #[inline]
     pub fn new(local_peer_id: PeerId) -> Self {
```
```suggestion
    pub fn new(local_peer_id: PeerId) -> Self
    where
        TRecordStorage: Default
    {
```

protocols/kad/src/behaviour.rs
```diff
@@ -170,35 +199,54 @@ impl QueryInfo {
...
-        Self::new_inner(local_peer_id)
+        Self::new_inner(local_peer_id, Default::default())
     }
 
     /// The same as `new`, but using a custom protocol name.
     ///
     /// Kademlia nodes only communicate with other nodes using the same protocol name. Using a
     /// custom name therefore allows to segregate the DHT from others, if that is desired.
     pub fn with_protocol_name(local_peer_id: PeerId, name: impl Into<Cow<'static, [u8]>>) -> Self {
```
```suggestion
    pub fn with_protocol_name(local_peer_id: PeerId, name: impl Into<Cow<'static, [u8]>>) -> Self
    where
        TRecordStorage: Default
    {
```

protocols/kad/src/behaviour.rs
```diff
@@ -170,35 +199,54 @@ impl QueryInfo {
...
+        Self::new_inner(local_peer_id, records)
+    }
+
     /// Creates a `Kademlia`.
     ///
     /// Contrary to `new`, doesn't perform the initialization queries that store our local ID into
     /// the DHT and fill our buckets.
     #[inline]
     #[deprecated(note="this function is now equivalent to new() and will be removed in the future")]
     pub fn without_init(local_peer_id: PeerId) -> Self {
```
```suggestion
    pub fn without_init(local_peer_id: PeerId) -> Self
        where TRecordStorage: Default
    {
```

protocols/kad/src/behaviour.rs
```diff
@@ -623,6 +723,93 @@ where
...
+                        .into_target_and_closest_peers();
+
+                    match query_info.inner {
+                        QueryInfoInner::GetValue { key: _, results, .. } => {
+                            let result = GetValueResult::Found { results };
+                            let event = KademliaOut::GetValueResult(result);
+
+                            self.queued_events.push(NetworkBehaviourAction::GenerateEvent(event));
+                        }
+                        _ => panic!("QueryInfoInner::GetValue was expected at the GetValueRes query id; QED.")
```
tomaka: That's not really a convincing proof, but it's ok if we change it later.
```suggestion
                        // TODO: write a better proof
                        _ => panic!("QueryInfoInner::GetValue was expected at the GetValueRes query id; QED.")
```

protocols/kad/src/behaviour.rs
```diff
@@ -753,6 +963,45 @@ where
...
+                            },
+                            _ => GetValueResult::Found{ results },
+                        };
+
+                        let event = KademliaOut::GetValueResult(result);
+
+                        break Async::Ready(NetworkBehaviourAction::GenerateEvent(event));
+                    },
+                    QueryInfoInner::PutValue { key, value } => {
+                        use std::iter::FromIterator;
```
tomaka: Let's put all the imports on top of the file.

```suggestion
```

protocols/kad/src/record.rs
```diff
@@ -0,0 +1,102 @@
...
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct Record {
+    /// Key of the record.
+    pub key: Multihash,
+    /// Value of the record.
+    pub value: Vec<u8>,
+}
+
+/// Trait for a record store.
+pub trait RecordStore {
```
tomaka: As mentioned, I think the methods here should be asynchronous. We should open an issue about that when the PR gets merged.

montekki:
When this is merged, it would be great to avoid including  whole history of 120+ commits into the squashed commit btw

protocols/kad/src/behaviour.rs
```diff
@@ -623,6 +731,93 @@ where
...
+                        .into_target_and_closest_peers();
+
+                    match query_info.inner {
+                        QueryInfoInner::GetValue { key: _, results, .. } => {
+                            let result = GetValueResult::Found { results };
+                            let event = KademliaOut::GetValueResult(result);
+
+                            self.queued_events.push(NetworkBehaviourAction::GenerateEvent(event));
+                        }
+                        _ => panic!("unexpected query_info.inner variant for a get_value result; QED.")
```
tomaka: This is still not a good proof. It does indicate what is wrong, but the point is to prove that things cannot be wrong.
I didn't intent that you try to fix it now, and fixing it probably requires code changes.

```suggestion
                        // TODO: write a better proof
                        _ => panic!("unexpected query_info.inner variant for a get_value result; QED.")
```

PR #1146, opened 2019-05-24
Add deflate protocol

### LINKED ISSUE #955: Add a DEFLATE protocol?
Similar to `plaintext`, but applies inflate/deflate on the packets.
I'm not sure whether tha's really useful, but it shouldn't be hard to do.

aknuds1:
Add deflate protocol; fixes #955.

I hope I did this right, although I'm not 100% sure being so new to rust-libp2p and the surrounding ecosystem.

### DIFF
--- a/protocols/deflate/Cargo.toml
+++ b/protocols/deflate/Cargo.toml
@@ -1,0 +1,23 @@
+[package]
+name = "libp2p-deflate"
+edition = "2018"
+description = "Deflate encryption protocol for libp2p"
+version = "0.1.0"
+authors = ["Parity Technologies <admin@parity.io>"]
+license = "MIT"
+repository = "https://github.com/libp2p/rust-libp2p"
+keywords = ["peer-to-peer", "libp2p", "networking"]
+categories = ["network-programming", "asynchronous"]
+
+[dependencies]
+futures = "0.1"
+libp2p-core = { version = "0.8.1", path = "../../core" }
+tokio-io = "0.1.12"
+flate2 = { version = "1.0", features = ["tokio"] }
+
+[dev-dependencies]
+env_logger = "0.6"
+libp2p-tcp = { version = "0.8.0", path = "../../transports/tcp" }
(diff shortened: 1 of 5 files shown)


protocols/deflate/tests/test.rs
```diff
@@ -0,0 +1,98 @@
...
+    let _ = env_logger::try_init();
+
+    fn prop(message: Vec<u8>) -> bool {
+        let server_transport = TcpConfig::new().with_upgrade(DeflateConfig {});
+        let client_transport = TcpConfig::new().with_upgrade(DeflateConfig {});
+        run(server_transport, client_transport, message);
+        true
+    }
+
+    QuickCheck::new()
```
aknuds1: I'm not sure if there's any need for using this, I copied it from another protocol test.

src/lib.rs
```diff
@@ -200,6 +200,8 @@ pub use libp2p_wasm_ext as wasm_ext;
 pub use libp2p_websocket as websocket;
 #[doc(inline)]
 pub use libp2p_yamux as yamux;
+#[doc(inline)]
```
tomaka: Nit: I think the crates are sorted alphabetically here.

aknuds1: Thanks @tomaka - will fix it.

aknuds1: Fixed.

protocols/deflate/Cargo.toml
```diff
@@ -0,0 +1,23 @@
+[package]
+name = "libp2p-deflate"
+edition = "2018"
+description = "Deflate encryption protocol for libp2p"
+version = "0.8.1"
```
tomaka: Oh, also, this should be `0.1.0`.

aknuds1: @tomaka Good catch! I actually thought it should be 0.8.1 since it looked as if the other protocols followed the version of the core crate. Should have asked you, but didn't think of it.

Issue #1103, opened 2019-05-05
Add a WebSocket transports that outputs Sink/Stream of packets


tomaka:
The WebSockets protocol is about sending frames: either ping/pong frames, text, or binary.

There should be a WebSockets transport in libp2p that gives access to that by having an `Output` that implements the `Stream` and `Sink` traits, where the item is some enum (ping, pong, text, binary).

The "real" WebSockets transport can then be implemented on top of that by wrapping the output in an `RwStreamSink`.

tomaka:
In addition to arguably being a better way to organize the code, the use-case is that I'd like to replace Substrate's telemetry with libp2p, and we need to be able to send precisely-delimited frames of data (as opposed to writing to a stream of bytes).

montekki:
does libp2p have a `Sink + Stream` trait somewhere btw?

tomaka:
> does libp2p have a Sink + Stream trait somewhere btw?

If you're asking about a trait that combines the two, then no.

twittner:
https://github.com/libp2p/rust-libp2p/pull/1150 implemented a strategy similar to this.

Issue #1094, opened 2019-04-29
We don't use TLS with /wss


tomaka:
When using WebSockets with `/wss`, we should connect to the server through HTTPS and not just HTTP. Right now we're using `async_connect_on`, which doesn't do that.

cc #849 

In general, I think that the libp2p-websocket crate should just be rewritten.

noahwolf2001:
hello

twittner:
Implemented.

burdges:
You've a reasonable TLS for this then?  :)

twittner:
We use https://crates.io/crates/tokio-rustls which in turn uses https://crates.io/crates/rustls and https://crates.io/crates/webpki.

burdges:
so ring under the hood, fine.  :)

Issue #369, opened 2018-07-25
The Send/Sync situation


tomaka:
Right now we use lots of `Box`es at various locations in the library, but most of them are not `Send` or `Sync`. This should be changed, or the boxes removed.

tomaka:
The only `Box<Future>` left are in `protocols/observed` and `transports/websocket/browser`.

romanb:
> The only Box<Future> left are in protocols/observed and transports/websocket/browser.

That code no longer exists. I think if there is something specific left to do here, we should create new issues that target specific uses of `Box`es. Feel free to reopen if you think this issue should still be kept open, despite its generality.

PR #1150, opened 2019-05-28
Reimplement the websocket transport.

twittner:
Actually two transports are implemented, one in the `framed` module whose output implements `Stream` and `Sink`, but not `AsyncRead` or `AsyncWrite`, and on top of that another transport which merely wraps the output of the former in a `RwStreamSink` to produce an output type which implements `AsyncRead` and `AsyncWrite`. The output of the `framed` transport produces and consumes just bytes which correspond to the actual application data in a websocket frame. (See https://github.com/libp2p/rust-libp2p/issues/1103 for details.)

The transport(s) now handle TLS, i.e. they use TLS for `/wss` multi-addresses (See https://github.com/libp2p/rust-libp2p/issues/1094 for details.)

I have created a version of the chat example to work with the websocket transport. If you want to try it you can find it here: https://gist.github.com/twittner/1c9ea9fcd538d5e6727cd6bb56c1a43d.

The transport uses the not yet published [soketto](https://github.com/paritytech/soketto) crate which is a heavily modified fork of the twist crate. If and when we agree to merge this PR soketto should be published to `crates.io` and the git reference in the `Cargo.toml` of `libp2p-websocket` should be replaced with a proper version number.


transports/websocket/src/framed.rs
```diff
@@ -0,0 +1,431 @@
...
+/// A [`Stream`] and [`Sink`] that produces and consumes [`BytesMut`] values
+/// which correspond to the payload data of websocket frames.
+#[derive(Debug)]
+pub struct BytesConnection<T> {
+    inner: Connection<EitherOutput<EitherOutput<client::TlsStream<T>, server::TlsStream<T>>, T>>
+}
+
+impl<T: AsyncRead + AsyncWrite> Stream for BytesConnection<T> {
+    type Item = BytesMut;
+    type Error = io::Error;
```
tomaka: Why is that `io::Error` instead of the error that `self.inner.poll` would return?

transports/websocket/src/framed.rs
```diff
@@ -0,0 +1,431 @@
...
+
+    fn poll(&mut self) -> Poll<Option<Self::Item>, Self::Error> {
+        let data = try_ready!(self.inner.poll().map_err(|e| io::Error::new(io::ErrorKind::Other, e)));
+        Ok(Async::Ready(data.map(base::Data::into_bytes)))
+    }
+}
+
+impl<T: AsyncRead + AsyncWrite> Sink for BytesConnection<T> {
+    type SinkItem = BytesMut;
+    type SinkError = io::Error;
```
tomaka: Same remark.

transports/websocket/src/lib.rs
```diff
@@ -1,4 +1,4 @@
-// Copyright 2017 Parity Technologies (UK) Ltd.
+// Copyright 2017 - 2019 Parity Technologies (UK) Ltd.
```
tomaka:
```suggestion
// Copyright 2017-2019 Parity Technologies (UK) Ltd.
```

transports/websocket/src/lib.rs
```diff
@@ -18,66 +18,182 @@
...
+
+impl<T> From<framed::WsConfig<T>> for WsConfig<T> {
+    fn from(framed: framed::WsConfig<T>) -> Self {
+        WsConfig {
+            transport: framed
+        }
+    }
+}
+
+impl<T> AsRef<framed::WsConfig<T>> for WsConfig<T> {
```
tomaka: Considering that the `WsConfig` here is supposed to be the mostly-used one, I think that all of its methods should simply be copy-pasted here as well. We want to make the user believe that they are two different types.

transports/websocket/src/framed.rs
```diff
@@ -0,0 +1,431 @@
...
+use url::Url;
+
+/// Max. number of payload bytes of a single frame.
+const MAX_DATA_SIZE: u64 = 256 * 1024 * 1024;
+
+/// A Websocket transport whose output type is a [`Stream`] and [`Sink`] of
+/// frame payloads which does not implement [`AsyncRead`] or
+/// [`AsyncWrite`]. See [`crate::WsConfig`] if you require the latter.
+#[derive(Debug, Clone)]
+pub struct WsConfig<T> {
```
tomaka: What about `FramedWsConfig`? So that you actually know that this is not the regular `WsConfig`.

transports/websocket/src/framed.rs
```diff
@@ -0,0 +1,431 @@
...
+    T: Transport + Send + Clone + 'static,
+    T::Error: Send + 'static,
+    T::Dial: Send + 'static,
+    T::Listener: Send + 'static,
+    T::ListenerUpgrade: Send + 'static,
+    T::Output: AsyncRead + AsyncWrite + Send + 'static
+{
+    type Output = BytesConnection<T::Output>;
+    type Error = Error<T::Error>;
+    type Listener = Box<dyn Stream<Item = ListenerEvent<Self::ListenerUpgrade>, Error = Self::Error> + Send>;
```
tomaka: Boxing goes in the opposite direction of #369.

transports/websocket/src/tls.rs
```diff
@@ -0,0 +1,176 @@
...
+use tokio_rustls::{
+    TlsConnector,
+    TlsAcceptor,
+    rustls,
+    webpki
+};
+
+/// TLS configuration.
+#[derive(Clone)]
+pub struct Config {
```
tomaka: `TlsConfig`?

transports/websocket/src/tls.rs
```diff
@@ -0,0 +1,176 @@
...
+
+/// Setup the rustls client configuration.
+fn client_config() -> rustls::ClientConfig {
+    let mut client = rustls::ClientConfig::new();
+    client.root_store.add_server_trust_anchors(&webpki_roots::TLS_SERVER_ROOTS);
+    client
+}
+
+/// TLS configuration builder.
+pub struct Builder {
```
tomaka: `TlsConfigBuilder`?

transports/websocket/src/framed.rs
twittner: I do not like the repetition on namespaces or module names in identifiers. Instead I prefer to import the module and refer to those identifiers with the module prefix, i.e. `io::Error` instead of `IoError`, `framed::WsConfig` instead of `FramedWsConfig`, `tls::Config` instead of `TlsConfig`. Otherwise names are just blown up for no good reason.

transports/websocket/src/tls.rs
twittner: https://github.com/libp2p/rust-libp2p/pull/1150#discussion_r289308861

transports/websocket/src/tls.rs
twittner: https://github.com/libp2p/rust-libp2p/pull/1150#discussion_r289308861

transports/websocket/src/framed.rs
twittner: I would not say it is the "opposite direction", but that it does not improve w.r.t. the boxing situation in the websocket transport, considering that it [uses Boxes here](https://github.com/libp2p/rust-libp2p/blob/6a2a401458529d79a5398f770711235ac6bddb11/transports/websocket/src/desktop.rs#L74-L78) as well.

transports/websocket/src/framed.rs
tomaka: It's not so much a problem of naming  as it is a problem of namespaces. I prefer having one `websocket` namespace with `websocket::TlsConfig`, `websocket::FramedConfig`, etc. rather than `framed::Config` or `tls::Config` which don't indicate much.

twittner: This naming scheme reintroduces namespaces as part of identifier names.
`websocket::framed::Config` and `websocket::tls::Config` use the language provided mechanism for name spacing, whereas `websocket::FramedConfig` and `websocket::TlsConfig` do not.

transports/websocket/src/framed.rs
twittner: The inner error is a `soketto::connection::Error` and I wanted to hide the particular websocket implementation we use.

transports/websocket/src/framed.rs
twittner: https://github.com/libp2p/rust-libp2p/pull/1150#discussion_r289321274

transports/websocket/src/framed.rs
twittner: Oh, and `RwStreamSink` requires `io::Error` anyway for the `Stream`/`Sink` type it is wrapping.

tomaka:
Any opposition to merging?

twittner:
No. We just need to publish `soketto` to crates.io before releasing the next version of `libp2p` and replace the git reference with a version number.

tomaka:
> No. We just need to publish soketto to crates.io before releasing the next version of libp2p and replace the git reference with a version number.

What about publishing now? Or do you have more changes to do to it?

twittner:
There are some breaking changes coming to soketto, so if you do not plan to publish another libp2p version today or tomorrow I would like to postpone the publishing of soketto a bit. But you could still merge this PR as it is and we update the Cargo.toml then.

folex:
@twittner Hi! Sorry for posting in an old PR, but do you still have websocket chat example available somewhere?

> I have created a version of the chat example to work with the websocket transport. If you want to try it you can find it here: https://gist.github.com/twittner/1c9ea9fcd538d5e6727cd6bb56c1a43d.

Gives 404 :(

Thanks in advance!

twittner:
@folex: I have adapted the old example to work with `master`. Have a look here: https://gist.github.com/twittner/e18c227cf5f0202225573d674f93d5fc

PR #1151, opened 2019-05-29
Change score calculation.

### LINKED ISSUE #1135: Add a time-to-live to external addresses
When, through `identify`, we receive the information about our IP address, we compute an "external address" that goes in a registry. See the `core::swarm::registry` module.

However, while there is a limit to the number of addresses, these addresses never expire and remain valid forever.

Since there is no mechanism that decreases the score of an address, once you reach 30 addresses, any new address will likely remain at the last position of the list even if the first 29 addresses are no longer valid.

This is especially problematic for IPv6, as some home routers automatically change your IP address every few hours for privacy reasons.

twittner:
When adding an address it now has an initial score of 1 instead of 0. Adding an existing address causes their score to be increased as before but we also decrement the score of every other address. This approach still favours most often reported addresses but helps with certain
cases, such as periodic address changes since expired addresses will reach 0 and a fresh address will immediately move to second position before eventually claiming the first spot.

Fixes #1135.

### DIFF
--- a/core/src/swarm/registry.rs
+++ b/core/src/swarm/registry.rs
@@ -64,16 +64,17 @@ impl Addresses {
     pub fn add(&mut self, a: Multiaddr) {
         for r in &mut self.registry {
             if &r.addr == &a {
-                r.score = r.score.saturating_add(1);
-                isort(&mut self.registry);
-                return ()
+                r.score = r.score.saturating_add(1)
+            } else {
+                r.score = r.score.saturating_sub(1)
             }
         }
         if self.registry.len() == self.limit.get() {
             self.registry.pop();
         }
-        let r = Record { score: 0, addr: a };
-        self.registry.push(r)
+        let r = Record { score: 1, addr: a };
+        self.registry.push(r);
+        isort(&mut self.registry)
     }
 
     /// Return an iterator over all [`Multiaddr`] values.


tomaka:
So if I report A, then B, then A, then B, one will be at 1 and one at 0 constantly?

twittner:
Yes, like that:
```
add:    A        B              A              B
seq: []   [A(1)]   [B(1), A(0)]   [A(1), B(0)]   [B(1), A(0)]
```

tomaka:
Isn't it a problem that an address that is constantly reported stays at 0?

twittner:
Why do you think so? Do you have an example in mind where this would be problematic?

tomaka:
Imagine I report A, then B, then A, then B, then A, then B, one thousand times. Then I report C twice. C will immediately jump on top.

However if I report A one thousand times, then report C twice, C will stay at the bottom.

twittner:
C would go to the second position, because either A or B has a score of 1. And if you continue to alternate between A and B, C would move to the third place in the next round.

tomaka:
But the whole behaviour doesn't make sense.

If I report both A and B a thousand times, then C once, it should stay at the bottom, since that's the whole point of this mechanism. If I report C twice after reporting A and B a thousand times, it shouldn't immediately jump in the first position.

the order in which A and B are reported shouldn't influence that much the position of C.

twittner:
It seems I misunderstood what you wanted in https://github.com/libp2p/rust-libp2p/issues/1135#issuecomment-494060852. We should probably have a more formal outline before attempting the implementation.

PR #1152, opened 2019-05-31
Some improvements to the docs of NetworkBehaviour

tomaka:


core/src/swarm/behaviour.rs
```diff
@@ -115,28 +149,35 @@ pub enum NetworkBehaviourAction<TInEvent, TOutEvent> {
...
     },
 
-    /// Instructs the `Swarm` to send a message to a connected peer.
+    /// Instructs the `Swarm` to send a message to the handler dedicated to this node.
     ///
     /// If the `Swarm` is connected to the peer, the message is delivered to the remote's
     /// protocol handler. If there is no connection to the peer, the message is ignored.
     /// To ensure delivery, the `NetworkBehaviour` must keep track of connected peers.
+    ///
+    /// Keep in mind, though, that even if we are connected to the peer, connections can get closed
```
romanb:
```suggestion
    /// Note that even if the peer is currently connected, connections can get closed
```

core/src/swarm/behaviour.rs
```diff
@@ -115,28 +149,35 @@ pub enum NetworkBehaviourAction<TInEvent, TOutEvent> {
...
+    ///
     /// On success, [`NetworkBehaviour::inject_connected`] is invoked.
     /// On failure, [`NetworkBehaviour::inject_dial_failure`] is invoked.
     DialPeer {
         /// The peer to try reach.
         peer_id: PeerId,
     },
 
-    /// Instructs the `Swarm` to send a message to a connected peer.
+    /// Instructs the `Swarm` to send a message to the handler dedicated to this node.
```
romanb:
```suggestion
    /// Instructs the `Swarm` to send a message to the handler dedicated to the connection with the peer.
```

PR #1155, opened 2019-06-02
Kademlia: Fix premature query termination.

romanb:
This is a minimal fix for https://github.com/libp2p/rust-libp2p/issues/1105 extracted from https://github.com/libp2p/rust-libp2p/pull/1154 that may be worthwhile incorporating on its own and potentially useful for https://github.com/libp2p/rust-libp2p/pull/1144 (context https://github.com/montekki/rust-libp2p/pull/1#issuecomment-497923800).

### DIFF
--- a/protocols/kad/src/query.rs
+++ b/protocols/kad/src/query.rs
@@ -319,10 +319,12 @@ where
             }
 
             if let QueryPeerState::Succeeded = state {
-                succeeded_counter.as_mut().map(|c| *c += 1);
-                // If we have enough results; the query is done.
-                if succeeded_counter.unwrap_or(0) >= num_results {
-                    return Async::Ready(QueryStatePollOut::Finished)
+                if let Some(ref mut cnt) = succeeded_counter {
+                    *cnt += 1;
+                    // If we have enough results; the query is done.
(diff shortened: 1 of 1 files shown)


romanb:
@montekki Good catch, `NotContacted` needs to be handled with an early return if the peer cannot currently be contacted (max. parallelism). I have added another commit to address that. This is also already done correctly (and tested) in #1154 ([relevant line](https://github.com/libp2p/rust-libp2p/pull/1154/files#diff-07deebbc98f98560efc08ecf45d7018aR392)).

protocols/kad/src/query.rs
```diff
@@ -312,15 +312,16 @@ where
...
                     Ok(Async::NotReady) => {
+                        succeeded_counter = None;
                         active_counter += 1
                     }
                 }
             }
 
             if let QueryPeerState::Succeeded = state {
-                succeeded_counter += 1;
+                succeeded_counter.as_mut().map(|c| *c += 1);
```
tomaka: I really prefer this form:
```suggestion
                if let Some(c) = succeeded_counter.as_mut() { *c += 1; }
```

To me, the functional style is for immutability.

PR #1161, opened 2019-06-04
Add AsRef<u8> for Multihash.

romanb:
Main motivation is to simplify the `From<Multihash>` instance for `libp2p_kad::kbucket::Key` [here](https://github.com/libp2p/rust-libp2p/blob/master/protocols/kad/src/kbucket/key.rs#L86-L91) so that it looks like the instance for `PeerId` and avoids a additional clone of the multihash.


tomaka:
Could you bump the patch version of `multihash` as well? Once the PR is merged we can publish a new version and depend on it in `libp2p-kad`.
We almost never touch `multihash`, and the change is too small to put in a changelog anyway.

PR #727, opened 2018-12-04
Run rustfmt on muxers

tomaka:
Runs `rustfmt`. Please say a word if you think you have ongoing work that conflicts with this PR.

PR #1164, opened 2019-06-05
Format code using 'cargo fmt'

Atul9:


twittner:
It seems the output of rustfmt has not improved since https://github.com/libp2p/rust-libp2p/pull/727.

twittner:
Sorry to close this, but we have decided not to use rustfmt for now.

PR #1174, opened 2019-06-13
Kademlia: Address some TODOs - Refactoring - API updates.

romanb:
Sits on top of https://github.com/libp2p/rust-libp2p/pull/1154, addressing some left-over TODOs, as well as  some refactoring and preparation for further work on the (still incomplete) record and provider system.

The following left-over issues from earlier work have been addressed:

  * The key for FIND_NODE requests is generalised to any Multihash, instead of just peer IDs.
  * All queries get a (configurable) timeout.
  * Finishing queries as soon as enough results have been received is simplified, avoiding code duplication.
  * No more panics in provider-API-related code paths. The provider API is however still untested and (I think) still incomplete (e.g. expiration of provider records).
  * Numerous smaller TODOs encountered in the code.

The following public API changes / additions have been made:

  * Introduced a `KademliaConfig` with new configuration options for the replication factor and query timeouts.
  * Renamed `find_node` to `get_closest_peers`.
  * Renamed `get_value` to `get_record` and `put_value` to `put_record`, introducing a `Quorum` parameter for both functions, replacing the existing `num_results` parameter with clearer semantics.
  * Renamed `add_providing` to `start_providing` and `remove_providing` to `stop_providing`.
  * Added a `bootstrap` function that implements a (almost) standard Kademlia bootstrapping procedure.
  * Renamed `KademliaOut` to `KademliaEvent` with an updated list of constructors (some renaming). All events that report query results now report a `Result` to uniformly permit reporting of errors.

The following refactorings have been made:

  * Introduced some constants.
  * Consolidated `query.rs` and `write.rs` behind a common interface to reduce duplication and facilitate better code reuse, introducing the notion of a query peer "iterator". `query/peers/closest.rs` contains the code that was formerly in `query.rs`. `query/peers/fixed.rs` contains a modified variant of `write.rs` (which is removed). The new `query.rs` provides an interface for working with a collection of queries, taking over some code from `behaviour.rs`.
  * Reduced code duplication in tests and switched to the current_thread runtime for polling swarms to avoid spurious errors in the test output due to aborted connections when a test finishes prematurely (e.g. because a quorum of results has been collected).
  * Some additions / improvements to the existing tests.


protocols/kad/src/query/peers/closest.rs
```diff
@@ -0,0 +1,694 @@
...
+        self.num_waiting
+    }
+
+    /// Returns true if the iterator is waiting for a response from the given peer.
+    pub fn is_waiting(&self, peer: &PeerId) -> bool {
+        self.waiting().any(|p| peer == p)
+    }
+
+    /// Advances the state of the iterator, potentially getting a new peer to contact.
+    pub fn next(&mut self, now: Instant) -> PeersIterState {
```
tomaka: How is the `peer_timeout` enforced? What tells the caller that they should call `next()` after a certain number of seconds have passed?

protocols/kad/src/behaviour.rs
```diff
@@ -71,188 +65,110 @@ pub struct Kademlia<TSubstream, TRecordStorage: RecordStore = MemoryRecordStorag
...
-        TRecordStorage: Default
-    {
-        let mut me = Kademlia::new_inner(local_peer_id, Default::default());
-        me.protocol_name_override = Some(name.into());
-        me
+    /// > **Note**: A single query usually comprises at least as many requests
+    /// > as the replication factor, i.e. this is not a request timeout.
+    ///
+    /// The default is 60 seconds.
+    pub fn set_query_timeout(&mut self, timeout: Duration) -> &mut Self {
```
tomaka: Why have a query timeout at all? The query will end at some point, and an extra timeout just means that we might fail if our network is poor and the value is far away.

protocols/kad/src/behaviour.rs
```diff
@@ -836,229 +922,232 @@ where
...
-    NotFound {
-        key: Multihash,
-        closest_peers: Vec<PeerId>
-    }
+/// A quorum w.r.t. the configured replication factor specifies the minimum number of distinct
+/// nodes that must be successfully contacted in order for a query to succeed.
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub enum Quorum {
+    One,
+    Two,
```
tomaka: I find `One`, `Two` and `Three` a bit weird, as they are redundant with `N`. Even `Majority` and `All` are redundant, since the user could calculate the actual value themselves.

protocols/kad/src/behaviour.rs
```diff
@@ -286,96 +202,116 @@ where
...
+        let multihash = key.into();
+        let info = QueryInfo::GetClosestPeers { key: multihash.clone() };
+        let target = kbucket::Key::new(multihash);
+        let peers = self.kbuckets.closest_keys(&target);
+        let inner = QueryInner::new(info);
+        self.queries.add_iter_closest(target.clone(), peers, inner);
     }
 
-    /// Starts an iterative `GET_VALUE` request.
+    /// Performs a lookup for a record in the DHT.
```
tomaka: I would explicitly write that this is not instant.
```suggestion
    /// Starts performing a lookup for a record in the DHT.
```

protocols/kad/src/behaviour.rs
```diff
@@ -286,96 +202,116 @@ where
...
     }
 
-    /// Starts an iterative `PUT_VALUE` request
-    pub fn put_value(&mut self, key: Multihash, value: Vec<u8>) {
-        if let Err(error) = self.records.put(Record { key: key.clone(), value: value.clone() }) {
-            self.queued_events.push(NetworkBehaviourAction::GenerateEvent(
-                KademliaOut::PutValueResult(
-                    PutValueResult::Err { key, cause: error }
-                )
+    /// Stores a record in the DHT.
```
```suggestion
    /// Starts the process of storing a record in the DHT.
```

protocols/kad/src/behaviour.rs
```diff
@@ -671,43 +768,44 @@ where
...
-                    addresses: provider_peer.multiaddrs.clone(),
-                    ty: provider_peer.connection_ty,
-                }));
+                self.queued_events.push_back(NetworkBehaviourAction::GenerateEvent(
+                    KademliaEvent::Discovered {
+                        peer_id: provider_peer.node_id.clone(),
+                        addresses: provider_peer.multiaddrs.clone(),
+                        ty: provider_peer.connection_ty,
+                    }));
+                // TODO: Expire provider records.
```
tomaka: (that seems pretty important)

protocols/kad/src/behaviour.rs
```diff
@@ -836,229 +922,232 @@ where
...
-        key: PeerId,
-        /// List of peers ordered from closest to furthest away.
-        closer_peers: Vec<PeerId>,
-    },
+/// The result of [`Kademlia::get_record`].
+pub type GetRecordResult = Result<GetRecordOk, GetRecordError>;
+
+/// The successful result of [`Kademlia::get_record`].
+#[derive(Debug, Clone)]
+pub struct GetRecordOk {
```
tomaka: Again, this is too much typing in my opinion.
Having to write things such as `GetRecordResult(Ok(GetRecordOk { records }))` is quite annoying to figure out when you're not the one who wrote the Kademlia code.

protocols/kad/src/behaviour.rs
```diff
@@ -836,229 +922,232 @@ where
...
+    LocalStorageError(RecordStorageError)
+}
+
+/// The result of [`Kademlia::bootstrap`].
+pub type BootstrapResult = Result<BootstrapOk, BootstrapError>;
+
+/// The successful result of [`Kademlia::bootstrap`].
+#[derive(Debug, Clone)]
+pub struct BootstrapOk {
+    pub peer: PeerId
```
tomaka: What is that value?

protocols/kad/src/behaviour.rs
```diff
@@ -836,229 +922,232 @@ where
...
+
+/// The successful result of [`Kademlia::put_record`].
+#[derive(Debug, Clone)]
+pub struct PutRecordOk {
+    pub key: Multihash
+}
+
+/// The error result of [`Kademlia::put_record`].
+#[derive(Debug, Clone)]
+pub enum PutRecordError {
```
tomaka: A method `into_key()` would probably be appreciated by the user who wants to try again.

examples/ipfs-kad.rs
```diff
@@ -45,42 +51,61 @@ fn main() {
...
-	    behaviour.add_address(&"QmSoLV4Bbm51jM9C4gDYZQ9Cy3U6aXMJDAbzgu2fzaDs64".parse().unwrap(), "/ip4/104.236.76.40/tcp/4001".parse().unwrap());
-	    behaviour.add_address(&"QmSoLer265NRgSp2LA3dPaeykiS1J6DifTC88f5uVQKNAd".parse().unwrap(), "/ip4/178.62.158.247/tcp/4001".parse().unwrap());
-	    behaviour.add_address(&"QmSoLPppuBtQSGwKDZT2M73ULpjvfd3aZ6ha4oFGL1KrGM".parse().unwrap(), "/ip6/2604:a880:1:20::203:d001/tcp/4001".parse().unwrap());
-	    behaviour.add_address(&"QmSoLSafTMBsPKadTEgaXctDQVcqN88CNLHXMkTNwMKPnu".parse().unwrap(), "/ip6/2400:6180:0:d0::151:6001/tcp/4001".parse().unwrap());
-	    behaviour.add_address(&"QmSoLV4Bbm51jM9C4gDYZQ9Cy3U6aXMJDAbzgu2fzaDs64".parse().unwrap(), "/ip6/2604:a880:800:10::4a:5001/tcp/4001".parse().unwrap());
-        behaviour.add_address(&"QmSoLer265NRgSp2LA3dPaeykiS1J6DifTC88f5uVQKNAd".parse().unwrap(), "/ip6/2a03:b0c0:0:1010::23:1001/tcp/4001".parse().unwrap());
-        libp2p::core::Swarm::new(transport, behaviour, local_peer_id)
+        /*behaviour.add_address(&"QmNnooDu7bfjPFoTZYxMNLWUQJyrVwtbZg5gBMjTezGAJN".parse().unwrap(), "/dnsaddr/bootstrap.libp2p.io".parse().unwrap());
```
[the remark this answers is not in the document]
romanb: Because they unnecessarily and permanently slow down running the example, which already runs into plenty of timeouts as it is.

protocols/kad/src/kbucket/key.rs
```diff
@@ -35,12 +35,40 @@ use sha2::{Digest, Sha256, digest::generic_array::{GenericArray, typenum::U32}};
...
-    hash: GenericArray<u8, U32>,
+    bytes: KeyBytes,
 }
 
-impl<T> PartialEq for Key<T> {
-    fn eq(&self, other: &Key<T>) -> bool {
-        self.hash == other.hash
+/// The raw bytes of a key in the DHT keyspace.
+#[derive(PartialEq, Eq, Clone, Debug)]
+pub struct KeyBytes(GenericArray<u8, U32>);
```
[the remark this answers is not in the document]
romanb: I understand how the motivation may be a bit weak in this PR alone, but this is actually expanded upon in https://github.com/libp2p/rust-libp2p/pull/1189. There, `KBucketsTable`, `KBucket`, `Entry`s etc. are generalised from working with `Key<T>` to any `TKey: AsRef<KeyBytes>`. That is, it allows e.g. to have a `KBucketsTable` where the key type is just `KeyBytes` or a custom type, as long as it provides `AsRef<KeyBytes>`. This is then also made use of in tests to construct random routing tables with `KeyBytes` as the key type, where keys can be deterministically generated for a certain bucket (via a random distance falling into the bucket) with the help of the function `KeyBytes::for_distance` added in this PR.

protocols/kad/src/lib.rs
```diff
@@ -37,4 +33,39 @@ mod addresses;
 mod behaviour;
 mod protobuf_structs;
 mod query;
-mod write;
+
+pub use behaviour::*;
```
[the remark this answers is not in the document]
romanb: Ok.

src/lib.rs
```diff
@@ -254,7 +254,7 @@ pub fn build_tcp_ws_secio_mplex_yamux(keypair: identity::Keypair)
             core::upgrade::apply(output.stream, upgrade, endpoint)
                 .map(|(id, muxer)| (id, core::muxing::StreamMuxerBox::new(muxer)))
         })
-        .with_timeout(Duration::from_secs(20))
+        .with_timeout(Duration::from_secs(10))
```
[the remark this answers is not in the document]
romanb: I admit that was only done to somewhat shorten the running times of the ipfs-kad example, which can run into _a lot_ of unresponsive nodes and connection timeouts during its iterative query. I can revert it if 20 seems better than 10, I didn't think this difference is meaningful for the development transport.

protocols/kad/src/query/peers/closest.rs
romanb: This module is agnostic to how time progresses (and at what granularity), that is entirely left up the code using it (which makes for side-effect-free code and easy testing). Note that these modules are private to the crate and this module in particular is really just the previous `query.rs` with some renaming, i.e. this isn't really new code.

protocols/kad/src/behaviour.rs
romanb: Running queries take up resources and all resource use should be bounded in a controlled fashion. While every query should "end at some point" if everything works correctly, that "at some point" can be a very, _very_ long time. If a user is willing to let queries run that long he can still do so by configuring a very large timeout. Furthermore, that a query "ends at some point" without a timeout depends on quite a few implementation details, some of which out of the control of `libp2p-kad`, namely that all the callbacks for connected / disconnected / dialing failed / etc. are all always invoked correctly for all peers participating in a query without a single mistake, otherwise there is already a memory leak that may go unnoticed for a long time. Explicit, hard timeouts put a bound here and make such issues more visible through the produced timeout errors.

In this manner, explicit timeouts provide some control over how many queries may actually be still running, e.g. if you want to issue a random query every minute say, as in substrate, having a query timeout of 1 minute ensures that only one such query is running at a time (or at most 2 for a timeout of 2 minutes, at most 3 for a timeout of 3 minutes or whatever is desired).

protocols/kad/src/behaviour.rs
romanb: The point is to have a more expressive and convenient (or as Rust folk like to say, _ergonomic_) API for the most-used quorums, and these are at least `One` (for read) and `All` (for write) which yields standard Kademlia semantics. I would be fine with removing `Two` and `Three` but would like to keep `Majority` as well.

> ... since the user could calculate the actual value themselves.

Sure, but it is inconvenient. The quorum should always be evaluated against the configured replication factor (it doesn't have meaningful semantics otherwise), which is done for the user in `put_record` and `get_record`. Instead of having to construct `NonZeroUsize`s, passing a value like `Quorum::One` is surely nicer and communicates more information.

protocols/kad/src/behaviour.rs
romanb: It's not new though, it was never done, I just added a TODO here. It is done as part of https://github.com/libp2p/rust-libp2p/pull/1189.

protocols/kad/src/behaviour.rs
romanb: While you can, you would usually not mention the inner `Ok` struct type in the pattern match though, i.e.

`GetRecordResult(Ok(result)) => { ... result.records ... }`

which surely isn't much worse than

`GetRecordResult(Ok(records)) => { ... records ... }`

or 

`GetRecordResult { records } => { ... records ... }`

the last of which does not even permit failure and is thus ruled out from my perspective. I did consider a single, unified error enum emitted by a single error event, but the only advantage I could see of that would be less effort on the implementors side (me) for loss of precision (i.e. any operation can supposedly yield any error, which isn't true) and there would then also need to be a way to identify what type of query actually produced the error. So I gladly took on the minor additional keyboard typing effort. 

If you simply suggest to drop `XyzOk` structs that (right now) just contain a single field, that seems possible but I like the consistency and extensibility of this API, where all DHT operations produce a `Result` with a struct for `Ok` and an enum for `Error`, making it easy to add new fields or variants specific to each type of query.

protocols/kad/src/behaviour.rs
romanb: The peer ID that is the target of the bootstrapping query - that is, either the local peer ID (first query) or a random peer ID for a specific bucket (follow-up queries). I can add a doc-comment.

src/lib.rs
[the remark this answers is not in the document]
tomaka: That method is supposed to be for production, while `build_development_transport` is for development. Right now the latter wraps around the former, but that's an implementation detail.

protocols/kad/src/query/peers/closest.rs
tomaka: > that is entirely left up the code using it

Which raises the question: how does the code using it know when the timeout has expired and that we should poll the `ClosestPeersIter` again?

protocols/kad/src/behaviour.rs
tomaka: To me it sounded obvious that `BootstrapOk` was only produced once at the very end of the bootstrapping process.

tomaka: Which, to me, shows that more documentation is needed here.

src/lib.rs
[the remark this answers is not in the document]
romanb: Oh, right. Anyways, I will revert that change then.

protocols/kad/src/behaviour.rs
romanb: How about the [existing documentation of `Kademlia::bootstrap`](https://github.com/libp2p/rust-libp2p/blob/ed07e679c8af9b2c298b3e4afcd066fb22e026c6/protocols/kad/src/behaviour.rs#L276-L290)? I think it does explain it in sufficient detail, but let me know what you think is missing.

protocols/kad/src/query/peers/closest.rs
romanb: If I understand the question correctly, it does not need to know / does not care. The code using it only cares about driving a query to completion by sending requests to all peers yielded by the query and reporting the results back. Concretely, `Kademlia::poll()` polls the `QueryPool` to obtain peers to send requests to. The QueryPool in turn looks through its queries, advancing the peer "iterators" to get the next peers to send requests to. Whenever in an execution of `Kademlia:poll` all queries in the pool yield no more peers (e.g. parallelism is at capacity), the pool itself yields `AtCapacity` or `Waiting(None)` and `Kademlia::poll` may return `NotReady` at this point. There is I/O in progress, all the necessary task wakeup is driven by the I/O. The QueryPool and the queries inside with their peer selection strategies just implement the state machines for advancing the queries. `Kademlia::poll` passes down the current `Instant` whenever it asks the `QueryPool` for new stuff to do and in the case of the `ClosestPeersIter`, this influences when the iterator terminates, as peers for whom a result was not delivered back within the `peer_timeout` are merely not being waited for when evaluating the termination conditions (but delayed results are still accepted, if they come in before the query is done).

I hope that makes sense.

tomaka: That doesn't answer my question that I had in mind :sweat_smile: but I'm going to trust you on this one.

protocols/kad/src/behaviour.rs
romanb: Hm, I'm not sure this is an improvement. That kind of documentation reminds me of those on `start_send` methods on futures and streams, where it is always required by the user to call another method (e.g. `poll_complete`) to actually finish the operation. Of course all Kademlia operations are driven by `Kademlia::poll()` but that does typically not need to be called by the user, as it is driven through a Swarm. In other words, `Kademlia::put_record` et al are supposed to _do_ these operation, from start to finish, not just to start them. The fact that the result is delivered asynchronously via `KademliaEvent`s (and thus that the operation is itself asynchronous) is (and needs to be) documented, of course.

Put differently, when I read documentation of the form "starts the process of ..." I think "ok, and what do I have to do to finish it?".

tomaka:
It looks like you don't want to put the mention in the docs that `get_record` and `put_record` only start the process instead of doing it instantly!

Other than that, looks good to me.

romanb:
I'm still unsure about the best wording in the documentation, but we can always change it later.

Thanks for the review!

PR #1178, opened 2019-06-18
Fix #1135

### LINKED ISSUE #1135: Add a time-to-live to external addresses

tomaka:
Fix #1135

Instead of just a list of addresses with a score, we now keep a list of addresses with score plus a queue of reports.

Every time a report is added, we pop the oldest record and decrease the score of that address.

In other words, the score of an address is now the number of times it's been reported in the last 200 reports.

Because we limit to 30 addresses, it is possible that this is not actually in sync. However the consequence of this is benign.

### DIFF
--- a/core/src/swarm/registry.rs
+++ b/core/src/swarm/registry.rs
@@ -20,7 +20,7 @@
 
 use crate::Multiaddr;
 use smallvec::SmallVec;
-use std::num::NonZeroUsize;
+use std::{collections::VecDeque, num::NonZeroUsize};
 
 /// Hold a ranked collection of [`Multiaddr`] values.
 ///
@@ -32,7 +32,10 @@ pub struct Addresses {
     /// Max. length of `registry`.
     limit: NonZeroUsize,
     /// The ranked sequence of addresses.
-    registry: SmallVec<[Record; 8]>
+    registry: SmallVec<[Record; 8]>,
+    /// Queue of last reports. Every new report is added to the queue. If the queue reaches its
+    /// capacity, we also pop the first element.
+    reports: VecDeque<Multiaddr>,
 }
 
 // An address record associates a score to a Multiaddr.
@@ -53,7 +56,8 @@ impl Addresses {
     pub fn new(limit: NonZeroUsize) -> Self {
         Addresses {
             limit,
-            registry: SmallVec::new()
+            registry: SmallVec::new(),
+            reports: VecDeque::with_capacity(200),
         }
     }
 
@@ -62,17 +66,41 @@ impl Addresses {
     /// Adding an existing address is interpreted as additional
     /// confirmation and thus increases its score.
     pub fn add(&mut self, a: Multiaddr) {
+        debug_assert_ne!(self.reports.capacity(), 0);
+
+        let oldest = if self.reports.len() == self.reports.capacity() {
+            self.reports.pop_front()
+        } else {
+            None
+        };
+
+        if let Some(oldest) = oldest {
+            if let Some(in_registry) = self.registry.iter_mut().find(|r| r.addr == oldest) {
+                in_registry.score = in_registry.score.saturating_sub(1);
+                isort(&mut self.registry);
+            }
+        }
+
+        // Remove addresses that have a score of 0.
+        while self.registry.last_mut().map(|e| e.score == 0).unwrap_or(false) {
+            self.registry.pop();
+        }
+
+        self.reports.push_back(a.clone());
+
         for r in &mut self.registry {
             if &r.addr == &a {
                 r.score = r.score.saturating_add(1);
                 isort(&mut self.registry);
-                return ()
+                return
             }
         }
+
         if self.registry.len() == self.limit.get() {
             self.registry.pop();
         }
-        let r = Record { score: 0, addr: a };
+
+        let r = Record { score: 1, addr: a };
         self.registry.push(r)
     }
 
@@ -159,7 +187,7 @@ fn isort(xs: &mut [Record]) {
 mod tests {
     use crate::Multiaddr;
     use quickcheck::QuickCheck;
-    use super::{isort, Record};
+    use super::{isort, Addresses, Record};
 
     #[test]
     fn isort_sorts() {
@@ -177,5 +205,24 @@ mod tests {
             true
         }
         QuickCheck::new().quickcheck(property as fn(Vec<u32>) -> bool)
+    }
+
+    #[test]
+    fn old_reports_disappear() {
+        let mut addresses = Addresses::default();
+
+        // Add an address a single time.
+        let single: Multiaddr = "/tcp/2108".parse().unwrap();
+        addresses.add(single.clone());
+        assert!(addresses.iter().find(|a| **a == single).is_some());
+
+        // Then fill `addresses` with random stuff.
+        let other: Multiaddr = "/tcp/120".parse().unwrap();
+        for _ in 0 .. 2000 {
+            addresses.add(other.clone());
+        }
+
+        // Check that `single` disappeared from the list.
+        assert!(addresses.iter().find(|a| **a == single).is_none());
     }
 }
--- a/core/src/swarm/registry.rs
+++ b/core/src/swarm/registry.rs
@@ -66,7 +66,6 @@ impl Addresses {
     /// Adding an existing address is interpreted as additional
     /// confirmation and thus increases its score.
     pub fn add(&mut self, a: Multiaddr) {
-        debug_assert_ne!(self.reports.capacity(), 0);
 
         let oldest = if self.reports.len() == self.limit.get() {
             self.reports.pop_front()
--- a/core/src/swarm/registry.rs
+++ b/core/src/swarm/registry.rs
@@ -29,10 +29,10 @@ use std::{collections::VecDeque, num::NonZeroUsize};
 /// score will be dropped first.
 #[derive(Debug, Clone)]
 pub struct Addresses {
-    /// Max. length of `registry`.
-    limit: NonZeroUsize,
     /// The ranked sequence of addresses.
     registry: SmallVec<[Record; 8]>,
+    /// Number of historical reports. Similar to `reports.capacity()`.
+    limit: NonZeroUsize,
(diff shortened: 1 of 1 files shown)


core/src/swarm/registry.rs
```diff
@@ -53,7 +56,8 @@ impl Addresses {
     pub fn new(limit: NonZeroUsize) -> Self {
         Addresses {
             limit,
-            registry: SmallVec::new()
+            registry: SmallVec::new(),
+            reports: VecDeque::with_capacity(200),
```
twittner: Note that 200 is only a lower bound and I guess `VecDeque` will allocate the nearest power of 2 bytes.

tomaka:
> To some extent this still exhibits the problem described in #1135, namely that with rotating addresses and a report number smaller than the report capacity there will be a certain number of outdated addresses in front of it. If an address is rotated every n reports it will only be able to move up to position capacity / n. So for n = 60 and capacity = 255 it will start at the end and move up to position 4 before becoming outdated itself which means there will be at best still 3 (outdated) addresses in front of it.

I don't understand that. At each report N, we're also decreasing the score of the address of the report N-200.

In other words the score of each address should be exactly the number of times it's been reported in the last 200 reports.

twittner:
Consider the example of number of reports `n` = 60 and capacity `c` = 240. Unless `c` reports were made, `c / n = 4` addresses (`a`,`b`,`c`,`d`), each with score `n`, are stored in the registry. Now the 5th address `e` is added `n` times. It starts at the end and with every report, only the score of `a` will be decremented as it was initially reported `n` times. This means that after `n / 2` times `a` will have moved past `e` to the last position and `e` is at position 4. The other addresses `b`, `c` and `d` are still in front of `e`. This pattern repeats for every new address.

tomaka:
Ah I see what you mean now.
I'm kind of assuming that c reports will have happened before we rotate the address.

With c = 200 and 25 connections and 2 minutes between identifications, that's between 14 and 16 minutes.

Vurich:
> Consider the example of number of reports `n` = 60 and capacity `c` = 240. Unless `c` reports were made, `c / n = 4` addresses (`a`,`b`,`c`,`d`), each with score `n`, are stored in the registry. Now the 5th address `e` is added `n` times. It starts at the end and with every report, only the score of `a` will be decremented as it was initially reported `n` times. This means that after `n / 2` times `a` will have moved past `e` to the last position and `e` is at position 4. The other addresses `b`, `c` and `d` are still in front of `e`. This pattern repeats for every new address.

Is that a problem? We remove the oldest reports first, so if the pattern is like `aaaabbbbccccdddd` then `a` hasn't been reported for a long time and so deserves to be decremented first.

tomaka:
> Is that a problem? We remove the oldest reports first, so if the pattern is like aaaabbbbccccdddd then a hasn't been reported for a long time and so deserves to be decremented first.

The problem is more that with a pattern like `aaaabbbbccccdddd`, `d` will always be below `b` and `c` in the list even though it's now the "main" address.
Assuming that addresses rotate at a fixed rate, the moment when `d` goes on top (equal to `b` and `c`) is the moment when we switch to `e`.

Vurich:
> > Is that a problem? We remove the oldest reports first, so if the pattern is like aaaabbbbccccdddd then a hasn't been reported for a long time and so deserves to be decremented first.
> 
> The problem is more that with a pattern like `aaaabbbbccccdddd`, `d` will always be below `b` and `c` in the list even though it's now the "main" address.
> Assuming that addresses rotate at a fixed rate, the moment when `d` goes on top (equal to `b` and `c`) is the moment when we switch to `e`.

core/src/swarm/registry.rs
```diff
@@ -62,17 +66,41 @@ impl Addresses {
     /// Adding an existing address is interpreted as additional
     /// confirmation and thus increases its score.
     pub fn add(&mut self, a: Multiaddr) {
+        debug_assert_ne!(self.reports.capacity(), 0);
+
+        let oldest = if self.reports.len() == self.reports.capacity() {
```
twittner: I suggest to add a separate field instead of using the capacity because the value passed to `VecDeque::with_capacity` may not be the actual capacity in use which makes it harder to understand the program. A field `max_history: usize` would be more straightforward.

core/src/swarm/registry.rs
twittner: Instead of hard-coding the max. report length which has implications w.r.t. some use cases (as discussed) I suggest to take it as an additional parameter and use the 200 in the `Default` impl.

core/src/swarm/registry.rs
```diff
@@ -62,17 +66,41 @@ impl Addresses {
...
+        };
+
+        if let Some(oldest) = oldest {
+            if let Some(in_registry) = self.registry.iter_mut().find(|r| r.addr == oldest) {
+                in_registry.score = in_registry.score.saturating_sub(1);
+                isort(&mut self.registry);
+            }
+        }
+
+        // Remove addresses that have a score of 0.
```
twittner: I suggest not to remove records with score 0 here. The important goal is that outdated addresses move to the end quickly where they can still be left lingering before being pushed out by newer addresses. There might still be a chance that they are reachable.

core/src/swarm/registry.rs
```diff
@@ -178,4 +206,23 @@ mod tests {
...
+
+        // Then fill `addresses` with random stuff.
+        let other: Multiaddr = "/tcp/120".parse().unwrap();
+        for _ in 0 .. 2000 {
+            addresses.add(other.clone());
+        }
+
+        // Check that `single` disappeared from the list.
+        assert!(addresses.iter().find(|a| **a == single).is_none());
+    }
```
twittner: Since the important property is that each score reflects the number of reports in the last `n` reports, I would replace this test with a corresponding property test:

```rust
    #[test]
    fn record_score_equals_last_n_reports() {
        use multiaddr::Protocol;
        use quickcheck::{Arbitrary, Gen};
        use rand::Rng;
        use std::num::NonZeroUsize;

        #[derive(PartialEq, Eq, Clone, Hash, Debug)]
        struct Ma(Multiaddr);

        impl Arbitrary for Ma {
            fn arbitrary<G: Gen>(g: &mut G) -> Self {
                Ma(Protocol::Tcp(g.gen::<u16>() % 16).into())
            }
        }

        fn property(xs: Vec<Ma>, n: u8) -> bool {
            let n = std::cmp::max(n, 1);
            let mut addresses = Addresses::new(NonZeroUsize::new(30).unwrap(), usize::from(n));
            for Ma(a) in &xs {
                addresses.add(a.clone())
            }
            for r in &addresses.registry {
                let count = xs.iter()
                    .rev()
                    .take(usize::from(n))
                    .filter(|Ma(x)| x == &r.addr)
                    .count();
                if r.score as usize != count {
                    return false
                }
            }
            true
        }

        QuickCheck::new().quickcheck(property as fn(Vec<Ma>, u8) -> bool)
    }
```

core/src/swarm/registry.rs
tomaka: Addresses with a score of 0 haven't been reported in the last 200 reports.
Removing addresses only when we reach the limit of 30 addresses assumes that your address actually rotates often enough, which might not necessarily be true.

If your address rotates only once every three months, you don't want to wait for 900 months in order for the oldest address to disappear.

twittner: The current address would take precedence long before that so the fact that there is this outdated address after the current one(s) is of no concern and with a limited report history a seemingly unreported address may after all still be useful.

tomaka: We're still reporting to everyone an extra address which we know is useless.

twittner: The code has no notion of time beyond the limited historical record so it can only be *assumed* that an address with score 0 is "useless".

tomaka: I don't really get the point. The fact that we're limiting ourselves to 30 addresses is also extremely arbitrary, and we can't be sure that the 31st address is not actually valid.

There's no perfect solution to that, and I'm just trying to do heuristics to solve the problem of rotating IPs and bad/malicious reports.

twittner: My point is that because we already have this arbitrary limit we do not need another one. It is there as a safeguard to ensure the number of addresses does not grow unbounded and certainly does not imply an address is useless as it will remove addresses even if their score is > 0. The report history on the other hand is an orthogonal mechanism to determine the order of addresses based on historic information but it can not judge the usefulness of addresses beyond its historical record.

tomaka: This PR kind of suggests to entirely change the mechanism so that the list of addresses **is** the historical reports of addresses. Having this limit of 30 addresses would then become a "secondary check" and could maybe be put inside the logic of `AddressesIter` instead.

twittner: > [...] the list of addresses **is** the historical reports of addresses [...]

In that case I would remove the registry limit altogether. The max. history length already determines a max. registry limit if we remove records with score 0. Client code using the iterator can then just `take` an arbitrary-length prefix of that (e.g. 30).

tomaka:
I went for removing the limit of 30 altogether, as it doesn't really make sense to me indeed.

core/src/swarm/registry.rs
```diff
@@ -62,17 +66,37 @@ impl Addresses {
     /// Adding an existing address is interpreted as additional
     /// confirmation and thus increases its score.
     pub fn add(&mut self, a: Multiaddr) {
-        for r in &mut self.registry {
-            if &r.addr == &a {
-                r.score = r.score.saturating_add(1);
+        debug_assert_ne!(self.reports.capacity(), 0);
```
twittner: This is no longer needed.

tomaka:
```suggestion
```

PR #1180, opened 2019-06-22
Make `<SubstreamRef as AsyncWrite>::shutdown` imply flush

sinkuu:
Documentation of `tokio::io::AsyncWrite::shutdown` says "Invocation of a shutdown implies an invocation of flush." But `SubstreamRef`'s impl of `AsyncWrite` did not.

This caused `Floodsub` (which uses `WriteOne` that justifiably assumes this invariant of `AsyncWrite` and `shutdown`s socket after writing) to fail to communicate via mplex.

### DIFF
--- a/core/src/muxing.rs
+++ b/core/src/muxing.rs
@@ -442,11 +442,11 @@ where
         loop {
             match self.shutdown_state {
                 ShutdownState::Shutdown => {
-                    let _ = try_ready!(self.muxer.shutdown_substream(s).map_err(|e| e.into()));
+                    try_ready!(self.muxer.shutdown_substream(s).map_err(|e| e.into()));
                     self.shutdown_state = ShutdownState::Flush;
                 }
                 ShutdownState::Flush => {
-                    let _ = try_ready!(self.muxer.flush_substream(s).map_err(|e| e.into()));
+                    try_ready!(self.muxer.flush_substream(s).map_err(|e| e.into()));
                     self.shutdown_state = ShutdownState::Done;
                 }
                 ShutdownState::Done => {
--- a/core/src/muxing.rs
+++ b/core/src/muxing.rs
@@ -338,6 +338,7 @@ where
     SubstreamRef {
         muxer,
         substream: Some(substream),
+        shutdown_state: ShutdownState::Shutdown,
     }
 }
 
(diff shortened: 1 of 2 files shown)


tomaka:
Oh good catch! I remember encountering some issues which may be caused by this.

core/src/muxing.rs
```diff
@@ -431,8 +439,33 @@ where
     #[inline]
     fn shutdown(&mut self) -> Poll<(), io::Error> {
         let s = self.substream.as_mut().expect("substream was empty");
-        self.muxer.shutdown_substream(s).map_err(|e| e.into())?;
-        Ok(Async::Ready(()))
+        loop {
+            match self.shutdown_state {
+                ShutdownState::Shutdown => {
+                    match self.muxer.shutdown_substream(s).map_err(|e| e.into())? {
```
tomaka: Could you do something like this?

```rust
let _ = try_ready!(self.muxer.shutdown_substream(s));
self.shutdown_state = ShutdownState::Flush;
```

(and same below)

It would make that code much more concise.

sinkuu: Done.

core/src/muxing.rs
```diff
@@ -431,8 +439,21 @@ where
     #[inline]
     fn shutdown(&mut self) -> Poll<(), io::Error> {
         let s = self.substream.as_mut().expect("substream was empty");
-        self.muxer.shutdown_substream(s).map_err(|e| e.into())?;
-        Ok(Async::Ready(()))
+        loop {
+            match self.shutdown_state {
+                ShutdownState::Shutdown => {
+                    let _ = try_ready!(self.muxer.shutdown_substream(s).map_err(|e| e.into()));
```
twittner:
```suggestion
                    try_ready!(self.muxer.shutdown_substream(s).map_err(|e| e.into()));
```

core/src/muxing.rs
```diff
@@ -431,8 +439,21 @@ where
...
+                    self.shutdown_state = ShutdownState::Flush;
+                }
+                ShutdownState::Flush => {
+                    let _ = try_ready!(self.muxer.flush_substream(s).map_err(|e| e.into()));
```
twittner:
```suggestion
                    try_ready!(self.muxer.flush_substream(s).map_err(|e| e.into()));
```

PR #1183, opened 2019-06-24
libp2p_tcp: Ignore sockets without local/remote addr.

### LINKED ISSUE #1182: `nc -z` effectively kills Swarm
Probing running Swarm with `nc -z` breaks it.
It is easily reproducible on v0.9.1 with `chat` example:
Terminal windows #1:
```
eric@copoka ~/C/S/rust-libp2p> cargo run --example chat
    Finished dev [unoptimized + debuginfo] target(s) in 1.05s
     Running `target/debug/examples/chat`
Local peer id: PeerId("Qmbe2cZb6gcQcCiHM7C5ferWK6zDTAXzTeu6JbqQn8ad3r")
Listening on "/ip4/127.0.0.1/tcp/63303"
[2019-06-24T08:31:36Z ERROR libp2p_tcp] Failed to get peer address: Os { code: 22, kind: InvalidInput, message: "Invalid argument" }
```
Terminal #2:
```
eric@copoka ~/C/S/rust-libp2p> nc -z localhost 63303
eric@copoka ~/C/S/rust-libp2p> telnet localhost 63303
Trying ::1...
telnet: connect to address ::1: Connection refused
Trying 127.0.0.1...
telnet: connect to address 127.0.0.1: Connection refused
telnet: Unable to connect to remote host
eric@copoka ~/C/S/rust-libp2p> 
```

After the first netcat connect, Swarm stops accepting any more connections.

twittner:
The transport should be able to continue processing other connections. An error to determine a socket's address is not a fatal condition but may happen when a connection is immediately reset after being established. By the time the programme asks for the remote address, the socket may already be gone.

Fixes #1182.

### DIFF
--- a/transports/tcp/src/lib.rs
+++ b/transports/tcp/src/lib.rs
@@ -50,7 +50,7 @@ use libp2p_core::{
     multiaddr::{Protocol, Multiaddr},
     transport::{ListenerEvent, TransportError}
 };
-use log::{debug, error, trace};
+use log::{debug, trace};
 use std::{
     collections::VecDeque,
     fmt,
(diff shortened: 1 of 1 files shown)


tomaka:
What about returning a `Async::Ready(Some(...))` with some event indicating that an "unknown" error happened on a socket?

I'm a bit conflicted, because in practice this error is totally useless.

twittner:
I do not think an error event would be useful. In order for clients to react in a meaningful way we would need to add a callback to `NetworkBehaviour` and then the question is what can they actually do? Without an address they have no means to identify the remote. They could decide to stop the listener if they receive too many errors, but that would also require extending the set of possible actions to actually enable them to do so. At this stage I think log-and-continue is the right approach.

PR #1187, opened 2019-06-26
multihash: Use `Bytes` instead of `Vec<u8>` internally.

twittner:
To improve the efficiency of cloning multi-hashes (e.g. as the representation of `PeerId`s), this PR replaces the `Vec<u8>` representation with `Bytes`. The API is largely kept backwards compatible and does not leak the representation type.

### DIFF
--- a/misc/multihash/src/hashes.rs
+++ b/misc/multihash/src/hashes.rs
@@ -37,8 +37,8 @@ pub enum Hash {
 
 impl Hash {
     /// Get the corresponding hash code.
-    pub fn code(&self) -> u16 {
-        match *self {
+    pub fn code(self) -> u16 {
+        match self {
             Hash::SHA1 => 0x11,
             Hash::SHA2256 => 0x12,
             Hash::SHA2512 => 0x13,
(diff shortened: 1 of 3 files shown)


misc/multihash/src/hashes.rs
```diff
@@ -37,8 +37,8 @@ pub enum Hash {
 
 impl Hash {
     /// Get the corresponding hash code.
-    pub fn code(&self) -> u16 {
-        match *self {
+    pub fn code(self) -> u16 {
```
tomaka: I think this is the only breaking change and it would force us to do a "major" bump, so it feels a bit stupid.

misc/multihash/src/hashes.rs
```diff
@@ -58,8 +58,8 @@ impl Hash {
     }
 
     /// Get the hash length in bytes.
-    pub fn size(&self) -> u8 {
-        match *self {
+    pub fn size(self) -> u8 {
```
tomaka: Same

Issue #1112, opened 2019-05-10
Consider extracting the Swarm/NetworkBehaviour/ProtocolsHandler to a different crate


tomaka:
Right now `libp2p-core` contains two different APIs: the one in the `nodes` module, and the one in the `swarm` and `protocols_handler` module. The latter is built on top of the former.

Maybe it is better if the content of `nodes` (plus `transport`, `upgrade`, `identity`, etc.) were in their own crate, and the content of `swarm` and `protocols_handler` were in a different one on top.
The higher-level crate would re-export the types from the lower-level crate.

romanb:
Can this be considered done?

PR #1188, opened 2019-06-27
Move swarm and protocols handler into swarm crate.

twittner:
In the context of #1112 this PR extracts the `swarm` and `protocols_handler` modules into a separate crate named `libp2p-swarm`. For now I chose not to re-export `libp2p-core` items from `libp2p-swarm` so that it is easier to see what is being pulled in from where. Also I am not sure if we really should re-export core items.


romanb:
FWIW I like it, but there is still a large chunk of crate-level documentation in `core/src/lib.rs` that talks about `Swarm`s and `NetworkBehaviour`s. I guess most of it could just be moved to the new `libp2p-swarm` crate and `libp2p-core` would need to get some new, top-level summary. Furthermore, for the cut to be clean, I think `libp2p_core` should then be free of the notion and any mentions of a "Swarm", which would also mean finding a better name for `RawSwarm`. That could certainly be done afterwards though.

Issue #1089, opened 2019-04-27
Kademlia::add_providing reaches unimplemented!()


rnbguy:
I get unimplemented error while using `add_providing` in Kademlia swarm, because of this.

https://github.com/libp2p/rust-libp2p/blob/ca58f8029c70db16774c988f409703f0191609d8/protocols/kad/src/behaviour.rs#L177-L180

It seems the mentioned TODO should work.

https://github.com/libp2p/rust-libp2p/blob/ca58f8029c70db16774c988f409703f0191609d8/protocols/kad/src/behaviour.rs#L127-L130

rnbguy:
Also, is there any reason why `add_providing` requires `key` to be `PeerId`, not `Multihash`? It does `key.clone().into()` or `key.into()` inside the function anyway.

https://github.com/libp2p/rust-libp2p/blob/ca58f8029c70db16774c988f409703f0191609d8/protocols/kad/src/behaviour.rs#L296-L298

tomaka:
`add_providing` is one of these features that is half implemented and not tested. While the code exists, you will run into issues.

As for `PeerId` vs `Multihash` for `add_providing`, this indeed looks like a mistake.

cc @elferdo

elferdo:
> Also, is there any reason why `add_providing` requires `key` to be `PeerId`, not `Multihash`? It does `key.clone().into()` or `key.into()` inside the function anyway.
> 
> https://github.com/libp2p/rust-libp2p/blob/ca58f8029c70db16774c988f409703f0191609d8/protocols/kad/src/behaviour.rs#L296-L298

This is indeed a mistake and should be addressed in https://github.com/libp2p/rust-libp2p/pull/1095

Issue #1159, opened 2019-06-04
[kad] RecordsStore trait should be async


tomaka:
Getting or writing a record should be able to not happen immediately.

montekki:
We should probably create a to-do list (meta) issue tracking all needed improvements for records

tomaka:
@montekki Note that you don't have to handle this yourself (unless you need it).

thomaseizinger:
Closed in favor of https://github.com/libp2p/rust-libp2p/issues/3035.

PR #1189, opened 2019-06-30
Kademlia: Somewhat complete the records implementation.

romanb:
Sits on top of https://github.com/libp2p/rust-libp2p/pull/1174 and relates to [libp2p-146] and [libp2p-1089].

The following are the major changes:
    
   * All records expire (by default, configurable).
   * Storage of provider records is also moved to the `RecordStore`, and the `RecordStore` API extended.
   * Background jobs for periodic (re-)replication and (re-)publication of records. Regular (value-)records are subject to re-replication and re-publication as per standard Kademlia. Provider records are only subject to re-publication.
   * For standard Kademlia value lookups (quorum = 1), if found, the record is cached at the closest peer to the key that did not return the value, as per standard Kademlia.
   * Expiration times of regular (value-)records are computed exponentially inversely proportional to the number of nodes between the local node and the closest node known to the key (beyond the k closest), as per standard Kademlia.
    
To that end, the protobuf structure for regular (value-)records is extended with the two fields `ttl` and `publisher`, in order to implement the different semantics of re-replication (by any of the k closest peers to the key, not affecting expiry) and re-publication (by the original publisher, resetting the expiry). This is not done yet in other libp2p Kademlia implementations (see e.g. [libp2p-go-323]). The new protobuf fields have been given somewhat unique identifiers to prevent future collision, should the libp2p spec and protocol be extended.
    
Similarly, periodic re-publication of provider records does not seem to be done yet in other implementations (see e.g. [libp2p-js-98]) but was already drafted in the existing Rust implementation, which I thus continued.
    
[libp2p-146]: https://github.com/libp2p/rust-libp2p/issues/146
[libp2p-1089]: https://github.com/libp2p/rust-libp2p/issues/1089
[libp2p-go-323]: https://github.com/libp2p/go-libp2p-kad-dht/issues/323
[libp2p-js-98]: https://github.com/libp2p/js-libp2p-kad-dht/issues/98


protocols/kad/src/behaviour.rs
```diff
@@ -315,134 +337,206 @@ where
...
-        if let Some(record) = self.records.get(key) {
-            results.push(record.into_owned());
-            if num_results == 1 {
-                self.queued_events.push(NetworkBehaviourAction::GenerateEvent(
-                    KademliaOut::GetValueResult(
-                        GetValueResult::Found { results }
-                )));
-                return;
+    /// The result of this operation is delivered in [`KademliaEvent::GetRecordResult`].
+    pub fn get_record(&mut self, key: &Multihash, quorum: Quorum) {
```
montekki: I think that we shouldn't be handing out to users the `Multihash` types in the api, since it would make a lot of code in substrate depentable on `Multihash` and different places in code can hash things differently and so on. What if the api took bytes and did the hashing?

romanb: Choosing multihashes as record keys was done in your original PR and didn't change here. I think it would be fine (even good) to further generalise over the record key type, but that should be done in another PR - this one is big enough as it is.

> What if the api took bytes and did the hashing?

Not sure what you mean by that. Whatever the key is, it is _always_ hashed into the Kademlia keyspace, see `protocols/kad/kbucket/key.rs`.

protocols/kad/src/query.rs
```diff
@@ -181,17 +185,17 @@ impl<TInner> QueryPool<TInner> {
...
-    pub replication_factor: usize,
+    pub replication_factor: NonZeroUsize,
 }
 
 impl Default for QueryConfig {
     fn default() -> Self {
         QueryConfig {
             timeout: Duration::from_secs(60),
-            replication_factor: K_VALUE
+            replication_factor: NonZeroUsize::new(K_VALUE).expect("K_VALUE > 0")
```
twittner: Maybe `K_VALUE` should eventually be a `NonZeroUsize`?

protocols/kad/src/addresses.rs
```diff
@@ -22,67 +22,77 @@ use libp2p_core::Multiaddr;
...
+    /// Removes the given address from the list.
+    ///
+    /// Returns true if the address was found and removed, false otherwise.
+    /// The last remaining address in the list cannot be remvoved.
+    ///
+    /// An address should only be removed if is determined to be invalid or
+    /// otherwise unreachable.
+    pub fn remove(&mut self, addr: &Multiaddr) -> bool {
+        if self.addrs.len() == 1 {
+            return false
```
twittner: I am not very familiar with the protocol, but AFAICS an address is removed only if the peer can not be reached using it. So instead of keeping a peer with a defunct address, should the peer not be removed? Alternatively maybe adding a comment to `Addresses` which motivates this behaviour would be appreciated.

protocols/kad/src/jobs.rs
```diff
@@ -0,0 +1,363 @@
...
+        }
+        false
+    }
+}
+
+/// The state of a background job run periodically.
+#[derive(Debug)]
+enum PeriodicJobState<T> {
+    Running(T),
+    Waiting(Delay)
```
twittner: Why contains `PeriodicJobState::Waiting` a future instead of a deadline value?

protocols/kad/dht.proto
```diff
@@ -18,6 +18,14 @@ message Record {
 
 	// Time the record was received, set by receiver
 	string timeReceived = 5;
+
+    // The original publisher of the record.
+    // Currently specific to rust-libp2p.
```
tomaka: Well, I'm not a fan of this. :upside_down_face:

romanb: That's ok, neither am I. Unfortunately I only know of one alternative that avoids protocol additions but is a step backwards w.r.t. Kademlia proper:

  1. Removing (re-)replication, leaving only re-publication of records.
  2. The `publisher` field of a record would (for now) only ever be either `None` or `Some(local_id)` if the local node considers itself the publisher. That is, every node only knows about which records it is itself the publisher of, but not who the alleged publisher of other records is.

Albeit a step backwards (from my perspective at least), it wouldn't preclude (re-)adding replication later, if the libp2p "specs" ever accommodate for it (I'm not holding my breath). Would that indeed be your preference or do you even see other options?

protocols/kad/src/addresses.rs
romanb: I generally feel uneasy about removing addresses from peers (in `Kademlia::inject_addr_reach_failure`) because from my understanding many such errors may be transient.  There is a tension, of course. It is also desirable to remove unreachable addresses because they slow down queries, so the problem is when to decide when an address is considered "permanently" unreachable.

I did think about removing the peer from the routing table when the last address is deemed "unreachable", but removal of peers seems even more problematic, because it would mean that temporary loss of connectivity, whether benign or malicious, would "flush" a node's routing table, subverting the important property of the Kademlia protocol of hanging on to existing peers, at least until other peers demonstrate better / longer connectivity, causing least-recently connected nodes to be evicted. Thus I opted for a compromise for now where the last address of a peer is just not removed and hence a peer not removed from the routing table as a result if dialing errors (i.e. just like before with the only difference that I don't allow peers without any addresses in the routing table, hence hanging on to the last address).

Let me know if you have further thoughts / advice on the matter. If my reasoning and choice (for now) sounds good, I can improve the code commentary as suggested.

protocols/kad/src/jobs.rs
romanb: The idea is that when a job is polled (see `PutRecordJob::poll` and `AddProviderJob::poll`) and returns `NotReady` the current task is woken up when the job is ready to run, since there may be nothing else going on otherwise. This polling of the jobs is done in `Kademlia::poll`. Does that make sense or did you have something else in mind with your question?

protocols/kad/src/lib.rs
```diff
@@ -61,9 +63,7 @@ pub use behaviour::{
     GetProvidersError,
 };
 pub use protocol::KadConnectionType;
-pub use record::{RecordStore, RecordStorageError, MemoryRecordStorage};
-
-use std::time::Duration;
+pub use record::*;
```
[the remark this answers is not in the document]
romanb: Ok.

protocols/kad/src/behaviour.rs
romanb: For what it's worth, I have a branch [here](https://github.com/romanb/rust-libp2p/compare/kad-records-providers...romanb:kad-key-bytes?expand=1) that sits on top of this PR and generalises record keys. I may open a PR for that once / if this PR is merged.

protocols/kad/src/addresses.rs
twittner: Yeah,  it is better to only remove peers if a bucket is full and they do not answer a ping.

protocols/kad/src/jobs.rs
```diff
@@ -0,0 +1,406 @@
...
+//!   2. (Re-)replication: Every node storing a replica of a record must
+//!      regularly re-replicate it to the closest nodes to the key in
+//!      order to ensure the record is present at these nodes.
+//!
+//! Re-publishing primarily ensures persistence of the record beyond its
+//! initial TTL, for as long as the publisher stores (or provides) the record,
+//! whilst (re-)replication primarily ensures persistence for the duration
+//! of the TTL in the light of topology changes. Consequently, replication
+//! intervals should be shorter than publication intervals and
+//! publication intervals should be shorter than the TTL.
```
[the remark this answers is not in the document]
romanb: That is just a matter of naming. In order to distinguish the hourly re-publishing done by every node from the 24h re-publishing done by the node that is the "original publisher", I call the former re-replication. The replication of values to a newly joined node can of course be given separate attention, but since that is also covered by the hourly re-republication, I didn't want to over-complicate things.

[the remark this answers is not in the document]
romanb: I guess my choice of naming is partly derived from the choice of naming of the constants for the intervals in [this design specification](http://xlattice.sourceforge.net/components/protocol/kademlia/specs.html#constants).

protocols/kad/src/addresses.rs
tomaka: I don't know if the code takes that into account, but keep in mind that when we receive an incoming, we don't "know" the address of the remote (it is determined through identify), yet we should maybe add it to the buckets already.

protocols/kad/src/handler.rs
```diff
@@ -271,6 +270,8 @@ impl From<ProtocolsHandlerUpgrErr<io::Error>> for KademliaHandlerQueryErr {
 /// Event to send to the handler.
 #[derive(Debug)]
 pub enum KademliaHandlerIn<TUserData> {
+    Reset(KademliaRequestId),
```
tomaka: Needs some documentation saying what that does.

protocols/kad/src/handler.rs
```diff
@@ -472,9 +470,18 @@ where
             .push(SubstreamState::InWaitingMessage(connec_unique_id, protocol));
     }
 
-    #[inline]
     fn inject_event(&mut self, message: KademliaHandlerIn<TUserData>) {
         match message {
+            KademliaHandlerIn::Reset(request_id) => {
```
tomaka: Can't we just wait for the requests to be answered or to time out instead?

protocols/kad/src/record/store.rs
```diff
@@ -0,0 +1,95 @@
...
+    fn get(&'a self, k: &Multihash) -> Option<Cow<Record>>;
+
+    /// Puts a record into the store.
+    fn put(&'a mut self, r: Record) -> Result<()>;
+
+    /// Removes the record with the given key from the store.
+    fn remove(&'a mut self, k: &Multihash);
+
+    /// Gets an iterator over all (value-) records currently stored.
+    fn records(&'a self) -> Self::RecordsIter;
```
tomaka: Is that a good idea? It works for an in-memory records store, but not a records store that stores them on disk.

protocols/kad/dht.proto
tomaka: Let's go with that.

protocols/kad/src/addresses.rs
```diff
@@ -22,67 +22,77 @@ use libp2p_core::Multiaddr;
...
-        self.addrs.is_empty()
-    }
+    /// Removes the given address from the list.
+    ///
+    /// Returns true if the address was found and removed, false otherwise.
+    /// The last remaining address in the list cannot be removed.
+    ///
+    /// An address should only be removed if is determined to be invalid or
+    /// otherwise unreachable.
+    pub fn remove(&mut self, addr: &Multiaddr) -> bool {
```
tomaka: What about returning `-> Result<(), ()>` instead? IMO it's more explicit that `Ok` means that the removal was successful.

protocols/kad/src/behaviour.rs
```diff
@@ -571,75 +787,237 @@ where
...
+                    value: record.value,
+                    request_id,
+                },
+            });
+            return
+        }
+
+        let now = Instant::now();
+
+        // Calculate the expiration exponentially inversely proportional to the
```
tomaka: I didn't even know that was in the paper.

protocols/kad/src/record/store.rs
```diff
@@ -0,0 +1,95 @@
...
+///      libp2p Kademlia specification and realise a "pull-model" for distributed
+///      content. Just like a regular record, a provider record is distributed
+///      to the closest nodes to the key.
+///
+pub trait RecordStore<'a> {
+    type RecordsIter: Iterator<Item = Cow<'a, Record>>;
+    type ProvidedIter: Iterator<Item = Cow<'a, ProviderRecord>>;
+
+    /// Gets a record from the store, given its key.
+    fn get(&'a self, k: &Multihash) -> Option<Cow<Record>>;
```
tomaka: Should eventually be asynchronous, but this would probably be extremely hard to implement as long as we don't have async/await, and not worth it.

protocols/kad/dht.proto
romanb: Do you have a final verdict on the matter? It may be worth mentioning that, despite these new fields, the protocol is still compatible with other libp2p Kademlia implementations, at least that is my intention. The protocol is also easy to (re-)adjust later, I think.

protocols/kad/src/handler.rs
romanb: This is currently used when the node receives a request to store a record, but for some reason (e.g. a storage error) is unable to (or refuses to) store it. We can just let the request time out on the remote, which is what was done before but there was a TODO left for it (by me). I thought it would be preferable to signal errors as quick as possible to not cause queries unnecessary delay in the normal case, and since the protocol itself has no explicit error responses, resetting the stream seemed the only option. Does that make sense?

protocols/kad/src/record/store.rs
romanb: Why do you think it can only be implemented by an in-memory store? That is certainly not my intention, but every record store must provide a means to iterate through all records - that is simply a requirement of the Kademlia protocol.

protocols/kad/src/addresses.rs
romanb: Sounds good to me.

protocols/kad/src/record/store.rs
romanb: I thought the same - besides we have https://github.com/libp2p/rust-libp2p/issues/1159 - but I am myself not sure whether the complications w.r.t. the implementation of the Kademlia behaviour that would come along with such a change would be worth it, especially not before async/await. I think a first step would be to actually allow even `RecordStore` implementations using synchronous I/O without `unwrap`ing, which will require all trait methods to return a `Result` and `store::Error` would need a constructor for `io::Error`. I left that for later as well.

protocols/kad/dht.proto
tomaka: I'm fine with doing that, for the exact reasons you mentioned.

protocols/kad/src/handler.rs
tomaka: Oh I see, I misunderstood the purpose.

protocols/kad/src/record/store.rs
tomaka: My impression is that the minimum requirement would be an iterator over the record keys/ids. My concerns is that the iterator also contains tons of metadata as well as the record's content, which might not be desirable to keep in memory.

This is a minor concern though.

romanb: I see - since I implemented the optimisation from the paper that records are skipped from hourly replication if they have recently been received from another peer (i.e. that effectively only 1 of the `k` peers closest to a key replicates a record hourly in most cases) just iterating over the keys may indeed be beneficial in terms of memory use, albeit requiring separate lookups from the storage for every record that is not skipped. We should certainly keep that in mind as a trade-off to (re)consider.

romanb:
The last feedback I got from him on Friday was that the PR overall looks good. He just had some thoughts around the implementation of the periodic jobs, e.g. for splitting the `PutRecordJob` up to separate re-replication (default hourly) from (re)publishing (default daily). I thought it was neat to combine these into the same job for reasons of efficiency but it may be worthwhile to experiment with a split. I don't want to do that in the context of this PR, however. So I think he is fine with merging and may even have intentionally left approval up to you.

Thank you both for the review and suggestions.

PR #1192, opened 2019-07-02
Adds a store_mut method to kademlia.

montekki:


romanb:
I don't think extending the API in this way is generally a good idea. The `RecordStore` trait should only have those APIs needed by `libp2p-kad`, not more. You could instead allow users to mutably borrow the store passed on construction, i.e. `impl Kademlia { ... pub fn store_mut(&mut self) -> &mut TStore ... }` and have `retain` only on the memory store.

tomaka:
I agree for `retain` in particular. I'm not so sure about making this a general rule, but I indeed prefer a `store_mut()` method.

protocols/kad/src/behaviour.rs
```diff
@@ -277,7 +277,7 @@ where
     }
 
     /// Get a mutable reference to internal record store.
-    pub fn store_mut(&mut self) -> &mut TRecordStorage {
+    pub fn store_mut(&mut self) -> &mut RecordStore {
```
romanb: Why now a trait object, i.e. what's wrong with `&mut TStore`? Also, is it not the case that you then cannot even call `retain` since its not e member of the trait?

montekki: Oh, you are right, good catch thanks

PR #1193, opened 2019-07-03
FloodSub: Add `publish_any`

sectore:
to emit messages even if we are not subscribed to the topic.

Currently a `floodsub` message can be sent only, if the topic has been subscribed before:  https://github.com/libp2p/rust-libp2p/blob/b6378ac526bd66ecb0ce582fcd3942f607bca9d2/protocols/floodsub/src/layer.rs#L185-L188

This PR introduces [`publish_any`](https://github.com/libp2p/rust-libp2p/blob/3c2e5c2e9ab00995b0e84c67ea5a1f0a1223ada4/protocols/floodsub/src/layer.rs#L170) and [`publish_many_any`](https://github.com/libp2p/rust-libp2p/blob/3c2e5c2e9ab00995b0e84c67ea5a1f0a1223ada4/protocols/floodsub/src/layer.rs#L183) to sent messages w/o the need of having subscriptions to specific topics.


protocols/floodsub/src/layer.rs
```diff
@@ -161,17 +161,20 @@ impl<TSubstream> Floodsub<TSubstream> {
...
+    pub fn publish_any(&mut self, topic: impl Into<TopicHash>, data: impl Into<Vec<u8>>) {
+        self.publish_many(iter::once(topic), data, false)
     }
 
     /// Publishes a message with multiple topics to the network.
     ///
-    /// > **Note**: Doesn't do anything if we're not subscribed to any of the topics.
-    pub fn publish_many(&mut self, topic: impl IntoIterator<Item = impl Into<TopicHash>>, data: impl Into<Vec<u8>>) {
+    /// > **Note**: By enabling `check_self_subscriptions`, message will be sent to the network, if we are subscribed to the topic only.
+    pub fn publish_many(&mut self, topic: impl IntoIterator<Item = impl Into<TopicHash>>, data: impl Into<Vec<u8>>, check_self_subscriptions: bool) {
```
tomaka: I think it should be a new method, named something like `publish_many_inner`, instead of a modification of the existing method.

PR #1197, opened 2019-07-09
Add some doc to ProtocolName

tomaka:


core/src/upgrade/mod.rs
```diff
@@ -81,8 +81,40 @@ pub use self::{
 };
 
 /// Types serving as protocol names.
+///
+/// # Why this trait?
```
romanb:
```suggestion
/// # Context
```

core/src/upgrade/mod.rs
```diff
@@ -81,8 +81,40 @@ pub use self::{
...
+///
+/// In situations where we provide a list of protocols that we support, the elements of that list are required to
+/// implement the [`ProtocolName`] trait.
+///
+/// Libp2p will call the [`ProtocolName::protocol_name`] trait method on each element of that list, and transmit the
+/// returned value on the network. If the remote accepts a given protocol, the element serves as the return value of
+/// the function that performed the negotiation.
```
romanb: Which function (and return value) is this referring to? Isn't the selected protocol just passed to `InboundUpgrade::upgrade_inbound` and `OutboundUpgrade::upgrade_outbound`?

tomaka: Yes, but I'm staying vague because the user typically specifies the upgrade as an associated type to their protocols handler, and therefore don't actually make any function call themselves.

PR #1198, opened 2019-07-09
Tweak ipfs-kad example.

romanb:
The `ipfs-kad` example typically runs into a lot of connection timeouts, which may cause the query to time out. A query timeout currently results in the example to be considered failed, but the example should only be considered failed if no closest peers are found, whether the query timed out or not.

This avoids spurious CI failures, since the examples are run as part of CI.


twittner:
> This avoids spurious CI failures, since the examples are run as part of CI.

Thanks a lot!

Issue #1199, opened 2019-07-10
there are no `KademliaConfig` and `KademliaEvent` in libp2p::kad


flyq:
code in examples/[ipfs-kad.rs](https://github.com/libp2p/rust-libp2p/blob/master/examples/ipfs-kad.rs#L33):
```rust 
use libp2p::kad::{Kademlia, KademliaConfig, KademliaEvent};
```
and `cargo run` failed:
```bash
error[E0432]: unresolved imports `libp2p::kad::KademliaConfig`, `libp2p::kad::KademliaEvent`
  --> src/main.rs:33:29
   |
33 | use libp2p::kad::{Kademlia, KademliaConfig, KademliaEvent};
   |                             ^^^^^^^^^^^^^^  ^^^^^^^^^^^^^
   |                             |               |
   |                             |               no `KademliaEvent` in `kad`
   |                             |               help: a similar name exists in the module: `KademliaOut`
   |                             no `KademliaConfig` in `kad`
```

and they don't exist in [rust doc](https://docs.rs/libp2p/0.10.0/libp2p/kad/index.html).

twittner:
The published version of libp2p  (0.10.0) does not contain those changes which were introduced in #1174. The current [master](https://github.com/libp2p/rust-libp2p/blob/ede114a13c0b423659f16e603dab8a8467cb6683/protocols/kad/src/lib.rs#L37) should work for you.

PR #1200, opened 2019-07-10
fix examples/ipfs-kad.rs

flyq:
fix KademliaConfig & KademliaEvent don't exist in kad and get_closest_peers not in swarm


twittner:
As mentioned in #1199, `KademliaConfig` et al. are not part of the published libp2p-0.10.0 version, but exist in master since #1174 got merged.

PR #1202, opened 2019-07-15
Implement Encode/Decode from parity-codec

cmichi:
The PR contains two commits. The first one should be cheaper, but requires an assertion, which I think we want to avoid. Hence the second one, using a `Vec`.

### DIFF
--- a/misc/multiaddr/src/lib.rs
+++ b/misc/multiaddr/src/lib.rs
@@ -23,6 +23,7 @@ use std::{
     result::Result as StdResult,
     str::FromStr
 };
+use parity_codec::{Compact, Encode, Decode, Input, Output};
 pub use self::errors::{Result, Error};
 pub use self::from_url::{FromUrlErr, from_url, from_url_lossy};
 pub use self::protocol::Protocol;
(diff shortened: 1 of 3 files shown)


misc/multiaddr/src/lib.rs
```diff
@@ -52,6 +53,11 @@ impl Multiaddr {
         Vec::from(&self.bytes[..])
     }
 
+    /// Builds a [`Multiaddr`] from a byte representation.
+    pub fn from_vec(vec: Vec<u8>) -> Self {
```
romanb: This allows creating invalid multi-addresses from arbitrary byte vectors, right? I don't think that is desirable. I think any such conversion should go through one of the existing `TryFrom` instances.

tomaka:
In general, I'm opposed to the change because this is very specific to Substrate, whereas rust-libp2p is kind of its own project.
Additionally, considering that this is stored on chain, I think that the internal representation should be precisely specified, as opposed to an implementation detail.
I don't see any drawback in performing a conversion manually within Substrate.

PR #1204, opened 2019-07-17
 core/src/translation: Support dns4 and dns6

mxinden:
When receiving an observed address on a tcp connection that we initiated, the
observed address contains our tcp dial port, not our tcp listen port. We know
which port we are listening on, thereby we can replace the port within the
observed address. This is done in `core/src/translation`.

So far `fn address_translation` would drop any non ipv4 or ipv6 protocols. This
is problematic in the following setup:

A multi node Substrate cluster where nodes discover each other via their dns
addresses. After A innitiated a connection to B, B sends a indentification
request to A. B is listed in A's `observed_addresses` by its dns address, as A
initiated the connection, thereby it sends back the dns address of B. When B
receives the response it runs it through the `address_translation` function to
clean up tcp listen and dial port switchup. The function drops the address,
given that it is not an ipv4 or ipv6 address.

I have tested this patch in Substrate by cherry-picking "core/src/translation:
Support dns4 and dns6" onto the `v0.10.0` git tag and building Substrate with
the local version.

---

Next to the bug fix above this patch also does:

- core/src/translation: Add unit tests 

- core/nodes/network: %s/nat_translation/replace_dial_with_listen_port/ 

(See individual commit messages for details)

### DIFF
--- a/core/src/nodes/network.rs
+++ b/core/src/nodes/network.rs
@@ -714,16 +714,23 @@ where
         self.incoming_limit
     }
 
-    /// Call this function in order to know which address remotes should dial in order to access
-    /// your local node.
+    /// Call this function in order to know which address remotes should dial to
+    /// access your local node.
     ///
-    /// `observed_addr` should be an address a remote observes you as, which can be obtained for
+    /// When receiving an observed address on a tcp connection that we initiated, the observed
+    /// address contains our tcp dial port, not our tcp listen port. We know which port we are
+    /// listening on, thereby we can replace the port within the observed address.
+    ///
+    /// When receiving an observed address on a tcp connection that we did **not** initiated, the
+    /// observed address should contain our listening port. In case it differs from our listening
+    /// port there might be a proxy along the path.
+    ///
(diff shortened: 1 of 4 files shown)


core/src/nodes/network.rs
```diff
@@ -714,14 +714,23 @@ where
...
+    /// port there might be a proxy along the path.
+    ///
+    /// # Arguments
+    ///
+    /// * `observed_addr` - should be an address a remote observes you as, which can be obtained for
     /// example with the identify protocol.
     ///
-    /// For each listener, calls `nat_traversal` with the observed address and returns the outcome.
-    pub fn nat_traversal<'a>(&'a self, observed_addr: &'a Multiaddr)
+    pub fn replace_dial_with_listen_port<'a>(&'a self, observed_addr: &'a Multiaddr)
```
tomaka: I'd rather call this `address_translation` as well, because this function is just a shortcut for calling `address_translation` multiple times.

mxinden:
@tomaka can you take another look?

PR #1205, opened 2019-07-18
Publish v0.11.0

tomaka:
To publish on crates.io after this PR is merged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,12 @@
+# Version 0.11.0 (2019-07-18)
+
+- Refactored `libp2p-kad`. Records storage is now properly implemented.
+- Extracted the `swarm` and `protocols_handler`-related contents from `libp2p-core` to a new `libp2p-swarm` crate.
+- Renamed `RawSwarm` to `Network`.
+- Added `Floodsub::publish_any`.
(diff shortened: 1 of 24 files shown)


tomaka:
I must admit that the CHANGELOG entry regarding Kademlia is a bit lazy. Ideally it would describe exactly what has changed, but that's a lot of work.

CHANGELOG.md
```diff
@@ -1,3 +1,12 @@
+# Version 0.11.0 (2019-07-18)
+
+- Refactored `libp2p-kad`. Records storage is now properly implemented.
```
romanb:
```suggestion
- `libp2p-kad`: Completed the core functionality of the record storage API, thereby extending the `RecordStore` for provider records. All records expire by default and are subject to regular republication and caching as per the Kademlia spec(s). Expiration and publication intervals are configurable through the `KademliaConfig`.
- `libp2p-kad`: The routing table now never stores peers without a known (listen) address. In particular, on receiving a new inbound connection, the Kademlia behaviour now emits `KademliaEvent::UnroutablePeer` to indicate that in order for the peer to be added to the routing table and hence considered a reachable node in the DHT, a listen address of the peer must be discovered and reported via `Kademlia::add_address`. This is usually achieved through the use of the `Identify` protocol on the same connection(s).
- `libp2p-kad`: Documentation updates.
```

I think that covers the main points of interest.

PR #1213, opened 2019-07-29
Move facade re-exports around

tomaka:
Refactors the `libp2p` facade crate to have re-exports that make more sense.

Instead of just dumping everything, this PR puts up some organization in place.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -47,7 +47,7 @@
 //! Example (Dialing a TCP/IP multi-address):
 //!
 //! ```rust
-//! use libp2p::{Multiaddr, Transport, tcp::TcpConfig};
+//! use libp2p::{Multiaddr, Transport, transport::tcp::TcpConfig};
 //! let tcp = TcpConfig::new();
 //! let addr: Multiaddr = "/ip4/98.97.96.95/tcp/20500".parse().expect("invalid multiaddr");
 //! let _conn = tcp.dial(addr);
@@ -86,7 +86,7 @@
 //!
 //! ```rust
 //! # #[cfg(all(not(any(target_os = "emscripten", target_os = "unknown")), feature = "libp2p-secio"))] {
-//! use libp2p::{Transport, tcp::TcpConfig, secio::SecioConfig, identity::Keypair};
+//! use libp2p::{Transport, transport::tcp::TcpConfig, encryption::secio::SecioConfig, identity::Keypair};
 //! let tcp = TcpConfig::new();
 //! let secio_upgrade = SecioConfig::new(Keypair::generate_ed25519());
 //! let tcp_secio = tcp.with_upgrade(secio_upgrade);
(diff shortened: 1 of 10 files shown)


src/lib.rs
```diff
@@ -161,78 +161,128 @@ pub use multihash;
...
-pub use libp2p_ping as ping;
-#[doc(inline)]
-pub use libp2p_plaintext as plaintext;
-#[doc(inline)]
-pub use libp2p_ratelimit as ratelimit;
-#[doc(inline)]
-pub use libp2p_secio as secio;
+/// We re-export libp2p-core for the `NetworkBehaviour` macro.
+#[doc(hidden)]
+pub use libp2p_core;
```
twittner:
```suggestion
pub use libp2p_core as core;
```

tomaka: Note that it's a hidden re-export (because of the procedural macro). Should I change that anyway?

twittner: Sorry, missed that. Please discard the suggestion.

romanb:
I am largely indifferent on the matter. However, the following are some considerations that come to mind that may be of interest in this context:

  1. It seems desirable to retain a symmetry between crate names and the top-level `libp2p`-crate modules. E.g. currently the top-level modules of crate `libp2p-xyz` can be found in `libp2p::xyz` in the `libp2p` facade crate. I feel like introducing intermediate modules like `transport` and `protocol` should thus also be reflected in the crate names. This may just be aesthetic preference, of course, but not doing so does have the consequence that despite allowing `libp2p::transport::abc` and `libp2p::protocol::abc` to coexist, the obvious crate names would conflict (`libp2p-abc`).

  2. Seeing authentication (usually accompanied by authenticated encryption) protocols as a special case with a dedicated namespace (in this PR `encryption`) may be a good idea, but I would probably opt for the terminology `auth(thenticate)`, since authentication is the primary job of these protocols even if that usually happens in the context of establishing a secure channel, i.e. for confidentiality (the plaintext protocol being the prominent exception). This terminology would then also be consistent with `libp2p_core::upgrade::Builder::authenticate`, through which these protocols are applied on a transport. To furthermore keep the symmetry discussed in (1), crate names like `libp2p-auth-noise`, `libp2p-auth-secio` and `libp2p-auth-plaintext` should be considered, re-exported in the facade crate as `libp2p::auth::noise` etc.

  3. Analogously to authentication, recognizing the special role of multiplexers in transport upgrades (`libp2p::upgrade::Builder::multiplex`) could lead to considering `libp2p::multiplex::yamux` (crate `libp2p-multiplex-yamux`) and `libp2p::multiplex::mplex` (crate `libp2p-multiplex-mplex`).

  4. Under consideration of the previous points, that would leave all other "protocols" in `libp2p::protocol::xyz` (crate `libp2p-protocol-xyz`) and the definition of a protocol could then actually be "a crate providing a (implementation of) a `NetworkBehaviour`". One could then of course argue about whether it should actually be `libp2p::behaviour::xyz` (crate `libp2p::behaviour::xyz`) or rather `trait NetworkBehaviour -> trait Protocol`.

tomaka:
I'm not going to work on this PR anymore and would start from scratch and thus I'm tempted to close.
On the other hand, this PR is useful as a reference point and a reminder to do this.

mxinden:
>  On the other hand, this PR is useful as a reference point and a reminder to do this.

I am closing here. Still useful as a reference point even when closed.

PR #1215, opened 2019-07-30
[libp2p-kad] Generalise record keys.

romanb:
This PR proposes a generalization of record keys from `Multihash` to a new opaque `record::Key` type, as previously mentioned and discussed in [the context of PR #1189](https://github.com/libp2p/rust-libp2p/pull/1189/files#r299686846).

### DIFF
--- a/protocols/kad/src/behaviour.rs
+++ b/protocols/kad/src/behaviour.rs
@@ -29,13 +29,12 @@ use crate::jobs::*;
 use crate::kbucket::{self, KBucketsTable, NodeStatus};
 use crate::protocol::{KadConnectionType, KadPeer};
 use crate::query::{Query, QueryId, QueryPool, QueryConfig, QueryPoolState};
-use crate::record::{store::{self, RecordStore}, Record, ProviderRecord};
+use crate::record::{self, store::{self, RecordStore}, Record, ProviderRecord};
 use fnv::{FnvHashMap, FnvHashSet};
 use futures::prelude::*;
 use libp2p_core::{ConnectedPoint, Multiaddr, PeerId};
 use libp2p_swarm::{NetworkBehaviour, NetworkBehaviourAction, PollParameters, ProtocolsHandler};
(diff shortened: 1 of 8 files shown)


romanb:
> Why not being generic over the key type and requiring the relevant trait bounds where required?

I'm currently shying away from it because such a type parameter would seem to penetrate all of the internal and external API types: `Kademlia`, `KademliaEvent`, `GetRecordResult`, ..., `KademliaHandlerEvent`, ... I'm not sure whether such an effort is worth it, considering that record keys are usually just compared for equality. It basically seems to be about whether conversion of a user's chosen key type to / from a "byte representation" is done by `libp2p-kad` by being generic over the key type with relevant trait bounds, or leaving these conversions between a `record::Key` and some custom type to be done by the user, the latter being the current approach.

romanb:
> Looks good to me.
> The only thing that slightly worries me is that it's now easier to accidentally break compatibility with go-libp2p and js-libp2p by building a `record::Key` that is _not_ a `Multihash`.
> 
> But that would be solved by having integration tests.

According to [the spec](https://github.com/libp2p/specs/blob/cb971aa17b2ee6bb7f20df6daed2e7cd4c7e861d/kad-dht/README.md#record-keys), for `GET_VALUE` and `PUT_VALUE` the key is actually "an unstructured array of bytes" and may be subject to some context-aware validation that even takes the value into account, or something like that. And then for provider records it is supposed to be validated as a CID. And lastly it states "There are intentions to move to multihash [5] keys **in the future**" (emphasis mine) which completes the confusion on my part. So I think we're good (and probably even more compatible) by accepting arbitrary bytes as keys, rather than restricting them to multihashes.

romanb:
@twittner Since I think generalizing from a `Multihash` to an opaque `Key` can be seen as a stepping-stone to a possible further generalization (from a Key to a type parameter) - i.e. this PR does not preclude taking the abstraction further later, I hope you're fine with merging it PR as-is.

PR #1217, opened 2019-08-01
[libp2p-kad] Scope pending RPCs to queries.

romanb:
Ensure that any still pending RPCs related to a query are removed once the query terminates (successfully or through timeout). It seemed most appropriate to scope pending RPCs to queries so that any such remaining events are dropped together with the query when the query terminates (for whatever reason).

### DIFF
--- a/protocols/kad/src/behaviour.rs
+++ b/protocols/kad/src/behaviour.rs
@@ -663,10 +663,10 @@ where
     }
 
     /// Handles a finished (i.e. successful) query.
-    fn query_finished(&mut self, query: Query<QueryInner>, params: &mut impl PollParameters)
+    fn query_finished(&mut self, q: Query<QueryInner>, params: &mut impl PollParameters)
         -> Option<KademliaEvent>
     {
-        let result = query.into_result();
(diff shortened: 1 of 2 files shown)


protocols/kad/src/behaviour.rs
```diff
@@ -1054,12 +1049,14 @@ where
...
     fn inject_connected(&mut self, peer: PeerId, endpoint: ConnectedPoint) {
-        while let Some(pos) = self.pending_rpcs.iter().position(|(p, _)| p == &peer) {
-            let (_, rpc) = self.pending_rpcs.remove(pos);
-            self.queued_events.push_back(NetworkBehaviourAction::SendEvent {
-                peer_id: peer.clone(),
-                event: rpc,
-            });
+        // Queue events for sending pending RPCs to the connected peer.
+        // There can be only one pending RPC for a particular peer and query per definition.
+        for (peer_id, event) in self.queries.iter_mut().filter_map(|q|
```
tomaka: I'm not a fan of these complicated nested functional-style getters, but it's also not a big deal.

Issue #1211, opened 2019-07-25
Generate an event when accept(2) returns an error


tomaka:
This morning I realized that our Polkadot bootnodes were frequently close to the limit of 1024 simultaneously-open file descriptors.

Once we reach that limit, incoming TCP connections will fail to get accepted. In other words, the `Stream` of incoming connections provided by the `TcpListener` will return an error.

In order to circumvent that issue, we are currently using the `tk-listen` crate. See here:

https://github.com/libp2p/rust-libp2p/blob/6999c36d781eac60ca80fcc2f944ddb2e16b4306/transports/tcp/src/lib.rs#L181

If such an error happens, `tk-listen` will silently ignore the error and force a little delay before trying again, with the hope that the number of file descriptors has decreased and the error goes away.

However, the drawback of that approach, is that on a higher level we are totally clueless of what is happening. The server might be completely unreachable, with no way to understand why.

I think that we should:

- Change the semantics of `Transport::Listener` to mention that returning an error is "normal" and doesn't mean that the listener should be closed. While this is technically a violation of the `Stream` API, it is **not** a violation of the new `Stream` API of `futures-preview`, in which the stream produces a list of `Result` items, as opposed to the old futures approach of failing to produce an item. When the `Stream` produces an error, it should continue getting polled.

- In `libp2p-tcp`, if the `TcpListener` returns an error, propagate that error to the upper layers. When the listener gets polled again immediately after (as it should be), force a small delay before querying the underlying `TcpListener`, in order to not flood the upper layers with errors.

twittner:
> * Change the semantics of `Transport::Listener` to mention that returning an error is "normal" and doesn't mean that the listener should be closed. While this is technically a violation of the `Stream` API, [...]

It is actually compatible with [futures-0.1 `Stream`s](https://docs.rs/futures/0.1.28/futures/stream/trait.Stream.html):

> Currently as of the 0.1 release of this library an error on a stream **does not terminate the stream**. That is, after one error is received, another error may be received from the same stream (it's valid to keep polling).

romanb:
Is this not closed by https://github.com/libp2p/rust-libp2p/pull/1218?

PR #1218, opened 2019-08-01
Return TCP listener errors.

twittner:
As described in #1211, we were previously silently waiting after errors that occurred on incoming TCP connections. This PR makes the actual error accessible by returning it from the listener stream and subsequently putting it into a listener event. This allows clients to log the error or base other decisions on it. To allow correlation of (error) events to listeners, each listener is now assigned an ID which is also part of listener events and can be used to remove individual listeners from the listeners stream if desired. By default we continue polling listeners which errored.


core/src/nodes/listeners.rs
```diff
@@ -83,7 +86,22 @@ where
...
+}
+
+/// The ID of a single listener.
+///
+/// It is part of most [`ListenersEvent`]s and can be used to remove
+/// individual listeners from the [`ListenersStream`].
+#[derive(Copy, Clone, Debug, PartialEq, Eq, Hash, PartialOrd, Ord)]
+pub struct ListenerId(u64);
+
+impl fmt::Display for ListenerId {
```
tomaka: Shouldn't that just be `Debug`?

core/src/nodes/listeners.rs
```diff
@@ -105,65 +125,93 @@ where
...
+    /// A listener closed.
     Closed {
+        /// The ID of the listener that closed.
+        listener_id: ListenerId,
         /// The listener that closed.
         listener: TTrans::Listener,
-        /// The error that happened. `Ok` if gracefully closed.
-        result: Result<(), <TTrans::Listener as Stream>::Error>,
     },
+    /// A listener errored.
```
tomaka: That's too vague. It should mention that the listener can continue running, and that the error is informative.

transports/tcp/src/lib.rs
```diff
@@ -353,10 +352,63 @@ enum Addresses {
...
 
+/// Incoming connection stream which pauses after errors.
+#[derive(Debug)]
+struct Listener<S> {
+    /// The incoming connections.
+    stream: S,
+    /// The current pause if any.
+    pause: Option<Delay>,
+    /// How long to pause after an error.
+    duration: Duration
```
tomaka:
```suggestion
    pause_duration: Duration
```

transports/tcp/src/lib.rs
```diff
@@ -353,10 +352,63 @@ enum Addresses {
...
+where
+    S: Stream,
+    S::Error: std::fmt::Display
+{
+    type Item = S::Item;
+    type Error = S::Error;
+
+    /// Polls for incoming connections, pausing if an error is encountered.
+    fn poll(&mut self) -> Poll<Option<S::Item>, S::Error> {
+        if let Some(mut delay) = self.pause.take() {
```
tomaka: We could use `as_mut()` here in order to not extract and push again the timer all the time.

core/src/nodes/listeners.rs
twittner: You mean, `ListenerId` should not have a `Display` impl?

transports/tcp/src/lib.rs
twittner: I think `as_mut` saves at most one move per pause. Normally, `pause` is `None` and `Delay` would only notify the task when the time is up.

core/src/nodes/listeners.rs
tomaka: Yes

twittner: Why not?

tomaka: More like: why? As far as I know, `Display` is for types that you may want to show to the end-user of your application. `Debug` is for types that you intend to show to the programmer.

twittner: When it comes to implementing `std` traits "Why not?" is actually the better question to ask I think. We also derive a couple of `std` traits like `PartialOrd` et al. which we do not use ourselves but as a public API type we should ask why we want to prevent users to use `ListenerId` in some particular context. Why should we make it more difficult than necessary to display `ListenerId` to users? `std::fmt` states in [fmt::Display vs fmt::Debug](https://doc.rust-lang.org/std/fmt/index.html#fmtdisplay-vs-fmtdebug):

> `fmt::Display` implementations assert that the type can be faithfully represented as a UTF-8 string at all times.

which is certainly true for `ListenerId` and the [Rust API Guidelines](https://rust-lang-nursery.github.io/api-guidelines/interoperability.html) recommend:

> [...] crates that define new types should eagerly implement all applicable, common traits

Therefore I think implementing `Display` is fair.

tomaka: But the internal representation of `ListenerId` is at the discretion of `listeners.rs`. It could change in the future.

In other words, a `ListenerId` is not a defined as being a number, but as being an opaque identifier. And opaque identifiers can't be faithfully represented as a UTF-8 string.

twittner: > But the internal representation of `ListenerId` is at the discretion of `listeners.rs`. It could change in the future.

Sure it can. But we make several commitments, for instance that a `ListenerId` can be compared for equality, can be ordered, can be copied or cloned, a hash value of it can be computed and that it has a UTF-8 representation. I do not understand why you take issue with the latter but none of the former commitments.

> In other words, a `ListenerId` is not a defined as being a number, but as being an opaque identifier that is valid in a certain context. And opaque identifiers can't be faithfully represented as a UTF-8 string.

Most identifiers, opaque or not, have UTF-8 representations, e.g. UUID, IBAN, etc. In fact a Version-4 UUID is a truly opaque identifier with a UTF-8 representation whereas  we make no effort to hide metadata and can hardly say that `ListenerId` is opaque. And while it is not itself a number it is nonetheless represented as a number.

All that being said I do not want to spend more time arguing about such a minor point and will therefore remove the `Display` impl of `ListenerId`.

tomaka:
For some reason I don't have the button to merge master on your branch.

twittner:
Should work now.

swarm/src/behaviour.rs
```diff
@@ -121,6 +121,14 @@ pub trait NetworkBehaviour {
     fn inject_new_external_addr(&mut self, _addr: &Multiaddr) {
     }
 
+    /// A listener experienced an error.
+    fn inject_listener_error(&mut self, _id: ListenerId, _err: &dyn std::error::Error) {
```
romanb: I'm wondering how useful a `&dyn Error` is, i.e. isn't the only thing that can be done with it to print it via `Debug` or `Display`? All methods on `std::error::Error` that allow inspecting the error (`downcast` / `is`) seem to require the `'static` lifetime. I would expect that a main reason to pass an error to a callback is to allow the callback to potentially inspect and handle it according to the type of the error.

romanb: I.e. wouldn't a custom enum `ListenerError` be desirable?

twittner: The `'static` bound should be added I think. As for 

> I.e. wouldn't a custom enum `ListenerError` be desirable?

What should the `ListenerError` enum contain?

romanb: > What should the `ListenerError` enum contain?

To start with it would probably just be
```rust
#[non_exhaustive]
pub enum ListenError {
    /// A listener failed due to an error encountered by the underlying transport.
    ///
    /// A transport error is often transient in nature, e.g. an I/O error.
    TransportError(Box<dyn Error + 'static>),
}
```
with the possibility to add `libp2p-swarm`-specific error cases later. I'm just mentioning it, because I've experimented with something similar for `inject_addr_reach_failure`, which currently also just takes a `&dyn Error`, [here](https://github.com/libp2p/rust-libp2p/commit/bcebde7b28c5a7b941ed7f86f8d9893d869d5092#diff-bd8e7718bb252fedaadead082bed8489R134). In that case, there are already `libp2p-swarm`-specific error cases other than a transport error. I see a few, admittedly small, advantages of an indirection through an explicit enum:

  1. It "annotates" the opaque error as a transport error, thereby documenting through the types that it makes sense to try to downcast / inspect the error as a e.g. TCP transport error.
  2. A little future-proving of the API with added type information, i.e. if there are other error cases in the future, they can be added through an explicit new constructor, instead of having to document that the `dyn Error` may be either a transport error or one of possibly a number of `libp2p-swarm`-specific errors.
  3. If the `DialError` idea linked to above is added later as well, there would be some symmetry.

This is just a suggestion that could also be investigated further later and need not delay merging this PR!

twittner: I would like to keep it as a trait object for the time being because it is rare in my experience that programmes base decisions on specific error cases instead of just failing if an error of any kind is encountered. But it could of course be the topic of another PR I guess. What is your take on this, @tomaka?

tomaka: I don't really have a strong opinion either. I would keep the `&dyn Error` because adding yet another `ListenerError` item adds confusion.

PR #1222, opened 2019-08-09
core: Replace `debug_assert!`s with log messages.

twittner:


tomaka:
Why the `cfg(debug_assertions)`?

twittner:
To preserve the "on debug builds only" semantics. But I have no horse in this race. If you want this to always happen, just tell me.

tomaka:
I would go for this always happening.

PR #1223, opened 2019-08-09
 Replace `listen_addr` with `local_addr`.

### LINKED ISSUE #1221: Change `ListenerEvent::Upgrade` to use the local address.
The current [`ListenerEvent::Upgrade`](https://github.com/libp2p/rust-libp2p/blob/5696b3eb4d285d1e6125f2cefbea130720a6b06e/core/src/transport/mod.rs#L241-L247) includes the remote address of the upgrade and the listen address. Since figuring out the listen address of an incoming connection is [more complicated](https://github.com/libp2p/rust-libp2p/pull/1132) and [fragile](https://github.com/paritytech/substrate/issues/3352) than perhaps initially thought, I would suggest we follow [this suggestion](https://github.com/libp2p/rust-libp2p/pull/1132#issuecomment-497664270) and replace the listen address with the local address of the incoming connection.

This would also mean changing [`ListenersEvent::Incoming`](https://github.com/libp2p/rust-libp2p/blob/5696b3eb4d285d1e6125f2cefbea130720a6b06e/core/src/nodes/listeners.rs#L117-L123) accordingly but, if https://github.com/libp2p/rust-libp2p/pull/1218 is merged, the newly added [`ListenerId`](https://github.com/libp2p/rust-libp2p/pull/1218/files#diff-934ec958acb4de656c28ea2d84af7aceR137) provides as better means to correlate the upgrade with the listener that produced it.

twittner:
PR on top of #1218 ~~(only the last commit https://github.com/libp2p/rust-libp2p/commit/715687bc7749986386362074e3fd94fc86755f94 is new)~~ to fix #1221.

### DIFF
--- a/core/src/transport/memory.rs
+++ b/core/src/transport/memory.rs
@@ -192,7 +192,7 @@ impl Stream for Listener {
         };
         let event = ListenerEvent::Upgrade {
             upgrade: future::ok(channel),
-            listen_addr: self.addr.clone(),
+            local_addr: self.addr.clone(),
             remote_addr: Protocol::Memory(self.port.get()).into()
         };
         Ok(Async::Ready(Some(event)))
--- a/transports/tcp/src/lib.rs
+++ b/transports/tcp/src/lib.rs
@@ -417,7 +417,7 @@ pub struct TcpListenStream {
 
 // If we listen on all interfaces, find out to which interface the given
 // socket address belongs. In case we think the address is new, check
-// all host interfaces again an report new and expired listen addresses.
+// all host interfaces again and report new and expired listen addresses.
 fn check_for_interface_changes(
     socket_addr: &SocketAddr,
     listen_port: u16,
--- a/transports/tcp/src/lib.rs
+++ b/transports/tcp/src/lib.rs
@@ -415,60 +415,59 @@ pub struct TcpListenStream {
     config: TcpConfig
 }
 
-// Map a `SocketAddr` to the corresponding `Multiaddr`.
-// If not found, check for host address changes.
-// This is a function rather than a method due to borrowing issues.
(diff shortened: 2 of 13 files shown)


core/src/transport/memory.rs
```diff
@@ -192,7 +192,7 @@ impl Stream for Listener {
         };
         let event = ListenerEvent::Upgrade {
             upgrade: future::ok(channel),
-            listen_addr: self.addr.clone(),
+            local_addr: self.addr.clone().with(Protocol::Memory(self.port.get())),
```
tomaka: That might be a stupid question, but doesn't that produce `/memory/N/memory/M`?

tomaka:
Looks good to me, modulo the question and the rebasing.

core/src/transport/memory.rs
twittner: No, you are quite right and it was a confused attempt to come up with a local address. I reverted it back to `local_addr: self.addr.clone(),`. Thanks!

transports/tcp/src/lib.rs
```diff
@@ -415,60 +415,59 @@ pub struct TcpListenStream {
...
-
-            // We should now be able to find the listen address of the local socket address,
-            // if not something is seriously wrong and we report an error.
-            if addrs.iter().find(|(i, j, _)| i == &addr.ip() || j.contains(&addr.ip())).is_none() {
-                let msg = format!("{} does not match any listen address", addr.ip());
-                return Err(io::Error::new(io::ErrorKind::Other, msg))
-            }
+// If we listen on all interfaces, find out to which interface the given
+// socket address belongs. In case we think the address is new, check
+// all host interfaces again an report new and expired listen addresses.
```
```suggestion
// all host interfaces again and report new and expired listen addresses.
```

Issue #1227, opened 2019-08-19
Multistream-select debug assertion failure


tomaka:
We've spotted some debug assertions failures in the CI logs of Substrate. You can search for "panicked" here: https://gitlab.parity.io/parity/substrate/-/jobs/213779/raw

```
thread 'tokio-runtime-worker-36' panicked at 'assertion failed: `(left == right)`
  left: `201`,
 right: `36`', /ci-cache/substrate/cargo/test-linux-stable-int/registry/src/github.com-1ecc6299db9ec823/multistream-select-0.5.0/src/negotiated.rs:228:29
```

```
thread 'tokio-runtime-worker-33' panicked at 'assertion failed: `(left == right)`
  left: `201`,
 right: `36`', /ci-cache/substrate/cargo/test-linux-stable-int/registry/src/github.com-1ecc6299db9ec823/multistream-select-0.5.0/src/negotiated.rs:228:29
```

```
thread 'tokio-runtime-worker-2' panicked at 'assertion failed: `(left == right)`
  left: `201`,
 right: `36`', /ci-cache/substrate/cargo/test-linux-stable-int/registry/src/github.com-1ecc6299db9ec823/multistream-select-0.5.0/src/negotiated.rs:228:29
```

cc @romanb

tomaka:
Fixed in #1228

PR #1228, opened 2019-08-19
[multistream-select] Fix restoring remaining buffer on write error in Negotiated I/O streams.

romanb:
This addresses https://github.com/libp2p/rust-libp2p/issues/1227. The wrong argument was being passed to `BytesMut::split_off`. The properties of the `remaining` buffer, in particular that it never increases, are now checked.


tomaka:
Could you please add a commit that bumps the crate version to `0.5.1`?

PR #1230, opened 2019-08-28
Add peer id inlining for small public keys

AgeManning:
This adds peer id inlining to match the behaviour of the go implementation. 

Specifically, it addresses the rust side of https://github.com/libp2p/go-libp2p-daemon/pull/137

A few points to note: 
- This changes the PeerId for `secp256k1` and `ed25519` keys
- The forked `multihash` crate needed to be updated to include the identity hash. The go-implementation of this allows for up to 2^32 byte identity objects and this is mimicked here. The current rust-implementation relied on fixed hash sizes, where the identity "hash" can have sizes up to this length. There is now split logic in this implementation, between true hashes and the `Identity` type (which isn't great, feel free modify it, if there is a cleaner solution). 

I've tested this with `RSA`, `ed25519` and `secp256k1` curves. This now connects and handshakes correctly with the go-daemon. May want to run further tests to ensure nothing upstream gets broken by these changes.


core/src/peer_id.rs
```diff
@@ -52,8 +56,13 @@ impl PeerId {
...
         let key_enc = key.into_protobuf_encoding();
-        let multihash = multihash::encode(multihash::Hash::SHA2256, &key_enc)
-            .expect("sha2-256 is always supported");
+        let hash_algorithm = if key_enc.len() <= MAX_INLINE_KEY_LENGTH {
+            multihash::Hash::Identity
+        } else {
+            multihash::Hash::SHA2256
+        };
+        let multihash =
+            multihash::encode(hash_algorithm, &key_enc).expect("identity and sha2-256 are always supported by known public key types");
```
twittner:
```suggestion
            multihash::encode(hash_algorithm, &key_enc)
                .expect("identity and sha2-256 are always supported by known public key types");
```

core/src/peer_id.rs
```diff
@@ -24,6 +24,10 @@ use quick_error::quick_error;
 use multihash;
 use std::{convert::TryFrom, fmt, str::FromStr};
 
+/// Public keys with byte-lengths smaller than `MAX_INLINE_KEY_LENGTH` will be
+/// automatically used as the peer id using an identity multihash.
+const MAX_INLINE_KEY_LENGTH: usize = 42;
```
twittner: Why 42?

core/src/peer_id.rs
```diff
@@ -63,7 +72,7 @@ impl PeerId {
     pub fn from_bytes(data: Vec<u8>) -> Result<PeerId, Vec<u8>> {
         match multihash::Multihash::from_bytes(data) {
             Ok(multihash) => {
-                if multihash.algorithm() == multihash::Hash::SHA2256 {
+                if multihash.algorithm() == multihash::Hash::SHA2256 || multihash.algorithm() == multihash::Hash::Identity {
```
twittner:
```suggestion
                if multihash.algorithm() == multihash::Hash::SHA2256
                    || multihash.algorithm() == multihash::Hash::Identity
                {
```

misc/multihash/src/lib.rs
```diff
@@ -57,24 +57,48 @@ macro_rules! match_encoder {
...
-        Keccak384 => sha3::Keccak384,
-        Keccak512 => sha3::Keccak512,
-        Blake2b512 => blake2::Blake2b,
-        Blake2s256 => blake2::Blake2s,
-    });
-
-    Ok(Multihash { bytes: output.freeze() })
+    // Custom length encoding for the identity multihash
+    if let Hash::Identity = hash {
+        if input.len() as u32 >= std::u32::MAX {
```
twittner:
```suggestion
        if u64::from(std::u32::MAX) < as_u64(input.len()) {
```

where `as_u64` is a little helper like this:

```rust
#[cfg(any(target_pointer_width = "32", target_pointer_width = "64"))]
fn as_u64(a: usize) -> u64 {
    a as u64
}
```

The alternative is using `TryFrom`, but it does not look pleasant.

misc/multihash/src/lib.rs
```diff
@@ -57,24 +57,48 @@ macro_rules! match_encoder {
...
+            return Err(EncodeError::UnsupportedInputLength);
+        }
+        let mut buf = encode::u16_buffer();
+        let code = encode::u16(hash.code(), &mut buf);
+        dbg!(code.len());
```
twittner:
```suggestion
```

misc/multihash/src/lib.rs
```diff
@@ -57,24 +57,48 @@ macro_rules! match_encoder {
...
+        let mut len_buf = encode::u32_buffer();
+        let size = encode::u32(input.len() as u32, &mut len_buf);
+
+        let total_len = code.len() + size.len() + input.len();
+
+        let mut output = BytesMut::with_capacity(total_len);
+        output.put_slice(code);
+        output.put_slice(size);
+        output.put_slice(input);
+        output.resize(total_len, 0);
```
twittner: Why is resizing necessary here?

misc/multihash/src/lib.rs
```diff
@@ -180,15 +204,25 @@ impl<'a> MultihashRef<'a> {
         let (code, bytes) = decode::u16(&input).map_err(|_| DecodeError::BadInputLength)?;
 
         let alg = Hash::from_code(code).ok_or(DecodeError::UnknownCode)?;
+
+        // handle the identity case
+        if alg == Hash::Identity {
+            let (hash_len, bytes) = decode::u32(&bytes).map_err(|_| DecodeError::BadInputLength)?;
+            if bytes.len() as u32 != hash_len {
```
twittner: Again using `as_u64` from above:

```suggestion
            if as_u64(bytes.len()) != u64::from(hash_len) {
```

core/src/peer_id.rs
AgeManning: To match the libp2p go-implementation: https://github.com/libp2p/go-libp2p-core/blob/6e566d10f4a5447317a66d64c7459954b969bdab/peer/peer.go#L34

misc/multihash/src/hashes.rs
```diff
@@ -60,6 +63,7 @@ impl Hash {
     /// Get the hash length in bytes.
     pub fn size(&self) -> u8 {
         match self {
+            Hash::Identity => 42,
```
AgeManning: Also note that this is entirely redundant. A go implementation uses this: 
https://github.com/multiformats/go-multihash/blob/64ffd94071a5bc58130ea88fd8404408aef993f2/multihash.go#L141

AgeManning:
> I am not sure if we want to support "identity" hashes for up to 4 GiB.

I would tend to agree. The only reason I did this, was because the more featured go-implementation of multi-hashes do it. https://github.com/multiformats/go-multihash/blob/master/multihash.go#L258

It should effect libp2p any really as we are only using this for known-sized keys. Happy to reduce this, am indifferent to the size of identity hashes to support.

misc/multihash/src/lib.rs
AgeManning: It's not. Removed.

core/src/peer_id.rs
twittner: Maybe worth mentioning in a comment.

tomaka: It's part of the libp2p specs. I don't think it's worth adding "this is part of the specs", because it's kind of implied everywhere.

twittner: > I don't think it's worth adding "this is part of the specs", because it's kind of implied everywhere.

No one suggested that.

tomaka: What would the comment be, then? "To match the Go implementation"? That would be wrong,  it's to match the specs.

twittner: I was asking "Why 42?" because I did not know where the number came from. The answer I got was "To match the libp2p go-*implementation*" (emphasis mine). Given this I would have considered it useful to state as much in a comment. Then you pointed out that it is specified like this, in other words *not* just implementation defined. Obviously a different answer, so my suggestion no longer applies.

Generally speaking though I would also appreciate comments referring to specs as they can be a great help for the reader.

tomaka:
The PR looks good. The only problem with it would break any existing network using rust-libp2p (which includes Substrate and Polkadot :sweat_smile: ).

If you don't mind, I think we should for now accept both hashed and non-hashed keys, but only send out our key hashed. In other words, `from_bytes` should accept both `Identity` and `SHA2256`, but `from_public_key` should be left untouched for now. Then it a later release, we change `from_public_key`.

The disadvantage of doing so is that during this transition period, two different `PeerId`s could represent the same node on the network. This isn't great, but I also don't really see any way this could be attacked.

AgeManning:
This originally came from compatibility issues with other implementations. 

If we broadcast a hashed version, the other implementations will reject that multiaddr when trying to connect on that multiaddr, which was the initial problem. 

It sounds like I am the only one really with this issue, and our client uses a forked version of rust-libp2p where I've already added these changes. So perhaps if this is not causing any issues for anyone else at the moment, perhaps we leave this PR, until a later release?

However if you want these in now, I'm happy to make the changes you suggest.

tomaka:
> So perhaps if this is not causing any issues for anyone else at the moment, perhaps we leave this PR, until a later release?

We definitely want this in master sooner or later.

Also, apparently js-libp2p introduced the inlining between 0.12.2 and 0.12.3, which broke everything.

AgeManning:
> Also, apparently js-libp2p introduced the inlining between 0.12.2 and 0.12.3, which broke compatibility with rust-libp2p until this PR is in.

Yeah, I usually use "the go implementation" as a synonym with "the libp2p-spec" as most of the specs are based off this and anything mismatching, we should consult the go-impl, so not surprising js-libp2p also matches this.

I'll leave this PR in your hands and let you merge/modify next release/when dependent projects are ready.

tomaka:
Small note: let's keep that PR open in order to not forget to finish this change.

tomaka:
I merged master on top of this PR and added a comment.

tomaka:
Tested on Polkadot. Unfortunately doesn't work, and nodes refuse our connections. Will need to figure out what is happening before merging.

tomaka:
The problem comes from here: https://github.com/libp2p/rust-libp2p/blob/481849e4f1fd17d94d6a926c5a84b047f7a08c6f/protocols/secio/src/lib.rs#L145
A node before this PR will generate a SHA256 `PeerId` and compare it against the expected identity `PeerId`, which will fail.
I'm not sure how to fix that.

twittner:
> Maybe we can make `id(pubkey)` compare equal to `sha256(hashed_pubkey)`?

I think this points into the right direction. The digest of an identity hash equals the original input, so we can always hash the digest with a different algorithm and compare the result with the other one at hand:

```rust
impl PartialEq<PeerId> for PeerId {
    fn eq(&self, other: &PeerId) -> bool {
        match (self.multihash.algorithm(), other.multihash.algorithm()) {
            (a, b) if a == b => {
                self.multihash.digest() == other.multihash.digest()
            }
            (multihash::Hash::Identity, b) => {
                let x = multihash::encode(b, self.multihash.digest());
                x.as_ref() == Ok(&other.multihash)
            }
            (a, multihash::Hash::Identity) => {
                let x = multihash::encode(a, other.multihash.digest());
                x.as_ref() == Ok(&self.multihash)
            }
            _ => false
        }
    }
}
```

tomaka:
> Maybe we can make id(pubkey) compare equal to sha256(hashed_pubkey)?

One thing to not forget is that we derive `Hash` on `PeerId`, and the hash of `peerid(identity(pub))` has to be the same as the one of `peerid(sha2(hashed_pub))`.

(nothing unfixable, but I thought I'd add a note to not forget)

rklaehn:
Any chance we can get this merged behind a disabled feature flag? It is essential for interop with go-libp2p and presumably also js-libp2p.

I tried to get the chat example to work with go-ipfs, and one problem is that the peer id on the rust side is different than on the go side. Working from this branch, the peer id is identical.

tomaka:
> Any chance we can get this merged behind a disabled feature flag? It is essential for interop with go-libp2p and presumably also js-libp2p.

This PR needs the changes so that `PeerId`s compare equal across identity and hashed versions.
I wanted to do the changes, but I haven't had the time to yet.

tomaka:
Finished in #1413

Issue #1219, opened 2019-08-04
The documentation of libp2p-identify is totally outdated


tomaka:

romanb:
Closed by https://github.com/libp2p/rust-libp2p/pull/1231.

PR #1231, opened 2019-08-28
Refactor the Identify protocol.

romanb:
This PR suggests a small refactoring of the `Identify` protocol, thereby updating the documentation (i.e. addressing https://github.com/libp2p/rust-libp2p/issues/1219). The low-level protocol and handler modules are no longer exposed, the two protocol handlers merged and some constructors of the `IdentifyEvent` renamed (not strictly necessary, feel free to object, but the removal of the public low-level modules is already a breaking change).


protocols/identify/src/protocol.rs
```diff
@@ -52,15 +52,26 @@ pub struct RemoteInfo {
...
     inner: Framed<T, codec::UviBytes<Vec<u8>>>,
 }
 
-impl<T> IdentifySender<T> where T: AsyncWrite {
-    /// Sends back information to the client. Returns a future that is signalled whenever the
-    /// info have been sent.
-    pub fn send(self, info: IdentifyInfo, observed_addr: &Multiaddr) -> IdentifySenderFuture<T> {
+impl<T> fmt::Debug for ReplySubstream<T> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "identify::protocol::ReplySubstream")
```
tomaka:
```suggestion
        f.debug_tuple("ReplySubstream").finish()
```

There are methods on `f` to write out debug implementations. Also the naming scheme is generally just the name of the struct.

protocols/identify/src/identify.rs
```diff
@@ -153,81 +165,91 @@ where
...
+                                let event = IdentifyEvent::Sent { peer_id: peer };
+                                return Async::Ready(NetworkBehaviourAction::GenerateEvent(event));
+                            },
+                            Ok(Async::NotReady) => {
+                                self.pending_replies.push_back(Reply::Sending { peer, io });
+                                if sending == to_send {
+                                    // All remaining futures are NotReady
+                                    break
+                                } else {
+                                    reply = self.pending_replies.pop_front();
```
tomaka: This whole system where you push and pop is quite complicated to understand.
I guess it's fine, because with async/await it should be possible to rewrite it in a very straight-forward way.

romanb: I admit I had wished for it to be simpler, but I hope it is still a small improvement over the two separate collections and loops that were there before, as it makes the state transition explicit in the `Reply` enum and a single loop, together with the use of named fields instead of tuples.

Issue #1208, opened 2019-07-22
Implement /plaintext/2.0.0


tomaka:
https://github.com/libp2p/specs/pull/186

burdges:
Do we actually need this for anything?

tomaka:
No.

tomaka:
> Do we actually need this for anything?

This would be useful for the C++ implementation of libp2p to be able to connect to our nodes for testing purposes, without forcing them to implement secio.

ackintosh:
Hi, I'm interested in this issue. ✨  Will tackle to this.

PR #1236, opened 2019-09-02
Implement /plaintext/2.0.0

ackintosh:
This PR closes https://github.com/libp2p/rust-libp2p/issues/1208 🚀 

TODO:
- [x] Keep PlainTextConfig as `PlainText1Config`
  - put the new config as `PlainText2Config`
- [x] [Validation](https://github.com/libp2p/specs/tree/master/plaintext#exchange)
  - the given peer id is consistent with the given public key


ackintosh:
Thank you for your review! 

> I think, however, that we should keep plaintext 1, as this is a very low-maintenance low-effort burden.

That makes sense. 💡 

I'll update entire the codes to compatible to [the plaintext/2.0.0 spec](https://github.com/libp2p/specs/tree/master/plaintext) (and to include your advice) as this PR is still draft. 😃

ackintosh:
I've added a `TODO` section to [the first comment](https://github.com/libp2p/rust-libp2p/pull/1236#issue-313099924) on this PR. 📝

ackintosh:
I've tested that [examples/ping](https://github.com/libp2p/rust-libp2p/blob/master/examples/ping.rs) works fine with the sample code below:

- sample code based on [libp2p::::build_tcp_ws_secio_mplex_yamux()](https://github.com/libp2p/rust-libp2p/blob/dbbf37d5e56612c4ac958ab2a5da86a8acb13061/src/lib.rs#L244)

```rust
pub fn build_tcp_ws_plaintext_mplex_yamux(keypair: identity::Keypair)
    -> impl Transport<Output = (PeerId, impl core::muxing::StreamMuxer<OutboundSubstream = impl Send, Substream = impl Send, Error = impl Into<io::Error>> + Send + Sync), Error = impl error::Error + Send, Listener = impl Send, Dial = impl Send, ListenerUpgrade = impl Send> + Clone
{
    // An another keys for testing the validation
    let _another_keys = identity::Keypair::generate_ed25519();

    CommonTransport::new()
        .with_upgrade(PlainText2Config { pubkey: keypair.public(), peer_id: keypair.public().into_peer_id() })
        .and_then(move |output, endpoint| {
            let peer_id = output.remote_key.into_peer_id();
            let peer_id2 = peer_id.clone();
            let upgrade = core::upgrade::SelectUpgrade::new(yamux::Config::default(), mplex::MplexConfig::new())
                // TODO: use a single `.map` instead of two maps
                .map_inbound(move |muxer| (peer_id, muxer))
                .map_outbound(move |muxer| (peer_id2, muxer));
            core::upgrade::apply(output.stream, upgrade, endpoint)
                .map(|(id, muxer)| (id, core::muxing::StreamMuxerBox::new(muxer)))
        })
        .with_timeout(Duration::from_secs(20))
}
```

- example/ping.rs

```diff
// ...
// ...
-    let transport = libp2p::build_development_transport(id_keys);
+    let transport = libp2p::build_tcp_ws_plaintext_mplex_yamux(id_keys);
// ...
// ...
```

---

Yes, there is no test codes about `plaintext/2.0.0` yet. I'm going to add test codes in another PR. 😄💦

ackintosh:
https://circleci.com/gh/libp2p/rust-libp2p/15485?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link
> error[E0599]: no variant or associated item named `Rsa` found for type `libp2p_core::identity::PublicKey` in the current scope
>   --> protocols/plaintext/src/handshake.rs:86:24
>    |
> 86 |             PublicKey::Rsa(_) => KeyType::RSA,
>    |                        ^^^ variant or associated item not found in `libp2p_core::identity::PublicKey`

Hmm... I couldn't find how fix it. In my local machine, it is done successfully. 🙄 
@tomaka  I'd appreciate it if you could help me. 🙏🙏

ackintosh:
> error[E0599]: no variant or associated item named Rsa found for type libp2p_core::identity::PublicKey in the current scope


It seems that the error is related to this 👀 : 

https://github.com/libp2p/rust-libp2p/blob/057b3795418ab183c6c6f1cafcb5f9c8e6f5ab4f/core/src/identity.rs#L24-L25

but still I can't find how fix it as I'm new to Rust and WASM...

ackintosh:
Ok, the CI has passed!

@tomaka This PR is ready for review. 🙏

protocols/plaintext/src/handshake.rs
```diff
@@ -0,0 +1,198 @@
...
+                        }
+                    };
+
+                    trace!("received exchange from remote; pubkey = {:?}", context.state.public_key);
+                    Ok((socket, context))
+                })
+        })
+        // Check the validity of the remote's `Exchange`.
+        .and_then(|(socket, context)| {
+            let is_valid = match context.state.peer_id.is_public_key(&context.state.public_key) {
```
tomaka: It's not the role of the plaintext protocol to check whether the public key matches the expected `PeerId`. That's already handled in the general logic of libp2p-core::nodes.
We don't necessarily know what the `PeerId` of the remote is going to be when we initiate the connection.

protocols/plaintext/src/lib.rs
```diff
@@ -55,3 +70,130 @@ impl<C> OutboundUpgrade<C> for PlainTextConfig {
     }
 }
 
+#[derive(Clone)]
+pub struct PlainText2Config {
+    pub peer_id: PeerId,
```
tomaka: I suppose that's the remote's `PeerId`? If so that should be removed.

protocols/plaintext/src/lib.rs
```diff
@@ -55,3 +70,130 @@ impl<C> OutboundUpgrade<C> for PlainTextConfig {
...
+    pub pubkey: identity::PublicKey,
```
tomaka: I guess that's our local public key? This fields needs some documentation explaining what it is.

protocols/plaintext/src/handshake.rs
romanb: This (checking that the transmitted peer ID is "consistent with" the transmitted public key) seems to be required by the spec though:

> Upon receiving an Exchange message from the remote peer, each side will validate that the given peer id is consistent with the given public key by deriving a peer id from the key and asserting that it's a match with the id field in the Exchange message.

This is _in addition to_:

> Dialing a peer in libp2p requires knowledge of the listening peer's peer id. As a result, the dialing peer ALSO verifies that the peer id presented by the listening peer matches the peer id that they attempted to dial.

romanb:
As a side node, if https://github.com/libp2p/rust-libp2p/pull/1240 is accepted, it would be nice if the output of the `Plaintext2Config` is `(PeerId, PlaintextOutput)`, so that it can be directly plugged into `transport::upgrade::Builder::authenticate` from https://github.com/libp2p/rust-libp2p/pull/1240. This was one of the main motivations for the Plaintext v2 protocol - that it can be used as an (insecure, development-only) "authenticator" / "security protocol". It can also be done as a follow-up, of course.

protocols/plaintext/src/handshake.rs
tomaka: Oh ok. I misunderstood the code and for some reason thought that `context.state` was the configuration.

protocols/plaintext/src/lib.rs
tomaka: From the other conversation, this seems to be our local `PeerId`. I think it should be automatically generated from the `pubkey` then, and not provided by the configuration.

ackintosh: Deleted the `PeerId` via https://github.com/libp2p/rust-libp2p/pull/1236/commits/abe8499e659374d878f7133e3473f98eb124f7cd

protocols/plaintext/src/lib.rs
ackintosh: Yes, that is our local public key. Is it better to rename from `pubkey` to `local_pubkey` ?

ackintosh:
@romanb 
Thank you for letting me know about that. It's awesome improvement. ✨ 

I'd like to update the output of the `PlainText2Config` if #1240 is accepted before this PR is merged.

ackintosh:
Now this PR has been catching up for #1240

@tomaka @romanb Please review this PR when you have time. 🙏

protocols/plaintext/src/lib.rs
tomaka: Yes please, rename to `local_public_key`.

protocols/plaintext/src/error.rs
```diff
@@ -0,0 +1,83 @@
...
+    /// Protocol buffer error.
+    ProtobufError(ProtobufError),
+
+    /// Failed to parse one of the handshake protobuf messages.
+    HandshakeParsingFailure,
+
+    /// There is no protocol supported by both the local and remote hosts.
+    NoSupportIntersection,
+
+    /// The peer id of the exchange isn't consist with the remote public key.
```
tomaka:
```suggestion
    /// The peer id of the exchange isn't consistent with the remote public key.
```

protocols/plaintext/src/error.rs
```diff
@@ -0,0 +1,83 @@
...
+            PlainTextError::IoError(e) =>
+                write!(f, "I/O error: {}", e),
+            PlainTextError::ProtobufError(e) =>
+                write!(f, "Protobuf error: {}", e),
+            PlainTextError::HandshakeParsingFailure =>
+                f.write_str("Failed to parse one of the handshake protobuf messages"),
+            PlainTextError::NoSupportIntersection =>
+                f.write_str("There is no protocol supported by both the local and remote hosts"),
+            PlainTextError::PeerIdValidationFailed =>
+                f.write_str("The peer id of the exchange isn't consist with the remote public key"),
```
tomaka:
```suggestion
                f.write_str("The peer id of the exchange isn't consistent with the remote public key"),
```

protocols/plaintext/src/error.rs
```diff
@@ -0,0 +1,83 @@
...
+    IoError(IoError),
+
+    /// Protocol buffer error.
+    ProtobufError(ProtobufError),
+
+    /// Failed to parse one of the handshake protobuf messages.
+    HandshakeParsingFailure,
+
+    /// There is no protocol supported by both the local and remote hosts.
+    NoSupportIntersection,
```
tomaka: This makes it look like there's actually a negotiation or something. It's actually just that we don't understand whatever the remote sent us, and I would merge that with `HandshakeParsingFailure`.

protocols/plaintext/src/handshake.rs
```diff
@@ -0,0 +1,198 @@
...
+                Some(b) => b,
+                None => {
+                    debug!("the remote's `PeerId`s hash algorithm is not unsupported");
+                    return Err(PlainTextError::NoSupportIntersection)
```
tomaka:
```suggestion
                    return Err(PlainTextError::HandshakeParsingFailure)
```

protocols/plaintext/src/handshake.rs
```diff
@@ -0,0 +1,198 @@
...
+            },
+        };
+        let peer_id = match PeerId::from_bytes(prop.take_id()) {
+            Ok(p) => p,
+            Err(_) => {
+                debug!("failed to parse remote's exchange's id protobuf");
+                return Err(PlainTextError::HandshakeParsingFailure);
+            },
+        };
+
```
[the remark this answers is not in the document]
ackintosh: Done via https://github.com/libp2p/rust-libp2p/pull/1236/commits/131740aa9a06cd26210cae6128893305b6287abc

protocols/plaintext/src/error.rs
```diff
@@ -0,0 +1,83 @@
...
+#[derive(Debug)]
+pub enum PlainTextError {
+    /// I/O error.
+    IoError(IoError),
+
+    /// Protocol buffer error.
+    ProtobufError(ProtobufError),
+
+    /// Failed to parse one of the handshake protobuf messages.
+    HandshakeParsingFailure,
```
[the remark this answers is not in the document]
ackintosh: I've addressed on https://github.com/libp2p/rust-libp2p/pull/1236/commits/1850fca90160bf273d6fbc49e01cfea9fbac6cc2

---
`HandshakeParsingFailure` is still remaining as these are not `ProtobufError`.

https://github.com/libp2p/rust-libp2p/blob/c9157f75e8d57bb63f8a9b79a93bf8865d39480f/protocols/plaintext/src/handshake.rs#L81-L94

ackintosh:
https://circleci.com/gh/libp2p/rust-libp2p/15741?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link
> ### ci/circleci: test-win32 
> error[E0721]: `await` is a keyword in the 2018 edition
>    --> /usr/local/cargo/registry/src/github.com-1ecc6299db9ec823/wasm-bindgen-backend-0.2.51/src/codegen.rs:454:79
>     |
> 454 |                         wasm_bindgen::__rt::IntoJsResult::into_js_result(#ret.await)
>     |                                                                               ^^^^^ help: you can use a raw identifier to stay compatible: `r#await`

It seems that the Docker image should be updated. cc: @tomaka 
https://hub.docker.com/r/tomaka/rust-mingw-docker

```sh
$ docker run --rm tomaka/rust-mingw-docker cargo -V
cargo 1.34.0 (6789d8a0a 2019-04-01)
```

ackintosh:
I've finished to update this PR but the CI failed, maybe the failure is not related to updates on this PR.

ackintosh:
The failure will be fixed in https://github.com/libp2p/rust-libp2p/pull/1261

ackintosh:
Thank you @tomaka , I've confirmed the CI has passed.

This PR is ready for your review. @tomaka @romanb 🙏

protocols/plaintext/src/error.rs
```diff
@@ -0,0 +1,76 @@
...
+
```
romanb: Another option is to collapse `HandshakeParsingFailure` and `ProtobufError` into the variant
```rust
InvalidPayload(Option<ProtobufError>)
```
Just a thought, because it seems weird to me to have two different error variants for basically the same kind of error (that all or part of the received message payload is invalid / cannot be parsed).

ackintosh:
@romanb Thanks for your review! I've updated.

romanb:
@tomaka Is this OK to merge from your point of view?

tomaka:
Yes!

Issue #953, opened 2019-02-18
Improve the transport creation experience


tomaka:
Creating a transport is a bit of a message because of modifiers everywhere (eg. https://github.com/paritytech/substrate/blob/ed2faf445f8849e8d90ddadc684ac2658819e55f/core/network-libp2p/src/transport.rs#L30-L60).

Needs some API design thoughts.

romanb:
What do you mean by "modifiers"?

elferdo:
In the same vein, I think we should be able to test the higher level logic without depending on specific transports, for example: https://github.com/libp2p/rust-libp2p/blob/603c7be7c21164a82b5f71ad746254a36a90c02b/protocols/kad/src/behaviour/test.rs#L67.We introduce TCP just to test if one node gives the correct result back. I think we could pack all the network complexity behind proxy objects that could be substituted for shallow doubles in tests.

tomaka:
@elferdo There's a `MemoryTransport` that does the equivalent of `TcpConfig` but keeps communications inside the process. But then you need some multiplexing on top, and writing a multiplexing layer is quite complicated.

elferdo:
Ok, thanks for the pointer of `MemoryTransport`. What I was thinking is more that to test whether Kademlia finds a given `PeerId` as the closest one to another `PeerId` we shouldn't really need any networking stuff. I feel that the coupling between logic and networking is too strong there.

romanb:
Closed by https://github.com/libp2p/rust-libp2p/issues/953.

PR #1240, opened 2019-09-05
Rework the transport upgrade API.

romanb:
This is a proposal for https://github.com/libp2p/rust-libp2p/issues/953.

## Motivation

Although transport upgrades must follow a specific pattern in order for the resulting transport to be usable with a `Network` or `Swarm`, that pattern is currently not well reflected in the transport upgrade API. Rather, transport upgrades are somewhat laborious and involve non-trivial code duplication.

## Overview

This PR introduces a `transport::upgrade::Builder` that is obtained from `Transport::upgrade`. The `Builder` encodes the previously implicit rules for transport upgrades:

  1. Authentication upgrades must happen first.
  2. Any number of upgrades may follow.
  3. A multiplexer upgrade must happen last.

Since multiplexing is the last (regular) transport upgrade (because that upgrade yields a `StreamMuxer` which is no longer a `AsyncRead` / `AsyncWrite` resource, which the upgrade process is based on),
the upgrade starts with `Transport::upgrade` and ends with `Builder::multiplex`, which drops back down to the `Transport`, providing a fluent API.

## Contract for authenticators and multiplexers

Authentication and multiplexer upgrades must adhere to a minimal contract w.r.t their outputs:

  1. An authentication upgrade is given an (async) I/O resource `C` and must produce a pair `(I, D)` where `I: ConnectionInfo` and  `D` is a new (async) I/O resource `D`.

  2. A multiplexer upgrade is given an (async) I/O resource `C` and must produce a `M: StreamMuxer`.

  To that end, two changes to the `secio` and `noise` protocols have been made:
    
  1. The `secio` upgrade now outputs a pair of `(PeerId, SecioOutput)`. The former implements `ConnectionInfo` and the latter `AsyncRead` /  `AsyncWrite`, fulfilling the `Builder` contract.
    
  2. A new `NoiseAuthenticated` upgrade has been added that wraps around any noise upgrade (i.e. `NoiseConfig`) and has an output of  `(PeerId, NoiseOutput)`, i.e. it checks if the `RemoteIdentity` from the handshake output is an `IdentityKey`, failing if that is not the case. This is the standard upgrade procedure one wants for integrating noise with libp2p-core/swarm.

## Examples

A typical example before/after:

### Before

```rust
   let transport = libp2p_tcp::TcpConfig::new()
            .with_upgrade(libp2p_secio::SecioConfig::new(local_key))
            .and_then(move |out, endpoint| {
                let peer_id = out.remote_key.into_peer_id();
                let peer_id2 = peer_id.clone();
                let upgrade = libp2p_mplex::MplexConfig::default()
                    .map_outbound(move |muxer| (peer_id, muxer))
                    .map_inbound(move |muxer| (peer_id2, muxer));
                upgrade::apply(out.stream, upgrade, endpoint)
            });
```

### After

```rust
  let transport = libp2p_tcp::TcpConfig::new()
            .upgrade()
            .authenticate(libp2p_secio::SecioConfig::new(local_key))
            .multiplex(libp2p_mplex::MplexConfig::new());
```

Fore more examples, see the diff, which removes all of the related TODOs.

Since `Transport::with_upgrade` is removed and `transport::upgrade::Upgrade` changed to fit into the `Builder` pipeline, these changes are not fully backward-compatible, but adapting existing code should be  easy.


core/tests/transport_upgrade.rs
```diff
@@ -0,0 +1,109 @@
+
```
tomaka: Needs the license.

PR #1244, opened 2019-09-13
Disable multistream optimization to be compatible

### LINKED ISSUE #1241: multistream-select interoperability issue with JS
When js-libp2p connects to a listener, it starts by sending out `/multistream-select/1.0.0`, then waits for the remote to send back `/multistream-select/1.0.0`, then only sends out the requested protocol.

However, the rust-libp2p listener reads `/multistream-select/1.0.0`, then sends out `/multistream-select/1.0.0`, but doesn't immediately flush that to the TCP socket. It only flushes after having sent back an answer to the requested protocol.

cc @raulk @romanb

tomaka:
Fix #1241 

See the issue for context.

### DIFF
--- a/misc/multistream-select/src/listener_select.rs
+++ b/misc/multistream-select/src/listener_select.rs
@@ -136,7 +136,7 @@ where
                     if io.start_send(Message::Header(Version::V1))?.is_not_ready() {
                         return Ok(Async::NotReady)
                     }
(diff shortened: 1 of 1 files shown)


romanb:
Superseded by https://github.com/libp2p/rust-libp2p/pull/1245.

PR #1246, opened 2019-09-16
Address some reviewing

tomaka:
Addresses some reviewing in #1196

### DIFF
--- a/core/src/nodes/tasks/manager.rs
+++ b/core/src/nodes/tasks/manager.rs
@@ -177,7 +177,7 @@ impl<I, O, H, E, HE, T, C> Manager<I, O, H, E, HE, T, C> {
 
         let task = Box::pin(Task::new(task_id, self.events_tx.clone(), rx, future, handler));
         if let Some(threads_pool) = &mut self.threads_pool {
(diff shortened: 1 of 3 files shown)


core/src/nodes/tasks/manager.rs
```diff
@@ -209,10 +209,6 @@ impl<I, O, H, E, HE, T, C> Manager<I, O, H, E, HE, T, C> {
         let (tx, rx) = mpsc::channel(4);
         self.tasks.insert(task_id, TaskInfo { sender: tx, user_data });
 
-        // TODO: we use `Pin<Box<Pending>>` instead of just `Pending` because `Pending` doesn't
-        // implement `Unpin` even though it should ; this is just a dummy template parameter and
-        // the `Box` is never actually created, so this has no repercusion whatsoever
-        // see https://github.com/rust-lang-nursery/futures-rs/pull/1746
         let task: Task<Pin<Box<futures::future::Pending<_>>>, _, _, _, _, _, _> =
```
mxinden: Wouldn't this also work now?

```suggestion
        let task: Task<futures::future::Pending<_>, _, _, _, _, _, _> =
```

tomaka: Yes, but also why change it?

PR #1247, opened 2019-09-18
misc/mdns: Update to futures-preview

mxinden:
This pull request updates `misc/mdns` to use `futures-preview` `0.3.0-alpha.17`. It is ready for review but still has a couple of `TODO` comments that I would like to have your input on.


misc/mdns/src/service.rs
```diff
@@ -147,21 +158,27 @@ impl MdnsService {
         Ok(MdnsService {
             socket,
             query_socket: UdpSocket::bind((Ipv4Addr::from([0u8, 0, 0, 0]), 0u16)).await?,
-            query_interval: Interval::new(Duration::from_secs(20)),
+            query_interval: Interval::new_at(Instant::now(), Duration::from_secs(20)),
```
tomaka: Why change this?

misc/mdns/src/service.rs
```diff
@@ -173,39 +190,70 @@ impl MdnsService {
...
             let to_send = self.query_send_buffers.remove(0);
-            match self.socket.send_to(&to_send, &From::from(([224, 0, 0, 251], 5353))).await {
-                Ok(bytes_written) => {
+
+            // TODO: The future borrows `to_send`. We can't insert `to_send`
+            // back into `send_buffers` before the borrow is dropped. We work
+            // around this with the additional scope on match. Is there a
+            // prettier way?
+            match {
+                let fut = self.query_socket.send_to(&to_send, *IPV4_MDNS_MULTICAST_ADDRESS);
```
tomaka: That doesn't work, you're dropping the future. There's a lot more code to write here for it to work unfortunately.

misc/mdns/src/service.rs
```diff
@@ -173,39 +190,70 @@ impl MdnsService {
...
             }
         }
 
         // TODO: block needs to be refactored
-        // Check for any incoming packet.
-        match AsyncDatagram::poll_recv_from(Pin::new(&mut self.socket), cx, &mut self.recv_buffer) {
+        match {
+            let fut = self.socket.recv_from(&mut self.recv_buffer);
+            futures::pin_mut!(fut);
+            fut.poll(cx)
```
tomaka: Same remark.

misc/mdns/src/service.rs
mxinden: Simply to not have to wait for 20 seconds within the unit tests. Other than adding a bit of complexity I don't see it having an impact at the compiled logic.

The same is done in current *master*: https://github.com/libp2p/rust-libp2p/blob/46c62b0145a45c899c4557dc5ac9512f779ae8f4/misc/mdns/src/service.rs#L161

misc/mdns/src/behaviour.rs
```diff
@@ -37,28 +37,46 @@ use wasm_timer::{Delay, Instant};
...
     /// The inner service.
-    service: MdnsService,
+    service: MaybeBusyMdnsService,
 
     /// List of nodes that we have discovered, the address, and when their TTL expires.
     ///
     /// Each combination of `PeerId` and `Multiaddr` can only appear once, but the same `PeerId`
     /// can appear multiple times.
+    //
+    // TODO: Why this optimization? Is mdns ever within the hot path of an
```
tomaka: This isn't an optimization. It is necessary in order to store the `Multiaddresses` of the nodes we know about. Without it, you can't implement `addresses_of_peer`.

misc/mdns/src/behaviour.rs
```diff
@@ -37,28 +37,46 @@ use wasm_timer::{Delay, Instant};
...
+    // application? Is the access latency on this data structure relevant?
     discovered_nodes: SmallVec<[(PeerId, Multiaddr, Instant); 8]>,
 
-    /// Future that fires when the TTL at least one node in `discovered_nodes` expires.
+    /// Future that fires when the TTL of at least one node in `discovered_nodes` expires.
     ///
     /// `None` if `discovered_nodes` is empty.
+    // TODO: Would a simple std Instant suffice as well?
```
tomaka: We currently report expired nodes as events of the behaviour. If we want these to be reported "pro-actively", then it has to be a `Delay`.

misc/mdns/src/behaviour.rs
```diff
@@ -264,17 +303,24 @@ where
                     break discovered;
                 },
                 MdnsPacket::ServiceDiscovery(disc) => {
-                    disc.respond(Duration::from_secs(5 * 60));
+                    // MaybeBusyMdnsService should always be Free.
+                    if let MaybeBusyMdnsService::Free(ref mut service) = self.service {
+                        // TODO: This should move to a constant, right? Also can it be consolidated with
+                        // the query response TTL?
+                        service.enqueue_response(disc.build_response(Duration::from_secs(5 * 60)));
+                    }
```
tomaka:
```suggestion
                    } else { debug_assert!(false); }
```

misc/mdns/src/behaviour.rs
```diff
@@ -284,8 +330,7 @@ where
 impl<TSubstream> fmt::Debug for Mdns<TSubstream> {
     fn fmt(&self, fmt: &mut fmt::Formatter<'_>) -> fmt::Result {
         fmt.debug_struct("Mdns")
-            .field("service", &self.service)
+            //.field("service", &self.service)
```
tomaka: Please either remove or make it work (both solutions are fine to me).

misc/mdns/src/service.rs
```diff
@@ -147,33 +153,60 @@ impl MdnsService {
...
             silent,
             recv_buffer: [0; 2048],
             send_buffers: Vec::new(),
             query_send_buffers: Vec::new(),
         })
     }
 
-    pub async fn next_packet(&mut self) -> MdnsPacket {
-        // TODO: refactor this block
+    pub fn enqueue_response(&mut self, rsp: Vec<u8>) {
```
tomaka: I don't think this is supposed to be public, right?

```suggestion
    fn enqueue_response(&mut self, rsp: Vec<u8>) {
```

misc/mdns/src/service.rs
```diff
@@ -147,33 +153,60 @@ impl MdnsService {
...
+        self.send_buffers.push(rsp);
+    }
+
+    /// Returns a future resolving to itself and the next received `MdnsPacket`.
+    ///
+    /// **Note**: Why does `next` take ownership of itself?
```
tomaka: I would explain this in a regular comment, but not in a doc-comment.

misc/mdns/src/service.rs
```diff
@@ -147,33 +153,60 @@ impl MdnsService {
...
+        //
         // Note that we don't use a loop here—it is pretty unlikely that we need it, and there is
         // no point in sending multiple requests in a row.
-        match Stream::poll_next(Pin::new(&mut self.query_interval), cx) {
-            Poll::Ready(_) => {
-                if !self.silent {
-                    let query = dns::build_query();
-                    self.query_send_buffers.push(query.to_vec());
-                }
+        if let Some(_) = self.query_interval.next().now_or_never() {
```
tomaka: That's not correct, unfortunately.
Once you reach `recv().await`, you will block forever until the socket has data for you.
Instead, what we want is for `query_interval` to "interrupt" that and dispatch a new request.

misc/mdns/src/service.rs
```diff
@@ -278,48 +300,43 @@ impl fmt::Debug for MdnsService {
...
     /// > **Note**: Keep in mind that we will also receive this response in an `MdnsResponse`.
     #[inline]
-    pub fn respond<TAddresses>(
+    pub fn build_response<TAddresses>(
         self,
         peer_id: PeerId,
         addresses: TAddresses,
         ttl: Duration,
-    ) -> Result<(), MdnsResponseError>
+    ) -> Result<Vec<u8>, MdnsResponseError>
```
tomaka: Well, that's not great.
The `respond` method was a way to tell **the service** to send a response.
Now that `MdnsQuery` is detached from the service, it doesn't make sense to keep the API similar to what it was before.

misc/mdns/src/service.rs
```diff
@@ -339,21 +356,18 @@ impl<'a> fmt::Debug for MdnsQuery<'a> {
...
 
-impl<'a> MdnsServiceDiscovery<'a> {
-    /// Respond to the query.
+impl MdnsServiceDiscovery {
+    /// Build service discovery response.
     #[inline]
-    pub fn respond(self, ttl: Duration) {
-        let response = dns::build_service_discovery_response(self.query_id, ttl);
-        self.send_buffers.push(response);
+    pub fn build_response(self, ttl: Duration) -> Vec<u8> {
```
tomaka: Same remark.

misc/mdns/src/service.rs
```diff
@@ -373,18 +387,15 @@ impl<'a> fmt::Debug for MdnsServiceDiscovery<'a> {
...
-impl<'a> MdnsResponse<'a> {
-    /// Returns the list of peers that have been reported in this packet.
-    ///
-    /// > **Note**: Keep in mind that this will also contain the responses we sent ourselves.
-    pub fn discovered_peers<'b>(&'b self) -> impl Iterator<Item = MdnsPeer<'b>> {
-        let packet = &self.packet;
-        self.packet.answers.iter().filter_map(move |record| {
+impl MdnsResponse {
+    /// Creates a new `MdnsResponse` based on the provided `Packet`.
+    pub fn new(packet: Packet, from: SocketAddr) -> MdnsResponse {
```
tomaka:
```suggestion
    fn new(packet: Packet, from: SocketAddr) -> MdnsResponse {
```

misc/mdns/src/behaviour.rs
mxinden: Why can this not be a simple heap allocated `Vec` instead of a *maybe* stack allocated `SmallVec`?

misc/mdns/src/behaviour.rs
mxinden: Makes sense. Thanks!

misc/mdns/src/service.rs
mxinden: This function is called within `Mdns` `Networkbehaviour` implementation, thus it needs to be `pub`.

misc/mdns/src/behaviour.rs
tomaka: I guess it doesn't really matter.

misc/mdns/src/service.rs
mxinden: Can you add some more details here? Due to  the fact that `MdnsQuery` does not have  a reference to the `send_buffer` anymore it only builds the response, which later on is send by the `Mdns` `NetworkBehaviour` implementation to the `MdnsService`.

tomaka: What I mean is that it doesn't make sense for the `MdnsQuery` object itself to build the response. With the exception of `query_id`, the `self` isn't necessary.

mxinden:
@tomaka Could you take another look at this?

misc/mdns/src/service.rs
```diff
@@ -410,13 +425,25 @@ impl<'a> MdnsResponse<'a> {
...
+        MdnsResponse {
+            peers,
+            from,
+        }
+    }
+
+    /// Returns the list of peers that have been reported in this packet.
+    ///
+    /// > **Note**: Keep in mind that this will also contain the responses we sent ourselves.
+    pub fn discovered_peers(&self) -> &Vec<MdnsPeer> {
```
tomaka:
```suggestion
    pub fn discovered_peers(&self) -> impl Iterator<Item = &MdnsPeer> {
```

misc/mdns/src/service.rs
```diff
@@ -435,41 +462,22 @@ impl<'a> fmt::Debug for MdnsResponse<'a> {
...
-    /// Returns the list of addresses the peer says it is listening on.
-    ///
-    /// Filters out invalid addresses.
-    pub fn addresses<'b>(&'b self) -> impl Iterator<Item = Multiaddr> + 'b {
-        let my_peer_id = &self.peer_id;
-        let record_value = &self.record_value;
-        self.packet
+impl MdnsPeer {
+    /// Creates a new `MdnsPeer` based on the provided `Packet`.
+    pub fn new(packet: &Packet, record_value: String, my_peer_id: PeerId, ttl: u32) -> MdnsPeer {
```
tomaka: `MdnsPeer` is supposed to be a borrow for a value inside of a `Packet`, similar to a `MutexGuard` being a borrow of a `Mutex` for example.

misc/mdns/src/service.rs
```diff
@@ -516,42 +549,87 @@ impl<'a> fmt::Debug for MdnsPeer<'a> {
...
-    use std::{io, task::Poll, time::Duration};
+    use std::{io::{Error, ErrorKind}, time::Duration};
+    use wasm_timer::ext::TryFutureExt;
     use crate::service::{MdnsPacket, MdnsService};
     use multiaddr::multihash::*;
 
     fn discover(peer_id: PeerId) {
-        let mut service = MdnsService::new().unwrap();
-        let stream = stream::poll_fn(move |cx| -> Poll<Option<Result<(), io::Error>>> {
+        block_on (async {
```
tomaka:
```suggestion
        block_on(async {
```

misc/mdns/src/service.rs
```diff
@@ -516,42 +549,87 @@ impl<'a> fmt::Debug for MdnsPeer<'a> {
...
                     MdnsPacket::Response(response) => {
                         for peer in response.discovered_peers() {
                             if peer.id() == &peer_id {
-                                return Poll::Ready(None);
+                                return;
                             }
                         }
                     }
-                    MdnsPacket::ServiceDiscovery(_) => {}
+                    MdnsPacket::ServiceDiscovery(_) => {panic!("did not expect a service discovery packet")}
```
tomaka:
```suggestion
                    MdnsPacket::ServiceDiscovery(_) => panic!("did not expect a service discovery packet")
```

misc/mdns/src/service.rs
```diff
@@ -516,42 +549,87 @@ impl<'a> fmt::Debug for MdnsPeer<'a> {
...
+                }
+            }
+        })
+    }
+
+    #[test]
```
tomaka:
```suggestion
```

misc/mdns/src/service.rs
```diff
@@ -516,42 +549,87 @@ impl<'a> fmt::Debug for MdnsPeer<'a> {
...
+    // As of today the underlying UDP socket is not stubbed out. Thus tests run in parallel to this
+    // unit tests inter fear with it. Test needs to be run in sequence to ensure test properties.
+    fn respect_query_interval() {
```
tomaka:
```suggestion
    #[test]
    fn respect_query_interval() {
```

misc/mdns/src/service.rs
mxinden: `Packet` is  not static, thus we need to do the parsing when constructing the `MdnsPeer` as the final `MdnsPeer` needs to be static.

mxinden:
Thanks for bearing with me @tomaka. The above comment and https://github.com/libp2p/rust-libp2p/pull/1247/commits/17fd131b24c646c9f16c7ca84bfad12ae305d7eb should address your feedback.

PR #1248, opened 2019-09-18
protocols/noise: Update to futures-preview

mxinden:
This pull request updates `protocols/noise` to use `futures-preview` `0.3.0-alpha.17`. It is ready for review but still has a couple of `TODO` comments that I would like to have your input on.

### DIFF
--- a/core/src/upgrade/map.rs
+++ b/core/src/upgrade/map.rs
@@ -49,7 +49,7 @@ impl<C, U, F, T> InboundUpgrade<C> for MapInboundUpgrade<U, F>
 where
     U: InboundUpgrade<C>,
     U::Future: Unpin,
-    F: FnOnce(U::Output) -> T + Send
+    F: FnOnce(U::Output) -> T
 {
     type Output = T;
     type Error = U::Error;
@@ -117,7 +117,7 @@ impl<C, U, F, T> OutboundUpgrade<C> for MapOutboundUpgrade<U, F>
 where
     U: OutboundUpgrade<C>,
     U::Future: Unpin,
-    F: FnOnce(U::Output) -> T + Send
+    F: FnOnce(U::Output) -> T
 {
     type Output = T;
     type Error = U::Error;
@@ -157,7 +157,7 @@ impl<C, U, F, T> InboundUpgrade<C> for MapInboundUpgradeErr<U, F>
 where
     U: InboundUpgrade<C>,
     U::Future: Unpin,
-    F: FnOnce(U::Error) -> T + Send
+    F: FnOnce(U::Error) -> T
 {
     type Output = U::Output;
     type Error = T;
@@ -211,7 +211,7 @@ impl<C, U, F, T> OutboundUpgrade<C> for MapOutboundUpgradeErr<U, F>
 where
     U: OutboundUpgrade<C>,
     U::Future: Unpin,
-    F: FnOnce(U::Error) -> T + Send
+    F: FnOnce(U::Error) -> T
 {
     type Output = U::Output;
     type Error = T;
--- a/core/src/upgrade/map.rs
+++ b/core/src/upgrade/map.rs
@@ -49,7 +49,7 @@ impl<C, U, F, T> InboundUpgrade<C> for MapInboundUpgrade<U, F>
 where
     U: InboundUpgrade<C>,
     U::Future: Unpin,
-    F: FnOnce(U::Output) -> T
+    F: FnOnce(U::Output) -> T + Send
 {
     type Output = T;
     type Error = U::Error;
@@ -117,7 +117,7 @@ impl<C, U, F, T> OutboundUpgrade<C> for MapOutboundUpgrade<U, F>
 where
     U: OutboundUpgrade<C>,
     U::Future: Unpin,
-    F: FnOnce(U::Output) -> T
+    F: FnOnce(U::Output) -> T + Send
 {
     type Output = T;
     type Error = U::Error;
@@ -157,7 +157,7 @@ impl<C, U, F, T> InboundUpgrade<C> for MapInboundUpgradeErr<U, F>
 where
     U: InboundUpgrade<C>,
     U::Future: Unpin,
-    F: FnOnce(U::Error) -> T
+    F: FnOnce(U::Error) -> T + Send
 {
     type Output = U::Output;
     type Error = T;
@@ -211,7 +211,7 @@ impl<C, U, F, T> OutboundUpgrade<C> for MapOutboundUpgradeErr<U, F>
 where
     U: OutboundUpgrade<C>,
     U::Future: Unpin,
-    F: FnOnce(U::Error) -> T
+    F: FnOnce(U::Error) -> T + Send
 {
     type Output = U::Output;
     type Error = T;
--- a/core/src/upgrade/mod.rs
+++ b/core/src/upgrade/mod.rs
@@ -145,7 +145,7 @@ pub trait InboundUpgrade<C>: UpgradeInfo {
     type Error;
     /// Future that performs the handshake with the remote.
     // TODO: remove Unpin
-    type Future: Future<Output = Result<Self::Output, Self::Error>> + Unpin;
+    type Future: Future<Output = Result<Self::Output, Self::Error>> + Unpin + Send;
 
     /// After we have determined that the remote supports one of the protocols we support, this
     /// method is called to start the handshake.
@@ -186,7 +186,7 @@ pub trait OutboundUpgrade<C>: UpgradeInfo {
     type Error;
     /// Future that performs the handshake with the remote.
     // TODO: remove Unpin
-    type Future: Future<Output = Result<Self::Output, Self::Error>> + Unpin;
+    type Future: Future<Output = Result<Self::Output, Self::Error>> + Unpin + Send;
 
     /// After we have determined that the remote supports one of the protocols we support, this
     /// method is called to start the handshake.
--- a/core/src/upgrade/mod.rs
+++ b/core/src/upgrade/mod.rs
@@ -145,7 +145,7 @@ pub trait InboundUpgrade<C>: UpgradeInfo {
     type Error;
     /// Future that performs the handshake with the remote.
     // TODO: remove Unpin
-    type Future: Future<Output = Result<Self::Output, Self::Error>> + Unpin + Send;
+    type Future: Future<Output = Result<Self::Output, Self::Error>> + Unpin;
 
     /// After we have determined that the remote supports one of the protocols we support, this
     /// method is called to start the handshake.
@@ -186,7 +186,7 @@ pub trait OutboundUpgrade<C>: UpgradeInfo {
     type Error;
     /// Future that performs the handshake with the remote.
     // TODO: remove Unpin
-    type Future: Future<Output = Result<Self::Output, Self::Error>> + Unpin + Send;
+    type Future: Future<Output = Result<Self::Output, Self::Error>> + Unpin;
 
     /// After we have determined that the remote supports one of the protocols we support, this
     /// method is called to start the handshake.
--- a/protocols/noise/src/io.rs
+++ b/protocols/noise/src/io.rs
@@ -337,7 +337,7 @@ impl<T: io::Write> io::Write for NoiseOutput<T> {
 }
 
 impl<T: AsyncRead + Unpin> AsyncRead for NoiseOutput<T> {
-    fn poll_read(self: core::pin::Pin<&mut Self>, cx: &mut futures::task::Context<'_>, buf: &mut [u8]) -> Poll<Result<usize, futures::io::Error>> {
+    fn poll_read(mut self: core::pin::Pin<&mut Self>, cx: &mut futures::task::Context<'_>, buf: &mut [u8]) -> Poll<Result<usize, futures::io::Error>> {
         // TODO: We could also do something along the lines of
         // https://docs.rs/tokio-io/0.1.12/src/tokio_io/async_read.rs.html#80
         // that way we are using our custom std::io::Read::read function defined
--- a/protocols/noise/src/io.rs
+++ b/protocols/noise/src/io.rs
@@ -342,7 +342,7 @@ impl<T: AsyncRead> AsyncRead for NoiseOutput<T> {
 }
 
 impl<T: AsyncWrite> AsyncWrite for NoiseOutput<T> {
-    fn shutdown(&mut self) -> Poll<(), io::Error> {
+    fn shutdown(&mut self) -> Poll<Result<(), io::Error>> {
         self.io.shutdown()
     }
 }
--- a/protocols/noise/src/io.rs
+++ b/protocols/noise/src/io.rs
@@ -26,7 +26,6 @@ use futures::Poll;
 use log::{debug, trace};
 use snow;
 use std::{fmt, io, pin::Pin};
-use futures::io::{AsyncReadExt, AsyncWriteExt};
 use futures::prelude::*;
 
 const MAX_NOISE_PKG_LEN: usize = 65535;
@@ -348,15 +347,15 @@ impl<T: AsyncRead + Unpin> AsyncRead for NoiseOutput<T> {
 
 // TODO: Can this be removed all together?
 impl<T: AsyncWrite + Unpin> AsyncWrite for NoiseOutput<T> {
-    fn poll_write(self: std::pin::Pin<&mut Self>, cx: &mut std::task::Context<'_>, buf: &[u8]) -> futures::task::Poll<std::result::Result<usize, std::io::Error>>{
+    fn poll_write(self: std::pin::Pin<&mut Self>, _cx: &mut std::task::Context<'_>, _buf: &[u8]) -> futures::task::Poll<std::result::Result<usize, std::io::Error>>{
         // TODO: Fix. See comment on poll_read.
         unimplemented!()
     }
-    fn poll_flush(self: std::pin::Pin<&mut Self>, cx: &mut std::task::Context<'_>) -> futures::task::Poll<std::result::Result<(), std::io::Error>> {
+    fn poll_flush(self: std::pin::Pin<&mut Self>, _cx: &mut std::task::Context<'_>) -> futures::task::Poll<std::result::Result<(), std::io::Error>> {
         // TODO: Fix. See comment on poll_read.
         unimplemented!()
     }
-    fn poll_close(self: std::pin::Pin<&mut Self>, cx: &mut std::task::Context<'_>) -> futures::task::Poll<std::result::Result<(), std::io::Error>>{
+    fn poll_close(self: std::pin::Pin<&mut Self>, _cx: &mut std::task::Context<'_>) -> futures::task::Poll<std::result::Result<(), std::io::Error>>{
         // TODO: Fix. See comment on poll_read.
        unimplemented!()
     }
--- a/protocols/noise/src/io.rs
+++ b/protocols/noise/src/io.rs
@@ -24,7 +24,6 @@ pub mod handshake;
 
 use futures::Poll;
 use futures::prelude::*;
-use futures::io::AsyncReadExt;
 use log::{debug, trace};
 use snow;
 use std::{fmt, io, pin::Pin};
@@ -149,7 +148,6 @@ impl<T: AsyncRead + Unpin> AsyncRead for NoiseOutput<T> {
                             return Poll::Ready(Err(e))
                         }
                         Poll::Pending => {
-                            // TODO: Is this needed?
                             this.read_state = ReadState::ReadLen { buf, off };
 
                             return Poll::Pending;
@@ -240,17 +238,20 @@ impl<T: AsyncWrite + Unpin> AsyncWrite for NoiseOutput<T> {
                     *off += n;
                     if *off == MAX_WRITE_BUF_LEN {
                         trace!("write: encrypting {} bytes", *off);
-                        if let Ok(n) = this.session.write_message(buffer.write, buffer.write_crypto) {
-                            trace!("write: cipher text len = {} bytes", n);
-                            this.write_state = WriteState::WriteLen {
-                                len: n,
-                                buf: u16::to_be_bytes(n as u16),
-                                off: 0
+                        match this.session.write_message(buffer.write, buffer.write_crypto) {
+                            Ok(n) => {
+                                trace!("write: cipher text len = {} bytes", n);
+                                this.write_state = WriteState::WriteLen {
+                                    len: n,
+                                    buf: u16::to_be_bytes(n as u16),
+                                    off: 0
+                                }
                             }
-                        } else {
-                            debug!("encryption error");
-                            this.write_state = WriteState::EncErr;
-                            return Poll::Ready(Err(io::ErrorKind::InvalidData.into()))
+                            Err(e) => {
+                                debug!("encryption error: {:?}", e);
+                                this.write_state = WriteState::EncErr;
+                                return Poll::Ready(Err(io::ErrorKind::InvalidData.into()))
+                            }
                         }
                     }
                     return Poll::Ready(Ok(n))
@@ -268,7 +269,6 @@ impl<T: AsyncWrite + Unpin> AsyncWrite for NoiseOutput<T> {
                             return Poll::Ready(Err(e))
                         }
                         Poll::Pending => {
-                            // TODO: Is this needed?
                             this.write_state = WriteState::WriteLen{ len, buf, off };
 
                             return Poll::Pending
@@ -280,8 +280,6 @@ impl<T: AsyncWrite + Unpin> AsyncWrite for NoiseOutput<T> {
                     let n = match Pin::new(&mut this.io).poll_write( cx, &buffer.write_crypto[*off .. len]) {
                         Poll::Ready(Ok(n)) => n,
                         Poll::Ready(Err(e)) => return Poll::Ready(Err(e)),
-                        // TODO: Do we need to persist the state like we do in
-                        // in Poll::Pending on `WriteState::WriteLen`?
                         Poll::Pending => return Poll::Pending,
                     };
                     trace!("write: wrote {}/{} bytes", *off + n, len);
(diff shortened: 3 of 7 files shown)


protocols/noise/src/io/handshake.rs
```diff
@@ -297,121 +307,74 @@ impl<T> State<T>
...
 
-// RecvEmpty -----------------------------------------------------------------
-
 /// A future for receiving a Noise handshake message with an empty payload.
-///
-/// Obtained from [`Handshake::recv_empty`].
 async fn recv_empty<T>(state: &mut State<T>) -> Result<(), NoiseError>
 where
-    T: AsyncRead
+    T: AsyncReadExt + Unpin
```
tomaka: Should be `AsyncRead`, not `AsyncReadExt`. `AsyncReadExt` is automatically implemented on anything that is `AsyncRead`. Same everywhere else you do that.

protocols/noise/tests/smoke.rs
```diff
@@ -18,6 +18,8 @@
 // FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
 // DEALINGS IN THE SOFTWARE.
 
+#![feature(async_closure)]
```
tomaka: We don't want to use features!

protocols/noise/tests/smoke.rs
```diff
@@ -145,55 +146,63 @@ fn run<T, U>(server_transport: T, client_transport: U, message1: Vec<u8>)
 where
     T: Transport<Output = Output>,
     T::Dial: Send + 'static,
-    T::Listener: Send + 'static,
+    T::Listener: Send + Unpin + futures::stream::TryStreamExt + 'static,
```
tomaka: `TryStream`. No `Ext`.

protocols/noise/src/lib.rs
```diff
@@ -157,42 +160,46 @@ where
...
 
     fn upgrade_inbound(self, socket: Negotiated<T>, _: Self::Info) -> Self::Future {
-        let session = self.params.into_builder()
-            .local_private_key(self.dh_keys.secret().as_ref())
-            .build_responder()
-            .map_err(NoiseError::from);
-        Handshake::rt1_responder(socket, session,
-            self.dh_keys.into_identity(),
-            IdentityExchange::Mutual)
+        Box::pin(async move {
```
romanb: Why is it necessary to wrap these blocks in another future that just waits on the inner future?

protocols/noise/src/io.rs
```diff
@@ -121,57 +121,74 @@ enum WriteState {
...
 }
 
-impl<T: io::Read> io::Read for NoiseOutput<T> {
-    fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
-        let buffer = self.buffer.borrow_mut();
+impl<T: AsyncRead + Unpin> AsyncRead for NoiseOutput<T> {
+    fn poll_read(mut self: core::pin::Pin<&mut Self>, cx: &mut futures::task::Context<'_>, buf: &mut [u8]) -> Poll<Result<usize, futures::io::Error>> {
+        // We use a `this` because the compiler isn't smart enough to allow
+        // mutably borrowing multiple different fields from the `Pin` at the
+        // same time.
```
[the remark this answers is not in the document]
tomaka: I don't think there's documentation about that.

The problem here is that `self` is a `Pin`, and `Pin<T>` implements `DerefMut<Target = T>`.
But when you do `self.foo` (to access the field `foo`), the compiler calls `deref_mut` on `self`, which entirely borrows `self`. If you try to access a different field, then it does the same and complains that `self` is already borrowed.

The fact that the compiler allows simultaneously borrowing multiple different fields from the same object is kind of a hack provided by the compiler, and doesn't work across the `DerefMut` trait.

protocols/noise/src/io.rs
```diff
@@ -121,57 +121,74 @@ enum WriteState {
     EncErr
 }
 
-impl<T: io::Read> io::Read for NoiseOutput<T> {
-    fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
-        let buffer = self.buffer.borrow_mut();
+impl<T: AsyncRead + Unpin> AsyncRead for NoiseOutput<T> {
```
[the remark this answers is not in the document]
tomaka: I don't think it's wrong to implement `Read` and `Write` as well (except for the fact that is bloats the code a bit), but in practice I don't really see why anyone would use a crate designed for asynchronous networking in order to do synchronous operations.

protocols/noise/src/lib.rs
tomaka: We can't express the type of the inner future. Async functions are similar to closures in the way that the type they generate is totally opaque.

protocols/noise/src/io.rs
```diff
@@ -121,57 +121,74 @@ enum WriteState {
...
+    fn poll_read(mut self: core::pin::Pin<&mut Self>, cx: &mut futures::task::Context<'_>, buf: &mut [u8]) -> Poll<Result<usize, futures::io::Error>> {
```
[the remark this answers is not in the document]
tomaka: `futures::io::Error` is the same as `std::io::Error`. I would just use `std::io::Error` everywhere to avoid confusions.

protocols/noise/src/io.rs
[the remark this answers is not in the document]
romanb: > The problem here is that self is a Pin, and Pin<T> implements DerefMut<Target = T>.

I see, I'm familiar with that kind of problem from calling other `&mut self` methods from inside a `&mut self` method even though both mutably borrow completely distinct fields of `Self`. I guess it was just wishful thinking on my part that de-referencing a `Pin` (or any type with `Deref`) on `Self` would be treated specially, i.e. could be elided like a newtype and not require going through regular deref-coercions.

> But when you do self.foo (to access the field foo), the compiler calls deref_mut [..]

To my understanding, insertion of `deref()` / `deref_mut()` calls (i.e. deref-coercions) are only inserted by the compiler if they resolve a type mismatch, so if `self: &mut Self` and `Self` indeed has the field `foo`, then `self.foo` does not involve `std::ops::Deref(Mut)`. On the other hand, if `self: Pin<&Self>` and it doesn't get special handling, it makes sense, of course, that `self.foo` is desugared to `(*Deref::deref(&self)).foo` and thus borrows `self` in its entirety. From that point of view, I probably wouldn't call the ability to simultaneously borrow multiple fields a hack though.

protocols/noise/src/lib.rs
romanb: It seems preferable to me though to keep `rt1_responder` & co return an opaque, boxed Future, as was done before with the [`Handshake`](https://github.com/libp2p/rust-libp2p/blob/master/protocols/noise/src/io/handshake.rs#L36) newtype, instead of declaring them as `async fn`s. Then the re-wrapping here in the declaration of the handshake patterns shouldn't be necessary, leaving the code and types much as before, with `type Future = Handshake<...>`.

protocols/noise/src/io.rs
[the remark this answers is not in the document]
romanb: So, I would suggest to remove these "the compiler isn't smart enough" comments, which I find slightly irritating, because this isn't really anything new, it's just the well-known difference between

```rust
let x = &mut self.deref_mut().x;
let y = &mut self.deref_mut().y; // error
```
and
```rust
let mut this = self.deref_mut();
let x = &mut this.x;
let y = &mut this.y; // ok
```

Explicitly writing `let mut this = self.deref_mut();` should make this clear enough.

protocols/noise/src/lib.rs
tomaka: The old code is very hard to upgrade to new futures, because all the reading/writing helper functions provided by the `futures` library (such as `write_all` or `read_exact`) now hold the socket by reference instead of taking ownership of it.

On the other hand, the async/await syntax makes it possible to write code that is very easy to read (a linear list of instructions, as opposed to a state machine). This is one of the main reasons why we're upgrading to stable futures.

romanb: > On the other hand, the async/await syntax makes it possible to write code that is very easy to read (a linear list of instructions, as opposed to a state machine). This is one of the main reasons why we're upgrading to stable futures.

I'm not suggesting not to use async/await for the implementation of `rt1_responder` et al, just to return an opaque `Handshake` future as before, to avoid the repeated re-wrapping in the definition of the handshake pattern configs. It may be my ignorance that I don't see why that wouldn't work and simplify things, maybe I have to try it myself to see what happens.

protocols/noise/src/io.rs
```diff
@@ -215,158 +238,185 @@ impl<T: io::Write> io::Write for NoiseOutput<T> {
...
                 }
-                WriteState::EncErr => return Err(io::ErrorKind::InvalidData.into())
+                WriteState::EncErr => return Poll::Ready(Err(io::ErrorKind::InvalidData.into()))
             }
         }
     }
-}
 
-impl<T: AsyncRead> AsyncRead for NoiseOutput<T> {
-    unsafe fn prepare_uninitialized_buffer(&self, _: &mut [u8]) -> bool {
```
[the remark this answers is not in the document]
mxinden: Do I understand correctly that as of right now the buffer would be zeroed on each call to `poll_read`? Does someone have experience with how expensive that is? Is it worth including `unsafe` code to skip the zeroing?

After doing some basic benchmarking with criterion, not zeroing gives me a performance benefit order of magnitude smaller than anything network io related.

This is in no way a profound or strong opinion from my side. I guess I am mostly curious what you think / experienced.

protocols/noise/src/io.rs
```diff
@@ -215,158 +238,185 @@ impl<T: io::Write> io::Write for NoiseOutput<T> {
...
-impl<T: AsyncWrite> AsyncWrite for NoiseOutput<T> {
-    fn shutdown(&mut self) -> Poll<(), io::Error> {
-        self.io.shutdown()
     }
+
 }
 
 /// Read 2 bytes as frame length from the given source into the given buffer.
 ///
+// TODO: This is not the case, right?
```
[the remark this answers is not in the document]
mxinden: > What do you mean by "This is not the case"? Since buf: &mut [u8; 2], passing an offset > 2 will panic since off is used as an index into buf. It does not panic for off == 2 though, I didn't know that. See also rust-lang/rust#32057.

Ah, right, I was a bit confused. Thanks for clarifying!

> That being said, since it's now using AsyncRead, this can of course be done much more conveniently through AsyncReadExt::read_exact, if you're inclined to do that.

@romanb `read_exact` returns a `Future`. Given that `read_frame_len` is called within a non-async function handling this `Future` would be difficult. Please correct me if I am wrong.

protocols/noise/src/io.rs
```diff
@@ -380,17 +430,26 @@ fn read_frame_len<R: io::Read>(io: &mut R, buf: &mut [u8; 2], off: &mut usize)
 /// be preserved for the next invocation.
 ///
 /// Returns `false` if EOF has been encountered.
-fn write_frame_len<W: io::Write>(io: &mut W, buf: &[u8; 2], off: &mut usize)
-    -> io::Result<bool>
+fn write_frame_len<W: AsyncWrite + Unpin>(mut io: &mut W, cx: &mut futures::task::Context<'_>, buf: &[u8; 2], off: &mut usize)
```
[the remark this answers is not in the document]
mxinden: See above. Happy to be proven wrong.

protocols/noise/src/io/handshake.rs
```diff
@@ -297,121 +307,74 @@ impl<T> State<T>
...
 {
+    // TODO: Do we want to `read` here? Don't we want to read nothing at all
+    // (empty)?
```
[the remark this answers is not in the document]
mxinden: Ah, never mind. I forgot about the two byte frame length bytes that we read within `read`.

protocols/noise/src/io.rs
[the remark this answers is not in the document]
mxinden: I am fine with either way. I just think we need to be consistent across the codebase.

mxinden:
@tomaka @romanb thanks for the thorough review. I either addressed or replied to each of your comments. Would you mind taking another look?

protocols/noise/src/io.rs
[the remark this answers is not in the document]
romanb: For reference, see https://github.com/libp2p/rust-libp2p/pull/1107. I think this is in general just a low-hanging fruit if you know the `AsyncRead` impl does not read from the buffer, which is very unusual. I personally haven't done any benchmarking, but apparently @tomaka did.

protocols/noise/src/io.rs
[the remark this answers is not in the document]
romanb: > > That being said, since it's now using AsyncRead, this can of course be done much more conveniently through AsyncReadExt::read_exact, if you're inclined to do that.
> 
> @romanb `read_exact` returns a `Future`. Given that `read_frame_len` is called within a non-async function handling this `Future` would be difficult. Please correct me if I am wrong.

The uses are inside `AsyncRead` / `AsyncWrite`, so the idea would be to just store the future in `ReadState::ReadLen` (respectively `WriteState::WriteLen`) instead of `buf` + `off` and polling it from within `poll_read` / `poll_write`. Where do you see the difficulties?

protocols/noise/src/io/handshake.rs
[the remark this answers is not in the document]
romanb: Not just that, the handshake messages transmit the key material used in the handshake. Payloads are entirely optional in the handshake messages. `NoiseOutput` reads the frame length and the opaque data of that length from the underlying I/O stream, passes that to [`snow::HandshakeState::read_message`](https://docs.rs/snow/0.6.1/snow/struct.HandshakeState.html#method.read_message) which maintains the handshake state and decrypts the payload, if any. `NoiseOutput` then copies the decrypted payload into the `buf` given to `read` (with intermediate buffering in `buffer`). Analogously for writing.

protocols/noise/src/io.rs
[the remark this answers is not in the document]
tomaka: I did some benchmarking of libp2p a looong time ago, and zero-ing memory was almost a third of the CPU usage during a typical work load.

protocols/noise/src/lib.rs
romanb: See https://github.com/mxinden/rust-libp2p/pull/1.

protocols/noise/src/io.rs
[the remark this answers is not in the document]
romanb: > I am fine with either way. I just think we need to be consistent across the codebase.

Which two competing approaches do you see here? I'm just saying that I think I would not have stumbled over this part of the code were it not for the presence of the (in my opinion misleading) code comment, which made me think for a moment that there is a compiler / borrow-checker limitation in place here that is somehow specific to the usage of `Pin` and deserves special mention, when in fact there is nothing unusual at all going on.

romanb:
I would still like to see the changes from https://github.com/mxinden/rust-libp2p/pull/1 included and preferably the code commentary around mutably borrowing from `Pin` removed (i.e. [this discussion](https://github.com/libp2p/rust-libp2p/pull/1248#discussion_r326865009)). Otherwise it looks good to me!

protocols/noise/src/io.rs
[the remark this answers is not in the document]
mxinden: Addressed in fa7d91985a3a0f0756ccfe2611e9cca4cdd1c6f3.

mxinden:
@tomaka any further comments from your side?

protocols/noise/src/io.rs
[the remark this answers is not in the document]
tomaka: This needs a change to `std::io::Error`.
Also, please improve the formatting. We're supposed to be limited to column 100.

protocols/noise/src/io.rs
```diff
@@ -181,32 +190,39 @@ impl<T: io::Read> io::Read for NoiseOutput<T> {
...
+    unsafe fn initializer(&self) -> futures::io::Initializer {
+        futures::io::Initializer::nop()
+    }
 }
 
-impl<T: io::Write> io::Write for NoiseOutput<T> {
-    fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
-        let buffer = self.buffer.borrow_mut();
+impl<T: AsyncWrite + Unpin> AsyncWrite for NoiseOutput<T> {
+    fn poll_write(mut self: std::pin::Pin<&mut Self>, cx: &mut std::task::Context<'_>, buf: &[u8]) -> futures::task::Poll<std::result::Result<usize, std::io::Error>>{
```
tomaka: Why all the namespaces?

```suggestion
    fn poll_write(mut self: Pin<&mut Self>, cx: &mut Context<'_>, buf: &[u8]) -> Poll<Result<usize, std::io::Error>>{
```

protocols/noise/src/io.rs
```diff
@@ -215,136 +231,148 @@ impl<T: io::Write> io::Write for NoiseOutput<T> {
...
+                            return Poll::Pending
                         }
-                        Ok(true) => ()
                     }
-                    self.write_state = WriteState::WriteData { len, off: 0 }
+                    this.write_state = WriteState::WriteData { len, off: 0 }
                 }
                 WriteState::WriteData { len, ref mut off } => {
-                    let n = self.io.write(&buffer.write_crypto[*off .. len])?;
+                    let n = match ready!(Pin::new(&mut this.io).poll_write( cx, &buffer.write_crypto[*off .. len])) {
```
tomaka:
```suggestion
                    let n = match ready!(Pin::new(&mut this.io).poll_write(cx, &buffer.write_crypto[*off .. len])) {
```

protocols/noise/src/io.rs
```diff
@@ -215,136 +231,148 @@ impl<T: io::Write> io::Write for NoiseOutput<T> {
...
                 }
-                WriteState::EncErr => return Err(io::ErrorKind::InvalidData.into())
+                WriteState::EncErr => return Poll::Ready(Err(io::ErrorKind::InvalidData.into()))
             }
         }
     }
 
-    fn flush(&mut self) -> io::Result<()> {
-        let buffer = self.buffer.borrow_mut();
+    fn poll_flush(mut self: std::pin::Pin<&mut Self>, cx: &mut std::task::Context<'_>) -> futures::task::Poll<std::result::Result<(), std::io::Error>> {
```
tomaka:
```suggestion
    fn poll_flush(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Result<(), io::Error>> {
```

protocols/noise/src/io.rs
```diff
@@ -215,136 +231,148 @@ impl<T: io::Write> io::Write for NoiseOutput<T> {
...
         }
     }
-}
 
-impl<T: AsyncRead> AsyncRead for NoiseOutput<T> {
-    unsafe fn prepare_uninitialized_buffer(&self, _: &mut [u8]) -> bool {
-        false
-    }
-}
+    fn poll_close(mut self: std::pin::Pin<&mut Self>, cx: &mut std::task::Context<'_>) -> futures::task::Poll<std::result::Result<(), std::io::Error>>{
```
tomaka: Same

protocols/noise/src/io.rs
```diff
@@ -356,17 +384,23 @@ impl<T: AsyncWrite> AsyncWrite for NoiseOutput<T> {
 /// for the next invocation.
 ///
 /// Returns `None` if EOF has been encountered.
-fn read_frame_len<R: io::Read>(io: &mut R, buf: &mut [u8; 2], off: &mut usize)
-    -> io::Result<Option<u16>>
+fn read_frame_len<R: AsyncRead + Unpin>(mut io: &mut R, cx: &mut futures::task::Context<'_>, buf: &mut [u8; 2], off: &mut usize)
```
tomaka: Same

protocols/noise/src/io.rs
```diff
@@ -380,17 +414,23 @@ fn read_frame_len<R: io::Read>(io: &mut R, buf: &mut [u8; 2], off: &mut usize)
 /// be preserved for the next invocation.
 ///
 /// Returns `false` if EOF has been encountered.
-fn write_frame_len<W: io::Write>(io: &mut W, buf: &[u8; 2], off: &mut usize)
-    -> io::Result<bool>
+fn write_frame_len<W: AsyncWrite + Unpin>(mut io: &mut W, cx: &mut futures::task::Context<'_>, buf: &[u8; 2], off: &mut usize)
```
tomaka: Same

protocols/noise/src/io/handshake.rs
```diff
@@ -86,129 +87,162 @@ pub enum IdentityExchange {
     None { remote: identity::PublicKey }
 }
 
-impl<T, C> Handshake<T, C>
+/// A future performing a Noise handshake pattern.
+pub struct Handshake<T, C>(
+    Pin<Box<dyn Future<
+        Output = <Handshake<T, C> as Future>::Output
```
tomaka:
```suggestion
        Output = Result<(RemoteIdentity<C>, NoiseOutput<T>), NoiseError>
```

protocols/noise/src/io/handshake.rs
```diff
@@ -86,129 +87,162 @@ pub enum IdentityExchange {
...
+    > + Send>>
+);
+
+impl<T, C> Future for Handshake<T, C> {
+    type Output = Result<(RemoteIdentity<C>, NoiseOutput<T>), NoiseError>;
+
+    fn poll(mut self: Pin<&mut Self>, ctx: &mut task::Context<'_>) -> task::Poll<Self::Output> {
```
tomaka:
```suggestion
    fn poll(mut self: Pin<&mut Self>, ctx: &mut Context<'_>) -> Poll<Self::Output> {
```

protocols/noise/src/lib.rs
```diff
@@ -319,23 +320,20 @@ where
...
+
 {
     type Output = (PeerId, NoiseOutput<Negotiated<T>>);
     type Error = NoiseError;
-    type Future = future::AndThen<
-        <NoiseConfig<P, C, R> as InboundUpgrade<T>>::Future,
-        FutureResult<Self::Output, Self::Error>,
-        fn((RemoteIdentity<C>, NoiseOutput<Negotiated<T>>)) -> FutureResult<Self::Output, Self::Error>
-    >;
+    type Future = Box<dyn Future<Output = Result<Self::Output, Self::Error>> + Unpin>;
```
tomaka:
```suggestion
    type Future = Pin<Box<dyn Future<Output = Result<Self::Output, Self::Error>> + Send>>;
```

protocols/noise/src/lib.rs
```diff
@@ -319,23 +320,20 @@ where
...
 
     fn upgrade_inbound(self, socket: Negotiated<T>, info: Self::Info) -> Self::Future {
-        self.config.upgrade_inbound(socket, info)
-            .and_then(|(remote, io)| future::result(match remote {
-                RemoteIdentity::IdentityKey(pk) => Ok((pk.into_peer_id(), io)),
-                _ => Err(NoiseError::AuthenticationFailed)
+        Box::new(self.config.upgrade_inbound(socket, info)
```
tomaka:
```suggestion
        Box::pin(self.config.upgrade_inbound(socket, info)
```

protocols/noise/src/lib.rs
```diff
@@ -345,24 +343,19 @@ where
...
     C: Protocol<C> + AsRef<[u8]> + Zeroize + Send + 'static,
 {
     type Output = (PeerId, NoiseOutput<Negotiated<T>>);
     type Error = NoiseError;
-    type Future = future::AndThen<
-        <NoiseConfig<P, C, R> as OutboundUpgrade<T>>::Future,
-        FutureResult<Self::Output, Self::Error>,
-        fn((RemoteIdentity<C>, NoiseOutput<Negotiated<T>>)) -> FutureResult<Self::Output, Self::Error>
-    >;
+    type Future = Box<dyn Future<Output = Result<Self::Output, Self::Error>> + Unpin>;
```
tomaka:
```suggestion
    type Future = Pin<Box<dyn Future<Output = Result<Self::Output, Self::Error>> + Send>>;
```

protocols/noise/src/lib.rs
```diff
@@ -345,24 +343,19 @@ where
...
 
     fn upgrade_outbound(self, socket: Negotiated<T>, info: Self::Info) -> Self::Future {
-        self.config.upgrade_outbound(socket, info)
-            .and_then(|(remote, io)| future::result(match remote {
-                RemoteIdentity::IdentityKey(pk) => Ok((pk.into_peer_id(), io)),
-                _ => Err(NoiseError::AuthenticationFailed)
+        Box::new(self.config.upgrade_outbound(socket, info)
```
tomaka:
```suggestion
        Box::pin(self.config.upgrade_outbound(socket, info)
```

protocols/noise/src/lib.rs
mxinden: @tomaka that would require `core::upgrade::InboundUpgrade::Future` to be `Send`. Should that be part of this pull request?

tomaka: We definitely want all futures to be send, as they can move around between threads of a threads pool while they are being executed, so yes.

mxinden: Sounds good. See https://github.com/libp2p/rust-libp2p/pull/1248/commits/555c2df315e44f21ad39d4408445ce2cb84dd1a4.

core/src/upgrade/map.rs
```diff
@@ -49,7 +49,7 @@ impl<C, U, F, T> InboundUpgrade<C> for MapInboundUpgrade<U, F>
 where
     U: InboundUpgrade<C>,
     U::Future: Unpin,
-    F: FnOnce(U::Output) -> T
+    F: FnOnce(U::Output) -> T + Send
```
tomaka: That's not what I was thinking of.

core/src/upgrade/mod.rs
```diff
@@ -145,7 +145,7 @@ pub trait InboundUpgrade<C>: UpgradeInfo {
     type Error;
     /// Future that performs the handshake with the remote.
     // TODO: remove Unpin
-    type Future: Future<Output = Result<Self::Output, Self::Error>> + Unpin;
+    type Future: Future<Output = Result<Self::Output, Self::Error>> + Unpin + Send;
```
tomaka: This neither.

protocols/noise/src/io/handshake.rs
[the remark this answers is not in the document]
mxinden: Mentioning discussion offline for *future me*: 

```rust
fn read_exact(&'a mut self, buf: &'a mut [u8]) -> ReadExact<'a, Self>
```

`read_exact` borrows the asynchronous reader in addition it takes a reference to
the destination buffer [2]. The borrow of the reader (`this.io`) is not allowed to
outlive the `poll` function lifetime, but it has to, as the `ReadExact` future
needs to outlive multiple `poll` calls.

One could transform the future into a future that takes ownership of the underlying I/O resource. But in that situation concurrent reading and writing wouldn't be possible, given that the resource is owned either by a reading or a writing state.

[2] https://docs.rs/futures-preview/0.3.0-alpha.18/futures/io/trait.AsyncReadExt.html#method.read_exact

core/src/upgrade/mod.rs
mxinden: @tomaka I am sorry, I must be misunderstanding something here.

You would like `<NoiseAuthenticated<P, C, R> as InboundUpgrade<T>>::Future` to be `Send`, right?

As far as I understood something can only be `Send` if everything below it is `Send`.

`<NoiseAuthenticated<P, C, R> as InboundUpgrade<T>>::upgrade_inbound` calls `<NoiseConfig<P, C, R> as InboundUpgrade<T>>::upgrade_inbound` which by our requirement above needs to return a `Send` `Future`, thus the added trait bound (`Send`) on the `InboundUpgrade` trait.

What am I missing?

tomaka: As you mentioned, the `InboundUpgrade` is `Send` if all its components are `Send`. So all you have to do is require `Send` on its components within the code of `libp2p-noise`.

> As far as I understood something can only be Send if everything below it is Send.

The exception is `Box<dyn Future>` which is never send and that you have to replace with `Box<dyn Future + Send>`. That's the only change that you might have to do outside of libp2p-noise.

mxinden:
@tomaka thanks for bearing with me. Could you take another look?

mxinden:
I don't have permissions to merge into *libp2p:stable-futures*. Given that both of you gave their *looks-good-to-me*, could you hit the merge button @tomaka @romanb?

PR #1254, opened 2019-09-23
Upgrade libp2p-kad to stable futures

tomaka:
I didn't check if the tests were working (they're most likely not) because they depend on `libp2p-yamux`, which isn't upgraded yet.

### DIFF
--- a/protocols/kad/src/handler.rs
+++ b/protocols/kad/src/handler.rs
@@ -36,8 +36,7 @@ use libp2p_core::{
     upgrade::{self, InboundUpgrade, OutboundUpgrade, Negotiated}
 };
 use log::trace;
-use std::{borrow::Cow, error, fmt, io, time::Duration};
-use tokio_io::{AsyncRead, AsyncWrite};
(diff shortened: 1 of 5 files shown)


protocols/kad/src/handler.rs
```diff
@@ -485,7 +484,10 @@ where
                     _ => false,
                 });
                 if let Some(pos) = pos {
-                    let _ = self.substreams.remove(pos).try_close();
+                    // TODO: we don't properly close down the substream
```
tomaka: Note that this is consistent with the current behaviour, but I realized that if the substream wasn't ready to be closed, we would just drop it immediately. That's not great.

tomaka: Closing a substream requires sending a message saying "close this substream" on the wire, and the closing process normally does that.

If a substream is dropped without being properly closed, the implementation of `StreamMuxer` can detect that and send the "close" message on the wire the next time it gets polled, but that's not an ideal way to do and I'm not sure that we're actually doing this.

tomaka: I'm ok with the current approach as long as we refactor it later with async/await.

PR #1260, opened 2019-09-30
Cargo.toml: Pin protobuf version

mxinden:
The protobuf crate introduced a breaking change within its semver minor
update from 2.8.1 to 2.9.0. This commit ensures libp2p uses anything
within the 2.8 minor releases.

This fixes `cargo build && cargo test` for me locally. Should we add this to other entry level Cargo.toml files as well?


mxinden:
Updating Protobuf files with protobuf 2.9 right now.

mxinden:
Closing here given that protobuf 2.9 has been removed.

PR #1261, opened 2019-09-30
Try fix CI

tomaka:
I suspect https://hub.docker.com/r/parity/rust-builder to be old.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -29,10 +29,6 @@ jobs:
       - restore_cache:
           key: test-cache
       - run:
-          name: Print Rust version
-          command: |
-            rustc --version
-      - run:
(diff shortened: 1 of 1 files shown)


tomaka:
I removed the `win32` build as it was using a personal Docker image, and I don't want to have to maintain that anymore. The win32 build was not actually testing win32, but rather checking if it is compiling on non-Unix environments.

thomaseizinger:
> We can choose a CI service which support windows environment (e.g. AppVayor) if needed.

FYI: GitHub actions also supports Windows.

tomaka:
> FYI: GitHub actions also supports Windows.

GitHub actions is extremely limited and afaik isn't capable of reproducing our CI configuration right now.

PR #1263, opened 2019-09-30
*: Remove usage of custom buffer initialization usage

mxinden:
With version `0.3.0-alpha.19` the futures-preview crate makes the
`AsyncRead::initializer` API unstable.

In order to improve interoperability with e.g. both a library depending
on alpha.18 as well as a library depending on alpha.19 and in order for
rust-libp2p to become stable again, this commit removes all usages of
the unstable `initializer` API.

### Status

- ~As of today *async-std* is [pinned to futures-preview `0.3.0-alpha.18 `](https://github.com/async-rs/async-std/commit/47daf555f33b4a4be7df546e89b883731cb0c579). Given that *rust libp2p* depends on *async-std*, we don't have much of a choice anyways but to stick with `alpha.18` for now.~


tomaka:
Is this ready? Shall we merge this? Or is there any blocker?

muxers/yamux/src/lib.rs
```diff
@@ -80,10 +80,6 @@ where
     fn destroy_outbound(&self, _: Self::OutboundSubstream) {
     }
 
-    unsafe fn prepare_uninitialized_buffer(&self, _: &mut [u8]) -> bool {
```
mxinden: Should this pull request touch this as well?

PR #1265, opened 2019-10-01
Update dependency versions

rschulman:
This PR updates a number of dependencies, particularly in the `noise` crate, that were keeping it from compiling alongside Rocket. It also makes a number of other changes due to changes in dependency APIs. No changes to `rust-libp2p` APIs were made.


core/Cargo.toml
```diff
@@ -38,7 +38,7 @@ void = "1"
 zeroize = "0.9"
 
 [target.'cfg(not(any(target_os = "emscripten", target_os = "unknown")))'.dependencies]
-ring = { version = "0.14", features = ["use_heap"], default-features = false }
+ring = { version = "^0.16", features = ["alloc"], default-features = false }
```
DemiMarie-parity: :+1:

core/src/identity/rsa.rs
```diff
@@ -69,10 +69,12 @@ pub struct PublicKey(Vec<u8>);
...
-        signature::verify(&RSA_PKCS1_2048_8192_SHA256,
-                          Input::from(&self.0),
-                          Input::from(msg),
-                          Input::from(sig)).is_ok()
+        let key = signature::UnparsedPublicKey::new(&RSA_PKCS1_2048_8192_SHA256, &self.0);
+        key.verify(msg, sig).is_ok()
+        // signature::verify(&RSA_PKCS1_2048_8192_SHA256,
+        //                   Input::from(&self.0),
+        //                   Input::from(msg),
+        //                   Input::from(sig)).is_ok()
```
DemiMarie-parity:
```suggestion
```

You don’t need to leave the old code commented out.

protocols/noise/src/io.rs
```diff
@@ -55,12 +56,48 @@ impl Buffer {
     }
 }
 
+/// A passthrough enum for the two kinds of state machines in `snow`
```
DemiMarie-parity: Thank you!!!!  I was stuck on this.

DemiMarie-parity:
@tomaka can you take a look?  I would like to get this merged ASAP.

protocols/noise/src/io.rs
rschulman: Yeah, this blocked me for a while too. I tried creating a trait for it, but ran into all kinds of trait bounds problems.

protocols/noise/src/io.rs
```diff
@@ -55,12 +56,48 @@ impl Buffer {
...
+pub(crate) enum SnowState {
```
romanb:
```suggestion
pub(crate) enum Session {
```

s3krit:
Hi. Would it be possible to also upgrade `libsecp256k1` to `0.3.1` as part of this PR, in order to take advantage of this fix: https://github.com/paritytech/libsecp256k1/commit/11ba23a9766a5079918cd9f515bc100bc8164b50 ? If not, happy to create a separate PR.

PR #1268, opened 2019-10-04
Bump most dependencies

DemiMarie-parity:
This actually builds 😊.


tomaka:
For libp2p-core, I'd actually prefer to do something like this:

```patch
diff --git a/core/Cargo.toml b/core/Cargo.toml
index db014ec0..9d9fb5d6 100644
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -13,7 +13,7 @@ categories = ["network-programming", "asynchronous"]
 asn1_der = "0.6.1"
 bs58 = "0.2.0"
 bytes = "0.4"
-ed25519-dalek = "1.0.0-pre.1"
+ed25519-dalek = "1.0.0-pre.2"
 failure = "0.1"
 fnv = "1.0"
 lazy_static = "1.2"
@@ -25,7 +25,7 @@ futures = "0.1"
 parking_lot = "0.8"
 protobuf = "2.3"
 quick-error = "1.2"
-rand = "0.6"
+rand = "0.7"
 rw-stream-sink = { version = "0.1.1", path = "../misc/rw-stream-sink" }
 libsecp256k1 = { version = "0.2.2", optional = true }
 sha2 = "0.8.0"
@@ -46,7 +46,7 @@ libp2p-swarm = { version = "0.2.0", path = "../swarm" }
 libp2p-tcp = { version = "0.12.0", path = "../transports/tcp" }
 libp2p-mplex = { version = "0.12.0", path = "../muxers/mplex" }
 libp2p-secio = { version = "0.12.0", path = "../protocols/secio" }
-rand = "0.6"
+rand = "0.7"
 quickcheck = "0.8"
 tokio = "0.1"
 wasm-timer = "0.1"
diff --git a/core/src/identity/ed25519.rs b/core/src/identity/ed25519.rs
index 1c6662c2..d1a2cbf1 100644
--- a/core/src/identity/ed25519.rs
+++ b/core/src/identity/ed25519.rs
@@ -22,6 +22,7 @@
 
 use ed25519_dalek as ed25519;
 use failure::Fail;
+use rand::RngCore;
 use super::error::DecodingError;
 use zeroize::Zeroize;
 
@@ -31,7 +32,9 @@ pub struct Keypair(ed25519::Keypair);
 impl Keypair {
     /// Generate a new Ed25519 keypair.
     pub fn generate() -> Keypair {
-        Keypair(ed25519::Keypair::generate(&mut rand::thread_rng()))
+        let mut bytes = [0u8; 32];
+        rand::thread_rng().fill_bytes(&mut bytes);
+        Keypair(ed25519::Keypair::from_bytes(&bytes).unwrap())
     }
 
     /// Encode the keypair into a byte array by concatenating the bytes
@@ -138,7 +141,9 @@ impl Clone for SecretKey {
 impl SecretKey {
     /// Generate a new Ed25519 secret key.
     pub fn generate() -> SecretKey {
-        SecretKey(ed25519::SecretKey::generate(&mut rand::thread_rng()))
+        let mut bytes = [0u8; 32];
+        rand::thread_rng().fill_bytes(&mut bytes);
+        SecretKey(ed25519::SecretKey::from_bytes(&bytes).unwrap())
     }
 
     /// Create an Ed25519 secret key from a byte slice, zeroing the input on success.
```

tomaka:
I derp'ed, as you can't build a valid `Keypair` from a random u64. But you can generate a `SecretKey`, then deduce the corresponding `PublicKey`.

romanb:
What is the relationship of this PR to https://github.com/libp2p/rust-libp2p/pull/1268/? Also, is it really a good idea to pin library dependencies to patch versions? As far as I'm aware, that is not desirable for a library, as it prevents consumers of the library from using the latest patch releases.

tomaka:
I'm going to merge #1265 as it came first and causes the least amount of conflicts with the `stable-futures` branch.

The only argument in favour of mentioning a patch version is that we might be inadvertently using a feature that has been added in a patch version of a crate (which theoretically shouldn't happen for crates after version >1.0.0, but we're mostly using crates v0.x.y).

On the other hand, backporting this PR to `stable-futures` might be hell on earth.

DemiMarie-parity:
@tomaka your patch was far, _far_ cleaner than my idea.  Thank you.

core/src/identity/ed25519.rs
```diff
@@ -138,7 +139,9 @@ impl Clone for SecretKey {
 impl SecretKey {
     /// Generate a new Ed25519 secret key.
     pub fn generate() -> SecretKey {
-        SecretKey(ed25519::SecretKey::generate(&mut rand::thread_rng()))
+        let mut bytes = [0u8; 32];
+        rand::thread_rng().fill_bytes(&mut bytes);
+        SecretKey(ed25519::SecretKey::from_bytes(&bytes).unwrap())
```
rphmeier: we have a strong policy against `unwrap`. don't use it. you need `expect` with a proof that it won't fail.

protocols/secio/src/handshake.rs
```diff
@@ -37,8 +37,8 @@ use sha2::{Digest as ShaDigestTrait, Sha256};
 use std::cmp::{self, Ordering};
 use std::io::{Error as IoError, ErrorKind as IoErrorKind};
 use crate::structs_proto::{Exchange, Propose};
-use tokio_io::codec::length_delimited;
-use tokio_io::{AsyncRead, AsyncWrite};
+#[allow(deprecated)]
```
tomaka: Please remove. Silencing warnings doesn't bring anything to the table.

core/src/identity/ed25519.rs
```diff
@@ -138,7 +139,13 @@ impl Clone for SecretKey {
 impl SecretKey {
     /// Generate a new Ed25519 secret key.
     pub fn generate() -> SecretKey {
-        SecretKey(ed25519::SecretKey::generate(&mut rand::thread_rng()))
+        loop {
+            let mut bytes = [0u8; 32];
+            rand::thread_rng().fill_bytes(&mut bytes);
+            if let Ok(key) = ed25519::SecretKey::from_bytes(&bytes) {
```
tomaka: This can only `Err` if the array size is wrong. The right fix is to add `expect("<some proof>")`.

DemiMarie-parity:
@rphmeier can you re-review this PR?

core/src/peer_id.rs
```diff
@@ -24,10 +24,6 @@ use quick_error::quick_error;
 use multihash;
 use std::{convert::TryFrom, fmt, str::FromStr};
 
-/// Public keys with byte-lengths smaller than `MAX_INLINE_KEY_LENGTH` will be
-/// automatically used as the peer id using an identity multihash.
-const MAX_INLINE_KEY_LENGTH: usize = 42;
```
tomaka: Please keep this.

protocols/noise/src/io/handshake.rs
```diff
@@ -20,6 +20,8 @@
 
 //! Noise protocol handshake I/O.
 
+// these are found in auto-generated code
+#[allow(bare_trait_objects)]
```
tomaka: Please remove.

muxers/yamux/Cargo.toml
```diff
@@ -14,4 +14,5 @@ futures = "0.1"
 libp2p-core = { version = "0.12.0", path = "../../core" }
 log = "0.4"
 tokio-io = "0.1"
-yamux = "0.2.1"
+# yamux = "0.2.1"
+yamux = { git = "https://github.com/paritytech/yamux", branch = "master" }
```
tomaka: Please revert.

PR #1271, opened 2019-10-07
Merge master into stable-futures

tomaka:
This doesn't necessarily need a review, considering that we will later need to review the `stable-futures` branch on top of master.


tomaka:
> I assume it is intentional that you want to merge with conflicts / conflict markers as well as with the change to the NULL cipher in secio for now.

Oh. Actually no. Sorry, I might have forgotten to run a `git add` at some point.

tomaka:
I fixed the additional conflicts.

PR #1272, opened 2019-10-09
fix can't run chat/ping examples

flyq:
update hmac's version, examples run well


tomaka:
What was the problem?

flyq:
There were some errors when ran examples before, and it runs well now, maybe https://github.com/libp2p/rust-libp2p/pull/1273 already fixes it

PR #1273, opened 2019-10-09
Revert CIPHERS set to null

tomaka:
That's a bit embarrassing. Good thing we didn't publish a version.


romanb:
Of course, unless both peers would actually support the NULL cipher, the handshakes would just have failed.

PR #1275, opened 2019-10-10
*: Consolidate protobuf scripts and update to rust-protobuf 2.8.1

mxinden:
This pull request does two things:

- Consolidate the Protobuf build scripts into a single one which is called with the *source* and *destination* from within each crate that depends on `.proto` files. Move the docker build logic into a separate Dockerfile to not rebuild on each generation.

- Update to `rust-protobuf` `v2.8.1` to reduce the Rust compiler warning noise.

Note:

- Due to consolidation some import paths changed. Let me know if this bothers you. I might be able to change it with temporary files in the build scripts.

- I think in the long run we should regenerate the files in our CI on each run to ensure the generated files stay in sync with the `.proto` files. I would prefer not to trust files generated on an engineers laptop.


mxinden:
With your comment above in mind, what do you think of https://github.com/libp2p/rust-libp2p/pull/1275/commits/9f0e098a2e3a05fd341b91e649e5fbc5ff890334 @romanb?

romanb:
Looks good to me!

mxinden:
@twittner @tomaka can you take a look at this as well?

twittner:
> * I think in the long run we should regenerate the files in our CI on each run to ensure the generated files stay in sync with the `.proto` files.

How about we switch to https://crates.io/crates/prost?

mxinden:
@twittner I am fine with either one. Are you fine merging this one first though?

twittner:
> @twittner I am fine with either one. Are you fine merging this one first though?

Sure. I had approved it already.

mxinden:
Will have to wait for https://github.com/libp2p/rust-libp2p/pull/1282.

PR #1278, opened 2019-10-18
Simple Warning Cleanup

peat:
Took a quick pass at cleaning up some of the compiler warnings to start getting familiar with the code. Very low hanging fruit! 😆

### DIFF
--- a/core/src/peer_id.rs
+++ b/core/src/peer_id.rs
@@ -215,7 +215,7 @@ impl Into<multihash::Multihash> for PeerId {
 quick_error! {
     #[derive(Debug)]
     pub enum ParseError {
-        B58(e: bs58::decode::DecodeError) {
+        B58(e: bs58::decode::Error) {
             display("base-58 decode error: {}", e)
             cause(e)
             from()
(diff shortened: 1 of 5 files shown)


core/src/peer_id.rs
```diff
@@ -24,10 +24,6 @@ use quick_error::quick_error;
 use multihash;
 use std::{convert::TryFrom, fmt, str::FromStr};
 
-/// Public keys with byte-lengths smaller than `MAX_INLINE_KEY_LENGTH` will be
-/// automatically used as the peer id using an identity multihash.
-const MAX_INLINE_KEY_LENGTH: usize = 42;
```
romanb: I think that should not be removed, but may be commented out, referring to the related comment that can be found in `PeerId::from_public_key` (where the code that uses it is commented out). This code path will eventually be enabled in an upcoming release. See https://github.com/libp2p/rust-libp2p/pull/1237/files.

peat: Awesome, thanks for the heads up. Reverted. 👍

PR #1281, opened 2019-10-19
examples: Add distributed key value store

mxinden:
This commit adds a basic distributed key value store supporting GET and
PUT commands using Kademlia and mDNS.

### DIFF
--- a/examples/distributed-key-value-store.rs
+++ b/examples/distributed-key-value-store.rs
@@ -1,0 +1,209 @@
+// Copyright 20l9 Parity Technologies (UK) Ltd.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a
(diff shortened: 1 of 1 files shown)


examples/distributed-key-value-store.rs
```diff
@@ -0,0 +1,209 @@
...
+            let record = Record {
+                key,
+                value,
+                publisher: None,
+                expires: None,
+            };
+            kademlia.put_record(record, Quorum::One);
+        }
+        _ => {
+            eprintln!("expected GET or SET");
```
DePizzottri: Typo: "expected GET or **PUT**"

mxinden: Great catch. Thanks @DePizzottri!

PR #1283, opened 2019-10-20
core/src/transport/memory: Add basic dialer listener test

mxinden:
Adds a basic test using two memory transport instances, one dialing the
other, sending a single message over the /wire/.


mxinden:
@romanb @twittner @tomaka friendly ping. What do you think of this pull request?

core/src/transport/memory.rs
```diff
@@ -308,5 +311,45 @@ mod tests {
...
     }
 
-    // TODO: test that is actually works
+    #[test]
+    fn communicating_between_dialer_and_listener() {
+        let msg = [1, 2, 3];
+
+        // Setup listener.
+
+        let rand_port: NonZeroU64;
```
romanb:
```suggestion
        let port = std::ord::max(1, rand::random::<u64>());
```

core/src/transport/memory.rs
```diff
@@ -308,5 +310,37 @@ mod tests {
...
     }
 
-    // TODO: test that is actually works
+    #[test]
+    fn communicating_between_dialer_and_listener() {
+        let msg = [1, 2, 3];
+
+        // Setup listener.
+
+        let t1_addr: Multiaddr = "/memory/1".parse().unwrap();
```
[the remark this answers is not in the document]
romanb: > Good point. I decided for randomizing the port to keep the test simple. I think we can take the risk of not hitting a collision of two _random_ `u64`'s, right?

_Probably_ :smile:

mxinden:
Delaying this pull request until the *stable-future* branch was merged into *master*.

tomaka:
Let's merge this now.

tomaka:
Compilation errors. Branch needs an update.

core/src/transport/memory.rs
```diff
@@ -311,5 +313,43 @@ mod tests {
...
     }
 
-    // TODO: test that is actually works
+    #[test]
+    fn communicating_between_dialer_and_listener() {
+        let msg = [1, 2, 3];
+
+        // Setup listener.
+
+        let rand_port = rand::random::<u64>() + 1;
```
tomaka: Why this `+ 1` by the way?

mxinden: Well, ... `0` would instruct the OS to bind to any port. I guess it is very unlikely. Do you want me to remove it @tomaka?

tomaka: Well, it is equally likely that we randomly generate `u64::max_value()` and that the `+1` either panics or wraps to 0.

romanb: I would go with `saturating_add(1)`.

tomaka: Here's the "Apply suggestion" button ;)

```suggestion
        let rand_port = rand::random::<u64>().saturating_add(1);
```

mxinden: Thanks @romanb and @tomaka!

PR #1285, opened 2019-10-23
Implement Debug for (ed25519/secp256k1)::(Keypair|SecretKey)

c410-f3r:
Sidenote: I see a bunch of deprecated functions, Rust 2015 stuff and `std::mem::uninitialized`. Is anyone going to update this project?

### DIFF
--- a/core/src/identity/ed25519.rs
+++ b/core/src/identity/ed25519.rs
@@ -24,6 +24,7 @@ use ed25519_dalek as ed25519;
 use failure::Fail;
 use super::error::DecodingError;
 use zeroize::Zeroize;
+use core::fmt;
 
 /// An Ed25519 keypair.
 pub struct Keypair(ed25519::Keypair);
@@ -63,6 +64,12 @@ impl Keypair {
     pub fn secret(&self) -> SecretKey {
         SecretKey::from_bytes(&mut self.0.secret.to_bytes())
             .expect("ed25519::SecretKey::from_bytes(to_bytes(k)) != k")
+    }
+}
+
+impl fmt::Debug for Keypair {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("Keypair").field("public", &self.0.public).finish()
     }
 }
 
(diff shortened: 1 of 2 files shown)


tomaka:
We preferably don't want secret keys to be accidentally printed on stdout, ever.
If we are to implement `Debug`, let's have a dummy manual implementation instead.

c410-f3r:
> If we are to implement Debug, let's have a dummy manual implementation instead.

Thank you. Can it be a bunch of zeros?

core/src/identity/secp256k1.rs
```diff
@@ -51,6 +52,12 @@ impl Keypair {
     }
 }
 
+impl fmt::Debug for Keypair {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "Keypair {{ secret: \"Secret Key\", public: \"{:?}\" }}", &self.public)
```
tomaka: The `Formatter` has helpers for outputting structs:

```suggestion
        f.debug_struct("Keypair").field("public", &self.public).finish()
```

core/src/identity/secp256k1.rs
```diff
@@ -70,6 +77,12 @@ impl From<Keypair> for SecretKey {
 #[derive(Clone)]
 pub struct SecretKey(secp256k1::SecretKey);
 
+impl fmt::Debug for SecretKey {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "Secret Key")
```
tomaka:
```suggestion
        write!(f, "SecretKey")
```

core/src/identity/ed25519.rs
```diff
@@ -66,6 +67,16 @@ impl Keypair {
     }
 }
 
+impl fmt::Debug for Keypair {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(
```
tomaka:
```suggestion
        f.debug_struct("Keypair").field("public", &self.0.public).finish()
```

tomaka:
> I never had the change to tell this but I would like thank you for all the awesome work you did for Rust, specially in the gaming ecosystem

Thank you!

c410-f3r:
Updated

Issue #1232, opened 2019-08-29
mdns peer discovery is not compliant to latest specification


shresthagrawal:
> If the encoding of the peer's ID exceeds 63 characters, then the Split at 63rd character workaround can be used.  

The [specs](https://github.com/libp2p/specs/blob/master/discovery/mdns.md) for encoding PeerID greater than 63 character requires to perform split and join with a '.' but the current [implementation](https://github.com/libp2p/rust-libp2p/blob/master/misc/mdns/src/dns.rs#L128) doesn't handle such condition!

tomaka:
Done in #1287 I believe.

PR #1287, opened 2019-10-24
[mdns] - Support for long mDNS names (Bug #1232)

peat:
Hey folks,

I believe this adds support for long mDNS names, as reported in issue https://github.com/libp2p/rust-libp2p/issues/1232 

There are a couple of additional changes that I made in order to test the scenario where long mDNS names are encountered, so I'd love to get feedback about unintended consequences: 

* ~~I added the `std` feature for `ring` in the top-level `Cargo.toml`. I wasn't able to build or test within the mdns directory without it (eg: `cd misc/mdns; cargo test` would fail on `ring` error traits).~~
* I added support for creating `PeerId` from the `Identity` multihash. It's in line with the spec, but I'm not familiar with the nuances. This was done so that I could generate long mDNS names for tests.

Cheers!

### DIFF
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -38,7 +38,7 @@ void = "1"
 zeroize = "1"
 
 [target.'cfg(not(any(target_os = "emscripten", target_os = "unknown")))'.dependencies]
-ring = { version = "^0.16", features = ["alloc"], default-features = false }
+ring = { version = "^0.16", features = ["alloc", "std"], default-features = false }
 untrusted = { version = "0.6" }
 
 [dev-dependencies]
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -38,7 +38,7 @@ void = "1"
 zeroize = "1"
 
 [target.'cfg(not(any(target_os = "emscripten", target_os = "unknown")))'.dependencies]
-ring = { version = "^0.16", features = ["alloc", "std"], default-features = false }
+ring = { version = "^0.16", features = ["alloc"], default-features = false }
 untrusted = { version = "0.6" }
 
 [dev-dependencies]
--- a/core/src/peer_id.rs
+++ b/core/src/peer_id.rs
@@ -219,7 +219,7 @@ impl Into<multihash::Multihash> for PeerId {
 quick_error! {
     #[derive(Debug)]
     pub enum ParseError {
-        B58(e: bs58::decode::DecodeError) {
+        B58(e: bs58::decode::Error) {
             display("base-58 decode error: {}", e)
             cause(e)
             from()
--- a/core/src/peer_id.rs
+++ b/core/src/peer_id.rs
@@ -100,7 +100,7 @@ impl PeerId {
     /// returns back the data as an error.
     #[inline]
     pub fn from_multihash(data: multihash::Multihash) -> Result<PeerId, multihash::Multihash> {
-        if data.algorithm() == multihash::Hash::SHA2256 {
+        if data.algorithm() == multihash::Hash::SHA2256 || data.algorithm() == multihash::Hash::Identity {
             Ok(PeerId { multihash: data })
         } else {
             Err(data)
--- a/core/src/peer_id.rs
+++ b/core/src/peer_id.rs
@@ -26,7 +26,9 @@ use std::{convert::TryFrom, fmt, str::FromStr};
 
 /// Public keys with byte-lengths smaller than `MAX_INLINE_KEY_LENGTH` will be
 /// automatically used as the peer id using an identity multihash.
-const MAX_INLINE_KEY_LENGTH: usize = 42;
+//
+// Note: see `from_public_key` for how this value will be used in the future.
+// const MAX_INLINE_KEY_LENGTH: usize = 42;
 
 /// Identifier of a peer of the network.
 ///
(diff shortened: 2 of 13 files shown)


tomaka:
I agree that should have this change, but I'm a bit confused by the specs. The fact that you have to modify the code of `PeerId` in order to be able to generate the long `PeerIds` that are relevant for this issue kind of shows that this couldn't happen in reality.

peat:
I don't think the Rust implementation would ever create an oversized DNS label because it only responds to queries with the locally generated SHA2256 `PeerId` -- so I'm happy to yank the code for using the `Identity` multihashes in the query response use case.

I would like to retain the code in `service.rs` -- the parsing is a little cleaner, and I think it's good to have the capability to parse the long format mDNS query responses for interop with other implementations.

@tomaka, would you be more comfortable seeing those changes, given your concerns about the spec being a bit loose?

tomaka:
> would you be more comfortable seeing those changes, given your concerns about the spec being a bit loose?

The scope of the changes is good (IMO). I'll probably review early next week.

core/Cargo.toml
```diff
@@ -38,7 +38,7 @@ void = "1"
 zeroize = "1"
 
 [target.'cfg(not(any(target_os = "emscripten", target_os = "unknown")))'.dependencies]
-ring = { version = "^0.16", features = ["alloc"], default-features = false }
+ring = { version = "^0.16", features = ["alloc", "std"], default-features = false }
```
tomaka: Please revert. One PR = one change. I opened https://github.com/libp2p/rust-libp2p/pull/1289.

```suggestion
ring = { version = "^0.16", features = ["alloc"], default-features = false }
```

core/src/peer_id.rs
```diff
@@ -26,7 +26,9 @@ use std::{convert::TryFrom, fmt, str::FromStr};
 
 /// Public keys with byte-lengths smaller than `MAX_INLINE_KEY_LENGTH` will be
 /// automatically used as the peer id using an identity multihash.
-const MAX_INLINE_KEY_LENGTH: usize = 42;
+//
+// Note: see `from_public_key` for how this value will be used in the future.
+// const MAX_INLINE_KEY_LENGTH: usize = 42;
```
tomaka: Please keep this.

misc/mdns/src/dns.rs
```diff
@@ -211,6 +203,37 @@ fn append_u16(out: &mut Vec<u8>, value: u16) {
     out.push((value & 0xff) as u8);
 }
 
+/// If a peer ID is longer than 63 characters, split it into segments to
+/// be compatible with RFC 1035.
+fn segment_peer_id(peer_id: &String) -> String {
```
tomaka:
```suggestion
fn segment_peer_id(peer_id: String) -> String {
```

It might be premature optimization, but we could add some pass-through:

```
if peer_id.as_bytes().len() <= 63 { return peer_id }
```

misc/mdns/src/dns.rs
```diff
@@ -211,6 +203,37 @@ fn append_u16(out: &mut Vec<u8>, value: u16) {
...
+    // This will only perform one allocation except in extreme circumstances.
```
tomaka: I think you can replace the function's body with something like this:

```suggestion
    String::from_utf8(peer_id.as_bytes().chunks(63).join(&b'.')).expect("we copy from a UTF8 string; qed")
```

peat: The problem is that `.chunks` can't `.join` (in 1.83.0) ... I went through several iterations to see if I could make that work elegantly, and ended up with the `for ... in` loop for clarity and minimal number of allocations.

misc/mdns/src/dns.rs
peat: Updated 👍

core/src/peer_id.rs
peat: Done 👍

misc/mdns/src/dns.rs
```diff
@@ -211,6 +206,40 @@ fn append_u16(out: &mut Vec<u8>, value: u16) {
...
+/// If a peer ID is longer than 63 characters, split it into segments to
+/// be compatible with RFC 1035.
+fn segment_peer_id(peer_id: String) -> String {
+    // Guard for the most common case
+    if peer_id.len() <= MAX_LABEL_LENGTH { return peer_id }
+
+    // This will only perform one allocation except in extreme circumstances.
+    let mut out = String::with_capacity(peer_id.len() + 8);
+
+    for (idx, chr) in peer_id.chars().enumerate() {
```
tomaka: This line bothers me a bit. We know the string only contains ASCII characters, so I guess it's fine. I would have preferred using bytes.

misc/mdns/src/dns.rs
mxinden: I looked further into this as I was curious, I guess not because it really matters in this case. I am sorry if this is just useless noise to the discussion.

Under the assumption that the string only contains ASCII characters, why not do:

```Rust
    let b = peer_id
        .as_bytes()
        .chunks(MAX_LABEL_LENGTH)
        .collect::<Vec<_>>()
        .join(&b'.');
    String::from_utf8(b).expect("we copy from a UTF8 string; qed")
```

Using a for-loop and manually preallocating the right amount of memory seems like an early optimization to me. I have benchmarked this [here](https://github.com/mxinden/segment/blob/master/benches/lib.rs). Using iterators with `join` allows one to not copy the characters one by one as in the for loop, but instead in chunks (See https://doc.rust-lang.org/src/alloc/slice.rs.html#664).

Running the above linked benchmarks in criterion shows that the above would be twice as fast (on my machine) as the for loop. In addition I find it easier to comprehend.

@peat @tomaka  what do you think?

PR #1288, opened 2019-10-28
Update the stable-futures branch to master

tomaka:
Another very painful update of the `stable-futures` branch.


protocols/plaintext/src/pb.rs
```diff
@@ -1,4 +1,4 @@
-// Copyright 2017 Parity Technologies (UK) Ltd.
```
mxinden: Interesting that git thinks this file moved.

protocols/plaintext/src/pb/structs.rs
```diff
@@ -0,0 +1,278 @@
+// This file is generated by rust-protobuf 2.3.0. Do not edit
```
mxinden: Protobuf version will be updated once plaintext moves to the consolidated scripts. (Seem comment above)

PR #1295, opened 2019-11-05
Remove using void to bypass ICE

tomaka:
It turns out that nightly Rust ICEs when using enum with zero variants.
See https://github.com/rust-lang/rust/issues/65462

We were hoping that this ICE would eventually be fixed, but it's about to hit stable and it still isn't fix.

The solution now is unfortunately to remove using empty enums from our API.

Should be merged before #1294

### DIFF
--- a/protocols/identify/Cargo.toml
+++ b/protocols/identify/Cargo.toml
@@ -22,7 +22,6 @@ tokio-codec = "0.1"
 tokio-io = "0.1.0"
 wasm-timer = "0.1"
 unsigned-varint = { version = "0.2.1", features = ["codec"] }
-void = "1.0"
 
 [dev-dependencies]
 libp2p-mplex = { version = "0.12.0", path = "../../muxers/mplex" }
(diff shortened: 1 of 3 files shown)


DemiMarie-parity:
@tomaka We can implement `Void` differently:

```rust
mod void {
	pub struct Void(());
	pub fn absurd(_i: Void) -> ! {
		// This could just as well be an unchecked, unsafe unreachable
		unreachable!("void cannot be constructed");
	}
}
```

This works because `void::Void` has a private constructor, so it cannot actually be constructed.

tomaka:
@DemiMarie-parity I'm kind of hesitating to keep `Void` at all. We used it because we thought that `!` was being stabilized and that we could replace `Void` with `!` briefly in the future. It turned out that `!` isn't stable yet.

DemiMarie-parity:
@tomaka In that case, can you go ahead and merge this and #1294?

PR #1301, opened 2019-11-12
Update `rw-stream-sink` to futures-0.3.

twittner:


twittner:
> [...] as long as the rest of rust-libp2p compiles with it.

Indeed, I just rebased (did not think you would review so quickly), as it was not working well with the rest of the code base. I am working my way through, core, tcp, secio, mplex, etc. The new version is far less disruptive as it is back to using a single type parameter.

> I still would like to review the logic of the code once the PR is final, in case I spot a mistake.

By all means. Please note that fwiw at least the tests in rw-stream-sink have been ported as well and run successfully.

twittner:
If you prefer we could also close this PR in preference for #1302 which is based on this one and accumulate the changes for futures-0.3 there first.

PR #1304, opened 2019-11-13
Publish 0.13.1

### LINKED ISSUE #1303: Compile errors on parity multihash versions higer than 0.1.1
Libp2p doesn't build if parity multi-address's version is higher than 0.1.1 because the error enum isn't fully specified on line 134 of the peer_id file.

On this line the enum is under specified:  
https://github.com/libp2p/rust-libp2p/blob/v0.12.0/core/src/peer_id.rs#L134
Because the error defined here has two enum types:
https://github.com/libp2p/rust-libp2p/blob/master/misc/multihash/src/errors.rs#L5

Which creates the error:
error[E0004]: non-exhaustive patterns: `Err(UnsupportedInputLength)` not covered
   --> /Users/pvienhage/.cargo/registry/src/github.com-1ecc6299db9ec823/libp2p-core-0.12.0/src/peer_id.rs:132:15
    |
132 |         match multihash::encode(alg, &enc) {
    |               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ pattern `Err(UnsupportedInputLength)` not covered
    |
    = help: ensure that all possible cases are being handled, possibly by adding wildcards or more match arms

When the build has fully updated versions of each sub package.

tomaka:
Close #1303


tomaka:
We need to yank multihash 0.1.4.

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# Version 0.13.1 (2019-11-13)
+
+- Maintenant release to bump dependencies and deal with an accidental breaking change in multihash 0.1.4.
```
twittner:
```suggestion
- Maintenance release to bump dependencies and deal with an accidental breaking change in multihash 0.1.4.
```

PR #1306, opened 2019-11-14
protocols/plaintext: Move to stable futures and use unsigned varints

mxinden:
The plaintext 2.0 specification requires to use unsigned varints for
frame length delimiting instead of fixed 4 byte integer frame length
delimiting. This commit aligns the implementation with the
specification.

### DIFF
--- a/protocols/plaintext/src/handshake.rs
+++ b/protocols/plaintext/src/handshake.rs
@@ -20,16 +20,11 @@
 
 use bytes::BytesMut;
 use std::io::{Error as IoError, ErrorKind as IoErrorKind};
-use futures::Future;
-use futures::future;
-use futures::sink::Sink;
-use futures::stream::Stream;
+use futures::prelude::*;
+use futures_codec::Framed;
 use libp2p_core::{PublicKey, PeerId};
 use log::{debug, trace};
 use crate::pb::structs::Exchange;
-use tokio_io::{AsyncRead, AsyncWrite};
-use tokio_io::codec::length_delimited;
-use tokio_io::codec::length_delimited::Framed;
 use protobuf::Message;
 use crate::error::PlainTextError;
 use crate::PlainText2Config;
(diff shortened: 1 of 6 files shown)


protocols/plaintext/src/lib.rs
```diff
@@ -18,15 +18,18 @@
...
 use crate::error::PlainTextError;
-use void::Void;
-use futures::future::FutureResult;
+use unsigned_varint::codec::UviBytes;
 use crate::handshake::Remote;
+use futures::future::{self, Ready};
+use futures::{Sink, Stream};
+use futures::prelude::*;
+use futures_codec::Framed;
+use libp2p_core::{identity, InboundUpgrade, OutboundUpgrade, UpgradeInfo, upgrade::Negotiated, PeerId, PublicKey};
```
twittner: This line is much longer than 100 characters (cf. [.editorconfig](https://github.com/libp2p/rust-libp2p/blob/master/.editorconfig))

protocols/plaintext/src/lib.rs
```diff
@@ -108,144 +111,144 @@ impl UpgradeInfo for PlainText2Config {
...
 
-#[inline]
 fn map_err(err: io::Error) -> io::Error {
     debug!("error during plaintext handshake {:?}", err);
     io::Error::new(io::ErrorKind::InvalidData, err)
 }
 
 pub struct PlainTextMiddleware<S> {
-    inner: Framed<S, BytesMut>,
+    inner: Framed<S, UviBytes<Vec<u8>>>,
```
twittner: Why did you decide to move from `BytesMut` to `Vec<u8>`?

protocols/plaintext/src/lib.rs
```diff
@@ -108,144 +111,144 @@ impl UpgradeInfo for PlainText2Config {
...
 {
-    type Item = BytesMut;
-    type Error = io::Error;
+    type Item = Result<Vec<u8>, io::Error>;
 
-    #[inline]
-    fn poll(&mut self) -> Poll<Option<Self::Item>, Self::Error> {
-        self.inner.poll()
+    fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        // TODO: Too much of a hack? (BytesMut -> Vec<u8>)
```
twittner: I would stick with `BytesMut` instead of copying the data.

protocols/plaintext/src/handshake.rs
```diff
@@ -109,45 +104,33 @@ impl HandshakeContext<Local> {
     }
 }
 
-pub fn handshake<S>(socket: S, config: PlainText2Config)
-    -> impl Future<Item = (Framed<S, BytesMut>, Remote), Error = PlainTextError>
+pub async fn handshake<S>(socket: S, config: PlainText2Config)
+    -> Result<(Framed<S, unsigned_varint::codec::UviBytes<Vec<u8>>>, Remote), PlainTextError>
```
twittner: I would import `unsigned_varint::codec::UviBytes` and shorted this line. Again, I would keep using `BytesMut`.

protocols/plaintext/src/handshake.rs
```diff
@@ -109,45 +104,33 @@ impl HandshakeContext<Local> {
...
+    S: AsyncRead + AsyncWrite + Send + Unpin,
 {
-    let socket = length_delimited::Builder::new()
-        .big_endian()
-        .length_field_length(4)
-        .new_framed(socket);
+    // The handshake messages all start with a variable-length integer indicating the size.
+    let mut socket = Framed::new(
+        socket,
+        unsigned_varint::codec::UviBytes::<Vec<u8>>::default()
```
twittner: I think you can write `Framed::new(socket, UviBytes::default())` here.

mxinden:
> Why did you decide to move from `BytesMut` to `Vec<u8>`?

This wasn't too much of a conscious decision, but rather me mechanically comparing the `plaintext` implementation to the `secio` implementation of `Stream` and `Sink`. Thanks for the catch!

Can you take another look?

protocols/plaintext/src/lib.rs
```diff
@@ -108,144 +121,138 @@ impl UpgradeInfo for PlainText2Config {
 
 impl<C> InboundUpgrade<C> for PlainText2Config
 where
-    C: AsyncRead + AsyncWrite + Send + 'static
+    C: AsyncRead + AsyncWrite + Send + Unpin + 'static
 {
     type Output = (PeerId, PlainTextOutput<Negotiated<C>>);
     type Error = PlainTextError;
-    type Future = Box<dyn Future<Item = Self::Output, Error = Self::Error> + Send>;
+    type Future = Pin<Box<dyn Future<Output = Result<Self::Output, Self::Error>> + Send>>;
```
twittner:
```suggestion
    type Future = BoxFuture<'static, Result<Self::Output, Self::Error>>;
```

(assuming an `use futures::future::BoxFuture;`)

protocols/plaintext/src/lib.rs
```diff
@@ -108,144 +121,138 @@ impl UpgradeInfo for PlainText2Config {
...
 
 impl<C> OutboundUpgrade<C> for PlainText2Config
 where
-    C: AsyncRead + AsyncWrite + Send + 'static
+    C: AsyncRead + AsyncWrite + Send + Unpin + 'static
 {
     type Output = (PeerId, PlainTextOutput<Negotiated<C>>);
     type Error = PlainTextError;
-    type Future = Box<dyn Future<Item = Self::Output, Error = Self::Error> + Send>;
+    type Future = Pin<Box<dyn Future<Output = Result<Self::Output, Self::Error>> + Send>>;
```
twittner: like above
```suggestion
    type Future = BoxFuture<'static, Result<Self::Output, Self::Error>>;
```

protocols/secio/src/codec/mod.rs
```diff
@@ -103,7 +103,7 @@ impl Hmac {
 }
 
 /// Takes control of `socket`. Returns an object that implements `future::Sink` and
-/// `future::Stream`. The `Stream` and `Sink` produce and accept `BytesMut` objects.
+/// `future::Stream`. The `Stream` and `Sink` produce and accept `Vec<u8>` objects.
```
twittner: This is obsolete, no?

mxinden: @twittner this change is not within the *plaintext* but within the *secio* module. As far as I can tell `FullCodec` (`EncoderMiddleware` and `DecoderMiddleware`) implement the `Stream` and `Sink` trait with `Vec<u8> and not `BytesMut`. Does that make sense?

twittner: Ah, sorry. I did not notice that this was in `secio`.

protocols/plaintext/src/lib.rs
mxinden: Ah, cool. Wasn't aware of this type alias. Thanks.

PR #1308, opened 2019-11-15
Update libp2p-uds to futures 0.3

tomaka:
Switched from `romio` (which is unmaintained and is still using futures-preview) to `async-std`.

### DIFF
--- a/transports/uds/src/lib.rs
+++ b/transports/uds/src/lib.rs
@@ -45,7 +45,7 @@
 #![cfg(all(unix, not(any(target_os = "emscripten", target_os = "unknown"))))]
 
 use async_std::os::unix::net::{UnixListener, UnixStream};
-use futures::{prelude::*, ready, future::Ready};
+use futures::{prelude::*, future::Ready};
 use futures::stream::Stream;
 use libp2p_core::{
     Transport,
@@ -53,7 +53,7 @@ use libp2p_core::{
     transport::{ListenerEvent, TransportError}
 };
 use log::debug;
-use std::{io, path::PathBuf, pin::Pin, task::Context, task::Poll};
+use std::{io, path::PathBuf, pin::Pin};
 
 /// Represents the configuration for a Unix domain sockets transport capability for libp2p.
 ///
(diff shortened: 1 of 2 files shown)


transports/uds/src/lib.rs
```diff
@@ -103,7 +114,7 @@ impl Transport for UdsConfig {
     fn dial(self, addr: Multiaddr) -> Result<Self::Dial, TransportError<Self::Error>> {
         if let Ok(path) = multiaddr_to_path(&addr) {
             debug!("Dialing {}", addr);
-            Ok(UnixStream::connect(&path))
+            Ok(Box::pin(async move { UnixStream::connect(&path).await }))
```
tomaka: The `Future` returned by `connect` unfortunately borrows `path`. This `async` block is here in order to bypass this problem.

transports/uds/src/lib.rs
```diff
@@ -74,27 +74,38 @@ impl UdsConfig {
 impl Transport for UdsConfig {
     type Output = UnixStream;
     type Error = io::Error;
-    type Listener = ListenerStream<romio::uds::Incoming>;
+    type Listener = Pin<Box<dyn Stream<Item = Result<ListenerEvent<Self::ListenerUpgrade>, Self::Error>> + Send>>;
     type ListenerUpgrade = Ready<Result<Self::Output, io::Error>>;
-    type Dial = romio::uds::ConnectFuture;
+    type Dial = Pin<Box<dyn Future<Output = Result<Self::Output, Self::Error>> + Send>>;
```
twittner:
```suggestion
    type Dial = BoxFuture<'static, Result<Self::Output, Self::Error>>;
```

transports/uds/src/lib.rs
```diff
@@ -74,27 +74,38 @@ impl UdsConfig {
 impl Transport for UdsConfig {
     type Output = UnixStream;
     type Error = io::Error;
-    type Listener = ListenerStream<romio::uds::Incoming>;
+    type Listener = Pin<Box<dyn Stream<Item = Result<ListenerEvent<Self::ListenerUpgrade>, Self::Error>> + Send>>;
```
twittner:
```suggestion
    type Listener = BoxStream<'static, Result<ListenerEvent<Self::ListenerUpgrade>, Self::Error>>;
```

PR #1309, opened 2019-11-17
protocols/plaintext: Adapt protobuf generation to consolidated scripts

mxinden:
https://github.com/libp2p/rust-libp2p/pull/1275 missed to update the Protobuf generation in the `plaintext` transport module. 

We can merge this before or after the `stable-futures` branch, as you like.


tomaka:
Let's merge this now that `stable-futures` is in.

tomaka:
Sorry I think I screwed up while resolving the merge conflict. There's probably an import missing.

mxinden:
@tomaka can you take another look?

PR #1316, opened 2019-11-21
Bump dependencies, respecting semver

DemiMarie-parity:
This is done using `cargo update && cargo upgrade --all --to-lockfile`.


tomaka:
Since this heavily conflicts with `stable-futures`, I'd prefer to do this after `stable-futures` has been merged.

PR #1317, opened 2019-11-22
Fix examples and update core-derive.

twittner:


Cargo.toml
```diff
@@ -45,9 +45,9 @@ libp2p-tcp = { version = "0.13.0", path = "transports/tcp" }
 libp2p-websocket = { version = "0.13.0", path = "transports/websocket", optional = true }
 
 [dev-dependencies]
+anyhow = "1.0"
```
tomaka: Since this is an opinionated way to handle errors, I'd be in favour of not using it in examples.

PR #1319, opened 2019-11-25
Add support for PINGs and PONGs to websockets.

twittner:
`websocket::framed::Connection` (formerly `websocket::framed::BytesConnection`) now uses the more structured types `IncomingData` and `OutgoingData` which mirror the data types in soketto (which are not exposed). This allows adding `Connection::send_ping` and `Connection::send_pong` methods.

The toplevel websocket transport defines `websocket::BytesConnection` as a wrapper around `websocket::framed::Connection` and handles only binary data.

Depends on https://github.com/paritytech/soketto/pull/9.

### DIFF
--- a/transports/websocket/src/lib.rs
+++ b/transports/websocket/src/lib.rs
@@ -24,9 +24,10 @@ pub mod error;
 pub mod framed;
 pub mod tls;
 
+use bytes::BytesMut;
 use error::Error;
-use framed::BytesConnection;
-use futures::{future::BoxFuture, prelude::*, stream::BoxStream};
+use framed::Connection;
+use futures::{future::BoxFuture, prelude::*, stream::BoxStream, ready};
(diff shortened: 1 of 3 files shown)


transports/websocket/src/lib.rs
```diff
@@ -118,78 +120,94 @@ where
...
+            match ready!(self.0.try_poll_next_unpin(cx)?) {
+                Some(framed::IncomingData::Binary(d)) => return Poll::Ready(Some(Ok(d))),
+                None => return Poll::Ready(None),
+                _ => {}
+            }
+        }
+    }
+}
+
+impl<T> Sink<BytesMut> for BytesConnection<T>
```
tomaka: As a basic comment (not requesting any change), we could also provide multiple implementations of the `Sink` but with different item types (text, ping, bytes).

PR #1320, opened 2019-11-25
protocols/kad: Fix tests

mxinden:
This pull request adjusts the Kademlia tests to the *stable future* refactoring.

(+ 2 minor fixes in noise and mdns)

### DIFF
--- a/protocols/kad/src/behaviour/test.rs
+++ b/protocols/kad/src/behaviour/test.rs
@@ -137,11 +137,9 @@ fn bootstrap() {
                         .map(|e| e.node.key.preimage().clone())
                         .collect::<HashSet<_>>();
                     assert_eq!(expected_known, known);
-                    return Poll::Ready(());
+                    return;
                 }
             }
-
-            Poll::Pending
         });
     }
 
--- a/protocols/kad/src/behaviour/test.rs
+++ b/protocols/kad/src/behaviour/test.rs
@@ -25,6 +25,8 @@ use super::*;
 use crate::K_VALUE;
 use crate::kbucket::Distance;
 use crate::record::store::MemoryStore;
+use futures::prelude::*;
+// TODO: still needed?
 use futures::future;
 use libp2p_core::{
     PeerId,
@@ -42,7 +44,6 @@ use libp2p_yamux as yamux;
 use quickcheck::*;
 use rand::{Rng, random, thread_rng};
 use std::{collections::{HashSet, HashMap}, io, num::NonZeroUsize, u64};
-use tokio::runtime::current_thread;
 use multihash::{Multihash, Hash::SHA2256};
 
 type TestSwarm = Swarm<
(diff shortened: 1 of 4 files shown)


protocols/kad/src/behaviour/test.rs
```diff
@@ -120,27 +123,26 @@ fn bootstrap() {
...
-                                    .map(|e| e.node.key.preimage().clone())
-                                    .collect::<HashSet<_>>();
-                                assert_eq!(expected_known, known);
-                                return Ok(Async::Ready(()));
-                            }
-                            Async::Ready(_) => (),
-                            Async::NotReady => break,
-                        }
-                    }
+        block_on(async move {
```
mxinden: I find the test easier to understand, when we first poll all swarms expect the first one, giving them a chance to make progress, and then check whether the first swarm has bootstrapped successfully.

What do you think @romanb? I would apply the same pattern to the similar tests below in case people are in favor of this style.

tomaka:
I'm not a fan of using `poll_fn` and adding `// TODO: refactor`.
It shouldn't be hard to use async/await properly here.

mxinden:
@tomaka see https://github.com/libp2p/rust-libp2p/pull/1320#discussion_r350160204. I would like to get your input on the style before I touch all following tests and remove their `poll_fn` and `TODO's`.

protocols/kad/src/behaviour/test.rs
romanb: I'm generally OK with it, though I don't find it simpler personally, but it seems to be semantically equivalent (i.e. each swarm is polled until pending or the expected result appears). However, I have two small comments:

  1. Is the `'outer` loop label necessary (i.e. why not `return` as before)?
  2. If I see it correctly, then while all swarms are "not ready" / "pending" (e.g. waiting for I/O), this variant seems to enter a "busy loop", not yielding control of the thread to do other useful things meanwhile. While this probably doesn't matter much especially since the tests use the `MemoryTransport`, I still find it cleaner to yield the thread when all swarms are pending.

romanb: Another small difference to the previous code seems to be that this version does not fail the test if any other swarm besides the first produces a `BootstrapResult` (which would naturally be incorrect and unexpected). Again, this may not be important, but it is a small difference.

mxinden:
I went for the `poll_fn` approach for the following reasons:

- Least intrusive, easiest to review.

- Using an `async` closure with `now_or_never` and `pending!` (to not enter a busy loop)  seems to defeat the purpose. Alternatively awaiting each swarm individually via a `FutureUnordered` introduces too much overhead having to synchronize access to shared data structures.

@tomaka @romanb let me know what you think.

tomaka:
Could you rebase or merge the latest `stable-futures`?

protocols/kad/src/jobs.rs
```diff
@@ -392,23 +397,26 @@ mod tests {
...
-            for r in store.provided().map(|r| r.into_owned()).collect::<Vec<_>>() {
-                if !r.is_expired(now) {
-                    assert_eq!(job.poll(&mut store, now), Poll::Ready(r));
-                    assert!(job.is_running());
+
+            block_on(poll_fn(|ctx| {
+                // Polling with an instant beyond the deadline for the next run
+                // is guaranteed to run the job, without the job needing to poll the `Delay`
+                // and thus without needing to run `poll` in the context of a task
+                // for testing purposes.
```
romanb: Just like the other case, this comment should now be removed.

protocols/kad/src/jobs.rs
```diff
@@ -362,20 +363,24 @@ mod tests {
...
-            for r in store.records().map(|r| r.into_owned()).collect::<Vec<_>>() {
-                if !r.is_expired(now) {
-                    assert_eq!(job.poll(&mut store, now), Poll::Ready(r));
-                    assert!(job.is_running());
+
+            block_on(poll_fn(|ctx| {
+                // Polling with an instant beyond the deadline for the next run
+                // is guaranteed to run the job, without the job needing to poll the `Delay`
+                // and thus without needing to run `poll` in the context of a task
+                // for testing purposes.
```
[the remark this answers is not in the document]
mxinden: Sorry for missing that, thanks.

mxinden:
@romanb could you take another look?

PR #1326, opened 2019-11-29
Fix identify not compiling for wasm

tomaka:
The WASM CI build should be green again now.

### DIFF
--- a/protocols/identify/Cargo.toml
+++ b/protocols/identify/Cargo.toml
@@ -10,7 +10,6 @@ keywords = ["peer-to-peer", "libp2p", "networking"]
 categories = ["network-programming", "asynchronous"]
 
 [dependencies]
-async-std = "1.0"
 bytes = "0.4"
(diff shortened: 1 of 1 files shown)


protocols/identify/Cargo.toml
```diff
@@ -24,6 +23,7 @@ wasm-timer = "0.2"
 unsigned-varint = { version = "0.2.3", features = ["futures-codec"] }
 
 [dev-dependencies]
+async-std = "1.0"
```
twittner: We should probably import `async-std` everywhere as `1` instead of `1.x`. According to their [stability policy](https://book.async.rs/overview/stability-guarantees.html) changes in minor and patch versions are backwards compatible.

tomaka: `async-std = "1"` and `async-std = "1.0"` should do exactly the same thing. Cargo will pull `async-std 1.1` even if we say `1.0`.

PR #1331, opened 2019-12-02
Update multistream-select for stable futures

tomaka:
Opening as a draft to show that I'm working on it. No totally finished yet, but close.

The diff right now is +250 -1205, but that's normal. I refactored most of the code to use async/await rather than polling.


tomaka:
Progress on this PR had stopped because of issues with pinning that #1384 should solve.

thomaseizinger:
#1384 is merged by now. What are the plans with this PR?

tomaka:
I haven't touched it in the past weeks, but it turned out to be more complicated than actually thought, as using async/await forces `Send + 'static` everywhere, which means that we would have to add these bounds to lots of places.

PR #1332, opened 2019-12-03
Update to current soketto `develop`.

twittner:
See https://github.com/paritytech/soketto/pull/11 for details.

### DIFF
--- a/transports/websocket/src/lib.rs
+++ b/transports/websocket/src/lib.rs
@@ -146,10 +146,12 @@ where
 
     fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context) -> Poll<Option<Self::Item>> {
         loop {
-            match ready!(self.0.try_poll_next_unpin(cx)?) {
-                Some(framed::IncomingData::Binary(d)) => return Poll::Ready(Some(Ok(d))),
-                None => return Poll::Ready(None),
-                _ => {}
+            if let Some(item) = ready!(self.0.try_poll_next_unpin(cx)?) {
+                if item.is_binary() {
+                    return Poll::Ready(Some(Ok(BytesMut::from(item.as_ref()))))
+                }
+            } else {
+                return Poll::Ready(None)
             }
         }
     }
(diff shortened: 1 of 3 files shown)


transports/websocket/src/lib.rs
```diff
@@ -146,10 +146,12 @@ where
...
     fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context) -> Poll<Option<Self::Item>> {
         loop {
-            match ready!(self.0.try_poll_next_unpin(cx)?) {
-                Some(framed::IncomingData::Binary(d)) => return Poll::Ready(Some(Ok(d))),
-                None => return Poll::Ready(None),
-                _ => {}
+            if let Some(item) = ready!(self.0.try_poll_next_unpin(cx)?) {
+                if item.is_binary() {
+                    return Poll::Ready(Some(Ok(BytesMut::from(item.as_ref()))))
+                }
```
tomaka: Can be maybe do something here so that sending a text frame doesn't silently vanish into the abyss?

twittner: Yes. The idea was that a `BytesConnection` would only allow binary data but of course we can forward text as binary too.

tomaka: What I mean is: if someone connects to some server that uses libp2p-websocket and sends out text frames, it will be silently ignored and they won't understand why.

I don't know whether we should add a log entry or treat the text as if it was binary, but IMO either of these is better than just discarding the text frames.

twittner: Did you see https://github.com/libp2p/rust-libp2p/pull/1332/commits/c43b5f8bc303d36a038af03f64bf25ae31a40f26?

tomaka: > Did you see c43b5f8?

From your answer I'm going to guess that `is_data` covers both binary and text.

twittner: Yes it does.

PR #1337, opened 2019-12-09
secio: Back to 4-byte BE length prefix.

twittner:
The secio spec states that each frame must be prefixed with a 32-bit big endian length prefix so we can not use an unsigned varint here.

### DIFF
--- a/protocols/secio/src/codec/mod.rs
+++ b/protocols/secio/src/codec/mod.rs
@@ -23,16 +23,17 @@
 
 mod decode;
 mod encode;
+mod len_prefix;
 
 use aes_ctr::stream_cipher;
 use crate::algo_support::Digest;
 use decode::DecoderMiddleware;
 use encode::EncoderMiddleware;
-use futures::{prelude::*, stream::BoxStream};
+use futures::prelude::*;
 use hmac::{self, Mac};
-use quicksink::Action;
 use sha2::{Sha256, Sha512};
-use std::{fmt, io, pin::Pin, task::{Context, Poll}};
+
+pub use len_prefix::LenPrefixCodec;
 
 /// Type returned by `full_codec`.
 pub type FullCodec<S> = DecoderMiddleware<EncoderMiddleware<LenPrefixCodec<S>>>;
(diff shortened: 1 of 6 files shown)


protocols/secio/src/codec/mod.rs
```diff
@@ -108,38 +108,129 @@ impl Hmac {
...
+        let stream = futures::stream::unfold(r, |mut r| async {
+            let mut len = [0; 4];
+            if let Err(e) = r.read_exact(&mut len).await {
+                if e.kind() == io::ErrorKind::UnexpectedEof {
+                    return None
+                }
+                return Some((Err(e), r))
+            }
+            let n = u32::from_be_bytes(len) as usize;
+            let mut v = vec![0; n as usize];
```
twittner: Maybe we should limit the vector size further than 2^32?

protocols/secio/src/codec/mod.rs
```diff
@@ -108,38 +108,128 @@ impl Hmac {
...
-    S: AsyncRead + AsyncWrite + Unpin,
+    S: AsyncRead + AsyncWrite + Unpin + Send + 'static
 {
     let encoder = EncoderMiddleware::new(socket, cipher_encoding, encoding_hmac);
     DecoderMiddleware::new(encoder, cipher_decoder, decoding_hmac, remote_nonce)
 }
 
+
+/// `Stream` & `Sink` that reads and writes a length prefix in front of the actual data.
+pub struct LenPrefixCodec<T> {
```
tomaka: Could we move this to a different module?

protocols/secio/src/lib.rs
```diff
@@ -95,7 +96,8 @@ impl SecioConfig {
             key: kp,
             agreements_prop: None,
             ciphers_prop: None,
-            digests_prop: None
+            digests_prop: None,
+            max_frame_len: 8 * 1024 * 1024
```
tomaka: cc'ing https://github.com/libp2p/specs/pull/236 for reference

PR #1338, opened 2019-12-09
Fix possible incorrect return value from <LengthDelimitedReader as io::Write>::write().

romanb:
Due to not taking into account `buf.len()` when computing `written`, it may be incorrectly judged less than `buf.len()`, which can result in duplicate data being written as well as potentially arithmetic overflow. This is essentially a backport of what was fixed for the `stable-futures` branch [here](https://github.com/libp2p/rust-libp2p/pull/1335#discussion_r355373145). However, instead of removing the optimisation of writing out the entire data in one go, as done there, this just corrects the oversight of not taking `buf.len()` into account when computing `written` in the `io::Write` impl of `LengthDelimitedReader`. The rest of the diff just adds some debug assertions for further clarity.

I'm not opposed to removing this kind of optimisation entirely, as it is evidently non-trivial to get right, though I think it should then also be removed from the `io::Write` impl of `Negotiated`, which does the same thing in the context of an owned buffer (and can be used to compare the implementations).

### DIFF
--- a/misc/multistream-select/src/length_delimited.rs
+++ b/misc/multistream-select/src/length_delimited.rs
@@ -328,17 +328,17 @@ where
             let n = self.inner.write_buffer.len();
             self.inner.write_buffer.extend_from_slice(buf);
             let result = self.inner.poll_write_buffer();
-            let written = n - self.inner.write_buffer.len();
+            let written = n + buf.len() - self.inner.write_buffer.len();
             if written == 0 {
+                self.inner.write_buffer.split_off(n); // Never grow the buffer.
                 if let Err(e) = result {
                     return Err(e)
                 }
                 return Err(io::ErrorKind::WouldBlock.into())
             }
             if written < buf.len() {
-                if self.inner.write_buffer.len() > n {
-                    self.inner.write_buffer.split_off(n); // Never grow the buffer.
-                }
+                debug_assert!(self.inner.write_buffer.len() > n);
+                self.inner.write_buffer.split_off(n); // Never grow the buffer.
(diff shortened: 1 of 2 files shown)


twittner:
> I'm not opposed to removing this kind of optimisation entirely, as it is evidently non-trivial to get right, though I think it should then also be removed from the `io::Write` impl of `Negotiated`, which does the same thing in the context of an owned buffer (and can be used to compare the implementations).

That would be my preferred outcome. I had a very close look at `Negotiated` as another suspect of the data issue we encountered before concluding that it is correct and the difficulty of understanding it just like the code in `length_delimited` leads me to the conclusion that it does not carry its weight. So I would strongly prefer to remove both instances.

tomaka:
I would indeed push towards what #1331 is doing: rewrite the code to be more simple and use async/await. The code is **much** easier to read afterwards.

tomaka:
Assuming that we merge `stable-futures` soon, I'm not sure that this backport is necessary?

romanb:
I removed the optimisation now from both `LengthDelimitedReader` and `Negotiated`, therbey updating the associated test.

> Assuming that we merge `stable-futures` soon, I'm not sure that this backport is necessary?

I leave that up to you. Just in case this issue is deemed severe enough to warrant a patch release before `stable-futures` is ready, this PR is available. I will also prepare a PR for `stable-futures` that similarly simplifies `Negotiated` and updates the test there.

twittner:
> Assuming that we merge `stable-futures` soon, I'm not sure that this backport is necessary?

I recommend a patch release.

PR #1342, opened 2019-12-10
Cleanups in libp2p-core in stable-futures branch

tomaka:
Removes a few left-overs, and lots of `Unpin` trait bounds replaced with `pin-project`.

### DIFF
--- a/core/src/either.rs
+++ b/core/src/either.rs
@@ -20,7 +20,8 @@
 
 use crate::{muxing::StreamMuxer, ProtocolName, transport::ListenerEvent};
 use futures::prelude::*;
-use std::{fmt, io::{Error as IoError, Read, Write}, pin::Pin, task::Context, task::Poll};
+use pin_project::{pin_project, project};
+use std::{fmt, io::{Error as IoError}, pin::Pin, task::Context, task::Poll};
 
 #[derive(Debug, Copy, Clone)]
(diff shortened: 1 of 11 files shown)


tomaka:
Note that, unless we do some design refactoring, the substream upgrade infrastructure has to require `Unpin` for the substream type, as we have to read/write to them, then give ownership to them to the upgrade function.

core/src/either.rs
```diff
@@ -56,99 +57,75 @@ where
...
-{
-    fn read(&mut self, buf: &mut [u8]) -> Result<usize, IoError> {
-        match self {
-            EitherOutput::First(a) => a.read(buf),
-            EitherOutput::Second(b) => b.read(buf),
+    #[project]
+    fn poll_read(self: Pin<&mut Self>, cx: &mut Context, buf: &mut [u8]) -> Poll<Result<usize, IoError>> {
+        #[project]
+        match self.project() {
+            EitherOutput::First(a) => AsyncRead::poll_read(a, cx, buf),
```
twittner: Here and elsewhere you switched to UFCS style. What is the motivation for this change and what is the advantage here over writing `a.poll_read(cx, buf)`?

tomaka: It's just a habit from early days of `futures-preview` where it wasn't possible to write `a.poll_read`.

PR #1345, opened 2019-12-10
Use `wasm-bindgen` rand feature to support wasm

expenses:
Without this feature, `thread_rng()` etc panics on `wasm32-unknown-unknown`.


tomaka:
We don't need to do that. Only the final binary has to activate the feature.

PR #1347, opened 2019-12-11
Fix noise/io.rs not flushing the underlying stream

tomaka:
Fixes `poll_flush` not calling the underlying `poll_flush` function.
Additionally, fixed `poll_close` to also flush, as required by the function.

Should be backported to master.

### DIFF
--- a/protocols/noise/src/io.rs
+++ b/protocols/noise/src/io.rs
@@ -424,7 +424,7 @@ impl<T: AsyncWrite + Unpin> AsyncWrite for NoiseOutput<T> {
         mut self: Pin<&mut Self>,
         cx: &mut Context<'_>,
     ) -> Poll<Result<(), std::io::Error>>{
-        ready!(Pin::new(&mut *self).poll_flush(cx))?;
+        ready!(self.as_mut().poll_flush(cx))?;
         Pin::new(&mut self.io).poll_close(cx)
     }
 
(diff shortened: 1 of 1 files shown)


protocols/noise/src/io.rs
```diff
@@ -425,6 +424,7 @@ impl<T: AsyncWrite + Unpin> AsyncWrite for NoiseOutput<T> {
         mut self: Pin<&mut Self>,
         cx: &mut Context<'_>,
     ) -> Poll<Result<(), std::io::Error>>{
+        ready!(Pin::new(&mut *self).poll_flush(cx))?;
```
twittner:
```suggestion
        ready!(self.as_mut().poll_flush(cx))?;
```

tomaka:
> +1 how did you catch this?

Contrary to libp2p-tcp, libp2p-websockets relies on `flush` being called. Connecting to a WebSockets server through Noise wasn't working and we investigated.

tomaka:
> I'm not aware of such a contract for AsyncWrite::poll_close, only for Sink::poll_close - do you have a link?

Hmm, it looks like this is specific to tokio:

- https://docs.rs/tokio-io/0.2.0-alpha.6/tokio_io/trait.AsyncWrite.html#tymethod.poll_shutdown
- https://docs.rs/tokio-io/0.1.12/tokio_io/trait.AsyncWrite.html#tymethod.shutdown

I would be on an overlook in async-std.

romanb:
I see, thanks. Taking `poll_close` to be the semantic equivalent to `shutdown` makes sense, I guess.

PR #1350, opened 2019-12-12
Bring back phantom types to yamux upgrade outputs.

twittner:


muxers/yamux/src/lib.rs
```diff
@@ -49,35 +49,41 @@ struct Inner<S> {
 #[derive(Debug)]
 pub struct OpenSubstreamToken(());
 
-impl Yamux<Incoming> {
+impl<C> Yamux<Incoming<C>>
+where
+    C: AsyncRead + AsyncWrite + Send + Unpin + 'static
```
tomaka: I think we can remove these bounds?

twittner: No, these are needed for the calls to `Yamux::new`, `into_stream` and `boxed`.

PR #1353, opened 2019-12-18
Several changes.

twittner:
- ~Pin `futures_codec` to version 0.3.3 as later versions require at least `bytes-0.5` which he have not upgraded to yet.~ Update to `bytes` v0.5 except for `multiaddr`.
- Replace `futures::executor::block_on` with `async_std::task::block_on` where `async-std` is already a dependency to work around an issue with `park`/`unpark` behaviour.
- Use the published versions of `quicksink`, `soketto` and `yamux`.
- Use `futures-timer` version 2 in `core` which removes the last (transitive) dependency to `futures-preview`.

Depends on https://github.com/paritytech/unsigned-varint/pull/23.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,7 +14,7 @@ default = ["secp256k1", "libp2p-websocket"]
 secp256k1 = ["libp2p-core/secp256k1", "libp2p-secio/secp256k1"]
 
 [dependencies]
-bytes = "0.4"
+bytes = "0.5"
 futures = "0.3.1"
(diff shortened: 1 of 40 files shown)


mxinden:
> to work around an issue with park/unpark behaviour.

@twittner I am curious. Can you add a link to the issue?

romanb:
> > to work around an issue with park/unpark behaviour.
> 
> @twittner I am curious. Can you add a link to the issue?

https://github.com/rust-lang/futures-rs/pull/2010

DemiMarie-parity:
@tomaka can you merge this?

PR #1354, opened 2019-12-19
Added Onion3 support to multiaddr

sdbondi:
Added support for `/onion3/xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx:{port}`
~~Extra: Added conversions from SocketAddr to multiaddr~~


misc/multiaddr/src/lib.rs
```diff
@@ -269,6 +270,31 @@ impl<'a> From<Protocol<'a>> for Multiaddr {
...
+            SocketAddr::V4(a) => a.into(),
+            SocketAddr::V6(a) => a.into()
+        }
+    }
+}
+
+impl From<SocketAddrV4> for Multiaddr {
+    fn from(v: SocketAddrV4) -> Multiaddr {
+        let mut addr: Multiaddr = v.ip().clone().into();
+        addr.push(Protocol::Tcp(v.port()));
```
twittner: Might as well be `Protocol::Udp`.

sdbondi: Ah of course - so this conversion won't work because we don't have enough information - ok I'll remove it

misc/multiaddr/src/protocol.rs
```diff
@@ -75,6 +76,7 @@ pub enum Protocol<'a> {
     /// Contains the "port" to contact. Similar to TCP or UDP, 0 means "assign me a port".
     Memory(u64),
     Onion(Cow<'a, [u8; 10]>, u16),
+    Onion3(Cow<'a, [u8]>, u16),
```
twittner: I wonder if we should use `Onion3(Cow<'a, [u8; 35]>, u16)` instead? It would more than double the size of `Protocol` and require custom impls of `PartialEq` and `Debug` but the type would be more descriptive and we would not require a heap allocation when parsing `Protocol::Onion3`. What do you think?

sdbondi: Hey, sorry was AFK - Agree with saving an allocation, I'll take another shot because I think I was rushed and just went with the easier option :)

sdbondi: I'll make a custom type `Onion3Addr(([u8; 35], u16))` and implement `PartialEq` and `Debug`

sdbondi: Done - let me know what you think :)

twittner:
Could you please undo the `cargo fmt` changes. Reviewing is hard with all those formatting changes.

sdbondi:
@twittner Cool will do - I will say though that I think this project would benefit from having a rustfmt.toml to keep things consistent across all code files

burdges:
Tor hidden service address I presume?

sdbondi:
@burdges I believe that an Onion v3 address is simply an address using a hash of a Curve25519 public key as opposed to v1 addresses which have much less entropy - they achieve the same thing

burdges:
Tor's newer hidden services use the whole public key.

misc/multiaddr/tests/lib.rs
```diff
@@ -104,7 +104,12 @@ impl Arbitrary for Proto {
                 let mut a = [0; 10];
                 g.fill(&mut a);
                 Proto(Onion(Cow::Owned(a), g.gen()))
-            }
+            },
+            23 => {
+                let mut a = Vec::with_capacity(35);
```
twittner:
```suggestion
                let mut a = [0; 35];
```

misc/multiaddr/src/onion_addr.rs
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    /// Consume this instance and create an owned version containing the same address
+    pub fn into_owned<'b>(self) -> Onion3Addr<'b> {
+        Self(Cow::Owned(self.0.into_owned()), self.1)
+    }
+}
+
+impl PartialEq for Onion3Addr<'_> {
+    fn eq(&self, other: &Self) -> bool {
```
twittner: How about:

```rust
impl PartialEq for Onion3Addr<'_> {
    fn eq(&self, other: &Self) -> bool {
        self.1 == other.1 && self.0[..] == other.0[..]
    }
}
```

misc/multiaddr/tests/lib.rs
```diff
@@ -208,35 +213,54 @@ fn construct_success() {
...
+
+    ma_valid(
+        "/onion/aaimaq4ygg2iegci:80",
+        "BC030010C0439831B48218480050",
+        vec![Onion(Cow::Owned([0, 16, 192, 67, 152, 49, 180, 130, 24, 72]), 80)],
+    );
+    ma_valid(
+        "/onion3/vww6ybal4bd7szmgncyruucpgfkqahzddi37ktceo3ah7ngmcopnpyyd:1234",
+        "BD03ADADEC040BE047F9658668B11A504F3155001F231A37F54C4476C07FB4CC139ED7E30304D2",
+        vec![Onion3((vec![173, 173, 236, 4, 11, 224, 71, 249, 101, 134, 104, 177, 26, 80, 79, 49, 85, 0, 31, 35, 26, 55, 245, 76, 68, 118, 192, 127, 180, 204, 19, 158, 215, 227, 3].as_slice(), 1234).into())],
```
twittner:
```suggestion
        vec![Onion3(([173, 173, 236, 4, 11, 224, 71, 249, 101, 134, 104, 177, 26, 80, 79, 49, 85, 0, 31, 35, 26, 55, 245, 76, 68, 118, 192, 127, 180, 204, 19, 158, 215, 227, 3], 1234).into())],
```

misc/multiaddr/src/onion_addr.rs
```diff
@@ -0,0 +1,82 @@
...
+
+impl<'a> From<(&'a [u8; 35], u16)> for Onion3Addr<'a> {
+    fn from(parts: (&'a [u8; 35], u16)) -> Self {
+        Self(Cow::Borrowed(parts.0), parts.1)
+    }
+}
+
+impl<'a> From<(&'a [u8], u16)> for Onion3Addr<'a> {
+    fn from(parts: (&'a [u8], u16)) -> Self {
+        Self(Cow::Borrowed(array_ref!(parts.0, 0, 35)), parts.1)
```
twittner: I would not want to add this `From` impl as the slice may easily not be 35 bytes long which causes a panic. Maybe just remove it or make it a `TryFrom` if you really think it would be necessary.

misc/multiaddr/src/onion_addr.rs
```diff
@@ -0,0 +1,82 @@
...
+    }
+}
+
+impl From<(Vec<u8>, u16)> for Onion3Addr<'_> {
+    fn from(parts: (Vec<u8>, u16)) -> Self {
+        let slice = array_ref!(parts.0, 0, 35);
```
twittner: Like with slices above, I would not want to have a `From` for arbitrary `Vec<u8>`s.

misc/multiaddr/src/onion_addr.rs
```diff
@@ -0,0 +1,82 @@
...
+        Self(Cow::Owned(*slice), parts.1)
+    }
+}
+
+impl Debug for Onion3Addr<'_> {
+    fn fmt(&self, f: &mut Formatter<'_>) -> Result<(), fmt::Error> {
+        let bytes_str = {
+            let s = self.0[..].iter().map(ToString::to_string).collect::<Vec<_>>();
+            format!("[{}]", s.join(", "))
+        };
```
twittner: This could be simplified to

```rust
let bytes_str = format!("{:?}", &self.0[..]);
```
or maybe with hex output even:

```rust
let bytes_str = format!("{:02x?}", &self.0[..]);
```

misc/multiaddr/src/onion_addr.rs
sdbondi: 🤦‍♂ Thanks

misc/multiaddr/src/onion_addr.rs
sdbondi: You are correct, they are not needed with your other suggestions in place - thanks

misc/multiaddr/src/onion_addr.rs
sdbondi: Ah, the "size not known at compile time" error got me - thanks I've learnt something

misc/multiaddr/src/protocol.rs
```diff
@@ -15,6 +15,8 @@ use std::{
 };
 use unsigned_varint::{encode, decode};
 
+pub use crate::onion_addr::Onion3Addr;
```
twittner: In here you can just `use crate::onion_addr::Onion3Addr`, the `pub use` must go into `lib.rs` for `Onion3Addr` to be exported. The `onion_addr` module itself is private.

misc/multiaddr/src/onion_addr.rs
```diff
@@ -0,0 +1,54 @@
...
+        self.0.as_ref()
+    }
+
+    /// Return the port
+    pub fn port(&self) -> u16 {
+        self.1
+    }
+
+    /// Consume this instance and create an owned version containing the same address
+    pub fn into_owned<'b>(self) -> Onion3Addr<'b> {
```
twittner: To align this with `Protocol`, could you change the name into `acquire`? `Onion3Addr::into_owned` is the better name but we can not change `Protocol::acquire` or we would break backwards compatibility and I think it would be nice to have consistent naming in this crate.

sdbondi: Sure thing 👍

sdbondi:
@twittner Appreciate all the help 🥇

PR #1355, opened 2019-12-20
Add protocol string to Error::UnknownProtocolString

ackintosh:
Added protocol string to `Error::UnknownProtocolString` so that we can review what the string causing the error.


tomaka:
Let's merge after #1328

Issue #1076, opened 2019-04-20
Add a transport that listens on either WebSockets or non-WebSockets


tomaka:
In `libp2p-websockets`, we should add a transport that, if the input is not HTTP, falls back to using the underlying transport instead of erroring.
This can also be implemented by adding an option in the existing `WsConfig`.

Might depend on #849.

Now that `listen_addr` can return multiple listened addresses, this transport can return both the "ws" and non-"ws" multiaddress.
For example, listening on `/ip4/0.0.0.0/tcp/80` would produce the addresses `/ip4/127.0.0.1/tcp/80`, `/ip4/127.0.0.1/tcp/80/ws`, `/ip4/192.168.1.12/tcp/80`, and `/ip4/192.168.1.12/tcp/80/ws` (and possibly more).

tomaka:
cc #1360

tomaka:
I agree that this was not a clean feature, but I was willing to do it anyway. If both @romanb and @twittner agree that this is a bad idea, then let's close this.

Another argument about closing is: in the future, we probably want to use QUIC as the primary transport mechanism rather than TCP, in which case this feature would become useless.

PR #1360, opened 2019-12-23
Allow falling back to the inner transport if the websocket handshaking fails.

expenses:
@tomaka @twittner 

It would be super useful to be able to switch to the inner transport when a websocket handshake fails. This would allow websocket connections to be used as a default and falling back to TCP when it's not supported.

This PR changes the websocket `Error::Handshake` type to contain an `Option<Fallback>`, which is a type implementing `AsyncRead` and `AsyncWrite`. This type is buffered to re-use bytes from the handshake server.

Required https://github.com/paritytech/soketto/pull/16 to be merged first.


transports/websocket/src/framed.rs
```diff
@@ -190,10 +190,18 @@ where
...
                         let ws_key = {
-                            let request = server.receive_request()
-                                .map_err(|e| Error::Handshake(Box::new(e)))
-                                .await?;
-                            request.into_key()
+                            let response = server.receive_request()
+                                .await;
+                            match response {
+                                Ok(request) => request.into_key(),
+                                Err(e) => {
```
twittner: Here one should probably inspect the error a bit more (cf. [`soketto::handshake::Error`](https://docs.rs/soketto/0.3.1/soketto/handshake/enum.Error.html)). I do not think it makes sense to fall back in case of an I/O error for example. Instead a fallback should maybe only be attempted in case of an [HTTP parsing error](https://docs.rs/soketto/0.3.1/soketto/handshake/enum.Error.html#variant.Http).

expenses:
Ready to be reviewed, but I think it requires https://github.com/libp2p/rust-libp2p/pull/1353 to be merged first to fix the whole repo test suite.

transports/websocket/src/framed.rs
```diff
@@ -191,9 +284,17 @@ where
...
                             let request = server.receive_request()
-                                .map_err(|e| Error::Handshake(Box::new(e)))
-                                .await?;
-                            request.into_key()
+                                .await;
+                            match request {
+                                Ok(request) => request.into_key(),
+                                Err(err) => return match err {
+                                    handshake::Error::Io(_) => Err(Error::Handshake(Box::new(err))),
+                                    _ => Ok(EitherConnection::Fallback(Fallback {
```
twittner: One way or another we handle errors the wrong way here. By continuing with the connection unless there is an I/O error we may ignore actual handshake errors. To minimise the risk we should only fall back if we encounter a `handshake::Error::Http` (HTTP parsing failed), or a `handshake::Error::HeaderNotFound(name)` with `name` equal to (case insensitive) "Upgrade" (missing upgrade header), or a `handshake::Error::UnexpectedHeader(name)` with `name` equal to (case insensitive) "Upgrade" (not a websocket upgrade). The last two conditions cover cases where the underlying transport sends well formed HTTP. The approach is still not very good protocol design as malformed HTTP requests may still be intended to be handshake messages.

transports/websocket/src/fallback.rs
```diff
@@ -0,0 +1,133 @@
...
+
+use std::pin::Pin;
+use std::task::{Poll, Context};
+use std::io::{self, Read, IoSliceMut};
+use futures::{AsyncRead, AsyncWrite, Stream, Sink};
+use bytes5::Buf;
+use crate::framed::{TlsOrPlain, Connection, OutgoingData};
+
+#[derive(Debug)]
+pub struct Fallback<T> {
```
twittner: `Fallback` could contain the original handshake error that caused the fall back. Then you do not need to add another error case `Error::FellBack` but you could return the more meaningful error in `WsConfig::listen_on`.

twittner:
I am not sure if falling back in case of errors during the handshake is good idea, mostly because of https://github.com/libp2p/rust-libp2p/pull/1360#discussion_r362430947.

transports/websocket/src/framed.rs
expenses: 👍

transports/websocket/src/fallback.rs
expenses: Yep, was thinking that.

expenses:
@twittner I think we're good to go here.

twittner:
> @twittner I think we're good to go here.

As I [mentioned already](https://github.com/libp2p/rust-libp2p/pull/1360#issuecomment-570173992) I am not overly confident that the approach to fall back based on some error heuristics is desirable. Maybe @mxinden, @romanb and @tomaka could weigh in?

romanb:
I also find such fallback strategies based on "best effort" error heuristics generally undesirable. If I may ask the very naive question: The PR description mentions that it would be "super useful" and https://github.com/libp2p/rust-libp2p/issues/1076 basically just says "it should be done", but could someone elaborate on the motivation and why upfront protocol negotiation is not an option?

tomaka:
> could someone elaborate more on the motivation and why upfront protocol negotiation is not an option?

We want to be able to compile Substrate and Polkadot (and applications that use libp2p in general) to WASM and run them in a browser.
Unfortunately browsers only allow WebSocket connections, and we therefore cannot do any protocol negotiation.

At the moment, nodes open two TCP ports: one for plain TCP connections, and one for WebSockets. The motivation is to make it possible to only open one port that would accept both.

The other solution would be to *only* allow WebSockets, but that feels wasteful.

twittner:
> At the moment, nodes open two TCP ports: one for plain TCP connections, and one for WebSockets.

This is much cleaner and preferable in my opinion.

expenses:
Closing

PR #1362, opened 2019-12-31
Enhance NetworkBehaviour docs

koivunej:
Small doc PR as I've now been trying to understand libp2p usage. Did not check if this applies to `stable-futures`. Edited the chat example ~for more commentary and~ to provide `#[behaviour(ignore)]` example.


examples/chat.rs
```diff
@@ -78,13 +78,30 @@ fn main() {
 
     // We create a custom network behaviour that combines floodsub and mDNS.
     // In the future, we want to improve libp2p to make this easier to do.
+    // Use the derive to generate delegating NetworkBehaviour impl and requires the
```
mxinden:
```suggestion
    // Use the derive to generate delegating NetworkBehaviour impl and require the
```

Not a native speaker, but this sounds better to me.

examples/chat.rs
```diff
@@ -163,6 +172,8 @@ fn main() {
                 Async::Ready(None) | Async::NotReady => {
                     if !listening {
                         if let Some(a) = Swarm::listeners(&swarm).next() {
```
mxinden: We might want to consider printing all the opened ports instead of just the first one. See: https://github.com/libp2p/rust-libp2p/pull/898/files#diff-0376ea63d50ca8feb5f8bec42cb2d469R146

examples/chat.rs
```diff
@@ -78,13 +78,30 @@ fn main() {
...
+    // NetworkBehaviourEventProcess implementations below.
     #[derive(NetworkBehaviour)]
     struct MyBehaviour<TSubstream: AsyncRead + AsyncWrite> {
         floodsub: Floodsub<TSubstream>,
         mdns: Mdns<TSubstream>,
+
+        // Struct fields which do not implement NetworkBehaviour need to be ignored
+        #[behaviour(ignore)]
+        #[allow(dead_code)]
+        ignored_member: bool,
```
mxinden: :+1:  I stumbled over this in the past as well. I think it is worth adding here.

swarm/src/behaviour.rs
```diff
@@ -27,6 +27,24 @@ use std::error;
...
+/// # Deriving `NetworkBehaviour`
+///
+/// Crate users should implement this trait by using the the `#[derive(NetworkBehaviour)]`
+/// proc macro re-exported by `libp2p` crate. The macro generates a delegating `trait`
+/// implementation for the `struct`, which delegates method calls to all trait members. Any events
+/// generated by struct members are delegated to
+/// [`NetworkBehaviourEventProcess`](trait.NetworkBehaviourEventProcess.html) implementations which
+/// are expected to be provided by the user.
+///
+/// Optionally `#[behaviour(poll_method = "poll")]` can be added on the struct to call last a custom
```
mxinden: What do you think of the following?

```suggestion
/// Optionally one can implement a custom `poll` function, which needs to be tagged with the `#[behaviour(poll_method = "poll")]` attribute, and would be called last with no parameters.
```

examples/chat.rs
koivunej: Thanks and not a native speaker either but the `s` definetly seems like a typo now that you pointed it out.

examples/chat.rs
koivunej: Thought about that but added only the comment but might as well make that a loop.

swarm/src/behaviour.rs
koivunej: Oh looks like even my sentence split and left open as I did not figure out an easy way to describe the returned type which is loooong. I'll change this to your suggestion, thanks.

koivunej:
Build fails because protobuf has been updated. I thought I saw a fix for that in some other PR?

mxinden:
> Build fails because protobuf has been updated. I thought I saw a fix for that in some other PR?

https://github.com/libp2p/rust-libp2p/pull/1368

koivunej:
I'll rebase on top of master soon-ish to get a green checkmark.

examples/chat.rs
```diff
@@ -157,13 +166,13 @@ fn main() {
 
         loop {
             match swarm.poll().expect("Error while polling swarm") {
-                Async::Ready(Some(_)) => {
-
-                },
+                Async::Ready(Some(_)) => {},
```
koivunej: Not sure if this should be changed; shouldn't there be a break and clean exit here? I am not quite sure about the return of [`Swarm::poll`](https://docs.rs/libp2p/0.13.2/libp2p/swarm/trait.NetworkBehaviour.html#tymethod.poll) or the custom `#[behaviour(poll_method = "...")]` poll method, which I why the sentence I added was left unfinished :)

examples/chat.rs
```diff
@@ -157,13 +166,13 @@ fn main() {
...
                 Async::Ready(None) | Async::NotReady => {
                     if !listening {
-                        if let Some(a) = Swarm::listeners(&swarm).next() {
-                            println!("Listening on {:?}", a);
+                        for addr in Swarm::listeners(&swarm) {
+                            // Since we used the OS assigned port the underlying TCP socket will not
+                            // be opened before the first `swarm.poll`.
```
romanb: I don't understand this comment. Whether you use OS assigned ports or not, the socket will always only be opened once the swarm is polled. The point of `listening` is to print "Listening on ..." _once_ and to only print it _after_ the swarm actually started listening on that socket.

koivunej: Originally I wanted to add this comment as I didn't quite understand why/how to implement this and didn't figure out search the examples. To be honest now that you pointed this out the comment does not make any sense for me either. I'll just remove this as I cannot figure out a better way to write it.

koivunej:
Rebased on top of latest master with stable-futures merged in.

swarm/src/behaviour.rs
```diff
@@ -26,6 +26,24 @@ use std::{error, task::Context, task::Poll};
...
 /// This trait has been designed to be composable. Multiple implementations can be combined into
 /// one that handles all the behaviours at once.
+///
+/// # Deriving `NetworkBehaviour`
+///
+/// Crate users should implement this trait by using the the `#[derive(NetworkBehaviour)]`
+/// proc macro re-exported by `libp2p` crate. The macro generates a delegating `trait`
+/// implementation for the `struct`, which delegates method calls to all trait members. Any events
+/// generated by struct members are delegated to
+/// [`NetworkBehaviourEventProcess`](trait.NetworkBehaviourEventProcess.html) implementations which
```
romanb:
```suggestion
/// [`NetworkBehaviourEventProcess`] implementations which
```
We already use [intra-rustdoc-links](https://github.com/rust-lang/rfcs/blob/master/text/1946-intra-rustdoc-links.md).

koivunej: Oh ok. Initially I thought these were on stable but apparently not and then added the links manually. I'll get rid of these and combine your other suggestions.

koivunej:
Rebase on top of latest master produces no conflicts, probably squash as there are number of non-important commits here?

romanb:
Thanks!

Issue #1351, opened 2019-12-13
Upgrade bytes to 0.5


expenses:
Version 0.5 of the bytes crate has several breaking changes that mean we lose backwards compatibility with older versions. This causes some problems as soketto uses bytes 0.5, while we use bytes 0.4.

twittner:
>  This causes some problems as soketto uses bytes 0.5, while we use bytes 0.4.

What problems exactly?

expenses:
@twittner The versions are incompatible so you can't pass bytes from soketto to libp2p. This prevents me from being able to downgrade a websocket transport to the inner transport and keep the bytes buffer.

twittner:
You can keep and use any bytes-0.5 types encapsulated within your transport. The transport's output type would most likely have to implement `AsyncRead` and `AsyncWrite` and when implementing these traits you are free to operate on those buffers.

tomaka:
#1353 mostly did the upgrade, but leaving open as there's still a `bytes 0.4` in multistream-select.

PR #1367, opened 2020-01-02
Bump bytes to 0.5

bltavares:
`bytes` 0.5 had a couple of breaking changes, specially the split of
`Bytes` and `BytesMut`.

Given how much this code use methods avaialable only on `BytesMut`, this
comit changes the internal field to use this struct and some smaller
code changes to adapt to it.

Closes https://github.com/libp2p/rust-libp2p/issues/1351

### DIFF
--- a/misc/multiaddr/src/lib.rs
+++ b/misc/multiaddr/src/lib.rs
@@ -7,7 +7,7 @@ mod errors;
 mod from_url;
 mod util;
 
-use bytes::Bytes;
+use bytes::BytesMut;
 use serde::{
     Deserialize,
     Deserializer,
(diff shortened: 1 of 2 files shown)


misc/multiaddr/src/lib.rs
```diff
@@ -29,17 +29,17 @@ pub use self::protocol::Protocol;
 
 /// Representation of a Multiaddr.
 #[derive(PartialEq, Eq, Clone, Hash)]
-pub struct Multiaddr { bytes: Bytes }
+pub struct Multiaddr { bytes: BytesMut }
```
twittner: Cloning `Bytes` is cheap, cloning `BytesMut` not so much. If we move away from `Bytes` we might as well use a `Vec<u8>` as we did in the past.

bltavares: I don't know much about memory cost of either. Would you like me to change to `Vec<u8>` on this PR?

twittner: No, I think we should leave the implementation as is for the time being and not upgrade to bytes-0.5. Eventually we will probably want a custom copy-on-write implementation with cheap clones for shared `Multiaddr` values.

PR #1368, opened 2020-01-02
Pin protobuf versin and publish 0.13.2

### LINKED ISSUE #1363: Protobuf version check does not play well with semver
This line does not work well with semver cargo expectation:
https://github.com/libp2p/rust-libp2p/blob/1aeb264cbc0e262af219c92e18eb0025724a9005/core/src/keys_proto.rs#L27 result in https://gitlab.parity.io/parity/substrate/-/jobs/341902
It would either require to have a fix versioning in cargo.toml or check over every version that match semver requirement. Not sure what is better.

tomaka:
(note: this PR is towards master)

Fix #1363 
Also includes https://github.com/libp2p/rust-libp2p/pull/1347

### DIFF
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -23,7 +23,7 @@ multihash = { package = "parity-multihash", version = "0.2.0", path = "../misc/m
 multistream-select = { version = "0.6.0", path = "../misc/multistream-select" }
 futures = "0.1"
 parking_lot = "0.9.0"
-protobuf = "=2.8.1" # note: there exists some semver issues with protobuf at the moment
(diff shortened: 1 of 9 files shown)


core/Cargo.toml
```diff
@@ -23,7 +23,7 @@ multihash = { package = "parity-multihash", version = "0.2.0", path = "../misc/m
 multistream-select = { version = "0.6.0", path = "../misc/multistream-select" }
 futures = "0.1"
 parking_lot = "0.9.0"
-protobuf = "2.8"
+protobuf = "=2.8.1" # note: there exists some semver issues with protobuf at the moment
```
mxinden: Could you reference https://github.com/libp2p/rust-libp2p/issues/1363 here?

PR #1371, opened 2020-01-06
muxers/mplex: Refactor noisy code

mxinden:
See: https://github.com/libp2p/rust-libp2p/pull/1328#discussion_r360745823


tomaka:
I'm actually not a fan of this. In my head, what we return is disjoint from what `start_send` returns, and this change makes the code slightly more difficult to understand.

mxinden:
I am fine either way. For anyone feeling strongly about this, feel free to reopen.

PR #1374, opened 2020-01-06
Add Swarm::next_extended

tomaka:
This PR is against `stable-futures`.
It fixes the situation with the `identify` test by adding a `SwarmEvent` struct containing more events such as "new listener address".

The `Swarm` now has two new async methods: `next()` and `next_extended()`. The former is the same as calling `Stream::next()` except that `None` is never returned. The latter is the same as `next()` except that it returns a `SwarmEvent`.
Additionally, polling the `Swarm` no longer returns a `Result`.

We can either merge before or after #1328, but this change should be in the CHANGELOG.


swarm/src/lib.rs
```diff
@@ -305,44 +332,32 @@ where TBehaviour: NetworkBehaviour<ProtocolsHandler = THandler>,
...
+                    return Poll::Ready(event);
+                }
+            }
+        }).await
+    }
+
+    /// Internal function used by everything event-related.
+    ///
+    /// Polls the `Swarm` for the next event.
+    fn poll_next_extended(mut self: Pin<&mut Self>, cx: &mut Context)
```
romanb:
```suggestion
    fn poll_next_event(mut self: Pin<&mut Self>, cx: &mut Context)
```

swarm/src/lib.rs
```diff
@@ -305,44 +332,32 @@ where TBehaviour: NetworkBehaviour<ProtocolsHandler = THandler>,
...
-      <<<THandler as IntoProtocolsHandler>::Handler as ProtocolsHandler>::OutboundProtocol as UpgradeInfo>::InfoIter: Send + 'static,
-      <<<<THandler as IntoProtocolsHandler>::Handler as ProtocolsHandler>::OutboundProtocol as UpgradeInfo>::InfoIter as IntoIterator>::IntoIter: Send + 'static,
-      <NodeHandlerWrapper<<THandler as IntoProtocolsHandler>::Handler> as NodeHandler>::OutboundOpenInfo: Send + 'static, // TODO: shouldn't be necessary
-      TConnInfo: ConnectionInfo<PeerId = PeerId> + fmt::Debug + Clone + Send + 'static,
-{
-    type Item = Result<TBehaviour::OutEvent, io::Error>;
+    /// Returns the next event that happens in the `Swarm`.
+    ///
+    /// Includes events from the `NetworkBehaviour` but also events about the connections status.
+    pub async fn next_extended(&mut self) -> SwarmEvent<TBehaviour::OutEvent> {
```
romanb:
```suggestion
    pub async fn next_event(&mut self) -> SwarmEvent<TBehaviour::OutEvent> {
```

PR #1376, opened 2020-01-07
Publish 0.14.0-alpha.1

tomaka:
I believe that `multihash` and `multiaddr` didn't get any breaking change.


misc/multiaddr/Cargo.toml
```diff
@@ -6,14 +6,14 @@ description = "Implementation of the multiaddr format"
 homepage = "https://github.com/libp2p/rust-libp2p"
 keywords = ["multiaddr", "ipfs"]
 license = "MIT"
-version = "0.6.0"
+version = "0.6.1"
```
twittner: The `TryFrom<Bytes>` and `TryFrom<BytesMut>` impls have been removed, so maybe we should go for 0.7.0 here.

tomaka:
I also bumped multistream-select. It's actually by accident, but let's do it just in case.

PR #1382, opened 2020-01-09
misc/mdns: Make MdnsService::new sync by using std::net::UdpSocket::bind

mxinden:
MdnsService uses an IP address to create a UDP socket. The address does
not need to be resolved. Therefore one can use std's UdpSocket::bind
instead of the async counterpart from async-std. As a result
MdnsService::new and MdnsService::silent don't need to be async.

https://github.com/libp2p/rust-libp2p/pull/1328#discussion_r360669993

(//CC @DemiMarie-parity)


misc/mdns/src/behaviour.rs
```diff
@@ -90,7 +90,7 @@ impl<TSubstream> Mdns<TSubstream> {
     /// Builds a new `Mdns` behaviour.
     pub async fn new() -> io::Result<Mdns<TSubstream>> {
```
twittner:
```suggestion
    pub fn new() -> io::Result<Mdns<TSubstream>> {
```
?

DemiMarie-parity:
Looks good to me, other than @twittner's correction.

misc/mdns/src/behaviour.rs
mxinden: Right. That makes a lot of sense. Thanks.

mxinden: Right. That makes a lot of sense. Thanks.

mxinden:
Should we merge this after we released `v0.14.0`, given that it is a breaking change to the API surface?

tomaka:
> Should we merge this after we released v0.14.0, given that it is a breaking change to the API surface?

We've already got a few breaking changes in, so let's go ahead with that.

PR #1383, opened 2020-01-09
Allow building an upgrade from an async function

tomaka:
This is a quality-of-life change, with the objective of trying to make rust-libp2p easier to use.

I think it's quite nice and comprehensible to be able to write upgrades as shown in the doc-comment example.

PR #1384, opened 2020-01-09
Remove some Unpin requirements on Futures

tomaka:
Opening as a draft.
The goal is to remove all `Unpin` requirements on futures on traits such as `*Upgrade` or `Transport`.
All the `T::Dial: Unpin`/`T::Listener: Unpin`/`T::ListenerUpgrade: Unpin` requirements still remain to be cleaned up.

This is surprisingly hard and my progress has stalled when chaining multiple futures is involved.
I've used unsafe code for `transport::and_then`, but that should ideally be avoided unless no other solution is found.

If someone wants to take over this work, let me know.


tomaka:
This is ready for review!

There's a breaking change: since the `Transport::Listener` no longer necessarily implements `Unpin`, I had to modify `remove_listener` to not return the `Listener` stream anymore.

tomaka:
Note that I removed all the `Unpin` requirements on futures and streams related to `Transport`s and upgrades, but I haven't touched the ones related to `AsyncRead`/`AsyncWrite`. There are around 180 occurrences of `Unpin` related to `AsyncRead`/`AsyncWrite` to remove, so that's for later.

twittner:
> The goal is to remove all `Unpin` requirements on futures on traits such as `*Upgrade` or `Transport`. [...] There are around 180 occurrences of `Unpin` related to `AsyncRead`/`AsyncWrite` to remove, so that's for later.

Do you have upgrade, transport or i/o types in mind which must not be moved and where our `Unpin` bounds present a problem?

tomaka:
> Do you have upgrade, transport or i/o types in mind which must not be moved and where our Unpin bounds present a problem?

Anything that uses `async`/`await` does not implement `Unpin`. #1383 is a concrete example.

Issue #476, opened 2018-09-13
Implement private networks


tomaka:
Add an equivalent to https://github.com/libp2p/js-libp2p-pnet/

rklaehn:
I might need this in the near future. Can we get a hint how we would go about implementing this?

tomaka:
The implementation would be very similar to libp2p-secio: a struct that implements `OutboundUpgrade` and `InboundUpgrade`, and that apply the cipher on the stream of data by wrapping around the `AsyncRead`/`AsyncWrite`.
The only major difference with secio is the lack of handshake, and that we pass the private shared key directly in the configuration.

rklaehn:
So I would sneak a manual upgrade into the place where the transport is built, like below, except with a yet to be written`PrivateSwarmConfig(swarm_key)` instead of the `PlainText1Config`?

```rust
    Ok(CommonTransport::new()?
        .and_then(move |io, endpoint| {
            libp2p_core::upgrade::apply(
                io,
                PlainText1Config{},
                endpoint,
                libp2p_core::transport::upgrade::Version::V1,
            )
        })
        .upgrade(core::upgrade::Version::V1)
        .authenticate(secio::SecioConfig::new(keypair))
        .multiplex(core::upgrade::SelectUpgrade::new(yamux::Config::default(), mplex::MplexConfig::new()))
        .map(|(peer, muxer), _| (peer, core::muxing::StreamMuxerBox::new(muxer)))
        .timeout(Duration::from_secs(20)))
```

tomaka:
Yes! Normally it should work.

rklaehn:
Very WIP, but does this go into the right direction?

https://github.com/libp2p/rust-libp2p/pull/1385

tomaka:
I haven't looked into the details of the logic, but the general structure is what I'd expect, yes.

tomaka:
Done in #1385

PR #1385, opened 2020-01-09
Implement private networks

rklaehn:
Implements private networks ( #476 )

This is meant to be interoperable with go-ipfs private swarms. For reference: the specification from PL is at https://github.com/libp2p/specs/blob/master/pnet/Private-Networks-PSK-V1.md , the implementation in go-ipfs is at https://github.com/libp2p/go-libp2p-pnet/ .

Tested by checking that two rust-ipfs nodes (pnet example) can chat with each other, and in addition checking that the pubsub traffic is visible in go-ipfs.


rklaehn:
I think the basic encryption is now reasonably correct. Here is the spec for reference: https://github.com/libp2p/specs/blob/master/pnet/Private-Networks-PSK-V1.md

However, I tried getting the chat example to work with go-ipfs with a private swarm key, but could not get it to work.

This might be related to the `UpgradeInfo`. I used a made up protocol name because I had to put something there. `b"/pnet/1.0.0"`. But I suspect that that is causing trouble. I tried implementing an UpgradeInfo that returns an empty collection, but with that I could not get it to work at all...

tomaka:
Oh I see, I didn't think about that.
I think that you will have to apply the handshake manually over the connections that the `Transport` produces.
You can do that with `transport.and_then(|socket, _| handshake(socket))`.

rklaehn:
Not exactly sure how to wire this up. I naively tried this:

```rust
pub fn build_tcp_ws_pnet_secio_mplex_yamux(keypair: identity::Keypair, psk: PSK)
    -> io::Result<impl Transport<Output = (PeerId, impl core::muxing::StreamMuxer<OutboundSubstream = impl Send, Substream = impl Send, Error = impl Into<io::Error>> + Send + Sync), Error = impl error::Error + Send, Listener = impl Send, Dial = impl Send, ListenerUpgrade = impl Send> + Clone>
{
    Ok(CommonTransport::new()?
        .and_then(|socket, endpoint| PnetConfig::new(psk).handshake(socket))
        // .and_then(move |io, endpoint| {
        //     libp2p_core::upgrade::apply(
        //         io,
        //         PnetConfig::new(psk),
        //         endpoint,
        //         libp2p_core::transport::upgrade::Version::V1,
        //     )
        // })
        .upgrade(core::upgrade::Version::V1)
        .authenticate(secio::SecioConfig::new(keypair))
        .multiplex(core::upgrade::SelectUpgrade::new(yamux::Config::default(), mplex::MplexConfig::new()))
        .map(|(peer, muxer), _| (peer, core::muxing::StreamMuxerBox::new(muxer)))
        .timeout(Duration::from_secs(20)))
}
```
...but then the result of the and_then is not a transport anymore so the `.upgrade(core::upgrade::Version::V1)` does not work anymore. I guess I have to call it a day...

tomaka:
> ...but then the result of the and_then is not a transport anymore

I think that's fixed by #1384 
Your `async` function probably produces a non-`Unpin` future that aren't supported now (but will be after #1384)

rklaehn:
Right. Got it to work now. Making handshake return a BoxFuture fixes it!

```rust
    pub fn pin_handshake<TSocket>(self, socket: TSocket) -> BoxFuture<'static, Result<PnetOutput<TSocket>, PnetError>>
        where
            TSocket: AsyncRead + AsyncWrite + Send + Unpin + 'static,
    {
        Box::pin(self.handshake(socket))
    }
```

Got the thing to talk to go-ipfs now. Thanks for helping me sort this out!

examples/pnet.rs
```diff
@@ -0,0 +1,190 @@
+// Copyright 2018 Parity Technologies (UK) Ltd.
```
tomaka: I'm not a fan of having this example. If the example is supposed to demonstrate how to use a private network, then there are only 4 relevant lines in the entire example, and the rest is the unfortunate complexity of floodsub/mdns.

protocols/pnet/src/lib.rs
```diff
@@ -0,0 +1,333 @@
...
+use std::fmt::{self, Write};
+use std::io::Error as IoError;
+
+const KEY_SIZE: usize = 32;
+const NONCE_SIZE: usize = 24;
+const WRITE_BUFFER_SIZE: usize = 1024;
+const FINGERPRINT_SIZE: usize = 16;
+
+#[derive(Copy, Clone, PartialEq, Eq)]
+pub struct PSK([u8; KEY_SIZE]);
```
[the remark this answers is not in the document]
rklaehn: :+1:

protocols/pnet/src/lib.rs
```diff
@@ -0,0 +1,333 @@
...
+        let size = std::cmp::min(buf.len(), self.write_buffer.len());
+        let write_buffer = &mut self.write_buffer[0..size];
+        write_buffer.copy_from_slice(&buf[..size]);
+        if self.write_cipher.current_pos() != self.write_offset {
+            trace!("seeking to {}", self.write_offset);
+            self.write_cipher.seek(self.write_offset);
+        }
+        trace!("encrypting {} bytes", size);
+        self.write_cipher.apply_keystream(write_buffer);
+        Pin::new(&mut self.inner).poll_write(cx, &self.write_buffer[..size])
```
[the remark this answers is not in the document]
rklaehn: It was the easiest thing I could get to work, given that I am not familiar with async-std.

So you want the thing to return the number of encrypted bytes even if the wrapped stream returns Pending, and then write the bytes the next time poll gets called? Are there higher level combinators for something like this or do I need to roll my own?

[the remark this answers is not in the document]
rklaehn: On further thought, not even sure how that is supposed to work. E.g. when the *last* write to an endpoint returns pending, I would then pretend to accept the bytes, but there would not be another poll call, so the data would be lost.

The closest thing in async std I could find was https://docs.rs/async-std/1.4.0/async_std/io/struct.BufWriter.html . What you want is essentially a BufWriter that also encrypts. But the BufWriter requires explicitly calling flush...

"Unlike the BufWriter type in std, this type does not write out the contents of its buffer when it is dropped. Therefore, it is absolutely critical that users explicitly flush the buffer before dropping a BufWriter."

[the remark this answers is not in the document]
tomaka: More or less.
What I mean is:

- Maintain a write buffer and range of data that is yet to be sent to `inner`.
- Create a method `flush_local_buffer()` that tries to write the content of `write_buffer` to `inner`.
- `poll_flush()` calls `flush_local_buffer` then `inner.poll_flush()`.
- `poll_write()` writes in the local buffer and returns success. If the local buffer is full, first call `flush_local_buffer`.

examples/pnet.rs
rklaehn: Will remove or come up with something smaller.

protocols/pnet/src/lib.rs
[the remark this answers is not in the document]
rklaehn: I have stripped down and adapted BufWriter from async_std ( https://docs.rs/async-std/1.4.0/async_std/io/struct.BufWriter.html ) now. Still works, and no more need to seek in the write cipher.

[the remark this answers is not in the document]
rklaehn: OK, I have now added a CryptWriter, which is a stripped down and adapted version of BufWriter from async_std. It still works, and there is no more seeking needed in the write cipher.

rklaehn:
I will remove or replace the pnet example before merging this. Now just want to keep it in for testing.

tomaka:
To me, multiple calls to `poll_write` should continue adding data at the end of the buffer, and only call `poll_flush_buf` if there is no space left.

However I think the current approach is fine.

Looks mostly good for me, but needs documentation.

tomaka:
Wait, I didn't realize that you had changed the logic of the buffer writing.

protocols/pnet/src/crypt_writer.rs
```diff
@@ -0,0 +1,146 @@
...
+    pub fn get_pin_mut(self: Pin<&mut Self>) -> Pin<&mut W> {
+        self.project().inner
+    }
+}
+
+/// Write the contents of a Vec<u8> into an AsyncWrite.
+///
+/// The handling 0 byte progress and the Interrupted error was taken from BufWriter in async_std.
+///
+/// If this fn returns Ready(Ok(())), the buffer will be completely flushed.
```
tomaka: That isn't true anymore.

rklaehn:
Happy to revert it if you want, but I think the new code is better.

protocols/pnet/src/crypt_writer.rs
```diff
@@ -0,0 +1,146 @@
...
+impl<W: AsyncWrite> AsyncWrite for CryptWriter<W> {
+    fn poll_write(
+        self: Pin<&mut Self>,
+        cx: &mut Context<'_>,
+        buf: &[u8],
+    ) -> Poll<io::Result<usize>> {
+        let mut this = self.project();
+        // completely flush the buffer, returning pending if not possible
+        ready!(poll_flush_buf(&mut this.inner, this.buf, cx))?;
+        // if we get here, the buffer is empty
```
tomaka: That isn't true anymore.

protocols/pnet/src/crypt_writer.rs
```diff
@@ -0,0 +1,146 @@
...
+        let res = Pin::new(&mut *this.buf).poll_write(cx, buf);
+        if let Poll::Ready(Ok(count)) = res {
+            this.cipher.apply_keystream(&mut this.buf[0..count]);
```
tomaka: The `0` here is therefore wrong because the buffer might not have been empty.

protocols/pnet/src/crypt_writer.rs
```diff
@@ -0,0 +1,146 @@
...
+            trace!("encrypted {} bytes", count);
+        };
+        // flush immediately afterwards, but if we get a pending we don't care
```
tomaka: Why flush? There's no reason to and it worsen performances.

protocols/pnet/src/crypt_writer.rs
```diff
@@ -0,0 +1,146 @@
...
+
+impl<W: AsyncWrite> AsyncWrite for CryptWriter<W> {
+    fn poll_write(
+        self: Pin<&mut Self>,
+        cx: &mut Context<'_>,
+        buf: &[u8],
+    ) -> Poll<io::Result<usize>> {
+        let mut this = self.project();
+        // completely flush the buffer, returning pending if not possible
+        ready!(poll_flush_buf(&mut this.inner, this.buf, cx))?;
```
tomaka: Ideally should only be done if there isn't enough space to fit all the data.

protocols/pnet/src/crypt_writer.rs
```diff
@@ -0,0 +1,146 @@
...
+        if let Poll::Ready(Err(e)) = poll_flush_buf(&mut this.inner, this.buf, cx) {
+            Poll::Ready(Err(e))
+        } else {
+            res
+        }
+    }
+
+    fn poll_flush(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<io::Result<()>> {
+        let mut this = self.project();
+        ready!(poll_flush_buf(&mut this.inner, this.buf, cx))?;
```
tomaka: Wrong if `poll_flush_buf` doesn't guarantee an empty buffer.

protocols/pnet/src/crypt_writer.rs
```diff
@@ -0,0 +1,146 @@
...
+        this.inner.poll_flush(cx)
+    }
+
+    fn poll_close(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<io::Result<()>> {
+        let mut this = self.project();
+        ready!(poll_flush_buf(&mut this.inner, this.buf, cx))?;
```
tomaka: Wrong if `poll_flush_buf` doesn't guarantee an empty buffer.

protocols/pnet/src/crypt_writer.rs
rklaehn: I don't know if flush will be called from upstream. If you tell me that it is safe to assume that flush will be called so no data is being stuck in the buffer, I am happy to remove this. But I am not familiar enough with the overall flow in rust-libp2p.

protocols/pnet/src/crypt_writer.rs
rklaehn: Am I correct that the `ready!` macro just returns pending on pending, and produces a value if the value is *not* pending? That is what I got from [the docs](https://docs.rs/futures-preview/0.3.0-alpha.19/futures/macro.ready.html)...

Then the `?` suffix bails out if the value produced by the `ready!(...)` is an error. So we only get here when poll_flush_buf has produced an `Poll::Ready(Result::Ok(whatever))` . At least that is my understanding.

protocols/pnet/src/crypt_writer.rs
rklaehn: Why not?

- If the while loop completes normally, everything up to `len` has been written, the buffer is drained, and therefore empty / completely flushed.

- All break cases are either for `Poll::Ready(Err(whatever))` or for `Poll::Pending`. In these cases the buffer might not be empty, but the part that has been written is nevertheless drained.

I guess the sentence should read "If this fn returns Ready(Ok(())), the buffer has been completely flushed and therefore is empty", otherwise it is a bit ambiguous.

protocols/pnet/src/crypt_writer.rs
rklaehn: As mentioned above, I think that poll_flush_buf does guarantee that when it returns Poll::Ready(Ok(())), the buffer is indeed empty. If this is not the case, this line is broken.

But if poll_flush_buf returns Pending or an error, isn't it OK here to pass through that error / pending using the ready! macro and the ? suffix? If you return Pending, whoever calls this method will try again until that is no longer the case. And if you return an error, whoever calls this method is supposed to stop. At least that is how I interpreted the contract for the poll_xxx methods.

protocols/pnet/src/crypt_writer.rs
```diff
@@ -0,0 +1,146 @@
...
+            }
+            Poll::Ready(Err(e)) => {
+                // Interrupted is the only error that we consider to be recoverable by trying again
+                if e.kind() != io::ErrorKind::Interrupted {
+                    // for any other error, don't try again
+                    ret = Poll::Ready(Err(e));
+                    break;
+                }
+            }
+            Poll::Pending => break,
```
bantonsson: I can't see that we are setting `ret` to `Poll::Pending` here.

protocols/pnet/src/crypt_writer.rs
rklaehn: Never mind. As my colleague pointed out, I missed setting ret to Pending.

rklaehn: I still have not found a way to write these poll loops that is not somewhat jarring... Tail recursion would be so much nicer...

protocols/pnet/src/crypt_writer.rs
rklaehn: Here is my thinking behind the "just do encryption" decision: the private swarms are an optional feature that is sometimes, but not always, enabled. If buffering is necessary it should be done somewhere else so we also have it when private swarms are disabled. If buffering is happening at a higher layer, data will be sent to / read from this layer already in large chunks, so no need for additional chunking.

protocols/pnet/src/crypt_writer.rs
```diff
@@ -0,0 +1,149 @@
...
+impl<W: AsyncWrite> AsyncWrite for CryptWriter<W> {
+    fn poll_write(
+        self: Pin<&mut Self>,
+        cx: &mut Context<'_>,
+        buf: &[u8],
+    ) -> Poll<io::Result<usize>> {
+        let mut this = self.project();
+        // completely flush the buffer, returning pending if not possible
+        ready!(poll_flush_buf(&mut this.inner, this.buf, cx))?;
+        // if we get here, the buffer is empty
```
tomaka:
```suggestion
        // if we get here, the buffer is empty
        debug_assert!(this.buf.is_empty());
```

protocols/pnet/src/crypt_writer.rs
```diff
@@ -0,0 +1,149 @@
...
+            Poll::Pending => {
+                ret = Poll::Pending;
+                break;
+            }
+        }
+    }
+    if written > 0 {
+        buf.drain(..written);
+    }
+    ret
```
tomaka:
```suggestion
    if let Poll::Ready(Ok(())) = ret { debug_assert!(buf.is_empty()); }
    ret
```

protocols/pnet/src/crypt_writer.rs
```diff
@@ -0,0 +1,149 @@
...
+    pub fn get_pin_mut(self: Pin<&mut Self>) -> Pin<&mut W> {
+        self.project().inner
+    }
+}
+
+/// Write the contents of a Vec<u8> into an AsyncWrite.
+///
+/// The handling 0 byte progress and the Interrupted error was taken from BufWriter in async_std.
+///
+/// If this fn returns Ready(Ok(())), the buffer has been completely flushed.
```
tomaka:
```suggestion
/// If this fn returns Ready(Ok(())), the buffer has been completely flushed and is empty.
```

protocols/pnet/src/crypt_writer.rs
```diff
@@ -0,0 +1,149 @@
...
+    ) -> Poll<io::Result<usize>> {
+        let mut this = self.project();
+        // completely flush the buffer, returning pending if not possible
+        ready!(poll_flush_buf(&mut this.inner, this.buf, cx))?;
+        // if we get here, the buffer is empty
+        let res = Pin::new(&mut *this.buf).poll_write(cx, buf);
+        if let Poll::Ready(Ok(count)) = res {
+            this.cipher.apply_keystream(&mut this.buf[0..count]);
+            trace!("encrypted {} bytes", count);
+        };
```
tomaka:
```suggestion
        } else {
            debug_assert!(false);
        };
```

rklaehn:
@tomaka I have taken all your suggestions and merged them into one commit, and rebased on current master.

PR #1390, opened 2020-01-13
Replace protobuf crate with prost!

twittner:
Replaces all uses of rust-protobuf with prost! using build files instead of committing generated sources into the repository.


tomaka:
Fwiw here's a link to how `prost` works: https://docs.rs/prost-build/0.6.0/prost_build/#sourcing-protoc

core/build.rs
```diff
@@ -0,0 +1,3 @@
+fn main() {
```
tomaka: All the `build.rs` files are missing copyright headers.

protocols/floodsub/src/rpc.proto
```diff
@@ -8,15 +8,15 @@ message RPC {
...
+		optional string topic_id = 2;
 	}
 }
 
 message Message {
 	optional bytes from = 1;
 	optional bytes data = 2;
 	optional bytes seqno = 3;
-	repeated string topicIDs = 4;
+	repeated string topic_ids = 4;
```
tomaka: Does prost not support camelcase? :thinking:

twittner: Maybe there is a tweak but IIRC `topicIDs` was turned into `topic_i_ds`. I suppose `topicIds` would have worked too.

core/build.rs
twittner: I will add those.

protocols/kad/src/protocol.rs
```diff
@@ -333,157 +332,118 @@ pub enum KadResponseMsg {
...
-            msg.set_record(record);
-
-            msg
+        KadResponseMsg::Pong => proto::Message {
+            r#type: proto::message::MessageType::Ping as i32,
+            .. proto::Message::default()
+        },
+        KadResponseMsg::FindNode { closer_peers } => proto::Message {
+            r#type: proto::message::MessageType::FindNode as i32,
+            cluster_level_raw: 10,
```
tomaka: This is normally 9 but this field is unused so it doesn't matter.

protocols/kad/src/protocol.rs
```diff
@@ -57,30 +57,26 @@ pub enum KadConnectionType {
...
+    fn into(self) -> proto::message::ConnectionType {
+        use proto::message::ConnectionType::*;
         match self {
-            KadConnectionType::NotConnected => NOT_CONNECTED,
-            KadConnectionType::Connected => CONNECTED,
-            KadConnectionType::CanConnect => CAN_CONNECT,
-            KadConnectionType::CannotConnect => CANNOT_CONNECT,
+            KadConnectionType::NotConnected => NotConnected,
+            KadConnectionType::Connected => Connected,
+            KadConnectionType::CanConnect => CannotConnect,
```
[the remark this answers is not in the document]
twittner: Thanks eagle eye!

twittner:
> By curiosity, did you test against existing nodes? If not, I briefly would like to test.

I had a quick run with substrate against Flaming Fir nodes. But feel free to run another test.

protocols/kad/src/protocol.rs
twittner: Sorry for those mistakes. A bit too much copy and paste. :blush:

protocols/floodsub/src/rpc.proto
rklaehn: Is this change backwards compatible with go-ipfs floodsub?

twittner: Yes, as far as protocol buffers is concerned, names do not matter, hence those index numbers.

rklaehn: Right, completely forgot about that. Sorry about the stupid question.

FlorianFranzen:
It might be worth mentioning somewhere that the build process now includes running a precompiled version of ```protoc```, which might result in odd build errors it that binary fails for any reason, as it does e.g. on NixOS.
In that case the environment variable ```PROTOC``` and ```PROTOC_INCLUDE``` have to be used to point to a working installation of ```protobuf``` and it compiler to build libp2p or any project depending on it successful.

General-Beck:
On Alpine working with `PROTOC=/usr/bin/protoc PROTOC_INCLUDE=/usr/include cargo build ....`
deps `apk add protoc`
(sourcing-protoc)[https://docs.rs/prost-build/0.5.0/prost_build/#sourcing-protoc]

whalelephant:
hi @General-Beck, I am trying to compile this on alpine3:12 and I am getting error from the core/build.rs. Have you seen this? 

```
  thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 2, kind: NotFound, message: "No such file or directory" }', 
/usr/local/cargo/registry/src/github.com-1ecc6299db9ec823/libp2p-core-0.21.0/build.rs:22:64
```

Issue #1318, opened 2019-11-25
Upgrade to latest tokio (0.2) or switch to async-std?


daogangtang:
Hi, is there some plans to upgrade to latest tokio (0.2) or switch to async-std?

Thank you.

mxinden:
Hi @daogangtang,

We are currently updating rust-libp2p to use *stable futures* as well as *async/await*. You can follow the progress via the `stable-futures` branch: https://github.com/libp2p/rust-libp2p/tree/stable-futures. Most of the modules only depend on the *async-std* and the *futures* crate.

auroraexecution:
Good News!

daogangtang:
Good news!

vorot93:
Any particular reason for choosing async-std over tokio?

tomaka:
> Any particular reason for choosing async-std over tokio?

Initially it's because tokio 0.2 wasn't even a thing.
Now I'd be more keep on picking async-std because they work independently of which executor polls them (contrary to tokio sockets that need a tokio reactor to poll them).

carllerche:
You are free to pick whichever you want. I will just clarify some points for future readers of this issue.

Both Tokio and async-std require a reactor to drive socket types. The primary difference is that Tokio __does not__ implicitly start a reactor where as async-std implicitly start background threads from a static context. The advantage of not implicitly starting a reactor is that it allows the user to control how, when, and where the reactor runs (background thread, current thread, inline w/ blocking fn calls, ...).

Tokio sockets support being used from any executor, it just requires you to specify **which** reactor to use to bind them. Roughly, it is something like:

```rust
// start a runtime that only includes an I/O driver
let rt = runtime::Builder::new()
    .enable_io()
    .build()
    .unwrap();

let rt_handle = rt.handle().clone();

// Run the runtime somewhere (background thread?)

let listener = rt_handle.enter(|| tokio::net::TcpListener::bind(...));
```

This has some advantages. For example, `rust-postgres` has seen significant performance improvements by controlling the details of how the I/O driver runs in its sync API bridge.

tomaka:
I totally agree that tokio has theoretically better performances. What I'm trying to avoid here is the surprise factor.

I hate the situation where everything compiles just fine, but then produces runtime errors because some initialization process hasn't been performed. In particular, I've been bitten multiple times when it comes to timers, since they appear to instantly succeed if not polled through a reactor of the same tokio version.

[...]

PR #1391, opened 2020-01-15
Allow configuring the tasks executor

tomaka:
Makes it possible to configure the executor for the background tasks that libp2p spawns.

Related to #1318, as it makes it possible to use the tokio executor.

The public-facing API is that `SwarmBuilder` gains an `executor` method.


tomaka:
I've added an `executor_fn` method, to make it easier to use tokio.
Typically, to use the tokio executor you'd call:

```
let handle = tokio_runtime.handle();
swarm_builder.executor_fn(|fut| handle.spawn(fut));
```

core/src/nodes/tasks/manager.rs
```diff
@@ -176,8 +174,14 @@ impl<I, O, H, E, HE, T, C> Manager<I, O, H, E, HE, T, C> {
...
-        if let Some(threads_pool) = &mut self.threads_pool {
-            threads_pool.spawn_ok(task);
+        if let Some(executor) = &self.executor {
+            if executor.status().is_ok() {
+                if executor.spawn(task).is_err() {
+                    log::error!("Failed to spawn task for peer");
+                }
+            } else {
+                self.local_spawns.push(task);
+            }
```
romanb: Just curious: Why not
```rust
if let Some(executor) = &self.executor {
    if let Err(e) = executor.spawn(task) {
        log::warn!("Failed to spawn task onto executor ({:?}). Resorting to in-place polling.", e);
        self.local_spawns.push(task);
    }           
} else {
    self.local_spawns.push(task);
}
```
?

swarm/src/lib.rs
```diff
@@ -624,6 +633,31 @@ where TBehaviour: NetworkBehaviour,
...
+        use futures::task::SpawnError;
+        struct SpawnImpl<F>(F);
+        impl<F: Fn(FutureObj<'static, ()>)> Spawn for SpawnImpl<F> {
+            fn spawn_obj(&self, future: FutureObj<'static, ()>) -> Result<(), SpawnError> {
+                (self.0)(future);
+                Ok(())
+            }
+            fn status(&self) -> Result<(), SpawnError> {
+                Ok(())
+            }
```
romanb: This is the default impl, so not really necessary, unless you prefer being explicit here.

core/src/nodes/tasks/manager.rs
```diff
@@ -63,9 +63,9 @@ pub struct Manager<I, O, H, E, HE, T, C = PeerId> {
     /// Identifier for the next task to spawn.
     next_task_id: TaskId,
 
-    /// Threads pool where we spawn the nodes' tasks. If `None`, then we push tasks to the
+    /// Custom executor where we spawn the nodes' tasks. If `None`, then we push tasks to the
     /// `local_spawns` list instead.
-    threads_pool: Option<ThreadPool>,
+    executor: Option<Sp>,
```
romanb: A pity about the rippling effect of the additional type parameter. Why not just `Option<Box<dyn Spawn>>`, since the Swarm already uses such a trait object (and thus dynamic dispatch) anyway?

core/src/nodes/tasks/manager.rs
tomaka: Calling `executor.spawn` entirely consumes the task.

core/src/nodes/tasks/manager.rs
tomaka: The idea is that `libp2p-core` is as zero-cost and generic as possible, while `libp2p-swarm` is a bit more opinionated.
I don't really have a strong opinion on extra template parameter vs `Box<dyn Spawn>` here though.

core/src/nodes/tasks/manager.rs
romanb: Oh, of course :man_facepalming:

core/src/nodes/tasks/manager.rs
romanb: I'd probably go with the more convenient option unless it proves to be a meaningful bottleneck, but I'm fine with keeping the additional type parameter, if you prefer that.

twittner: +1 for the trait object.

core/src/nodes/tasks/manager.rs
```diff
@@ -212,8 +217,14 @@ impl<I, O, H, E, HE, T, C> Manager<I, O, H, E, HE, T, C> {
         let task: Task<Pin<Box<futures::future::Pending<_>>>, _, _, _, _, _, _> =
             Task::node(task_id, self.events_tx.clone(), rx, HandledNode::new(muxer, handler));
 
-        if let Some(threads_pool) = &mut self.threads_pool {
-            threads_pool.spawn_ok(Box::pin(task));
+        if let Some(executor) = &self.executor {
+            if executor.status().is_ok() {
+                if executor.spawn(Box::pin(task)).is_err() {
+                    log::error!("Failed to spawn task for peer");
```
twittner: We should not return a `TaskId` if we did not actually spawn a task.

swarm/src/lib.rs
```diff
@@ -611,9 +613,16 @@ where TBehaviour: NetworkBehaviour,
       TConnInfo: ConnectionInfo<PeerId = PeerId> + fmt::Debug + Clone + Send + 'static,
 {
     pub fn new(transport: TTransport, behaviour: TBehaviour, local_peer_id: PeerId) -> Self {
+        let executor = ThreadPoolBuilder::new()
```
thomaseizinger: This could be moved to the `build` method, then you wouldn't have to pay the cost of constructing a `ThreadPoolBuilder` if the user overwrites it with `executor` or `executor_fn` anyway.

tomaka:
>  think we should not use the Spawn trait but define our own executor trait which gets implemented for tokio and async-std

But then we'd have to depend on `async-std` and `tokio`, whereas the `executor` trait is in the `futures` crate.

twittner:
> But then we'd have to depend on `async-std` and `tokio`, whereas the `executor` trait is in the `futures` crate.

The impls should be feature-gated. Users who want a particular executor pick the corresponding feature.

tomaka:
> The impls should be feature-gated. Users who want a particular executor pick the corresponding feature.

The problem is that to me this looks strictly worse than what this PR does.

twittner:
> The problem is that to me this looks strictly worse than what this PR does.

Would you mind explaining why? [Here](https://github.com/twittner/rust-libp2p/commit/d09c6490bb395150cfd102d5836f80372762b59c) is a sketch implementing what I had in mind ([diff to master](https://github.com/libp2p/rust-libp2p/compare/master...twittner:exec)). In particular I would be glad to see [this block](https://github.com/twittner/rust-libp2p/commit/d09c6490bb395150cfd102d5836f80372762b59c#diff-df0155e646a4bc9e417984c29541ec94L178) go away. Do you think this would be difficult to use?

tomaka:
I like your sketch, and the fact that we are guaranteed success when spawning tasks.
I'm not a fan of having `tokio` and `async-std` features in the core, which is supposed to be as "pure" as possible.

What about asking the user to pass `Fn(Pin<Box<dyn Future<Output = ()>>>)`, rather than having a custom `Executor` trait? That way, we would be depending on other crates.

twittner:
> What about asking the user to pass `Fn(Pin<Box<dyn Future<Output = ()>>>)`, rather than having a custom `Executor` trait? That way, we wouldn't be depending on other crates.

Worth a shot. We could stick with `FutureObj`, as it already has some nice `From` impls.

twittner:
> I'm not a fan of having `tokio` and `async-std` features in the core, which is supposed to be as "pure" as possible.

In theory we could add another crate `libp2p-executor` and put the trait and impls in there. This way we could also remove the `thread-pool` and `executor` feature flags from `libp2p-core`.

tomaka:
Now ready for another review pass.

PR #1392, opened 2020-01-15
Remove the extra memcpies in tasks.rs

tomaka:
Every single time something happens on a connection, we `mem::replace` the state with `Undefined`, then put the state back afterwards.
This is a lot of copies, and 5% of Substrate's CPU use is here.

In the past it's been tedious to fix because of non-lexical lifetime issues in Rust. But Rust has gotten some improvements since then.

### DIFF
--- a/core/src/nodes/tasks/task.rs
+++ b/core/src/nodes/tasks/task.rs
@@ -27,7 +27,7 @@ use crate::{
 };
 use futures::{prelude::*, channel::mpsc, stream};
 use smallvec::SmallVec;
-use std::{pin::Pin, task::Context, task::Poll};
+use std::{mem, pin::Pin, task::Context, task::Poll};
 use super::{TaskId, Error};
 
 /// Message to transmit from the public API to a task.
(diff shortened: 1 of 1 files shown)


tomaka:
Tried on Polkadot. Everything working fine.

core/src/nodes/tasks/task.rs
```diff
@@ -268,10 +273,13 @@ where
                                 this.taken_over.push(take_over),
                             Poll::Ready(None) =>
                                 // Node closed by the external API; start closing.
-                                if let Some(n) = node {
-                                    this.state = State::Closing(n.close());
+                                if node.is_some() {
```
mxinden: Wouldn't this also work, given that `extract_node` returns `None` when there is no node?

```rust
                                if let Some(node) = this.state.extract_node() {
                                    this.state = State::Closing(node.close());
                                    continue 'poll
                                } else {
                                    this.state = State::Undefined;
                                    return Poll::Ready(()) // end task
                                }
```

romanb:
> This is a lot of copies, and 5% of Substrate's CPU use is here.

How did you measure this and pinpoint it to this particular use of `mem::replace`? Just so others may try to reproduce.

On first glance, this doesn't exactly look like an improvement to the code (mainly judging by LOC and new occurrences of `expect()` and `unreachable`), so if this is an important optimisation, I agree with @mxinden in that it should be at least prominently documented and ideally accompanied by a reproducible benchmark to avoid future regressions.

tomaka:
It is indeed not as obvious that there's an issue here.
In the past, profiling has shown a ~5% CPU usage on some symbol whose name ressembles `avx_unaligned_memcpy` when called directly by `Task::poll`. This is the reason why I suspected this `mem::replace`.

However it is indeed possible that this instead comes from an inlined function that isn't shown in the backtrace.

I haven't had time to run before/after benchmarks.

tomaka:
Going to close this after previous message.

PR #1393, opened 2020-01-15
[swarm] Log connection close in swarm. Simplify some bounds.

romanb:
This is just a small PR that:

  1. Adds logging for `NodeClosed` events, as otherwise the contained `error` is forever lost. I chose `trace` level for, but I guess `debug` would be ok, too. This is primarily useful when troubleshooting connection issues.
  2. Simplifies some trait bounds.


romanb:
> Maybe we should pass the error in `SwarmEvent::Disconnected`

I thought about it, but went with logging to start with as the least intrusive option (i.e. no API changes). I can follow up with another PR that proposes the change to the event.

tomaka:
I'd go with the API change. This can be either in this PR or a different one.

PR #1279, opened 2019-10-18
floodsub - Removes hashing for topic identifier

peat:
If I understood the conversation around #473, the decision was to push the "hash or not to hash" decision up to the application, in order to be compatible with the Go and JS implementations and let app developers decide how they want to broadcast topics.

### Changes

I stripped everything down to a thin wrapper around the `rpc_proto::TopicDescriptor` and used `String` for the topic identifiers instead of creating another struct or type alias.

### Testing

Existing tests pass, and `examples/chat` works as expected. Please let me know if there's anything else expected. 😁

Issue #473, opened 2018-09-12
rust-libp2p hashes pubsub topics which prevents Go/JS interoperability


mgoelzer:
I'm trying to demonstrate floodsub interoperability between Go, JS and Rust.  In all cases, the topic name is `libp2p-demo-chat`.  Go and JS use that topic string verbatim, but rust-libp2p hashes it to `RDEpsjSPrAZF9JCK5REt3tao`.

Thoughts?  I'd really like to demo the 3 languages inter-operating at some conferences this fall.  Guessing you did this for privacy reasons @tomaka?

Example output below.  This one is a Go peer connecting to my bootstrapper:

```
<NOTICE> Got connection from /ip4/127.0.0.1/tcp/46023
handleIncomingRPC: subopt.GetTopicid() = 'libp2p-demo-chat'
```

Now a JS peer connecting:

```
<NOTICE> Got connection from /ip4/127.0.0.1/tcp/56098
handleIncomingRPC: subopt.GetTopicid() = 'libp2p-demo-chat'
```

Now a rust peer connecting:

```
<NOTICE> Got connection from /ip4/127.0.0.1/tcp/56076
handleIncomingRPC: subopt.GetTopicid() = 'RDEpsjSPrAZF9JCK5REt3tao'
```

mgoelzer:
@vyzo Any thoughts here?  Rust is doing the right thing here by hashing*.  But Go and JS are both doing the wrong thing in a compatible way.  What should we do?

@Stebalien and I talked about asking Rust to add a flag to do the wrong thing (no hashing) while simultaneously upgrading Go and JS to do the right thing (hash the topic) and then all 3 would be both forward and backwards compatible.  But getting Go and JS to make a change like at the same time is hard.

(*Assuming the hash is computed over the descriptor and the topic string, which I didn't check.)

tomaka:
> Guessing you did this for privacy reasons @tomaka?

I did this out of ignorance. I mistakenly guessed that it was supposed to be hashed.

mgoelzer:
I think you actually did the "right" thing (according to the consensus opinion here in Palo Alto.

Let's see what @vyzo thinks.  For now, I can workaround in my demo by giving Go and JS the hashed topic to match rust.  (I have all three languages communicating on a pubsub topic now!)

vyzo:
The original intent was for topic descriptors to be CIDs, which would be hashed.
[...]

PR #1395, opened 2020-01-17
Addressing #473 ... if I understood the ticket right, we want to pass…

rklaehn:
… through whatever the application provides as a topic identifier, leaving hashing (or not hashing) up to the application.

This is https://github.com/libp2p/rust-libp2p/pull/1279, just rebased and adapted to latest master.

### DIFF
--- a/protocols/floodsub/src/layer.rs
+++ b/protocols/floodsub/src/layer.rs
@@ -52,7 +52,7 @@ pub struct Floodsub<TSubstream> {
     /// List of peers the network is connected to, and the topics that they're subscribed to.
     // TODO: filter out peers that don't support floodsub, so that we avoid hammering them with
     //       opened substreams
-    connected_peers: HashMap<PeerId, SmallVec<[String; 8]>>,
+    connected_peers: HashMap<PeerId, SmallVec<[Topic; 8]>>,
 
     // List of topics we're subscribed to. Necessary to filter out messages that we receive
     // erroneously.
@@ -85,13 +85,13 @@ impl<TSubstream> Floodsub<TSubstream> {
     pub fn add_node_to_partial_view(&mut self, peer_id: PeerId) {
         // Send our topics to this node if we're already connected to it.
         if self.connected_peers.contains_key(&peer_id) {
-            for topic in self.subscribed_topics.iter() {
+            for topic in self.subscribed_topics.iter().cloned() {
                 self.events.push_back(NetworkBehaviourAction::SendEvent {
                     peer_id: peer_id.clone(),
                     event: FloodsubRpc {
                         messages: Vec::new(),
                         subscriptions: vec![FloodsubSubscription {
-                            topic: topic.id().to_owned(),
+                            topic: topic,
                             action: FloodsubSubscriptionAction::Subscribe,
                         }],
                     },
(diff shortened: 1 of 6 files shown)


tomaka:
I'm not so much a fan of the API, but I also didn't have time to look into what I think we should do.
The `Topic` struct after this PR is pretty useless, and I'm not a fan of using a simple `String` (rather than some more strongly-typed struct) for the topic.

rklaehn:
OK, just wanted to get the old PR up to date first.

So should I turn `Topic` into a simple newtype wrapper and use that everywhere where it now uses `String` for topic? What's the whole TopicDescriptor protobuf msg for, anyway?

rklaehn:
So it seems that a TopicDescriptor is basically a string and some options that are currently not being used. It is also nowhere referenced from the RPC messages, so I am inclined to remove the whole thing from the .proto until PL makes up their mind what they want to do with these future options...

https://github.com/libp2p/specs/tree/master/pubsub#the-topic-descriptor

protocols/floodsub/src/topic.rs
```diff
@@ -18,93 +18,27 @@
...
-        self.builder.encode(&mut buf).expect("Vec<u8> provides capacity as needed");
-        // TODO: https://github.com/libp2p/rust-libp2p/issues/473
-        let hash = TopicHash {
-            hash: bs58::encode(&buf).into_string(),
-        };
-        Topic {
-            descriptor: self.builder,
-            hash,
-        }
+impl Into<String> for Topic {
```
tomaka: Should be the other way around

```suggestion
impl From<Topic> for String {
```

protocols/floodsub/src/layer.rs
```diff
@@ -168,40 +167,40 @@ impl<TSubstream> Floodsub<TSubstream> {
     }
 
     /// Publishes a message to the network, if we're subscribed to the topic only.
-    pub fn publish(&mut self, topic: impl Into<TopicHash>, data: impl Into<Vec<u8>>) {
+    pub fn publish(&mut self, topic: Topic, data: impl Into<Vec<u8>>) {
```
tomaka: Could you use `impl Into<Topic>` everywhere in the public API? (where there was `impl Into<TopicHash>` before)
And also add `impl From<String> for Topic` and `impl From<&str> for Topic`? This way, users can just pass a string here.

protocols/floodsub/src/layer.rs
```diff
@@ -85,13 +85,13 @@ impl<TSubstream> Floodsub<TSubstream> {
...
         if self.connected_peers.contains_key(&peer_id) {
-            for topic in self.subscribed_topics.iter() {
+            for topic in self.subscribed_topics.iter().cloned() {
                 self.events.push_back(NetworkBehaviourAction::SendEvent {
                     peer_id: peer_id.clone(),
                     event: FloodsubRpc {
                         messages: Vec::new(),
                         subscriptions: vec![FloodsubSubscription {
-                            topic: topic.hash().clone(),
+                            topic: topic,
```
tomaka:
```suggestion
                            topic,
```

rklaehn:
I think I have addressed all feedback. By the way: do you use cargo fmt? I tried running cargo fmt and got a huge diff....

PR #1398, opened 2020-01-20
Don't set KeepAlive to Until all the time

tomaka:
Right now whenever `poll()` of Kademlia is called, we set `keep_alive` to either `Until` or `Yes`.
The intent of this code code is that is updates `keep_alive` is response to something related to Kademlia happening.
Unfortunately, `poll()` can be called even if nothing happened.

I also changed the initial value of `keep_alive`. Right now it is set to `Yes` by default, but if nothing ever happens then that value will stay `Yes` forever and the connections will never die.

### DIFF
--- a/protocols/kad/src/handler.rs
+++ b/protocols/kad/src/handler.rs
@@ -645,7 +645,9 @@ where
     ) -> Poll<
         ProtocolsHandlerEvent<Self::OutboundProtocol, Self::OutboundOpenInfo, Self::OutEvent, Self::Error>,
     > {
-        let had_any_substream = !self.substreams.is_empty();
+        if self.substreams.is_empty() {
+            return Poll::Pending;
+        }
 
         // We remove each element from `substreams` one by one and add them back.
         for n in (0..self.substreams.len()).rev() {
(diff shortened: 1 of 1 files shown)


protocols/kad/src/handler.rs
```diff
@@ -645,6 +645,8 @@ where
     ) -> Poll<
         ProtocolsHandlerEvent<Self::OutboundProtocol, Self::OutboundOpenInfo, Self::OutEvent, Self::Error>,
     > {
+        let had_any_substream = !self.substreams.is_empty();
+
         // We remove each element from `substreams` one by one and add them back.
         for n in (0..self.substreams.len()).rev() {
             let mut substream = self.substreams.swap_remove(n);
```
romanb: I just realised that with the early return at the top, this loop probably needs to be adjusted for the match on `(None, Some(event), _) => {` to also do
```rust
if self.substreams.is_empty() {
    self.keep_alive = KeepAlive::Until(Instant::now() + Duration::from_secs(10));
}
```
before it does not put the substream back in the list and then `return`s from within the loop, which may otherwise result in a keep-alive previously set to `Yes` to remain so.

PR #1400, opened 2020-01-21
Remove all mod.rs files

tomaka:
As part of Rust edition 2018, you are now encouraged to use `foo.rs` rather than `foo/mod.rs`.
Half of the code of `rust-libp2p` was using the old idiom, and the other half the new idiom. This PR unifies all of this.


mxinden:
> As part of Rust edition 2018, you are now encouraged to use foo.rs rather than foo/mod.rs.

@tomaka can you cite a reference here? I could only find [doc.rust-lang.org](https://doc.rust-lang.org/nightly/edition-guide/rust-2018/module-system/path-clarity.html#no-more-modrs) and this [reddit article](https://www.reddit.com/r/rust/comments/b65ecn/modrs_in_rust_2018_yes_or_no/) but neither say that one is encouraged over the other.

:+1: for being consistent across the project, whichever style we use.

tomaka:
Ping for reviews?

PR #1405, opened 2020-01-23
Simplify trait bounds on NetworkBehaviour

tomaka:
This PR introduces several quality-of-life changes to `NetworkBehaviour` and `Swarm`:

- Removes the `type Substream;` associated type from `ProtocolsHandler`. Consequently, implementations of `ProtocolsHandler` and `NetworkBehaviour` no longer need a generic parameter (often called `TSubstream`).

- Consequently, one now has to implement `InboundUpgrade<NegotiatedBoxSubstream>` and `OutboundUpgrade<NegotiatedBoxSubstream>` rather than `InboundUpgrade<Self::Substream>` and `OutboundUpgrade<Self::Substream>`. `NegotiatedBoxSubstream` is a substream of a `StreamMuxerBox`.

- Removed the `TTransport` parameter from `Swarm`. The transport in the `Swarm` is now always a `BoxTransport<(TConnInfo, StreamMuxerBox), io::Error>`. The transport is automatically boxed when you build the swarm, and the user doesn't have to call `.map` and stuff.

- The `Send + 'static` bounds are now required on some types of the `ProtocolsHandler`.


tomaka:
A long time ago, I tried to make it possible to inject an "external" `NetworkBehaviour` into the Substrate network. It was deemed impossible, because ultimately you can't write `struct Foo<T: NetworkBehaviour<ProtocolsHandler::Substream = Bar>`. However I don't remember the exact process that I went through.

Before merging, I'd like to try re-apply to Substrate some changes, and remember exactly why it wasn't possible, and see if these changes now make it possible.

rklaehn:
Do you know how high the performance impact of the additional boxing is? It looks awesome from an usability POV, but it does introduce quite a lot of boxing and virtual calls.

tomaka:
> Do you know how high the performance impact of the additional boxing is?

My bet is that it is negligible. We unfortunately don't have any up-to-date benchmark, and it's probably quite hard debatable what to actually measure.

We could however easily (later? :pray:) add a benchmark that compares the throughput of using a transport directly vs using that transport through the swarm.

rklaehn:
The issue is probably not so much the actual boxing but the fact that due to the virtual call boundary there are some optimizations no longer happening that could be happening before. But my completely uneducated guess would also be that given that there are so many heavy things going on like encryption etc, the overhead will be very low unless you write lots of very tiny things over the wire, in which case you got other problems. Was just curious.

tomaka:
> However I don't remember the exact process that I went through.

So the thing that's been blocking this is:

- `Swarm<_, B>` requires `B` to implement `NetworkBehaviour`.

- If `B` is implemented through the custom derive, then `B` implements `NetworkBehaviour` only if all its members have a `type Substream;` in their protocols handler equal to the same value. Which means that if the fields of `B` are for example `C` and `Ping<Substream<StreamMuxerBox>>`, then `B` implements `NetworkBehaviour` only if `C` implements `NetworkBehaviour` with a substream equal to `Substream<StreamMuxerBox>`.

- Since `Swarm<_, B<C>>` is a field of a struct, then we would need to enforce that `C` implements `NetworkBehaviour` with a substream of `Substream<StreamMuxerBox>`, which isn't possible to express on a struct (structs don't have `where` fields).

This PR now makes this possible.

swarm/src/lib.rs
```diff
@@ -77,28 +77,33 @@ pub use protocols_handler::{
     SubstreamProtocol
 };
 
+
+pub type BoxSubstream = Substream<StreamMuxerBox>;
```
twittner: This name is a bit misleading I think. The `StreamMuxer` is boxed, but the `Substream` is not. Considering that this type alias is not used often, maybe we can do without it?

swarm/src/lib.rs
```diff
@@ -77,28 +77,33 @@ pub use protocols_handler::{
...
+pub type NegotiatedBoxSubstream = Negotiated<BoxSubstream>;
```
twittner: In light of the previous comment, maybe `pub type NegotiatedSubstream = Negotiated<Substream<StreamMuxerBox>>` could be an alternative?

tomaka:
I pushed two other commits that were necessary for Substrate.
It is now possible to "just" write `T: NetworkBehaviour` and use it in the `Swarm` with no additional bounds to enforce.

swarm/src/lib.rs
tomaka: It's a bit confusing. The `Substream<T>` alias is defined as `muxing::SubstreamRef<Arc<TMuxer>>`, and `SubstreamRef` holds a `Arc<TMuxer>` and one of the substreams.

Essentially, `BoxSubstream` contains an `Arc` and a `usize` representing the substream.

I'll remove both `BoxSubstream` and `Substream`.

tomaka:
Should be ready now.
I've had to add three new traits: `UpgradeInfoSend`, `InboundUpgradeSend` and `OutboundUpgradeSend`.

They are implemented automatically on types that implement respectively `UpgradeInfo`, `InboundUpgrade` and `OutboundUpgrade`. They are kind of similar to `TryFuture` and `TryStream` in the `futures` crate in this regard, in the sense that they exist for convenience.

twittner:
> > I've had to add three new traits: `UpgradeInfoSend`, `InboundUpgradeSend` and `OutboundUpgradeSend`.
> 
> Could you please explain why you had to do this and what the alternative would look like?

@tomaka?

tomaka:
The problem I encountered is that I want to write this:

```rust
trait ProtocolsHandler {
    type InboundProtocol: InboundUpgrade<Output: Send + 'static, Future: Send + 'static, Info: Send + 'static, InfoIter::Iterator: Send + 'static>;

    // (same for OutboundProtocol)

    // ...
}
```

But Rust doesn't let you do that.

There are three ways to bypass this problem:

- Enforce the trait bounds "outside", [which is what we're doing right now (before this PR)](https://github.com/libp2p/rust-libp2p/blob/7d0c80fce1e02076e2788be730e13022cdf4f874/swarm/src/lib.rs#L515-L527).
- Add the `Send`/`'static` requirements on the `InboundUpgrade` and `OutboundUpgrade` themselves.
- Introduce `UpgradeInfoSend`, `InboundUpgradeSend` and `OutboundUpgradeSend`, which are automatically implemented on the subset of types that implement `InboundUpgrade`/`OutboundUpgrade` and `Send`/`'static`.

The choice between the last two options is kind of debatable, but I went for the last option in order to not make `libp2p-core` too opinionated about `Send` and `'static`.
In a perfect world, `Send` and `'static` would be optional, which is why I'm reluctant to enforce them everywhere.

PR #1407, opened 2020-01-24
Fix WouldBlock being returned in wasm-ext

tomaka:
This got lost in translation when converting from the `Read` trait (where `WouldBlock` is the equivalent of `Pending`) to the `AsyncRead` trait.


tomaka:
cc @expenses

expenses:
LGTM

PR #1409, opened 2020-01-24
Add metadata to gossipsub

tomaka:
Otherwise I can't publish it.


tomaka:
I've already published a version with this PR in, so it's better get merged!

PR #1411, opened 2020-01-24
Fix some broken links in rustdoc

DemiMarie-parity:


DemiMarie-parity:
Workaround for https://github.com/rust-lang/rust/issues/64193

core/src/lib.rs
```diff
@@ -23,17 +23,26 @@
 //! The main concepts of libp2p-core are:
 //!
 //! - A [`PeerId`] is a unique global identifier for a node on the network.
-//!   Each node must have a different `PeerId`. Normally, a `PeerId` is the
+//!   Each node must have a different [`PeerId`]. Normally, a [`PeerId`] is the
```
romanb: It is often intentional that not every mentioning of a (linkable) name is actually a link. Specifically, it is a common practice to only link the first occurrence of such a name in a single paragraph or doc section, using `regular monospaced text` for subsequent occurrences. Too many links in a block of text, especially for the same name, can be quite irritating.

core/src/lib.rs
```diff
@@ -23,17 +23,26 @@
...
+//!   incoming remote connections. See the [`transport`] module.
 //! - The [`StreamMuxer`] trait is implemented on structs that hold a connection
 //!   to a remote and can subdivide this connection into multiple substreams.
-//!   See the `muxing` module.
+//!   See the [`muxing`] module.
 //! - The [`UpgradeInfo`], [`InboundUpgrade`] and [`OutboundUpgrade`] traits
 //!   define how to upgrade each individual substream to use a protocol.
 //!   See the `upgrade` module.
+//!
+//! [`PeerId`]: <struct.PeerId.html>
```
romanb: I don't think we should start using these kinds of links now, as they are hard to maintain and not currently used anywhere in rust-libp2p. We rely on intra-rustdoc-links. Bugs with the implementation of intra-rustdoc-links are unfortunate but I don't think it is worthwhile to maintain workarounds for them.

DemiMarie-parity: From what I can tell, intra-rustdoc links are quite broken.

PR #1412, opened 2020-01-25
protocols/kad: Implement S-Kademlia's lookup over disjoint paths

mxinden:
The extension paper S-Kademlia [1] includes a proposal for lookups over
disjoint paths. Within vanilla Kademlia, queries keep track of the
closest nodes in a single bucket. Any adversary along the path can thus
influence all future paths, in case they can come up with the
next-closest (not overall closest) hops. S-Kademlia tries to solve the
attack above by querying over disjoint paths using multiple buckets.

To adjust the libp2p Kademlia implementation accordingly this change-set
introduces the notion of `PathId`s into the `ClosestPeersIter` and the
ability to specify the amount of disjoint paths to pursue in the
`ClosestPeersIterConfig`.

`ClosestPeersIter` explores each path in a round-robin fashion. It
iterates the not-yet-contacted nodes one by one starting with the
closest one. When the given peer origins from the initial set or from a
past query with the path id currently selected, it is tagged with that
path id and returned to the upper layer to be queried.

Once iterated through all peers without success, one chooses the one
with the lowest distance from the target detached from its path id (a
bit like a work-steeling queue).

Setting the `disjoint_paths` to `1` within the `ClosestPeersIterConfig`
effectively disables the feature.

[1] http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.68.4986&rep=rep1&type=pdf

### Status

This is still a work-in-progress pull request. I think the overall design of the change is already worth discussing (e.g. do we want this in the first place, should this be a separate `ClosestPeersIter`, ...). This is the least intrusive change, probably not as resilient as one could be and definitely not complete in regards to @infinity0 [research](https://forum.web3.foundation/t/s-kademlia-and-multipaths/310). There is still a borrowing issue, some clean-up and plenty more tests to write.

### DIFF
--- a/protocols/kad/src/query/peers/closest.rs
+++ b/protocols/kad/src/query/peers/closest.rs
@@ -325,7 +325,7 @@ impl ClosestPeersIter {
                         *cnt += 1;
                         // If `num_results` successful results have been delivered for the
                         // closest peers, the iterator is done.
-                        if *cnt > self.config.num_results {
+                        if *cnt >= self.config.num_results {
                             self.state = State::Finished;
                             return PeersIterState::Finished
                         }
--- a/protocols/kad/src/query/peers/closest.rs
+++ b/protocols/kad/src/query/peers/closest.rs
@@ -278,24 +278,23 @@ impl ClosestPeersIter {
         self.waiting().any(|p| peer == p)
     }
 
-
     /// Advances the state of the iterator, potentially getting a new peer to contact.
     pub fn next(&mut self, now: Instant) -> PeersIterState {
         if let State::Finished = self.state {
             return PeersIterState::Finished
         }
 
-        // Count the number of peers that returned a result. If there is a
-        // request in progress to one of the `num_results` closest peers, the
-        // counter is set to `None` as the iterator can only finish once
-        // `num_results` closest peers have responded (or there are no more
-        // peers to contact, see `num_waiting`).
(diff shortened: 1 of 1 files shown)


protocols/kad/src/query/peers/closest.rs
```diff
@@ -296,32 +320,72 @@ impl ClosestPeersIter {
...
                     }
                 }
-
                 PeerState::Succeeded =>
                     if let Some(ref mut cnt) = result_counter {
                         *cnt += 1;
                         // If `num_results` successful results have been delivered for the
                         // closest peers, the iterator is done.
-                        if *cnt >= self.config.num_results {
+                        if *cnt > self.config.num_results {
```
mxinden: Given that I first iterate all peers looking for timeouts and successes, the function does not behave as before anymore even when `disjoint_paths` is set to `1` (disabled). Only tests failing without the line above are `put_record` and `add_provider`. I still need to investigate this further.

infinity0: TBH for simplicity I would not even keep the old behaviour as it is obviously insecure and I can't see a use case for ever not enabling disjoint paths, even for testing/debugging purposes.

mxinden:
Closing in favor of https://github.com/libp2p/rust-libp2p/pull/1473.

PR #1413, opened 2020-01-27
Finish PeerId inlining change

tomaka:
Supersedes #1230 

Adds custom implementations of `PartialEq` and `Hash` for `PeerId`s so that they compare equal between hash algorithm.

This is intended to be removed at some point, because the hash algorithm should be derived in a predictable way from the public key.


tomaka:
Substrate works after this change.

tomaka:
> Any particular reason why you did not use the property test from #1230 (comment)?

I had forgotten about it.

PR #1420, opened 2020-02-03
Ipfs private swarms interop example

rklaehn:
Add small example for interop with public or private ipfs nodes via pubsub and ping

The only notable change outside the example is that I have allowed parsing ipfs addresses such as `/ip4/104.131.131.82/tcp/4001/ipfs/QmaCpDMGvV2BGHeYERUEnRQAwe3N8SzbUtfsmvsqQLuvuJ` as a multiaddr. I don't see a big downside of this, but I am also happy to remove it again.

If you find the either transport sufficiently useful I can integrate it with core, but for now I have just left it in the example.


tomaka:
> I don't see a big downside of this, but I am also happy to remove it again.

A year or so ago we went the other way and renamed "ipfs" to "p2p".
"ipfs" is still supported for backwards compatibility reasons in the Go and JS implementations, but we went for not supporting it in Rust as there was no such concern.

rklaehn:
> > I don't see a big downside of this, but I am also happy to remove it again.
> 
> A year or so ago we went the other way and renamed "ipfs" to "p2p".
> "ipfs" is still supported for backwards compatibility reasons in the Go and JS implementations, but we went for not supporting it in Rust as there was no such concern.

I think protocol labs will also switch to `p2p`. This is just so that all the /ipfs/ addresses out there are supported. E.g. if you run the example with some addresses from your ipfs bootstrap list, like

```
cargo run --example ipfs-private -- /ip4/104.131.131.82/tcp/4001/ipfs/QmaCpDMGvV2BGHeYERUEnRQAwe3N8SzbUtfsmvsqQLuvuJ
```

If you don't want to support ipfs links I will sanitize them another way, no problem.

rklaehn:
On the subject of whether to support /ipfs/, it is specified as identical to /p2p/ in [here](https://github.com/multiformats/multiaddr/blob/master/protocols.csv).

tomaka:
> I don't think there is a downside in supporting it.

The downside is that we can't ever remove support for it without potentially breaking existing code.

rklaehn:
> > I don't think there is a downside in supporting it.
> 
> The downside is that then we can't ever remove support for it without potentially breaking existing code.

That is obvious. But since it is exactly the same as p2p, which you are going to support anyway, the only change is the additional string "ipfs" that gets recognized. And since https://github.com/multiformats/multiaddr/blob/master/protocols.csv seems to be the standard format table, the protocol name is already reserved forever.

But I also understand if you don't want to support it. In that case I will just do a sanitizer that replaces /p2p/ with /ipfs/ before converting the string to a multiaddr. Just let me know what you want.

examples/ipfs-private.rs
```diff
@@ -0,0 +1,354 @@
...
+        loop {
+            match stdin.try_poll_next_unpin(cx)? {
+                Poll::Ready(Some(line)) => {
+                    swarm.gossipsub.publish(&gossipsub_topic, line.as_bytes());
+                }
+                Poll::Ready(None) => panic!("Stdin closed"),
+                Poll::Pending => break,
+            }
+        }
+        loop {
```
rklaehn: @tomaka This is as good a place as any to ask this: would it be also an option to use `select!` on these two streams, or is it preferable to poll the swarm until it returns Poll::Pending in a loop?

I have been wondering if using `select!` as mentioned here https://github.com/ipfs-rust/rust-ipfs/blob/master/src/lib.rs#L268 is actually a good idea?

michaelvoronov: Based on the comment from the provided link: hmm, why not just use `swarm.next().fuse()` in select macro? I have successfully used it our project like this: 
```rust
    task::spawn(async move {
        loop {
            select! {
                from_peer = peer_service_out_receiver.next().fuse() => {
                    match from_peer {
                    }
                },

                from_swarm = node_service.swarm.next().fuse() => {
                    match from_swarm {
                    }
                },

                _ = exit_receiver.next().fuse() => {
                    break
                }
            }
        }
    });
```

Also I find this approach more expressive then several loops. And also wandering why not use such approach with `select!` at least in rust-libp2p examples instead of loops?

rklaehn: That is what I am currently doing in some mini ipfs like PoC at work, except that I am `select!` ing on `swarm.next_event().fuse()` and some other stream.

It seems to work, but I just wanted to make sure that the overall approach is sound and does not just work by chance/accident.

I think it would be useful to have a fused stream of swarm events for use within select, and to make the various next() methods fused... See https://github.com/libp2p/rust-libp2p/issues/1434 and https://github.com/libp2p/rust-libp2p/pull/1433 ...

examples/ipfs-private.rs
```diff
@@ -0,0 +1,354 @@
...
+    task::{Context, Poll},
+};
+
+/// An Either combination of two transports
+///
+/// this is useful if you have a different transport depending on configuration.
+/// In this example we have a different transport type depending on whether private swarms
+/// are enabled or not.
+#[derive(Debug, Copy, Clone)]
+pub enum EitherTransport<A, B> {
```
twittner: We should probably get this into `either.rs`. @tomaka, what do you think?

rklaehn:
@twittner that is actually a good point. Will remove the ipfs legacy thing and fix the string before conversion...

examples/ipfs-private.rs
tomaka: Agree

rklaehn: Moved it as it is to core/src/either.rs. I think we don't need .left() and .right() methods on the transport trait.

rklaehn:
Removed the change to multiaddr and did the sanitization just in the example.

rklaehn:
I adapted this to the latest master, removing some type constraints. Was less painful than expected. Looks much simpler now!

protocols/pnet/src/lib.rs
```diff
@@ -161,6 +161,22 @@ pub enum KeyParseError {
...
 
+impl fmt::Display for KeyParseError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{:?}", self)
+    }
+}
+
+impl error::Error for KeyParseError {
+
+    fn cause(&self) -> Option<&dyn error::Error> {
```
tomaka:
```suggestion
    fn source(&self) -> Option<&(dyn error::Error + 'static)> {
```

protocols/pnet/src/lib.rs
```diff
@@ -161,6 +161,22 @@ pub enum KeyParseError {
...
 }
 
+impl fmt::Display for KeyParseError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{:?}", self)
+    }
+}
+
+impl error::Error for KeyParseError {
+
```
tomaka:
```suggestion
```

examples/ipfs-private.rs
michaelvoronov: It seems that implementing `FusedStream` for Swarm would make sense for optimization - `Fuse::poll_next` contains a little overhead (some if's that could lead to cache misses) for managing `self.done`. But in all other aspects `swarm.next().fuse()` is good.

PR #1427, opened 2020-02-05
Make errors on listener non-fatal

tomaka:
In #1218, we decided to make listener errors non-fatal.
There was however something missing, and errors were actually fatal.
This fixes it.


tomaka:
> effectively this means that listeners must never fail.

I think we debated this in the past at the time of #1218, and the conclusion was that yes, listeners must never fail. There isn't really a fundamental reason why a listener would fail (as an example, the TCP one doesn't), and in these rare situations I think it's fine to log an error and either return `None` or return `Poll::Pending` all the time.

twittner:
If listeners must not fail, would this not better be expressed by a `Stream<Item = ListenerEvent<Self::ListenerUpgrade>>` or, if they should be able to report errors, by a `Stream<Item = ListenerEvent<Self::ListenerUpgrade, Self::Error>>`?

tomaka:
I was almost finished with the changes when I realized a problem:
`TcpListener::bind` is asynchronous, and if it fails we return the error as part of the listener.
In this specific situation, it should indeed be fatal.

twittner:
> Should we go all the way and make `listen_on` return a `Future` to a `Stream`?

Do you mean
```rust
Future<Output = Result<Stream<Item = ListenerEvent<Self::ListenerUpgrade, Self::Error>>, Self::Error>
```
?

tomaka:
Yes. It seems like a lot of complexity and would require adding yet another associated type to `Transport`.

twittner:
I am in favour of this proposal.

tomaka:
I reworked this PR to instead add a `ListenerEvent::Error` variant that is non-fatal.

core/src/nodes/listeners.rs
```diff
@@ -269,15 +269,17 @@ where
...
+                    let id = *listener_project.id;
+                    self.listeners.push_front(listener);
+                    return Poll::Ready(ListenersEvent::Error {
+                        listener_id: id,
+                        error,
                     })
                 }
-                Poll::Ready(Some(Err(err))) => {
-                    return Poll::Ready(ListenersEvent::Error {
+                Poll::Ready(None) | Poll::Ready(Some(Err(_))) => {
```
twittner: I would be nice if we could log this error should it occur.

core/src/transport.rs
```diff
@@ -318,6 +343,27 @@ impl<T> ListenerEvent<T> {
             None
         }
     }
+
+    /// Returns `true` if this is an `Error` listener event.
+    pub fn is_non_fatal_error(&self) -> bool {
```
twittner:
```suggestion
    pub fn is_error(&self) -> bool {
```

core/src/transport.rs
```diff
@@ -318,6 +343,27 @@ impl<T> ListenerEvent<T> {
...
+        } else {
+            false
+        }
+    }
+
+    /// Try to turn this listener event into the `Error` part.
+    ///
+    /// Returns `None` if the event is not actually a `Error`,
+    /// otherwise the error.
+    pub fn into_non_fatal_error(self) -> Option<TErr> {
```
twittner:
```suggestion
    pub fn into_error(self) -> Option<TErr> {
```

core/src/transport/dummy.rs
```diff
@@ -56,7 +56,7 @@ impl<TOut> Clone for DummyTransport<TOut> {
 impl<TOut> Transport for DummyTransport<TOut> {
     type Output = TOut;
     type Error = io::Error;
-    type Listener = futures::stream::Pending<Result<ListenerEvent<Self::ListenerUpgrade>, io::Error>>;
+    type Listener = futures::stream::Pending<Result<ListenerEvent<Self::ListenerUpgrade, io::Error>, io::Error>>;
```
twittner:
```suggestion
    type Listener = futures::stream::Pending<Result<ListenerEvent<Self::ListenerUpgrade, Self::Error>, Self::Error>>;
```

transports/tcp/src/lib.rs
```diff
@@ -93,7 +93,7 @@ impl $tcp_config {
 impl Transport for $tcp_config {
     type Output = $tcp_trans_stream;
     type Error = io::Error;
-    type Listener = Pin<Box<dyn Stream<Item = Result<ListenerEvent<Self::ListenerUpgrade>, io::Error>> + Send>>;
+    type Listener = Pin<Box<dyn Stream<Item = Result<ListenerEvent<Self::ListenerUpgrade, Self::Error>, io::Error>> + Send>>;
```
twittner:
```suggestion
    type Listener = Pin<Box<dyn Stream<Item = Result<ListenerEvent<Self::ListenerUpgrade, Self::Error>, Self::Error>> + Send>>;
```

PR #1430, opened 2020-02-05
Try to use GitHub actions

tomaka:


tomaka:
Well, this didn't trigger any action. I guess it's maybe because the actions have to be in the upstream repo? But how would you then modify them and know that they work?

dignifiedquire:
@tomaka I had the same issue before, you need to merge a dummy thing, or just your first pass, to `master` and then you can iterate in PRs

tomaka:
My intent when opening this PR was to check whether it was possible to replicate the current CI with GitHub Actions.
Looks like we would have to actually merge something in order to try that :grimacing:

PR #1433, opened 2020-02-06
Implement FusedStream for Swarm

rklaehn:
The stream for swarm will never terminate, and therefore does not have to
keep track of termination.

Implementing Fused markers can be done for several other futures as well, e.g. next() and next_event(). However, this would require turning them into structs that implicitly provide Future instead of async fns, which means a sea of type constraint boilerplate. Not sure if it is worth it...

### DIFF
--- a/swarm/src/lib.rs
+++ b/swarm/src/lib.rs
@@ -85,7 +85,7 @@ pub use protocols_handler::{
 pub type NegotiatedSubstream = Negotiated<Substream<StreamMuxerBox>>;
 
 use protocols_handler::{NodeHandlerWrapperBuilder, NodeHandlerWrapperError};
-use futures::{prelude::*, executor::{ThreadPool, ThreadPoolBuilder}};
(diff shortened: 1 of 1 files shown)


rklaehn:
@mxinden agree that the complexity is regrettable - although it is not so much complexity as boilerplate.

The issue is that once you call .fuse() you no longer have the swarm directly available, but have to call .get_mut() on the fused stream, which makes things more ugly than it has to be, especially since swarm already is a fused stream...

mxinden:
I don't have an opinion here. What do others think?

michaelvoronov:
Having some experience of using `rust-libp2p` I think that it is worth it. These 38 lines of code gives better user experience and provide a little optimization by removing nessecerity of using `poll_next` wrapper from the `Fuse` struct. Yes, this overhead is very tiny, but for someone these several `if`'s could matter (especially considering that this is a library).

rklaehn:
This looks so much nicer now that the type constraints are simpler!

PR #1438, opened 2020-02-06
Add a limit to the number of incoming connections

tomaka:
Right now the `Swarm` has an "incoming limit" system that allows enforcing a limit to the number of incoming connections.

I think that this is bad design, and this PR instead implements a layer around the `Transport` trait that serves the same purpose.

Opening as a draft because the test never finishes. There's probably an issue with waking tasks.

### DIFF
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -28,6 +28,7 @@ prost = "0.6.1"
 rand = "0.7"
 rw-stream-sink = { version = "0.2.0", path = "../misc/rw-stream-sink" }
 sha2 = "0.8.0"
+slab = "0.4.2"
 smallvec = "1.0"
 thiserror = "1.0"
 unsigned-varint = "0.3"
--- a/core/src/transport/incoming_limit.rs
+++ b/core/src/transport/incoming_limit.rs
@@ -350,7 +350,7 @@ mod tests {
                                 let mut n = 0;
                                 future::poll_fn(|cx| {
                                     n += 1;
-                                    if n >= 3 {
+                                    if n >= 50 {
                                         Poll::Ready(())
                                     } else {
                                         cx.waker().wake_by_ref();
--- a/core/src/transport/incoming_limit.rs
+++ b/core/src/transport/incoming_limit.rs
@@ -1,0 +1,372 @@
+// Copyright 2020 Parity Technologies (UK) Ltd.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a
+// copy of this software and associated documentation files (the "Software"),
+// to deal in the Software without restriction, including without limitation
+// the rights to use, copy, modify, merge, publish, distribute, sublicense,
+// and/or sell copies of the Software, and to permit persons to whom the
+// Software is furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS
+// OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+// FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+// DEALINGS IN THE SOFTWARE.
+
+//! Applies a limit to the number of simultaneous incoming connections of
+//! a transport.
+//!
+//! The limit is enforced by not polling the underlying listener.
+//!
+//! By design, we don't hold any pending undelivered incoming connection. In other words, if the
+//! underlying listener produces a connection then we always immediately deliver it.
+//!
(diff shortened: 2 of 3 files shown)


core/src/transport/incoming_limit.rs
```diff
@@ -0,0 +1,372 @@
...
+{ 
+    type Output = TInner::Output;
+
+    fn poll(self: Pin<&mut Self>, cx: &mut Context) -> Poll<Self::Output> {
+        let this = self.project();
+        this.inner.poll(cx)
+    }
+}
+
+#[pin_project::pinned_drop]
```
mxinden: Cool! Wasn't aware of `pin_project::pinned_drop`.

core/Cargo.toml
```diff
@@ -28,6 +28,7 @@ prost = "0.6.1"
 rand = "0.7"
 rw-stream-sink = { version = "0.2.0", path = "../misc/rw-stream-sink" }
 sha2 = "0.8.0"
+slab = "0.4.2"
```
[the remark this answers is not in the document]
tomaka: It's not *really* an optimization. It's an alternative to `HashMap` except you don't get to choose the keys. I don't need any of the properties of `HashMap` here.

tomaka:
So the test was actually not passing because `futures::pending!()` doesn't do what I expected. Fixed now.

tomaka:
Let me know if the implementation is too complicated.

core/Cargo.toml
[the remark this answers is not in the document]
mxinden: Introducing a new dependency in order to use a data structure with less features even though we don't know whether having those features would result in any performance impact seems like a premature optimization to me.

tomaka:
It's kind of hard to benchmark this, as the "optimization" consists in making it possible to run `poll` multiple times concurrently. In other words, it's not an optimization in the sense that we replace slow code by faster code, but a gain made possible by changing the behaviour of the code.

The expected performance difference therefore depend almost entirely on how much parallelism we expect and how long `poll` takes to call.

mxinden:
> It's kind of hard to benchmark this, as the "optimization" consists in making it possible to run poll multiple times concurrently. In other words, it's not an optimization in the sense that we replace slow code by faster code, but a gain made possible by changing the behaviour of the code.

Then my suggestion would be to go with the *slow* version and replace it with the *fast* version in case it turns out to be a performance bottleneck in future benchmarking.

> We could obviously benchmark slab vs HashMap for example, but that should be a pretty minor difference and to me the expected performance depends almost entirely on how much parallelism we expect and how long poll takes to call.

If you expect only a minor difference, then I would suggest using a HashMap for the sake of less dependencies and faster compile times.

---

What do others think? Feel free to ignore my comments in case you think the performance improvements are worth the complexity.

core/src/transport/incoming_limit.rs
```diff
@@ -0,0 +1,401 @@
...
+
+/// Reservation for a slot towards the limit.
+struct NumApproxGuard<'a> {
+    limit: &'a IncomingLimit,
+}
+
+impl<'a> NumApproxGuard<'a> {
+    fn acquire(limit: &'a IncomingLimit) -> (Self, usize) {
+        let current_num = limit.current_num_approx.fetch_add(1, Ordering::Acquire);
+        debug_assert_ne!(current_num, usize::max_value()); // Check for overflows.
```
twittner: Debug builds already check for overflows. Why this additional check here?

core/src/transport/incoming_limit.rs
```diff
@@ -0,0 +1,401 @@
...
+        })))
+    }
+}
+
+#[pin_project::pinned_drop]
+impl<TInner> PinnedDrop for IncomingLimitListen<TInner> {
+    fn drop(self: Pin<&mut Self>) {
+        let this = self.project();
+        if let Some(waker_slot_num) = this.waker_slot_num {
+            this.limit.inner.lock().wakers.remove(*waker_slot_num);
```
twittner: Should we wake up any `Waker` that is stored in here?

core/src/transport/incoming_limit.rs
```diff
@@ -0,0 +1,401 @@
...
+                return Poll::Ready(Some(Ok(ListenerEvent::AddressExpired(addr)))),
+            Poll::Ready(Some(Err(err))) =>
+                return Poll::Ready(Some(Err(err))),
+            Poll::Ready(Some(Ok(ListenerEvent::Upgrade { upgrade, local_addr, remote_addr }))) => {
+                (upgrade, local_addr, remote_addr)
+            }
+        };
+
+        // Properly reserve our slot. Requires a mutex lock.
+        mem::forget(_guard);
```
twittner: Should we add a comment stating that at this point we do not want the destructor to run to not decrement the counter since we now have a new incoming connection?

core/src/transport/incoming_limit.rs
```diff
@@ -0,0 +1,401 @@
...
+#[must_use]
+#[pin_project::pin_project(PinnedDrop)]
+pub struct IncomingLimitUpgrade<TInner> {
+    #[pin]
+    inner: TInner,
+
+    /// True if we already free'd a slot in `limit`.
+    cleaned_up: bool,
+
+    /// This struct implicitely holds a slot in both `current_num` and `current_num_approx`.
```
twittner:
```suggestion
    /// This struct implicitly holds a slot in both `current_num` and `current_num_approx`.
```

core/src/transport/incoming_limit.rs
tomaka: For atomics, the documentation explicitly says that it wraps around on overflow.

romanb:
I didn't look at the implementation details, but is this PR still relevant given the built-in configurable connection limits that are now in `libp2p-core` (including an incoming connection limit)?

tomaka:
I don't really have a plan to come back to this PR, so closing it.

PR #1439, opened 2020-02-07
Switch noise from the RingResolver to the DefaultResolver

### LINKED ISSUE #1414: Make libp2p-noise compile for wasm-in-the-browser
Secio already compiles for the browser, so let's do the same for noise.

expenses:
This is the easiest way to fix #1414. The comment saying that the default resolver uses `rust-crypto` appears to be out-of-date, but in the event that we don't want to use it's current dependencies, we could use an updated fork of https://github.com/mcginty/snow/pull/20 instead.


protocols/noise/Cargo.toml
```diff
@@ -16,7 +16,7 @@ log = "0.4"
 prost = "0.6.1"
 rand = "0.7.2"
 ring = { version = "0.16.9", features = ["alloc"], default-features = false }
```
tomaka:
```suggestion
```

twittner:
I would not want to see `ring` replaced with a set of rust crypto crates which have not been updated for many years and are barely used, like for instance `chacha20-poly1305-aead`. This does not necessarily mean that those crates are broken, but when it comes to crypto crates I would prefer to depend on widely used ones.

twittner:
As a way forward we should either get the libsodium-based resolver into snow first and use it or use the default resolver only for wasm, or experiment with https://crates.io/crates/noise-protocol which already has a sodiumoxide backend.

expenses:
I'm happy to either `use the default resolver only for wasm` or to `get the libsodium-based resolver into snow`.

protocols/noise/Cargo.toml
expenses: We're currently only using `ring` directly for SHA256. Should i switch the code to the `sha2` crate?

burdges:
I think both ring and libsodium sound good.  It's possible ring's constant-time protections keep pace with intel's ongoing shit show slightly better, so keep support for ring even if you add libsodium.   

I think https://github.com/RustCrypto/AEADs/tree/master/chacha20poly1305 and https://github.com/RustCrypto/hashes/tree/master/blake2 along with x25519-dalek sound fine.  I think RustCrypto crates are slower than libsodium and lack some constant-time protections, so sure restrict them to the browser, but they make an effort and they're maintained anyways.  I'd assume ring is less vulnerable to timing attacks than dalek, but dalek makes a good effort, except for not covering all Intel's SGX vulnerabilities.

I donno if its easier to just add libsodium or use the chacha20poly1305 and blake2 crates, but either sounds fine.

expenses:
Ok, I've created https://github.com/mcginty/snow/pull/75 but sodiumoxide doesn't support wasm at the moment (not sure how hard this would be to do).

expenses:
And https://github.com/mcginty/snow/pull/76. @burdges do you mean that we should switch snow from `blake2-rfc` to `blake2`?

burdges:
I think blake2 is a fork of blake2-rfc but blake2 has seen zero commits that improve the crypto so far:  https://github.com/RustCrypto/hashes/commits/master/blake2/src 

I'd expect all RustCrypto crates should improve past the "stagnant" crates from which they were forked after https://github.com/rust-lang/rfcs/pull/2859 lands and stabilizes.

I've always found the RustCrypto traits kinda painful due to using GenericArray, so I personally avoid updating code using crates like blake2-rfc to blake2 until after const generics land and stabilizes.  It's might not be that painful for this case however, so maybe an upgrade now works. 

I'm ambivalent.  :)

expenses:
Well, it was easy enough to switch from `blake2-rfc` to `blake2` so I'll make a PR for that anyway. If those PRs got merged into snow, would you be happy enough to use snow with the default resolver on wasm, while keeping the ring resolver on native?

burdges:
Sure. Yes keep ring whenever it works, but replace otherwise. 

I think @twittner brought up chacha20poly1305-aead specifically.  I'd assume the chacha20poly1305 crate's poly1305 was lifted from chacha20poly1305-aead, but I'm less sure the chacha20 crate's genesis although the salsa one they did themselves:  https://github.com/RustCrypto/stream-ciphers/commits/b71cf742112a1c62904fcdfaeabed0db5f7ccf30/salsa

Any big shortcomings of chacha20poly1305 should be fixed eventually since Tony Arcieri helps with this crate.  If you're switching them then maybe switch chacha20poly1305-aead over to chacha20poly1305 too.

tomaka:
For what it's worth I'm happy with whatever changes compiles to WASM and is compatible with the existing network.
I don't have the bandwidth/motivation to dig into cryptographic libraries quality.

expenses:
Should be ready for review now.

tomaka:
Any opinion about the PR now @twittner @burdges ?

expenses:
Yeah, I'm gonna finish that, but most for completion's sake. Sodiumoxide doesn't currently support wasm, so we have to use the native resolver on wasm for now.

protocols/noise/Cargo.toml
```diff
@@ -15,8 +15,8 @@ libp2p-core = { version = "0.15.0", path = "../../core" }
 log = "0.4"
 prost = "0.6.1"
 rand = "0.7.2"
-ring = { version = "0.16.9", features = ["alloc"], default-features = false }
-snow = { version = "0.6.1", features = ["ring-resolver"], default-features = false }
+sha2 = "0.8.0"
+snow = { version = "0.6.1", features = ["default-resolver", "ring-resolver"], default-features = false }
```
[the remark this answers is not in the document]
expenses: Yeah, that works.

expenses:
Are we good to merge this?

Issue #912, opened 2019-02-03
Allow multiple simultaneous connections to the same peer


tomaka:
Right now, in order to simplify the implementations of `NetworkBehaviour`, we only have allow connection per peer ID.
Allowing multiple connections would, in my opinion, considerably increase the difficulty of implementing this trait. If we provide helpers for `NetworkBehaviour`, though, we could allow multiple connections per peer.

Note that we should have ways to properly identify each connection. Having multiple connections shouldn't be implicit.

tomaka:
> Allowing multiple connections would, in my opinion, considerably increase the difficulty of implementing this trait. If we provide helpers for `NetworkBehaviour`, though, we could allow multiple connections per peer.

I just realized that is not necessarily true. We can tell the behaviour that "connected" means "connected for the first time", and "disconnected" means "zero connections remaining".
There could be either one `NodeHandler` for all the connections, or one `NodeHandler` per connection, I'm not sure.

tomaka:
Bumping to "important". See https://github.com/paritytech/substrate/issues/4272 for context.

romanb:
This is now in `libp2p-core`, but `libp2p-swarm` does not yet permit explicitly creating multiple connections per peer. I guess this ticket should stay open until `libp2p-swarm` fully "embraces" the notion of multiple connections per peer?

romanb:
Considered closed with https://github.com/libp2p/rust-libp2p/pull/1519.

Issue #925, opened 2019-02-06
Add some dialing parallelism


tomaka:
Right now we can't dial multiple addresses belonging to the same peer at the same time.
Allowing this means solving the problem of what happens if we successfully open multiple connections, which is similar to the simultaneous dialing problem of #907.

romanb:
Closed by https://github.com/libp2p/rust-libp2p/pull/1506.

PR #1440, opened 2020-02-07
Multiple connections per peer

romanb:
## Motivation

This is a proposal for https://github.com/libp2p/rust-libp2p/issues/912. Instead of trying to enforce a single connection per peer, which involves quite a bit of additional complexity e.g. to prioritise simultaneously opened connections and can have other undesirable consequences (e.g. paritytech/substrate#4272) besides potentially prohibiting other valid use-cases where multiple connections per peer are desired, this PR makes multiple connections per peer a feature in `libp2p-core`.

## Overview

The concept of a "node" with an implicit 1-1 correspondence to a connection has been replaced with the "first-class" concept of a "connection". The code from `src/nodes` has moved (with varying degrees of modification) to `src/connection` and `src/network`. A `HandledNode` has become a `Connection`, a `NodeHandler` a `ConnectionHandler`, the `CollectionStream` was the basis for the new `connection::Pool`, and so forth.

Conceptually, a `Network` contains a `connection::Pool` which in turn internally employs the `connection::Manager` for handling the background `connection::manager::Task`s, one per connection, as before. These are all considered implementation details. On the public API, `Peer`s are managed as before through the `Network`, except now the API has changed with the shift of focus to (potentially multiple) connections per peer. The `NetworkEvent`s have accordingly also undergone changes.

## Backward Compatibility

The API of `libp2p-core` obviously has changed significantly in detail and semantics, though not so much in the structure (there is still the `Network` emitting events and `Network::peer()` for managing peers and almost all existing functions have been preserved, though often with changed semantics due to the nature of permitting multiple connections).

The API of `libp2p-swarm` should be backward compatible. There is the subtle semantic change that a single `NetworkBehaviour` can now be associated with many `ProtocolsHandler` instances for the same peer due to multiple connections, ~~but unless an implementation makes some hard assumptions about such a 1-1 correspondence between peers and `ProtocolsHandler` instances, which seems unlikely, existing implementations should remain largely unaffected~~ (**EDIT**: This is incorrect and needs to be addressed. See [this comment](https://github.com/libp2p/rust-libp2p/pull/1440#issuecomment-584041814)). `libp2p-swarm` still does not by itself create multiple connections per peer (i.e. a `DialPeer` received from a behaviour only results in a connection attempt if the peer is currently disconnected, whereas `libp2p-core` now also allows additional `connect()`s for already connected peers), it just may occur as a result of simultaneous dialing. Lastly, `inject_connected` and `inject_disconnected` are only called when the first connection is established, respectively when the last connection is lost to a peer, to preserve the existing semantics. `inject_replaced` is no longer called, since connections are no longer replaced and has thus been deprecated.

## Dialing Concurrency

While these changes permit multiple established connections per peer, what I have preserved is the constraint to a single pending outgoing (i.e. "dialing") connection per peer at a time. I think that is still desirable, especially because while dialing is in progress, new addresses to try can be added, e.g. as was and is still done by `libp2p-swarm`.

## Additions

  * Instead of just an `incoming_limit`, there is now also a configurable outgoing limit (pending outgoing connections to _any_ peer) and a configurable limit for the established connections per peer. These limits are enforced by the connection `Pool`, i.e. the incoming limit is no longer checked on every invocation of `Network::poll()` to determine if the listeners stream should be polled but rather at the time when they are added to the pool and whenever a pending connection is established. The defaults are no limits for any of these options. Furthermore these are currently hard limits, i.e. there is no throttling based on delays and additional queuing involved.
  * There is a `NetworkInfo` obtained by `Network::info()` containing some basic connection statistics from the connection pool.
  * There is a `NetworkConfig` for configuring the task executor as well as the connection pool limits.

## How to Review

It seems most sensible to first look through the new / adapted modules in `src/connection/**` and `src/connections.rs`, followed by `src/network/**.` and `src/network.rs` and lastly looking at the actual diffs for `libp2p-swarm`. A lot of the code and modules in `src/connection` and `src/network` will look familiar as it stems from previous `src/nodes` modules.

## Remaining Work

At this point I'm only planning to go over the documentation again,think about new tests and verify with the current polkadot/substrate (e.g. that it indeed addresses paritytech/substrate#4272), but there are no major changes on my backlog other than those that may result from feedback here or from issues encountered in further testing, hence I'm opening this now for review and general feedback.


romanb:
> There is the subtle semantic change that a single NetworkBehaviour can now be associated with many ProtocolsHandler instances for the same peer due to multiple connections, but unless an implementation makes some hard assumptions about such a 1-1 correspondence between peers and ProtocolsHandler instances, which seems unlikely, existing implementations should remain largely unaffected.

After taking a look at `libp2p-kad` again, I think this is very much incorrect. Indeed, likely any request-reply protocol requires the reply to be sent on the same substream and connection as the original request. That is what e.g. `libp2p-kad` uses the `KademliaRequestId`s for which get assigned to incoming requests and are sent back by the behaviour to the handler with the reply.

The least intrusive solution that comes to mind is to let `ProtocolsHandler::inject_event` (and possibly `ConnectionHandler::inject_event` as well) signal in the return value whether the given event was accepted / handled. The `Swarm` can then iterate over the connections to the peer and call `inject_event` until a connection (handler) accepts it, otherwise drop it.

The alternatives that come to mind involve making a `NetworkBehaviour` aware of `ConnectionID`s, which is likely more intrusive and may be a change that goes well together with removing the `ProtocolsHandler` altogether later.

tomaka:
> The alternatives that come to mind involve making a NetworkBehaviour aware of ConnectionIDs, which is likely more intrusive and may be a change that goes well together with removing the ProtocolsHandler altogether later.

I would go more in this direction in general.

romanb:
> > The alternatives that come to mind involve making a NetworkBehaviour aware of ConnectionIDs, which is likely more intrusive and may be a change that goes well together with removing the ProtocolsHandler altogether later.
> 
> I would go more in this direction in general.

While this is clearly where we are heading in general, I'd really like to keep the changes and especially API changes outside of `libp2p-core` to a minimum in this first iteration.

core/src/connection/manager.rs
```diff
@@ -0,0 +1,479 @@
...
+            InEvent = I,
+            OutEvent = O,
+            Error = HE
+        > + Send + 'static,
+        <H::Handler as ConnectionHandler>::OutboundOpenInfo: Send + 'static,
+    {
+        let task_id = self.next_task_id;
+        self.next_task_id.0 += 1;
+
+        let (tx, rx) = mpsc::channel(4);
```
mxinden: Out of curiosity, what is the number 4 based on? Why not just `channel(0)`, given that each sender is guaranteed a slot?

core/src/connection/manager.rs
```diff
@@ -0,0 +1,479 @@
...
+            match &task.state {
+                TaskState::Pending => None,
+                TaskState::Established(c) => Some(c),
+            }
+        } else {
+            None
+        }
+    }
+
+    /// Provides an API similar to `Stream`, except that it cannot produce an error.
```
mxinden: [futures 0.3.1 `Stream`](https://docs.rs/futures/0.3.1/futures/stream/trait.Stream.html) is also not aware of the notion of an `Error`, no?

core/src/connection/manager.rs
```diff
@@ -0,0 +1,479 @@
...
+        ConnectionId(*self.task.key())
+    }
+
+    /// Aborts the pending connection attempt.
+    pub fn abort(self)  {
+        self.task.remove();
+    }
+}
+
+impl<'a, I, C> EstablishedEntry<'a, I, C> {
```
mxinden: Minor: Why not put this below `pub struct EstablishedEntry`? Why interleave `PendingEntry`?

core/src/connection/manager/task.rs
```diff
@@ -0,0 +1,333 @@
...
+        // TODO: don't pin this Future; this requires deeper changes though
+        future: Pin<Box<F>>,
+        /// The intended handler for the established connection.
+        handler: H,
+        /// While we are dialing the future, we need to buffer the events received via
+        /// `Command::NotifyHandler` so that they get delivered to the `handler`
+        /// once the connection is established. We can't leave these in `Task::receiver`
+        /// because we have to detect if the connection attempt has been aborted (by
+        /// dropping the corresponding `sender` owned by the manager).
+        events: Vec<I>
```
mxinden: Should we limit the amount of events one can buffer here? Or can this not cause any issues, given that the pending connection eventually times out?

core/src/connection/pool.rs
```diff
@@ -0,0 +1,907 @@
...
+pub struct Pool<TInEvent, TOutEvent, THandler, TTransErr, THandlerErr, TConnInfo = PeerId, TPeerId = PeerId> {
+    local_id: TPeerId,
+
+    /// The configuration of the pool.
+    limits: PoolLimits,
+
+    /// The connection manager that handles the connection I/O for both
+    /// established and pending connections.
+    ///
+    /// For every established connection there is a corresponding entry in `connected`.
```
mxinden: Not sure, but shouldn't it be:

```suggestion
    /// For every established connection there is a corresponding entry in `established`.
```

core/src/connection/pool.rs
```diff
@@ -0,0 +1,907 @@
...
+            .field("limits", &self.limits)
+            .finish()
+    }
+}
+
+impl<TInEvent, TOutEvent, THandler, TTransErr, THandlerErr, TConnInfo, TPeerId> Unpin
+for Pool<TInEvent, TOutEvent, THandler, TTransErr, THandlerErr, TConnInfo, TPeerId> {}
+
+/// Event that can happen on the `Pool`.
+pub enum PoolEvent<'a, TInEvent, TOutEvent, THandler, TTransErr, THandlerErr, TConnInfo, TPeerId> {
```
mxinden: We call an event emitted by a task an `Event`. Why not call a `PoolEvent` `Event` as well?

core/src/connection/manager.rs
```diff
@@ -0,0 +1,479 @@
...
+                return Poll::Pending;
+            }
+        }
+
+        for task in self.tasks.values_mut() {
+            let cmd = task::Command::NotifyHandler(event.clone());
+            match task.sender.start_send(cmd) {
+                Ok(()) => {},
+                Err(ref err) if err.is_full() => unreachable!("by (*)"),
+                Err(_) => {},
```
mxinden: As far as I know this is only reached on disconnection, right? Would this also be unreachable, given that we control both ends?

core/src/connection/manager.rs
tomaka: I haven't reviewed the code yet, but this comment is probably a reliquate from the era of futures 0.1.

core/src/connection/manager.rs
romanb: Good question, I don't know! This is taken from pre-existing code introduced in https://github.com/libp2p/rust-libp2p/pull/1191 so maybe @twittner can shed more light on it. My naive guess would be that it is simply to not require complete "fairness" from the executor in polling the task that runs code using functions of the manager vs polling the background tasks of the connection(s) to consume the commands sent to the task(s), i.e. such a buffer limit of N could allow one task to take N steps before an executor is forced to switch to another task to make progress, in a sense controlling how far ahead a producer task can get before execution must continue on the consumer side. That is just a guess though.

core/src/connection/manager.rs
romanb: Good point, thanks!

core/src/connection/manager.rs
romanb: No reason, just the result of a lot of back-and-forth during the implementation, I guess.

core/src/connection/pool.rs
romanb: Thanks!

core/src/connection/pool.rs
romanb: No particular reason. We could do that for pool events as well I suppose, because they are internal, it's just that on the other end we already have types like `NetworkEvent` in the public API, instead of `network::Event`, which I didn't change here, because there is also e.g. `ConnectionError` and not `connection::Error`.

core/src/connection/manager.rs
romanb: To my understanding the receiver corresponding to `task.sender` is owned by the background task, which may in principle drop the receiver any time. How do you come to the conclusion that "we control both ends" (assuming by "we" you mean the current thread executing this function)?

core/src/connection/manager/task.rs
romanb: This is also pre-existing code, so I'm not sure. Since dropping events is out of the question, as it may lead to arbitrary misbehaviour on behalf of the connection handler, the only options seem to be either to establish another mechanism for backpressure towards the manager, which I'm not sure how to do here, or to at least have a hard limit that avoids theoretical unbounded growth and when hit, the task aborts the pending connection and logs an error. This last option sounds the most appealing to me. What do you think?

romanb:
> > > The alternatives that come to mind involve making a NetworkBehaviour aware of ConnectionIDs, which is likely more intrusive and may be a change that goes well together with removing the ProtocolsHandler altogether later.
> > 
> > 
> > I would go more in this direction in general.
> 
> While this is clearly where we are heading in general, I'd really like to keep the changes and especially API changes outside of `libp2p-core` to a minimum in this first iteration.

Turns out it was necessary to do the following two small API changes to allow sending of responses on the same connection:

  * Add a connection ID parameter to `inject_node_event` (and renamed to `inject_event` for consistency with the rest of the changes).
  * Add a connection ID parameter to `NetworkBehaviorAction::SendEvent` (and renamed to `NetworkBehaviourAction::NotifyHandler` for consistency with the rest of the changes).

These changes are very straight-forward in terms of code migration. See: https://github.com/libp2p/rust-libp2p/pull/1440/commits/ae6bfe9023d682332ce82c21c5509aa7c872fd4e.

core/src/network/peer.rs
```diff
@@ -0,0 +1,523 @@
...
+        TInEvent: Send + 'static,
+        TOutEvent: Send + 'static,
+
+    {
+        if self.network.dialing.contains_key(&self.peer_id) {
+            let mut peer = DialingPeer {
+                network: self.network,
+                peer_id: self.peer_id
+            };
+            peer.connection().add_address(address);
```
tomaka: This behaviour seems a bit weird to me. We add the new addresses to a queue for dialing?
Does that mean that if I call `connect` three times I will still have only one new connection?

core/src/connection/manager.rs
```diff
@@ -0,0 +1,484 @@
...
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS
+// OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+// FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+// DEALINGS IN THE SOFTWARE.
+
+mod task;
```
tomaka: Nit: should go below the `use` statements.

core/src/connection/manager.rs
```diff
@@ -0,0 +1,484 @@
...
+// manager.
+
+/// The result of a pending connection attempt.
+type ConnectResult<C, M, HE, TE> = Result<(Connected<C>, M), ConnectionError<HE, TE>>;
+
+/// Connection identifier.
+#[derive(Debug, Copy, Clone, Hash, PartialEq, Eq, PartialOrd, Ord)]
+pub struct ConnectionId(TaskId);
+
+/// A connection `Manager` orchestrates the I/O of a set of connections,
```
tomaka:
```suggestion
/// A connection `Manager` orchestrates the I/O of a set of connections.
```

core/src/connection/manager.rs
tomaka: > given that each sender is guaranteed a slot?

We never clone this sender.

And yes, it is so that we can push multiple events without having to wait for the background task to make progress.

I don't think there's any drawback in increasing this value if we wanted to, except for the increase in memory usage. `4` is kind of arbitrary, but seems like a good value to me: unlikely to be reached in a normal situation, but small enough to not be a burden in memory.

core/src/connection/manager.rs
```diff
@@ -0,0 +1,484 @@
...
+#[derive(Debug)]
+pub struct EstablishedEntry<'a, I, C> {
+    task: hash_map::OccupiedEntry<'a, TaskId, TaskInfo<I, C>>,
+}
+
+impl<'a, I, C> EstablishedEntry<'a, I, C> {
+    /// (Asynchronously) sends an event to the connection handler.
+    ///
+    /// **Note:** Must only be called after `poll_ready_notify_handler`
+    /// was successful and without interference by another thread.
```
tomaka:
```suggestion
    /// was successful and without interference by another thread, otherwise the event is discarded.
```

core/src/connection/pool.rs
```diff
@@ -0,0 +1,907 @@
...
+    local_id: TPeerId,
+
+    /// The configuration of the pool.
+    limits: PoolLimits,
+
+    /// The connection manager that handles the connection I/O for both
+    /// established and pending connections.
+    ///
+    /// For every established connection there is a corresponding entry in `established`.
+    manager: Manager<TInEvent, TOutEvent, THandler, TTransErr, THandlerErr, TConnInfo>, // tasks
```
tomaka:
```suggestion
    manager: Manager<TInEvent, TOutEvent, THandler, TTransErr, THandlerErr, TConnInfo>,
```

core/src/connection/pool.rs
tomaka: I prefer `PoolEvent`, `NetworkEvent`, and so on, otherwise the identifier `Event` would refer to different things depending on which code you're reading, which to me makes it very painful to read.
Nothing more fun than getting errors such as "expected Event, but got Event instead".

core/src/connection/pool.rs
```diff
@@ -0,0 +1,907 @@
...
+}
+
+impl<TInEvent, TOutEvent, THandler, TTransErr, THandlerErr, TConnInfo, TPeerId> Unpin
+for Pool<TInEvent, TOutEvent, THandler, TTransErr, THandlerErr, TConnInfo, TPeerId> {}
+
+/// Event that can happen on the `Pool`.
+pub enum PoolEvent<'a, TInEvent, TOutEvent, THandler, TTransErr, THandlerErr, TConnInfo, TPeerId> {
+    /// A new connection has been established.
+    ///
+    /// In order for the connection to remain in the pool, it must be accepted.
```
tomaka: I believe this comment is no longer true?

core/src/connection/pool.rs
```diff
@@ -0,0 +1,907 @@
...
+    ///
+    /// Can only happen after a node has been successfully reached.
+    ConnectionError {
+        id: ConnectionId,
+        /// Information about the connection that errored.
+        connected: Connected<TConnInfo>,
+        /// The error that happened.
+        error: ConnectionError<THandlerErr, TTransErr>,
+        /// A reference to the pool that used to manage the connection.
+        pool: &'a mut Pool<TInEvent, TOutEvent, THandler, TTransErr, THandlerErr, TConnInfo, TPeerId>,
```
tomaka: We normally don't need these fields, but I also see why you added them. I'm fine with merging this PR then seeing afterwards if we can satisfy the borrow checker without making the code significantly dirtier.

core/src/connection/pool.rs
```diff
@@ -0,0 +1,907 @@
...
+    /// If a connection ID is given and no established connection with that
+    /// ID is known to the pool, `None` is returned.
+    ///
+    /// If no connection ID is given and the pool contains at least one
+    /// established connection to the given peer, an unspecified selection
+    /// is made among these connections.
+    ///
+    /// If no connection ID is given and no established connection exists
+    /// for the given peer, `None` is returned.
+    pub fn get_established(&mut self, peer: &TPeerId, id: Option<ConnectionId>)
```
tomaka: That is an extremely weird API. I'd prefer if this function didn't exist and the user has to explicitly select which connection is wanted.

core/src/connection/pool.rs
```diff
@@ -0,0 +1,907 @@
...
+    /// If a connection ID is given and no pending outgoing connection with that
+    /// ID is known to the pool, `None` is returned.
+    ///
+    /// If no connection ID is given and the pool contains at least one
+    /// pending outgoing connection to the given peer, an unspecified selection
+    /// is made among these connections.
+    ///
+    /// If no connection ID is given and no pending outgoing connection exists
+    /// to the given peer, `None` is returned.
+    pub fn get_outgoing(&mut self, peer: &TPeerId, id: Option<ConnectionId>)
```
tomaka: Same

core/src/connection/pool.rs
```diff
@@ -0,0 +1,907 @@
...
+                }
+            }
+        }
+    }
+
+    /// Asynchronously notifies a connection handler for the given peer of an event.
+    ///
+    /// If multiple connections exist to the peer, it is unspecified which handler
+    /// receives the event.
+    pub fn notify_handler(&mut self, peer: &TPeerId, event: TInEvent)
```
tomaka: Same remark. I think we will run into issues if we just leave selecting the connection to unspecified behaviours. This is the low-level API where we everything has to be strictly defined and we don't try to provide convenience methods.

core/src/connection/manager/task.rs
tomaka: Since this is pre-existing code and the solution isn't obvious, I'd suggest we open an issue and tackle it afterwards.

core/src/connection/manager/task.rs
```diff
@@ -0,0 +1,333 @@
...
+/// Commands that can be sent to a [`Task`].
+#[derive(Debug)]
+pub enum Command<T> {
+    /// Notify the connection handler of an event.
+    NotifyHandler(T),
+}
+
+/// Events that a task can emit to its manager.
+#[derive(Debug)]
+pub enum Event<T, H, TE, HE, C> {
```
tomaka:
```suggestion
pub enum TaskEvent<T, H, TE, HE, C> {
```

As per my other comment.

swarm/src/behaviour.rs
```diff
@@ -218,18 +221,26 @@ pub enum NetworkBehaviourAction<TInEvent, TOutEvent> {
...
+    /// at any time and thus the event may not reach a handler.
+    NotifyHandler {
+        /// The peer for whom a `ProtocolsHandler` should be notified.
         peer_id: PeerId,
-        /// The message to send.
+        /// The ID of the connection whose `ProtocolsHandler` to notify.
+        ///
+        /// If `None` and there exist multiple connections to the peer,
+        /// an unspecified choice is made.
+        connection: Option<ConnectionId>,
```
tomaka: I'm ok with the idea of leaving the choice to the swarm here, but I'd suggest splitting this into two variants `NotifyAnyHandler` and `NotifyHandler`.
Alternatively, replacing the `Option` with an enum would be fine as well, but to me it seems more annoying to use from an end-user perspective.

core/src/connection/pool.rs
romanb: Correct, thanks!

swarm/src/behaviour.rs
romanb: Sounds ok to me, I would go with the additional `NotifyAnyHandler` constructor then. It is certainly a bit clearer than seeing `connection: None`.

swarm/src/behaviour.rs
```diff
@@ -93,19 +93,22 @@ pub trait NetworkBehaviour: Send + 'static {
...
+    fn inject_replaced(&mut self, _: PeerId, _: ConnectedPoint, _: ConnectedPoint) {
     }
 
     /// Informs the behaviour about an event generated by the handler dedicated to the peer identified by `peer_id`.
     /// for the behaviour.
     ///
     /// The `peer_id` is guaranteed to be in a connected state. In other words, `inject_connected`
     /// has previously been called with this `PeerId`.
-    fn inject_node_event(
+    fn inject_event(
```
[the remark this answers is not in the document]
romanb: I have two minor arguments for this change though:

  1. We have `ConnectionHandler::inject_event` and `ProtocolsHandler::inject_event`. `NetworkBehaviour::inject_event` immediately suggests that it belongs to the same family of functions for communicating events back and forth between a `NetworkBehaviour` and `ProtocolsHandler` / `ConnectionHandler`.
  2. There is no other part of the API left that uses the `node` terminology, since it has been removed in favor of `connection` terminology. It could be `NetworkBehaviour::inject_connection_event`, but again I think the symmetry mentioned in the first point is worth something.

core/src/connection/manager/task.rs
romanb: While my personal preference goes in the other direction, i.e. not to repeat enclosing module names within type names in the module, I'm also  not really keen on debating it, I really just tried to stay somewhat consistent with the current naming scheme and not change names where I don't see a strong enough incentive. In this case, it was actually just called `Event` before, i.e. I didn't change it. Given that this is also not in the public API, would you still like me to change it in this PR?

core/src/connection.rs
```diff
@@ -0,0 +1,400 @@
...
+    pub(crate) current: Multiaddr,
+    /// Multiaddresses to attempt if the current one fails.
+    pub(crate) next: Vec<Multiaddr>,
+}
+
+/// A `DialingConnection` is a [`PendingConnection`] where the local peer
+/// has the role of the dialer (i.e. initiator) and the (expected) remote
+/// peer ID is known.
+pub struct DialingConnection<'a, TInEvent, TConnInfo, TPeerId> {
+    pub(crate) peer_id: &'a TPeerId,
```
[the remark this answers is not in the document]
romanb: The `DialingConnection` is only constructed in the `network::peer` module. I was going back and forth between putting in in the `connection` module or `network::peer` module. I could either create a `pub(crate)` constructor here instead (since the `connection` module is public) or indeed move the whole definition over to the `network::peer` module. I suppose you prefer the latter?

core/src/connection.rs
```diff
@@ -0,0 +1,400 @@
...
+            }
+        }
+    }
+}
+
+/// Attempt to reach a peer.
+#[derive(Debug, Clone)]
+pub struct DialingAttempt {
+    /// Identifier for the reach attempt.
+    pub(crate) id: ConnectionId,
```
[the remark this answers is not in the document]
romanb: It is only constructed in the `network` module and the type used in the `DialingConnection` and the `Network`, so similarly to the other case, I could either provide a `pub(crate)` constructor here or move the type with a `pub(super)` constructor into the `network::peer` module. I suppose you prefer the latter?

core/src/connection/manager.rs
```diff
@@ -0,0 +1,483 @@
...
+pub struct EstablishedEntry<'a, I, C> {
+    task: hash_map::OccupiedEntry<'a, TaskId, TaskInfo<I, C>>,
+}
+
+impl<'a, I, C> EstablishedEntry<'a, I, C> {
+    /// (Asynchronously) sends an event to the connection handler.
+    ///
+    /// **Note:** Must only be called after `poll_ready_notify_handler`
+    /// was successful and without interference by another thread, otherwise the event is discarded.
+    pub fn notify_handler(&mut self, event: I) {
```
twittner: Why not replacing this racy `poll_ready_notify_handler`/`notify_handler` combo with a non-racy version, e.g.

```rust
pub fn notify_handler(&mut self, event: I, cx: &mut Context) -> Option<I> {
    if self.poll_ready_notify_task(cx).is_pending() {
        return Some(event)
    }
    let cmd = task::Command::NotifyHandler(event);
    self.notify_task(cmd);
    None
}
```

core/src/network/peer.rs
romanb: If we keep the constraint of a single dialing (i.e. pending outgoing) connection per peer, which I have done so far as mentioned in the PR description, this seemed to me to be the natural thing to do, though it should be properly documented, of course. I.e. calling `connect()` on an already `ConnectedPeer` expresses the intention to establish another (new) connection and if such an attempt is already in progress, that same attempt is "picked up", just adding the given addresses as potential further candidates. In other words, if you really want multiple new connections you have to wait for the previous dialing connection to finish.

What alternatives do you see? Returning `Either<PendingPeer, ConnectionLimit>` in the "error" case? Renaming the function? Or changing the semantics in another way? Or would you even prefer if we remove the constraint of a single dialing attempt per peer?

core/src/connection/manager.rs
tomaka: This would be similar to the old `AsyncSink` enum, and it forces you to potentially store the event somewhere for later when the handler is ready. Meanwhile, with this "racy" API (it's not racy, since it requires `&mut`), you can generate the event on-the-fly.

twittner: We had this discussion in https://github.com/libp2p/rust-libp2p/pull/1373 already.

core/src/connection/pool.rs
romanb: Ok, done.

core/src/connection/pool.rs
romanb: Done.

core/src/connection/pool.rs
romanb: Done.

core/src/connection/error.rs
```diff
@@ -0,0 +1,73 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+// FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+// DEALINGS IN THE SOFTWARE.
+
+use crate::transport::TransportError;
+use std::{io, fmt};
+
+/// Errors that can occur in the context of a pending or established `Connection`.
+#[derive(Debug)]
+pub enum ConnectionError<THandlerErr, TTransErr> {
```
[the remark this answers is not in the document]
romanb: Will do, thanks!

core/src/connection.rs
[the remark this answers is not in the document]
tomaka: I would indeed move the definition, if that's ok. At the very least, use `pub(super)` rather than `pub(crate)`.

core/src/connection.rs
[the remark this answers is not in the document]
tomaka: (same idea as for `DialingConnection`)

core/src/connection/manager.rs
romanb: My preference would be to start with the simpler and safer API as per the suggestion of @twittner and potentially add the `poll_ready_x` / `x` combination later as an alternative if needed. That is, we may at the end offer both, just like `futures::channel::mpsc::Sender` does in form of `try_send`, `start_send` and `poll_ready`.

To elaborate, I'm sure we all understand the motivation and trade-off behind these two styles of API and we just have different priorities w.r.t. what we consider more important. Given that the `poll_ready_x` / `x` design is arguably both more "complex" (in that you have to use 2 methods instead of 1) and more error-prone (in that you have to use these two methods in a certain way, subject as well to pitfalls w.r.t multi-threading as already discussed in https://github.com/libp2p/rust-libp2p/pull/1373) I personally prefer to start with just providing `try_notify_handler` for now. The `poll_ready_notify_handler` / `notify_handler` combination could be added later if there is a strong demand or use-case (which as far as I can tell even the `Swarm` does not currently have).

Does that make sense and sound reasonable?

core/src/network/peer.rs
tomaka: I think that in the future we should remove the single dialing attempt per peer (cc https://github.com/libp2p/rust-libp2p/issues/925).
I don't think it should be part of this PR, because I don't think it's an easy task. For instance, we don't want to open 20 TCP connections whenever we connect to a node with 20 addresses.

My complaint here is indeed more about:

- The behaviour of this function being a bit strange.
- The function name not really reflecting its behaviour.

However I don't really have a suggestion, and I'm fine with adding a comment in #925 about this function needing some tweaking.

core/src/connection/manager.rs
romanb: @tomaka Could you elaborate on the :-1: w.r.t. what aspect you disagree with? Do you insist on offering `poll_ready_notify_handler` / `notify_handler` from the start or do you even have objections to offering `try_notify_handler`?

twittner: > it's not racy, since it requires `&mut`

This proposition has been refuted in #1373 and I am surprised to read it here again. I also do not understand why one would want to ignore the safer alternative to the current API. If one really wants to offer the more dangerous poll-then-send API I think @romanb's suggestion makes perfect sense.

tomaka: > This proposition has been refuted in #1373

It hasn't been refuted, I just gave up arguing.

> the safer alternative

The definition "safe" in the context of Rust is "cannot trigger undefined behaviour". In no way misusing this API can trigger any undefined behaviour. The word "safe" is totally irrelevant here.

The worst that can happen is a panic resulting in us misusing the API. There are lots of places in the Rust standard library and in the Rust ecosystem where misusing the API will result in a panic. Polling a `Future` after it has returned `Ready`, to give the most prominent example.

Using this API in a correct way is as racy as `try_notify_handler` is, in the sense that sometimes it will return `Pending` and sometimes it will return `Ready`, depending on the timing of the other tasks/threads.

Misuing this API is racy in a similar way as using this API correct, with the exception that the implementation might panic (because you misused the API).

> the more dangerous

In order for me to change my point of view, I would like to get a precise definition of what "dangerous" means here for you.

tomaka: My issue with `try_notify_handler` is the fact that there is no way for you to know whether your event will be accepted before generating said event.

An API that I think is a good is:

```
impl Manager {
    /// Try to give out a slot for sending an event.
    pub fn poll_ready_notify(&mut self, ...) -> Poll<EventSend> { ... }
}

impl<'a, T> EventSend<'a, T> {
    /// Sends the event, consuming `self`.
    pub fn send_event(self, event: T) { ... }
}
```

romanb: > The worst that can happen is a panic resulting from us misusing the API. There are lots of places in the Rust standard library and in the Rust ecosystem where misusing the API will result in a panic. Polling a `Future` after it has returned `Ready`, to give the most prominent example.

Note though that the underlying API we're talking about here, `futures::channel::mpsc::Sender`, never panics (and neither does our current API). Indeed the `Sender` API says `poll_ready` _should_ be called before `start_send` and it consequently still returns an error if the channel is full. It is a bit telling, to me anyway, that they did not go with saying "`poll_ready` _must_ be called before `start_send`" combined with panicking due to programmer error on `start_send` if the channel is full. And they also provide `try_send`.

> 
> Using this API in a correct way is as racy as `try_notify_handler` is, in the sense that sometimes it will return `Pending` and sometimes it will return `Ready`, depending on the timing of the other tasks/threads.

Here we are again at the point which I thought we clarified e.g. [here](https://github.com/libp2p/rust-libp2p/pull/1373#issuecomment-571192286). To recap, the crucial difference is in the assumptions that the method that must be called second (e.g. `notify_handler`) places on the possible errors that can occur and how it deals with them. If it assumes certain errors can never happen _because of a requirement that another method must be called first_ then calls to such two functions which may be interleaved by multiple threads if e.g. `Mutex`es are used can lead to unexpected behaviour, hence why I already augmented the commentary on `notify_handler` to say that not only must `poll_ready_notify_handler` be called first, but also that no other thread must be allowed to interleave between `poll_ready_notify_handler` and `notify_handler`. Of course that is easy to satisfy if, as a user, you just call both these `&mut` methods in sequence on the same thread without e.g. going through a `Mutex`, but nevertheless it is another thing to watch out for. We may agree to disagree on whether that justifies calling an API "racy", but in any case the `futures::channel::mpsc::Sender` API is actually not racy, whatever definition you choose, whereas the API provided by us, currently, is considered racy by myself and @twittner because it may silently (well, now it logs at least) ignore errors due to the channel being full when calling `notify_handler`.

I have added `try_notify_handler` in a new commit in the hope that it is uncontroversial since it follows the example of the `futures::channel::mpsc::Sender` API, leaving the other functions in place. Can we all live with that?

twittner: > > This proposition has been refuted in #1373
> 
> It hasn't been refuted, I just gave up arguing.

I would love to hear a new argument then. If "it's not racy, since it requires `&mut`" would be true why would we even have this discussion? Since the outcome of concurrent calls to `poll_ready_notify_handler` and `notify_handler` depends on relative timings there clearly is a race condition here that the individual `&mut`s in each of these methods can not prevent.

> 
> > the safer alternative
> 
> The definition of "safe" in the context of Rust is "cannot trigger undefined behaviour". In no way misusing this API can trigger any undefined behaviour. The word "safe" is totally irrelevant here.

[...]

> > the more dangerous
> 
> In order for me to change my point of view, I would like to get a precise definition of what "dangerous" means here for you.

Rust does not have a monopoly on the word "safe" and since you were asking for a definiton of "dangerous", I am happy to elaborate: Assuming there are undesirable program states, e.g. a panic, an API is safe w.r.t. those states if it can by construction guarantee that those states can never occur. Correspondingly an API is dangerous w.r.t. those states if it can not give those guarantees.

> Misuing this API is racy in a similar way as using this API correctly, with the exception that you might get a panic (because you misused the API).

This "exception" being an undesirable program state and the fact that the API can not guarantee its own correct use but allows for degrees of freedom that can lead to such a state makes it more dangerous in my book.

> My issue with `try_notify_handler` is the fact that there is no way for you to know whether your event will be accepted before generating said event.

We could combine both worlds by keeping `poll_ready_notify_handler` as an *indication* that *most likely* `notify_handler` will be able to accept the event, but in case it can not (because of API misuse) we can at least return the event:


```rust
pub fn poll_ready_notify_handler(&mut self, cx: &mut Context) -> Poll<()>;

pub fn notify_handler(&mut self, event: I, cx: &mut Context) -> Option<I>;
```

romanb: > We could combine both worlds by keeping `poll_ready_notify_handler` as an _indication_ that _most likely_ `notify_handler` will be able to accept the event, but in case it can not (because of API misuse) we can at least return the event:
> 
> ```rust
> pub fn poll_ready_notify_handler(&mut self, cx: &mut Context) -> Poll<()>;
> 
> pub fn notify_handler(&mut self, event: I, cx: &mut Context) -> Option<I>;
> ```

I would be ok with that as well, and it would make `try_notify_handler` superfluous (`notify_handler` would then internally use `Sender::try_send` which is used  by `Sender::start_send` internally anyway).

core/src/connection.rs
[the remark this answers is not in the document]
romanb: Done.

core/src/connection.rs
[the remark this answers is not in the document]
romanb: Done.

core/src/network/peer.rs
romanb: > I think that in the future we should remove the single dialing attempt per peer (cc #925).
I don't think it should be part of this PR, because I don't think it's an easy task. For instance, we don't want to open 20 TCP connections whenever we connect to a node with 20 addresses.

I'm not sure it would be that difficult actually and though I initially thought keeping this constraint would be a good thing, I tend to agree now that it may actually not be desirable in the long run. My naive approach would be that every call to `connect()` would simply result in a new dialing connection for the given addresses. If you have a connection ID of a dialing connection you can still add addresses while dialing is ongoing. I may give it a try and either get back here with a diff or abort if it gets too cumbersome and rather adjust the function name / documentation.

core/src/connection/manager/task.rs
tomaka: Since it's a private API it doesn't matter so much, so let's skip this.

core/src/connection/error.rs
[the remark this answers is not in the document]
romanb: Since the use of `thiserror` for such an error type with type parameters requires trait bounds to be present on the error type itself, which in turn propagate through the code base, there actually does not seem to be a lot of gain in using it here. Nevertheless the experiment can be seen [here](https://github.com/romanb/rust-libp2p/commit/9466a4526a62b2cdc8bb58a9ef6258c91e615761).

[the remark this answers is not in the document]
twittner: Thanks for giving it a go.

core/src/network/peer.rs
romanb: [The latest commit](https://github.com/libp2p/rust-libp2p/pull/1440/commits/87452c82d7fff046d8127a0bac47fbd1bcf5422c) shows my current proposal. It essentially establishes more symmetry in the `Peer` API due to:

  1. A `Peer` can `connect()` in any representation, whether disconnected, connected or dialing. Thus there is now `Peer::connect()`.
  2. A `DialingPeer` now has methods `connections()`, `connection(id)` and `num_connections()` analogously to `ConnectedPeer`, which operate on `DialingConnection`s instead of `EstablishedConnection`s.

romanb: I'm open to extracting that work as a separate, follow-up PR for easier review and sicussion, as the changeset got a bit larger than I anticipated. Let me know.

tomaka: > I'm open to extracting that work as a separate, follow-up PR for easier review and discussion, as the changeset got a bit larger than I anticipated. Let me know.

I'm ok with that.

core/src/connection/manager.rs
romanb: @mxinden Do you have further comments on the matter?

core/src/network/peer.rs
romanb: Your original critique w.r.t. this function is now intended to be addressed in this PR by [this commit](https://github.com/libp2p/rust-libp2p/pull/1440/commits/35405880ea281f49cac2fedcf9e444b901a26e5c). Let me know if that is acceptable. I will follow-up with another PR w.r.t. removing the constraint of a single ongoing dialing attempt per peer.

core/src/connection/manager/task.rs
romanb: See https://github.com/libp2p/rust-libp2p/issues/1465.

core/src/connection/manager.rs
mxinden: @romanb sorry, I confused this with a different `start_send`. Marking as resolved.

swarm/src/lib.rs
```diff
@@ -346,37 +381,43 @@ where TBehaviour: NetworkBehaviour<ProtocolsHandler = THandler>,
...
-                Poll::Ready(NetworkEvent::NodeClosed { conn_info, endpoint, error }) => {
-                    log::trace!("Connection {:?} with endpoint {:?} closed by {:?}",
-                                conn_info, endpoint, error);
-                    this.behaviour.inject_disconnected(conn_info.peer_id(), endpoint);
-                    return Poll::Ready(SwarmEvent::Disconnected(conn_info.peer_id().clone()));
-                },
-                Poll::Ready(NetworkEvent::Replaced { new_info, closed_endpoint, endpoint, .. }) => {
-                    this.behaviour.inject_replaced(new_info.peer_id().clone(), closed_endpoint, endpoint);
+                Poll::Ready(NetworkEvent::ConnectionError { connected, error, num_established }) => {
+                    log::error!("Connection {:?} closed by {:?}", connected, error);
```
tomaka: `error!` should be reserved for exceptional issues, not when a connection closes, which happens all the time.

swarm/src/behaviour.rs
```diff
@@ -93,19 +93,22 @@ pub trait NetworkBehaviour: Send + 'static {
...
     /// `inject_connected`. This is a logically safe way to implement this behaviour. However, you
     /// may want to overwrite this method in the situations where this isn't appropriate.
-    fn inject_replaced(&mut self, peer_id: PeerId, closed_endpoint: ConnectedPoint, new_endpoint: ConnectedPoint) {
-        self.inject_disconnected(&peer_id, closed_endpoint);
-        self.inject_connected(peer_id, new_endpoint);
+    #[deprecated(since = "0.16.0", note =
+        "Connections are no longer replaced and this function no longer called. \
+        It will eventually be removed from the trait."
+    )]
+    fn inject_replaced(&mut self, _: PeerId, _: ConnectedPoint, _: ConnectedPoint) {
```
tomaka: Considering that we're renaming `inject_node_event` to `inject_event`, I'd remove that method immediately and therefore remove confusion.

swarm/src/lib.rs
romanb: Absolutely, sorry, this is just a left-over from some testing sessions with substrate/polkadot. It should not have been committed. I will change it back to debug level shortly.

swarm/src/behaviour.rs
romanb: Alright.

romanb:
Thanks to all reviewers for the time-consuming reviews of this large changeset! My current status is that I'm still doing some final testing with substrate / polkadot, i.e. I'm preparing a corresponding PR for `substrate` with the necessary changes. When no further issues are uncovered I will make a final rebase / merge with `master` here.

However, in the context of the substrate integration, it seems like there is one more addition to this PR required, namely a `NetworkBehaviourAction::NotifyAllHandlers` in addition to the existing `NotifyAnyHandler` and `NotifyHandler`, due to the fact that the connection handlers in substrate seem to have a `enable` / `disable` facility and it seems now necessary for these `enable` / `disable` events to be propagated to all handlers, even though as per the `libp2p-swarm` semantics there shouldn't be multiple connections normally, but as of this PR there may now, at least temporarily, be further connections as a result of e.g. simultaneous connections attempts. So to be on the safe side and preserve the existing semantics, it seems prudent to provide the ability for a network behaviour to send an event to all handlers, however many there may be. Please let me know if there are objections to that or alternative proposals.

twittner:
What is the conclusion regarding https://github.com/libp2p/rust-libp2p/pull/1440#discussion_r380236472? AFAICS this is not resolved.

romanb:
> What is the conclusion regarding [#1440 (comment)](https://github.com/libp2p/rust-libp2p/pull/1440#discussion_r380236472)? AFAICS this is not resolved.

True, it is not. Personally I would prefer to incorporate your last suggestion for that API (i.e. removing `try_notify_handler` and adjusting `notify_handler`) if @tomaka has no objections. However, I took his silence on the matter to mean that while he is ok with the status quo (`notify_handler` + `poll_notify_handler` + `try_notify_handler`) he is still contemplating the latest comments in that discussion.

tomaka:
The test is repeatedly failing.

romanb:
> The test is repeatedly failing.

Which test do you mean? They all look green right now. Relatedly, my proposal for the substrate integration is now also ready for review (but obviously not for merging): https://github.com/paritytech/substrate/pull/5066.

tomaka:
> Which test do you mean?

Hah, I restarted the tests once again right after posting my comment, and now it's green.

romanb:
> > Which test do you mean?
> 
> Hah, I restarted the tests once again right after posting my comment, and now it's green.

I see, I can look at the ipfs-kad example again, but I would probably as a first guess blame IPFS nodes. It may help if we increase the acceptable maximum payload size of Kademlia requests / responses from 4KiB to 8KiB. I remember often seeing failures due to that size limit, sometimes even from the first node. There may be other issues as well, of course.

tomaka:
> I would probably as a first guess blame IPFS nodes.

That would also be my guess.

PR #1442, opened 2020-02-10
add poll parameters to custom poll function

### LINKED ISSUE #1431: Make custom_poll to take polling parameters
Are there any reasons not to take the `PollParameters` argument in custom polling function of `NetworkBehaviour` ?

If no, I would like to make a PR, because it will simplify my project a little bit.

michaelvoronov:
Resolves #1431.


tomaka:
I'll merge that tomorrow if there's no other opinion by then.

PR #1443, opened 2020-02-11
yamux: Re-export `WindowUpdateMode`.

twittner:
Also implement `Deref` and `DerefMut` on `Config` to allow transparent configuration of `yamux::Config`.


twittner:
Re-exporting `yamux::Config` would pollute the namespace I think and we would anyway need to rename it so it does not collide with `libp2p_yamux::Config`. In the generated doc you would then see `Config`, `LocalConfig`, and another `YamuxConfig` which looks confusing, no?

Duplicating the fields is better but needs to be kept in sync with changes in `yamux` which I wanted to avoid.

PR #1444, opened 2020-02-11
Use SHA256 again for from_public_key

tomaka:
Reverts parts of #1413 (which hasn't been published yet)

Summary of the story:

Libp2p 0.12 only hashes public keys using SHA2256, which is not conforming to specs. Keys less than 42 bytes have to be encoded using the identity hash.

Libp2p 0.13 and above no longer sees an identity peer ID as an error by itself.
However we forgot a detail: if we try to connect to a node that has an identity peer ID, it would always fail because the public key sent by secio/noise is always locally encoded using SHA256 before comparison.

#1413 fixes that and makes it possible to connect to nodes with an "identity" hash, and also encodes small public key using the "identity" hash by default.

Unfortunately, upgrading from a network before #1413 to a network after #1413 is kind of tricky. Nodes with #1413 will send peer IDs with the identity hash to nodes without #1413, which will then fail to connect. It could cause a netsplit.

What this PR does is revert the identity hashing part of #1413, but keep the part where we compare keys equal.
The intent is that we release version 0.16 with this PR, then later we fully apply #1413 and release 0.17.
Then 0.16 will be fully compatible with both 0.15 and 0.17.


tomaka:
Hopefully we're nearing the end of this saga.

PR #1446, opened 2020-02-11
Propagate rustdoc links check to GH actions

tomaka:
I'm opening this PR quickly after merging #1430 in order to see if CI works.
If CI fails, I'll also fix any failure as part of this PR.


tomaka:
Alright, GitHub Actions are set up. We can remove CircleCI in a couple weeks/months if we see that everything fine.

PR #1356, opened 2019-12-20
Support /dnsaddr in multiaddr

ackintosh:
closes https://github.com/libp2p/rust-libp2p/issues/967

📝 
- [multiaddr/DNSADDR.md at master · multiformats/multiaddr](https://github.com/multiformats/multiaddr/blob/master/protocols/DNSADDR.md)
- https://github.com/multiformats/go-multiaddr-dns/blob/0eb9e18d2fb65518efd1583d7bb91c6a12ffb1b4/resolve.go#L146

PR #1450, opened 2020-02-12
Add transport support for dnsaddr

ackintosh:
Related issue: https://github.com/libp2p/rust-libp2p/issues/967
Related PR: https://github.com/libp2p/rust-libp2p/pull/1356

https://github.com/multiformats/multiaddr/blob/master/protocols/DNSADDR.md

### DIFF
--- a/transports/dns/src/lib.rs
+++ b/transports/dns/src/lib.rs
@@ -33,7 +33,7 @@
 //! replaced with respectively an `/ip4/` or an `/ip6/` component.
 //!
 
-use futures::{prelude::*, future::BoxFuture};
+use futures::{prelude::*, future::BoxFuture, stream::FuturesOrdered};
 use libp2p_core::{
     Transport,
     multiaddr::{Protocol, Multiaddr},
@@ -46,7 +46,6 @@ use trust_dns_client::rr::{DNSClass, RecordType};
 use trust_dns_client::udp::UdpClientConnection;
 use trust_dns_client::client::{SyncClient, Client};
 use trust_dns_proto::rr::domain::Name;
-use futures::stream::FuturesOrdered;
 
 /// Represents the configuration for a DNS transport capability of libp2p.
 ///
--- a/transports/dns/src/lib.rs
+++ b/transports/dns/src/lib.rs
@@ -33,7 +33,7 @@
 //! replaced with respectively an `/ip4/` or an `/ip6/` component.
 //!
 
-use futures::{prelude::*, channel::oneshot, future::BoxFuture, FutureExt};
+use futures::{prelude::*, future::BoxFuture, FutureExt};
 use libp2p_core::{
     Transport,
     multiaddr::{Protocol, Multiaddr},
@@ -46,6 +46,7 @@ use trust_dns_client::rr::{DNSClass, RecordType};
 use trust_dns_client::udp::UdpClientConnection;
 use trust_dns_client::client::{SyncClient, Client};
 use trust_dns_proto::rr::domain::Name;
+use futures::stream::FuturesOrdered;
 
 /// Represents the configuration for a DNS transport capability of libp2p.
 ///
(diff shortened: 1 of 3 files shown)


transports/dns/src/lib.rs
```diff
@@ -170,7 +176,56 @@ where
                             })
                             .next()
                             .ok_or_else(|| DnsErr::ResolveFail(name))
-                    }.left_future()
+                    }.boxed().left_future()
```
tomaka: You normally don't have to box the future. Just put an `async` block around the entire code instead of having two distinct `async` blocks.

Since this is a lot of code, it might be worth moving it to a separate freestanding function.

transports/dns/src/lib.rs
```diff
@@ -299,10 +370,59 @@ mod tests {
...
+            fn listen_on(self, _: Multiaddr) -> Result<Self::Listener, TransportError<Self::Error>> {
+                unreachable!()
+            }
+
+            fn dial(self, addr: Multiaddr) -> Result<Self::Dial, TransportError<Self::Error>> {
+                let addr = addr.iter().collect::<Vec<_>>();
+                assert_eq!(addr.len(), 3);
+                // TODO: ipfs
+//                match addr[2] {
+//                    Protocol::Ipfs(_) => (),
```
ackintosh: I'll file another PR to add the ipfs support.

transports/dns/src/lib.rs
```diff
@@ -238,6 +225,90 @@ where TErr: error::Error + 'static
...
+                    }
+                })
+                .collect::<Vec<_>>();
+
+            // if resolved_addrs.len() == 0 {} // TODO
+
+            if suffix.len() == 0 {
+                Ok(resolved_addrs[0].clone())
+            } else {
+                // TODO: suffix matching
```
tomaka: I'd that this is out of scope of this function.

transports/dns/src/lib.rs
```diff
@@ -238,6 +225,90 @@ where TErr: error::Error + 'static
...
+                    } else {
+                        None
+                    }
+                })
+                .next()
+                .ok_or_else(|| DnsErr::ResolveFail(name))
+        }
+        Protocol::Dnsaddr(ref n) => {
+            let conn = UdpClientConnection::new("8.8.8.8:53".parse().unwrap()).unwrap(); // TODO: error handling
+            let client = SyncClient::new(conn); // TODO: should use async client?
```
tomaka: > // TODO: should use async client?

Yes!

transports/dns/src/lib.rs
ackintosh: https://github.com/libp2p/rust-libp2p/pull/1463

transports/dns/src/lib.rs
```diff
@@ -238,6 +225,90 @@ where TErr: error::Error + 'static
     }
 }
 
+async fn resolve_dns<T>(p: Protocol<'_>, suffix: Multiaddr) -> Result<Multiaddr, DnsErr<T::Error>>
+    where
```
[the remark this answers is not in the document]
ackintosh: Updated at https://github.com/libp2p/rust-libp2p/pull/1450/commits/fcc3d6a0f9dd1d0e0d2d6245c56f41bbd882d0cc .

If `T: Transport` is removed we'll get the error:
```text
error[E0220]: associated type `Error` not found for `T`
   --> transports/dns/src/lib.rs:228:89
    |
228 | async fn resolve_dns<T>(p: Protocol<'_>, suffix: Multiaddr) -> Result<Multiaddr, DnsErr<T::Error>>
    |                                                                                         ^^^^^^^^ associated type `Error` not found
```

transports/dns/src/lib.rs
ackintosh: Hmm... 

I've added [trust-dns](https://github.com/bluejekyll/trust-dns) in order to query TXT records. I'm trying to switch the SyncClient to AsyncClient but I think AsyncClient can not be used for rust-libp2p because a background future is required for the client. 🤔

https://docs.rs/trust-dns-client/0.19.3/trust_dns_client/#async-usage

```rust
// Create a new client, the bg is a background future which handles
//   the multiplexing of the DNS requests to the server.
//   the client is a handle to an unbounded queue for sending requests via the
//   background. The background must be scheduled to run before the client can
//   send any dns requests
let client = AsyncClient::connect(stream);

// await the connection to be established
let (mut client, bg) = runtime.block_on(client).expect("connection failed");

// make sure to run the background task
runtime.spawn(bg);
```

======

Do you know other library what can query TXT records asynchronously? 💦

tomaka: > I think AsyncClient can not be used for rust-libp2p because a background future is required for the client.

That makes it more difficult to implement, but normally not impossible.

tomaka: The idea is not to run the background future in the background, but to run it alongside with the "front" one. I don't have more precise guidelines without looking deeply into details (which I don't really have the time right now to do) though.

ackintosh: I've thought for a while about this but I can not come up anything to address the issue... hmm... 🤔💦

romanb:
Superseded by https://github.com/libp2p/rust-libp2p/pull/1931.

PR #1454, opened 2020-02-13
Add a js implementation of `wasm-ext`.

expenses:
This PR just copies the implementation of the web socket transport we use for the substrate browser node to libp2p and binds to it via wasm-bindgen.


tomaka:
One problem is that you're assuming websockets or even browser. This is a generic WebAssembly transport.

expenses:
Right, so we should write js bindings to some things (`multiaddr_to_ws` etc) but not others? It might be better to plug in the implementation into `browser-utils` in substrate then.

tomaka:
IMO we should instead add a `websocket_transport` feature to `libp2p-wasm-ext` and provide this transport only if the feature is enabled.

Also, the `transport` function should be named `websocket_transport`.

expenses:
Yep, something like 
```
#[cfg_attr(feature = "websocket_transport", wasm_bindgen(module = "/src/implementation.js"))]
#[cfg_attr(not(feature = "websocket_transport"), wasm_bindgen)]
```
perhaps?

tomaka:
Can you not split the `extern` block in two?

expenses:
That might also work

transports/wasm-ext/src/lib.rs
```diff
@@ -126,19 +123,20 @@ pub mod ffi {
         #[wasm_bindgen(method, getter)]
         pub fn local_addr(this: &ConnectionEvent) -> String;
     }
+
+    #[cfg(feature = "websockets")]
+    #[wasm_bindgen(module = "/src/websockets.js")]
+    extern "C" {
+        /// Returns a `Transport` implemented using websockets.
+        pub fn wasm_transport() -> Transport;
```
tomaka:
```suggestion
        pub fn websockets_transport() -> Transport;
```

expenses: Done

transports/wasm-ext/Cargo.toml
```diff
@@ -16,3 +16,6 @@ libp2p-core = { version = "0.16.0", path = "../../core" }
 parity-send-wrapper = "0.1.0"
 wasm-bindgen = "0.2.42"
 wasm-bindgen-futures = "0.4.4"
+
+[features]
+websockets = []
```
tomaka:
```suggestion
websocket = []
```

transports/wasm-ext/src/lib.rs
```diff
@@ -123,6 +123,13 @@ pub mod ffi {
         #[wasm_bindgen(method, getter)]
         pub fn local_addr(this: &ConnectionEvent) -> String;
     }
+
+    #[cfg(feature = "websockets")]
```
tomaka:
```suggestion
    #[cfg(feature = "websocket")]
```

tomaka:
Ugh, obviously I forgot to merge before the 0.16.1 PR.

PR #1455, opened 2020-02-14
Revert bumping multistream-select and bump multihash

tomaka:
This is a fuck-up from my 0.16 PR.

We didn't change anything in multistream-select: https://github.com/libp2p/rust-libp2p/commits/master/misc/multistream-select


tomaka:
Pushed another commit that bumps multihash.

PR #1457, opened 2020-02-16
Small libp2p-kad tweaks.

romanb:
* Replace two remaining `AsRef` constraints for DHT keys with `Borrow`.
  * Add a bit of debug/trace logging.
  * Tiny refactoring and a debug assertion for the `bucket` module.


romanb:
>     * Replace two remaining `AsRef` constraints for DHT keys with `Borrow`.

Due to this bullet, it would be good if this goes into a `0.16.1` or similar, so that there are no surprises once `0.17` is released with fully enabled "identity hashing", though the way substrate uses `get_closest_peers` with random IDs it would likely not be affected by any such difference.

PR #1458, opened 2020-02-16
Move mdns

### LINKED ISSUE #1452: Mdns location
I am just interested why mdns is located in `rust-libp2p/misc` not in `rust-libp2p/protocols`?

michaelvoronov:
Resolves #1452.

It seems more convinient to have `mdns` in `protocols` folder then in `misc`.


Cargo.toml
```diff
@@ -55,7 +55,7 @@ env_logger = "0.7.1"
 members = [
     "core",
     "misc/core-derive",
-    "misc/mdns",
+    "protocols/mdns",
```
tomaka: Alphabetic ordering here

michaelvoronov: done

tomaka:
I don't really have an opinion on that.

PR #1460, opened 2020-02-18
Finish changes regarding PeerId, again

tomaka:
See https://github.com/libp2p/rust-libp2p/pull/1444 and https://github.com/libp2p/rust-libp2p/pull/1449 for context.


tomaka:
I'm going to merge this if there's no opposition.

PR #1463, opened 2020-02-21
multiaddr: Add ipfs support

ackintosh:
Related PR: https://github.com/libp2p/rust-libp2p/pull/1450#pullrequestreview-362059811

The bootstrap address is resolved to the address which contains `ipfs` below:

```sh
$ dig +short _dnsaddr.ams-2.bootstrap.libp2p.io txt

"dnsaddr=/ip4/147.75.83.83/tcp/4001/ipfs/QmbLHAnMoJPWSCR5Zhtx6BHJX9KiKNN6tpvbUcqanj75Nb"
"dnsaddr=/ip6/2604:1380:2000:7a00::1/tcp/4001/ipfs/QmbLHAnMoJPWSCR5Zhtx6BHJX9KiKNN6tpvbUcqanj75Nb"
```

We need to support ipfs. However, I think ipfs should be parsed as p2p since [the code](https://github.com/multiformats/multiaddr/blob/master/protocols.csv) is shared between the two protocols and p2p is prefered over ipfs.


tomaka:
You're not the first one to try to add support for `/ipfs/` in multiaddresses, and I think the general consensus is that:
- It's not great to re-introduce something that has been deprecated a long time ago
- It's additionally not great that doing `"/ipfs/foo".parse().unwrap().to_string()` would produce `"/p2p/foo"` (i.e. not the original string).

Preferably it would be IPFS updating the values they store in their DNS records, or even add code in libp2p-dns that can specifically handle `/ipfs`.

romanb:
I tend to agree with what @tomaka already said, i.e. it doesn't seem desirable to add support for a legacy identifier that should no longer be used. This is not strictly necessary for the work in #1450, is it?

ackintosh:
@tomaka @romanb  Thank you for your comment! Now I understand it's not desirable to add support for a legacy identifier.
closing this PR.

PR #1468, opened 2020-02-24
Onion3Addr::acquire: fix bug due to differing lifetimes

Centril:
Because the `Self` type is `Onion3Addr<'a>` and the return type uses `'b`, the usage of `Self` as a constructor is illegal as per https://crater-reports.s3.amazonaws.com/pr-69340/try%239ff4d07208097950831ed4ea9d76feb1eafc8baa/reg/libp2p-plaintext-0.15.0/log.txt and the bug fix in https://github.com/rust-lang/rust/pull/69340. This PR fixes that problem for you.


Centril:
CI seems happy :)

PR #1469, opened 2020-02-24
kad: Export `QueryId` and add `Kademlia::protocol_name`.

twittner:
The `QueryId` type should be exported as it is used in the `NetworkBehaviour::ProtocolsHandler` type of `Kademlia`.

`Kademlia::protocol_name` is added for convenience.


protocols/kad/src/protocol.rs
```diff
@@ -40,6 +40,8 @@ use std::{io, iter};
 use unsigned_varint::codec;
 use wasm_timer::Instant;
 
+pub(crate) const DEFAULT_PROTO_NAME: Cow<'static, [u8]> = Cow::Borrowed(b"/ipfs/kad/1.0.0");
```
romanb:
```suggestion
pub const DEFAULT_PROTO_NAME: Cow<'static, [u8]> = Cow::Borrowed(b"/ipfs/kad/1.0.0");
```
Would it not make sense to export this constant from the crate? Just an idea, since its not necessary we don't need to do that at the moment.

protocols/kad/src/behaviour.rs
```diff
@@ -222,6 +222,13 @@ where
         Self::with_config(id, store, Default::default())
     }
 
+    /// Get the protocol name of this kademlia instance.
+    pub fn protocol_name(&self) -> Cow<'static, [u8]> {
```
tomaka: This API is a bit weird. If it's the default, then we return a `&'static [u8]`, and if it's not the default, we clone a `Vec<u8>`.
A more proper way to me would be to return a reference in both cases. In other words an enum of either `&'static [u8]` or `&'a [u8]`.
But doing some seems a bit too cumbersome, and in the end returning `&[u8]` looks like the best approach to me.

protocols/kad/src/protocol.rs
```diff
@@ -40,6 +40,9 @@ use std::{io, iter};
 use unsigned_varint::codec;
 use wasm_timer::Instant;
 
+/// The protocol name used for negotiating with multistream-select.
+pub const DEFAULT_PROTO_NAME: Cow<'static, [u8]> = Cow::Borrowed(b"/ipfs/kad/1.0.0");
```
tomaka: Since it's public, I would also do that:

```suggestion
pub const DEFAULT_PROTO_NAME: &'static [u8] = b"/ipfs/kad/1.0.0";
```

protocols/kad/src/protocol.rs
```diff
@@ -152,7 +155,7 @@ impl KademliaProtocolConfig {
 impl Default for KademliaProtocolConfig {
     fn default() -> Self {
         KademliaProtocolConfig {
-            protocol_name: Cow::Borrowed(b"/ipfs/kad/1.0.0"),
+            protocol_name: DEFAULT_PROTO_NAME
```
tomaka:
```suggestion
            protocol_name: Cow::Borrowed(DEFAULT_PROTO_NAME)
```

PR #1481, opened 2020-03-02
Add prefix to `KeypairIdentity` signature data in noise protocol

pawanjay176:
For the signature in the noise handshake payload, the [spec](https://github.com/libp2p/specs/tree/master/noise#the-libp2p-handshake-payload) states that

> The data to be signed is the UTF-8 string `noise-libp2p-static-key:`, followed by the Noise static public key, encoded according to the rules defined in section 5 of RFC 7748.

This PR prefixes the data for signing and verification with the prefix string to conform with the spec.

### DIFF
--- a/protocols/noise/src/protocol.rs
+++ b/protocols/noise/src/protocol.rs
@@ -27,6 +27,8 @@ use libp2p_core::identity;
 use rand::SeedableRng;
 use zeroize::Zeroize;
 
+const NOISE_STATIC_KEY_SIGNATURE_PREFIX: &str = "noise-libp2p-static-key:";
+
 /// The parameters of a Noise protocol, consisting of a choice
 /// for a handshake pattern as well as DH, cipher and hash functions.
 #[derive(Clone)]
@@ -92,8 +94,12 @@ pub trait Protocol<C> {
         C: AsRef<[u8]>
     {
         Self::linked(id_pk, dh_pk)
-            ||
-        sig.as_ref().map_or(false, |s| id_pk.verify(dh_pk.as_ref(), s))
+            || sig.as_ref().map_or(false, |s| {
+                id_pk.verify(
+                    &[NOISE_STATIC_KEY_SIGNATURE_PREFIX.as_bytes(), dh_pk.as_ref()].concat(),
+                    s,
+                )
+            })
     }
 }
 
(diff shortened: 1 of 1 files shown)


tomaka:
Since libp2p-noise is used in production, this is a big breaking change that we can't just merge as is.
There needs to be some way to keep the old behaviour.

As a more general note, I think it's a better idea to not incrementally change libp2p-noise, but instead to make it fully conforming to the specs at once. I don't exactly know what "conforming to the specs" involves, however, as I don't know it in details.

pawanjay176:
> As a more general note, I think it's a better idea to not incrementally change libp2p-noise, but instead to make it fully conforming to the specs at once

Thanks for the feedback :) I agree. This was the only issue I found in rust-libp2p while doing interop testing with nim-libp2p noise. Apart from this, rust seems to be fully compliant with the spec but I'll do some more testing against nim-libp2p and go-libp2p to make sure. 

> There needs to be some way to keep the old behaviour.

We could could add an additional condition in the `verify` function to verify the old signature format without the prefix so it would work with code already running in production. Is that what you meant by keeping old behaviour?

AgeManning:
One (not very elegant) suggestion would be to maintain a duplicate `noise-spec` protocol (or alternatively we can keep spec changes in our fork). 

As @tomaka suggests, we could keep this updated with the spec (as the spec seems to be changing) as a separate protocol (although their protocol-id's collide) and users can use either protocol. 

Not entirely fan of this idea, but could be one way of handling future differences.

tomaka:
> rust seems to be fully compliant with the spec

Oh that's great!

> Is that what you meant by keeping old behaviour?

I was thinking of having both `LegacyNoiseConfig` and `NoiseConfig` (or similar names). API breaking changes are fine as long as there's a way to use the current wire protocol.

> as the spec seems to be changing

Err, that's less great!

pawanjay176:
> I was thinking of having both `LegacyNoiseConfig` and `NoiseConfig` (or similar names).

This looks like the best option in order to keep the old behaviour. So if we want to network with the old code, we would use `LegacyNoiseConfig` else the spec'd `NoiseConfig.

Is that right @tomaka ?

tomaka:
Yes

romanb:
Superseded by https://github.com/libp2p/rust-libp2p/pull/1545.

PR #1484, opened 2020-03-05
Update multistream-select to stable futures

tomaka:
Close #1331 
This is a less sexy version of #1331, where we just do a one-to-one conversion to futures 0.3, without attempting to use `async`/`await`.

Using `async`/`await` at such a low level means that we have to box futures, which means that we have to enforce `Send` and `'static` everywhere, which isn't great.

After this PR, `futures 0.1` has disappeared from my `Cargo.lock`.


tomaka:
I'm currently building Polkadot with this PR (as a way to fuzz the code), and will report how it goes.

Issue #1482, opened 2020-03-03
Expire listening addresses if a listener is closed


tomaka:
In `listeners.rs`, if a listener is closed, we emit a `ListenersEvent::Closed` event, but we don't mark the addresses the listener was listening on as "expired".

The `ListenersEvent::Closed` event should contain a list of addresses, and then we should appropriately call `NetworkBehaviour::inject_expired_listen_addr` and emit `SwarmEvent::ExpiredListenAddr`.

disfenoid:
Hello, I would like to take this one, if you don't mind.

tcharding:
[Draft PR](https://github.com/libp2p/rust-libp2p/pull/1485) is an attempt at resolving this.  It is draft because I have some `rustfmt` dramas.

Issue #1323, opened 2019-11-25
Example ipfs-kad is broken


FedericoPonzi:
Command:
`cargo run --example ipfs-kad`
Expected behaviour:
"Query finished with closest peers: ..."
Actual behaviour:
```
Searching for the closest peers to PeerId("QmZy9TwDKSZiMDPUeDnsNvMKfX6GFsCHJqmmLQs91ecyEe")
thread 'tokio-runtime-worker-2' panicked at 'Query finished with no closest peers.', examples/ipfs-kad.rs:104:33
```
I've also tried to replace the ip address / peer id with others from `ipfs bootstrap` but still getting the same error.

tomaka:
The last time I've checked (a few months ago), there were two ongoing issues:

- Most of the IPFS bootnode addresses are not responsive.
- All the responsive bootnodes only support RSA with 1024bits key.

Unfortunately, the Rust library for RSA that we use (`ring`) doesn't support 1024bits keys anymore, as they are considered insecure. We therefore can't connect to any node.

koivunej:
An almost offtopic update related to this but likely interest of anyone looking into ipfs compatilibity:

Tl;DR: Don't try with ed25519 keys, use rsa (for now).

With https://github.com/dvc94ch/rust-ipfs/pull/37 I managed to transfer a single block from a local go-ipfs node to local rust-ipfs (small steps) and turned out the major incompatibility is with using ed25519 keys, which turn into different peerids by different libp2p impls. If this is worked around for secio the nodes will still use their incompatible versions of peerids for dht which simply does not work (this part has not been confirmed other than that "it does not work").

On the linked PRs commits and changes there are more links on the compatibility if you are interested.

tomaka:
> which turn into different peerids by different libp2p impls

This might be related to peer ids inlining. See https://github.com/libp2p/rust-libp2p/pull/1230 for context.

koivunej:
> This might be related to peer ids inlining. See #1230 for context.

Yes. Had not noticed the PR. Also for more context https://github.com/libp2p/specs/issues/138 ... Couldn't find an issue related to this in `go-libp2p` but the relevant code: https://github.com/libp2p/go-libp2p-core/blob/dc718fa4dab1866476fd9f379718fdd619455a4f/peer/peer.go#L23-L34

FlorianFranzen:
[...]

PR #1485, opened 2020-03-09
Add addresses field for closing listeners

tcharding:
We would like to be able expire addresses when a listener closes.

Add an `addresses` field to the `ListenersEvent` and the `ListenerClosed` to hold the addresses of a listener that has just closed.  When we return a `ListenerClosed` network event, loop over the addresses and call `inject_expired_listen_address()` on each one.

Fixes: https://github.com/libp2p/rust-libp2p/issues/1482

### DIFF
--- a/core/src/connection/listeners.rs
+++ b/core/src/connection/listeners.rs
@@ -285,14 +285,14 @@ where
                 Poll::Ready(None) => {
                     return Poll::Ready(ListenersEvent::Closed {
                         listener_id: *listener_project.id,
-                        addresses: listener.addresses.to_vec(),
+                        addresses: listener_project.addresses.drain(..).collect(),
                         reason: Ok(()),
                     })
                 }
                 Poll::Ready(Some(Err(err))) => {
                     return Poll::Ready(ListenersEvent::Closed {
                         listener_id: *listener_project.id,
-                        addresses: listener.addresses.to_vec(),
+                        addresses: listener_project.addresses.drain(..).collect(),
                         reason: Err(err),
                     })
                 }
--- a/core/src/connection/listeners.rs
+++ b/core/src/connection/listeners.rs
@@ -149,7 +149,7 @@ where
         /// The ID of the listener that closed.
         listener_id: ListenerId,
         /// The addresses that the listener was listening on.
-        addresses: SmallVec<[Multiaddr; 4]>,
+        addresses: Vec<Multiaddr>,
         /// Reason for the closure. Contains `Ok(())` if the stream produced `None`, or `Err`
         /// if the stream produced an error.
         reason: Result<(), TTrans::Error>,
@@ -285,14 +285,14 @@ where
                 Poll::Ready(None) => {
                     return Poll::Ready(ListenersEvent::Closed {
                         listener_id: *listener_project.id,
-                        addresses: listener.addresses.clone(),
+                        addresses: listener.addresses.to_vec(),
                         reason: Ok(()),
                     })
                 }
                 Poll::Ready(Some(Err(err))) => {
                     return Poll::Ready(ListenersEvent::Closed {
                         listener_id: *listener_project.id,
-                        addresses: listener.addresses.clone(),
+                        addresses: listener.addresses.to_vec(),
                         reason: Err(err),
                     })
                 }
--- a/core/src/connection/listeners.rs
+++ b/core/src/connection/listeners.rs
@@ -148,6 +148,8 @@ where
     Closed {
         /// The ID of the listener that closed.
         listener_id: ListenerId,
+        /// The addresses that the listener was listening on.
+        addresses: SmallVec<[Multiaddr; 4]>,
         /// Reason for the closure. Contains `Ok(())` if the stream produced `None`, or `Err`
         /// if the stream produced an error.
         reason: Result<(), TTrans::Error>,
@@ -283,12 +285,14 @@ where
                 Poll::Ready(None) => {
                     return Poll::Ready(ListenersEvent::Closed {
                         listener_id: *listener_project.id,
+                        addresses: listener.addresses.clone(),
                         reason: Ok(()),
                     })
                 }
                 Poll::Ready(Some(Err(err))) => {
                     return Poll::Ready(ListenersEvent::Closed {
                         listener_id: *listener_project.id,
+                        addresses: listener.addresses.clone(),
                         reason: Err(err),
                     })
                 }
@@ -351,9 +355,10 @@ where
                 .field("listener_id", listener_id)
                 .field("local_addr", local_addr)
                 .finish(),
-            ListenersEvent::Closed { listener_id, reason } => f
+            ListenersEvent::Closed { listener_id, addresses, reason } => f
                 .debug_struct("ListenersEvent::Closed")
                 .field("listener_id", listener_id)
+                .field("addresses", addresses)
                 .field("reason", reason)
                 .finish(),
             ListenersEvent::Error { listener_id, error } => f
--- a/core/src/network/event.rs
+++ b/core/src/network/event.rs
@@ -44,6 +44,7 @@ use crate::{
 };
 use futures::prelude::*;
 use std::{error, fmt, hash::Hash};
+use smallvec::SmallVec;
 
 /// Event that can happen on the `Network`.
 pub enum NetworkEvent<'a, TTrans, TInEvent, TOutEvent, THandler, TConnInfo, TPeerId>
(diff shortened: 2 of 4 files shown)


tomaka:
Our code isn't formatted with `rustfmt`, so please remove the first commit!

tcharding:
> Our code isn't formatted with `rustfmt`, so please remove the first commit!

Oh, that would explain the indentation for your `macro_use!` blocks (which is super nice).

Will change my editor setup to match you project and re-submit.

Thanks @tomaka

tcharding:
> Our code isn't formatted with `rustfmt`

I see you guys use `.editorconfig` instead.  In order to educate myself; what is the reason behind using this instead of `rustfmt` please?

tcharding:
FTR, I did not configure my editor to use `.editorconfig`.  If CI fails (and before patching this repo again) I will invest the time to do so.

tcharding:
I'm not sure how to proceed with the CI integration test failure.

tcharding:
Researching further, `rustfmt` - is there a reason libp2p does not use `rustfmt.toml` to get a formatting style we like?

Discussed here: https://www.reddit.com/r/rust/comments/9jl6a9/pro_tip_if_you_use_cargo_fmtrustfmt_use_a/

mxinden:
> I'm not sure how to proceed with the CI integration test failure.

As far as I can tell this is unrelated to your patch set. (See https://github.com/libp2p/rust-libp2p/issues/1323.)

> In order to educate myself; what is the reason behind using this instead of rustfmt please?

You might find some answers in https://github.com/libp2p/rust-libp2p/pull/1164 and the resources linked within.

mxinden:
Retriggered the tests. All green now.

core/src/connection/listeners.rs
```diff
@@ -148,6 +148,8 @@ where
     Closed {
         /// The ID of the listener that closed.
         listener_id: ListenerId,
+        /// The addresses that the listener was listening on.
+        addresses: SmallVec<[Multiaddr; 4]>,
```
tomaka: This is a nit, but I'd rather go with `Vec` in order to not expose a third-party dependency in our API.
Alternatively, a custom iterator type that wraps around `smallvec::IntoIter` would work as well.

core/src/connection/listeners.rs
```diff
@@ -283,12 +285,14 @@ where
                 Poll::Ready(None) => {
                     return Poll::Ready(ListenersEvent::Closed {
                         listener_id: *listener_project.id,
+                        addresses: listener.addresses.clone(),
```
tomaka: Since we're destroying the listener, you don't need to clone.
```suggestion
                        addresses: mem::replace(listener.addresses, SmallVec::new()),
```

(or, for `Vec`, `listener.addresses.drain(..).collect()`)

core/src/network/event.rs
```diff
@@ -55,6 +56,8 @@ where
     ListenerClosed {
         /// The listener ID that closed.
         listener_id: ListenerId,
+        /// The addresses that the listener was listening on.
+        addresses: SmallVec<[Multiaddr; 4]>,
```
tomaka: Same remark

core/src/connection/listeners.rs
tcharding: No worries, will fix and re-spin.  Thanks.

core/src/connection/listeners.rs
tcharding: I went with a `Vec` and used the `SmallVec` method `to_vec()` for the  conversion.

tcharding:
Whats the rust-libp2p projects protocol on squashing/rebasing of PRs?  Would you prefer the second commit (added after review) to be squashed into the first or left as it is?  Thanks for the reviews.

mxinden:
> Whats the rust-libp2p projects protocol on squashing/rebasing of PRs? Would you prefer the second commit (added after review) to be squashed into the first or left as it is? Thanks for the reviews.

@tcharding no action required from your side. We use Github's _Squash and merge_ feature.

core/src/connection/listeners.rs
```diff
@@ -283,12 +285,14 @@ where
                 Poll::Ready(None) => {
                     return Poll::Ready(ListenersEvent::Closed {
                         listener_id: *listener_project.id,
+                        addresses: listener.addresses.to_vec(),
```
tomaka: It's still cloning the content of the container, which is what I was trying to avoid with my suggestion.
```suggestion
                        addresses: listener.addresses.drain(..).collect(),
```

core/src/connection/listeners.rs
```diff
@@ -283,12 +285,14 @@ where
...
                         reason: Ok(()),
                     })
                 }
                 Poll::Ready(Some(Err(err))) => {
                     return Poll::Ready(ListenersEvent::Closed {
                         listener_id: *listener_project.id,
+                        addresses: listener.addresses.to_vec(),
```
tomaka:
```suggestion
                        addresses: listener.addresses.drain(..).collect(),
```

core/src/connection/listeners.rs
tcharding: Its hard to get good help :)  I did try your suggestion but it didn't seem to work.  Will explore further.  Thanks

tcharding: I've had a bit more of a play with it.  Please correct me if I'm wrong but it seems we cannot get a mutable reference to `listener` (needed if we want to drain `addresses`) because we have already taken a mutable reference.  We would need the following code to be possible (I think, open to correction):
```
    /// Provides an API similar to `Stream`, except that it cannot end.
    pub fn poll(mut self: Pin<&mut Self>, cx: &mut Context) -> Poll<ListenersEvent<TTrans>> {
        let mut remaining = self.listeners.len();
        while let Some(mut listener) = self.listeners.pop_back() {
            let mut listener_project = listener.as_mut().project();
            let mut addresses = listener.as_mut().addresses;
```
This is clearly not possible.  Is there another way to do this?

tomaka: Ah sorry, you just have to use `listener_project`
```suggestion
                        addresses: listener_project.addresses.drain(..).collect(),
```

tcharding: Cool, thanks.  I learned (a little bit) about pin_project, seems I still have a lot to learn about pinning :) Thanks for your patient revies @tomaka.  One final thing, just to make sure, are you happy with the remaining usage of `SmallVec` in `connection/listener.rs` introduced in this PR?
```
Closed {
         /// The ID of the listener that closed.
         listener_id: ListenerId,
+        /// The addresses that the listener was listening on.
+        addresses: SmallVec<[Multiaddr; 4]>,
```

And, since I'm new and its good to start on the right foot; can you please confirm that the typical pre-PR-push tasks you would expect to pass are only:
- `cargo build`
- `cargo test`

(i.e., no clippy, no e2e tests etc)

Thanks,
Tobin.

tomaka: > are you happy with the remaining usage of SmallVec in connection/listener.rs introduced in this PR?

Ah! No, I missed that. Please turn that into a `Vec` as well.

There are more tests than `cargo test`, but running `cargo test` locally is good enough before opening a PR.

tcharding: Done, I think we are getting dangerously close to a mergeable PR :)

tomaka: Wait. Please revert that last change. Having `SmallVec` in a private field is good.
(you mentioned "introduced in this PR", but your PR didn't touch this.)

tcharding: Oh boy, you are right.  I hope you meant 'revert' figuratively, I removed last patch and force pushed.  Thanks for your patience.

swarm/src/lib.rs
```diff
@@ -445,9 +445,12 @@ where TBehaviour: NetworkBehaviour<ProtocolsHandler = THandler>,
                     this.behaviour.inject_expired_listen_addr(&listen_addr);
                     return Poll::Ready(SwarmEvent::ExpiredListenAddr(listen_addr));
                 }
-                Poll::Ready(NetworkEvent::ListenerClosed { listener_id, reason }) => {
+                Poll::Ready(NetworkEvent::ListenerClosed { listener_id, addresses, reason }) => {
                     log::debug!("Listener {:?}; Closed by {:?}.", listener_id, reason);
                     this.behaviour.inject_listener_closed(listener_id);
+                    for addr in addresses.iter() {
+                        this.behaviour.inject_expired_listen_addr(addr);
+                    }
```
tomaka: One last change: could you move this above `inject_listener_closed`?
The listener and its addresses technically expire at the same time, but since here we have to pick an order, it makes more sense that the addresses expire before.

tcharding: No worries, done!

PR #1486, opened 2020-03-09
Add logging to the integration test

tomaka:
cc #1323 

Should help understand what's going on.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -105,7 +105,7 @@ jobs:
           command: |
             rustc --version
       - run:
-          command: RUST_LOG=libp2p_kad cargo run --example ipfs-kad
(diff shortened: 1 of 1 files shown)


tomaka:
I think it's the first time I re-run CI because it's green and I want it to be red.

romanb:
Using `RUST_LOG=debug` may give more insight, because the curious thing is that when the example fails it is due to 10 minutes without output, supposedly, yet the query timeout is set to 5 minutes. If the query timeout isn't utterly broken, this may even mean that a code path somewhere outside libp2p-kad stalls / hangs. I've also never been able to reproduce such a 10 minute hang running the example locally, which may furthermore indicate that the environment in which the example is run in CI is another contributing factor.

romanb:
Superseded by https://github.com/libp2p/rust-libp2p/pull/1493.

Issue #1491, opened 2020-03-11
Panic in libp2p-kad


arkpar:
Polkadot panics when running debug build after a while.

```
2020-03-11 03:44:04 Imported #1398447 (0x5bb8…4864)
2020-03-11 03:44:06 Polling the network future took 8.256043471s
2020-03-11 03:44:07 Idle (65 peers), best: #1398447 (0x39c1…b096), finalized #1398445 (0x5ea6…7b95), ⬇ 163.5kiB/s ⬆ 24.3kiB/s
2020-03-11 03:44:09 Imported #1398448 (0x9bf7…91d8)

====================

Version: 0.7.22-61e5d69e-x86_64-linux-gnu

stack backtrace:
   0: sp_panic_handler::panic_hook
             at /home/arkadiy/.cargo/git/checkouts/substrate-7e08433d4c370a21/a439a7a/primitives/panic-handler/src/lib.rs:148
   1: sp_panic_handler::set::{{closure}}
             at /home/arkadiy/.cargo/git/checkouts/substrate-7e08433d4c370a21/a439a7a/primitives/panic-handler/src/lib.rs:58
   2: std::panicking::rust_panic_with_hook
             at src/libstd/panicking.rs:475
   3: rust_begin_unwind
             at src/libstd/panicking.rs:375
   4: core::panicking::panic_fmt
             at src/libcore/panicking.rs:84
   5: core::panicking::panic
             at src/libcore/panicking.rs:51
   6: libp2p_kad::behaviour::Kademlia<TStore>::record_received::{{closure}}
             at /home/arkadiy/.cargo/registry/src/github.com-1ecc6299db9ec823/libp2p-kad-0.16.2/src/behaviour.rs:947
   7: core::option::Option<T>::map
             at /rustc/5e1a799842ba6ed4a57e91f7ab9435947482f7d8/src/libcore/option.rs:450
   8: libp2p_kad::behaviour::Kademlia<TStore>::record_received
             at /home/arkadiy/.cargo/registry/src/github.com-1ecc6299db9ec823/libp2p-kad-0.16.2/src/behaviour.rs:946
   9: <libp2p_kad::behaviour::Kademlia<TStore> as libp2p_swarm::behaviour::NetworkBehaviour>::inject_node_event
             at /home/arkadiy/.cargo/registry/src/github.com-1ecc6299db9ec823/libp2p-kad-0.16.2/src/behaviour.rs:1295
[...]

PR #1492, opened 2020-03-11
protocols/kad: Fix right shift overflow panic in record_received 

mxinden:
Within `Behaviour::record_received` the exponentially decreasing
expiration based on the distance to the target for a record is
calculated as following:

1. Calculate the amount of nodes between us and the record key beyond
the k replication constant as `n`.

2. Shift the configured record time-to-live `n` times to the right to
calculate an exponentially decreasing expiration.

The configured record time-to-live is a u64. If `n` is larger or equal
to 64 the right shift will lead to an overflow which panics in debug
mode.

This patch uses a checked right shift instead, defaulting to 0 (`now +
0`) for the expiration on overflow.

Fixes https://github.com/libp2p/rust-libp2p/issues/1491.

### DIFF
--- a/protocols/kad/src/behaviour.rs
+++ b/protocols/kad/src/behaviour.rs
@@ -946,8 +946,6 @@ where
             return
         }
 
-        let now = Instant::now();
-
         // Calculate the expiration exponentially inversely proportional to the
         // number of nodes between the local node and the closest node to the key
         // (beyond the replication factor). This ensures avoiding over-caching
@@ -956,9 +954,7 @@ where
         let num_between = self.kbuckets.count_nodes_between(&target);
         let k = self.queries.config().replication_factor.get();
         let num_beyond_k = (usize::max(k, num_between) - k) as u32;
-        let expiration = self.record_ttl.map(|ttl|
-            now + Duration::from_secs(ttl.as_secs().checked_shr(num_beyond_k).unwrap_or(0))
-        );
+        let expiration = exp_decr_expiration(self.record_ttl, num_beyond_k);
         // The smaller TTL prevails. Only if neither TTL is set is the record
         // stored "forever".
         record.expires = record.expires.or(expiration).min(expiration);
@@ -1028,6 +1024,15 @@ where
             }
         }
     }
+}
+
+/// Calculate exponentially decreasing expiration from a default time-to-live by a factor.
+fn exp_decr_expiration(default_ttl: Option<Duration>, factor: u32) -> Option<Instant> {
+    default_ttl.map(|ttl| Instant::now() + Duration::from_secs(
+        ttl.as_secs()
+            .checked_shr(factor)
+            .unwrap_or(0),
+    ))
 }
 
 impl<TStore> NetworkBehaviour for Kademlia<TStore>
--- a/protocols/kad/src/behaviour.rs
+++ b/protocols/kad/src/behaviour.rs
@@ -957,7 +957,7 @@ where
         let k = self.queries.config().replication_factor.get();
         let num_beyond_k = (usize::max(k, num_between) - k) as u32;
         let expiration = self.record_ttl.map(|ttl|
-            now + Duration::from_secs(ttl.as_secs() >> num_beyond_k)
+            now + Duration::from_secs(ttl.as_secs().checked_shr(num_beyond_k).unwrap_or(0))
         );
         // The smaller TTL prevails. Only if neither TTL is set is the record
         // stored "forever".
@@ -1911,4 +1911,3 @@ impl QueryInfo {
         }
     }
 }
-
--- a/protocols/kad/src/behaviour.rs
+++ b/protocols/kad/src/behaviour.rs
@@ -946,6 +946,8 @@ where
             return
         }
 
+        let now = Instant::now();
+
         // Calculate the expiration exponentially inversely proportional to the
         // number of nodes between the local node and the closest node to the key
         // (beyond the replication factor). This ensures avoiding over-caching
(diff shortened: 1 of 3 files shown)


protocols/kad/src/behaviour/test.rs
```diff
@@ -682,3 +684,44 @@ fn exceed_jobs_max_queries() {
         })
     )
 }
+
+#[test]
```
twittner: I think in libp2p the convention is to put attributes below comments.

protocols/kad/src/handler.rs
```diff
@@ -364,6 +364,15 @@ pub struct KademliaRequestId {
     connec_unique_id: UniqueConnecId,
 }
 
+#[cfg(test)]
+impl Default for KademliaRequestId {
```
twittner: As long as `KademliaRequestId` implements `Clone`, this `Default` impl might as well be available all the time.

tomaka: This is an opaque implementation detail. Implementing `Default` on it doesn't make any sense. Maybe it could contain a channel or an `Arc` in the future.

protocols/kad/src/behaviour/test.rs
```diff
@@ -682,3 +684,44 @@ fn exceed_jobs_max_queries() {
...
+    }
+
+    let connection_id = ConnectionId::new(1234);
+    let request_id = KademliaRequestId::default();
+    let record = Record::new(
+        record::Key::from(target.clone().into_bytes()),
+        vec![],
+    );
+
+    kad.record_received(target, connection_id, request_id, record);
```
tomaka: We shouldn't be testing private implementation details. The parameter of this method are totally arbitrary. The fact that it's a separate method is also totally arbitrary.

protocols/kad/src/handler.rs
twittner: I understood `KademliaRequestId` to be some sort of unique one-time token and `Clone` defeats that purpose but I did not look much into it so feel free to ignore the criticism.

protocols/kad/src/behaviour.rs
```diff
@@ -957,7 +957,7 @@ where
         let k = self.queries.config().replication_factor.get();
         let num_beyond_k = (usize::max(k, num_between) - k) as u32;
         let expiration = self.record_ttl.map(|ttl|
-            now + Duration::from_secs(ttl.as_secs() >> num_beyond_k)
+            now + Duration::from_secs(ttl.as_secs().checked_shr(num_beyond_k).unwrap_or(0))
```
arkpar: Does it make sense to put records that are immediately expired in the store on line 987

protocols/kad/src/handler.rs
mxinden: I am fine either way. I can't come up with a use-case for `Default` outside of testing though.

protocols/kad/src/behaviour/test.rs
mxinden: I understand your concern. What are you advocating for instead: (1) remove the test or (2) write a larger test that reproduces the same behaviour?

> All we're doing with this test is locking the current implementation as it is.

I don't think a test should ever _lock_ an implementation. While I am in favor of having a test for each regression, I am also in favor of removing outdated tests if necessary.

protocols/kad/src/behaviour.rs
mxinden: As far as I can tell the record would be removed by the next `PutRecordJob::poll` invocation. Maybe @romanb has a better answer.

Never the less I would advocate for addressing this in a follow up pull-request.

protocols/kad/src/handler.rs
romanb: > I understood `KademliaRequestId` to be some sort of unique one-time token and `Clone` defeats that purpose but I did not look much into it so feel free to ignore the criticism.

Just to give some context: `KademliaRequestId` did not use to be `Clone` but it had to be as a result of https://github.com/libp2p/rust-libp2p/pull/1440 requiring `Clone` for the `TInEvent` of any handler, since it is now possible to "broadcast" a single such event to multiple handlers (which in turn was a requirement discovered during the substrate integration in https://github.com/paritytech/substrate/pull/5066) and `KademliaRequestId` is part of these events in `libp2p-kad`. It would of course be better to not have that `Clone`, but I didn't come up with a way to do so.

protocols/kad/src/behaviour.rs
romanb: Yes, expired records are regularly purged (and a `RecordStore` may of course just drop expired records at any time anyway), but it may nevertheless be a good optimisation that may also make it a bit more difficult to somewhat maliciously or accidentally (temporarily) fill a node's store with already expired records. @mxinden if you could create an issue for it, that would be great, thanks!

mxinden: See https://github.com/libp2p/rust-libp2p/pull/1496.

romanb:
Thanks for fixing this, I wasn't even aware bit shifts could overflow in this way. A regression test is a good idea but I don't see why it needs to be so complicated. Why not extract the code that calculates the expiry into a standalone function and writing small property and or / regresstion test for that? Even if you want the function to actually take a `KBucketsTable` as argument, so that it uses `count_nodes_between` internally, there shouldn't be a need for such a probabilistic generation of so many peer IDs, since `Key::for_distance` can be used in combination with the `ClosestBucketsIter` to deterministically fill the buckets as desired ([this test](https://github.com/libp2p/rust-libp2p/blob/master/protocols/kad/src/kbucket.rs#L641-L661) may give inspiration). In other words, when testing a particular edge case requires seemingly too much surrounding boilerplate in order to test what you want, it is often a good idea to partition the code further so that the part you're interested in testing in isolation is easier to test, without so much surrounding ceremony. This seems to be such a case.

mxinden:
@romanb thanks for the input. Can you take another look?

protocols/kad/src/behaviour.rs
```diff
@@ -946,8 +946,6 @@ where
             return
         }
 
-        let now = Instant::now();
```
romanb: I think it is preferable to leave this here as it can be reused multiple times (cf. https://github.com/libp2p/rust-libp2p/pull/1496). And since `Instant::now()` is a side-effecting function for obtaining some notion of system time, usually both testability and performance is improved by using it mostly in top-level functions and passing the instant down to others where needed (see also the other comments).

mxinden:
Suggested simplifications look good to me. @romanb would you mind taking another look?

PR #1493, opened 2020-03-11
Fix regression w.r.t. reporting of dial errors.

romanb:
https://github.com/libp2p/rust-libp2p/pull/1440 introduced a regression w.r.t. the reporting of dial errors. In particular, if a connection attempt fails due to an invalid remote peer ID, any remaining addresses for the same peer would not be tried (intentional) but the dial failure would not be reported to the behaviour, causing e.g. libp2p-kad queries to potentially stall.

In hindsight, I figured it is better to preserve the previous behaviour to still try alternative addresses
of the peer even on invalid peer ID errors on an earlier address. In particular because in the context of libp2p-kad it is not uncommon for peers to report localhost addresses while the local node actually has e.g. an ipfs node running on that address, obviously with a different peer ID, which is the scenario causing frequent invalid peer ID (mismatch) errors when running the ipfs-kad example in the go-ipfs docker container.

This PR thus restores the previous behaviour w.r.t. trying all remaining addresses on invalid peer ID errors, as well as making sure `inject_dial_error` is always called when the last attempt failed, regardless of whether the peer is already connected or not (e.g. as a result of a simultaneous incoming connection).

Overall this is a slight simplification, though requiring some additional cloning of peer IDs (as was also done before https://github.com/libp2p/rust-libp2p/pull/1440).

This PR restores the ipfs-kad example as an integration test for CI, which should now be stable again.

### DIFF
--- a/core/src/connection/pool.rs
+++ b/core/src/connection/pool.rs
@@ -555,7 +555,7 @@ where
         PoolEvent<'a, TInEvent, TOutEvent, THandler, TTransErr, THandlerErr, TConnInfo, TPeerId>
     > where
         TConnInfo: ConnectionInfo<PeerId = TPeerId> + Clone,
-        TPeerId: Clone + fmt::Debug,
+        TPeerId: Clone
     {
         loop {
             let item = match self.manager.poll(cx) {
@@ -611,9 +611,13 @@ where
                         }
                         // Peer ID checks must already have happened. See `add_pending`.
                         if cfg!(debug_assertions) {
-                            assert_ne!(&self.local_id, entry.connected().peer_id());
+                            if &self.local_id == entry.connected().peer_id() {
+                                panic!("Unexpected local peer ID for remote.");
+                            }
                             if let Some(peer) = peer {
-                                assert_eq!(&peer, entry.connected().peer_id());
+                                if &peer != entry.connected().peer_id() {
+                                    panic!("Unexpected peer ID mismatch.");
+                                }
                             }
                         }
                         // Add the connection to the pool.
--- a/core/src/connection/pool.rs
+++ b/core/src/connection/pool.rs
@@ -112,7 +112,7 @@ pub enum PoolEvent<'a, TInEvent, TOutEvent, THandler, TTransErr, THandlerErr, TC
         error: PendingConnectionError<TTransErr>,
         /// The handler that was supposed to handle the connection,
         /// if the connection failed before the handler was consumed.
-        handler: Option<THandler>,
+        handler: THandler,
         /// The (expected) peer of the failed connection.
         peer: Option<TPeerId>,
         /// A reference to the pool that managed the connection.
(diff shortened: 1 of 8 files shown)


core/src/connection/pool.rs
```diff
@@ -581,40 +600,22 @@ where
...
-                                    peer: Some(connected.info.peer_id().clone()),
-                                    error: PendingConnectionError::InvalidPeerId,
-                                    pool: self,
-                                    handler: None,
-                                })
+                        // Peer ID checks must already have happened. See `add_pending`.
+                        if cfg!(debug_assertions) {
+                            assert_ne!(&self.local_id, entry.connected().peer_id());
+                            if let Some(peer) = peer {
+                                assert_eq!(&peer, entry.connected().peer_id());
```
tomaka: I suppose that this is because you require `Debug` above?
If so, I'd prefer to not enforce `Debug` if possible.

```suggestion
                            assert_ne!(&self.local_id, entry.connected().peer_id(), "Unexpected local peer ID");
                            if let Some(peer) = peer {
                                assert_eq!(&peer, entry.connected().peer_id(), "PeerId mismatch");
```

core/src/connection/pool.rs
```diff
@@ -536,7 +555,7 @@ where
         PoolEvent<'a, TInEvent, TOutEvent, THandler, TTransErr, THandlerErr, TConnInfo, TPeerId>
     > where
         TConnInfo: ConnectionInfo<PeerId = TPeerId> + Clone,
-        TPeerId: Clone,
+        TPeerId: Clone + fmt::Debug,
```
tomaka:
```suggestion
        TPeerId: Clone,
```

core/src/network.rs
```diff
@@ -331,7 +330,7 @@ where
         THandler::Handler: ConnectionHandler<Substream = Substream<TMuxer>, InEvent = TInEvent, OutEvent = TOutEvent> + Send + 'static,
         <THandler::Handler as ConnectionHandler>::Error: error::Error + Send + 'static,
         TConnInfo: Clone,
-        TPeerId: AsRef<[u8]> + Send + 'static,
+        TPeerId: fmt::Debug + Send + 'static,
```
tomaka:
```suggestion
        TPeerId: Send + 'static,
```

core/src/connection/pool.rs
romanb: `fmt::Debug` is required by `assert_eq!` and `assert_ne!` for the arguments, so are you suggesting something like:
```rust
if &self.local_id == entry.connected().peer_id() {
    panic!("...")
}
if let Some(peer) = peer {
    if peer != entry.connected().peer_id() {
        panic!("...")
    }
}
```
? It is unfortunate though not to have the problematic peer IDs in the output. Why is it undesirable to require `fmt::Debug`?

romanb: I have now done this: https://github.com/libp2p/rust-libp2p/pull/1493/commits/5807155992cba55e18a1cd5e41d2b1b9b68e6163

tomaka: > Why is it undesirable to require fmt::Debug?

I just see it as "more correct" to only enforce the minimum possible set of traits.
In an ideal world, we would print the `PeerId` if it implements `Debug` and not print it if it doesn't. But that's unfortunately not possible.

PR #1494, opened 2020-03-12
Add Kademlia::num_active_iterative_queries

tomaka:
Small getter for diagnostics purposes.

I also renamed `size()` to `len()`, as in idiomatic Rust `size()` normally designates a number of bytes while `len()` is a number of elements.


protocols/kad/src/behaviour.rs
```diff
@@ -441,6 +441,11 @@ where
         &mut self.store
     }
 
+    /// Returns the number of iterative queries that are currently being performed.
+    pub fn num_active_iterative_queries(&self) -> usize {
```
```suggestion
    pub fn num_active_queries(&self) -> usize {
```

A query could also be a [`FixedPeersIter`](https://github.com/libp2p/rust-libp2p/blob/master/protocols/kad/src/query/peers/fixed.rs#L28) which only operates on a predefined set of peers.

```suggestion
    pub fn num_queries(&self) -> usize {
```
Because there is no such thing as an inactive query, I don't think `active` is a useful distinction and may be misleading.

tomaka:
About the naming: I don't care so much about the word `active`, but in my opinion it'd be better to have the word `iterative` in it, to distinguish that from so-called "RPC queries" (a single query to a node).

romanb:
> About the naming: I don't care so much about the word `active`, but in my opinion it'd be better to have the word `iterative` in it, to distinguish that from so-called "RPC queries" (a single query to a node).

But the point is that `self.queries` is a pool of all kinds of queries, not just iterative queries. Furthermore, iterative queries are often just one part / half of a "query" from the user's perspective, e.g. writing a record to the DHT is an iterative query _followed by_ `k` direct RPCs to the resulting peers. The latter step is also a query in `self.queries`, a query that has an underlying `FixedPeersIter` going over these peers, as @mxinden mentioned. Even more: When a lookup of a record succeeds, the record is stored (cached) at the closest node that _did not return the record_, which is also a query in `self.queries`, with a singleton `FixedPeersIter`.

romanb:
> > About the naming: I don't care so much about the word `active`, but in my opinion it'd be better to have the word `iterative` in it, to distinguish that from so-called "RPC queries" (a single query to a node).
> 
> But the point is that `self.queries` is a pool of all kinds of queries, not just iterative queries. Furthermore, iterative queries are often just one part / half of a "query" from the user's perspective, e.g. writing a record to the DHT is an iterative query _followed by_ `k` direct RPCs to the resulting peers. The latter step is also a query in `self.queries`, a query that has an underlying `FixedPeersIter` going over these peers, as @mxinden mentioned. Even more: When a lookup of a record succeeds, the record is stored (cached) at the closest node that _did not return the record_, which is also a query in `self.queries`, with a singleton `FixedPeersIter`.

So the current definition of "query" is essentially just any number of RPCs performed in the context of a particular operation / for a particular purpose.

romanb:
Maybe it would be more useful to expose some kind of `QueryStats` struct which goes over `self.queries` and counts e.g. the number of queries based on their inner "QueryInfo" type, so you would get numbers on "bootstrap", "get_closests_peers", etc.

tomaka:
> Maybe it would be more useful to expose some kind of QueryStats struct which goes over self.queries and counts e.g. the number of queries based on their inner "QueryInfo" type, so you would get numbers on "bootstrap", "get_closests_peers", etc.

That's indeed probably a better idea.

However I went for `num_queries(&self) -> usize` because it is a simple API.

If we go into the details of the queries, then rather than just giving raw numbers of queries, I'd rather instead have a `queries(&self) -> impl Iterator<Item = QueryInfo>` method that explicitly enumerates ongoing queries, so that users can gather whatever statistics they want.

romanb:
> > Maybe it would be more useful to expose some kind of QueryStats struct which goes over self.queries and counts e.g. the number of queries based on their inner "QueryInfo" type, so you would get numbers on "bootstrap", "get_closests_peers", etc.
> 
> That's indeed probably a better idea.
> 
> However I went for something like `num_queries(&self) -> usize` because it is a very simple API.
> 
> If we go into the details of the queries, then rather than just giving raw numbers of queries, I'd rather instead have a `fn queries(&self) -> impl Iterator<Item = QueryInfo>` method that explicitly enumerates ongoing queries, so that users can gather whatever statistics they want.
> 
> Does what make sense?

The only problem I see with that is that `QueryInfo` as it currently is should not be exposed, as it contains implementation details, e.g. some enum variants have fields like `external_addresses` and `context`, meaning this enum is supposed to also be able to hold internal state for a query. So I would suggest to at least map the `QueryInfo`s to another, public enum which only exposes what we want to expose (maybe just the query type to start with?).

romanb:
@tomaka I can take over this one and see if I can come up with something along the lines discussed above.

tomaka:
> I can take over this one and see if I can come up with something along the lines discussed above.

Thanks, feel free to do that.

romanb:
I will open a new PR by the end of the day that is supposed to supersede this one.

romanb:
Closing in favour of #1567

PR #1495, opened 2020-03-12
.circleci: Add script to test that Rust attributes are below comments

mxinden:
To safe me a couple of minutes next time I wonder whether attributes go above or below comments.

### DIFF
--- a/.circleci/attributes_below_comments.sh
+++ b/.circleci/attributes_below_comments.sh
@@ -1,0 +1,32 @@
+#!/usr/bin/env bash
+# only exit with zero if all commands of the pipeline exit successfully
+set -o pipefail
+# error on unset variables
+set -u
+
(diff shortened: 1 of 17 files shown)


tomaka:
We're kind of switching to GitHub Actions and out of CircleCI, so this would need to be done through GH actions.

(also, this would also be enforced by rustfmt, which I'm still in favour of)

mxinden:
@tomaka could you take another look?

tomaka:
I find it a bit exaggerated to have a test just for this, and I'd be in favour of just using rustfmt, but I'm also not really opposed to it.

You have to rename the script though, as we removed the `.circleci` directory.

mxinden:
> I find it a bit exaggerated to have a test just for this.

Yes, agreed. In addition the more time I spend on this, the more I feel like falling victim to the sunk cost fallacy.

> and I'd be in favour of just using rustfmt

I second this. Even though this particular case would not be covered (as far as I can tell from a quick test).

PR #1496, opened 2020-03-12
protocols/kad: Do not attempt to store expired record in record store

mxinden:
`Kademlia::record_received` calculates the expiration time of a record
before inserting it into the record store. Instead of inserting the
record into the record store in any case, with this patch the record is
only inserted if it is not expired. If the record is expired a
`KademliaHandlerIn::Reset` for the given (sub) stream is triggered.

This would serve as a tiny defense mechanism against an attacker trying
to fill a node's record store with expired records before the record
store's clean up procedure removes the records.

---

This is a follow up for https://github.com/libp2p/rust-libp2p/pull/1492#discussion_r391765926. Instead of an issue I created a pull request right away given that the change set should be reasonably small.

**Major discussion point: Should we trigger a `KademliaHandlerIn::PutRecordRes` or a `KademliaHandlerIn::Reset` when the record is expired?**

 As far as I can tell the former would inform the remote that the value [was indeed put](https://github.com/libp2p/rust-libp2p/blob/master/protocols/kad/src/handler.rs#L573) into the record store whereas the latter would simply have the remote [remove the substream](https://github.com/libp2p/rust-libp2p/blob/master/protocols/kad/src/handler.rs#L461). The former would lead to an `on_success` call on the query, the latter would have the query retry another node.

Consulting the [specification](https://github.com/libp2p/specs/pull/108) did not reveal any suggestions. Reading the go implementation I failed to determine where the [handler for a put event](https://github.com/libp2p/go-libp2p-kad-dht/blob/95964c0eac86887e54559a689f7e8f68b4a0f69f/handlers.go#L157) determines whether a given record is expired, given that it only seems to instantiate a [`PublicKeyValidator`](https://github.com/libp2p/go-libp2p-record/blob/master/pubkey.go#L20).

I decided for resetting the connection and thereby not (wrongly) acknowledging that the node cached the record. What do you think?

### DIFF
--- a/protocols/kad/src/behaviour.rs
+++ b/protocols/kad/src/behaviour.rs
@@ -981,31 +981,35 @@ where
         // overridden as it avoids having to load the existing record in the
         // first place.
 
-        // The record is cloned because of the weird libp2p protocol requirement
-        // to send back the value in the response, although this is a waste of
-        // resources.
-        match self.store.put(record.clone()) {
-            Ok(()) => {
-                debug!("Record stored: {:?}; {} bytes", record.key, record.value.len());
-                self.queued_events.push_back(NetworkBehaviourAction::NotifyHandler {
-                    peer_id: source,
-                    handler: NotifyHandler::One(connection),
-                    event: KademliaHandlerIn::PutRecordRes {
-                        key: record.key,
-                        value: record.value,
-                        request_id,
-                    },
-                })
-            }
-            Err(e) => {
-                info!("Record not stored: {:?}", e);
-                self.queued_events.push_back(NetworkBehaviourAction::NotifyHandler {
-                    peer_id: source,
-                    handler: NotifyHandler::One(connection),
-                    event: KademliaHandlerIn::Reset(request_id)
-                })
+        if !record.is_expired(Instant::now()) {
+            // The record is cloned because of the weird libp2p protocol requirement
+            // to send back the value in the response, although this is a waste of
+            // resources.
+            match self.store.put(record.clone()) {
+                Ok(()) => {
+                    debug!("Record stored: {:?}; {} bytes", record.key, record.value.len());
+                    self.queued_events.push_back(NetworkBehaviourAction::NotifyHandler {
+                        peer_id: source,
+                        handler: NotifyHandler::One(connection),
+                        event: KademliaHandlerIn::PutRecordRes {
+                            key: record.key,
+                            value: record.value,
+                            request_id,
+                        },
+                    });
+                    return
+                }
+                Err(e) => {
+                    info!("Record not stored: {:?}", e);
+                }
             }
         }
+
+        self.queued_events.push_back(NetworkBehaviourAction::NotifyHandler {
+            peer_id: source,
+            handler: NotifyHandler::One(connection),
+            event: KademliaHandlerIn::Reset(request_id)
+        })
     }
 
     /// Processes a provider record received from a peer.
@@ -1911,4 +1915,3 @@ impl QueryInfo {
         }
     }
 }
-
--- a/protocols/kad/src/behaviour.rs
+++ b/protocols/kad/src/behaviour.rs
@@ -981,7 +981,7 @@ where
         // overridden as it avoids having to load the existing record in the
         // first place.
 
-        if !record.is_expired(Instant::now()) {
+        if !record.is_expired(now) {
             // The record is cloned because of the weird libp2p protocol
             // requirement to send back the value in the response, although this
             // is a waste of resources.
@@ -1004,10 +1004,7 @@ where
         // case where the record is discarded due to being expired. Given that
         // the remote sent the local node a [`KademliaHandlerEvent::PutRecord`]
         // request, the remote perceives the local node as one node among the k
-        // closest nodes to the target. Returning a [`KademliaHandlerIn::Reset`]
-        // instead of an [`KademliaHandlerIn::PutRecordRes`] to have the remote
-        // try another node would only result in the remote node to contact an
-        // even more distant node. In addition returning
+        // closest nodes to the target. In addition returning
         // [`KademliaHandlerIn::PutRecordRes`] does not reveal any internal
         // information to a possibly malicious remote node.
         self.queued_events.push_back(NetworkBehaviourAction::NotifyHandler {
--- a/protocols/kad/src/behaviour.rs
+++ b/protocols/kad/src/behaviour.rs
@@ -982,33 +982,42 @@ where
         // first place.
 
         if !record.is_expired(Instant::now()) {
-            // The record is cloned because of the weird libp2p protocol requirement
-            // to send back the value in the response, although this is a waste of
-            // resources.
+            // The record is cloned because of the weird libp2p protocol
+            // requirement to send back the value in the response, although this
+            // is a waste of resources.
             match self.store.put(record.clone()) {
-                Ok(()) => {
-                    debug!("Record stored: {:?}; {} bytes", record.key, record.value.len());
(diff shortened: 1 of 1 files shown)


romanb:
> I decided for resetting the connection and thereby not (wrongly) acknowledging that the node cached the record. What do you think?

There are a few reasons why I think the correct thing to do is to send a regular acknowledgement:

  1. It is the current behaviour.
  2. An honest node `a` always tries to send a record to the `k` closest nodes to the key. If some node `b` receives such a store request and considers itself so far away from the key that the expiry is instant, then either `a` or `b` has a very distorted view of the overlay network, which may even be maliciously induced. In either case, it is not for `b` to judge by rejecting the record. Since `b` received the request, then evidently `a` considers `b` to be among the `k` closest to the key, so `b` stores it subject to its own modifications to the expiry according to its own view of the network. If that expiry is instant, so be it, that's not an error.
  3. Sending a regular acknowledgement does not reveal any information to a potentially maliciously motivated sender trying to temporarily flood a node's storage, i.e. it is not told through an error that the record key is considered "too far away" by the targeted node. Not unnecessarily giving away such information is usually beneficial.

romanb:
4. Requests to store a record are also received as a result of the "outwards" caching of records via targeted requests to single nodes, that happens during regular record lookups. Silently instantly expiring records when such caching reaches the perimeter of the eligible caching region is, I think, precisely what is desired.

mxinden:
Not triggering a retry because:

- The local node is already among the k closest nodes the remote node is aware of.

- Reaching "the perimeter of the eligible caching region" should not trigger further retries.

At the same time reducing the amount of information returned to possibly malicious actors seems plausible.

I added an additional commit to return a regular acknowledgement. @romanb can you take another look?

protocols/kad/src/behaviour.rs
```diff
@@ -981,31 +981,44 @@ where
...
-                })
-            }
-            Err(e) => {
-                info!("Record not stored: {:?}", e);
-                self.queued_events.push_back(NetworkBehaviourAction::NotifyHandler {
-                    peer_id: source,
-                    handler: NotifyHandler::One(connection),
-                    event: KademliaHandlerIn::Reset(request_id)
-                })
+        if !record.is_expired(Instant::now()) {
```
```suggestion
        if !record.is_expired(now) {
```

protocols/kad/src/behaviour.rs
```diff
@@ -981,31 +981,44 @@ where
...
+                Ok(()) => debug!("Record stored: {:?}; {} bytes", record.key, record.value.len()),
+                Err(e) => {
+                    info!("Record not stored: {:?}", e);
+                    self.queued_events.push_back(NetworkBehaviourAction::NotifyHandler {
+                        peer_id: source,
+                        handler: NotifyHandler::One(connection),
+                        event: KademliaHandlerIn::Reset(request_id)
+                    });
+
+                    return
```
romanb: I would personally prefer to put the remaining code in this branch into an `else` clause for `if !record.is_expired(now)`, because I find early `return`s in the middle of larger code blocks / functions to often trip me up in reasoning about the control flow (as compared to early returns at the very beginning of a function).

protocols/kad/src/behaviour.rs
```diff
@@ -981,31 +981,44 @@ where
...
             }
         }
+
+        // The remote receives a [`KademliaHandlerIn::PutRecordRes`] even in the
+        // case where the record is discarded due to being expired. Given that
+        // the remote sent the local node a [`KademliaHandlerEvent::PutRecord`]
+        // request, the remote perceives the local node as one node among the k
+        // closest nodes to the target. Returning a [`KademliaHandlerIn::Reset`]
+        // instead of an [`KademliaHandlerIn::PutRecordRes`] to have the remote
+        // try another node would only result in the remote node to contact an
```
romanb: In the current implementation, there would not be a retry on another node. Once the iterative query produces the `k` closest nodes to put a record to, this fixed set receives these requests. If one of these requests fails, it counts as (and is reported as) a failure w.r.t. the require quorum for the write operation. So I would suggest to the second part of this comment. The primary reason for not returning an error is because that is precisely how the outwards caching of records with decreasing TTL is supposed to work.

protocols/kad/src/behaviour.rs
mxinden: But then I would need to duplicate the happy case return logic. We want to return `PutRecordRes` both if:

1. The record is expired.

2. The record is successfully put into the record store.

I would end up with:

```diff
diff --git a/protocols/kad/src/behaviour.rs b/protocols/kad/src/behaviour.rs
index 0e6035ad..b34580af 100644
--- a/protocols/kad/src/behaviour.rs
+++ b/protocols/kad/src/behaviour.rs
@@ -986,7 +986,19 @@ where
             // requirement to send back the value in the response, although this
             // is a waste of resources.
             match self.store.put(record.clone()) {
-                Ok(()) => debug!("Record stored: {:?}; {} bytes", record.key, record.value.len()),
+                Ok(()) => {
+                    debug!("Record stored: {:?}; {} bytes", record.key, record.value.len());
+
+                    self.queued_events.push_back(NetworkBehaviourAction::NotifyHandler {
+                        peer_id: source,
+                        handler: NotifyHandler::One(connection),
+                        event: KademliaHandlerIn::PutRecordRes {
+                            key: record.key,
+                            value: record.value,
+                            request_id,
+                        },
+                    })
+                },
                 Err(e) => {
                     info!("Record not stored: {:?}", e);
                     self.queued_events.push_back(NetworkBehaviourAction::NotifyHandler {
@@ -994,31 +1006,29 @@ where
                         handler: NotifyHandler::One(connection),
                         event: KademliaHandlerIn::Reset(request_id)
                     });
-
-                    return
                 }
             }
+        } else {
+            // The remote receives a [`KademliaHandlerIn::PutRecordRes`] even in the
+            // case where the record is discarded due to being expired. Given that
+            // the remote sent the local node a [`KademliaHandlerEvent::PutRecord`]
+            // request, the remote perceives the local node as one node among the k
+            // closest nodes to the target. Returning a [`KademliaHandlerIn::Reset`]
+            // instead of an [`KademliaHandlerIn::PutRecordRes`] to have the remote
+            // try another node would only result in the remote node to contact an
+            // even more distant node. In addition returning
+            // [`KademliaHandlerIn::PutRecordRes`] does not reveal any internal
+            // information to a possibly malicious remote node.
+            self.queued_events.push_back(NetworkBehaviourAction::NotifyHandler {
+                peer_id: source,
+                handler: NotifyHandler::One(connection),
+                event: KademliaHandlerIn::PutRecordRes {
+                    key: record.key,
+                    value: record.value,
+                    request_id,
+                },
+            })
         }
-
-        // The remote receives a [`KademliaHandlerIn::PutRecordRes`] even in the
-        // case where the record is discarded due to being expired. Given that
-        // the remote sent the local node a [`KademliaHandlerEvent::PutRecord`]
-        // request, the remote perceives the local node as one node among the k
-        // closest nodes to the target. Returning a [`KademliaHandlerIn::Reset`]
-        // instead of an [`KademliaHandlerIn::PutRecordRes`] to have the remote
-        // try another node would only result in the remote node to contact an
-        // even more distant node. In addition returning
-        // [`KademliaHandlerIn::PutRecordRes`] does not reveal any internal
-        // information to a possibly malicious remote node.
-        self.queued_events.push_back(NetworkBehaviourAction::NotifyHandler {
-            peer_id: source,
-            handler: NotifyHandler::One(connection),
-            event: KademliaHandlerIn::PutRecordRes {
-                key: record.key,
-                value: record.value,
-                request_id,
-            },
-        })
     }
```

Am I missing something?

romanb: I overlooked that, apologies.

mxinden:
Recent round of suggestions applied. Thanks for the feedback. Let me know if you have any further comments.

PR #1497, opened 2020-03-16
Add `protocols_handler::multi` module.

twittner:
An implementation of `ProtocolsHandler` that contains multiple other `ProtocolsHandler`s indexed by some key type.

Factored out from https://github.com/paritytech/substrate/pull/5045.


swarm/src/protocols_handler/multi.rs
```diff
@@ -0,0 +1,260 @@
...
+        self.handlers.values()
+            .map(|h| h.connection_keep_alive())
+            .max()
+            .unwrap_or(KeepAlive::No)
+    }
+
+    fn poll(&mut self, cx: &mut Context)
+        -> Poll<ProtocolsHandlerEvent<Self::OutboundProtocol, Self::OutboundOpenInfo, Self::OutEvent, Self::Error>>
+    {
+        for (k, h) in self.handlers.iter_mut() {
```
mxinden: This might starve the last handler in case any handler other than the last returns `Poll::Ready` for each call to `poll`, right? Is that an issue or negligible given its low probability?

twittner: True, with larger numbers of handlers this might become a problem.

twittner: NB that no matter in which order we perform the polling starvation always becomes a problem with a larger number of handlers.

swarm/src/protocols_handler/multi.rs
```diff
@@ -0,0 +1,277 @@
...
+
+/// A [`ProtocolsHandler`] for multiple other `ProtocolsHandler`s.
+#[derive(Clone)]
+pub struct Handler<K, H> {
+    handlers: HashMap<K, H>
+}
+
+impl<K, H> fmt::Debug for Handler<K, H> {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        f.write_str("Handler")
```
tomaka:
```suggestion
impl<K: fmt::Debug, H: fmt::Debug> fmt::Debug for Handler<K, H> {
    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
        fmt::Debug::fmt(&self.handlers, f)
```

swarm/src/protocols_handler/multi.rs
```diff
@@ -0,0 +1,277 @@
...
+    H::InboundProtocol: InboundUpgradeSend,
+    H::OutboundProtocol: OutboundUpgradeSend
+{
+    /// Create a new empty handler.
+    pub fn new() -> Self {
+        Handler { handlers: HashMap::new() }
+    }
+
+    /// Insert a [`ProtocolsHandler`] at index `key`.
+    pub fn add(&mut self, key: K, handler: H) {
```
tomaka:
```suggestion
    pub fn insert(&mut self, key: K, handler: H) {
```

swarm/src/protocols_handler/multi.rs
```diff
@@ -0,0 +1,277 @@
...
+    }
+}
+
+impl<K, H> Handler<K, H>
+where
+    K: Clone + std::fmt::Debug + Hash + Eq + Send + 'static,
+    H: ProtocolsHandler,
+    H::InboundProtocol: InboundUpgradeSend,
+    H::OutboundProtocol: OutboundUpgradeSend
+{
```
tomaka: I don't think you need any of these here.

```suggestion
impl<K, H> Handler<K, H> {
```

swarm/src/protocols_handler/multi.rs
```diff
@@ -0,0 +1,277 @@
...
+    }
+}
+
+impl<K, H> FromIterator<(K, H)> for Handler<K, H>
+where
+    K: Clone + std::fmt::Debug + Hash + Eq + Send + 'static,
+    H: ProtocolsHandler,
+    H::InboundProtocol: InboundUpgradeSend,
+    H::OutboundProtocol: OutboundUpgradeSend
+{
```
tomaka:
```suggestion
impl<K, H> FromIterator<(K, H)> for Handler<K, H> {
```

swarm/src/protocols_handler/multi.rs
```diff
@@ -0,0 +1,277 @@
...
+
+/// Inbound and outbound upgrade for all `ProtocolsHandler`s.
+#[derive(Clone)]
+pub struct Upgrade<K, H> {
+    upgrades: HashMap<K, H>
+}
+
+impl<K, H> fmt::Debug for Upgrade<K, H> {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        f.write_str("Upgrade")
```
tomaka:
```suggestion
impl<K: fmt::Debug, H: fmt::Debug> fmt::Debug for Upgrade<K, H> {
    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
        fmt::Debug::fmt(&self.upgrades, f)
```

swarm/src/protocols_handler/multi.rs
```diff
@@ -0,0 +1,277 @@
...
+    UpgradeInfoSend
+};
+use futures::{future::BoxFuture, prelude::*};
+use libp2p_core::upgrade::ProtocolName;
+use rand::Rng;
+use std::{collections::HashMap, fmt, hash::Hash, iter::{self, FromIterator}, task::{Context, Poll}};
+
+/// A [`ProtocolsHandler`] for multiple other `ProtocolsHandler`s.
+#[derive(Clone)]
+pub struct Handler<K, H> {
```
tomaka: Maybe a name that's a bit more explicit?

```suggestion
pub struct MultiHandler<K, H> {
```

swarm/src/protocols_handler/multi.rs
```diff
@@ -0,0 +1,277 @@
...
+
+    /// Insert a [`ProtocolsHandler`] at index `key`.
+    pub fn add(&mut self, key: K, handler: H) {
+        self.handlers.insert(key, handler);
+    }
+
+    /// Remove a [`ProtocolsHandler`] at index `key`.
+    pub fn del(&mut self, key: &K) {
+        self.handlers.remove(key);
+    }
```
tomaka: I wouldn't provide any way to remove a handler.
There is no real use case for that, because it is logically incorrect to remove support for a protocol once a connection is open. And removing a handler also means that we will trigger the `log::error!`s below.

swarm/src/protocols_handler/multi.rs
```diff
@@ -0,0 +1,277 @@
...
+        self.handlers.values()
+            .map(|h| h.connection_keep_alive())
+            .max()
+            .unwrap_or(KeepAlive::No)
+    }
+
+    fn poll(&mut self, cx: &mut Context)
+        -> Poll<ProtocolsHandlerEvent<Self::OutboundProtocol, Self::OutboundOpenInfo, Self::OutEvent, Self::Error>>
+    {
+        // Not always polling handlers in the same order should give anyone the chance to make progress.
```
tomaka: I'm really not sure if it's worth the added complexity.
We turn a bug into a bug where things work a bit?

swarm/src/protocols_handler/multi.rs
```diff
@@ -0,0 +1,277 @@
...
+    where
+        T: IntoIterator<Item = (K, H)>
+    {
+        Handler { handlers: HashMap::from_iter(iter) }
+    }
+}
+
+impl<K, H> ProtocolsHandler for Handler<K, H>
+where
+    K: Clone + std::fmt::Debug + Hash + Eq + Send + 'static,
```
tomaka: I'm really not a fan of enforcing `Debug` here.

swarm/src/protocols_handler/multi.rs
tomaka: I also think it would be a very good idea to detect if we override an existing key.
However I'd be more in favour of simply not providing any `insert` method at all.

swarm/src/protocols_handler/multi.rs
twittner: I actually switched from `MultiHandler` to `Handler` because I thought that it would be used in code like this:

```rust
use libp2p::swarm::protocols_handler::multi;

type ProtocolsHandler = multi::Handler<Foo, Bar>;
```

If you prefer I can go back to `MultiHandler`.

swarm/src/protocols_handler/multi.rs
twittner: Without `Debug` we can not include the key when we log an error for a failed lookup, making the log message less useful. I also do not think `Debug` is a big burden, but if you insist I remove it.

swarm/src/protocols_handler/multi.rs
twittner: The concern was that a handler that frequently becomes ready would prevent other handlers from making progress, so I would say this is more a matter of fairness, no?

twittner:
> You also need to provide an implementation of `IntoProtocolsHandler`.

My understanding was that there is a blanket impl of `IntoProtocolsHandler` for types that implement `ProtocolsHandler`.

tomaka:
> My understanding was that there is a blanket impl of IntoProtocolsHandler for types that implement ProtocolsHandler.

All `ProtocolsHandler`s implement `IntoProtocolsHandler`, but not the invert.
Right now you can't use a "multi handler" with an `H` that implements only `IntoProtocolsHandler` and not `ProtocolsHandler`.

To rephrase: the "protocols handler" thing is really two steps (the "into", then the actual handler), and for convenience we allow short-circuiting the first step. But in that PR you make it mandatory to short-circuit the first step, which is not great for something generic.

swarm/src/protocols_handler/multi.rs
tomaka: There's no right answer to that question, but idiomatic Rust consists in importing directly types (in other words, write out `Handler` and not `multi::Handler`), which points in the direction of having something like `MultiHandler`.

swarm/src/protocols_handler/multi.rs
tomaka: I can see two situations where that happens:

- One of the inner `ProtocolsHandler`s has a bug and returns `Ready` all the time, and I don't think it's worth trying to mitigate that.
- The inner `ProtocolHandler`s work fine, but we poll too slowly (because the CPU is busy doing something else).

I guess that for situation #2 it is indeed justified.

swarm/src/protocols_handler/multi.rs
twittner: > [...] idiomatic Rust [...]

Source? A counterexample seems to be `io::Error`, no?

tomaka: It's something I read a long time ago in some official "code style guide". The best I could find is https://doc.rust-lang.org/1.0.0/style/style/imports.html#prefer-fully-importing-types/traits-while-module-qualifying-functions. but that looks old.

Yes, `io::Error` is a counter-example. The Rust standard library doesn't necessarily follow the guidelines that have been written at a later point in time.

swarm/src/protocols_handler/multi.rs
```diff
@@ -0,0 +1,362 @@
...
+                set.insert(v);
+            }
+        }
+    }
+    Ok(())
+}
+
+/// It is an error if two handlers share the same protocol name.
+#[derive(Debug, Clone)]
+pub struct DuplicateProtoname(pub Vec<u8>);
```
tomaka:
```suggestion
pub struct DuplicateProtonameErr(Vec<u8>);
```

Should be named with an `Err` or `Error` suffix.
Also its content should either be explained (I don't even know what is contained here) or be private.

swarm/src/protocols_handler/multi.rs
```diff
@@ -0,0 +1,362 @@
...
+    }
+}
+
+/// Check that no two protocol names are equal.
+fn uniq_proto_names<I, T>(iter: I) -> Result<(), DuplicateProtoname>
+where
+    I: Iterator<Item = T>,
+    T: UpgradeInfoSend
+{
+    let mut set = std::collections::HashSet::new();
```
tomaka:
```suggestion
    let mut set = HashSet::new();
```

Imports on top

swarm/src/protocols_handler/multi.rs
```diff
@@ -0,0 +1,362 @@
...
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        if let Ok(s) = std::str::from_utf8(&self.0) {
+            write!(f, "duplicate protocol name: {}", s)
+        } else {
+            write!(f, "duplicate protocol name: {:?}", self.0)
+        }
+    }
+}
+
+impl std::error::Error for DuplicateProtoname {}
```
tomaka:
```suggestion
impl error::Error for DuplicateProtoname {}
```

swarm/src/protocols_handler/multi.rs
```diff
@@ -0,0 +1,362 @@
...
+impl<K, H: ProtocolName> ProtocolName for KeyedProtoName<K, H> {
+    fn protocol_name(&self) -> &[u8] {
+        self.1.protocol_name()
+    }
+}
+
+/// Inbound and outbound upgrade for all `ProtocolsHandler`s.
+#[derive(Clone)]
+pub struct Upgrade<K, H> {
+    upgrades: HashMap<K, H>
```
tomaka: I feel like this `HashMap` here is overkill.

```suggestion
    upgrades: Vec<(K, H)>
```

Can we not index the upgrades with a simple `usize` rather than with `K`?

swarm/src/protocols_handler/multi.rs
```diff
@@ -0,0 +1,362 @@
...
+    }
+}
+
+
+impl<K, H> IntoMultiHandler<K, H>
+where
+    K: Hash + Eq,
+    H: IntoProtocolsHandler
+{
+    pub fn try_from_iter<I>(iter: I) -> Result<Self, DuplicateProtoname>
```
tomaka: Needs documentation, especially about the fact that we check for duplicates.

swarm/src/protocols_handler/multi.rs
```diff
@@ -0,0 +1,362 @@
...
+            .finish()
+    }
+}
+
+impl<K, H> MultiHandler<K, H>
+where
+    K: Hash + Eq,
+    H: ProtocolsHandler
+{
+    /// Create and populate a `MultiHandler` from the given handler iterator.
```
tomaka: Needs documentation about checking for duplicates.

swarm/src/protocols_handler/multi.rs
tomaka: To me these log messages are not expected to realistically be triggered, so it's not a big deal to make them less useful.
If we see them triggered, that means there's a bug in these 300 lines of code, and it then shouldn't be hard to find.

swarm/src/protocols_handler/multi.rs
twittner: > Should be named with an `Err` or `Error` suffix.

Why?

swarm/src/protocols_handler/multi.rs
twittner: Maybe you should write a style guide so that contributors can anticipate your requirements.

swarm/src/protocols_handler/multi.rs
tomaka: That's what every single crate in the Rust ecosystem does

swarm/src/protocols_handler/multi.rs
twittner: What would be the advantage of restricting ourselves to `usize`? The substrate PR for instance, where this originates from, uses the `ProtocolId` for indexing.

tomaka: Note that I'm talking only about `Upgrades` here. The `MultiHandler` should keep a `HashMap`.

If we compare:
With a `HashMap` in `Upgrades`:

- When a substream is opened by the remote, we go through each child handler, extract their inbound upgrade, clone the keys, and create a `HashMap`.
- The list of protocols is then gathered, and we clone the keys again.
- When a protocol is negotiated, we index the hashmap for the corresponding key.

With a `Vec` in `Upgrades`:

- When a substream is opened by the remote, we go through each child handler, extract their inbound upgrade, clone the keys, and create a `Vec`.
- The list of protocols is then gathered, and we simply put `usize`s in the upgrades. We don't clone the key.
- When a protocol is negotiated, we look in the `Vec` for corresponding index.

Overall the `Vec` solution looks overall better as we don't clone the keys, and we don't go through the overhead of a `HashMap`.

twittner: Clients which are worried about expensive `Clone`s can use a `usize` or some other "cheap" type directly.

Using `usize` only in `Upgrades` requires another `K` <-> `usize` mapping because the output of an upgrade is given to handler methods where we need to map back from `usize` to `K` in order to get back the corresponding handler from the `HashMap` of `MultiHandler`.

I am fine with using a `Vec<(K, H)>` even though I would rather like to see this presumed overhead to be quantified.

tomaka: Ah I see from your changes that we didn't think of the same thing.

To me the idea is that `protocol_info` should become:
```rust
    fn protocol_info(&self) -> Self::InfoIter {
        self.upgrades.iter().enumerate().map(|(n, (_, i))| iter::repeat(n).zip(i.protocol_info()))
            .flatten()
            .map(|(n, i)| KeyedProtoName(n, i))
            .collect::<Vec<_>>()
            .into_iter()
    }
```

And that `KeyedProtoName` should become:
```rust
pub struct KeyedProtoName<H>(usize, H);
```

Issue #1465, opened 2020-02-23
Potential unbounded buffer growth in background task of pending connection.


romanb:
It was brought up in https://github.com/libp2p/rust-libp2p/pull/1440#discussion_r378945869 that there is a theoretical unbounded growing buffer of pending events for a connection handler in a background `Task`s `events` buffer while the connection is still pending (i.e. being established). While in practice every connection attempt is eventually bound to time out, it may nevertheless be preferable to bound the buffer. My current suggestion would be to add a hard upper limit for the size of the buffer which, when reached, results in the connection attempt to be aborted and the background task dropped.

mxinden:
Implementation suggestion in https://github.com/libp2p/rust-libp2p/pull/1501.

PR #1501, opened 2020-03-19
core/src/connection: Bound task event buffer during conn establishment

mxinden:
This commit adds a size limit to the event buffer that is filled during
connection establishment and forwarded in bulk to the handler once the
connection is established.

When the size limit is reached an `EventBufferLimitReached` error is
triggered and send to the manager which will in turn drop its connection
to the task signaling it to shutdown.

The above buffer is already bound by the connection establishment
timeout. This commit adds an additional size bound to the buffer.

Fixes https://github.com/libp2p/rust-libp2p/issues/1465.

### DIFF
--- a/core/src/connection/manager/task.rs
+++ b/core/src/connection/manager/task.rs
@@ -339,3 +339,91 @@ where
     }
 }
 
+#[cfg(test)]
+mod test {
+    use super::*;
+    use crate::{
+        muxing::SingletonMuxer,
+        connection::{
+            ConnectionHandlerEvent,
+            Substream,
+            SubstreamEndpoint,
+        },
+    };
+    use futures::io::AllowStdIo;
+    use std::io::Cursor;
+
+    // `Vec<u8>` implements `std::io::Write`. `Cursor<Vec<u8>>` implements `std::io::Read`.
+    // `AllowStdIo<_>` transforms a type implementing `Read` and `Write` to a type implementing
+    // `AsyncRead` and `AsyncWrite`.
+    type TestSocket = AllowStdIo<Cursor<Vec<u8>>>;
+
+    struct TestConnectionHandler {}
+
+    impl ConnectionHandler for TestConnectionHandler {
+        type InEvent = ();
+        type OutEvent = ();
+        type Error = ();
+        type Substream = Substream<SingletonMuxer<TestSocket>>;
+        type OutboundOpenInfo = ();
+
+        fn inject_substream(
+            &mut self,
+            _: Self::Substream,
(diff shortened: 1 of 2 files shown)


core/src/connection/manager/task.rs
```diff
@@ -35,6 +35,8 @@ use futures::{prelude::*, channel::mpsc, stream};
 use std::{pin::Pin, task::Context, task::Poll};
 use super::ConnectResult;
 
+const MAX_EVENT_BUFFERING: usize = 1000;
```
mxinden: What would be a good limit? `1_000` might be a bit conservative.

romanb:
I am wondering if it is actually still necessary to buffer these events in the background task at all, since the "take over" commands sent to pending connections in the context of the single-connection-per-peer constraint have been removed. The API `poll_ready_notify_handler` and `notify_handler` is now only exposed on an `EstablishedEntry` by the `manager` module. The only other code that sends events seems to be in `poll_broadcast`, which could (and maybe even should?) be changed to skip over pending connections. Just thinking out loud here as it would naturally be preferable to further simplify things.

tomaka:
I agree with @romanb. Being able to send events to connections that haven't been established yet was probably a mistake. `poll_broadcast` as well has not proven to be useful and could be removed as well.

mxinden:
Thanks for the feedback @romanb @tomaka.

To summarize we have two alternative approaches both relieving us off the need to buffer events in the first place.

1. Have `Manager::poll_broadcast` skip `TaskState::Pending` tasks.

2. Remove `Manager::poll_broadcast`.

In both cases one would still signal a `Task` to stop by dropping the `TaskInfo::sender` `mspc::Sender`, correct?

Do you have preferences on (1) or (2)?

tomaka:
> In both cases one would still signal a Task to stop by dropping the TaskInfo::sender mspc::Sender, correct?

Yes.

> Do you have preferences on (1) or (2)?

I'd just go with 2.

romanb:
If `poll_broadcast` is not needed it is certainly preferable to remove it, especially since it already necessarily changed semantics with #1440.

> In both cases one would still signal a Task to stop by dropping the TaskInfo::sender mspc::Sender, correct?

I'm not sure what you mean by that w.r.t. this PR? Do you mean what a `Task` should do if it receives an event for the connection handler while it is still pending? I suppose it would just panic, referring to the requirement that the `Manager` must not permit sending such events in that state.

mxinden:
Replaced by https://github.com/libp2p/rust-libp2p/pull/1527. Thanks for the input!

PR #1502, opened 2020-03-19
Allow customizing the Kademlia maximum packet size

tomaka:


protocols/kad/src/behaviour.rs
```diff
@@ -115,6 +119,7 @@ impl Default for KademliaConfig {
             provider_publication_interval: Some(Duration::from_secs(12 * 60 * 60)),
             provider_record_ttl: Some(Duration::from_secs(24 * 60 * 60)),
             connection_idle_timeout: Duration::from_secs(10),
+            max_protocol_packet_size: 4096,
```
romanb: Would it be an idea to replace `protocol_name_override` and `max_protocol_packet_size` with directly embedding a `KademliaProtocolConfig` in the `KademliaConfig` (e.g. like it is done with `QueryConfig`)? I may simplify things a bit, now and for the future if more options are added. For example, the default packet size would not be duplicated but only exist in the `Default` impl for `KademliaProtocolConfig`. One step further would be to even embed the `KademliaHandlerConfig` in the `KademliaConfig`. These embeddings are implementation details, of course, the `KademliaConfig` has the public API, but it keeps the number of potentially duplicated fields to a minimum, e.g. the embedded configs can be moved into fields of the behaviour on construction as a whole.

PR #1506, opened 2020-03-23
Permit concurrent dialing attempts per peer.

romanb:
This is a follow-up to https://github.com/libp2p/rust-libp2p/pull/1440 and relates to https://github.com/libp2p/rust-libp2p/issues/925. This PR permits multiple dialing attempts per peer.  The essence of the changes are that the `Peer` API in `libp2p-core` now provides `Peer::dial()`, i.e. regardless of the state in which the peer is. Previously dialing was only permitted on a `ConnectedPeer` or `DisconnectedPeer`. A dialing attempt is always made up of one or more addresses tried sequentially, as before, but now there can be multiple dialing attempts per peer. A configurable per-peer limit for outgoing connections and thus concurrent dialing attempts is also included. A `NetworkBehaviour` can now always request a new dialing attempt via `DialPeerCondition::Always` in `NetworkBehaviourAction::DialPeer`.

Another change included here is that `Swarm::dial` now returns `Result<(), DialError>` instead of `Result<bool, ConnectionLimit>`, i.e. the case of dialing failing because `NetworkBehaviour::addresses_of_peer` returning no addresses is now treated as an error (`DialError::NoAddresses`). This has the (I think) desirable effect of having a `NetworkBehaviourAction::DialPeer` request always be paired with a final call to `inject_connection_established` or `inject_dial_failure`. A behaviour may otherwise potentially keep state around for a dialing attempt that it thinks it emitted but for which it never got informed about a result.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Version ???
 
+- `libp2p-core`, `libp2p-swarm`: Added support for multiple dialing
+  attempts per peer, with a configurable limit.
+  [PR 1506](https://github.com/libp2p/rust-libp2p/pull/1506)
+
 - `libp2p-noise`: Added the `X25519Spec` protocol suite which uses
   libp2p-noise-spec compliant signatures on static keys as well as the
   `/noise` protocol upgrade, hence providing a libp2p-noise-spec compliant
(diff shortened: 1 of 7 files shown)


romanb:
I integrated the latest changes from master.

romanb:
Thanks for the review. I still need to update the changelog, but otherwise if there are no further objections / comments by next Monday, I plan to merge this then.

PR #1509, opened 2020-03-24
Fix potential bug when polling the Swarm

tomaka:
In the situation where `this.network.poll` returns `Poll::Ready(_)` and then `this.behaviour.poll` returns `Poll::Ready(NetworkBehaviourAction::NotifyHandler { .. })`, we would return `Poll::Pending`.

That is wrong, as nothing in the stack has registered the `Waker` to wake it up later. In other words, the task would then get frozen forever.

Same problem if `this.network.poll` returns `Ready` and we have an event buffered up that we fail to deliver.

### DIFF
--- a/swarm/src/lib.rs
+++ b/swarm/src/lib.rs
@@ -487,7 +487,7 @@ where TBehaviour: NetworkBehaviour<ProtocolsHandler = THandler>,
             // the pending event emitted by the behaviour in the previous iteration
             // to the connection handler(s). The pending event must be delivered
             // before polling the behaviour again. If the targeted peer
-            // meanwhie disconnected, the event is discarded.
+            // meanwhile disconnected, the event is discarded.
             if let Some((peer_id, handler, event)) = this.pending_event.take() {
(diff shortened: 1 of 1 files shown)


romanb:
> That is wrong, as nothing in the stack has registered the Waker to wake it up later. In other words, the task would then get frozen forever.

I'm not sure I follow, I will need to take a closer look later, but all the occurrences of `return Poll::Pending` that you removed here are in the context of `notify_one` / `notify_any` / `notify_all` having returned `Some`, which only happens when the current task is registered to be woken up because at least one connection returned `Poll::Pending` from `poll_ready_notify_handler`. At least that is the intention and how I documented it.

tomaka:
Oh ok, I see. I admit that it didn't occur to me that `notify` was polling, as I still had the former design with the unbounded channels in my head.

romanb:
No problem, I actually found an unrelated issue on re-reviewing the code: https://github.com/libp2p/rust-libp2p/pull/1513.

PR #1510, opened 2020-03-24
Multiaddr: Replace Arc<Vec<u8>> with inline storage

rklaehn:
Storage stores small bytes inline, and spills to the heap only once that is no longer possible.

A typical small multiaddr consisting of e.g. ip4 host/port, ip6 host/port etc. will be stored inline. Only some complex multiaddrs containing multiple peer ids will be stored on the heap as an Arc<[u8]>.

The `push` and `with` fns allocate temporary vectors, but I think this is an acceptable tradeoff for making the in memory representation of multiaddrs much smaller. Many of the multiadds that you have you don't ever manually build but get from somewhere else.

### DIFF
--- a/misc/multiaddr/src/lib.rs
+++ b/misc/multiaddr/src/lib.rs
@@ -6,6 +6,7 @@ mod protocol;
 mod onion_addr;
 mod errors;
 mod from_url;
+mod storage;
 
 use serde::{
     Deserialize,
@@ -18,12 +19,13 @@ use std::{
     convert::TryFrom,
     fmt,
     io,
+    hash,
     iter::FromIterator,
     net::{IpAddr, Ipv4Addr, Ipv6Addr},
     result::Result as StdResult,
     str::FromStr,
-    sync::Arc
 };
+use storage::Storage;
 pub use self::errors::{Result, Error};
 pub use self::from_url::{FromUrlErr, from_url, from_url_lossy};
 pub use self::protocol::Protocol;
@@ -36,28 +38,23 @@ static_assertions::const_assert! {
 }
 
 /// Representation of a Multiaddr.
-#[derive(PartialEq, Eq, Clone, Hash)]
-pub struct Multiaddr { bytes: Arc<Vec<u8>> }
+#[derive(Clone)]
+pub struct Multiaddr { storage: Storage }
 
 impl Multiaddr {
     /// Create a new, empty multiaddress.
     pub fn empty() -> Self {
-        Self { bytes: Arc::new(Vec::new()) }
-    }
-
-    /// Create a new, empty multiaddress with the given capacity.
-    pub fn with_capacity(n: usize) -> Self {
-        Self { bytes: Arc::new(Vec::with_capacity(n)) }
+        Self { storage: Storage::from_slice(&[]) }
     }
 
     /// Return the length in bytes of this multiaddress.
     pub fn len(&self) -> usize {
-        self.bytes.len()
+        self.as_ref().len()
     }
 
     /// Return a copy of this [`Multiaddr`]'s byte representation.
     pub fn to_vec(&self) -> Vec<u8> {
-        Vec::from(&self.bytes[..])
+        self.as_ref().to_vec()
     }
 
     /// Adds an already-parsed address component to the end of this multiaddr.
@@ -73,9 +70,10 @@ impl Multiaddr {
     /// ```
     ///
     pub fn push(&mut self, p: Protocol<'_>) {
-        let mut w = io::Cursor::<&mut Vec<u8>>::new(Arc::make_mut(&mut self.bytes));
+        let mut w = io::Cursor::new(self.to_vec());
         w.set_position(w.get_ref().len() as u64);
-        p.write_bytes(&mut w).expect("Writing to a `io::Cursor<&mut Vec<u8>>` never fails.")
+        p.write_bytes(&mut w).expect("Writing to a `io::Cursor<&mut Vec<u8>>` never fails.");
+        self.storage = Storage::from_slice(&w.into_inner());
     }
 
     /// Pops the last `Protocol` of this multiaddr, or `None` if the multiaddr is empty.
@@ -89,7 +87,7 @@ impl Multiaddr {
     /// ```
     ///
     pub fn pop<'a>(&mut self) -> Option<Protocol<'a>> {
-        let mut slice = &self.bytes[..]; // the remaining multiaddr slice
+        let mut slice = self.as_ref(); // the remaining multiaddr slice
         if slice.is_empty() {
             return None
         }
@@ -100,16 +98,17 @@ impl Multiaddr {
             }
             slice = s
         };
-        let remaining_len = self.bytes.len() - slice.len();
-        Arc::make_mut(&mut self.bytes).truncate(remaining_len);
+        let remaining_len = self.as_ref().len() - slice.len();
+        self.storage = Storage::from_slice(&self.as_ref()[..remaining_len]);
         Some(protocol)
     }
 
     /// Like [`Multiaddr::push`] but consumes `self`.
     pub fn with(mut self, p: Protocol<'_>) -> Self {
-        let mut w = io::Cursor::<&mut Vec<u8>>::new(Arc::make_mut(&mut self.bytes));
+        let mut w = io::Cursor::new(self.to_vec());
         w.set_position(w.get_ref().len() as u64);
         p.write_bytes(&mut w).expect("Writing to a `io::Cursor<&mut Vec<u8>>` never fails.");
+        self.storage = Storage::from_slice(&w.into_inner());
         self
     }
 
@@ -130,7 +129,7 @@ impl Multiaddr {
     /// ```
     ///
     pub fn iter(&self) -> Iter<'_> {
-        Iter(&self.bytes)
+        Iter(&self.as_ref())
     }
 
     /// Replace a [`Protocol`] at some position in this `Multiaddr`.
@@ -145,7 +144,7 @@ impl Multiaddr {
     where
         F: FnOnce(&Protocol) -> Option<Protocol<'a>>
     {
-        let mut address = Multiaddr::with_capacity(self.len());
+        let mut address = Multiaddr::empty();
         let mut fun = Some(by);
         let mut replaced = false;
 
@@ -192,9 +191,23 @@ impl fmt::Display for Multiaddr {
     }
 }
 
+impl PartialEq for Multiaddr {
+    fn eq(&self, other: &Self) -> bool {
+        self.as_ref() == other.as_ref()
+    }
+}
+
+impl Eq for Multiaddr {}
+
+impl hash::Hash for Multiaddr {
+    fn hash<H: hash::Hasher>(&self, state: &mut H) {
+        self.as_ref().hash(state);
+    }
+}
+
 impl AsRef<[u8]> for Multiaddr {
     fn as_ref(&self) -> &[u8] {
-        self.bytes.as_ref()
+        self.storage.bytes()
     }
 }
 
@@ -203,7 +216,7 @@ impl<'a> IntoIterator for &'a Multiaddr {
     type IntoIter = Iter<'a>;
 
     fn into_iter(self) -> Iter<'a> {
-        Iter(&self.bytes)
+        Iter(&self.as_ref())
     }
 }
 
@@ -216,7 +229,7 @@ impl<'a> FromIterator<Protocol<'a>> for Multiaddr {
         for cmp in iter {
             cmp.write_bytes(&mut writer).expect("Writing to a `Vec` never fails.");
         }
-        Multiaddr { bytes: Arc::new(writer) }
+        Multiaddr { storage: Storage::from_slice(&writer) }
     }
 }
 
@@ -237,7 +250,7 @@ impl FromStr for Multiaddr {
             p.write_bytes(&mut writer).expect("Writing to a `Vec` never fails.");
         }
 
-        Ok(Multiaddr { bytes: Arc::new(writer) })
+        Ok(Multiaddr { storage: Storage::from_slice(&writer) })
     }
 }
 
@@ -264,7 +277,7 @@ impl<'a> From<Protocol<'a>> for Multiaddr {
     fn from(p: Protocol<'a>) -> Multiaddr {
         let mut w = Vec::new();
         p.write_bytes(&mut w).expect("Writing to a `Vec` never fails.");
-        Multiaddr { bytes: Arc::new(w) }
+        Multiaddr { storage: Storage::from_slice(&w) }
     }
 }
 
@@ -299,7 +312,7 @@ impl TryFrom<Vec<u8>> for Multiaddr {
             let (_, s) = Protocol::from_bytes(slice)?;
             slice = s
         }
-        Ok(Multiaddr { bytes: Arc::new(v) })
+        Ok(Multiaddr { storage: Storage::from_slice(&v) })
     }
 }
 
--- a/misc/multiaddr/src/lib.rs
+++ b/misc/multiaddr/src/lib.rs
@@ -70,10 +70,9 @@ impl Multiaddr {
     /// ```
     ///
     pub fn push(&mut self, p: Protocol<'_>) {
-        let mut w = io::Cursor::new(self.to_vec());
-        w.set_position(w.get_ref().len() as u64);
-        p.write_bytes(&mut w).expect("Writing to a `io::Cursor<&mut Vec<u8>>` never fails.");
-        self.storage = Storage::from_slice(&w.into_inner());
+        let mut w = Buffer::from_ref(self.as_ref());
+        p.write_bytes(&mut w).expect("Writing to a `Buffer` never fails.");
+        self.storage = w.to_storage();
     }
 
     /// Pops the last `Protocol` of this multiaddr, or `None` if the multiaddr is empty.
@@ -105,10 +104,9 @@ impl Multiaddr {
 
     /// Like [`Multiaddr::push`] but consumes `self`.
     pub fn with(mut self, p: Protocol<'_>) -> Self {
-        let mut w = io::Cursor::new(self.to_vec());
-        w.set_position(w.get_ref().len() as u64);
-        p.write_bytes(&mut w).expect("Writing to a `io::Cursor<&mut Vec<u8>>` never fails.");
-        self.storage = Storage::from_slice(&w.into_inner());
+        let mut w = Buffer::from_ref(self.as_ref());
+        p.write_bytes(&mut w).expect("Writing to a `Buffer` never fails.");
+        self.storage = w.to_storage();
         self
     }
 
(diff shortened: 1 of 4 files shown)


misc/multiaddr/src/storage.rs
```diff
@@ -0,0 +1,71 @@
+use std::sync::Arc;
+
+/// MAX_INLINE is the maximum size of a multiaddr that can be stored inline
+const MAX_INLINE: usize = 38;
```
rklaehn: This means that the entire struct will be 40 bytes large (+1 byte enum discriminator and +1 byte length). I guess you could argue that for the typical multiaddr that does not contain a peer id, 32 bytes total length / 30 bytes payload length would be enough...

rklaehn:
Note: this is basically identical to the optimization that has been done in rust-multihash in https://github.com/multiformats/rust-multihash/pull/47

misc/multiaddr/src/storage.rs
```diff
@@ -0,0 +1,71 @@
...
+
+#[derive(Clone)]
+pub(crate) enum Storage {
+    /// hash is stored inline if it is smaller than MAX_INLINE
+    Inline(u8, [u8; MAX_INLINE]),
+    /// hash is stored on the heap. this must be only used if the hash is actually larger than
+    /// MAX_INLINE bytes to ensure an unique representation.
```
mxinden:
```suggestion
    /// MAX_INLINE bytes to ensure a unique representation.
```

misc/multiaddr/src/storage.rs
mxinden: I would say having these assumptions documented similar to as  comments similar to https://github.com/multiformats/rust-multihash/blob/master/src/storage.rs#L5 would be useful for the future.

rklaehn: Will push a documentation update!

rklaehn:
@mxinden I can add a clone benchmark that compares cloning multiaddrs that are slightly below the limit to multiadds that are slightly above the limit. But it will be hard to integrate this into CI, so it would just be something you run on demand when you want to mess with this part of the code...

rklaehn:
Added a small bench and a test that shows where the limits of inline storage are.

misc/multiaddr/src/storage.rs
```diff
@@ -0,0 +1,109 @@
...
+// DEALINGS IN THE SOFTWARE.
+use std::sync::Arc;
+
+/// MAX_INLINE is the maximum size of a multiaddr that can be stored inline.
+/// There is an overhead of 2 bytes, 1 for the length and 1 for the enum discriminator.
+/// 30 is chosen so that the overall size is 32. This should still be big enough to fit
+/// a multiaddr containing an ipv4 or ipv6 address and port.
+///
+/// More complex multiaddrs like those containing peer ids will be stored on the heap.
+const MAX_INLINE: usize = 30;
```
twittner: Why not `62`? Most peer IDs would fit in there as well and peer IDs are quite a common case.

misc/multiaddr/src/lib.rs
```diff
@@ -73,9 +70,10 @@ impl Multiaddr {
     /// ```
     ///
     pub fn push(&mut self, p: Protocol<'_>) {
-        let mut w = io::Cursor::<&mut Vec<u8>>::new(Arc::make_mut(&mut self.bytes));
+        let mut w = io::Cursor::new(self.to_vec());
```
twittner: Can we not avoid this allocation if `p` fits into `Storage::Inline`?

misc/multiaddr/src/lib.rs
```diff
@@ -100,16 +98,17 @@ impl Multiaddr {
...
-        Arc::make_mut(&mut self.bytes).truncate(remaining_len);
+        let remaining_len = self.as_ref().len() - slice.len();
+        self.storage = Storage::from_slice(&self.as_ref()[..remaining_len]);
         Some(protocol)
     }
 
     /// Like [`Multiaddr::push`] but consumes `self`.
     pub fn with(mut self, p: Protocol<'_>) -> Self {
-        let mut w = io::Cursor::<&mut Vec<u8>>::new(Arc::make_mut(&mut self.bytes));
+        let mut w = io::Cursor::new(self.to_vec());
```
twittner: Can we not avoid this allocation if `p` fits into `Storage::Inline`?

misc/multiaddr/src/lib.rs
rklaehn: I guess so, but it would be a bit more tricky, and I was going for a minimal code change.

If you write a thing that behaves like a vector and gracefully switches from inline to heap storage, at some point you are basically reimplementing https://crates.io/crates/smallvec . You could just have a dependency on smallvec, then this would be super easy to write efficiently. But I have chosen to go with fewer dependencies.

I doubt that it matters, since typically you build or deconstruct a multiaddr once, but then keep it around for a long time. There are lots of multiaddrs that you *never* build but just get over the wire somehow.

misc/multiaddr/src/storage.rs
rklaehn: We could make it 62. I have no idea about the size distribution of the typical peer id. From what I have seen in rust-libp2p, and for what I intend to use this for, you often retain addresses for other nodes, which are just host/port/protocol. If you say that the ones containing a peer id are sufficiently common, I will change this to 62.

misc/multiaddr/src/lib.rs
rklaehn: Same as above.

misc/multiaddr/src/lib.rs
twittner: > You could just have a dependency on smallvec, then this would be super easy to write efficiently. But I have chosen to go with fewer dependencies.

I would go with `smallvec` which is already used in quite a few places within `rust-libp2p`.

rklaehn: Using smallvec now.

twittner:
> Use smallvec to avoid some allocations in push and with

Sorry, what I had in mind was to use `SmallVec` everywhere instead of `Storage`. From my perspective all that is needed compared to master is this change:

```diff
diff --git a/misc/multiaddr/Cargo.toml b/misc/multiaddr/Cargo.toml
index c30b4504..a6ce960d 100644
--- a/misc/multiaddr/Cargo.toml
+++ b/misc/multiaddr/Cargo.toml
@@ -16,6 +16,7 @@ data-encoding = "2.1"
 multihash = "0.10"
 percent-encoding = "2.1.0"
 serde = "1.0.70"
+smallvec = { version = "1.0", features = ["write"] }
 static_assertions = "1.1"
 unsigned-varint = "0.3"
 url = { version = "2.1.0", default-features = false }
diff --git a/misc/multiaddr/src/lib.rs b/misc/multiaddr/src/lib.rs
index f81b8ada..58906a48 100644
--- a/misc/multiaddr/src/lib.rs
+++ b/misc/multiaddr/src/lib.rs
@@ -14,10 +14,10 @@ use serde::{
     Serializer,
     de::{self, Error as DeserializerError}
 };
+use smallvec::SmallVec;
 use std::{
     convert::TryFrom,
     fmt,
-    io,
     iter::FromIterator,
     net::{IpAddr, Ipv4Addr, Ipv6Addr},
     result::Result as StdResult,
@@ -35,19 +35,22 @@ static_assertions::const_assert! {
     std::mem::size_of::<usize>() <= std::mem::size_of::<u64>()
 }
 
+/// Size to use with `SmallVec`.
+const INLINE_SIZE: usize = 32;
+
 /// Representation of a Multiaddr.
 #[derive(PartialEq, Eq, Clone, Hash)]
-pub struct Multiaddr { bytes: Arc<Vec<u8>> }
+pub struct Multiaddr { bytes: Arc<SmallVec<[u8; INLINE_SIZE]>> }
 
 impl Multiaddr {
     /// Create a new, empty multiaddress.
     pub fn empty() -> Self {
-        Self { bytes: Arc::new(Vec::new()) }
+        Self { bytes: Arc::new(SmallVec::new()) }
     }
 
     /// Create a new, empty multiaddress with the given capacity.
     pub fn with_capacity(n: usize) -> Self {
-        Self { bytes: Arc::new(Vec::with_capacity(n)) }
+        Self { bytes: Arc::new(SmallVec::with_capacity(n)) }
     }
 
     /// Return the length in bytes of this multiaddress.
@@ -73,9 +76,8 @@ impl Multiaddr {
     /// ```
     ///
     pub fn push(&mut self, p: Protocol<'_>) {
-        let mut w = io::Cursor::<&mut Vec<u8>>::new(Arc::make_mut(&mut self.bytes));
-        w.set_position(w.get_ref().len() as u64);
-        p.write_bytes(&mut w).expect("Writing to a `io::Cursor<&mut Vec<u8>>` never fails.")
+        let mut w = Arc::make_mut(&mut self.bytes);
+        p.write_bytes(&mut w).expect("Writing to a `SmallVec` never fails.")
     }
 
     /// Pops the last `Protocol` of this multiaddr, or `None` if the multiaddr is empty.
@@ -107,9 +109,8 @@ impl Multiaddr {
 
     /// Like [`Multiaddr::push`] but consumes `self`.
     pub fn with(mut self, p: Protocol<'_>) -> Self {
-        let mut w = io::Cursor::<&mut Vec<u8>>::new(Arc::make_mut(&mut self.bytes));
-        w.set_position(w.get_ref().len() as u64);
-        p.write_bytes(&mut w).expect("Writing to a `io::Cursor<&mut Vec<u8>>` never fails.");
+        let mut w = Arc::make_mut(&mut self.bytes);
+        p.write_bytes(&mut w).expect("Writing to a `SmallVec` never fails.");
         self
     }
 
@@ -212,7 +213,7 @@ impl<'a> FromIterator<Protocol<'a>> for Multiaddr {
     where
         T: IntoIterator<Item = Protocol<'a>>,
     {
-        let mut writer = Vec::new();
+        let mut writer = SmallVec::new();
         for cmp in iter {
             cmp.write_bytes(&mut writer).expect("Writing to a `Vec` never fails.");
         }
@@ -224,7 +225,7 @@ impl FromStr for Multiaddr {
     type Err = Error;
 
     fn from_str(input: &str) -> Result<Self> {
-        let mut writer = Vec::new();
+        let mut writer = SmallVec::new();
         let mut parts = input.split('/').peekable();
 
         if Some("") != parts.next() {
@@ -262,7 +263,7 @@ impl<'a> Iterator for Iter<'a> {
 
 impl<'a> From<Protocol<'a>> for Multiaddr {
     fn from(p: Protocol<'a>) -> Multiaddr {
-        let mut w = Vec::new();
+        let mut w = SmallVec::new();
         p.write_bytes(&mut w).expect("Writing to a `Vec` never fails.");
         Multiaddr { bytes: Arc::new(w) }
     }
@@ -299,7 +300,7 @@ impl TryFrom<Vec<u8>> for Multiaddr {
             let (_, s) = Protocol::from_bytes(slice)?;
             slice = s
         }
-        Ok(Multiaddr { bytes: Arc::new(v) })
+        Ok(Multiaddr { bytes: Arc::new(SmallVec::from(v)) })
     }
 }
 ```

**Comment by twittner** [comment]
Ah sorry, that makes no sense. We still do the Arc.

File: misc/multiaddr/src/lib.rs [hunk 6]
```diff
@@ -412,3 +423,26 @@ macro_rules! multiaddr {
         }
     }
 }
+
+struct Buffer(smallvec::SmallVec<[u8; 32]>);
```
twittner: `smallvec` has a `write` feature that would allow you to remove `Buffer`.

misc/multiaddr/src/storage.rs
```diff
@@ -0,0 +1,109 @@
...
+impl Storage {
+    /// The raw bytes.
+    pub fn bytes(&self) -> &[u8] {
+        match self {
+            Storage::Inline(len, bytes) => &bytes[..(*len as usize)],
+            Storage::Heap(data) => &data,
+        }
+    }
+
+    /// creates storage from a vec. For a size up to MAX_INLINE, this will not allocate.
```
twittner:
```suggestion
    /// Creates storage from a slice.
    /// For a size up to MAX_INLINE, this will not allocate.
```

misc/multiaddr/src/storage.rs
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    #[test]
+    fn struct_size() {
+        // this should be true for both 32 and 64 bit archs
+        assert_eq!(std::mem::size_of::<Storage>(), 32);
+    }
+
+    #[test]
+    fn roundtrip() {
```
twittner: Is this test not redundant given the property test below?

rklaehn: Well, this one is guaranteed to hit all the interesting cases, whereas with the property based test you depend on luck. But this was copied from multihash, which is an independent repository. So if you want to reduce the number of tests for rust-libp2p, which is a bit bigger, I can remove it.

misc/multiaddr/src/lib.rs
rklaehn: :+1:

misc/multiaddr/src/storage.rs
rklaehn: Unless large multiaddrs are way more popular than I think, 64 would be too big. If you go to the arced version for large multihashes, it is not the end of the world. And there are situations where having lots of 64 byte objects would be worse than having a bunch of pointers to arcs.

misc/multiaddr/src/storage.rs
rklaehn: If you are concerned about the test runtime, it would probably be best to keep this one and drop the property based test.

tomaka:
This comment is kind of irrelevant here since we're also using `Bytes`, but I'm really not a fan of this kind of magic and tend to prefer explicitness.

To me, instead of optimizing the insides of the `Multiaddr` we should either avoid cloning multiaddresses or use a `Arc<Multiaddr>`.

I also strongly think that we're micro-optimizing here. Benchmarking how much time it takes to clone a `Multiaddr` is based upon the assumption that we're spending a lot of time cloning multiaddresses, while my guess is that this is probably using not even 0.1% of the CPU time of a typical libp2p application.

That being said I don't care that much and I'm not opposed to merging this.

rklaehn:
Just to be clear, the purpose of this change is not to make clone faster. That is just a welcome but not very important side effect. I am perfectly happy with the speed of cloning an `Arc`.

The purpose of the optimization is to reduce the memory usage of a multiaddr, make memory usage more deterministic, and to avoid touching the heap at all in the average case. (The worst case memory usage of an Arc<Vec<u8>> can be arbitrarily large and even in the average case will be significantly more than just the raw bytes, because it is **two** objects on the heap.

I agree that this is an optimization that does not matter at all unless you are running on very low powered hardware and/or have a very large number of multiaddrs sticking around. But if multiaddr is supposed to be a general purpose crate and not just a support library for parity, then these situations can happen. E.g. some of the ipfs nodes used by pinning services like ipfs pinata are keeping a quite large number of connections open to work around deficiencies of the current content routing.

Besides, I think as far as microoptimizations go, this one is pretty harmless and self contained. It does not add unsafe code or additional dependencies (except for the SmallVec, which I could also remove again). The optimization was originally done for multi *hash*, for which it definitely matters because an ipfs node will keep a very large number of multihashes around. I thought it would be nice to carry it over to multiaddr, where admittedly it matters a bit less.

tomaka:
> But if multiaddr is supposed to be a general purpose crate and not just a support library for parity, then these situations can happen. E.g. some of the ipfs nodes used by pinning services like ipfs pinata are keeping a quite large number of connections open to work around deficiencies of the current content routing.

If anything, this change ties the multiaddr *more* to a specific usage.
If you are storing a lot `Multiaddr`s that are larger than 30 bytes, then this PR is strictly negative, as every single instance of `Multiaddr` is now at least 30 bytes in size.

> some of the ipfs nodes used by pinning services like ipfs pinata are keeping a quite large number of connections open to work around deficiencies of the current content routing.
> for which it definitely matters because an ipfs node will keep a very large number of multihashes around

I don't understand. If you're saying that you are keeping a lot of multihashes alive, then this PR will worsen your memory usage, and cloning an inline multiaddr/multihash is presumably more expensive than cloning an `Arc`.

rklaehn:
> > But if multiaddr is supposed to be a general purpose crate and not just a support library for parity, then these situations can happen. E.g. some of the ipfs nodes used by pinning services like ipfs pinata are keeping a quite large number of connections open to work around deficiencies of the current content routing.
> 
> If anything, this change ties the multiaddr _more_ to a specific usage.
> If you are storing a lot `Multiaddr`s that are larger than 30 bytes, then this PR is strictly negative, as every single instance of `Multiaddr` is now at least 30 bytes in size.
> Anything different than `Multiaddr { _: Vec<u8> }` is opinionated over how it is being used.
> 
If you have large multiaddrs and do not benefit from the inline case, a Multiaddr that is essentially an `Arc<[u8]>` is still more memory efficient than a multiaddr that is a `Vec<u8>`. The heap allocated byte array is often larger than the actual storage that is required. E.g. you have a multiaddr that is 34 bytes, the buffer backing the vec will probably be 64 bytes due to the vector being grown in increments of 2. If your multiaddr is a result of some push and pop, the buffer backing the vec might even be much larger.

> > some of the ipfs nodes used by pinning services like ipfs pinata are keeping a quite large number of connections open to work around deficiencies of the current content routing.
> > for which it definitely matters because an ipfs node will keep a very large number of multihashes around
> 
> I don't understand. If you're saying that you are keeping a lot of multihashes alive, then this PR will worsen your memory usage,

If you have a lot of distinct multiaddrs that fit the inline case, this will make memory usage better since you don't touch the heap at all. If you have a lot of distinct multiaddrs that _do not_ fit the inline case, it will still make things better over an `Arc<Vec<u8>>`, which is the current state, due to 1 instead of 2 objects on the heap.

> and cloning an inline multiaddr/multihash is presumably more expensive than cloning an `Arc`.

Unless you go to very large inline multiaddrs, it is significantly cheaper than cloning an `Arc`. See attached benchmark which compares the inline case (copying 32 bytes) to the non-inline case (increasing the strong count of the arc).

But this whole thing is not worth having a heated discussion about. I think this is a small but in some cases significant improvement. If you think this is a bad tradeoff and don't want to merge this, I am fine with that as well.

rklaehn:
Just out of curiosity, I ran a small example that just allocates a bunch of multiaddrs, on this branch and on master, for the inline and non-inline case, and ran heaptrack on it.

```rust
use multiaddr::Multiaddr;
fn main() {
    let inline = "/dns4/01234567890123456789123/tcp/80/ws";
    let heap = "/dns4/0123456789012345678901234/tcp/80/ws";
    let mut t: Vec<Multiaddr> = Vec::new();
    for _ in 0..1000000 {
        // t.push(heap.parse().unwrap());
        t.push(inline.parse().unwrap());
    }
}
```

This branch / inline
```
peak heap memory consumption:
33,6 MB after 1.737s
```

This branch / heap
```
peak heap memory consumption:
81,6 MB after 3.423s
```

master / same size as inline case above
```
peak heap memory consumption:
98,5 MB after 3.469s
```

master / same size as heap case above
```
peak heap memory consumption:
102,5 MB after 3.352s
```

So in the inline case, this branch consumes quite a bit less memory than master. But also in the heap case, this branch consumes a bit less memory than master, presumably because `Arc<[u8]>` is just a single object on the heap, vs. 2 for `Arc<Vec<u8>>`.

Not sure if there is a better way to measure the actual heap consumption including overhead. Was looking for a working version of https://docs.rs/heapsize/0.4.2/heapsize/ that works with current rust, but could not find anything quickly.

rklaehn:
@tomaka I am going to close this. I think doing this in a way that is more agreeable to you is beyond my current rust capabilities.

Ideally you would want to separate the notion of a multiaddr (which is just a refinement of `[u8]`) and the concern of storing things on the stack if they are too small for the heap completely, correct?

That would be a bit more work, which might not be worth it...

Issue #1511, opened 2020-03-24
OneShotHandler: timeout ignored


sireliah:
I accidentally discovered that specifying timeout on the `OneShotHandler` is not working properly: the timeout defaults to 10 seconds for all new substreams even if you set different value on your handler.

It's a bit troublesome to reproduce, but I'll try.

1. Implement your new `NetworkBehaviour` and set in timeouts in `new_handler` in a network behaviour that uses `OneShotHandler`.
```rust
    fn new_handler(&mut self) -> Self::ProtocolsHandler {
        let timeout = Duration::from_secs(60);
        let message = SomeMessage::new();
        let proto = SubstreamProtocol::new(message).with_timeout(timeout);
        Self::ProtocolsHandler::new(proto, timeout)
    }
```
as in:
 https://github.com/libp2p/rust-libp2p/blob/master/protocols/floodsub/src/layer.rs#L231
2. Send large payload using the protocol (something guaranteed to take more than 10 seconds)
3. `inject_dial_upgrade_error` method in the handler will receive `Timeout` after precisely 10 seconds.

Root cause is in the poll, which re-instantiates `SubstreamProtocol` with default 10 seconds timeout.
https://github.com/libp2p/rust-libp2p/blob/master/swarm/src/protocols_handler/one_shot.rs#L227

Please let me know if this makes sense!

PR #1512, opened 2020-03-24
Fix timeout on OneShotHandler

sireliah:
Fixes: https://github.com/libp2p/rust-libp2p/issues/1511

`OneShotHandler` opened new `SubstreamProtocol`s with default timeout, however the handler could have been configured with different timeout.

### DIFF
--- a/swarm/src/protocols_handler/one_shot.rs
+++ b/swarm/src/protocols_handler/one_shot.rs
@@ -224,7 +224,8 @@ where
                 self.dial_negotiated += 1;
                 return Poll::Ready(
                     ProtocolsHandlerEvent::OutboundSubstreamRequest {
-                        protocol: SubstreamProtocol::new(self.dial_queue.remove(0)),
+                        protocol: SubstreamProtocol::new(self.dial_queue.remove(0))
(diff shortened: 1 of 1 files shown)


romanb:
The `inactive_timeout` is actually not intended to be a timeout for outbound substreams, onle for `connection_keep_alive`. In other words, I don't think this is a bug, but rather a feature request for making the timeout on outbound substream requests configurable for the `OneShotHandler`, which should probably be done with a new configurable setting.

sireliah:
Ok, that makes sense. Do you think we should have this kind of feature? My use case involved outbound substream that needs very long timeout (for the reason of big payload), but I wonder if anyone would use `OneShotHandler` similarly.

tomaka:
> Do you think we should have this kind of feature? 

To me it's never a bad idea to make things more configurable.
(with some exceptions, for example you don't want for a weird configuration value to be capable of breaking the logic of the code)

sireliah:
Great, then I'll prepare PR for that!

sireliah:
Here we go: https://github.com/libp2p/rust-libp2p/pull/1521

PR #1513, opened 2020-03-24
[libp2p-swarm] Correct returned connections from notify_all.

romanb:
This addresses a potential issue with `notify_all` in the swarm discovered on another review of the code triggered by https://github.com/libp2p/rust-libp2p/pull/1509: If at least one connection was not ready (i.e. pending), only those (i.e. the pending) connections would be returned and considered on the next task wakeup, whereas those which were ready should also remain in the list of connections to notify on task wakeup and retry of `notify_all`. I admit that this deserves better testing, as does `libp2p-swarm` in general, which I hope to spend some more time on soon.

### DIFF
--- a/swarm/src/lib.rs
+++ b/swarm/src/lib.rs
@@ -681,9 +681,9 @@ where
 /// Notify all of the given connections of a peer of an event.
 ///
 /// Returns `Some` with the given event and a new list of connections if
-/// at least one of the given connections is not currently able to receive the event
-/// but is not closing, in which case the current task is scheduled to be woken up.
-/// The returned connections are those which are not closing.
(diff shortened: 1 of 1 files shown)


romanb:
It occurred to me after a question from @twittner that is seems unnecessary to implement `notify_all` via "poll all" -> "send all" semantics. Rather, the event can be delivered to a handler as soon as the handler is ready, only tracking those remaining connections whose handlers currently not ready. This further simplifies things and was done in https://github.com/libp2p/rust-libp2p/pull/1513/commits/c23075a3e84c4c3e57346bc4bf4d83d298b5d821.

Issue #1489, opened 2020-03-10
Replace get_if_addrs with a pure Rust library


tomaka:
For the sake of not having to compile any C code, let's try use only pure Rust dependencies.

mxinden:
We could use libpnet's [`interfaces`](https://github.com/libpnet/libpnet/blob/master/pnet_datalink/src/unix_interfaces.rs#L25) instead which for unix would use the systems libc instead of custom C code.

DemiMarie-parity:
Once https://github.com/paritytech/ip-watch is ready, this will be solved.

kallisti5:
It looks like this one is solved?  Having trouble building apps using rust-libp2p 0.22.0 because of poor "non-linux" support.  get_if_addrs is also now read-only and not accepting patches / fixes.

mxinden:
> It looks like this one is solved?

Not quite. I am afraid it gotten even more complicated.

With https://github.com/libp2p/rust-libp2p/pull/1830  `libp2p-mdns` now uses [`if-wtach`](https://github.com/dvc94ch/if-watch/) which uses a Rust implementation on Linux and Windows and [falls back to `if-addrs` on e.g. Mac](https://github.com/dvc94ch/if-watch/blob/master/src/fallback.rs).

`libp2p-tcp` still uses [`if-addr`](https://github.com/messense/if-addrs) which might change to `if-watch` with https://github.com/libp2p/rust-libp2p/pull/1845.

dvc94ch:
this one can be closed now, no?

PR #1514, opened 2020-03-25
Use pure rust libraries to get host interfaces

tcharding:
For the sake of not having to compile any C code, let's try use only pure Rust dependencies.  To get the host interface information we can use the `pnet` and `ipnetwork` libraries.  These are both pure Rust.

`pnet` v0.25 depends on `ipnetwork` v0.15 so we use that also instead of the latest `ipnetwork` v0.16.

Resolves: #1489

### DIFF
--- a/transports/tcp/Cargo.toml
+++ b/transports/tcp/Cargo.toml
@@ -13,10 +13,10 @@ categories = ["network-programming", "asynchronous"]
 async-std = { version = "1.0", optional = true }
 futures = "0.3.1"
 futures-timer = "3.0"
-get_if_addrs = "0.5.3"
-ipnet = "2.0.0"
+ipnetwork = "0.15"
 libp2p-core = { version = "0.16.0", path = "../../core" }
 log = "0.4.1"
+pnet = "0.25"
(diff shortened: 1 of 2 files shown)


tomaka:
Thanks. The PR in itself is good, but I'm not super confident about `pnet`.
It's quite a huge beast for "just" getting the interface addresses, and [this section](https://github.com/libpnet/libpnet#windows) suggests that building libp2p (and anything that depends on it) for Windows would become extremely painful.

tcharding:
> Thanks. The PR in itself has my green mark, but I'm not super confident about `pnet`.
> It's quite a huge beast for "just" getting the interface addresses, and [this section](https://github.com/libpnet/libpnet#windows) suggests that building libp2p (and anything that depends on it) for Windows would become extremely painful.

Thanks for the review, I was a bit unsure at the choice of libraries in this PR also.  I wanted to put up a PR to start the conversation to demonstrate some effort instead of just talking.  I'm more than happy to redo it with other libraries if you (or anyone) can suggest a better alternative.  Was fun to write this anyways.

tcharding:
Not really sure what to do about this right now, closing it so as to not clutter the place up with stale open PRs.

PR #1515, opened 2020-03-25
Make the SwarmEvent report everything

tomaka:
Adds new variants and improves the existing variants of the `SwarmEvent` in order to report everything that happens in the swarm in terms of connectivity.

This way, the outside can track exactly what happens.

Some notes:

- Reporting the errors enums (rather than opaque `Box<Error>`s) is going to be quite helpful in Substrate to detect connections closed because they were idle, and dial failures because of peer id mismatch.

- I changed some `usize`s into `u32`s. In general, `usize` should only be used for things that depend on the platform's pointer width.

- I purposefully left aside the concept of "external addresses", as the way we handle them is still quite blurry at the moment.


core/src/connection/pool.rs
```diff
@@ -580,7 +580,7 @@ where
                     let num_established =
                         if let Some(conns) = self.established.get_mut(connected.peer_id()) {
                             conns.remove(&id);
-                            conns.len()
+                            conns.len() as u32
```
romanb: It seems to me that the only scenario where this may realistically result in a lossy cast is one where we have a bug in the connection management / counting, so it may be worthwhile triggering a panic?

core/src/connection/pool.rs
```diff
@@ -600,7 +600,7 @@ where
                                             .map_or(0, |conns| conns.len());
                         if let Err(e) = self.limits.check_established(current) {
                             let connected = entry.close();
-                            let num_established = e.current;
+                            let num_established = e.current as u32;
```
romanb: Same question / suggestion as for the other case.

core/src/network.rs
```diff
@@ -517,7 +517,7 @@ where
         // A pending outgoing connection to a known peer failed.
         let mut attempt = dialing.remove(&peer_id).expect("by (1)");
 
-        let num_remain = attempt.next.len();
+        let num_remain = attempt.next.len() as u32;
```
romanb: Same question / suggestion as for the other cases.

core/src/network/event.rs
```diff
@@ -98,15 +98,18 @@ where
         /// Address used to send back data to the remote.
         send_back_addr: Multiaddr,
         /// The error that happened.
+        // TODO: This enum contains `PeerIdMismatch`, which cannot possibly happen for an incoming
+        // connection. Ideally we should use a different enum.
```
romanb: That's not quite right because I collapsed the cases `PeerIdMismatch` and `LocalPeerId` into `PendingConnectionError::InvalidPeerId` (i.e. the remote identified with an invalid peer ID, which either does not match the expected one or is equal to the local peer ID and thus similarly unexpected, or otherwise invalid). If you consider the distinction(s) important, you may want to open an issue or add a todo about splitting these errors again.

swarm/src/lib.rs
```diff
@@ -135,29 +142,111 @@ pub type NegotiatedSubstream = Negotiated<Substream<StreamMuxerBox>>;
...
-    /// We are now connected to the given peer.
-    Connected(PeerId),
-    /// We are now disconnected from the given peer.
-    Disconnected(PeerId),
-    /// One of our listeners has reported a new local listening address.
-    NewListenAddr(Multiaddr),
-    /// One of our listeners has reported the expiration of a listening address.
-    ExpiredListenAddr(Multiaddr),
+    /// A connection to the given peer has been opened.
+    NewConnection {
```
romanb:
```suggestion
    ConnectionEstablished {
```
The reason for these naming suggestions (and also `ConnectionClosed`) is two-fold:

  * It shows more symmetry, also to the underlying network events.
  * Keeping `Disconnected` can be misleading, especially since it is now emitted for every closed connection.
  * These names would merge quite well with a PR that I have almost ready that introduces `inject_connection_established` / `inject_connection_closed` to the `NetworkBehaviour`.

swarm/src/lib.rs
```diff
@@ -135,29 +142,111 @@ pub type NegotiatedSubstream = Negotiated<Substream<StreamMuxerBox>>;
...
+        /// Identity of the peer that we have connected to.
+        peer_id: PeerId,
+        /// Endpoint of the connection that has been opened.
+        endpoint: ConnectedPoint,
+        /// Number of established connections to this peer, including the one that has just been
+        /// opened.
+        num_established: NonZeroU32,
+    },
+    /// A connection with the given peer has been closed.
+    Disconnected {
```
romanb:
```suggestion
    ConnectionClosed {
```

swarm/src/lib.rs
```diff
@@ -135,29 +142,111 @@ pub type NegotiatedSubstream = Negotiated<Substream<StreamMuxerBox>>;
...
+        /// Identity of the peer that we have connected to.
+        peer_id: PeerId,
+        /// Endpoint of the connection that has been closed.
+        endpoint: ConnectedPoint,
+        /// Number of other remaining connections to this same peer.
+        num_established: u32,
+        /// Reason for the disconnection.
+        cause: ConnectionError<NodeHandlerWrapperError<THandleErr>>,
+    },
+    /// A new connection arrived on a listener and is now starting to handshake.
```
romanb:
```suggestion
    /// A new connection arrived on a listener and is in the process of protocol negotiation.
```

swarm/src/lib.rs
```diff
@@ -135,29 +142,111 @@ pub type NegotiatedSubstream = Negotiated<Substream<StreamMuxerBox>>;
...
+        peer_id: PeerId,
+        /// Endpoint of the connection that has been closed.
+        endpoint: ConnectedPoint,
+    },
+    /// Starting to try to reach the given peer.
+    ///
+    /// We are trying to connect to this peer until a [`NewConnection`](SwarmEvent::NewConnection)
+    /// event is reported, or a [`UnreachableAddr`](SwarmEvent::UnreachableAddr) event is reported
+    /// with `attempts_remaining` equal to 0.
+    StartConnect(PeerId),
```
romanb:
```suggestion
    Dialing(PeerId),
```

swarm/src/lib.rs
```diff
@@ -135,29 +142,111 @@ pub type NegotiatedSubstream = Negotiated<Substream<StreamMuxerBox>>;
...
         /// Error that has been encountered.
-        error: Box<dyn error::Error + Send>,
+        error: PendingConnectionError<io::Error>,
+        /// Number of remaining connection attempts that are being tried for this peer.
+        attempts_remaining: u32,
+    },
+    /// Tried to dial an address but it ended up being unreachaable.
+    /// Contrary to `UnreachableAddr`, we don't know the identity of the peer that we were trying
+    /// to reach.
+    // TODO: This variant can be in an inconsistent state if `error == PeerIdMismatch`
```
romanb: As per the other comment, this is not quite right, because there is only `PendingConnectionError::InvalidPeerId` which can also be triggered if the remote unexpectedly "identified" with the local peer ID.

core/src/network/event.rs
tomaka: Oh right. Sorry, like yesterday I still have the automatism of the old code.

Issue #1475, opened 2020-02-27
Pass the error to NetworkBehaviour::inject_listened_closed()


tomaka:

tcharding:
Taking a punt at this; Issue is meant to read `NetworkBehaviour::inject_listener_closed()`.  Ticket wants to pass the error, if there was one, from the `NetworkEvent::ListenerClosed` `reason` field to the `inject_listener_closed()` method.

PR #1517, opened 2020-03-25
Pass the error to inject_listener_closed method

tcharding:
If there is an error when the listener closes, found in the `NetworkEvent::ListenerClosed` `reason` field, we would like to pass it on to the `inject_listener_closed()` method so that implementors of this method have access to it.

Add an error parameter to `inject_listener_closed`.  Convert the `reason` field from a `Result` to an `Option` and if there is an error pass `Some(error)` at the method call site.

Resolves: #1475


swarm/src/behaviour.rs
```diff
@@ -130,7 +130,7 @@ pub trait NetworkBehaviour: Send + 'static {
     }
 
     /// A listener closed.
-    fn inject_listener_closed(&mut self, _id: ListenerId) {
+    fn inject_listener_closed(&mut self, _id: ListenerId, _err: Option<std::io::Error>) {
```
tomaka: Could you leave the `Result` for consistency with the rest of the library?

```suggestion
    fn inject_listener_closed(&mut self, _id: ListenerId, reason: Result<(), io::Error>) {
```

tcharding: Sure thing, can do.

tomaka:
Please pass the error by reference to `inject_listener_closed`

tomaka:
I took the liberty to finish the change, I hope that's ok.
It would be nice to get this PR in the next release, and this next release has already been waiting for a long time.

tcharding:
> Please pass the error by reference to `inject_listener_closed`

Face palm - obviously.  Thanks for finishing it.  FTR feel free to take, modify, paraphrase, or drop anything I contribute.  I'm just here to learn a bit and help out where it helps :)

Thanks.

PR #1519, opened 2020-03-25
Full support for multiple connections per peer in libp2p-swarm.

romanb:
This is a follow-up to #1440 that introduces "first class" support for multiple connections per peer to `libp2p-swarm`. Concretely:

  * New callbacks `inject_connection_established` and `inject_connection_closed` are added to the `NetworkBehaviour` with empty default impls.
  * `inject_connected` and `inject_disconnected` no longer get the `endpoint` as an argument. This is primarily because these callbacks are invoked for the first established connection, respectively last closed connection, whose endpoints may now differ, as opposed to the previous behaviour. Network behaviours that made use of the endpoint will naturally have to implement the new callbacks and split the code previously in `inject_connected` / `inject_disconnected`, as is demonstrated in some of the behaviours updated in this PR.
  * `NetworkBehaviourAction::DialPeer` as well es `ExpandedSwarm::dial`now take another parameter, a `DialPeerCondition`, which specifies the conditions under which a new dialing attempt should be initiated. The default (and what most behaviours migrating old code will want to use) is `DialPeerCondition::Disconnected`.
  * The `NetworkConfig` of a swarm (which includes connection limits) can now be fully customised via `SwarmBuilder::network_config`, in order to avoid code duplication. For example, substrate will probably want to configure the maximum connections per peer to `2` (just to permit concurrent dialing attempts) and possibly also configure limits for concurrent inbound and outbound connections.

It is worth mentioning that I took the opportunity to change `inject_connected` to take the peer ID by reference, rather than by value, just like `inject_disconnected` and most other methods. This simplifies things a bit in the derive-macro code and can avoid a bit of unnecessary cloning. I further did this because the signature of `inject_connected` changed anyway in this PR.

This PR is expected to be rebased over https://github.com/libp2p/rust-libp2p/pull/1515 eventually.


swarm/src/behaviour.rs
```diff
@@ -253,3 +271,33 @@ pub enum NotifyHandler {
     All
 }
 
+/// The available conditions under which a new dialing attempt to
+/// a peer is initiated when requested by [`NetworkBehaviourAction::DialPeer`].
+#[derive(Debug, Copy, Clone)]
+pub enum DialPeerCondition {
```
twittner: Should we mark this with `#[non_exhaustive]` perhaps? Client code will most likely not match on this anyway and judging by the TODO below we may want to add other cases.

romanb: Sounds good to me!

swarm/src/lib.rs
```diff
@@ -283,26 +293,69 @@ where TBehaviour: NetworkBehaviour<ProtocolsHandler = THandler>,
...
+    /// to the ongoing dialing attempt, ignoring duplicates.
     ///
-    /// Has no effect if we are already connected to that peer, or if no address is known for the
-    /// peer.
-    pub fn dial(me: &mut Self, peer_id: PeerId) {
+    /// If no addresses are reported for the peer by the behaviour, this is a noop.
+    ///
+    /// Returns `Some` with the connection ID of the first connection
+    /// attempt if a new dialing attempt has been initiated.
+    pub fn dial(me: &mut Self, peer_id: PeerId, condition: DialPeerCondition) -> Option<ConnectionId> {
```
tomaka: I'm not a fan of this API.
I agree that it is better than the current one, but in my opinion it would be better to have something more explicit.
I don't have any precise opinion, but a function that is called `dial` should *always* dial, without any condition attached to it.
If we want the user to be able to dial only under certain conditions, then we should provide an API similar to the one in `Network`, where you can first ask for the state of the connectivity with the peer.

swarm/src/lib.rs
```diff
@@ -156,8 +166,8 @@ pub enum SwarmEvent<TBvEv> {
         /// Error that has been encountered.
         error: Box<dyn error::Error + Send>,
     },
-    /// Startng to try to reach the given peer.
-    StartConnect(PeerId),
+    /// A new dialing attempt has been initiated.
+    Dialing(PeerId, ConnectionId),
```
tomaka: There are two different ways to go for this event, in my opinion:

- Either it is emitted every single time we start an outgoing connection, in which case it should include the `Multiaddr` that we're trying to reach.
- Or we emit it when we start trying to reach the given peer (as was the case for `StartConnect`), in which case I don't think it should contain the `ConnectionId`.

Right now this `ConnectionId` refers to the first connection attempt (out of possibly many), and is not very useful.

swarm/src/lib.rs
```diff
@@ -828,35 +893,17 @@ where TBehaviour: NetworkBehaviour,
...
-    pub fn executor_fn(mut self, executor: impl Fn(Pin<Box<dyn Future<Output = ()> + Send>>) + Send + 'static) -> Self {
-        struct SpawnImpl<F>(F);
-        impl<F: Fn(Pin<Box<dyn Future<Output = ()> + Send>>)> Executor for SpawnImpl<F> {
-            fn exec(&self, f: Pin<Box<dyn Future<Output = ()> + Send>>) {
-                (self.0)(f)
-            }
-        }
-        self.network.set_executor(Box::new(SpawnImpl(executor)));
+    /// Configures the underlying network for the swarm.
+    pub fn network_config(mut self, cfg: NetworkConfig) -> Self {
```
tomaka: This doesn't look so great to me in terms of usability.

Instead of:
```rust
SwarmBuilder::new(...).executor(...).build()
```

One now has to do:
```rust
let mut cfg = NetworkConfig::default();
cfg.set_executor(...);
SwarmBuilder::new(...).network_config(cfg).build();
```

I've had that remark already for `PingConfig`.
It can really become very annoying to have to browse half of the documentation and import half of the library to create types that pretty much all the users of libp2p will have to create.

romanb: Well, `libp2p-swarm` can certainly re-export things to make imports easier, but I really don't think essentially duplicating the `NetworkConfig` API on the `SwarmBuilder` is a better strategy, at least I wouldn't want to maintain that.

tomaka:
The only thing that I think should be resolved before merging is the `SwarmEvent::Dialing`. My other two remarks are mostly for discussion.

swarm/src/lib.rs
romanb: Ok, it sounds a bit like a stylistic preference which I'm not sure I can appreciate, but you are fine with the `DialPeerCondition` on `NetworkBehaviourAction::DialPeer`? Since the `NetworkBehaviour` has no means to figure out whether a peer is currently being dialed as it only interacts via callbacks, `Swarm::dial` is essentially modeled analogous to emitting `NetworkBehaviourAction::DialPeer`, thus showing both a certain symmetry on these APIs and reuse of data types, which to me seemed a good thing, but I'm not married to it.

swarm/src/lib.rs
romanb: > Or we emit it when we start trying to reach the given peer (as was the case for StartConnect), in which case I don't think it should contain the ConnectionId.

Note that `StartConnect` was previously always emitted, even if `ExpandedSwarm::dial` didn't actually do anything. It is only now that it is only emitted when a new dialing attempt has indeed been started. I can see why you may not consider the connection ID currently useful, since the Swarm does not permit direct access to the underlying `Network` API (which I thought may be useful in the future, but is certainly not necessary now). However, `ConnectionEstablished` and `ConnectionClosed` also carry the `ConnectionId` and these are at least mandatory if we want to do https://github.com/libp2p/rust-libp2p/issues/1522. However, if you prefer for this PR we could remove all connection IDs from the `SwarmEvent`, if that sounds ok to you?

swarm/src/behaviour.rs
romanb: On second thought, I think precisely because client code is not expected to match on it, but only libp2p-swarm code, we would want the exhaustiveness check when introducing new variants. `#[non_exhaustive]` seems to me primarily useful if one does not want to break client code matching on an enum which is known to be "open-ended" in the potential number of variants (and whereby handling of the new variants is optional in the vast majority of cases). Does that make sense?

swarm/src/lib.rs
romanb: I guess we could justify the API duplication by stating that, in principle, we want to keep open the possibility of `libp2p-swarm` to make predetermined (i.e. opinionated) choices on some of the underlying network configuration, even if we do not currently do so. A very similar justification could be made for not exposing the underlying `Network` of a `Swarm`, where additionally direct access to the network APIs may even cause potential state kept in a `Swarm` to become out-of-sync.

swarm/src/lib.rs
tomaka: > Note that StartConnect / Dialing was previously always emitted, even if ExpandedSwarm::dial didn't actually do anything.

That looks wrong to me.

> I can see why you may not consider the connection ID currently useful, since the Swarm does not permit direct access to the underlying Network API

Correct me if I'm wrong, but each dialing attempt gets its own `ConnectionId`, right?

Imagine if a certain node has two addresses A and B. When we start dialing address A, we emit a `Dialing` event with `A`'s `ConnectionId`. Supposing that dialing address A fails, we immediately start dialing B, which has a different `ConnectionId`, except that we don't report this new `ConnectionId` to the user in any way.

That's the problem that I was pointing out, but I'm not sure if I'm correct here.

swarm/src/behaviour.rs
romanb: Nevermind, I was under the mistaken impression `non_exhaustive` would also affect the crate in which the enum is defined. Given that this is not so, my previous comment is nonsensical.

swarm/src/lib.rs
romanb: > > Note that StartConnect / Dialing was previously always emitted, even if ExpandedSwarm::dial didn't actually do anything.
> 
> That looks wrong to me.

I agree, which is why I changed it, but this has been the case [at least for a couple months](https://github.com/libp2p/rust-libp2p/blame/84b6a7d04ded33216d5579b629731988c721fb1f/swarm/src/lib.rs#L469-L470).

> > I can see why you may not consider the connection ID currently useful, since the Swarm does not permit direct access to the underlying Network API
> 
> Correct me if I'm wrong, but each dialing attempt gets its own `ConnectionId`, right?

A "dialing attempt" is by definition a sequence of "connection attempts" that terminates with the first successful connection or with the last failed connection, and each connection attempt gets its own `ConnectionId`, yes. There is no such thing as a "dialing attempt ID" (there never was).

> Imagine if a certain node has two addresses A and B. When we start dialing address A, we emit a `Dialing` event with `A`'s `ConnectionId`. Supposing that dialing address A fails, we immediately start dialing B, which has a different `ConnectionId`, except that we don't report this new `ConnectionId` to the user in any way.

That is correct. While the user is not actively notified though, all current dialing connections can be seen at any time in the `Peer` API of the `Network`. It is just the result of introducing the notion of `ConnectionId`s (and multiple connections in general) while preserving the existing dialing attempt semantics. We could certainly introduce another network event for each successive dialing attempt, but that seems out of scope of this PR.

For this PR, would you then prefer to just remove the `ConnectionId` from the `Dialing` event?

tomaka: > A "dialing attempt" is by definition a sequence of "connection attempts" to different addresses that terminates with the first successful connection or with the last failed connection

That was poor wording from me. By "dialing attempt" here I was thinking of connection attempt.

> For this PR, would you then prefer to just remove the ConnectionId from the Dialing event?

I would indeed prefer.

swarm/src/lib.rs
tomaka: > I guess we could justify the API duplication by stating that, in principle, we want to keep open the possibility of libp2p-swarm to make predetermined (i.e. opinionated) choices on some of the underlying network configuration, even if we do not currently do so.

That was indeed my reasoning.

The `Swarm` configuration and the `Network` configuration do intersect in some way, but aren't necessarily a subset or a superset of one another.

swarm/src/lib.rs
romanb: Resolved by https://github.com/libp2p/rust-libp2p/pull/1519/commits/4fe71ed7166739602fe3760b7266c87ce92081ba. I also kept the connection IDs out of `SwarmEvent::ConnectionEstablished` and `SwarmEvent::ConnectionClosed` for now, since there is no point as long as the Swarm provides no APIs to work with connections directly.

swarm/src/behaviour.rs
romanb: Resolved by https://github.com/libp2p/rust-libp2p/pull/1519/commits/5c016989ecf4fc275c9f23c5ad729095efa7d7ac.

swarm/src/lib.rs
romanb: Resolved by https://github.com/libp2p/rust-libp2p/pull/1519/commits/f90f548dfd81bb4da762e6bd870d947a3f9902e0.

swarm/src/lib.rs
romanb: I reworked the `Swarm::dial` API with some related adjustments to make `Swarm::dial` and `Swarm::dial_addr` (as well as the underlying `Network` functions) more alike in https://github.com/libp2p/rust-libp2p/pull/1519/commits/e047ff93b488ae68a4191d1338fc63318654837b. There are some subtleties to preserve the existing behaviour of always adding the result of `addresses_of_peer` to the current dialing attempt, if there is one, though I don't think we can keep this behaviour for `Swarm::dial` if https://github.com/libp2p/rust-libp2p/pull/1506 is accepted without introducing new parameters to `Swarm::dial` due to ambiguity of meaning (if we want to distinguish "add addresses to current dialing attempt if there is one, otherwise start dialing" vs "always start new dialing attempt"; intuitively and for simplicity, https://github.com/libp2p/rust-libp2p/pull/1506 would have to always do the latter for `Swarm::dial`.)

PR #1520, opened 2020-03-25
feat: allow sent messages seen as subscribed

koivunej:
minor feature to allow mimicing the behaviour expected by ipfs http api tests. default behaviour is not changed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Version ???
 
+- `libp2p-floodsub`: Allow sent messages seen as subscribed
+  [PR 1520](https://github.com/libp2p/rust-libp2p/pull/1520)
 
 # Version 0.17.0 (2020-04-02)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Version ???
 
-- `libp2p-floodsub`: Allow sent messages seen as subscribed
+- `libp2p-floodsub`: Allow sent messages seen as subscribed.
   [PR 1520](https://github.com/libp2p/rust-libp2p/pull/1520)
 
 # Version 0.17.0 (2020-04-02)
--- a/protocols/floodsub/src/layer.rs
+++ b/protocols/floodsub/src/layer.rs
@@ -20,6 +20,7 @@
 
 use crate::protocol::{FloodsubConfig, FloodsubMessage, FloodsubRpc, FloodsubSubscription, FloodsubSubscriptionAction};
 use crate::topic::Topic;
+use crate::FloodsubOptions;
 use cuckoofilter::CuckooFilter;
 use fnv::FnvHashSet;
 use libp2p_core::{Multiaddr, PeerId, connection::ConnectionId};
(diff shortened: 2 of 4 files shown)


protocols/floodsub/src/lib.rs
```diff
@@ -33,3 +35,22 @@ mod rpc_proto {
 pub use self::layer::{Floodsub, FloodsubEvent};
 pub use self::protocol::{FloodsubMessage, FloodsubRpc};
 pub use self::topic::Topic;
+
+/// Configuration options for the Floodsub protocol.
+pub struct FloodsubOptions {
```
tomaka: Could you please name this "config", for consistency with the rest of the library?
```suggestion
pub struct FloodsubConfig {
```

protocols/floodsub/src/layer.rs
```diff
@@ -63,11 +63,16 @@ pub struct Floodsub {
...
 
 impl Floodsub {
-    /// Creates a `Floodsub`.
+    /// Creates a `Floodsub` with default options.
     pub fn new(local_peer_id: PeerId) -> Self {
+        Self::from_options(FloodsubOptions::new(local_peer_id))
+    }
+
+    /// Creates a `Floodsub` with the given options.
+    pub fn from_options(options: FloodsubOptions) -> Self {
```
tomaka:
```suggestion
    pub fn from_config(options: FloodsubConfig) -> Self {
```

protocols/floodsub/src/lib.rs
koivunej: I went with "options" because the `libp2p_floodsub::protocol::FloodsubConfig` exists already for protocol (upgrade).. I don't have strong opinion on this naming, but having two of the same name sounds odd. Could you confirm this should be `FloodsubConfig`?

koivunej: Just found https://github.com/libp2p/rust-libp2p/pull/1015 which does propose the `libp2p_floodsub::FloodsubConfig` and renaming the exiting `FloodsubConfig` to `FloodsubListen`... Should I try to bring up the #1015 first and then extend with this behavioural change?

tomaka: #1015 is extremely old, but yes in my opinion we should rename `protocol::FloodsubConfig` to something else.

koivunej:
Rebased on top of latest master, renames:

 * `floodsub::protocol::Floodsub{Config => Protocol}`
 * `floodsub::Floodsub{Options => Config}`
 * `floodsub::layer::Floodsub::from_{options => config}`
 * `floodsub::layer::Floodsub::{options => config}`

I think this covers everything noted.

CHANGELOG.md
```diff
@@ -45,6 +45,9 @@
 - `multihash`: Removed the crate in favour of the upstream crate.
   [PR 1472](https://github.com/libp2p/rust-libp2p/pull/1472)
 
+- `libp2p-floodsub`: Allow sent messages seen as subscribed
```
tomaka, line 48: Please merge master and move this at the top, since we published a version in between!

koivunej: Done and done.

koivunej:
Rebased on top of master. Looks like the #1532 didn't include an update to CHANGELOG as there was no conflict.

CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 # Version ???
 
+- `libp2p-floodsub`: Allow sent messages seen as subscribed
```
mxinden, line 3:
```suggestion
- `libp2p-floodsub`: Allow sent messages seen as subscribed.
```

PR #1521, opened 2020-03-25
Allow configuration of outbound substream in OneShotHandler

sireliah:
When opening a new outbound substream in `OneShotHandler`, you should be able to specify the timeout for the substream. 

Follow up of discussion here: https://github.com/libp2p/rust-libp2p/pull/1512

Do you like this solution or do you prefer the `.with_substream_timeout(timeout)` way?


swarm/src/protocols_handler/one_shot.rs
```diff
@@ -67,7 +69,8 @@ where
     #[inline]
     pub fn new(
         listen_protocol: SubstreamProtocol<TInProto>,
-        inactive_timeout: Duration
+        inactive_timeout: Duration,
+        substream_timeout: Duration
```
tomaka: I saw this PR before you force-pushed to it, and I preferred the way it was before, with a configuration struct!

Could you please extract these two fields?

```suggestion
        config: OneShotHandlerConfig,
```

sireliah: Done!

swarm/src/protocols_handler/one_shot.rs
```diff
@@ -236,3 +240,25 @@ where
         Poll::Pending
     }
 }
+
+/// Configuration parameters for the `OneShotHandler`
+pub struct OneShotHandlerConfig {
```
tomaka:
```suggestion
#[derive(Debug)]
pub struct OneShotHandlerConfig {
```

swarm/src/protocols_handler/one_shot.rs
```diff
@@ -236,3 +240,25 @@ where
...
+    inactive_timeout: Duration,
+    /// Timeout duration for each newly opened outbound substream.
+    substream_timeout: Duration,
+}
+
+impl OneShotHandlerConfig {
+    fn new(inactive_timeout: Duration, substream_timeout: Duration) -> Self {
+        OneShotHandlerConfig { inactive_timeout, substream_timeout }
+    }
+}
```
tomaka:
```suggestion
```

I don't really see the value of this `new` function.

swarm/src/protocols_handler/one_shot.rs
```diff
@@ -236,3 +240,25 @@ where
...
+
+impl Default for OneShotHandlerConfig {
+    fn default() -> Self {
+        let inactive_timeout = Duration::from_secs(10);
+        let substream_timeout = Duration::from_secs(10);
+        OneShotHandlerConfig::new(inactive_timeout, substream_timeout)
```
tomaka:
```suggestion
        OneShotHandlerConfig { inactive_timeout, substream_timeout }
```

swarm/src/protocols_handler/one_shot.rs
```diff
@@ -236,3 +240,25 @@ where
...
+    }
+}
```
tomaka:
```suggestion
}

```

romanb:
I may be blind, but how does one now actually configure a `OneShotHandlerConfig`? I only see private fields and a `Default` impl.

romanb:
Moreover, the `one_shot` is a private module and `OneShotHandlerConfig` is not re-exported by `protocols_handler`, which only leaves `OneShotHandler::default()` usable, since `OneShotHandler::new` requires a `OneShotHandlerConfig`. I'm a bit puzzled.

PR #1524, opened 2020-03-26
Update quicksink dependency.

twittner:


protocols/secio/Cargo.toml
```diff
@@ -20,7 +20,7 @@ libp2p-core = { version = "0.16.0", path = "../../core" }
 log = "0.4.6"
 prost = "0.6.1"
 pin-project = "0.4.6"
-quicksink = "0.1"
+quicksink = { git = "https://github.com/paritytech/quicksink", branch = "develop" }
```
twittner: If this PR is approved, the new quicksink version will be published to crates.io and this entry will become `quicksink = "0.2"`.

transports/websocket/Cargo.toml
```diff
@@ -16,7 +16,7 @@ either = "1.5.3"
 futures = "0.3.1"
 libp2p-core = { version = "0.16.0", path = "../../core" }
 log = "0.4.8"
-quicksink = "0.1"
+quicksink = { git = "https://github.com/paritytech/quicksink", branch = "develop" }
```
twittner: If this PR is approved, the new quicksink version will be published to crates.io and this entry will become `quicksink = "0.2"`.

transports/websocket/src/framed.rs
```diff
@@ -410,7 +410,7 @@ fn location_to_multiaddr<T>(location: &str) -> Result<Multiaddr, Error<T>> {
 /// The websocket connection.
 pub struct Connection<T> {
     receiver: BoxStream<'static, Result<data::Incoming, connection::Error>>,
-    sender: Pin<Box<dyn Sink<OutgoingData, Error = connection::Error> + Send>>,
+    sender: Pin<Box<dyn Sink<OutgoingData, Error = quicksink::Error<connection::Error>> + Send>>,
```
mxinden: This leaks a dependency into the public interface, correct? As the users are likely not aware of what `quicksink` is, how about doing the same trick as above converting it to an `io` error?

twittner: No worries, the type stays private. The only API `Connection` exposes is `Stream<Item = std::io::Result<_, _>>` and `Sink<Error = std::io::Error>` plus some methods that utilise the `Sink` impl.

twittner:
The motivation for the `quicksink` change ― avoiding a panic when polled after an error occurred ― are somewhat defeated by `SinkExt::sink_map_err` which uses an `FnOnce` and panics on subsequent mapping attempts.

On the upside, this clarifies the somewhat generic "In most cases, if the sink encounters an error, the sink will permanently be unable to receive items." from the Sink API documentation, i.e. one should read that without the "In most cases" part if one is interested in avoiding a panic by a second error mapping.

twittner:
As mentioned adding `quicksink::Error` does not fit well into the `Sink` design space so I close this PR.

PR #1527, opened 2020-03-30
core/src: Remove poll_broadcast connection notification mechanism

### LINKED ISSUE #1465: Potential unbounded buffer growth in background task of pending connection.

mxinden:
The `Network::poll_broadcast` function has not proven to be useful. This
commit removes the mechanism all the way down to the connection manager.

With `poll_broadcast` gone there is no mechanism left to send commands
to pending connections. Thereby command buffering for pending
connections is not needed anymore and is thus removed in this commit as
well.

---

Fixes #1465.

Replaces https://github.com/libp2p/rust-libp2p/pull/1501.

### DIFF
--- a/core/src/connection/manager.rs
+++ b/core/src/connection/manager.rs
@@ -134,9 +134,6 @@ where
 #[derive(Debug)]
 struct TaskInfo<I, C> {
     /// Channel endpoint to send messages to the task.
-    //
-    // Note: Only established tasks can handle commands. Do not send commands to
-    // pending tasks (see `[<Task as Future>::poll]`).
     sender: mpsc::Sender<task::Command<I>>,
     /// The state of the task as seen by the `Manager`.
     state: TaskState<C>,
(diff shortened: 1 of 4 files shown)


core/src/connection/manager.rs
```diff
@@ -133,7 +133,10 @@ where
 /// the associated user data.
 #[derive(Debug)]
 struct TaskInfo<I, C> {
-    /// channel endpoint to send messages to the task
+    /// Channel endpoint to send messages to the task.
+    //
+    // Note: Only established tasks can handle commands. Do not send commands to
+    // pending tasks (see `[<Task as Future>::poll]`).
```
tomaka:
```suggestion
```

Nit: seems a bit too specific.

mxinden:
Thanks for the quick review. I will leave this open for another day in case anyone out there depends on `poll_broadcast` and has objections.

Issue #1523, opened 2020-03-26
Panic in deep of libp2p Sink::poll_ready applied to a closed sink.


vldm:
Hi.
I have an application that use libp2p 0.13 with custom written protocols. It uses floodsub with some modification and other protocols.  I tried to port it on new async/await, my plan was to add gossip support during this port.
I made everything compile, tested my app with only gossip, and its work. But when i tried to make behaviour with multiple protocols, i caught a panic.

```
thread 'libp2p-task-1' panicked at 'Sink::poll_ready applied to a closed sink.', /home/vladimir/.cargo/registry/src/github.com-1ecc6299db9ec823/quicksink-0.1.1/src/lib.rs:151:30
stack backtrace:
   0: backtrace::backtrace::libunwind::trace
             at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.40/src/backtrace/libunwind.rs:88
   1: backtrace::backtrace::trace_unsynchronized
             at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.40/src/backtrace/mod.rs:66
   2: std::sys_common::backtrace::_print_fmt
             at src/libstd/sys_common/backtrace.rs:84
   3: <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt
             at src/libstd/sys_common/backtrace.rs:61
   4: core::fmt::write
             at src/libcore/fmt/mod.rs:1030
   5: std::io::Write::write_fmt
             at src/libstd/io/mod.rs:1412
   6: std::sys_common::backtrace::_print
             at src/libstd/sys_common/backtrace.rs:65
   7: std::sys_common::backtrace::print
             at src/libstd/sys_common/backtrace.rs:50
   8: std::panicking::default_hook::{{closure}}
             at src/libstd/panicking.rs:188
   9: std::panicking::default_hook
             at src/libstd/panicking.rs:205
  10: std::panicking::rust_panic_with_hook
             at src/libstd/panicking.rs:464
  11: std::panicking::begin_panic
             at /rustc/412f43ac5b4ae8c3599e71c6972112e9be4758fa/src/libstd/panicking.rs:400
  12: <quicksink::SinkImpl<S,F,T,A,E> as futures_sink::Sink<A>>::poll_ready
             at ./<::std::macros::panic macros>:3
  13: <core::pin::Pin<P> as futures_sink::Sink<Item>>::poll_ready
[...]

Issue #1504, opened 2020-03-23
Substream limits exceeded


michaelvoronov:
I have stumbled with problem of reaching the maximum number of substreams inside multiplexors. 

I have complex test, and will try to describe it: from 20 nodes using libp2p-kad for connectivity I constructed 176 pairs (with `tuple_combinations`), then created 176 clients connected to this nodes. In each pair one client sends some info to other client. Each node has two separated services based on different libp2p's swarms: one for clients, other for nodes in the network. Node swarm behaviour combines kademlia, identity, ping and my own protocol. My own protocol is pretty simple and is based on the `OneShotHandler`. Peer swarm just controls clients and sends requests from them to node swarm. Clients are separated entities and also built on rust-libp2p.

Node swarm after receiveing request from a client uses the `get providers` capability of libp2p-kad for finding other client (second in the corresponding pair). After that it tries to connect to it by my own protocol and send some short message. So the flow looks like this: 
`client_1` -> `node_1` {`client swarm` -> `node swarm` (FindProviders-Connect-Send)} -> `node_2` {`node swarm` -> `client swarm`} -> `client_2`.

The first client from each pair sends 100 message to the second client. All pairs are placed in `FuturesUnordered` so sending beteen pairs is parallel but inside one pair is sequential - the test waits for recevieng message by the second client with timeout 5 sec.

All of these works fine (without any timeouts) on my 176 pairs only If I increase a number of maximum substreams on multiplexors (both mplex and yamux). But after several tries it stucks also after reaching the set maximum of substreams or smth like that - I have seen several corresponding errors of reaching substream limits from mplex in logs. And for example if I increase substream counts from default 128 for mplex to 1024 the test completes thefirst time successfully. Also from logs it could be seen that message is successfully passed to node swarm of the `node_1` connected to the `client_1`. 

I have instrumented mplex a little bit and it seems that finding providers generates several new substreams (5-6 on my network) but only one of them will be closed. 

All of this looks like the root case in the stream management on the libp2p-kad side, but after very quick analysis couldn't find the cause.

Also very rarely I have such panic:
```rust
thread 'libp2p-task-1' panicked at 'Sink::poll_ready applied to a closed sink.', <::std::macros::panic macros>:2:4
stack backtrace:
   0: backtrace::backtrace::libunwind::trace
             at ./cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.40/src/backtrace/libunwind.rs:88
[...]

PR #1529, opened 2020-03-30
mplex: Check for error and shutdown.

twittner:
Issues #1504 and #1523 reported panics caused by polling the sink of `secio::LenPrefixCodec` after it had entered its terminal state, i.e. after it had previously encountered an error or was closed. According to the reports this happened only when using mplex as a stream multiplexer. It seems that because mplex always stores and keeps the `Waker` when polling, a wakeup of any of those wakers will resume the polling even for those cases where the previous poll did not return `Poll::Pending` but resolved to a value.

To prevent polling after the connection was closed or an error happened we check for those conditions prior to every poll.


romanb:
I may be missing something, but given that a `Sink` is deemed unusable after an error (and operations may panic), shouldn't we set `inner.error` in [poll_outbound](https://github.com/libp2p/rust-libp2p/pull/1529/files#diff-04961af40722cf221f6c77d3c1df6034R438), [write_substream](https://github.com/libp2p/rust-libp2p/pull/1529/files#diff-04961af40722cf221f6c77d3c1df6034R545), [flush_substream](https://github.com/libp2p/rust-libp2p/pull/1529/files#diff-04961af40722cf221f6c77d3c1df6034R555), [shutdown_substream](https://github.com/libp2p/rust-libp2p/pull/1529/files#diff-04961af40722cf221f6c77d3c1df6034R569), [close](https://github.com/libp2p/rust-libp2p/pull/1529/files#diff-04961af40722cf221f6c77d3c1df6034R600) and [flush_all](https://github.com/libp2p/rust-libp2p/pull/1529/files#diff-04961af40722cf221f6c77d3c1df6034R614) so that `ensure_no_error_no_close` can take effect, i.e. whenever a `Sink` operation returns an error?

twittner:
> I may be missing something, but given that a `Sink` is deemed unusable after an error (and operations may panic), shouldn't we set `inner.error` in [...] so that `ensure_no_error_no_close` can take effect, i.e. whenever a `Sink` operation returns an error?

I think this would help ;-)

PR #1535, opened 2020-03-31
Fix parsed is null error

tomaka:
cc @expenses 

This JS code hasn't really been properly reviewed by anyone unfortunately, and jumped a bit suddenly from "quickly written code to prototype" to "thing that we expect to be robust".

One consequence is that for some reason we checked `parsed != null` only after using `parsed`.


tomaka:
> Why reinvent the wheel instead of using https://github.com/multiformats/js-multiaddr ?

Note that we're supporting the `x-parity-ws` and `x-parity-wss` protocols (extensions so that one can add a URL to WebSocket addresses), which aren't a thing in `js-multiaddr`.

PR #1538, opened 2020-04-01
Re-export OneShotHandlerConfig. Make fields public.

romanb:
This is supposed to address the issues I ran into with the recent introduction of `OneShotHandlerConfig`: https://github.com/libp2p/rust-libp2p/pull/1521#issuecomment-607254753.


sireliah:
Thanks for finding this out @romanb!

PR #1539, opened 2020-04-01
[Gossipsub] Correct max transmit size error handling

AgeManning:
## Description

If a user sends a message that is over the maximum transmission size gossipsub will disconnect from the peer being sent the message.

This PR updates the logic to simply emit an error, not send the over-sized message but maintain the long-lived streams for future messages.

### DIFF
--- a/protocols/gossipsub/src/handler.rs
+++ b/protocols/gossipsub/src/handler.rs
@@ -27,7 +27,7 @@ use libp2p_swarm::protocols_handler::{
     KeepAlive, ProtocolsHandler, ProtocolsHandlerEvent, ProtocolsHandlerUpgrErr, SubstreamProtocol,
 };
 use libp2p_swarm::NegotiatedSubstream;
-use log::{debug, trace, warn};
+use log::{debug, error, trace, warn};
 use smallvec::SmallVec;
 use std::{
     borrow::Cow,
@@ -274,7 +274,13 @@ impl ProtocolsHandler for GossipsubHandler {
                                         Some(OutboundSubstreamState::PendingFlush(substream))
                                 }
                                 Err(e) => {
-                                    return Poll::Ready(ProtocolsHandlerEvent::Close(e));
+                                    if let io::ErrorKind::PermissionDenied = e.kind() {
+                                        error!("Message over the maximum transmission limit was not sent.");
+                                        self.outbound_substream =
+                                            Some(OutboundSubstreamState::WaitingOutput(substream));
+                                    } else {
+                                        return Poll::Ready(ProtocolsHandlerEvent::Close(e));
+                                    }
                                 }
                             }
                         }


protocols/gossipsub/src/handler.rs
```diff
@@ -274,7 +274,13 @@ impl ProtocolsHandler for GossipsubHandler {
                                         Some(OutboundSubstreamState::PendingFlush(substream))
                                 }
                                 Err(e) => {
-                                    return Poll::Ready(ProtocolsHandlerEvent::Close(e));
+                                    if let io::ErrorKind::PermissionDenied = e.kind() {
+                                        error!("Message over the maximum transmission limit was not sent.");
```
mxinden, line 278: Am I right assuming that you encountered this using secio [here](https://github.com/libp2p/rust-libp2p/blob/master/protocols/secio/src/codec/len_prefix.rs#L60)?

I am not sure it is safe to assume that `PermissionDenied` is always related to maximum transmission limit failures. Thus I would suggest including the original error message here.

I do think that `PermissionDenied` always implies a mistake on the local node's end, thus not disconnecting sounds good to me.

AgeManning:
Hey guys, 

I also agree that checking this earlier on would be nicer (although this currently checks before sending any data on the stream). The reason I haven't done this is because the max size is based the size of the protobuf encoded GossipMessage:
https://github.com/libp2p/rust-libp2p/blob/master/protocols/gossipsub/src/protocol.rs#L204

This can contain many control message, subscriptions and actual messages.  IIRC in the heartbeat we try and group these messages into a single GossipsubRpc. 
I don't think it's trivial to just include a check in `publish_many()` as we could still hit the limit with subscriptions and control messages. Future optimisations (and extra items in the protobuf message due to gossipsub 1.1) may also add on subscriptions/control messages to a standard message, so a check in the publish would also have to account for this. 

I'm open to suggestions to how better address this. This current solution handles it before sending it. The message hits the codec, fails, logs a warning and leaves the stream intact. 

The `PermissionDenied` kind occurs from the UVICodec when the bytes exceed the max size:
https://docs.rs/unsigned-varint/0.3.2/src/unsigned_varint/codec.rs.html#156

Again, open to suggestions to handle this in a nicer way.

protocols/gossipsub/src/handler.rs
AgeManning: I can't recall, I use secio and noise often. 

The `PermissionDenied` is an error that the UVICodec emits when we go over the maximum size. This is independent of the substream type, transport or authentication layer. 

What is happening, is the message gets sent to the `Framed` stream. It uses the GossipsubCodec to encode the data. The UVICodec realises the encoded message is too large and returns `PermissionDenied`. This will always occur regardless of any underlying stream.

This also happens locally, no information is sent to the underlying substream.

mxinden:
> The reason I haven't done this is because the max size is based the size of the protobuf encoded GossipsubRpc:
https://github.com/libp2p/rust-libp2p/blob/master/protocols/gossipsub/src/protocol.rs#L204

But one could estimate the encoded size based on the un-encoded size, no? At least set an upper limit.

> This can contain many control message, subscriptions and actual messages. IIRC in the heartbeat we try and group these messages into a single GossipsubRpc.

In this case not hitting the limit would need to be part of whatever [bin packing](https://en.wikipedia.org/wiki/Bin_packing_problem) logic we use, right? We would just need to ensure that no single message already exceeds the limit.

> I'm open to suggestions to how better address this. This current solution handles it before sending it. The message hits the codec, fails, logs a warning and leaves the stream intact.

True, but as far as I can tell it drops the message itself, right?

AgeManning:
> But one could estimate the encoded size based on the un-encoded size, no? At least set an upper limit.

Yep we could do this. We could work out the overhead around a single message. The cost per subscription and control message. When grouping messages, as you point out, be clever around packing them into a message. 

> True, but as far as I can tell it drops the message itself, right?

Yes. This could be avoided in the case where (in the current code) we over-group messages and instead should split into smaller chunks. If the user requests to send a message that is too large, there is no way to send it, so I imagine the desired logic is to drop it right?

If we are opting to go down the path of trying to evaluate sizes before being sent and handling these cases when grouping messages, then I think I'll leave these changes for a future PR. I'm intending on building gossipsub 0.1.1 and need to make a few changes to the current implementation also. I
ll likely put these changes in with those. 

This current PR, is however a quick solution to notify user's of the error rather than leaving them in the dark and randomly disconnecting peers. 

I'll leave it up to you as to whether you think its best to close this PR and wait for the 1.1 upgrade or use this as a temporary measure.

mxinden:
> If the user requests to send a message that is too large, there is no way to send it, so I imagine the desired logic is to drop it right?

I would expect `publish` to return an error in that case.

> I'll leave it up to you as to whether you think its best to close this PR and wait for the 1.1 upgrade or use this as a temporary measure.

I think we should merge this pull request. Logging an error is definitely better than silently ignoring it and terminating the connection. Thanks @AgeManning! **In case there are no objections by others I will merge next week Monday.**

@AgeManning with `1.1` are you referring to the [v1.1 gossipsub spec](https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/gossipsub-v1.1.md). If so I am looking forward to reviewing your changes!

AgeManning:
Yep, am talking about the v1.1 update. 

I've also noticed another issue with the current implementation, where you can form duplicate message loops if the lru cache size is smaller than the mcache. Because nodes can request old messages via iwant and repropagate them. I'll address this in the update as well.

mxinden:
Merged through https://github.com/libp2p/rust-libp2p/pull/1558.

PR #1545, opened 2020-04-07
Spec-compliant Noise Protocol

romanb:
This PR is a proposal for adding a [libp2p-spec compliant](https://github.com/libp2p/specs/pull/260) Noise `Protocol` implementation for `x25519`, called `X25519Spec`. The differences to the existing `X25519` protocol are:

  1. The spec-compliant protocol neither supports `IK` nor `IX`.
  2. The spec-compliant protocol adds the `noise-libp2p-static-key:` prefix for signing and verifying static DH keys, i.e. a domain separator.
  3. The spec-compliant protocol uses just `/noise` as the name for the protocol upgrade.

Please shout if I missed something required for spec compliance.

As an upgrade path, both a `NoiseConfig::xx` based on a `Keypair<X25519Spec>` as well as some other `NoiseConfig` based on a `Keypair<X25519>` could be used with a `SelectUpgrade`.

The protobuf changes are purely cosmetical, since luckily the same protobuf is used for both protocols.

If we eventually _only_ want to support the spec-compliant protocol, we could just replace `X25519Spec -> X25519` together with a few simplifications, like removing `IK` and `IX`.

### DIFF
--- a/protocols/noise/src/protocol/x25519_spec.rs
+++ b/protocols/noise/src/protocol/x25519_spec.rs
@@ -1,0 +1,148 @@
+// Copyright 2019 Parity Technologies (UK) Ltd.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a
+// copy of this software and associated documentation files (the "Software"),
+// to deal in the Software without restriction, including without limitation
+// the rights to use, copy, modify, merge, publish, distribute, sublicense,
+// and/or sell copies of the Software, and to permit persons to whom the
+// Software is furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS
+// OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+// FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+// DEALINGS IN THE SOFTWARE.
+
+//! libp2p-spec compliant Noise protocols based on X25519.
+
+use crate::{NoiseConfig, NoiseError, Protocol, ProtocolParams};
+use libp2p_core::UpgradeInfo;
+use libp2p_core::identity;
+use rand::Rng;
+use x25519_dalek::{X25519_BASEPOINT_BYTES, x25519};
+use zeroize::Zeroize;
+
+use super::{*, x25519::X25519};
+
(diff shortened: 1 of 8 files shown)


romanb:
> If we eventually only want to support the spec-compliant protocol, we could just replace X25519Spec -> X25519 together with a few simplifications, like removing IK and IX.

In that context, should we already `#[deprecate]` the existing `X25519` with this PR to motivate people to upgrade?

protocols/noise/src/protocol/x25519_spec.rs
```diff
@@ -0,0 +1,148 @@
...
+        Keypair { secret, public }
+    }
+}
+
+impl UpgradeInfo for NoiseConfig<XX, X25519Spec> {
+    type Info = &'static [u8];
+    type InfoIter = std::iter::Once<Self::Info>;
+
+    fn protocol_info(&self) -> Self::InfoIter {
+        std::iter::once(b"/noise")
```
tomaka, line 74: Is this the official protocol name?
If so, the fact that it is different from the existing one means that the upgrade path for Substrate/Polkadot will be easy.

tomaka:
> In that context, should we already #[deprecate] the existing X25519 with this PR to motivate people to upgrade?

My opinion is yes. We could also immediately rename `x25519_spec.rs` to `x25519.rs`, and the current one to `x25519_legacy.rs`.

protocols/noise/src/protocol/x25519_spec.rs
romanb: Yes, it's a bit hidden in the spec: https://github.com/libp2p/specs/tree/master/noise#negotiation.

I can only guess that https://github.com/libp2p/rust-libp2p/pull/1481 did not run into a protocol name issue because of doing only prefix matching on protocol names, but I really don't know.

romanb:
> If I understand correctly, we will reach the `unimplemented!()` if we try to combine the specs version with the IK/IX handshakes, but considering that no real decisions exists here as to whether these handshakes will be added, I'm ok with that trade-off.

We essentially have two options going forward:

    1. We discontinue support for IK/IX and only reintroduce them if they are added to the spec. Then any use of `libp2p-noise` will be spec-compliant, i.e. you cannot use it in a way that won't be interoperable with other libp2p implementations.

    2. We continue to support IK/IX, documenting them as non-interoperable with other libp2p implementations (but they may well be used - and be useful - if someone is not interested in interop but only in using rust-libp2p).

I really have no preference. I think in general rust-libp2p, and maybe any libp2p implementation, provide a feature set that is at best a superset of what is in the libp2p specs but more commonly just an intersection. I'm not sure whether an ideal implementation of the specs is not supposed to offer anything beyond the specs (since any such functionality would not be interoperable). Are there any guidelines towards implementation-specific features?

tomaka:
> Are there any guidelines towards implementation-specific features?

Not really as far as I know.

There are already a large number of configuration options that break interoperability if not left to their default value.

To me, the general guideline would be that by default everything is interoperable, and that tutorials/high-level documentation should showcase only interoperable features, but don't restraint ourselves only to them.

romanb:
Ok, then I guess it makes sense to leave support for IK/IX in place also for the new spec-compliant version, just updating the documentation appropriately, and maybe adding a `Default` instance for `NoiseConfig<XX, X25519Spec>`.

tomaka:
Could you add an entry in the CHANGELOG as well?

And then I'd suggest to merge this PR, unless someone else wants to review or comment.

romanb:
I don't really know what the problem is with the `clang-9` installation in one of the CI builds. Any hints welcome.

tomaka:
I think I broke the CI by doing a manual restart.
You can either merge #1565 first and then update this branch, or just push an empty commit.

PR #1548, opened 2020-04-14
protocols/kad/query/peers/closest: Make at_capacity use max 

mxinden:
When not making progress for `parallelism` time a `ClosestPeersIter`
becomes `State::Stalled`. When stalled an iterator is allowed to make
more parallel requests up to `num_results`. If `num_results` is smaller
than `parallelism` make sure to still allow up to `parallelism` requests
in-flight.

### DIFF
--- a/protocols/kad/src/query/peers/closest.rs
+++ b/protocols/kad/src/query/peers/closest.rs
@@ -695,76 +695,27 @@ mod tests {
     }
 
     #[test]
-    fn stalled_iter_at_capacity_use_max_of_parallelism_and_num_results() {
-        let target: KeyBytes = Key::from(PeerId::random()).into();
+    fn stalled_iter_at_capacity_max_of_parallelism_and_num_results() {
+        fn prop(mut iter: ClosestPeersIter) {
+            iter.state = State::Stalled;
 
-        let mut iter = ClosestPeersIter::with_config(
-            ClosestPeersIterConfig {
-                num_results: 4,
-                parallelism: 8,
-                ..ClosestPeersIterConfig::default()
-            },
-            target,
-            std::iter::empty(),
-        );
-        iter.state = State::Stalled;
-        iter.num_waiting = 7; // Smaller than `parallelism`, but larger than `num_results`.
-
-        assert!(!iter.at_capacity());
-    }
-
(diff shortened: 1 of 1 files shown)


protocols/kad/src/query/peers/closest.rs
```diff
@@ -691,4 +693,77 @@ mod tests {
...
+            target,
+            std::iter::empty(),
+        );
+        iter.state = State::Stalled;
+        iter.num_waiting = 7; // Smaller than `parallelism`, but larger than `num_results`.
+
+        assert!(!iter.at_capacity());
+    }
+
+    /// When not making progress for `parallelism` time a [`ClosestPeersIter`] becomes
```
mxinden, line 716: This test is likely a bit over the top. I left it here to showcase the issue in a more in-depth scenario. I would suggest removing it before merging. What do you think?

protocols/kad/src/query/peers/closest.rs
```diff
@@ -691,4 +693,78 @@ mod tests {
 
         QuickCheck::new().tests(10).quickcheck(prop as fn(_) -> _)
     }
+
+    #[test]
+    fn stalled_iter_at_capacity_use_max_of_parallelism_and_num_results() {
```
romanb, line 698: Wouldn't this make for a good property test? The property being the following: "For any `ClosestPeersIter`, if it is in `State::Stalled` then it must be at capacity only if `num_waiting = max{num_results, parallelism}`" ?

protocols/kad/src/query/peers/closest.rs
romanb: Yes, I don't think it adds much to the other test, especially if the other is turned into a property test. Thanks for writing it though.

mxinden:
Property based test is a good idea. Would you mind taking another look @romanb?

protocols/kad/src/query/peers/closest.rs
```diff
@@ -691,4 +693,29 @@ mod tests {
 
         QuickCheck::new().tests(10).quickcheck(prop as fn(_) -> _)
     }
+
+    #[test]
+    fn stalled_iter_at_capacity_max_of_parallelism_and_num_results() {
```
romanb, line 698:
```suggestion
    fn stalled_at_capacity() {
```

PR #1549, opened 2020-04-17
Make sure inject_dial_failure is called

tomaka:
Context: https://github.com/paritytech/substrate/issues/5684

There is an assumption right now in Substrate's code (and probably other) that, after we call `dial` or emit a `DialPeer`, the `inject_dial_failure` method is called if we know that we won't be able to reach the given node.

This is not the case right now in the situation where `addresses_of_peer` returns an empty list. In that situation, we ask the `Swarm` to connect to a node, and don't get any event whatsoever about that node ever again.

This PR modifies `dial` to emit the event if necessary, and simplifies the handling of `DialPeer` to rely more on the behaviour of `dial`.

In generally, I wish it was precisely documented and explained what happens in which situation, and that it is possible in any situation to track the state of each connection attempt, but that's a later change.
While I think it's be reasonable to ask @romanb to handle and review that, he's unfortunately on vacation right now.

### DIFF
--- a/swarm/src/lib.rs
+++ b/swarm/src/lib.rs
@@ -740,7 +740,7 @@ where TBehaviour: NetworkBehaviour<ProtocolsHandler = THandler>,
                                     this.behaviour.inject_dial_failure(&peer_id);
                                 }
                             }
-                        };
+                        }
                     }
                 },
                 Poll::Ready(NetworkBehaviourAction::NotifyHandler { peer_id, handler, event }) => {
--- a/swarm/src/lib.rs
+++ b/swarm/src/lib.rs
@@ -733,8 +733,6 @@ where TBehaviour: NetworkBehaviour<ProtocolsHandler = THandler>,
                             if let Some(mut peer) = this.network.peer(peer_id.clone()).into_dialing() {
                                 let addrs = this.behaviour.addresses_of_peer(peer.id());
                                 peer.connection().add_addresses(addrs);
-                            } else {
-                                this.behaviour.inject_dial_failure(&peer_id);
                             }
                         };
                     }
--- a/swarm/src/lib.rs
+++ b/swarm/src/lib.rs
@@ -730,9 +730,15 @@ where TBehaviour: NetworkBehaviour<ProtocolsHandler = THandler>,
                         } else {
                             log::trace!("Condition for new dialing attempt to {:?} not met: {:?}",
                                 peer_id, condition);
-                            if let Some(mut peer) = this.network.peer(peer_id.clone()).into_dialing() {
-                                let addrs = this.behaviour.addresses_of_peer(peer.id());
-                                peer.connection().add_addresses(addrs);
+                            match this.network.peer(peer_id.clone()) {
+                                Peer::Dialing(mut peer) => {
+                                    let addrs = this.behaviour.addresses_of_peer(peer.id());
+                                    peer.connection().add_addresses(addrs);
+                                },
+                                Peer::Connected(_) => {},
+                                Peer::Disconnected(..) | Peer::Local => {
+                                    this.behaviour.inject_dial_failure(&peer_id);
+                                }
                             }
                         };
                     }
--- a/swarm/src/lib.rs
+++ b/swarm/src/lib.rs
@@ -730,15 +730,9 @@ where TBehaviour: NetworkBehaviour<ProtocolsHandler = THandler>,
                         } else {
                             log::trace!("Condition for new dialing attempt to {:?} not met: {:?}",
                                 peer_id, condition);
-                            match this.network.peer(peer_id.clone()) {
-                                Peer::Dialing(mut peer) => {
-                                    let addrs = this.behaviour.addresses_of_peer(peer.id());
-                                    peer.connection().add_addresses(addrs);
-                                },
-                                Peer::Connected(_) => {},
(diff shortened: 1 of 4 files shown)


tomaka:
I've reverted a change in [the last commit](https://github.com/libp2p/rust-libp2p/pull/1549/commits/c10ef27e0255850339b5fdb182a648cef0b0762c) because it makes the libp2p-kad tests fail.

To me the change is correct: if we emit a `DialPeer` with a `condition` that evaluates to `false`, we should emit an `inject_dial_failure`.
While I think the change is correct, an even better change would be to make this non-ambiguous in the future.

I'm testing whether this fixes https://github.com/paritytech/substrate/issues/5684.

swarm/src/lib.rs
```diff
@@ -701,36 +714,29 @@ where TBehaviour: NetworkBehaviour<ProtocolsHandler = THandler>,
...
+                            log::trace!("Condition for new dialing attempt to {:?} not met: {:?}",
+                                peer_id, condition);
+                            if let Some(mut peer) = this.network.peer(peer_id.clone()).into_dialing() {
+                                let addrs = this.behaviour.addresses_of_peer(peer.id());
+                                peer.connection().add_addresses(addrs);
+                            } else {
                                 this.behaviour.inject_dial_failure(&peer_id);
                             }
-                        }
+                        };
```
twittner, line 737:
```suggestion
                        }
```

tomaka:
> To me the change is correct: if we emit a DialPeer with a condition that evaluates to false, we should emit an inject_dial_failure.
While I think the change is correct, an even better change would be to make this non-ambiguous in the future.

After investigating a bit, I was mistaken. The problem is that my change would call `inject_dial_failure` even if we're already connected to that node.
I pushed another commit that does it properly.

romanb:
> After investigating a bit, I was mistaken. The problem is that my change would call inject_dial_failure even if we're already connected to that node.

Why would that not be correct? I was actually aware of the current behaviour described in this PR w.r.t. `inject_dial_failure` (See the second paragraph of the PR description of https://github.com/libp2p/rust-libp2p/pull/1506), but I thought this was actually pre-existing behaviour, i.e. predating #1440  - am I mistaken and I introduced this behaviour? I was quite sure that e.g. `inject_dial_failure` was never called when `DialPeer` was emitted with the local peer ID, but I may have missed something.

swarm/src/lib.rs
```diff
@@ -701,34 +714,31 @@ where TBehaviour: NetworkBehaviour<ProtocolsHandler = THandler>,
...
+                            log::trace!("Condition for new dialing attempt to {:?} not met: {:?}",
+                                peer_id, condition);
+                            match this.network.peer(peer_id.clone()) {
+                                Peer::Dialing(mut peer) => {
                                     let addrs = this.behaviour.addresses_of_peer(peer.id());
                                     peer.connection().add_addresses(addrs);
+                                },
+                                Peer::Connected(_) => {},
+                                Peer::Disconnected(..) | Peer::Local => {
+                                    this.behaviour.inject_dial_failure(&peer_id);
```
romanb, line 740: I'm really not sure about calling `inject_dial_failure` when the dialing condition _was not met_. I would only expect a call to `inject_dial_failure` if my behaviour emitted `DialPeer` and the dialing condition was actually met, i.e. a new dialing attempt was actually initiated, for which I want to know about the result. In other woirds, I would expect `inject_dial_failure` for every failed dialing attempt, not if no attempt was made. Admittedly, as I wrote in https://github.com/libp2p/rust-libp2p/pull/1506, if a behaviour actually wants to know if the dialing condition is not met because it would otherwise still keep some state, another callback may be needed then.

romanb: I guess it would be ok if we clearly document that `inject_dial_failure` is called also if the dialing  condition is not met, i.e. such that every `DialPeer` is guaranteed to be paired with `inject_connection_established` or `inject_dial_failure`, but then there should be no exception for when the peer is already connected, again because the behaviour may otherwise keep some state around.

tomaka:
> am I mistaken and I introduced this behaviour? I was quite sure that e.g. inject_dial_failure was never called when DialPeer was emitted with the local peer ID, but I may have missed something.

I don't know when, but in libp2p 0.16.2, when `addresses_of_peer` returned an empty list, we would indeed call `inject_dial_failure`:

- https://github.com/libp2p/rust-libp2p/blob/2c40f2880e1ad4d9dd378c5866f04daf81911380/core/src/nodes/network.rs#L1804-L1808
- https://github.com/libp2p/rust-libp2p/blob/2c40f2880e1ad4d9dd378c5866f04daf81911380/swarm/src/lib.rs#L259-L261

swarm/src/lib.rs
tomaka: I can revert that.

romanb:
> > am I mistaken and I introduced this behaviour? I was quite sure that e.g. inject_dial_failure was never called when DialPeer was emitted with the local peer ID, but I may have missed something.
> 
> I don't know when, but in libp2p 0.16.2, when `addresses_of_peer` returned an empty list, we would indeed call `inject_dial_failure`: [..]

I see, thanks for digging that out. This was a regression then, my apologies.

tomaka:
No worries! Is that PR good for you?

Issue #1552, opened 2020-04-20
On a Google Cloud node, listening on an IPv6 address triggers a lot of listener errors


tomaka:
When listening for example on `/ip6/::/tcp/30333`, there seems to be a periodic `ListenerError` whose message is something like:

> ::ffff:10.128.0.9 does not match any listen address

https://github.com/libp2p/rust-libp2p/blob/bbed28b3ec837f0cc24e61b904c3ce04926ee4d1/transports/tcp/src/lib.rs#L469

This only happens on my Google Cloud node, and not on my local machine.

tomaka:
I logged the output of calling the `host_addresses` function:

```
2020-04-20 08:54:53.207 tokio-runtime-worker DEBUG libp2p_tcp  host_addresses = [(V4(127.0.0.1), 127.0.0.1/8, "/ip4/127.0.0.1/tcp/30333"), (V4(10.128.0.9), 10.128.0.9/32, "/ip4/10.128.0.9/tcp/30333"), (V4(172.17.0.1), 172.17.0.1/16, "/ip4/172.17.0.1/tcp/30333"), (V6(::1), ::1/128, "/ip6/::1/tcp/30333")]
... (unrelevant logs here) ...
2020-04-20 08:54:55.611 tokio-runtime-worker DEBUG libp2p_tcp  host_addresses = [(V4(127.0.0.1), 127.0.0.1/8, "/ip4/127.0.0.1/tcp/30333"), (V4(10.128.0.9), 10.128.0.9/32, "/ip4/10.128.0.9/tcp/30333"), (V4(172.17.0.1), 172.17.0.1/16, "/ip4/172.17.0.1/tcp/30333"), (V6(::1), ::1/128, "/ip6/::1/tcp/30333")]
2020-04-20 08:54:55.611 tokio-runtime-worker ERROR sub-libp2p  Error on libp2p listener ListenerId(1): ::ffff:10.128.0.9 does not match any listen address
```

twittner:
#1555 separates IPv4 and IPv6 listeners, so this mixed comparison should not happen anymore.

PR #1555, opened 2020-04-23
tcp: Set IPV6_V6ONLY for IPv6 listeners.

twittner:
The current behaviour of listening on an IPv6 address varies depending on the operating system's IP address stack implementation. Some support IPv4-mapped IPv6 addresses (e.g. Linux and newer versions of Windows) so a single IPv6 address would support IPv4-mapped addresses too. Others do not (e.g. OpenBSD). If they do, then some support them by default (e.g. Linux) and some do not (e.g. Windows).

This PR attempts to implement the same behaviour accross operating systems. The strategy is as follows:

Disable IPv4-mapped IPv6 addresses, hence the socket option IPV6_V6ONLY is always set to true.

This allows binding two sockets to the same port and also avoids the problem of comparing mixed addresses which leads to issues such as #1552.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
   option `IPV6_V6ONLY` is set to true. Instead of relying on IPv4-mapped IPv6
   address support, two listeners can be started if IPv4 and IPv6 should both
   be supported. IPv4 listener addresses are not affected by this change.
+  [PR 1555](https://github.com/libp2p/rust-libp2p/pull/1555)
 
 # Version 0.18.1 (2020-04-17)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@
   initialization. Unless `KademliaConfig::set_replication_factor` is used change
   has no effect.
   [PR 1536](https://github.com/libp2p/rust-libp2p/pull/1536)
+- `libp2p-tcp`: On listeners started with an IPv6 multi-address the socket
+  option `IPV6_V6ONLY` is set to true. Instead of relying on IPv4-mapped IPv6
+  address support, two listeners can be started if IPv4 and IPv6 should both
+  be supported. IPv4 listener addresses are not affected by this change.
 
 # Version 0.18.1 (2020-04-17)
 
--- a/transports/tcp/src/lib.rs
+++ b/transports/tcp/src/lib.rs
@@ -123,10 +123,8 @@ impl Transport for $tcp_config {
             socket.bind(&socket_addr.into())?;
             socket.listen(1024)?; // we may want to make this configurable
 
-            let listener = match <$tcp_listener>::try_from(socket.into_tcp_listener()) {
-                Ok(listener) => listener,
-                Err(e) => return Err(io::Error::new(io::ErrorKind::Other, e))
-            };
+            let listener = <$tcp_listener>::try_from(socket.into_tcp_listener())
+                .map_err(|e| io::Error::new(io::ErrorKind::Other, e))?;
 
             let local_addr = listener.local_addr()?;
             let port = local_addr.port();
--- a/transports/tcp/src/lib.rs
+++ b/transports/tcp/src/lib.rs
@@ -39,8 +39,10 @@ use libp2p_core::{
     transport::{ListenerEvent, TransportError}
 };
 use log::{debug, trace};
+use socket2::{Socket, Domain, Type};
 use std::{
     collections::VecDeque,
+    convert::TryFrom,
     io,
     iter::{self, FromIterator},
     net::{IpAddr, SocketAddr},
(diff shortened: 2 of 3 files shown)


transports/tcp/src/lib.rs
```diff
@@ -108,7 +110,24 @@ impl Transport for $tcp_config {
...
+                s.set_only_v6(true)?;
+                s
+            };
+            if cfg!(target_family = "unix") {
+                socket.set_reuse_address(true)?;
+            }
+            socket.bind(&socket_addr.into())?;
+            socket.listen(1024)?; // we may want to make this configurable
+
+            let listener = match <$tcp_listener>::try_from(socket.into_tcp_listener()) {
```
mxinden, line 126: Nit: Why not use the *try* operator and `map_err` here?

transports/tcp/src/lib.rs
```diff
@@ -108,7 +110,24 @@ impl Transport for $tcp_config {
...
+            } else {
+                let s = Socket::new(Domain::ipv6(), Type::stream(), Some(socket2::Protocol::tcp()))?;
+                s.set_only_v6(true)?;
+                s
+            };
+            if cfg!(target_family = "unix") {
+                socket.set_reuse_address(true)?;
+            }
+            socket.bind(&socket_addr.into())?;
+            socket.listen(1024)?; // we may want to make this configurable
```
tomaka, line 124: The default in Rust seems to be 128
https://github.com/rust-lang/rust/blob/c06e4aca19046b07d952b16e9f002bfab38fde6b/src/libstd/sys/unix/ext/net.rs#L823

```suggestion
            socket.listen(128)?; // we may want to make this configurable
```

mxinden, line 124: As far as I can tell [the libstd default is `128`](https://github.com/rust-lang/rust/blob/master/src/libstd/sys_common/net.rs#L373). Why `1024` in particular here?

transports/tcp/src/lib.rs
```diff
@@ -108,7 +110,24 @@ impl Transport for $tcp_config {
...
-            let listener = <$tcp_listener>::bind(&socket_addr).await?;
+            let socket = if socket_addr.is_ipv4() {
+                Socket::new(Domain::ipv4(), Type::stream(), Some(socket2::Protocol::tcp()))?
+            } else {
+                let s = Socket::new(Domain::ipv6(), Type::stream(), Some(socket2::Protocol::tcp()))?;
+                s.set_only_v6(true)?;
+                s
+            };
+            if cfg!(target_family = "unix") {
+                socket.set_reuse_address(true)?;
```
mxinden, line 121: Does that imply that on non-unix platforms one can not use dualstack nor two listeners (ipv4 and ipv6 on the same port)? (If so, not sure whether it would be important.)

tomaka, line 121: I have no idea of the implications of that change. What are the reasons for that?

transports/tcp/src/lib.rs
twittner: I was following [MIOs](https://crates.io/crates/mio) lead here as it is what tokio and async-std use.

transports/tcp/src/lib.rs
twittner: I do not know why it is restricted to Unix, but again this is what MIO does which in turn does [what libstd does](https://github.com/rust-lang/rust/blob/c06e4aca19046b07d952b16e9f002bfab38fde6b/src/libstd/sys_common/net.rs#L393-L398).

transports/tcp/src/lib.rs
twittner: See [this comment](https://github.com/libp2p/rust-libp2p/pull/1555#discussion_r413873891).

transports/tcp/src/lib.rs
twittner: It depends on the particular platform. On Windows for example [IPV6_V6ONLY seems to be true by default](https://docs.microsoft.com/en-us/windows/win32/winsock/ipproto-ipv6-socket-options) and use of `SO_REUSEADDR` is discouraged if I read [this](https://docs.microsoft.com/en-us/windows/win32/winsock/using-so-reuseaddr-and-so-exclusiveaddruse) correctly, but frankly I do not know Windows and if someone has ideas how to configure things more optimally on this and other non-unix platforms we can make those changes in subsequent PRs.

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@
   initialization. Unless `KademliaConfig::set_replication_factor` is used change
   has no effect.
   [PR 1536](https://github.com/libp2p/rust-libp2p/pull/1536)
+- `libp2p-tcp`: On listeners started with an IPv6 multi-address the socket
+  option `IPV6_V6ONLY` is set to true. Instead of relying on IPv4-mapped IPv6
+  address support, two listeners can be started if IPv4 and IPv6 should both
+  be supported. IPv4 listener addresses are not affected by this change.
```
tomaka, line 10:
```suggestion
  be supported. IPv4 listener addresses are not affected by this change.
  [PR 1555](https://github.com/libp2p/rust-libp2p/pull/1555)
```

PR #1566, opened 2020-05-07
Update to latest multihash.

dvc94ch:
Also fixes an unused warning when compiled with no-default-features


tomaka:
Since the `multihash` crate doesn't have any changelog:
How does it work for the hasher of a certain `multihash` to be always available? Does that mean that every single hash algorithm is now included in the crate? Or do we fail to parse unknown multihashes with unsupported algorithms? Or is there a hidden panic?

dvc94ch:
It hasn't changed, all multihashes have always been included. The only difference is that now digest is called on `Code` directly instead of heap allocating a hasher, which only makes sense if you want to use the `Digest` api.

tomaka:
Could you add an entry in CHANGELOG.md if you don't mind? Since multihash is exposed on the public API, that's a breaking change that should be mentioned

dvc94ch:
Added an entry, let me know if you want it to contain more information.

```diff
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,8 @@
   be supported. IPv4 listener addresses are not affected by this change.
   [PR 1555](https://github.com/libp2p/rust-libp2p/pull/1555)
 
+- `libp2p-core`: Updated to multihash 0.11.0.
+
 # Version 0.18.1 (2020-04-17)
 
 - `libp2p-swarm`: Make sure inject_dial_failure is called in all situations.
```

src/lib.rs
```diff
@@ -299,7 +297,7 @@ pub fn build_tcp_ws_secio_mplex_yamux(keypair: identity::Keypair)
         .authenticate(secio::SecioConfig::new(keypair))
         .multiplex(core::upgrade::SelectUpgrade::new(yamux::Config::default(), mplex::MplexConfig::new()))
         .map(|(peer, muxer), _| (peer, core::muxing::StreamMuxerBox::new(muxer)))
-        .timeout(Duration::from_secs(20)))
+        .timeout(std::time::Duration::from_secs(20)))
```
mxinden, line 300: Nit pick: Seems like we always import `Duration` at the top. Why change that?

dvc94ch: To fix an unused warning when compiled with no-default-features

mxinden: Ah, yes, you mentioned that on the top. Sorry for the noise.

PR #1567, opened 2020-05-07
More insight into Kademlia queries.

romanb:
This is an extended proposal for providing more insight into and control over Kademlia queries, based on earlier discussions at https://github.com/libp2p/rust-libp2p/pull/1494. 

  * More insight: The API allows iterating over the active queries and inspecting their state and execution statistics (`Kademlia::iter_queries` via `QueryRef` and `Kademlia::iter_queries_mut` via `QueryMut`). Both `QueryRef` and `QueryMut` give access to the `QueryInfo` and `QueryStats`. I went a bit back-and-forth about whether to expose `QueryInfo` or not but in the end decided to do so. Any query state that is purely internal can still kept in the internal `QueryInner`, which is not exposed. Whenever a query finishes, successfully or with an error, execution statistics are also reported in the `KademliaEvent`. The `QueryStats` include: Number of requests initiated, number of successful requests, number of failed requests, as well as query duration.

  * More control: The API allows aborting queries prematurely at any time: `Kademlia::query_mut(id)` yields a `QueryMut` that provides `QueryMut::finish()`. This functionality existed before internally, but was not exposed. 

To that end, API operations that initiate new queries return the query ID and multi-phase queries such as `put_record` retain the query ID across all phases, each phase being executed by a new (internal) query.

Tangentially, the behaviour of `Kademlia::bootstrap` has changed (for the better): Previously, once the initial self-lookup finished, queries to refresh all buckets beyond the first non-empty bucket would be initiated all at the same time, now they are initiated one after the other, which is more sane. The bootstrap events reported via a `KademliaEvent` also report the remaining number of bootstrap queries. When that counter reaches 0, bootstrapping is complete. This was not really detectable before.


protocols/kad/src/behaviour.rs
```diff
@@ -476,13 +499,19 @@ where
     ///
     /// > **Note**: Bootstrapping requires at least one node of the DHT to be known.
     /// > See [`Kademlia::add_address`].
-    pub fn bootstrap(&mut self) {
+    pub fn bootstrap(&mut self) -> Result<QueryId, ()> {
```
romanb, line 502: Should probably still get a better error type.

examples/distributed-key-value-store.rs
```diff
@@ -188,7 +199,7 @@ fn handle_input_line(kademlia: &mut Kademlia<MemoryStore>, line: String) {
                 publisher: None,
                 expires: None,
             };
-            kademlia.put_record(record, Quorum::One);
+            kademlia.put_record(record, Quorum::One).expect("Failed to store record locally.");
```
mxinden, line 202: Instead of leaving the interpretation of the error up to the user (in this case the `expect` string, what do you think of having `RecordStore::put` return a proper error? Not saying this should happen within this pull request. I am happy to do that as a follow up.

protocols/kad/src/behaviour.rs
```diff
@@ -286,6 +289,26 @@ where
         }
     }
 
+    /// Gets an iterator over immutable references to all running queries.
```
mxinden, line 292: If I am not mistaken `queries.iter` can as well yield queries that are in `Finished` state, right? I guess that could still count as `running`?

protocols/kad/src/behaviour.rs
```diff
@@ -476,13 +499,19 @@ where
     ///
     /// > **Note**: Bootstrapping requires at least one node of the DHT to be known.
     /// > See [`Kademlia::add_address`].
-    pub fn bootstrap(&mut self) {
+    pub fn bootstrap(&mut self) -> Result<QueryId, ()> {
         let local_key = self.kbuckets.local_key().clone();
-        let info = QueryInfo::Bootstrap { peer: local_key.preimage().clone() };
+        let info = QueryInfo::Bootstrap {
+            peer: local_key.preimage().clone(),
+            remaining: Vec::new().into_iter()
```
mxinden, line 506: Instead of modelling the uninitialized state with an empty vector, I would find an `Option<Vec<_>>::None` easier to understand, what do you think?

protocols/kad/src/behaviour.rs
```diff
@@ -732,72 +770,114 @@ where
...
+                    let info = QueryInfo::Bootstrap {
+                        peer: target.clone().into_preimage(),
+                        remaining
+                    };
+                    let peers = self.kbuckets.closest_keys(&target);
+                    let inner = QueryInner::new(info);
+                    self.queries.continue_iter_closest(query_id, target.clone(), peers, inner);
+                }
+
+                Some(KademliaEvent::QueryResult {
```
mxinden, line 789: Do I understand correctly that instead of a call to `bootstrap` triggering a single `QueryResult` it now triggers one per bootstrap bucket? If so, could you add a note in the changelog?

protocols/kad/src/behaviour.rs
```diff
@@ -732,72 +770,114 @@ where
                                 target = kbucket::Key::new(PeerId::random());
                             }
                             target
-                        }).collect::<Vec<_>>();
+                        }).collect::<Vec<_>>().into_iter();
```
mxinden, line 773: Why not make this iterator lazy delaying the _heavy_ computations to the time they are needed.

protocols/kad/src/query.rs
```diff
@@ -293,6 +335,91 @@ pub struct QueryResult<TInner, TPeers> {
...
+    pub fn pending(&self) -> u32 {
+        self.requests - (self.success + self.failure)
+    }
+
+    /// Gets the duration of the query.
+    ///
+    /// If the query has not yet finished, the duration is measured from the
+    /// start of the query to the current instant.
+    ///
+    /// If the query did not yet start (i.e. yield the first peer to contact),
```
mxinden, line 392: I would prefer returning an `Option<Duration>` to model the case where the query has not yet started. Using `0` to model the uninitialized value seems error prone to me. What do you think?

protocols/kad/src/query/peers/closest.rs
```diff
@@ -211,16 +213,16 @@ impl ClosestPeersIter {
     /// If the iterator is finished, it is not currently waiting for a
```
mxinden, line 213: Can you document the case where `on_failure` returns `false`?

protocols/kad/src/behaviour.rs
romanb: That was already the case before, the only change is in when these queries are started. Previously there were essentially just two stages: 1) start a query to lookup myself 2) start one query per bucket that is further away than the first non-empty bucket. The latter is obviously not so great, apart from the fact that a user doesn't really know when bootstrapping is complete (the events are emitted, but you never know when you got the last one). Now step 2) is made up of N sequential queries, where the query for "refreshing" bucket `n+1` is started only when the query for the "refresh" of bucket `n` has finished.

protocols/kad/src/behaviour.rs
romanb: What is the meaning of `Some(Vec::new())` then? Wouldn't we add another type inhabitant that is never supposed to be instantiated? `Option<NonEmptyVec<_>>` would be a good isomorphic type but I don't think it is worth introducing such a thing here, though I would generally like it if we had non-empty collection types for many purposes.

romanb: Oh, I think I misunderstood. You suggest intentionally distinguishing `Some(Vec::new())` from `None` because the query in this state may be visible to the user. I wrongly remembered this to only be an internal intermediate state. That makes sense, thanks for the suggestion!

protocols/kad/src/behaviour.rs
romanb: True, there can certainly be queries which are already "finished" but have not yet been seen as finished by the `Kademlia` behaviour, since that can only be detected on `Kademlia::poll`. I guess we could filter them out here, what do you think?

examples/distributed-key-value-store.rs
romanb: The interpretation isn't up to the user, the error type is a `store::Error`.

protocols/kad/src/behaviour.rs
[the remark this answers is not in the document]
romanb: Of course, thanks.

protocols/kad/src/query.rs
```diff
@@ -293,6 +335,91 @@ pub struct QueryResult<TInner, TPeers> {
...
+    /// Gets the number of failed requests.
+    pub fn num_failures(&self) -> u32 {
+        self.failure
+    }
+
+    /// Gets the number of pending requests.
+    ///
+    /// > **Note**: A query can finish while still having pending
+    /// > requests, if the termination conditions are already met.
+    pub fn pending(&self) -> u32 {
```
mxinden, line 383:
```suggestion
    pub fn num_pending(&self) -> u32 {
```

Would be consistent with the getters above.

protocols/kad/src/behaviour.rs
romanb: What exactly do you have in mind? Note the following:

  * The absolute maximum size of this vector is 254 elements, but this is _extremely_ unlikely (probabilistically). The realistic sizes are 1-10 elements.
  * There are no heavy computations (at least I don't consider hashing, xor and sampling bytes from a PRNG to be heavy computations).
  * We need to know the length of this vector / iterator immediately below, so if we wanted to leave it as an iterator, it must be an `ExactSizeIterator`.
  * `remaining` is put into the `QueryInfo` for which I'd like to avoid adding type parameters.

Nevertheless, please let me know if you have a concrete approach in mind that is not invasive, maybe I'm missing something!

protocols/kad/src/query.rs
```diff
@@ -293,6 +335,91 @@ pub struct QueryResult<TInner, TPeers> {
...
+        self.requests - (self.success + self.failure)
```
mxinden, line 384:
```
thread 'main' panicked at 'attempt to subtract with overflow', /home/mxinden/.cargo/git/checkouts/rust-libp2p-800b9a
7842b0accc/ae27c37/protocols/kad/src/query.rs:413:9
```

```suggestion
        self.requests.checked_sub(self.success + self.failure).unwrap_or(0)
```

romanb: Why would `success + failure` ever be larger than `requests`? At least that is not the intention, so there may be a bug somewhere.

romanb: Can you provide more context for the panic?

protocols/kad/src/query.rs
romanb: Sounds good.

protocols/kad/src/query/peers/closest.rs
romanb: Will do.

protocols/kad/src/query.rs
mxinden: Sorry for the rather mechanical error report.

I have a guess what is happening: Failing to contact a node always notifies all `ClosestPeersIter`s. Say there are two `ClosestPeersIter`, both query some unavailable node. Thus each iterator `on_failure` will be called twice, increasing its `failures` stat twice instead of once.


By returning false on `PeerState` `Failed` and `Succeeded` one can ensure not to double count failures.

```diff
diff --git a/protocols/kad/src/query/peers/closest.rs b/protocols/kad/src/query/peers/closest.rs
index 78be44d2..a8f0fb13 100644
--- a/protocols/kad/src/query/peers/closest.rs
+++ b/protocols/kad/src/query/peers/closest.rs
@@ -235,8 +235,7 @@ impl ClosestPeersIter {
                 PeerState::Unresponsive => {
                     e.get_mut().state = PeerState::Failed
                 }
-                _ => {}
-            }
+                _ => return false,
         }
 
         true
```

Does this make sense @romanb?

romanb: Yes, that looks like the bug, good catch.

mxinden:
I have [updated](https://github.com/mxinden/kademlia-exporter/pull/1) the [Kademlia exporter](https://github.com/mxinden/kademlia-exporter) with this patch and made it expose the new statistics as metrics. In addition I added a new graph (_Number of hops per query_) to the two (_all-dhts_, _specific dht_) dashboards at kademlia-exporter.max-inden.de.

(It ignores `num_pending` for now.)

protocols/kad/src/behaviour.rs
romanb: I decided to filter finished queries on this API.

romanb:
> Woud you mind mentioning the changes of this pull request in the changelog?

I will update the changelog once the PR has passed review just before merging. That way I don't need adjust the changelog together with ongoing changes resulting from reviews. Just a personal preference.

mxinden:
New version of exporter is deployed on kademlia-exporter.max-inden.de.

tomaka:
I don't really have the time/motivation to dig into the Kademlia code, and I'd trust Max's review here.

mxinden:
This is ready to be merged from my side (just missing the changelog entries).

95% of all get queries on Kusama without disjoint paths involve ~180 failed and ~60 succeeded requests.

image

95% of all get queries on Kusama without disjoint paths involve ~240 failed and ~60 succeeded requests.

image

(`num_{requests, failure, success}` don't add up, keep in mind that these graphs show the 95th percentile for each of the values independently.)

mxinden:
https://github.com/romanb/rust-libp2p/pull/4 should fix the existing merge conflicts.

romanb:
@mxinden Thanks a lot.

PR #1570, opened 2020-05-11
Add a method to create a pre shared key when you have the raw bytes

rklaehn:
This was an oversight of the original impl. Currently the only way to make a PSK is by parsing a PSK file. But sometimes you already have the raw bytes.

The raw bytes are just random data, so there is no harm in having a method to create a PreSharedKey
from the raw bytes.


rklaehn:
Is the failed WASM build actually related to this PR, or a fluke? It seems very unlikely that this would cause a build failure...

PR #1571, opened 2020-05-11
protocols/kad/query/closest: Make quickcheck deterministic

mxinden:
Instead of creating unconstrained random number generators in quickcheck
tests to generate test data, have quickcheck provide a `Seed` to seed
those random number generators and thus make the test execution
deterministic / reproducible.


mxinden:
Thanks for the review. I will merge later today to give everyone interested a chance to take a look (not implying that this commit is in any way particularly interesting).

PR #1573, opened 2020-05-13
protocols/kad: Remove unnecessary PeerId conversion for fixed iter

mxinden:
`FixedPeersIter` requires the initial set of peers to be passed as
`PeerId`s and not as `Key<PeerId>`s. This commit removes the unnecessary
conversion.

(*I would expect rustc or llvm to optimize this away anyways. I find it still worth
 it to reduce complexity.*)

### DIFF
--- a/protocols/kad/src/query/peers/fixed.rs
+++ b/protocols/kad/src/query/peers/fixed.rs
@@ -57,7 +57,12 @@ enum PeerState {
 }
 
 impl FixedPeersIter {
-    pub fn new(peers: Vec<PeerId>, parallelism: usize) -> Self {
+    pub fn new<I>(peers: I, parallelism: usize) -> Self
+    where
+        I: IntoIterator<Item = PeerId>
+    {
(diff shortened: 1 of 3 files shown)


protocols/kad/src/query/peers/fixed.rs
```diff
@@ -57,7 +57,12 @@ enum PeerState {
 }
 
 impl FixedPeersIter {
-    pub fn new(peers: Vec<PeerId>, parallelism: usize) -> Self {
+    pub fn new<I>(peers: I, parallelism: usize) -> Self
+    where
+        I: IntoIterator<Item = PeerId>
+    {
+        let peers = peers.into_iter().collect::<Vec<_>>();
```
mxinden, line 64: Happy for alternative suggestions. Using dynamic dispatching requires us to set `Send` and `Sync` bounds.

Doing the `collect` here instead of within `query.rs` moves the *hack* closer to its usage.

This might be optimized away by the compiler as well.

PR #1574, opened 2020-05-14
Make the number of events buffered to/from tasks configurable

tomaka:
cc https://github.com/paritytech/substrate/issues/6009

Hints seem to indicate that we have some CPU issues in Substrate from the low number of events that can be buffered. This PR makes this limit configurable.

Additionally, this PR increases these values: the channel size from tasks to the swarm is now 32 (instead of 1!), and the channel size from the swarm to each task is now 7 (so 8 in practice).

When reviewing, keep in mind that channels reserve an extra element for each active sender, as documented: https://docs.rs/futures/0.3.5/futures/channel/mpsc/fn.channel.html

(I also slightly fixed the CHANGELOG to be alphabetically-ordered)

### DIFF
--- a/core/src/network.rs
+++ b/core/src/network.rs
@@ -644,7 +644,7 @@ impl NetworkConfig {
     /// longer be able to deliver events to the associated `ConnectionHandler`,
     /// thus exerting back-pressure on the connection and peer API.
     pub fn set_notify_handler_buffer_size(&mut self, n: NonZeroUsize) -> &mut Self {
-        self.manager_config.to_task_channel_size = n.get() - 1;
+        self.manager_config.task_command_buffer_size = n.get() - 1;
         self
     }
 
@@ -653,9 +653,9 @@ impl NetworkConfig {
     ///
     /// When the buffer is full, the background tasks of all connections will stall.
     /// In this way, the consumers of network events exert back-pressure on
-    /// the network connection I/O. 
+    /// the network connection I/O.
     pub fn set_connection_event_buffer_size(&mut self, n: usize) -> &mut Self {
-        self.manager_config.from_task_channel_size = n;
+        self.manager_config.task_event_buffer_size = n;
         self
     }
 
--- a/core/src/network.rs
+++ b/core/src/network.rs
@@ -605,11 +605,9 @@ pub struct NetworkInfo {
 /// `notify_handler` buffer size of 8.
 #[derive(Default)]
 pub struct NetworkConfig {
-    /// Note that `ManagerConfig` doesn't contain the number of buffered
-    /// events, but rather the size of the channel (which is an
-    /// implementation detail). When it comes to the "to task buffer events"
-    /// configuration option, the size of the channel is the number of
-    /// buffered events minus one.
+    /// Note that the `ManagerConfig`s task command buffer always provides
+    /// one "free" slot per task. Thus the given total `notify_handler_buffer_size`
+    /// exposed for configuration on the `Network` is reduced by one.
     manager_config: ManagerConfig,
     pool_limits: PoolLimits,
 }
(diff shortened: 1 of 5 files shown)


swarm/src/lib.rs
```diff
@@ -1002,6 +1002,48 @@ where TBehaviour: NetworkBehaviour,
...
+    /// Configures the number of events from the [`NetworkBehaviour`] in
+    /// destination to the [`ProtocolsHandler`] that can be buffered before
+    /// the [`Swarm`] has to wait. An individual buffer with this number of
+    /// events exists for each individual connection.
+    ///
+    /// The ideal value depends on the executor used, the CPU speed, and the
+    /// volume of events. If this value is too low, then the [`Swarm`] will
+    /// be sleeping more often than necessary. Increasing this value increases
+    /// the overall memory usage.
+    pub fn to_task_buffered_events(mut self, n: NonZeroUsize) -> Self {
```
romanb, line 1014:
```suggestion
    pub fn notify_handler_buffer_size(mut self, n: NonZeroUsize) -> Self {
```

swarm/src/lib.rs
```diff
@@ -1002,6 +1002,48 @@ where TBehaviour: NetworkBehaviour,
...
+    /// first-served basis.
+    ///
+    /// The ideal value depends on the executor used, the CPU speed, the
+    /// average number of connections, and the volume of events. If this value
+    /// is too low, then the [`ProtocolsHandler`]s will be sleeping more often
+    /// than necessary. Increasing this value increases the overall memory
+    /// usage, and more importantly the latency between the moment when an
+    /// event is emitted and the moment when it is received by the
+    /// [`NetworkBehaviour`].
+    pub fn from_task_extra_buffered_events(mut self, n: usize) -> Self {
```
romanb, line 1042:
```suggestion
    pub fn connection_event_buffer_size(mut self, n: usize) -> Self {
```

core/src/network.rs
```diff
@@ -598,17 +600,23 @@ pub struct NetworkInfo {
 
 /// The (optional) configuration for a [`Network`].
 ///
-/// The default configuration specifies no dedicated task executor
-/// and no connection limits.
+/// The default configuration specifies no dedicated task executor, no
+/// connection limits, a "from task" extra buffer size of 32, and a
+/// "to task" buffer size of 8.
```
romanb, line 605:
```suggestion
/// `notify_handler` buffer size of 8.
```

core/src/network.rs
```diff
@@ -598,17 +600,23 @@ pub struct NetworkInfo {
...
 #[derive(Default)]
 pub struct NetworkConfig {
-    executor: Option<Box<dyn Executor + Send>>,
+    /// Note that `ManagerConfig` doesn't contain the number of buffered
+    /// events, but rather the size of the channel (which is an
+    /// implementation detail). When it comes to the "to task buffer events"
+    /// configuration option, the size of the channel is the number of
+    /// buffered events minus one.
```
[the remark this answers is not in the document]
tomaka: The important information in that comment is that `manager_config.to_task_channel_size` (or `task_command_buffer_size`) is equal to the configured option minus 1.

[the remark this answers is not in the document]
romanb: Ok, if you deem it worthwhile mentioning here, no objections.

core/src/network.rs
```diff
@@ -598,17 +600,23 @@ pub struct NetworkInfo {
...
+/// connection limits, a connection event buffer size of 32, and a
+/// `notify_handler` buffer size of 8.
 #[derive(Default)]
 pub struct NetworkConfig {
-    executor: Option<Box<dyn Executor + Send>>,
+    /// Note that `ManagerConfig` doesn't contain the number of buffered
+    /// events, but rather the size of the channel (which is an
+    /// implementation detail). When it comes to the "to task buffer events"
+    /// configuration option, the size of the channel is the number of
+    /// buffered events minus one.
```
romanb, line 612:
```suggestion
    /// Note that the `ManagerConfig`s task command buffer always provides
    /// one "free" slot per task. Thus the given total `notify_handler_buffer_size`
    /// exposed for configuration on the `Network` is reduced by one.
```

PR #1575, opened 2020-05-14
Support /dns protocol in multiaddr

mzabaluev:
The `/dns` protocol has been added to the spec and has had a de-facto meaning for years (citing https://github.com/multiformats/multiaddr/pull/100)

This adds address parsing and encoding support for `/dns` to the multiaddr library.

### DIFF
--- a/misc/multiaddr/src/from_url.rs
+++ b/misc/multiaddr/src/from_url.rs
@@ -70,7 +70,7 @@ fn from_url_inner_http_ws(url: url::Url, lossy: bool) -> std::result::Result<Mul
         if let Ok(ip) = hostname.parse::<IpAddr>() {
             Protocol::from(ip)
         } else {
-            Protocol::Dns4(hostname.into())
+            Protocol::Dns(hostname.into())
         }
     } else {
         return Err(FromUrlErr::BadUrl);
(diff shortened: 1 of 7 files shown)


mzabaluev:
Address resolution for this is not yet implemented in libp2p, but other users of the `parity-multiaddr` crate could make use of the protocol component in their own ways, e.g. translating a `/dns` multiaddr to an URL to give to a Happy Eyeballs capable HTTP stack.

tomaka:
Support for it is probably very easy to add to `libp2p-dns`.
I don't want to lose track of aa semi-implemented feature, and I don't think it's worth opening an issue just for this, so support should probably be added to this PR, or whoever merges this should quickly submit a PR afterwards.

mzabaluev:
OK, I will try to do here shortly.

mzabaluev:
@tomaka I have added `/dns` support to libp2p-dns and a few other places. I get a unit test failure in libp2p-dns, but the same test is also failing in master.

Commits c02bf8d and possibly f76331f may be too disruptive, as they change the behavior of URL-to-multiaddr translations, albeit improving it with regard to IPv6 support.

misc/multiaddr/src/from_url.rs
```diff
@@ -70,7 +70,7 @@ fn from_url_inner_http_ws(url: url::Url, lossy: bool) -> std::result::Result<Mul
         if let Ok(ip) = hostname.parse::<IpAddr>() {
             Protocol::from(ip)
         } else {
-            Protocol::Dns4(hostname.into())
+            Protocol::Dns(hostname.into())
```
tomaka, line 73: Oh, that's actually a great change :tada:

transports/websocket/src/framed.rs
```diff
@@ -371,7 +373,7 @@ fn location_to_multiaddr<T>(location: &str) -> Result<Multiaddr, Error<T>> {
             let mut a = Multiaddr::empty();
             match url.host() {
                 Some(url::Host::Domain(h)) => {
-                    a.push(Protocol::Dns4(h.into()))
+                    a.push(Protocol::Dns(h.into()))
```
tomaka, line 376: Same

PR #1580, opened 2020-05-18
Readd lost noise upgrades.

romanb:
This readds the (non-standardised/specced) `UpgradeInfo` implementations for `IK` and `IX` to `X25519Spec` which got lost in the back-and-forth of https://github.com/libp2p/rust-libp2p/pull/1545. It also enables the `async-std` feature for the `libp2p-tcp` dependency in the tests, so that the `libp2p-noise` tests compile again on their own.

### DIFF
--- a/protocols/noise/Cargo.toml
+++ b/protocols/noise/Cargo.toml
@@ -28,7 +28,7 @@ snow = { version = "0.6.1", features = ["default-resolver"], default-features = 
 
 [dev-dependencies]
 env_logger = "0.7.1"
-libp2p-tcp = { version = "0.19.0", path = "../../transports/tcp" }
+libp2p-tcp = { version = "0.19.0", path = "../../transports/tcp", features = ["async-std"] }
 quickcheck = "0.9.0"
 sodiumoxide = "^0.2.5"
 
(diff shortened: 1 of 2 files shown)


romanb:
Any volunteers? Shouldn't be controversial.

tomaka:
> Shouldn't be controversial.

Well it kind of is, since the comment saying that this isn't standard is easy to miss.
If this PR had been in when I upgraded Substrate to libp2p 0.19, we would probably accidentally be using a non-standard noise right now.

romanb:
If there was something controversial about continuing to support other handshake patterns, that was the topic of https://github.com/libp2p/rust-libp2p/pull/1545 - this PR just corrects an oversight in that PR, where I understood that you're ok with it. I don't think we (or other libp2p implementations) have limited ourselves strictly to what is defined in the specs before, so I think it is unproblematic to support features beyond those specs. I'm ok with changing that direction, but then it would need to be enforced much more strictly that anything that isn't specced and requires support from all nodes participating in a protocol mustn't be implemented due to the possibility of incompatibility.

tomaka:
I'm not familiar enough with the general state of the libp2p specs regarding noise to have an actual opinion on this, I'm just trusting you here.

PR #1581, opened 2020-05-18
Add get providers example.

dvc94ch:
Adds an example


dvc94ch:
Feel free to close if you don't think it's a worthwhile addition. Just thought it might be useful to someone

tomaka:
The problem is not really that it's not worth adding example, the problem is that our examples are quite a mess. There's a lot of duplicate code involved for example. I didn't answer because I first wanted to look into what can be done to fix that, but I don't have enough time to really dig into it.

PR #1583, opened 2020-05-19
Gossipsub message signing

AgeManning:
## Description

This adds optional message signing to the gossipsub protocol as per the libp2p specifications.

This also removes the `LruCache` `received` cache and simply uses the `memcache` in it's place.


burdges:
As an aside, we've gossip messages that benefit from exotic signatures, like VRFs for approval assignments (A&V) or maybe even aggregation tricks in GRANDPA.  

It's normally fastest if gossip messages run over authenticated channels that we drop on harmless but incorrect and spammy messages, but then do their own authentication, probably polite, possibly weird, before being sent back out.  It's also bad for layering, auditing, etc. to do this.  shrug

protocols/gossipsub/src/protocol.rs
```diff
@@ -319,6 +400,76 @@ impl Decoder for GossipsubCodec {
     }
 }
 
+/// Signs a gossipsub message, adding the key and signature fields.
+fn sign_message(keypair: &Keypair, message: &mut rpc_proto::Message) -> Result<(), io::Error> {
```
mxinden, line 404: TODO mxinden: Why a function and not a method?

protocols/gossipsub/src/protocol.rs
```diff
@@ -112,6 +128,46 @@ where
...
+        keypair: Option<Keypair>,
+        allow_unsigned: bool,
+    ) -> Self {
+        // determine if the public key needs to be included in each message
+        let key = {
+            if let Some(keypair) = keypair.as_ref() {
+                // signing is required
+                let key_enc = keypair.public().into_protobuf_encoding();
+                if key_enc.len() <= 42 {
+                    // public key can be inlined, so we don't include it in the protobuf
```
mxinden, line 152:
```suggestion
                    // public key can be inlined in [`rpc_proto::Message::from`], so we don't include it in the protobuf
```

protocols/gossipsub/src/protocol.rs
```diff
@@ -132,9 +188,20 @@ impl Encoder for GossipsubCodec {
...
+                signature: None,
+                key: None,
             })
             .collect::<Vec<_>>();
 
+        // sign the messages if required
+        if let Some(keypair) = self.keypair.as_ref() {
+            for message in publish.iter_mut() {
+                sign_message(keypair, message)?;
+                // add the peer_id key if required
```
mxinden, line 200:
```suggestion
                // add the public key if not already inlined via the peer id in [`rpc_proto::Message::from`]
```

protocols/gossipsub/src/protocol.rs
```diff
@@ -222,9 +289,22 @@ impl Decoder for GossipsubCodec {
         let rpc = rpc_proto::Rpc::decode(&packet[..])?;
 
         let mut messages = Vec::with_capacity(rpc.publish.len());
-        for publish in rpc.publish.into_iter() {
+        for message in rpc.publish.into_iter() {
+            // verify message signatures if required
+            if !self.allow_unsigned {
```
mxinden, line 294: Do I understand correctly that when `allow_unsigned` is `true` we also allow invalid signatures?

protocols/gossipsub/src/protocol.rs
```diff
@@ -319,6 +400,76 @@ impl Decoder for GossipsubCodec {
     }
 }
 
+/// Signs a gossipsub message, adding the key and signature fields.
```
mxinden, line 403:
```suggestion
/// Signs a gossipsub message, adding [`rpc_proto::Message::signature`] field.
```

protocols/gossipsub/src/protocol.rs
```diff
@@ -319,6 +400,76 @@ impl Decoder for GossipsubCodec {
...
+fn sign_message(keypair: &Keypair, message: &mut rpc_proto::Message) -> Result<(), io::Error> {
+    let mut buf = Vec::with_capacity(message.encoded_len());
+    message
+        .encode(&mut buf)
+        .expect("Buffer has sufficient capacity");
+    // the signature is over the bytes "libp2p-pubsub:<protobuf-message>"
+    let mut signature_bytes = b"libp2p-pubsub:".to_vec();
```
mxinden, line 410: Out of curiosity, what is the purpose of prefixing the signature with `libp2p-pubsub`? I can't find an explanation in the [spec](https://github.com/libp2p/specs/tree/master/pubsub#message-signing).

protocols/gossipsub/src/protocol.rs
```diff
@@ -397,3 +548,51 @@ pub enum GossipsubControlAction {
         topic_hash: TopicHash,
     },
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn sign_and_verify_message_peer_inline() {
```
mxinden, line 557: I think this could greatly benefit of random test input generation e.g. via quickcheck. I am happy to provide a commit in case you agree.

protocols/gossipsub/src/rpc.proto
```diff
@@ -19,6 +19,8 @@ message Message {
 	optional bytes data = 2;
 	optional bytes seqno = 3;
 	repeated string topic_ids = 4;
+    optional bytes signature = 5;
```
mxinden, line 22: Formatting.

protocols/gossipsub/src/config.rs
```diff
@@ -73,9 +73,16 @@ pub struct GossipsubConfig {
...
     /// allows a user to validate the messages before propagating them to their peers. If set to
     /// true, the user must manually call `propagate_message()` on the behaviour to forward message
-    /// once validated (default is false).
+    /// once validated (default is `false`).
     pub manual_propagation: bool,
 
+    /// When set to `true` all published messages are signed by the libp2p key (default is `true`).
+    pub sign_messages: bool,
+
+    /// Determines whether unsigned messages will be accepted. If set to false, unsigned messages
```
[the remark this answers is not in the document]
AgeManning: Yes. There is a config in go, around whether to use message signing or not. 

I put this in, to mimic the current implementation which completely ignores the signature field. I was concerned with interoperability with other implementations and wanted this maximally configurable such that it could accept messages with other implementations that were not signing messages, even if we were. 

However, now that you have pointed this out, perhaps this will cause more trouble than it saves in practice and we should just have a boolean that dictates we either sign/verify or don't sign/verify.

protocols/gossipsub/src/config.rs
```diff
@@ -182,48 +199,84 @@ impl GossipsubConfigBuilder {
...
+    /// The maximum byte size for each gossip (default is 2048 bytes).
     pub fn max_transmit_size(&mut self, max_transmit_size: usize) -> &mut Self {
         self.config.max_transmit_size = max_transmit_size;
         self
     }
 
-    pub fn hash_topics(&mut self) -> &mut Self {
-        self.config.hash_topics = true;
+    /// Flag determining if gossipsub topics are hashed or sent as plain strings (default is false).
+    pub fn hash_topics(&mut self, value: bool) -> &mut Self {
```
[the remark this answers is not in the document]
AgeManning: I agree. I preferred the original version also, but had been playing with other builder implementations and somehow convinced myself that the boolean signature was a more canonical pattern in the rust community (for some reason I thought tokio used this pattern, on inspection some other packages do, like hyper for example: https://docs.rs/hyper/0.13.5/hyper/server/struct.Builder.html). 

I also prefer the previous implementation, so its good to know someone else thinks so. 
Will change back.

protocols/gossipsub/src/mcache.rs
```diff
@@ -72,17 +72,22 @@ impl MessageCache {
...
             mid: message_id.clone(),
             topics: msg.topics.clone(),
         };
 
-        self.msgs.insert(message_id, msg);
-
-        self.history[0].push(cache_entry);
+        let seen_message = self.msgs.insert(message_id, msg);
+        if seen_message.is_none() {
+            // don't add duplicates entries to the cache
```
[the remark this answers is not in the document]
AgeManning: Yep, your're right, will change

protocols/gossipsub/src/protocol.rs
AgeManning: To save you some time. Originally these were inlined, I pulled them out to their own logic functions and they were easier to test. 

There's no real reason why they are not methods, except that the verify message function requires no struct fields of the codec. 

None the less, I've re-written these as methods, because, why not :)

protocols/gossipsub/src/protocol.rs
AgeManning: Neither could I. 
Just speculating, but it ensures this signature can only be used in this protocol I guess. If some other protocol where using the same protobuf format, the signature would be invalid unless those protocols also required this string prefix.

protocols/gossipsub/src/protocol.rs
AgeManning: Yeah I agree, I'm about to go off and build out 1.1. If you have the time to add this in, it would be great

AgeManning:
Thanks for the quick review @mxinden. 

On closer inspection of your comments, I've realised I've made a few pretty significant errors. 

Firstly, I was signing all messages and didn't consider forwarded messages, which obviously we shouldn't sign. There are two ways to address this imo:
1. Perform the signing at the behaviour level. This is nice because we could sign messages in the `publish_many()` function and prevent the cloning of `keypair` in the encoding as well as a few other variables. The downside of this is that the signing requires the protobuf encoding of the messages which is mainly dealt with in `protocol.rs`. For this reason, I went with (2) (also because it required less code changes). 
2. Check outgoing messages. If the source_id is us, then we should be signing it (provided no signature already exists). 

This logic, also brings to light a vulnerability in using `allow_unsigned_messages`. If a peer sent us a message using us as the source id, and we ignored the signature, then signed the message, a malicious peer could get us to sign arbitrary messages. Therefore, I've removed that option, and signing and verification is binary, we either sign and verify, or we don't. 

If we are not signing, we also don't bother checking the signatures (even if one exists). I prefer this functionality, somewhat selfishly because in my use, I'm not signing and don't care about signatures and don't want to waste any computation checking signatures that I don't care about, regardless if they are valid or not.

protocols/gossipsub/src/protocol.rs
thomaseizinger: The go implementations does add this prefix: https://github.com/libp2p/go-libp2p-pubsub/blob/49274b0e8aecdf6cad59d768e5702ff00aa48488/sign.go#L12

protocols/gossipsub/src/protocol.rs
AgeManning: Correct, I've left a message about this below

protocols/gossipsub/src/protocol.rs
AgeManning: Yep, it's in the spec, so they wouldn't be following the spec if they didn't.

thomaseizinger: Ah sorry, I misread the initial comment. Don't mind me :grimacing:

mxinden:
Do I understand correctly that currently if a node is signing messages `disable_message_signing == false` there is no way to allow unsigned messages from other nodes?

Just checking for now, haven't made up my mind whether that is good or not.

protocols/gossipsub/src/protocol.rs
mxinden: @AgeManning see https://github.com/sigp/rust-libp2p/pull/31.

AgeManning:
> Do I understand correctly that currently if a node is signing messages (`disable_message_signing == false`) there is no way to allow unsigned messages from other nodes?

I changed it to this, yes. I originally allowed the other behaviour for debugging. After a bit of thought, I decided this current logic was best. Signing is either enabled or not. The negative case can accept and read signed messages and will propagate them but the affirmative case rejects all unsigned/invalid messages.

The reason was to try and avoid networks of mixed cases. If a node wants to enable signing, I think it wouldn't be great that it propagates unsigned messages also. 

I haven't completed the scoring in 1.1 in relation to how a peer would score another peer that sent unsigned messages. But if in a network where peers expect signed messages, and one peer was not signing and another peer was forwarding unsigned messages (i.e we enabled reading unsigned messages) I think both peers would be scored negatively (need to confirm this). 

We could add flags that allow reading of unsigned messages without propagation and then with propagation, but the complexity grows and I'm not entirely sure we gain much from this added complexity of allowing mixed networks. 

For these reasons I decided to leave it binary. I guess the correct logic would be to follow other implementations. If other clients support it, we should do. From memory the go-client has it binary, but perhaps there is a flag that allows it, will need to check.

I don't use the signing part of gossipsub, so if others have some use for mixed networks, I'm happy to allow it add the flags in, but we might end up with more issues.

protocols/gossipsub/src/behaviour.rs
```diff
@@ -83,28 +84,59 @@ pub struct Gossipsub {
...
 impl Gossipsub {
     /// Creates a `Gossipsub` struct given a set of parameters specified by `gs_config`.
-    pub fn new(local_peer_id: PeerId, gs_config: GossipsubConfig) -> Self {
-        let local_peer_id = if gs_config.no_source_id {
+    pub fn new(keypair: Keypair, gs_config: GossipsubConfig) -> Self {
+        // Set up the router given the configuration settings.
+
+        // Sets up the source_id of published messages.
+        let message_source_id = if gs_config.no_source_id {
             PeerId::from_bytes(crate::config::IDENTITY_SOURCE.to_vec()).expect("Valid peer id")
```
mxinden, line 102: I can't find the concept of the identity source id in the specification. Could you point me to it?

protocols/gossipsub/src/config.rs
```diff
@@ -73,9 +73,13 @@ pub struct GossipsubConfig {
     /// When set to `true`, prevents automatic forwarding of all received messages. This setting
     /// allows a user to validate the messages before propagating them to their peers. If set to
     /// true, the user must manually call `propagate_message()` on the behaviour to forward message
-    /// once validated (default is false).
+    /// once validated (default is `false`).
     pub manual_propagation: bool,
 
+    /// Message signing is on by default.  When this parameter is set,
+    /// published messages are not signed by the libp2p key.
+    pub disable_message_signing: bool,
```
mxinden, line 81: Do I understand correctly that one has to pass a keypair to `GossipSub::new` even when setting `disable_message_signing` to `true`?

Wouldn't something along the following lines work as well?

```suggestion
    pub enum Signing {
        Enabled(Keypair),
        Disabled(PeerId),
    }
    
    // ...

    pub disable_message_signing: Signing,
```

protocols/gossipsub/src/behaviour.rs
```diff
@@ -83,28 +84,59 @@ pub struct Gossipsub {
...
+            None
+        } else {
+            Some(keypair)
+        };
+
+        // If we are signing and the public key cannot be inlined in the `PeerId`, we store the
+        // public key to attach to published messages.
+        let key = {
+            if let Some(keypair) = keypair.as_ref() {
+                // signing is required
```
mxinden, line 118:
```suggestion
                // Signing is required.
```

protocols/gossipsub/src/behaviour.rs
```diff
@@ -83,28 +84,59 @@ pub struct Gossipsub {
...
+                let key_enc = keypair.public().into_protobuf_encoding();
+                if key_enc.len() <= 42 {
+                    // The public key can be inlined in [`rpc_proto::Message::from`], so we don't include it
+                    // specifically in the [`rpc_proto::Message::key`] field.
+                    None
+                } else {
+                    // Include the protobuf encoding of the public key in the message
```
mxinden, line 125:
```suggestion
                    // Include the protobuf encoding of the public key in the message.
```

protocols/gossipsub/src/behaviour.rs
```diff
@@ -83,28 +84,59 @@ pub struct Gossipsub {
...
+                    Some(key_enc)
+                }
+            } else {
+                // Signing is not required, no key needs to be added
```
mxinden, line 129:
```suggestion
                // Signing is not required, no key needs to be added.
```

protocols/gossipsub/src/behaviour.rs
```diff
@@ -222,34 +256,45 @@ impl Gossipsub {
...
-            topics: topic.into_iter().map(|t| self.topic_hash(t)).collect(),
+        let message = match self.build_message(
+            topics.into_iter().map(|t| self.topic_hash(t)).collect(),
+            data.into(),
+        ) {
+            Ok(m) => m,
+            Err(e) => {
+                error!("{}", e);
+                return;
+            }
```
mxinden, line 270: Why not have `publish_many` return a `Result<(), PublishManyError>` and then do:

```suggestion
        let message = self.build_message(
            topics.into_iter().map(|t| self.topic_hash(t)).collect(),
            data.into(),
        )?;
```

As this is a library I think it is better to have the caller handle an error.

protocols/gossipsub/src/behaviour.rs
```diff
@@ -222,34 +256,45 @@ impl Gossipsub {
...
+            );
+            return;
+        }
+
+        debug!("Publishing message: {:?}", msg_id);
 
-        // forward the message to mesh peers
-        let local_peer_id = self.local_peer_id.clone();
-        self.forward_msg(message.clone(), &local_peer_id);
+        // Forward the message to mesh peers
```
mxinden, line 287:
```suggestion
        // Forward the message to mesh peers.
```

protocols/gossipsub/src/behaviour.rs
```diff
@@ -222,34 +256,45 @@ impl Gossipsub {
...
+        let message_source = &self.message_source_id.clone();
+        self.forward_msg(message.clone(), message_source);
 
         let mut recipient_peers = HashSet::new();
         for topic_hash in &message.topics {
-            // if not subscribed to the topic, use fanout peers
+            // If not subscribed to the topic, use fanout peers
```
mxinden, line 293:
```suggestion
            // If not subscribed to the topic, use fanout peers.
```

protocols/gossipsub/src/behaviour.rs
```diff
@@ -921,16 +963,63 @@ impl Gossipsub {
...
         }
         debug!("Completed forwarding message");
     }
 
+    /// Constructs a `GossipsubMessage` performing message signing if required.
+    pub(crate) fn build_message(
+        &self,
+        topics: Vec<TopicHash>,
+        data: Vec<u8>,
+    ) -> Result<GossipsubMessage, String> {
```
mxinden, line 982:
```suggestion
    ) -> Result<GossipsubMessage, SigningError> {
```

While not opinionated on which error handling pattern to use I would prefer not to pass `String`s around as one can not programmatically react on them.

protocols/gossipsub/src/behaviour.rs
```diff
@@ -921,16 +963,63 @@ impl Gossipsub {
...
+            from: Some(self.message_source_id.clone().into_bytes()),
+            data: Some(data),
+            seqno: Some(sequence_number.to_be_bytes().to_vec()),
+            topic_ids: topics.clone().into_iter().map(|t| t.into()).collect(),
+            signature: None,
+            key: None,
+        };
+
+        // If a signature is required, generate it
+        let signature = if let Some(keypair) = self.keypair.as_ref() {
```
mxinden, line 994: `message` is only needed when `Some(keypair) == self.keypair.as_ref()` right?

```suggestion
        // If a signature is required, generate it
        let signature = if let Some(keypair) = self.keypair.as_ref() {
              let message = rpc_proto::Message {
                from: Some(self.message_source_id.clone().into_bytes()),
                data: Some(data),
                seqno: Some(sequence_number.to_be_bytes().to_vec()),
                topic_ids: topics.clone().into_iter().map(|t| t.into()).collect(),
                signature: None,
                key: None,
            };
```

protocols/gossipsub/src/mcache.rs
```diff
@@ -72,17 +72,22 @@ impl MessageCache {
...
             mid: message_id.clone(),
             topics: msg.topics.clone(),
         };
 
-        self.msgs.insert(message_id, msg);
-
-        self.history[0].push(cache_entry);
+        let seen_message = self.msgs.insert(message_id, msg);
+        if seen_message.is_none() {
+            // Don't add duplicate entries to the cache
```
mxinden, line 87:
```suggestion
            // Don't add duplicate entries to the cache.
```

protocols/gossipsub/src/config.rs
AgeManning: This is interesting. You are correct in that currently we unnecessarily require the keypair. 
I'm happy to go with whatever you think will be the best ergonomics to use this. 

If we create a `Signing` enum in the config, the `config()` will no longer implement `Default` and a user will be required to enter a `Signing` variant during the config initialisation. 

Alternatively, I could take signing out of the config, and take an Option<Keypair> when creating the main struct, however this may not be intuitive when initialising the struct that you are optionally invoking signing. 

Unless I'm mistaken you are suggesting something like:
```rust
let config = GossipsubConfig::new(Signing::Enabled(keypair));
Gossipsub::new(config);
```

Which seems fine. I'll add it in and we can change later if we need.

protocols/gossipsub/src/behaviour.rs
AgeManning: Agreed

protocols/gossipsub/src/behaviour.rs
AgeManning: This is true, but if i recall correctly, I was trying to avoid the clone on data. 
I'll change it, but it now clones the data.

AgeManning:
I finally got back around to this. 

I think I've addressed all your comments, I've left comments on ones that I think require them. 

We've since noticed a bug in gossipsub where subscriptions were not being sent to all peers. I've added that correction in too, whilst I'm here.

protocols/gossipsub/src/behaviour.rs
AgeManning: This was something I added as extra pseudo-privacy. Since this addition go has introduced an "author" which can be modified. 

This has since been removed in favour of the `Signing` enum, which when sent to disabled gives you the ability to set the `author`. 

The identity was just an arbitrary peer ID i was using, which is easily identifiable.

protocols/gossipsub/src/config.rs
AgeManning: After writing this, and attempting to integrate the new version, I realise it's really kind of an annoyance to not have `Default` implemented for the configuration, or to not allow the configuration be instantiated without a `PeerId` or `Keypair`.

The two solutions I see immediately are: 
1. Move the signing enum to the initialisation of `Gossipsub`. i.e `Gossipsub::new(Signing, GossipsubConfig)`
2. Add a default Signing variant, which could be Signing::Disabled(IDENTITY_PEERID). 

Both of these don't seem great.

AgeManning: I went with 1.

core/src/identity.rs
```diff
@@ -29,7 +29,7 @@ pub mod secp256k1;
 pub mod error;
 
 use self::error::*;
-use crate::{PeerId, keys_proto};
```
mxinden, line 32: I am guessing that the changes in `identity.rs` and `core/src/identity/error.rs` were done by rustfmt. Would you mind reverting them in this pull-request as they seem unrelated? If you feel like the suggestions would improve code readability please open a separate pull request.

protocols/gossipsub/src/error.rs
```diff
@@ -0,0 +1,40 @@
...
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS
+// OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+// FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+// DEALINGS IN THE SOFTWARE.
+
+//! Error types than can result from gossipsub.
```
mxinden, line 21:
```suggestion
//! Error types that can result from gossipsub.
```

protocols/gossipsub/src/behaviour.rs
mxinden: I see. I guess as this is only scoped to the function itself we can delay the decision. In case it shows up in benchmarks we can easily change it. Rustc or LLVM might as well just optimize it away, not sure.

core/src/identity.rs
AgeManning: Yep, editor has it in by default. I usually catch these, seems I missed one.

core/src/identity/error.rs
```diff
@@ -27,16 +27,22 @@ use std::fmt;
 #[derive(Debug)]
 pub struct DecodingError {
     msg: String,
-    source: Option<Box<dyn Error + Send + Sync>>
+    source: Option<Box<dyn Error + Send + Sync>>,
```
mxinden, line 30: @AgeManning can you revert these formatting changes as well?

burdges:
As an aside..

We've some internal discussions from which I gleaned that one might improve performance by signing gossip messages with 256 byte Rabin-Williams signatures, not that Rust has any good implementation, but hey..

In short, Schnorr needs 64 bytes, but requires 40ms for verification, while 256 byte Rabin-Williams signatures have negligible verification time.  Is 192 bytes worth 25ms?  It's maybe not on-chain, but only gossiped maybe yes.  I'm not sure this is the layer in which to test this of course.  Also, I'd expect the silicon traversed and energy burn by that 192 bytes maybe worse than 25ms, but that's an externality.

AgeManning:
Sorry, took a bit of a detour on this one. 

The extra changes since last review are: 
- Added extensible configuration options to provide a range of privacy and validation modes. This is also to accommodate: https://github.com/libp2p/go-libp2p-pubsub/pull/359. A user can choose the level of message validation and author `authenticity`/ privacy. Invalid mixing of these configurations result in a panic on initialisation. 
- Fixes that resolve #1674 and #1671
- Adds back a separate duplicates cache. The idea is that this can be much larger than the memcache and only holds message_ids rather than the entire message. 
- When doing message validation, messages that have not been validated are excluded from gossip and hence propagation.

I'll be doing some tests on the current version, but I think this is ready for another review now.

protocols/gossipsub/src/behaviour.rs
```diff
@@ -18,40 +18,137 @@
...
+    Signed(Keypair),
+    /// Message signing is disabled.
+    ///
+    /// The specified `PeerId` will be used as the author of all published messages. The sequence
+    /// number will be randomized.
+    Author(PeerId),
+    /// Message signing is disabled.
+    ///
+    /// A random `PeerId` will be used when publishing each message. The sequence number will be a
+    /// random number.
```
mxinden, line 74:
```suggestion
    /// A random `PeerId` will be used when publishing each message. The sequence number will be
    /// randomized.
```

Nit pick to align the working with the one above.

protocols/gossipsub/src/behaviour.rs
```diff
@@ -18,40 +18,137 @@
...
+    RandomAuthor,
+    /// Message signing is disabled.
+    ///
+    /// The author of the message and the sequence numbers are excluded from the message.
+    ///
+    /// NOTE: Excluding these fields may make these messages invalid by other nodes who enforce validation of these
+    /// fields. See [`ValidationMode`] in the `GossipsubConfig` for how to customise this for rust-libp2p gossipsub. A custom message_id function will need to be set to prevent all messages from a peer being filtered as duplicates.
```
mxinden, line 81: Can you wrap these lines?

protocols/gossipsub/src/behaviour.rs
```diff
@@ -18,40 +18,137 @@
...
+    fn is_anonymous(&self) -> bool {
+        match self {
+            MessageAuthenticity::Anonymous => true,
+            _ => false,
+        }
+    }
+}
+
+/// A data structure for storing information for publishing messages.
+enum PublishInfo {
```
mxinden, line 103:
```suggestion
/// A data structure for storing configuration for publishing messages.
enum PublishConfig {
```

I would rather call this a _configuration_ than an _information_, no?

protocols/gossipsub/src/behaviour.rs
```diff
@@ -18,40 +18,137 @@
...
+    Signing {
+        keypair: Keypair,
+        author: PeerId,
+        inline_key: Option<Vec<u8>>,
+    },
+    // Signing is disabled, but this author is used to publish messages.
+    Author(PeerId),
+    /// The author is radomized each message.
+    RandomAuthor,
+    /// The from and sequence number fields are excluded from the message
```
mxinden, line 115: I would suggest linking to the doc comments of `MessageAuthenticity` instead of duplicating the information here.

protocols/gossipsub/src/behaviour.rs
```diff
@@ -62,64 +159,66 @@ pub struct Gossipsub {
...
 
 impl Gossipsub {
-    /// Creates a `Gossipsub` struct given a set of parameters specified by `gs_config`.
-    pub fn new(local_peer_id: PeerId, gs_config: GossipsubConfig) -> Self {
-        let local_peer_id = if gs_config.no_source_id {
-            PeerId::from_bytes(crate::config::IDENTITY_SOURCE.to_vec()).expect("Valid peer id")
-        } else {
-            local_peer_id
-        };
+    /// Creates a `Gossipsub` struct given a set of parameters specified by via a `GossipsubConfig`.
```
mxinden, line 192:
```suggestion
    /// Creates a `Gossipsub` struct given a set of parameters specified via a `GossipsubConfig`.
```

AgeManning:
Thanks for the review. Have updated based on comments.

romanb:
Are these changes backward-compatible? Could someone update the changelog?

mxinden:
@AgeManning do you have time to open up a pull request with [a changelog entry](https://github.com/libp2p/rust-libp2p/blob/master/protocols/gossipsub/CHANGELOG.md) for these changes? Otherwise I will prepare one in the next couple of days.

AgeManning:
Hey yep, can do. 

I've also discovered that the `lru_time_cache` used for duplicates here does an `O(N)` lookup for each message. For a network with a lot of gossipsub messages this isn't great. I've got corrections for this on my own branch. I can make a PR for them, but I'm tempted to wait for the 1.1 PR which we'll make within a week and should contain this modifications also.

PR #1587, opened 2020-05-23
Fully qualified std::result::Result in macro

ureeves:
Currently writing code where one defines a type alias named `Result` combined with a type that uses `#[derive(NetworkBehaviour)]` sometimes leads to [E0107](https://doc.rust-lang.org/error-index.html#E0107).

```rust
type Result<T> = std::result::Result<T, Error>;
struct Error {}

#[derive(NetworkBehaviour)]
struct Behaviour<TStore> {
    mdns: Mdns,
    kad: Kademlia<TStore>,
}

// NetworkBehaviourEventProcess implementations...
```

```texterror[E0107]: wrong number of type arguments: expected 1, found 2
   --> src/lib.rs:223:10
    |
223 | #[derive(NetworkBehaviour)]
    |          ^^^^^^^^^^^^^^^^ unexpected type argument
    |
    = note: this error originates in a derive macro (in Nightly builds, run with -Z macro-backtrace for more info)
```

In my estimation this happens due to the use of `Result` in the definition of the `NetworkBehaviour` macro. Using the fully qualified name `std::result::Result` fixes this issue.

### DIFF
--- a/misc/core-derive/src/lib.rs
+++ b/misc/core-derive/src/lib.rs
@@ -469,7 +469,7 @@ fn build_struct(ast: &DeriveInput, data_struct: &DataStruct) -> TokenStream {
                 #(#inject_listener_error_stmts);*
             }
 
-            fn inject_listener_closed(&mut self, id: #listener_id, reason: Result<(), &std::io::Error>) {
+            fn inject_listener_closed(&mut self, id: #listener_id, reason: std::result::Result<(), &std::io::Error>) {
                 #(#inject_listener_closed_stmts);*
             }
 


mxinden:
Retriggered pipeline as I doubt intra-doc-link failure is related to this change set.

ureeves:
The intra-doc-link check was succeeding at first, but the "Build and test" check was failing due to cancellation (taking too long). 

After looking at the normal run time for this check (~11m) and taking into account that it builds just fine after change, I can only assume that something in the environment running the pipeline is making the tests take too long.

Maybe I'm looking at it wrong but it seems too small a change to cause this.

Issue #1588, opened 2020-05-25
CI failing because of async-std 1.6


tomaka:
It seems that the upgrade to `async-std` v1.6 broke our tests.

I'm opening this issue mostly to track down whether the problem has been figured out.

tomaka:
One should revert https://github.com/libp2p/rust-libp2p/pull/1589 before this issue is closed.

dignifiedquire:
please let me know if there is anything that needs fixing on the async-std side, 1.6 was a big change but trying hard not to break codes

romanb:
@dignifiedquire We didn't look into it yet, but if you're curious or want to lend a hand, it is the [libp2p-noise integration tests](https://github.com/libp2p/rust-libp2p/blob/master/protocols/noise/tests/smoke.rs) that stall with `1.6` as if there is a missing task wakeup. It is maybe not a minimal test case but the tests are nevertheless relatively small and only use `libp2p-core`. Just running `cargo test` in `protocols/noise` with async-std 1.6 should be sufficient to reproduce.

tomaka:
We're maybe a bit too much on the cautious side here, but I would also like to run libp2p-with-async-std-1.6 with a production load and see if any issue arises.

romanb:
Closed by https://github.com/libp2p/rust-libp2p/pull/1618. See also https://github.com/libp2p/rust-libp2p/pull/1612.

PR #1589, opened 2020-05-25
Temporarily bound all async-std dependencies.

romanb:
This PR temporarily and conservatively bounds all `async-std` dependencies in libp2p to `< 1.6` as at least the noise smoke tests seem to stall with `1.6`. This is just a temporary defensive measure to stabilise CI and avoid surprises in releases, until the underlying issue has been figured out and https://github.com/libp2p/rust-libp2p/issues/1588 has been resolved.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -87,7 +87,7 @@ libp2p-tcp = { version = "0.19.0", path = "transports/tcp", optional = true }
 libp2p-websocket = { version = "0.19.0", path = "transports/websocket", optional = true }
 
 [dev-dependencies]
-async-std = "1.0"
+async-std = "< 1.6"
 env_logger = "0.7.1"
 
 [workspace]
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -39,7 +39,7 @@ zeroize = "1"
 ring = { version = "0.16.9", features = ["alloc", "std"], default-features = false }
 
 [dev-dependencies]
-async-std = "1.0"
+async-std = "< 1.6"
 libp2p-mplex = { version = "0.19.0", path = "../muxers/mplex" }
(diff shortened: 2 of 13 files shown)


romanb:
Is the rustdoc-links check failure a nightly rust issue maybe? Any hints?

Cargo.toml
```diff
@@ -87,7 +87,7 @@ libp2p-tcp = { version = "0.19.0", path = "transports/tcp", optional = true }
 libp2p-websocket = { version = "0.19.0", path = "transports/websocket", optional = true }
 
 [dev-dependencies]
-async-std = "1.0"
+async-std = "< 1.6"
```
mxinden, line 90: Would this allow versions below `1.0` to be used? If so how about `>=1.0, < 1.6`.

tomaka: I can't find anything about that in the docs, but it might indeed be the case

romanb: I don't think it really matters as cargo prefers newer versions, doesn't it? Do you think the lower bound is important?

tomaka: There's a general problem in the Rust ecosystem: you might be using features introduced in `foo v1.1` but only have `foo v1.0` in your dependencies. Cargo is theoretically allowed to depend on `foo v1.0`, which will make the compilation fail.

This issue is a bit everywhere, but people don't realize it because as you said Cargo prefers newer versions.

I'd say we're in the same situation here, so I'm tempted to not care.

dvc94ch: Mmh, due to semver libp2p selects a version smaller than 1.0, so 0.99 gets selected currently. I think the correct fix would be `=1.5`

tomaka: Yeah we've actually been bitten by it in Polkadot and released two versions with an old async-std (that had the bug with a high number of threads being spawned) without realizing.

PR #1594, opened 2020-05-28
Add PartialOrd and Ord for PeerId

rklaehn:
for using PeerId in order based collections, and for other niceties like deterministic debug output when e.g. printing a set of peer ids.

Ord uses borrow to be consistent with equality.


tomaka:
You seem to have unchecked the "Allow maintainers to modify my branch" checkbox, so please merge master so we can merge this.

rklaehn:
Sorry, was on holiday.

PR #1602, opened 2020-06-02
[libp2p-swarm] Opt-in connection-oriented backpressure for NetworkBehaviours.

romanb:
This is a simplified variant of https://github.com/libp2p/rust-libp2p/pull/1586 since I figured that the introduction of the configurable `event_lead` in https://github.com/libp2p/rust-libp2p/pull/1586 may have obscured the primary intention of the PR: To permit `NetworkBehaviour`s to opt-in to more fine-grained connection-oriented backpressure, thereby improving upon the `pending_event` bottleneck. So here is a variant that removes the notion of allowing the network to "run ahead" for a configurable limit when there is a `pending_event`. Below is a redacted description of the PR description of #1586 that accommodates for the simplification and elaborates on the motivation.

## Motivation

Currently, if a single background task of a connection is slow to process its events for whatever reason, the effect is that the `Swarm` stalls all progress until this background task is ready to receive the event (into its queue). The increase of the underlying buffers in `libp2p-core` are a means to delay this happening but it does not change the fact that the `NetworkBehaviour` is unaware of any such issues of particular connections / connection handlers until the entire Swarm stalls. A `NetworkBehaviour` should be allowed to opt-in to handle such backpressure from the `Network` more gracefully, which is what this proposal of `NetworkBehaviour::inject_connections_busy` provides. In such an implementation, a behaviour can e.g. react by prioritising events emitted for other peers, delay (re-)issuing events for peers of the affected connections, initiate the opening of new connections with these peers etc. The behaviour can still, the moment it deems itself unable to handle more backpressure, just start emitting `Err` from `inject_connections_busy` which will result in the swarm stalling progress via the `pending_event` as usual.

## `Network` to `NetworkBehaviour`

When network connections become slow to process events, which can currently either happen because `ConnectionHandler::inject_event` is slow or because the task scheduler is busy or overwhelmed, the `Swarm` starts being unable to deliver new events emitted by the `NetworkBehaviour` to the respective connection(s). When that happens, it first tries to propagate backpressure to the `NetworkBehaviour` by calling `NetworkBehaviour::inject_connections_busy`, i.e. giving the behaviour the event back to be emitted again at a later time. If the behaviour rejects that event (i.e. does not implement backpressure), the `Swarm` buffers the `pending_event` as usual.
 
The default implementation of `NetworkBehaviour::inject_connections_busy` is `Err(event)`, meaning these changes are fully backward-compatible and without implementing `NetworkBehaviour::inject_connections_busy`, the existing behaviour is preserved, allowing a gradual transition and experimentation.

**Missing**: What I think is really missing to make this backpressure pipeline complete is a backward-incompatible change to `ConnectionHander::inject_event` by allowing it to outright return (i.e. reject) the given event and thus exert backpressure on the background task which in turn would propagate backpressure back through the `Network` to the `Swarm` by not consuming new events from its inbound channel until the `inject_event` accepted the event.

## `NetworkBehaviour` to `Network`

The `NetworkBehaviour` has no _direct_ means for backpressure on `NetworkBehaviour::inject_event` as it is expected from the contract of the trait that whenever the behaviour returns `Poll::Pending`, it is ready to receive more events, and the single `pending_event` prevents further calls to `inject_event`.

## Polling Priority

I changed the polling behaviour of the swarm such that it always polls the `NetworkBehaviour` first, until it returns `Poll::Pending`, which signals that the behaviour is waiting for progress and input from the `Network` (or the network connections signal backpressure, in which case the behaviour is also no longer polled). This has the effect that events buffered in the behaviour as a result of network activity resulting in calls to `inject_event`, `inject_connected` etc. are drained before further such calls happen, thus preventing potentially unbounded buffer growth in the behaviour if there is a lot of network activity. I clarified the documentation on the matter that a behaviour that from some point onwards never again emits `Poll::Pending` usually is misbehaved, as this is a behaviour that, from some point onwards, no longer needs any input or progress from the `Network`. I also think that it makes for easier to understand semantics of how the `Swarm` does the polling, as it makes clear that the `NetworkBehaviour` takes the lead and the underlying `Network` is only an input/output facility driven on demand by the behaviour.


romanb:
(I intentionally request reviews from all of @mxinden, @twittner and @mxinden  because I would like to get feedback / opinions from (at least) all of you :)

tomaka:
If I understand correctly, one correct way to implement `inject_connections_busy` is to discard the packet and return `Ok`? Is there another potentially valid way?

romanb:
> If I understand correctly, one correct way to implement inject_connections_busy is to discard the packet and return Ok? Is there another potentially valid way?

There are many options. Discarding the event, while certainly an option, is not what I would usually expect. The more common approach would be, I think, to re-enqueue these events to be emitted later again by `poll`, possibly based on a timer and meanwhile prioritizing emitting events for other peers. I would also certainly expect that the behaviour would bound the number of events it can "take back" for a peer or total and eventually start either discarding them or returning `Err`, thereby forcing the Swarm to wait on the connection (background tasks), as is currently always done.

Your question seems to imply that you actually think there are no other valid ways of implementing `inject_connections_busy` or am I misreading the subtext?

swarm/src/behaviour.rs
```diff
@@ -101,11 +101,57 @@ pub trait NetworkBehaviour: Send + 'static {
     fn inject_connection_closed(&mut self, _: &PeerId, _: &ConnectionId, _: &ConnectedPoint)
     {}
 
-    /// Informs the behaviour about an event generated by the handler dedicated to the peer identified by `peer_id`.
-    /// for the behaviour.
+    /// Informs the behaviour that one or more connections to a peer are
+    /// currently busy and unable to receive the given event which was
+    /// previously emitted by `NetworkBehaviour::poll`.
```
mxinden, line 106:
```suggestion
    /// emitted by the last call to `NetworkBehaviour::poll`.
```

How about documenting and thus guaranteeing that the event passed to `inject_connection_busy` is always the last event received from `NetworkBehaviour::poll` and not any of the previous ones from the past.

romanb:
> I am wondering how much complexity this introduces on the NetworkBehaviour side, e.g. the need of bookkeeping of the events that were re-enqueued.

Well, obviously as much or little complexity as the behaviour is willing to add for better handling of backpressure. The point here is that right now the behaviour has no choice at all, it has no way to improve upon the quite radical handling of backpressure via the `pending_event` in the Swarm.

> @romanb have you implemented inject_connections_busy for any of the existing NetworkBehaviours in libp2p/rust-libp2p? If so, would you mind sharing that implementation?

No, this is a choice that I think should be left to applications as there is a large spectrum of possible implementations with different trade-offs.

swarm/src/behaviour.rs
romanb: What would be the point of doing that?

mxinden: A `NetworkBehaviour` can thus be sure that all previously emitted events via `NetworkBehaviour::poll` have been handled and won't be passed back via `inject_connection_busy` out-of-order at a later point in time.

romanb: > A `NetworkBehaviour` can thus be sure that all previously emitted events via `NetworkBehaviour::poll` have been handled and won't be passed back via `inject_connection_busy` out-of-order at a later point in time.

Naturally, but my question was more aiming at: How is it useful for the behaviour to know that it is always the last event emitted from `poll`? Note that the `NetworkBehaviour` already guarantees that all events emitted for the connection of a peer are delivered in the same order to the handler as they are emitted. That implies that if the behaviour gets an event back via `inject_connections_busy` all previous events have either been delivered or also given to `inject_connections_busy`. This is already very close to the guarantee you want to give, but avoids being as specific as referring to the "last call to poll", which I think is an unnecessary over-specification. The ordering guarantees w.r.t. emitted events could maybe be made clearer in the docs, what do you think?

mxinden: > The ordering guarantees w.r.t. emitted events could maybe be made clearer in the docs, what do you think?

Yes, with the already existing ordering guarantees in mind my initial suggestion is superfluous. Where are these documented?

mxinden:
I am having difficulties reviewing this code without seeing a larger implementation of `inject_connections_busy`, especially with composed `NetworkBehaviour`s. This change seems to only introduce little complexity within `rust-libp2p/swarm` but I wonder what the impact would be e.g. for Substrate's [`Behaviour`](https://github.com/paritytech/substrate/blob/master/client/network/src/behaviour.rs) (which I am guessing this patch is intended for).

*I don't think my opinion should have much influence here, thus feel free to ignore this comment.*

romanb:
> I am having difficulties reviewing this code without seeing a larger implementation of `inject_connections_busy`, especially with composed `NetworkBehaviour`s. [..]

So you're wondering how feasible it is to actually implement backpressure in a `NetworkBehaviour` and thus wondering if there will be any implementation deviating from the default implementation at all? I cannot answer that, but library features often need to be judged without being able to anticipate all potential ways they may be used. The reason this is opt-in with no change to the default behaviour is to find that out. W.r.t. composed behaviours, there is always one "top-level" behaviour which controls the entire `NetworkBehaviour` API. In particular, it is aware and can control all emitted `NotifyHandler` actions and their targeted peers and connections emitted from `poll()`. As far as I can say that is mainly what you would need to implement a certain degree of backpressure, other than just always returning `Err` or `Ok`, in your top-level behaviour by re-prioritising the emitted events if you're informed of busy connections/peers so that the Swarm doesn't stall on a single slow / busy / waiting connection background task.

What I would like to get your (and everyone else's) opinion on is:

  a) Do you agree with the motivation, i.e. that there is a problem to solve in the first place?
  b) If the answer to a) is yes, do you find this proposal for opt-in improvement on the problem worthwhile trying out or do you see alternatives?

I understand an approving review as a "Yes" to both. If there is someone whose answer to a) is "No" I understand that as a veto and we can bury this PR since I think everyone should at least be on the same page w.r.t. the motivation.

tomaka:
I'm not opposed to merging this.

However you have to update the code in `misc/core-derive` to add the new `NetworkBehaviour` method.

romanb:
Since I'm no longer motivated to bring this PR to an end - the reserved responses tell me that I would be rather forcing these changes onto people just to get my way, rather than the changes being welcome - I'm closing this PR. If someone still finds parts of this PR useful, feel free to extract them!

Issue #555, opened 2018-10-10
If peer id length <= 42, use identity hashing


tomaka:
If the length of the serialized peer id protobuf is inferior or equal to 42 bytes, use the identity hashing instead.

https://github.com/libp2p/specs/pull/100

mgoelzer:
Claiming this.  I'll have a PR in 2 weeks.

tomaka:
This has been done in #1230/#1237

PR #1608, opened 2020-06-15
Remove temporary peer ID compatibility mode.

romanb:
This removes the temporary compatibility mode between peer IDs using identity hashing and sha256, thus being
the last step in the context of https://github.com/libp2p/rust-libp2p/issues/555. This has low priority, I just wanted to get this out as a PR as a reminder to do this.


tomaka:
I think we can go ahead with this now.

core/src/peer_id.rs
```diff
@@ -77,57 +72,35 @@ impl PeerId {
...
-        if data.algorithm() == multihash::Code::Sha2_256 {
-            Ok(PeerId { multihash: data, canonical: None })
-        } else if data.algorithm() == multihash::Code::Identity {
-            let canonical = Sha2_256::digest(data.digest());
-            Ok(PeerId { multihash: data, canonical: Some(canonical) })
-        } else {
-            Err(data)
+    pub fn from_multihash(data: Multihash) -> Result<PeerId, Multihash> {
+        match data.algorithm() {
+            Code::Sha2_256 | Code::Identity => Ok(PeerId { multihash: data }),
```
tomaka, line 102: This isn't the case right now, but maybe we should check that `data.len() <= MAX_INLINE_KEY_LENGTH` when the algorithm is `identity`?

romanb: Sounds good to me.

romanb: See https://github.com/libp2p/rust-libp2p/pull/1608/commits/c251d8f730fdec90fb82d53ecb274fd474270936.

core/src/peer_id.rs
```diff
@@ -77,57 +72,40 @@ impl PeerId {
...
-                    Ok(PeerId { multihash, canonical: Some(canonical) })
-                } else {
-                    Err(multihash.into_bytes())
-                }
-            }
+        match Multihash::from_bytes(data) {
+            Ok(multihash) => match multihash.algorithm() {
+                Code::Sha2_256 | Code::Identity => Ok(PeerId { multihash }),
+                _ => Err(multihash.into_bytes())
+            },
```
twittner, line 93:
```suggestion
            Ok(multihash) => PeerId::from_multihash(multihash).map_err(Multihash::into_bytes),
```

twittner (approved):

tomaka (approved):

core/src/peer_id.rs
```diff
@@ -77,57 +72,37 @@ impl PeerId {
...
         let key_enc = key.into_protobuf_encoding();
 
-        // Note: before 0.12, this was incorrectly implemented and `SHA2256` was always used.
-        // Starting from version 0.13, rust-libp2p accepts both hashed and non-hashed keys as
-        // input (see `from_bytes`). Starting from version 0.16 rust-libp2p will compare
-        // `PeerId`s of different hashes equal, which makes it possible to connect through
-        // secio or noise to nodes with an identity hash. Starting from version 0.17, rust-libp2p
-        // will switch to not hashing the key (i.e. the correct behaviour).
-        // In other words, rust-libp2p 0.16 is compatible with all versions of rust-libp2p.
-        // Rust-libp2p 0.12 and below is **NOT** compatible with rust-libp2p 0.17 and above.
```
tomaka, line 87: Maybe we could leave that comment, but I don't have a strong opinion.

romanb: I incorporated it into the changelog (see https://github.com/libp2p/rust-libp2p/pull/1608/commits/750476445539dcc157462c2529340d9345e83b65).

PR #1609, opened 2020-06-15
Avoid recursive dependencies due to tests.

twittner:
~~Integration tests from `core` and `core-derived` are moved into a top-level `tests` directory.~~ Edit: See https://github.com/libp2p/rust-libp2p/pull/1609#discussion_r441617944


twittner:
If the `core_` file prefix is deemed annoying I can rename the test files.

Cargo.toml
```diff
@@ -89,6 +89,9 @@ libp2p-websocket = { version = "0.19.0", path = "transports/websocket", optional
 [dev-dependencies]
 async-std = "~1.5.0"
 env_logger = "0.7.1"
+libp2p = { path = ".", features = ["identify", "kad", "ping"] }
```
tomaka, line 92: I must admit that I'm surprised that the CI passes.
It seems to me that this will also fail at the `cargo publish` step, which is a shame since the objective of this PR is to fix that.

I think that the idiomatic solution is to use `required-features`: https://doc.rust-lang.org/cargo/reference/cargo-targets.html#the-required-features-field
Unfortunately that means that you have to manually add one `[[test]]` section per integration test.

twittner: > It seems to me that this will also fail at the cargo publish step, which is a shame since the objective of this PR is to fix that.

No, it would not. `cargo package` ignores `dev-dependencies` without a version.

>I think that the idiomatic solution is to use required-features: https://doc.rust-lang.org/cargo/reference/cargo-targets.html#the-required-features-field Unfortunately that means that you have to manually add one [[test]] section per integration test.

The documentation of `required-features` states that "If any of the required features are not enabled, the target will be skipped" which means the tests would not be executed for `cargo test --no-default-features`. Not to mention that any doc tests (which can not be named AFAIK) which require features fail.

I just realised that the path dependency re-enables the default features of `libp2p` and I could just have added `libp2p = { path = "." }` to the same effect. Now the only problem with all this is that if clients run `cargo test -p libp2p --no-default-features` in their own project the tests would fail to compile and if the tests would require features which are not enabled by default a `cargo test -p libp2p` would too.

tomaka: > No, it would not. cargo package ignores dev-dependencies without a version.

Oh I didn't know that. What about fixing the existing code by removing the version, instead of moving the tests, then?

twittner: How would users be able to execute the tests then, e.g. of `libp2p-core`?

tomaka: I don't understand what you're asking.
I was suggesting the idea of keeping the tests where they are, but for example replace `[dev-dependencies] libp2p-mplex = { version = "0.19.0", path = "../muxers/mplex" }` with `libp2p-mplex = { path = "../muxers/mplex" }`.
Would that not work?

twittner: It would mean that the published crate lacks `libp2p-mplex` as a `dev-dependency` and if a user in their own project does a `cargo test -p libp2p-core` it would fail to compile because the tests of `libp2p-core` do depend on `libp2p-mplex`.

twittner: Actually it seems one can only execute `cargo test -p <crate>` if `crate` has zero `dev-dependencies` or is part of a workspace.

twittner: So, I removed the `version` fields from workspace members in `dev-dependencies`. NB that in some cases (e.g. `core-derive`, `libp2p-websocket`) the published crate will therefore have 0 `dev-dependencies` and can thus be tested in external projects with `cargo test -p` and the tests will fail to compile because the required test dependencies are missing. However since this use-case seems to be rare and cargo would outright refuse to test dependencies with any `dev-dependencies` I suppose we are fine with this.

PR #1612, opened 2020-06-17
Upgrade async-std to 1.6.1

### LINKED ISSUE #1588: CI failing because of async-std 1.6

secretfader:
Resolves #1588, and allows `libp2p` to be used in projects that rely on the latest version of `async-std`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -87,7 +87,7 @@ libp2p-tcp = { version = "0.19.1", path = "transports/tcp", optional = true }
 libp2p-websocket = { version = "0.19.0", path = "transports/websocket", optional = true }
 
 [dev-dependencies]
-async-std = "~1.5.0"
+async-std = "1.6.1"
 env_logger = "0.7.1"
 
 [workspace]
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -39,7 +39,7 @@ zeroize = "1"
 ring = { version = "0.16.9", features = ["alloc", "std"], default-features = false }
 
 [dev-dependencies]
-async-std = "~1.5.0"
+async-std = "1.6.1"
 libp2p-mplex = { version = "0.19.0", path = "../muxers/mplex" }
 libp2p-secio = { version = "0.19.0", path = "../protocols/secio" }
 libp2p-tcp = { version = "0.19.0", path = "../transports/tcp" }
--- a/misc/multistream-select/Cargo.toml
+++ b/misc/multistream-select/Cargo.toml
@@ -18,7 +18,7 @@ smallvec = "1.0"
 unsigned-varint = "0.3.2"
 
 [dev-dependencies]
-async-std = "~1.5.0"
+async-std = "1.6.1"
 quickcheck = "0.9.0"
 rand = "0.7.2"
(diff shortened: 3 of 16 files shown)


mxinden:
Have the failures related to https://github.com/libp2p/rust-libp2p/issues/1588 been addressed?

secretfader:
@mxinden I'm not familiar with the issues encountered earlier in #1588. In fact, I hadn't seen that issue until now. My primary goal was to reconcile dependency versions of `async-std` with requirements of a project that uses `libp2p`.

If I understand correctly, the original issue was that CI broke with the update of `async-std` to `1.6.0.`. Based on what I can see here, it seems all CI checks have passed except one.

secretfader:
After re-reading #1588, I see the problem is with the Noise protocol smoke test, which I'll attempt to resolve here.

secretfader:
@mxinden Tests are passing with `async-std` version `1.6.1`, which should resolve #1588. The problem? Tests were assuming an available runtime to schedule async tasks against, but found none and stalled execution instead.

While including `async-std` everywhere async tests will need to run is an imperfect solution, but one I'm comfortable accepting temporarily (until we find a more ergonomic and performant method).

tomaka:
It is extremely unsettle-ing If `async-std` now forces you to use their runtime for the sockets to work.
The sole and entire reason why we were using `async-std` in the first place is that you didn't have to do that.
`tokio` at least panics if no runtime is found, which is way better than silently stalling futures.

dignifiedquire:
@tomaka @secretfader when you are saying you are not using the runtime, are you manually polling the sockets, or how exactly is the setup you are expecting to work?

dignifiedquire:
> if it is an expected behaviour that sockets don't work if you don't use `#[async_std::test]`

The goal is to have them work, when polling manually, independent of the runtime that is polling them. The current situation for TcpStream::connect is the following

- in 1.5 and earlier SocketAddrs were resolved using a manual implementation, which would only spawn a blocking task for resolution if it couldn't resolve them locally, see https://github.com/async-rs/async-std/blob/v1.5.0/src/net/addr.rs for details 
- in 1.6 we use delegate the connection to `smol::Async`, which immediately spawns a blocking task, no matter of the input data: https://github.com/stjepang/smol/blob/master/src/async_io.rs#L661 so the behaviour is the same for cases where resolution needs to happen, but not where the resolution can be done locally. 

> This isn't mentioned in their CHANGELOG

Because I did not realize this behaviour had chaneged, and it was not my intention to bind the sockets to the runtime in this way. I am considering this a bug in the 1.6, and effectively for the cases it is happening in 1.5 too. I am sorry for the troubles this is causing.

dignifiedquire:
After speaking @stjepang I realized my analysis above was wrong. We need to ensure that the reactor is running when creating sockets, which is simply in bug in async-std atm. Going to attempt a fix and see if I can get libp2ps test working out of the box

dignifiedquire:
Good news, I have a working fix for async-std, which I will be publishing once CI and review are done. This will allow you to upgrade without any other changes to the tests, as it was intended.

PR #728, opened 2018-12-04
Run rustfmt on transports

tomaka:
Runs rustfmt. Please say a word if you think you have ongoing work that conflicts with this PR.

PR #1613, opened 2020-06-18
chore: format with rustfmt

secretfader:
While contributing, I found myself accidentally committing changes that were made by `rustfmt`. To my surprise, most of the project hasn't been formatted, so I ran `cargo fmt` with the goal of increasing readability.

Should the project enforce code style via CI? I'd like to suggest that, as I've seen formatting enforcement make pull requests much easier on other projects, but I'd like feedback before proceeding further.


mxinden:
Hi @secretfader,

Thanks for the contribution. This has been discussed many times in the past (https://github.com/libp2p/rust-libp2p/pull/1164, https://github.com/libp2p/rust-libp2p/pull/728, https://github.com/libp2p/rust-libp2p/pull/727#issuecomment-444426422) and thus far the decision is to not format rust-libp2p with rustfmt.

I myself don't have a strong opinion on this (anymore). I agree with @twittner that [we don't have "constant arguments about the placements of commas, braces and newlines"](https://github.com/libp2p/rust-libp2p/pull/727#issuecomment-444426422) today.

If you feel strongly about this it is probably best to involve @twittner  here.

secretfader:
I'm always surprised by the topics that often become controversial. The arguments for and against `rustfmt` have been discussed ad nauseam, and I don't care to reopen that Pandora's Box once again.

My effort is better used elsewhere, so I'll just close without further discussion.

PR #1614, opened 2020-06-18
Add CHANGELOG and increment version to 0.19.2

twittner:
Prepare bugfix release of `libp2p-ping`.

### DIFF
--- a/protocols/ping/CHANGELOG.md
+++ b/protocols/ping/CHANGELOG.md
@@ -1,4 +1,4 @@
-# 0.19.2
+# 0.19.2 [2020-06-18]
 
 - Close substream in inbound upgrade
   [PR 1606](https://github.com/libp2p/rust-libp2p/pull/1606).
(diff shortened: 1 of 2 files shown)


protocols/ping/CHANGELOG.md
```diff
@@ -0,0 +1,5 @@
+# 0.19.2
```
tomaka, line 1: A date wouldn't be a bad idea I think.

```suggestion
# 0.19.2 (2002-06-18)
```

tomaka:
If I understand correctly, if tomorrow we wanted to release a new minor version of, say, `libp2p-kad`, we would release it under `0.19.1`, is that right?

Consequently, we wouldn't tag releases anymore, I suppose?

protocols/ping/CHANGELOG.md
mxinden: "2020-06-18"

romanb:
> If I understand correctly, if tomorrow we wanted to release a new minor version of, say, libp2p-kad, we would release it under 0.19.1, is that right?

I think that is the idea - any subproject / crate can be released independently (except for necessary releases of dependencies, if any).

> Consequently, we wouldn't tag releases anymore, I suppose?

What is the problem with tags of the form `libp2p-ping-v0.19.2`?

PR #1618, opened 2020-06-19
feat: update async-std to 1.6.2

dignifiedquire:
~~(still testing, will mark as ready when I am confident it is working)~~

Ref #1612


dignifiedquire:
Ready for review, all tests are now passing out of the box, thanks to due diligent work from @stjepang

tomaka:
I'm trying this on Polkadot and it seems to work so far. :rocket:

tomaka:
I'm still going to test more, but I think it's ok to merge this.
I think we should wait for a couple days to put it in a release though.

twittner:
Then we should wait with the merge until we have published the new libp2p version which will require `async-std` 1.5.0. If we merge this now we would use 1.6.2 as a lower bound.

tomaka:
It's been running for 4 hours now, and there isn't any anomaly in sight.

tomaka:
Let's merge this?

Issue #1584, opened 2020-05-19
Banning a connected peer does not result in calls to NetworkBehaviour methods


koivunej:
Banning a connected peer with `ExpandedSwarm::ban_peer_id` does not seem to result in any NetworkBehaviour method calls by the swarm on the banning peer side. On the banned peer the closed connection(s) and disconnected peer are forwarded to NetworkBehaviour's respective methods.

Can't see any changes to the behaviour in #1440 and the original #1065 which introduced the ban/unban feature seems to work in a similar way. I think the expected behaviour would be to notify the swarm's NetworkBehaviour of the disconnection to allow it/them to cleanup any possible state they may have had regarding the peer. This could probably be done through `inject_connection_closed` and finally `inject_disconnected`.

romanb:
This should be fixed in `master` and thus in the next release.

PR #1619, opened 2020-06-19
[core/swarm] Emit events for active connection close and fix `disconnect()`.

romanb:
The `Network` does currently not emit events for actively closed connections, e.g. via `EstablishedConnection::close`
or `ConnectedPeer::disconnect()`. As a result, when actively closing connections, there will be `ConnectionEstablished`
events emitted without eventually a matching `ConnectionClosed` event. This seems undesirable and has the consequence that
the `Swarm::ban_peer_id` feature in `libp2p-swarm` does not result in appropriate calls to `NetworkBehaviour::inject_connection_closed` and `NetworkBehaviour::inject_disconnected`. Furthermore, the `disconnect()` functionality in `libp2p-core` is currently broken as it leaves the `Pool` in an inconsistent state.

This PR does the following:

  1. When connection background tasks are dropped (i.e. removed from the `Manager`), they always terminate immediately, without attempting an orderly close of the connection.

  2. An orderly close is sent to the background task of a connection as a regular command. The background task emits a `Closed` event before terminating.

  3. `Pool::disconnect()` removes all connection tasks for the affected peer from the `Manager`, i.e. without an orderly close, thereby also fixing the discovered state inconsistency due to not removing the corresponding entries in the `Pool` itself after removing them from the `Manager`. The `Pool` ensures that `ConnectionClosed` events are emitted for these connections. The former `NetworkEvent::ConnectionError` has been renamed to `NetworkEvent::ConnectionClosed` with the `error` field being an `Option`, thus `error: None` indicates an active (but not necessarily orderly) close.

  4. A new test is added to `libp2p-swarm` that exercises the ban/unban functionality (currently somewhat broken, see https://github.com/libp2p/rust-libp2p/issues/1584) and places assertions on the number and order of calls to the `NetworkBehaviour`. In that context some new testing utilities have been added to `libp2p-swarm` as well.

Points (1)-(3) ensure that each `NetworkEvent::ConnectionEstablished` is eventually paired with a `NetworkEvent::ConnectionClosed`, also in the case of actively closed connections (with orderly shutdown or not), thus addressing https://github.com/libp2p/rust-libp2p/issues/1584. (1) and (2) went along with some internal simplifications in the state machine implementation of the background `Task`s.


tomaka:
I haven't read the code changes yet, but it is completely intentional that no event is emitted if `close` or `disconnect` is called.

Unless that has been changed, the API of `close` and `disconnect` guarantees that the connection in question will no longer be visible in the API of `Network` afterwards. The user can assume that calling `close`/`disconnect` implicitly generates an instantaneous `ConnectionClosed` event.

Unless there is a reason why the closing has to be asynchronous, I do strongly prefer when disconnecting appears to be synchronous.

romanb:
> I haven't read the code changes yet, but it is completely intentional that no event is emitted if close or disconnect is called.

I think that is not a good idea, as can be seen by https://github.com/libp2p/rust-libp2p/issues/1584 which is just incorrect behaviour. The `Swarm` relies on `Established/Closed` events coming in pairs and it seems cleaner to not make a distinction there between active/passive close.

> Unless that has been changed, the API of close and disconnect guarantees that the connection in question will no longer be visible in the API of Network afterwards.

That is still the case for `disconnect` (used by `Swarm::ban_peer_id`), which is essentially an ungraceful "go away" signal. A proper `close`, however, is only complete once the background task has completed the orderly close. Implying an immediate "connection closed" success here, as is done right now, is just incorrect in my opinion, as the orderly shutdown may also fail. As I wrote in the PR description, this PR is making a clear distinction between an orderly shutdown/close of a connection and an abrupt disconnect.

> Unless there is a reason why the closing has to be asynchronous on the API level, I do strongly prefer when disconnecting appears to be synchronous.

Nothing much changed on the API-level, since the background tasks are still there. The difference is that a) `disconnect()` ensures there is never an attempt at a graceful close in the background task, i.e. it stops immediately and b) For each closed connection that has been actively closed (disconnect or close) an event is also emitted that pairs up with the earlier `ConnectionEstablished`.

romanb:
I want to hightlight again here the motivations of the PR description: a) The `disconnect` functionality in `libp2p-core` is currently broken and b) The banning functionality in the `Swarm` is also broken, partly as a consequence of a) but also because the `NetworkBehaviour` callbacks are not invoked.

tomaka:
> I think that is not a good idea, as can be seen by #1584 which is just incorrect behaviour. The Swarm relies on Established/Closed events coming in pairs and it seems cleaner to not make a distinction there between active/passive close.

My argument is that we should change the Swarm to no longer rely on that. The Swarm knows when `close`/`disconnect` is called and can generate a `SwarmEvent` itself. I don't see any reason to have a 1:1 mapping between the events from the Network and the events from the Swarm.

> Implying an immediate "connection closed" success here, as is done right now, is just incorrect in my opinion, as the orderly shutdown may also fail.

I don't understand why it is important to know whether the shutdown has been successful or not. There is nothing the user can do if for example we fail to shut down a connection because of an I/O error.

> a) The disconnect functionality in libp2p-core is currently broken 

That's not clear to me. I would expect `disconnect()` (and `close`) to perform a clean shut down.

> but also because the NetworkBehaviour callbacks are not invoked for closed connections resulting from an active close (including disconnect).

Which can be fixed by making the `NetworkBehaviour` call the callback from within `ban_peer_id`. We don't need to modify the behaviour of `disconnect()` for that.

romanb:
> I don't see any reason to have a 1:1 mapping between the events from the Network and the events from the Swarm.

Such a an overall 1:1 mapping between Network and Swarm events is not the motivation or goal here, this is only about connection established/closed events. I do see a good reason for having `Network::ConnectionEstablished`and `Network::ConnectionClosed` (and similarly `Swarm::ConnectionEstablished`/`Closed`) always come in pairs, because the symmetry makes it easy to understand and write correct code that is handling these events. As witnessed by https://github.com/libp2p/rust-libp2p/issues/1584, it is easy to make the mistake of forgetting that you don't get `Closed`-events for connections that have been closed as a result of some code path using `close()` or `disconnect()`. Now as a user of `libp2p-core`'s `Network` like the `Swarm` here, instead of just writing code that handles closed connections in a single place, where you handle `Network::ConnectionClosed`, you need to remember that this code won't run if you explicitly `close()` or `disconnect()`. That is just not a nice API to program against in my opinion.

> I don't understand why it is important to know whether the shutdown has been successful or not. There is nothing the user can do if for example we fail to shut down a connection because of an I/O error.

How do you know? If I try to perform a clean shutdown of a connection, which implies flushing and sending all remaining buffered data, I may certainly be interested in whether that succeeded. If it doesn't, I certainly have to assume that some of the data was not sent and may want to take action on that. If I don't care or know at all about whether an orderly connection shutdown succeeds, there isn't much point in doing it in the first place, I can just drop the connection. The current code (before this PR) always tries to perform an orderly `close()` in the background task when it notices that the `Manager` has dropped the task, but at that point no-one will or can know about the outcome, which seems pointless to me. If errors on close/shutdown were irrelevant as you say, surely there would be no need to even return an error from [futures::io::AsyncWrite::poll_close](https://docs.rs/futures/0.3.5/futures/io/trait.AsyncWrite.html#tymethod.poll_close) or [tokio::io::AsyncWrite::poll_shutdown](https://docs.rs/tokio/0.2.21/tokio/io/trait.AsyncWrite.html#tymethod.poll_shutdown).

> >  a) The disconnect functionality in libp2p-core is currently broken

> That's not clear to me. I would expect disconnect() (and close) to perform a clean shut down.

What is not clear? Why it is broken? As I hinted at in the PR description and should be seen in the diff (and has been reproduced by the included new `libp2p-swarm` test), `Pool::disconnect()` currently leaves the `Pool` in an inconsistent state because it does not remove the corresponding connection entries it removed from the `Manager` from its own `established` and `pending` lists.
If instead you mean it is not clear why I made `disconnect()` and `close()` behave differently, that is motivated by two things:

  1. The current use of `disconnect()` in the context of `Swarm::ban_peer_id`. When you ban a peer, I don't think you want to have the connections to that peer potentially hanging around for quite some time to complete the orderly `close()`, which is what the background tasks will currently try to do. Rather, you want to drop the connections to the peer immediately.

  2. If you want to cleanly shut down all connections to a peer, you can already do that by iterating through the connections to that peer and either `abort()` (for pending connections) or `close()` them (for established connections). There is really no need for a separate `disconnect()` functionality just to achieve that.

> My argument is that we should change the Swarm to no longer rely on that.

The thing is, the `Swarm` already does and needs to rely on `Established` / `Closed` coming in pairs _at least for the case where connections are not actively closed_, since that is instrumental to deciding when to call `inject_disconnected()`, i.e. when the last connection closes. You're thus proposing it shouldn't rely on them coming in pairs for the case of actively closed connections, as you say:

> The Swarm knows when close/disconnect is called and can generate a SwarmEvent itself.
> [..]
> Which can be fixed by making the NetworkBehaviour call the callback from within ban_peer_id. We don't need to modify the behaviour of disconnect() for that.

Yes, it can, as long as you can get access to all the information you need to feed into the callbacks and / or events. I think that is possible now, but must also always remain possible in the future (that is, in principle, any information you get in a `Network::ConnectionClosed` event must be obtainable through other APIs, since any such information may need to be passed to `inject_connection_closed` or `_disconnected`). However, because of that and for reasons stated at the beginning of this post whereby you have to repeat the code that needs to run on connection closure in multiple places, I don't think this is a good solution in terms of API design and hence did not propose it but rather an approach that I find preferable, which is the pairing of established/closed events no matter how connections are closed. If you find it preferable to address this in the `Swarm` only in the way you suggest, feel free to take from this PR what you like and leave what you don't (just don't forget the necessary changes to `core::connection::Pool::disconnect()`, which I think are necessary in any case).

romanb:
Any further comments / questions? If my arguments are not convincing, and no-one else seems to have an opinion, feel free to close the PR. Otherwise I'm happy to rebase it.

core/src/connection/manager/task.rs
```diff
@@ -218,112 +229,114 @@ where
...
+                                if result.is_ok() {
+                                    if let Ok(()) = this.events.start_send(event) {
+                                        this.state = State::Established { connection, event: None };
+                                        continue 'poll
+                                    }
+                                }
+                                // The manager is no longer reachable, maybe due to
+                                // application shutdown. Try a graceful shutdown of the
+                                // connection before terminating the task.
+                                this.state = State::Closing(connection.close());
```
twittner, line 273: I know we did it like this so far but I wonder if child resource should linger around and perform I/O when their managing parent has been dropped?

core/src/connection/pool.rs
```diff
@@ -551,6 +592,20 @@ where
         TConnInfo: ConnectionInfo<PeerId = TPeerId> + Clone,
         TPeerId: Clone
     {
+        // Drain events resulting from forced disconnections.
+        while let Some(Disconnected {
+            id, connected, num_established
+        }) = self.disconnected.pop() {
```
twittner, line 598: A comment about the importance of `.pop()` w.r.t. the correctness of `num_established` would be appreciated.

core/src/connection/pool.rs
```diff
@@ -888,3 +921,29 @@ impl PoolLimits {
...
+
+impl<'a, TInEvent, TConnInfo> Future for StartClose<'a, TInEvent, TConnInfo> {
+    type Output = ();
+
+    fn poll(mut self: std::pin::Pin<&mut Self>, cx: &mut Context) -> Poll<()> {
+        self.0.poll_start_close(cx)
+    }
+}
+
+struct Disconnected<TConnInfo> {
```
twittner, line 945: Some documenting comment would be nice.

swarm/src/lib.rs
```diff
@@ -521,8 +523,12 @@ where TBehaviour: NetworkBehaviour<ProtocolsHandler = THandler>,
                         });
                     }
                 },
-                Poll::Ready(NetworkEvent::ConnectionError { id, connected, error, num_established }) => {
-                    log::debug!("Connection {:?} closed: {:?}", connected, error);
+                Poll::Ready(NetworkEvent::ConnectionClosed { id, connected, error, num_established }) => {
+                    if error.is_some() {
```
twittner, line 527:
```suggestion
                    if let Some(error) = error {
```

core/src/connection/manager.rs
```diff
@@ -426,10 +426,11 @@ impl<'a, I, C> EstablishedEntry<'a, I, C> {
     /// > task _may not be notified_ if sending the event fails due to
     /// > the connection handler not being ready at this time.
     pub fn notify_handler(&mut self, event: I) -> Result<(), I> {
-        let cmd = task::Command::NotifyHandler(event);
+        let cmd = task::Command::NotifyHandler(event); // (*)
         self.task.get_mut().sender.try_send(cmd)
             .map_err(|e| match e.into_inner() {
-                task::Command::NotifyHandler(event) => event
+                task::Command::NotifyHandler(event) => event,
+                _ => unreachable!("by (*)")
```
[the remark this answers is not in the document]
romanb: Apart from very simple cases, I think the message alone is rarely sufficient to troubleshoot a panic - you always need to take a look at the code immediately surrounding the line where the panic triggered. The wording of the message here just reflects the currently established practice of supplying a "proof" when `expect()`ing combined with my approach to short proofs. So in the "proof-style" this could also be `unreachable!("Because `cmd` is `Command::NotifyHandler`")`, but I personally don't find it useful to "describe" the surrounding code in the message in this way, rather than just referring to a marked line that states that same assumption. If we used a different practice, where `expect()`, `unreachable()` etc. messages were supposed to describe what went wrong rather than why it cannot happen, they may indeed tend to be a bit more helpful, e.g. "Expected Command::NotifyHandler, got ...". I still find it helpful to succinctly refer to the lines in the code that establish the assumptions for an `expect()` or `unreachable()` but that could of course also be achieved via regular comments, outside of the panic message. So if we don't follow the proof-as-panic-message here, this could be:

```rust
x => unreachable!("Expected `Command::NotifyHandker`, got {:?}", x); // by (*)
```

I wouldn't mind doing this, if you agree that it is an improvement. Or did you have something else in mind?

core/src/connection/manager/task.rs
romanb: I wondered about this as well and am in favour of no longer doing it, because with this PR there is now actually a way to do a "clean" shutdown if you want to by explicitly closing the connections and waiting for the corresponding events.

core/src/connection/manager/task.rs
```diff
@@ -218,112 +229,114 @@ where
...
-                                // graceful shutdown of the connection, if given.
-                                if let Some(c) = connection {
-                                    this.state = State::Closing(c.close());
-                                    continue 'poll
-                                } else {
-                                    return Poll::Ready(())
-                                }
+                State::Closing(mut closing) => {
+                    // Ignore further commands received from the manager, if any.
+                    while let Poll::Ready(Some(_)) = this.commands.poll_next_unpin(cx) {}
```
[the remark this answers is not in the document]
romanb: Yes, it is probably better to close `commands` before entering the `Closing` state.

core/src/connection/manager.rs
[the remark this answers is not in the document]
romanb: Resolved by https://github.com/libp2p/rust-libp2p/pull/1619/commits/603cb2b29958170ba42375b6c5bd439274eb6a71.

romanb:
@twittner Thanks for the view, I incorporated pretty much all your suggestions.

@tomaka I didn't hear back from you yet - do you also have a final verdict here? I would go ahead with merging eventually unless you still have strong objections and my answers to earlier questions were not convincing. If you just didn't find the time to take a closer look yet, I'm happy to leave this PR open for a while longer. The gist of the changes is still that `*Established` events are always paired with `*Closed` events, regardless of how connections are closed. It is true that with these changes connections that are gracefully closed (`EstablishedConnection::start_close`) do not immediately disappear from the API until they are, well, _actually_ closed, which surely isn't non-sensical. This shouldn't really be a drastic change for a user either, because due to the background tasks any "established" connection one has at hand can suddenly actually be closed, which shows in the API (e.g. the possible errors of `notify_handler`). It may make even more sense if I mention here that in https://github.com/libp2p/rust-libp2p/pull/1682 while the connection is performing a graceful shutdown but before it is ultimately closed, the handler is still allowed to receive and emit events (but not get or create any new substreams, see that PR). Of course, whether this is in itself a good idea or not should be discussed in that PR and not here.

PR #1621, opened 2020-06-19
Allow StreamMuxer to notify changes in the address

tomaka:
In the QUIC protocol, remotes can change IP addresses while still retaining the same connection. When that happens, we need, amongst potentially other things, to update the content of DHT.

This PR renames `StreamMuxer::poll_inbound` to `poll_event` and makes the function return either a "new substream" event or an "address change" event. This "address change" event is bubbled up and reported through two new methods: `ProtocolsHandler::inject_address_change` and `NetworkBehaviour::inject_address_change`. I didn't add any `SwarmEvent` because the `SwarmEvent` doesn't expose any `ConnectionId`.

I didn't do the changes to `libp2p-kad` that update the DHT, as I'm not familiar with the code. I'd suggest to open an issue and do them later, unless someone wants to tackle that.


muxers/mplex/src/lib.rs
```diff
@@ -174,6 +175,9 @@ where
 }
 
 /// Multiplexer. Implements the `StreamMuxer` trait.
+///
+/// This implementation isn't capable of detecting when the underlying socket changes its address,
+/// and no [`StreamMuxerEvent::AddressChange`] event is ever emitted.
```
twittner, line 180: This may eventually change with [Multipath TCP](https://en.wikipedia.org/wiki/Multipath_TCP).

core/src/network.rs
```diff
@@ -652,7 +659,7 @@ impl NetworkConfig {
     /// When the buffer is full, the background tasks of all connections will stall.
     /// In this way, the consumers of network events exert back-pressure on
     /// the network connection I/O.
-    pub fn set_connection_event_buffer_size(&mut self, n: usize) -> &mut Self {
+    pub fn set_poll_event_buffer_size(&mut self, n: usize) -> &mut Self {
```
romanb, line 662: What motivated this change? Address-change events still count as connection events to me, even if they're not emitted directly by a `ConnectionHandler`. `poll` seems a bit more ambiguous, but if you think its an improvement, why not. But maybe it would make sense to update the doc-comments then to reflect the change in terminology (and clarify what a "poll event" refers to).

tomaka: Err... that wasn't intended and is an accident.

romanb:
What is the plan w.r.t the `libp2p-kad` integration? Presumably it can be done as a follow-up PR or did you plan to do integrate it here @mxinden? Or asked differently, can / should this PR be merged and the changes potentially released before this is done? Presumably this will only be tested in combination with QUIC.

mxinden:
> Or asked differently, can / should this PR be merged and the changes potentially released before this is done?

I don't see an issue doing it after this pull request is merged. As you already said it will only be used in combination with QUIC (... and eventually Multipath TCP) which is not yet supported. Thus a release without the corresponding libp2p-kad changes won't do any damage.

Any objections? Am I missing something?

PR #1626, opened 2020-06-24
Support /dns/ (in addition to /dns4/ and /dns6/) in the websocket browser transport

expenses:
This fixes the telemetry working in the substrate browser node that was broken in https://github.com/paritytech/substrate/pull/6369.


expenses:
@tomaka

Issue #1560, opened 2020-04-29
Make it possible to disable automatically inserting nodes we connect to to the DHT


tomaka:
In libp2p-kad, whenever we successfully connect to a node, we immediately insert its address in the DHT.

Unfortunately, in Substrate (and probably for other users for libp2p) we don't necessarily want that to happen.
In particular, light clients shouldn't be part of the DHT (https://github.com/paritytech/substrate/issues/3303) and, now that we have multiple DHTs, we should only insert nodes in the DHT after having verified that they have the correct chain root (https://github.com/paritytech/substrate/issues/5825).

There should be an option that disables the auto-insertion.

mxinden:
*Summarizing a discussion in Riot there are two simple and one complex paths forward:*

#### 1. Disable automatic routing table updates

Offering an option in libp2p Kademlia to disable routing table updates when a new connection is established i.e. a simple condition in `inject_connection_established`.

https://github.com/libp2p/rust-libp2p/blob/0443fea157ab7d81feb4952cdf8aaae934130377/protocols/kad/src/behaviour.rs#L1104-L1115

 It would remain enabled by default. When disabled, no entry will get into the routing table without an explicit call to `Kademlia::add_address`, which would be called by Substrate after identifying a peer as (a) part of the right protocol and (b) not a light-client.

https://github.com/paritytech/substrate/blob/c90dcc8d8b3c3661ce4b4e0fb6521fb7c4f9281c/client/network/src/behaviour.rs#L183-L199

#### 2. Provide a method to remove nodes after-the-fact

Offering `remove_address` function, thus being able to remove a peer from the routing table after-the-fact.

Drawbacks of this approach are related to keeping the Kademlia internal peer blacklist in-sync. E.g. we connect to a node, thus it is added to the Kademlia routing table, but before identifying the node we disconnect. In this scenario the peer would stay in the Kademlia routing table forever.

#### 3. Introduce structured prioritization to Kademlias routing table

See details in https://github.com/libp2p/rust-libp2p/issues/1352. This would require us to have an authenticated *address book*, which we currently only have in an eventually consistent form with the authority discovery module.

romanb:
Note that `1.` and `2.` are not intended to be alternatives to choose from but rather both should be offered. Even when automatic insertions into the routing table are disabled and you use `add_address`, there can be legitimate reasons that you want to remove an address (or entire peer) again later (e.g. it is considered misbehaved, changed its role, whatever).

[...]

Issue #1352, opened 2019-12-17
Authenticated stratified k-buckets


infinity0:
I propose that the kademlia implementation should provide more structure in the k-buckets so that "more trusted nodes" (as configured by the higher layer) are guaranteed available resources in the k-buckets and cannot be evicted in favour of less trusted nodes. This would help to greatly raise the cost and lower the effectiveness of an eclipse attack (paritytech/substrate#332) and indirectly and automatically avoid wasting resources storing light clients in the DHT (paritytech/substrate#3303).

Proposal is as follows, the full text is being worked on as part of w3f/research#61:

1. The kademlia implementation should offer an API for the higher-layer application to define different "authenticated roles" for peers that can potentially be inserted into k-buckets, along with the maximum fraction of the whole k-bucket that peers belonging to that role are allowed to occupy. Roles must be sortable. So for example a typical configuration might look like:

    - ``{ role 2: 0.5, role 1: 0.3 }``

    The sum of the fractions should <= 1, with the remainder being allocated to the default unauthenticated role ("role 0"). So in the above example, role 0 would be left with a fraction of 0.2.

    This should be called sporadically e.g. at the start of the process run.

2. The kademlia implementation should offer an API for the higher-layer application to define which cryptographic identities belong to which roles. This should come with an expiry time and be refreshed continually. For example a typical API call might look like:

    - ``{ role 2: member_of { some set }, role 1: some predicate, expires: 1 hour }``

When the k-bucket is not full, these configurations have no effect and the behaviour is the same as ordinary Kademlia. That is, a new node is simply added to the k-bucket, and an existing node has their timestamp refreshed.

When the k-bucket is full however, this configuration serves to affect which other node is potentially evicted in favour of the new node. In ordinary Kademlia it is the node with the oldest timestamp that is nominated for eviction - a PING is sent and if they don't reply then they are evicted in favour of the new node. In our new proposal, the node to maybe-evict is instead selected as follows:

1. from lowest to highest role (starting with the unauthenticated role 0):
    a. if the k-bucket contains more than k*fraction entries for that role, then
        i. select the oldest node *in that role* for eviction
2. if the above loop did not select any node, then
    a. select the oldest node in *the same role* as the new node for eviction

This should serve to guarantee the following security properties:

1. If resources are available (i.e. the k-bucket is not full) then everyone is added, without prejudice.

[...]

PR #1628, opened 2020-06-24
[libp2p-kad] More control & insight for k-buckets.

romanb:
This implements option(s) `1.` and `2.` from https://github.com/libp2p/rust-libp2p/issues/1560 and possibly even allows implementing the role management in option `3.` from https://github.com/libp2p/rust-libp2p/issues/1352 externally.

## More Control
It is now possible to disable automatic insertions of peers into the routing table via a new`KademliaBucketInserts` configuration option. The default is `OnConnected`, but it can be set to `Manual`, in which case `add_address` must be called explicitly. In order to communicate all situations in which a user of `Kademlia` may want to manually update the routing table, two new events are introduced:

  * `KademliaEvent::RoutablePeer`: When a connection to a peer with a known listen address is established
    which may be added to the routing table. This is also emitted when automatic inserts are allowed but
    the corresponding k-bucket is full.
  * `KademliaEvent::PendingRoutablePeer`: When a connection to a peer with a known listen address is established
    which is pending insertion into the routing table (but may not make it). This is only emitted when `OnConnected` (i.e. automatic inserts) are used.

These complement the existing `UnroutablePeer` and `RoutingUpdated` events. It is now also possible to explicitly remove peers and addresses from the routing table via `Kademlia::remove_address` and `Kademlia::remove_peer` respectively.

## More Insight
`Kademlia::kbuckets` now gives an iterator over `KBucketRef`s and `Kademlia::bucket` a particular `KBucketRef`. A `KBucketRef` in turn allows iteration over its entries. In this way, the full contents of the routing table can be inspected, e.g. in order to decide which peer(s) or address(es) to remove.

### DIFF
--- a/protocols/kad/src/behaviour.rs
+++ b/protocols/kad/src/behaviour.rs
@@ -52,10 +52,14 @@ use wasm_timer::Instant;
 
 pub use crate::query::QueryStats;
 
-/// Network behaviour that handles Kademlia.
+/// `Kademlia` is a `NetworkBehaviour` that implements the libp2p
+/// Kademlia protocol.
 pub struct Kademlia<TStore> {
     /// The Kademlia routing table.
     kbuckets: KBucketsTable<kbucket::Key<PeerId>, Addresses>,
+
+    /// The k-bucket insertion strategy.
+    kbucket_inserts: KademliaBucketInserts,
 
     /// Configuration of the wire protocol.
     protocol_config: KademliaProtocolConfig,
@@ -92,6 +96,30 @@ pub struct Kademlia<TStore> {
     store: TStore,
 }
 
+/// The configurable strategies for the insertion of peers
+/// and their addresses into the k-buckets of the Kademlia
+/// routing table.
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub enum KademliaBucketInserts {
+    /// Whenever a connection to a peer is established as a
+    /// result of a dialing attempt and that peer is not yet
+    /// in the routing table, it is inserted as long as there
+    /// is a free slot in the corresponding k-bucket. If the
+    /// k-bucket is full but still has a free pending slot,
+    /// it may be inserted at a later time if an unresponsive
+    /// disconnected peer is evicted from the bucket.
+    OnConnected,
+    /// New peers and addresses are only added to the routing table via
+    /// explicit calls to [`Kademlia::add_address`].
+    ///
+    /// > **Note**: Even though peers can only get into the
+    /// > routing table as a result of [`Kademlia::add_address`],
+    /// > routing table entries are still updated as peers
+    /// > connect and disconnect (i.e. the order of the entries
+    /// > as well as the network addresses).
+    Manual,
+}
+
 /// The configuration for the `Kademlia` behaviour.
 ///
 /// The configuration is consumed by [`Kademlia::new`].
@@ -106,6 +134,7 @@ pub struct KademliaConfig {
     provider_record_ttl: Option<Duration>,
     provider_publication_interval: Option<Duration>,
     connection_idle_timeout: Duration,
+    kbucket_inserts: KademliaBucketInserts,
 }
 
 impl Default for KademliaConfig {
@@ -120,6 +149,7 @@ impl Default for KademliaConfig {
             provider_publication_interval: Some(Duration::from_secs(12 * 60 * 60)),
             provider_record_ttl: Some(Duration::from_secs(24 * 60 * 60)),
             connection_idle_timeout: Duration::from_secs(10),
+            kbucket_inserts: KademliaBucketInserts::OnConnected,
         }
     }
 }
@@ -275,6 +305,12 @@ impl KademliaConfig {
         self.protocol_config.set_max_packet_size(size);
         self
     }
+
+    /// Sets the k-bucket insertion strategy for the Kademlia routing table.
+    pub fn set_kbucket_inserts(&mut self, inserts: KademliaBucketInserts) -> &mut Self {
+        self.kbucket_inserts = inserts;
+        self
+    }
 }
 
 impl<TStore> Kademlia<TStore>
@@ -312,6 +348,7 @@ where
         Kademlia {
             store,
             kbuckets: KBucketsTable::new(local_key, config.kbucket_pending_timeout),
+            kbucket_inserts: config.kbucket_inserts,
             protocol_config: config.protocol_config,
             queued_events: VecDeque::with_capacity(config.query_config.replication_factor.get()),
             queries: QueryPool::new(config.query_config),
@@ -381,7 +418,7 @@ where
     ///
     /// If the routing table has been updated as a result of this operation,
     /// a [`KademliaEvent::RoutingUpdated`] event is emitted.
-    pub fn add_address(&mut self, peer: &PeerId, address: Multiaddr) {
+    pub fn add_address(&mut self, peer: &PeerId, address: Multiaddr) -> RoutingUpdate {
         let key = kbucket::Key::new(peer.clone());
         match self.kbuckets.entry(&key) {
             kbucket::Entry::Present(mut entry, _) => {
@@ -394,9 +431,11 @@ where
                         }
                     ))
                 }
+                RoutingUpdate::Success
             }
             kbucket::Entry::Pending(mut entry, _) => {
                 entry.value().insert(address);
+                RoutingUpdate::Pending
             }
             kbucket::Entry::Absent(entry) => {
                 let addresses = Addresses::new(address);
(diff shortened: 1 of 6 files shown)


romanb:
As a follow-up I would actually like to remove the `KademliaEvent::Discovered` events. They are very noisy and overlap heavily with the `RoutingUpdated` and `RoutablePeer` events. IIRC substrate does not even make use of `Discovered` and I wonder how useful it really is to have an event that is emitted for every single peer found in an iterative query (even if the query will never connect to it). The other events related to actual or potential routing table updates seem much more useful in comparison and I would hope cover all use-cases.

tomaka:
> I wonder how useful it really is to have an event that is emitted for every single peer found in an iterative query (even if the query will never connect to it).

My impression is that it could be used for diagnostic/informative purposes, such as if you want to write a DHT crawler.
But I don't have any actual opinion on that removal.

mxinden:
Re: Removing `KademliaEvent::Discovered`

I kind of wrote a DHT crawler with https://github.com/mxinden/kademlia-exporter where I [deliberately ignore the `Discovered` event](https://github.com/mxinden/kademlia-exporter/blob/master/src/exporter.rs#L271) (apart from the single metric increase done for consistency, not for actual need.)

I don't have a strong opinion. I can't think of a serious use-case for it, but also I don't think it clutters the interface much. Do you think the implied frequent message passing is a performance problem @romanb?

protocols/kad/src/behaviour.rs
```diff
@@ -92,6 +96,30 @@ pub struct Kademlia<TStore> {
...
+/// and their addresses into the k-buckets of the Kademlia
+/// routing table.
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub enum KademliaBucketInserts {
+    /// Whenever a connection to a peer is established as a
+    /// result of a dialing attempt and that peer is not yet
+    /// in the routing table, it is inserted as long as there
+    /// is a free slot in the corresponding k-bucket. If the
+    /// k-bucket is full but still has a free pending slot,
+    /// it may be inserted at a later time if an unresponsive
```
mxinden, line 109:
```suggestion
    /// it may be inserted into the routing table at a later time if an unresponsive
```

To disambiguate between inserting into the routing table and inserting into the pending slot.

protocols/kad/src/behaviour.rs
```diff
@@ -2294,3 +2453,22 @@ impl fmt::Display for NoKnownPeers {
 }
 
 impl std::error::Error for NoKnownPeers {}
+
+/// The possible outcomes of [`Kademlia::add_address`].
+pub enum RoutingUpdate {
+    /// The given peer and address has been added to the routing
+    /// table.
```
mxinden, line 2460:
```suggestion
    /// table or was already part of the routing table before in which case any additional addresses were added to the entry.
```

protocols/kad/src/behaviour.rs
```diff
@@ -769,20 +892,24 @@ where
                                     NetworkBehaviourAction::GenerateEvent(event));
                             },
                             kbucket::InsertResult::Full => {
-                                debug!("Bucket full. Peer not added to routing table: {}", peer)
+                                debug!("Bucket full. Peer not added to routing table: {}", peer);
+                                let address = addresses.first().clone();
+                                self.queued_events.push_back(NetworkBehaviourAction::GenerateEvent(
+                                    KademliaEvent::PendingRoutablePeer { peer, address }
```
mxinden, line 898: Doesn't returning a `PendingRoutablePeer` and the doc comment below contradict?

```rust
    /// A connection to a peer has been established for whom a listen address
    /// is known but the peer is only pending insertion into the routing table
    /// if the least-recently disconnected peer is unresponsive, i.e. the peer
    /// may not make it into the routing table.
```

In this case the node is neither added to the routing table nor is it pending insertion.

romanb, line 898: Yes, this is a mistake. Thanks!
```suggestion
                                    KademliaEvent::RoutablePeer { peer, address }
```

romanb: Yes, this is a mistake and should be `RoutablePeer`, thanks!

protocols/kad/src/behaviour.rs
romanb: In what way do you think this is an improvement? This is the result of `add_address` which takes a peer and an address. Success means the given peer and address are now in the routing table, I don't think it matters whether they were already in there.

romanb:
> Re: Removing KademliaEvent::Discovered
> [..]
> I kind of wrote a DHT crawler [..] where I deliberately ignore the Discovered event [..]
> [..]
> I don't have a strong opinion. I can't think of a serious use-case for it, but also I don't think it clutters the interface much. Do you think the implied frequent message passing is a performance problem @romanb?

It's really just that there is significant overlap with the other events (i.e. for every peer we connect to during an iterative query there is both a `Discovered` and one of `RoutingUpdated`, `RoutablePeer` or `PendingRoutablePeer` emitted. `Discovered` is of course also emitted for every peer encountered in a query that we do _not_ connect to, i.e. it is the most-emitted event. Considering that every iterative query needs at least to successfully contact 20 nodes and each of these returns 20 closer peers (assuming a non-trivial sufficiently large network) there will always at least be 400 `Discovered` events emitted for a single query, even if all 20 peers return the same 20 closer peers! Seeing this and that `Discovered` is neither used by substrate nor by your crawler makes me question whether it is really pulling its "weight". I admittedly can see that it _may_ be interesting to record all peers that are _potentially_ reachable (which is what `Discovered` shows), but it is certainly much more interesting also for a DHT crawler to see which nodes are _actually reachable and queried_, i.e. the paths that queries take on the network. There is simply a lot of redundancy in the emission of `Discovered` events, both in context of the other events as well as among the `Discovered` events themselves.

mxinden:
With the above in mind and especially the fact that we are not aware of a user of the event, I am in favor of removing the `Discovered` event. Anyone objecting?

romanb:
Yes, I intend to do that before merging.

protocols/kad/src/behaviour.rs
```diff
@@ -415,26 +454,97 @@ where
...
+                Some(entry.remove())
+            }
+            kbucket::Entry::Absent(..) | kbucket::Entry::SelfEntry => {
+                None
+            }
+        }
+    }
+
+    /// Returns an iterator over all non-empty buckets in the routing table.
+    pub fn kbuckets(&mut self)
```
mxinden, line 533: I am currently updating Substrate to this change-set. What do you think of an additional `kbuckets_ref(&self)` to iterate immutably? That would help in the [`DiscoveryBehaviour::known_peers`](https://github.com/paritytech/substrate/blob/master/client/network/src/discovery.rs#L233) method.

romanb: The problem is that an immutable borrow can show a stale view of the buckets, because the pending entries are applied "lazily", i.e. whenever there is an access to a bucket we check whether the timeout for the pending entry has expired and if so evict another entry from the bucket.

protocols/kad/src/behaviour.rs
```diff
@@ -769,20 +892,24 @@ where
...
+                                let address = addresses.first().clone();
+                                self.queued_events.push_back(NetworkBehaviourAction::GenerateEvent(
+                                    KademliaEvent::RoutablePeer { peer, address }
+                                ));
                             },
                             kbucket::InsertResult::Pending { disconnected } => {
                                 debug_assert!(!self.connected_peers.contains(disconnected.preimage()));
+                                let address = addresses.first().clone();
+                                self.queued_events.push_back(NetworkBehaviourAction::GenerateEvent(
+                                    KademliaEvent::PendingRoutablePeer { peer, address }
```
folex, line 905: Doc says
>  A connection to a peer has been established

I'm not sure if that's true at this point. `DialPeer` is being scheduled, it may fail. Also, `DialPeer` may result in dialing not `addresses.first()`, but some other address.

Just wanted to state that this is a bit misleading :) Maybe reflect that in the doc comment on `PendingRoutablePeer`?

romanb: > I'm not sure if that's true at this point. DialPeer is being scheduled, it may fail. Also, DialPeer may result in dialing not addresses.first(), but some other address.

A `DialPeer` is scheduled for the peer that is potentially replaced if it does not respond, i.e. for `disconnected` (this is a `PeerId`). `peer` is the one who just connected via `address` and is pending insertion (if `disconnected` turns out unresponsive). So I think this is correct. Does that help? If you have suggestions for clarifying the documentation, feel free to open a PR!

folex: Thanks for clarifying! I definitely missed that `DialPeer` is issued to `disconnected`, not to freshly connected `peer`. I was wrong, thank you for explanation!

PR #1632, opened 2020-06-27
[libp2p-kad] Remove the `Discovered` event.

romanb:
This is a follow-up to (and thus branched off from) https://github.com/libp2p/rust-libp2p/pull/1628 - see the discussion there. It probably shouldn't be rushed in order to give users a chance to voice their concerns here early, in case there are any.


mxinden:
@romanb can you add an entry to https://github.com/libp2p/rust-libp2p/blob/master/protocols/kad/CHANGELOG.md?

romanb:
I will do that as part of an upcoming release PR.

PR #1637, opened 2020-06-30
Document all features on docs.rs

### LINKED ISSUE #1636: Build doc for TokioTcpConfig on docs.rs
docs.rs generates the docs with default features by default (duh), so to get documentation on `TokioTcpConfig`, the feature should be enabled with [metadata](https://docs.rs/about#metadata) in `Cargo.toml`.

The same probably applies to other non-default features, but I'm not yet familiar enough with the project to suggest to enable any other features for docs.rs.

mzabaluev:
To have the documentation built for the optional feature, it has to be enabled in package metadata for docs.rs.

Resolves #1636.


tomaka:
Should we not mention *all* the features?
I'm a bit afraid that passing just `tcp-tokio` disables everything else.

mzabaluev:
> Should we not mention _all_ the features?

It's up to you; I would prefer documentation for all features, unless some are so experimental that the documentation for them has to be hidden.

> I'm a bit afraid that passing just `tcp-tokio` disables everything else.

No, the default features are on unless explicitly disabled.

tomaka:
Ah I've found the documentation: https://docs.rs/about

> It's up to you; I would prefer documentation for all features, unless some are so experimental that the documentation for them has to be hidden.

Yeah, could you please do `all-features = true` instead?

PR #1639, opened 2020-06-30
RequestResponse updates

austinabell:
Made some changes to allow this to be more easily used in an async context. Let me know if you guys are open to this, would love to use.

- The module wasn't exposed, so this primarily adds that in `src/lib`
- Exposes some functionality of the `ResponseChannel` (specifically `send` function). Very limiting having to pass the whole channel back to the libp2p service just to send data through the channel.
- Adds `send_request_with_id` for allowing to send requests with a specific `RequestId` so that I don't have to rely on RequestResponse to assign the ID 
- Makes the `RequestId` integer public so it can be used externally

These changes allow this to be used on the block and chain I work on. Sorry about the rustfmt changes in the file, is a habit, but can change back and reapply changes if you'd like

Thanks!

Edit: I also noticed there isn't a clean way of allowing multiple protocols for the `RequestResponse` service (meaning I had to attach multiple of them in my behaviour) so if I'm not missing anything about how to do this, then I'm happy to tackle that in this PR or in another one if it's preferred to be broken up (@romanb maybe there is some insight you can give here, or opinions since you setup the request-response stuff)


romanb:
> The module wasn't exposed, so this primarily adds that in src/lib

Thanks, I addressed that directly in the imminent [release PR](https://github.com/libp2p/rust-libp2p/pull/1640) in https://github.com/libp2p/rust-libp2p/pull/1640/commits/316b83ed205450652029be829ced2977fbccf059.

> Exposes some functionality of the ResponseChannel (specifically send function). Very limiting having to pass the whole channel back to the libp2p service just to send data through the channel.

In what way is that limiting? There are two reasons for having `RequestResponse::send_response`, one more of an aesthetic nature and the other of a very practical one:

  1. Having `send_request` and `send_response` close together on the same "level" of the API `RequestResponse` makes them easily discoverable and establishes a symmetry. This is mostly an aesthetic point of API design but has the following practical consequence.
  2. Having `send_response` on `RequestResponse` gives it access to the same scope of the implementation internals as `send_request`, namely `RequestResponse`. That means there is a lot of leeway in how the implementation may be changed or augmented in the future without breaking the API. And this is not theoretical: In the course of the implementation of the crate, it went through a few different implementations before ending up with the current use of `oneshot` channels. `ResponseChannel` is supposed to be an opaque "handle" for a channel, in previous implementations it didn't even contain an actual channel in the implementation but just an internal identifier. By providing `ResponseChannel::send` we deprive ourselves of any future changes that may need access to the `RequestResponse` behaviour without breaking the API.

I'm not completely opposed to it, but it comes at non-negligible cost in terms of giving up a good degree of abstraction.

> Adds send_request_with_id for allowing to send requests with a specific RequestId so that I don't have to rely on RequestResponse to assign the ID
> [..]
> Makes the RequestId integer public so it can be used externally

Why do you want to do that? It is intentional that `RequestId`s are only assigned internally by the crate, both to abstract over how they are implemented (numbers, strings, ...) as well as to ensure their uniqueness. Permitting construction of arbitrary request IDs makes the API more fragile in the sense of possible accidental ID collision or reuse and again imposes constraints on what we can change in the future without breaking the API, so this is again about abstraction. Are you missing specific trait implementations for `RequestId` like `Ord` or `PartialOrd`? I suppose those could be added.

austinabell:
First of all, thanks so much for considering this PR and for the detailed response.
 
> 1. Having `send_request` and `send_response` close together on the same "level" of the API `RequestResponse` makes them easily discoverable and establishes a symmetry. This is mostly an aesthetic point of API design but has the following practical consequence.

So small point wrt that, is just that it means (in my case, sorry to bring in personal benefit but seems like a general use case) having to send response and oneshot channel through a channel to the libp2p service to be able to call this function, instead of just passing in the response through the oneshot in the seperate task. I don't see another way around this, but sorry if I'm missing something.

> 2. Having `send_response` on `RequestResponse` gives it access to the same scope of the implementation internals as `send_request`, namely `RequestResponse`. That means there is a lot of leeway in how the implementation may be changed or augmented in the future without breaking the API. And this is not theoretical: In the course of the implementation of the crate, it went through a few different implementations before ending up with the current use of `oneshot` channels. `ResponseChannel` is supposed to be an opaque "handle" for a "channel" in the abstract sense; in previous implementations it didn't even contain an actual channel in the implementation but just an internal identifier. By providing `ResponseChannel::send` we deprive ourselves of any future changes that may need access to the `RequestResponse` behaviour without breaking the API.

Yeah, I totally get that, and was something I had in mind when doing this. As I mentioned before, it's just that it becomes really janky to use in an async context and also impossible to mock. Is there an interface that you would consider, or is this a full stop on this?

> > Adds send_request_with_id for allowing to send requests with a specific RequestId so that I don't have to rely on RequestResponse to assign the ID
> > [..]
> > Makes the RequestId integer public so it can be used externally
> 
> Why do you want to do that? It is intentional that `RequestId`s are only assigned internally by the crate, both to abstract over how they are implemented (numbers, strings, ...) as well as to ensure their uniqueness. Permitting construction of arbitrary request IDs makes the API more fragile in the sense of possible accidental ID collision or reuse and again imposes constraints on what we can change in the future without breaking the API, so this is again about abstraction. Are you missing specific trait implementations for `RequestId` like `Ord` or `PartialOrd`? I suppose those could be added.

Yeah so this point I'm very okay with conceding because it could open the API to be misused. Again my issue is that I need to call this through a separate task, so I can't get the RequestId from when the request is sent. I essentially had the id handling done the same as is in the `RequestResponse` in the separate task, so that I didn't have to use some sort of intermediary to map the IDs. 

The idea behind this is just that if you use the function, you are just accepting that all Ids will be handled externally, which I do see isn't the best for this lib. I don't need any other trait impls on it, but I can add it if you'd like

thomaseizinger:
> As I mentioned before, it's just that it becomes really janky to use in an async context and also impossible to mock.

In my experience, using rust-libp2p in an async/await context can be tricky to wrap your head around. One thing that works for us is to simply wrap the `Swarm` in an `Arc<Mutex>` which means you can pass it around as much as you want.

It is not the cleanest solution but it works.

If you are interested, there is some relevant discussion going on over here on how to avoid the `Arc<Mutex>`: https://github.com/libp2p/rust-libp2p/pull/1630

For your concrete case, having an a reference to the swarm (and thereby to the underlying behaviour) through `Arc<Mutex>` should work. Whichever task gets the `ResponseChannel` just also needs to have an instance of `Arc<Mutex>`.

Make sure to use an async Mutex like the [tokio one](https://docs.rs/tokio/0.2.21/tokio/sync/struct.Mutex.html) though, otherwise you are starving the executor :)

romanb:
> Yeah, I totally get that, and was something I had in mind when doing this. As I mentioned before, it's just that it becomes really janky to use in an async context and also impossible to mock. Is there an interface that you would consider, or is this a full stop on this?

I understand the appeal of being able to send responses wherever you have a `ResponseChannel` without needing access to the `RequestResponse` behavior, I'm just on the fence about the trade-offs in terms of what we lose in implementation flexibility due to the loss of access to `RequestResponse` when sending a response (though that could possibly still be achieved with yet another internal oneshot channel, if necessary, so that the `ResponseChannel` will actually send the response to the behaviour). My suggestion would be that you open a new PR or at least a ticket just for this proposal where it can be discussed further on its own (PRs without `rustfmt` changes please, since we simply don't use it). It would be good to collect some more input from more people before making a decision. This crate is very new so there hasn't been a lot of opportunity yet for integrating it into other projects (e.g. as far as I know it is also supposed to be used in [substrate](https://github.com/paritytech/substrate).

> Yeah so this point I'm very okay with conceding because it could open the API to be misused. Again my issue is that I need to call this through a separate task, so I can't get the RequestId from when the request is sent. I essentially had the id handling done the same as is in the RequestResponse in the separate task, so that I didn't have to use some sort of intermediary to map the IDs.

What is not entirely clear to me is that if you can call `RequestResponse::send_request_with_id` you can also call `RequestResponse::send_request`, so I'm not sure how that is related to separate tasks. Is it just that you want or already have your own notion of request IDs, maybe with a larger scope, that you would want to reuse for `libp2p-request-response`? If the scope of these request IDs is different, it is probably cleaner to have an injective (but if necessary invertable) mapping then from the `RequestResponse` IDs to your own ID space.

austinabell:
> What is not entirely clear to me is that if you can call `RequestResponse::send_request_with_id` you can also call `RequestResponse::send_request`, so I'm not sure how that is related to separate tasks. Is it just that you want or already have your own notion of request IDs, maybe with a larger scope, that you would want to reuse for `libp2p-request-response`? If the scope of these request IDs is different, it is probably cleaner to have an injective (but if necessary invertable) mapping then from the `RequestResponse` IDs to your own ID space.

Yeah the task doesn't call the function directly and there just isn't a clean way to handle those mappings. The main issue is just that this API is very restrictive, especially when it comes to mocking requests, as there is no way to construct this `ResponseChannel` so the logic built on top of this can't be tested. 

That's fine though, if not open to any changes like this I can just fork that crate specifically and just come back to revisit if the API changes to something I can use as the interface should be close enough to switch back

PR #1641, opened 2020-07-02
Tor transport

tcharding:
Add a Tor transport based on the Tokio TCP Transport. This Transport allows users of libp2p to run all TCP traffic over the Tor network.

For the listener side of the application, using this transport, one can pass an onion address to the application. libp2p will pass this address down to the transport via the call to `listen_on`. In order to create a listening socket to redirect the onion service to we must be able to map from the onion `Multiaddr` to the port number to use. In order to do this we add a `HashMap` to the `TorTokioTcpConfig`, the user must set up this map when building the transport.

For the dialer side things are considerably more simple. When the `dial` method is called we set up an IPv4 TCP connection to the locally running Tor instance's SOCKS5 proxy. We then simply return the `TcpStream` that was created.

This works with upgrades and is tested with multiplexing via yamux and authentication via secio.

Patch 1 adds the new Transport. Patch 2 adds an example application to show how the new transport can be used.

Creating as a draft because this PR introduces a lot of duplicate code copied from the TCP transport. I have some ideas how to reduce this but would like to see I'm on the right track first.

Thanks

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -101,6 +101,7 @@ env_logger = "0.7.1"
 [workspace]
 members = [
     "core",
+    "examples/tor",
     "misc/core-derive",
     "misc/multiaddr",
     "misc/multistream-select",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -48,6 +48,7 @@ request-response = ["libp2p-request-response"]
 secio = ["libp2p-secio"]
 tcp-async-std = ["libp2p-tcp", "libp2p-tcp/async-std"]
 tcp-tokio = ["libp2p-tcp", "libp2p-tcp/tokio"]
+tor =  ["libp2p-tor",]
 uds = ["libp2p-uds"]
 wasm-ext = ["libp2p-wasm-ext"]
 websocket = ["libp2p-websocket"]
@@ -90,6 +91,7 @@ libp2p-deflate = { version = "0.20.0", path = "protocols/deflate", optional = tr
 libp2p-dns = { version = "0.20.0", path = "transports/dns", optional = true }
 libp2p-mdns = { version = "0.20.0", path = "protocols/mdns", optional = true }
 libp2p-tcp = { version = "0.20.0", path = "transports/tcp", optional = true }
+libp2p-tor = { version = "0.20.0", path = "transports/tor", optional = true }
 libp2p-websocket = { version = "0.21.0", path = "transports/websocket", optional = true }
 
 [dev-dependencies]
@@ -118,6 +120,7 @@ members = [
     "swarm",
     "transports/dns",
     "transports/tcp",
+    "transports/tor",
     "transports/uds",
     "transports/websocket",
     "transports/wasm-ext"
--- a/examples/tor/Cargo.toml
+++ b/examples/tor/Cargo.toml
@@ -1,0 +1,21 @@
+[package]
+name = "ping-pong"
+version = "0.2.0"
+authors = ["Tobin C. Harding <tobin@coblox.tech>"]
+edition = "2018"
+
+[dependencies]
+anyhow = "1.0"
+clap = "2.32"
+data-encoding = "2.2"
+futures = "0.3"
+futures-timer = "3.0"
+get_if_addrs = "0.5"
+ipnet = "2.3"
+libp2p = { version = "0.20", default-features = false, features = [ "secio", "yamux", "mplex", "dns", "tcp-tokio", "ping"] }
+log = "0.4"
+simple_logger = "1.6"
+socket2 = "0.3"
+structopt = "0.3"
+tokio = { version = "0.2", features = ["rt-threaded", "macros", "tcp"] }
+tokio-socks = "0.2"
--- a/examples/tor/README.md
+++ b/examples/tor/README.md
@@ -1,0 +1,25 @@
+# libp2p over Tor - Ping-Pong
+
+Basic TCP peer to peer connectivity using libp2p over the Tor network.
+
+## Usage
+
+0. Install Tor
+1. Configure an onion service using the Tor run file. You can use
+`./torrc` and run `tor` using `sudo /usr/bin/tor --defaults-torrc tor-service-defaults-torrc -f torrc --RunAsDaemon 0`
+2. Once you have run `tor` for the first time get the onion address
+   from the `hostname` file (if you used the `tor` invocation above
+   this will be in `/var/lib/tor/hidden_service/hostname`). Set the
+   onion address `const ONION` in `main.rs`.
+3. Set the log level in `main.rs` if you wish.
+
+Now to run this demo application run tor in one terminal, the listener
+in another terminal, and the dialer in a third terminal.
+
+- Run the listener with: `ping-pong --listener`.
+- Run the dialer with: `ping-pong --dialer`.
+
+See `ping-pong --help` for more information.
+
+
+Tested on Ubuntu LTS 18 using Tor version 0.4.3.5
--- a/examples/tor/src/cli.rs
+++ b/examples/tor/src/cli.rs
@@ -1,0 +1,17 @@
+use structopt::StructOpt;
+
+#[derive(Debug, StructOpt)]
+#[structopt(name = "ping-pong", about = "libp2p ping-pong application over Tor.")]
+pub struct Opt {
+    /// Run as the dialer i.e., do the ping
+    #[structopt(short, long)]
+    pub dialer: bool,
+
+    /// Run as the listener i.e., do the pong (default)
+    #[structopt(short, long)]
+    pub listener: bool,
+
+    /// Onion mulitaddr to use (only required for dialer)
+    #[structopt(long)]
+    pub onion: Option<String>,
+}
--- a/examples/tor/src/lib.rs
+++ b/examples/tor/src/lib.rs
@@ -1,0 +1,151 @@
+mod cli;
+pub mod transport;
+
+pub use cli::Opt;
+
+use std::{
+    io,
+    net::{Ipv4Addr, SocketAddr, SocketAddrV4},
+    pin::Pin,
+    task::{Context, Poll},
+    time::Duration,
+    collections::HashMap,
+};
+
+use anyhow::{Result};
+use futures::{future, prelude::*};
+use libp2p::{
+    core::{
+        either::EitherError,
+        muxing::StreamMuxerBox,
+        transport::{boxed::Boxed, timeout::TransportTimeoutError},
+        upgrade::{SelectUpgrade, Version},
+        UpgradeError,
+    },
+    dns::{DnsConfig, DnsErr},
+    identity,
+    mplex::MplexConfig,
+    ping::{Ping, PingConfig},
+    secio::{SecioConfig, SecioError},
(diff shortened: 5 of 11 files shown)


burdges:
As a rule, you'd want a Tor transport to exist in two layers:
1. a SOCKS5 transport that's usable for `ssh -D`, etc. and reused by 
2. a tor transport that also invokes the Tor control port as appropriate https://crates.io/search?q=tor%20 and
3. maybe later a stenographic transport similar to Tor pluggable transports but does not go through Tor. 

You need the Tor control port for things like creating a .onion address, telling Tor you want a new circuit whenever you want a new identity, etc.  It's also possible you'll want the Tor control port to be managed by an outside process if you fear the shear size of some code base resting on libp2p directly.

tcharding:
Thanks for your response, I don't really understand some of what you say. Commenting in line.

> As a rule, you'd want a Tor transport to exist in two layers:
> 
> 1. a SOCKS5 transport that's usable for `ssh -D`, etc. and reused by

Do you mean by this that we should have a separate transport for dialing through a SOCKS5 proxy?

> 2. a tor transport that also invokes the Tor control port as appropriate [crates.io/search?q=tor%20](https://crates.io/search?q=tor%20) and

At first I thought this as well, (see https://github.com/comit-network/ping-pong/tree/v0.1). I later realized that leaving the configuration of onion services up to the user via a torrc file is way simpler. There is then no need for libp2p to know about the Tor control port. It can be used, if desired, at a higher level within the application. This is beneficial also because, IIUC, one can have only one connection to the control port so if we use it in libp2p applications will no be able to use it.

> 3. maybe later a stenographic transport similar to Tor pluggable transports but does not go through Tor.

I don't understand this bit sorry. Are you referring to Tor over libp2p as apposed to libp2p over Tor?

> You need the Tor control port for things like creating a .onion address, telling Tor you want a new circuit whenever you want a new identity, etc. It's also possible you'll want the Tor control port to be managed by an outside process if you fear the shear size of some code base resting on libp2p directly.

Please say if I have missed your points. Thanks for looking at this.

burdges:
> Do you mean by this that we should have a separate transport for dialing through a SOCKS5 proxy?

No no, one SOCKS5 proxy is fine, but..

> At first I thought this as well, (see https://github.com/comit-network/ping-pong/tree/v0.1). I later realized that leaving the configuration of onion services up to the user via a torrc file is way simpler. 

Yes, it's mostly different layer for the control port I guess, fair enough.  You do influence outgoing circuit usage with the account name and password though.  It's possible your SOCKS5 proxy type exposes the user name and password in two ways, the usual way, and the tor way, or maybe you can just expose them as a closure or something. 

> I don't understand this bit sorry. Are you referring to Tor over libp2p as apposed to libp2p over Tor?

No. Tor pluggable transports are simply something else that uses SOCK5.  They need not necessarily send traffic over Tor.  Ignore this for now I guess.  :)

tcharding:
Thanks @burdges, I'm not sure from your comments if you'd like to see any changes in this PR. Please excuse me if I am acting brain dead.

burdges:
You convinced me it's fine for now.  You should double check that you understand how tor uses the username and password.  If you did not see that in the docs, then I'll ask mike perry or something.

tcharding:
> You should double check that you understand how tor uses the username and password. 

I searched the Tor docs for `socks` and read around every (75) resulting instance of that term. I'm not sure how this effects this PR though. I'm intentionally leaving all Tor config up to the user. For the listener side of applications that use this transport that means configuring the onion service using the Tor run file. For the dialer we just connect to the SOCKS5 proxy. Any other configuration options should be handled by the user also and AFAICT shouldn't adversely effect usage of this transport as long as the Tor instance starts with the required functionality i.e., SOCKS5 port open.

burdges:
I'll see what I find and let you know.

burdges:
In short, "socks connections with different usernames or passwords are kept on separate tor circuits by default", which tor browser uses.  I think a socks module could just expose a username and password option and the code above it could know to use this to signal tor if it needs a new identity.  I suppose most code using libp2p would be complex things that's kinda hard to audit well enough to exploit this..

tcharding:
Cool, thanks.

tcharding:
This ones been open for two weeks now, is there anything I need to do to help with review please?

burdges:
Sorry, LGTM :)

romanb:
Thanks for your effort. I think it generally does no harm to explicitly request reviews from specific people - even if they then have to decline or just forward the request to others. Speaking for myself, I didn't comment so far because I only have superficial knowledge of TOR and thus don't really feel qualified to do a proper review, but I think I can give some initial high-level feedback:

  * I'm a bit irritated by `examples/tor` being a full-fledged crate with `Cargo.toml` and all. I don't think that's typical for examples. I think it would be good to turn it into a "regular" example like the others (examples can have multiple files). I'm particularly irritated by the presence of `examples/tor/src/transport.rs`, which on first glance looks like an almost duplicate of `transports/tor/src/lib.rs`.

  * If tokio cannot be made an implementation detail of `libp2p-tor`, i.e. cannot be fully encapsulated, then I think it would be preferable to either: a) Rename the crate to `libp2p-tor-tokio` or b) provide both `tokio` and `async-std` support similar to `libp2p-tcp`, e.g. via feature flags.

  * The version of the new crate should start with `0.1.0`.

I think an important question for this PR, which introduces a new crate, is: Do you intend to commit yourself to ongoing maintenance of this crate as part of the `libp2p` project or do you see this PR as a handover of this code? Since this project has in general very few maintainers, inclusion of new crates obviously has a much better chance if the new crate comes along with a new maintainer. There is also the option of keeping this crate as a dedicated project of yours, outside of the `libp2p` umbrella project, which you probably already considered? I myself am not opposed to inclusion of new crates, especially if they come with new maintainers, but one needs to be aware that if a crate like this happens to become incompatible with future versions of other libp2p crates and is not in the focus of the few core maintainers, the task of updating it may fall on you and if it ends up being unmaintained it may get its dependencies pinned to older versions and hence "left behind" and potentially removed again if no maintainers volunteer.

tcharding:
> Thanks for your effort. I think it generally does no harm to explicitly request reviews from specific people - even if they then have to decline or just forward the request to others.

Point noted, thanks.  Will do in future.

> Speaking for myself, I didn't comment so far because I only have superficial knowledge of TOR and thus don't really feel qualified to do a proper review, but I think I can give some initial high-level feedback:

I appreciate the feedback, thanks.

> * I'm a bit irritated by `examples/tor` being a full-fledged crate with `Cargo.toml` and all. I don't think that's typical for examples. I think it would be good to turn it into a "regular" example like the others (examples can have multiple files). I'm particularly irritated by the presence of `examples/tor/src/transport.rs`, which on first glance looks like an almost duplicate of `transports/tor/src/lib.rs`.

This is a fail by me. I should have presented this PR differently. My intention with having two commits, the example second, was for the example to be validation of the initial commit. I did not expect it to be merged as is. My mistake for not making that explicit.

> * If tokio cannot be made an implementation detail of `libp2p-tor`, i.e. cannot be fully encapsulated, then I think it would be preferable to either: a) Rename the crate to `libp2p-tor-tokio` or b) provide both `tokio` and `async-std` support similar to `libp2p-tcp`, e.g. via feature flags.

Points noted. If this new transport is deemed valuable enough for merge consideration I can definitely look into an `async-std` version. Otherwise I will do as you suggest.

> * The version of the new crate should start with `0.1.0`.

Now here I _was_ trying to fit in with the project, I got it wrong :) Will change to `0.1.0`. Thanks.
 
> I think an important question for this PR, which introduces a new crate, is: Do you intend to commit yourself to ongoing maintenance of this crate as part of the `libp2p` project or do you see this PR as a handover of this code? Since this project has in general very few maintainers and a lot of crates, inclusion of new crates obviously has a much better chance if the new crate comes along with a new maintainer. There is also the option of keeping this crate as a dedicated project of yours, outside of the `libp2p` umbrella project, which you probably already considered? I myself am not opposed to inclusion of new crates, especially if they come with new maintainers, but one needs to be aware that if a crate like this happens to become incompatible with future versions of other core libp2p crates and is not in the focus of the few core maintainers, the task of updating it may fall on you and if it ends up being unmaintained it may get its dependencies pinned to older versions and hence "left behind" and potentially removed again if no maintainers volunteer.

I am happy to maintain this crate. I wrote this as part of my day job, we are using it currently there and I can allocate time for maintenance. I'd prefer to see it upstream rather than as a separate project so others can find it easily. Thanks for pointing this out, I had not thought of these considerations until you commented.

I going to convert this to draft and take time to fix the example and look into async-std.

Thanks again @romanb.

tcharding:
On further thought I am now leaning towards making this a separate crate and putting it up on crates.io. Reasoning is as follows
- Taking me on as a maintainer is a risk to the libp2p project since I'm an unknown quantity.
- A fully fledged example is not easy to write in the same fashion as the rest of the libp2p examples, this smells to me like there shouldn't even be an example but the code is useful for others to see, a separate crate on crates.io does not have this conflict
- The async_std/tokio stuff is not trivial to sort out, until its sorted it feels wrong to put this in libp2p since we only just got libp2p to be async library agnostic
- Adding this to libp2p correctly would really mean refactoring the current transports so there is not all this duplicate code, again this is difficult when the tokio/async_std stuff is not sorted out
- This is not part of the 'spec', the go-onion-transport is hosted in a separate location to the go implementation of libp2p
- There is no real harm in having it as a separate crate, hosting it on crates.io for a year or two will give us time to see if it is useful and also will verify that I am a competent maintainer of the crate then we can merge it into rust-libp2p if its deem useful

I'll leave this up as draft for a while for any comments then proceed as outlined above.

Any and all suggestions welcome and most appreciated.

tomaka:
As part of the crate naming: I would expect a crate named `libp2p-tor` to embed a Tor client itself.
Would it make sense to instead name this `libp2p-tor-ipc` (or something similar)?

burdges:
We've no special handling for tor here, so it's libp2p-socks5 really.

If you want to control over which tor circuit your traffic flows, then you use different usernames or passwords when opening socks5 connections to tor, but this necessarily remains a decision for application above libp2p since only the application knows which data exposes the linkages between tor circuits.  And the username and password could take their normal meaning in socks5 if using another socks5 server besides tor.

burdges:
As an aside, Polkadot itself should soon-ish become quite a UDP heavy protocol.  I know socks5 supports UDP, so did that get implemented here?  I'm afraid however the two most common socks5 proxies `tor` and `ssh -D` do not support UDP because they push everything over TCP streams, but other dedicated socks5 proxies do support UDP.  I donno if I2P handles UDP acceptably or if it's simply too hard to anonymize.  

In any case, I'd foresee tor being used primarily for parachains, especially tracking zk parachains, but maybe some unplanned future sentry node or remote signer configuration would've validators talking to their own sentry nodes over only TCP/`tor`/`ssh -D`.

tcharding:
How about crate name `libp2p-tokio-socks5`? The async-std stuff would then have to be done in a separate repo/crate. The examples could then show how to use it with Tor.

tcharding:
Closing. Moving to https://github.com/comit-network/rust-libp2p-tokio-socks5

tcharding:
Now published on crates.io: https://crates.io/crates/libp2p-tokio-socks5

PR #1642, opened 2020-07-02
Re-release websockets 0.20.1 as 0.21.

romanb:
This PR suggests to yank `libp2p-websockets-0.20.1` and re-release it as `0.21.0` due to the minor dependency bump in its `libp2p-core` dependency. Otherwise the patch upgrade of `libp2p-websockets` "pulls in" unwanted newer, incompatible versions of other `libp2p` crates.


romanb:
CC @tomaka

PR #1644, opened 2020-07-02
Improve the code of the WebSocket browser binding

tomaka:
Reworks a bit the binding between `libp2p-wasm-ext` and the browser's `WebSocket` object.

I've used https://html.spec.whatwg.org/multipage/web-sockets.html to review the current code and make it more robust.

Notably:

- We now directly ask the browser to produce `ArrayBuffer` objects rather than obtaining `Blob` objects and converting them manually into `ArrayBuffer`.
- The threshold for sending more data is now 8kiB. In other words, as long as the out buffer (the one handled by the browser and whose length is accessed through `bufferedAmount`) is smaller than 8kiB, we report to the upper layers that we are ready to accept more data.
- `onclose` and `onerror` now unconditionally irreversibly shut down both the opening attempt and the incoming stream of bytes, without checking the state of the connection.
- Sending now returns a `Promise` that repeatedly checks `WebSocket.readyState` and returns an error if it's no longer open.

The changes in `lib.rs` are only documentation clarifications, no change to the API itself has been made.

### DIFF
--- a/transports/wasm-ext/CHANGELOG.md
+++ b/transports/wasm-ext/CHANGELOG.md
@@ -1,4 +1,4 @@
-# 0.20.1 [????-??-??]
+# 0.20.1 [2020-07-06]
 
 - Improve the code quality of the `websockets.js` binding with the browser's `WebSocket` API.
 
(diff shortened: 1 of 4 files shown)


tomaka:
I've tested this with Polkadot-in-the-browser on both Chrome and Firefox.

tomaka:
This PR doesn't really fix any specific known bug, it's just a general clean-up. The one bugfix is that, when connecting to a peer fails, the previous code would assume that `onerror` is called, while in reality the browser calls `onclose` according to the specs.

It's not particularly urgent, but there's also no reason to not release it right now.

PR #1645, opened 2020-07-03
Allow gossipsub Topic construction from any string

mzabaluev:
Extend the constructor to allow any `Into<String>` input.
Unfortunately, this invalidates the current call sites with `"blah".into()`, so this should only be rolled into a new 0.x release.

### DIFF
--- a/examples/gossipsub-chat.rs
+++ b/examples/gossipsub-chat.rs
@@ -72,7 +72,7 @@ fn main() -> Result<(), Box<dyn Error>> {
     let transport = libp2p::build_development_transport(local_key)?;
 
     // Create a Gossipsub topic
-    let topic = Topic::new("test-net".into());
+    let topic = Topic::new("test-net");
 
     // Create a Swarm to manage peers and events
     let mut swarm = {
--- a/examples/ipfs-private.rs
+++ b/examples/ipfs-private.rs
@@ -168,7 +168,7 @@ fn main() -> Result<(), Box<dyn Error>> {
     let transport = build_transport(local_key.clone(), psk);
 
     // Create a Gosspipsub topic
-    let gossipsub_topic = gossipsub::Topic::new("chat".into());
+    let gossipsub_topic = gossipsub::Topic::new("chat");
 
     // We create a custom network behaviour that combines gossipsub, ping and identify.
     #[derive(NetworkBehaviour)]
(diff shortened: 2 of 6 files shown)


AgeManning:
Hey, yep, I have noticed this. 

I've started on gossipsub 1.1 (have had other things come up, but will get back around to it soon). In that update, I've modified the topics to be a bit more ergonomic, so have also addressed this. See for example: https://github.com/sigp/rust-libp2p/commit/238e30565758fbe1aafd3a3170399eacbd63c996#diff-e508e1fe2df1aa713d7d22989bd2b1a8

I'm happy with this change, but potentially we may re-address with the gossipsub 1.1 update (which should be soon, order of weeks, I promise). 

Depends on when you need it @mzabaluev. If you can wait for the 1.1 update, we can roll it in and can have your input on how you would like the topics to be handled there.

mzabaluev:
Thanks, @AgeManning, I'm looking forward to gossipsub 1.1!

This PR is just a quality-of-life improvement, it's in no way critical for me.

AgeManning:
If that's the case, I suggest we do all the commits into the 1.1 update, save multiple releases. 

I'll ping you when I make the PR to master and get your input on the topics, if that works for you.

mriise:
@AgeManning though a bit off-topic (similar vein), would it be possible to implement multihash (and potentially multicodec) digests of strings instead of strictly SHA?

AgeManning:
@mriise - We can. The issue is interoperability. Waaay back, Rust used hashing and the go (spec) implementation didn't. See this issue: https://github.com/libp2p/rust-libp2p/issues/473

I think go now has configurable hashing (although I had a quick skim and couldn't find it). The current hash implementation was ported from the old version and is not interoperable (I only included it as an alternative as it was ideally the "correct" way of doing topic subscriptions". 

If we added multihashes then only the rust implementation would be able to communicate (potentially other clients could subscribe to the topics by supplying the resulting hash). 

I think this is more of a spec issue. If its important I could probably add it in as a configuration parameter, but it should probably be raised as an issue on the libp2p specs repo. If it get's spec'd then all clients should implement it. 

An alternative way would be to implement it at the application layer. And use the resulting topic hash as a topic "string".

mxinden:
@mzabaluev I am closing here as it seems to be not-critical and already included in the 1.1 effort of Age. Feel free to ping me in case you would still like to have it included earlier.

PR #1646, opened 2020-07-06
[libp2p-swarm] Ignore a node's own addresses on dialing.

romanb:
Dialing attempts of a local node to one of its own addresses for what appears to be a different peer ID are futile and bound to fail with an `InvalidPeerId` error. To avoid such futile dialing attempts, filter out the node's own addresses from the addresses
reported by the `NetworkBehaviour` for any peer.

There can be a few reasons why a `NetworkBehaviour` may think an address belongs to a different peer, e.g.:

  1. In the context of e.g. `libp2p-kad`, the local node may have changed its network identity (e.g. key rotation)
     and "discovers" its former identity in the DHT, with the same address(es).
  2. Another peer may erroneously or intentionally, possibly even maliciously, report one of the local node's addresses as its own, making the node try to connect to itself.

Note: It is already the case (i.e. before this PR) that an attempt to dial the local peer ID is an error. This PR deals with the situation of trying to dial one of the local node's own addresses for what appears to be a different peer ID. Since many such situations are not necessarily programmer error, or it is at least unreasonably cumbersome to have each `NetworkBehaviour` track the local node's addresses, which the `Swarm` already does, just to avoid returning a local address for a different peer ID from `addresses_of_peer`, it seems sensible to just ignore such addresses directly in the `Swarm` which already keeps track of these.

Relates to https://github.com/paritytech/stakingops-issues/issues/18.

### DIFF
--- a/swarm/CHANGELOG.md
+++ b/swarm/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 - Documentation updates.
 
+- Ignore addresses returned by `NetworkBehaviour::addresses_of_peer`
+that the `Swarm` considers to be listening addresses of the local node. This
+avoids futile dialing attempts of a node to itself, which can otherwise
+even happen in genuine situations, e.g. after the local node changed
+its network identity and a behaviour makes a dialing attempt to a
+former identity using the same addresses.
+
 # 0.20.0 [2020-07-01]
 
 - Updated the `libp2p-core` dependency.
--- a/swarm/src/lib.rs
+++ b/swarm/src/lib.rs
@@ -382,13 +382,17 @@ where TBehaviour: NetworkBehaviour<ProtocolsHandler = THandler>,
     /// dialing attempt or `addresses_of_peer` reports no addresses, `Ok(false)`
     /// is returned.
     pub fn dial(me: &mut Self, peer_id: &PeerId) -> Result<(), DialError> {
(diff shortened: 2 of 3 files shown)


tomaka:
I fear that this is attackable.
The list of external addresses of the local node is built by asking nodes we are connected to which IP they see for the connection.

A malicious peer could report the IP of the attack target instead of the actual IP, and prevent the attack target from being dialed.

romanb:
> I fear that this is attackable.
> The list of external addresses of the local node is built by asking nodes we are connected to which IP they see for the connection.
> 
> A malicious peer could report the IP of the attack target instead of the actual IP, and prevent the attack target from being dialed.

I see, but that problem does not affect the `listened_addrs`, right? To cover cases like changes of node identity, I don't think the external addresses are even required.

romanb:
> I don't think the external addresses are even required to check.

To elaborate on that, my understanding is that "external" addresses differ from the listener addresses mostly due to NAT, and as far as I'm aware dialing your own "external" address won't work anyway, unless the NAT-device supports something like [hairpinning](https://en.wikipedia.org/wiki/Hairpinning)), and thus won't lead to "invalid peer ID" errors. So just checking the `listened_addrs` should already be an improvement. I tested this with two local nodes, one of which keeps restarting with new node identities and thereby discovering all its former identities in the DHT. With these changes, it won't attempt to dial these and thus no spurious `InvalidPeerId` errors occur.

PR #1649, opened 2020-07-07
protocols/kad: Implement NetworkBehaviour::inject_address_change

mxinden:
With 826f513 a `StreamMuxer` can notify that the address of a remote
peer changed. This is needed to support the transport protocol QUIC as
remotes can change their IP addresses within the lifetime of a single
connection.

This commit implements the `NetworkBehaviour::inject_address_change`
handler to update the Kademlia routing table accordingly.

Follow up for https://github.com/libp2p/rust-libp2p/pull/1621.


protocols/kad/src/behaviour.rs
```diff
@@ -1449,6 +1449,34 @@ where
...
+        _: &ConnectionId,
+        old: &ConnectedPoint,
+        new: &ConnectedPoint
+    ) {
+        // Only addresses of outbound connections are inserted into the routing
+        // table in `inject_connection_established`. Thereby only address
+        // changes of outbound connections require updating.
+        let new_address = match new {
+            ConnectedPoint::Dialer { address } => Some(address.clone()),
+            ConnectedPoint::Listener { .. } => return,
```
romanb, line 1464: The reason `inject_connection_established` only directly tries to attempt insertion into the routing table for `Dialer` addresses, is that these are known to be "routable" / "dialable". In contrast, the listener address only _may_ be dialable and we emit an `UnroutablePeer` event. User code may subsequently add this address to the routing table if it is deemed dialable. Whether the `Listener` address is routable / dialable in general depends on the transport protocol. With TCP it isn't, as it contains a connection-specific ephemeral client port number but in general it may be (e.g. with QUIC?). So I think the safer approach here for `inject_address_change` is: If the `old` address is in the routing table for `peer`, replace it with `new`, regardless of whether it is a `Dialer` or `Listener` address. It is probably also fine to always insert `new` even if `old` is not present (but only if the peer is already in the routing table, of course).

protocols/kad/src/behaviour.rs
```diff
@@ -1449,6 +1449,34 @@ where
...
+        };
+
+        // First insert the new address, then remove the old address. A peer
+        // with a single address might lose its slot within the routing table if
+        // one removes the old address first.
+        self.connection_updated(peer.clone(), new_address, NodeStatus::Connected);
```
romanb, line 1470: I don't think you can do this, because `connection_updated` with an address and `NodeStatus::Connected` will even insert the peer into the routing table if it is not present. I don't think `inject_address_change` should do that.

mxinden:
Thanks for the review @romanb!

With my recent changes a new address is only added to the `Addresses` of a peer iff the old address was present.

> It is probably also fine to always insert new even if old is not present (but only if the peer is already in the routing table, of course).

Can you expand on the scenario where the old address would not be present even though we are connected to the node?

core/src/connection.rs
```diff
@@ -133,6 +133,13 @@ impl ConnectedPoint {
         }
     }
 
+    pub fn get_remote_address(&self) -> &Multiaddr {
```
romanb, line 136: A short doc-comment may be good, possibly with a hint / reminder that this address may be "private" to the connection, i.e. may not be a listening address of the remote and hence not usable for establishing new connections.

protocols/kad/src/behaviour.rs
```diff
@@ -1449,6 +1449,46 @@ where
...
+        } else {
+            debug!(
+                "Address '{}' not replaced with '{}' for peer '{}' as peer is not present in the \
+                 routing table.",
+                old, new, peer,
+            );
+        }
+
+        // Update query address cache.
+        for query in self.queries.iter_mut() {
```
romanb, line 1481: I don't think there is a point in doing this, but I may be mistaken. The query address "cache" is used to make sure when a query wants to dial a certain address  that often may not (yet) be in the routing table, that address is returned by `addresses_of_peer`. Addresses in the routing table are always returned from `addresses_of_peer`. So when the new address is successfully put in the routing table, it will be considered for dialing. Please correct me if I'm wrong.

romanb:
> Can you expand on the scenario where the old address would not be present even though we are connected to the node?

I don't remember what I had in mind at the time. Your new approach seems fine to me.

protocols/kad/src/behaviour.rs
romanb: > Given two connected nodes: local node A and remote node B. Say node B is not in node A's routing table. Additionally node B is part of the QueryInner::addresses list of an ongoing query on node A. Say Node B triggers an address change and then disconnects. Later on the earlier mentioned query on node A would like to connect to node B. Without replacing the address in the QueryInner::addresses set node A would attempt to dial the old and not the new address.

Fair enough, I guess given the right timing, this situation can occur. I still wonder though whether this is worth this special attention and effort. Maybe there is no good reason, but changing the addresses that a query discovered like this seems strange to me. At least it adds a subtle twist to reasoning about which addresses a query uses. The other aspect is that, depending on how often addresses change like this (certainly increasingly often the more peers you are connected to), iterating through all discovered addresses of that peer in all currently ongoing queries seems like a search for a needle in a haystack, if there is substantial traffic. Neither of these points is a good technical reason against doing this, I just had to express that I feel a bit uncomfortable about it and why that is. I guess the decision is yours, ultimately.

mxinden: > At least it adds a subtle twist to reasoning about which addresses a query uses.

That is a good point.

Still I would prefer to uphold correctness and revisit the decision once we have a transport protocol that supports address changes and the given logic proves to be a performance problem. I have included a note in https://github.com/libp2p/rust-libp2p/pull/1649/commits/ad81fd6146853d9fa08148a70705891317eaaf29 summarizing this discussion.

mxinden:
Thank you for the help @romanb.

PR #1651, opened 2020-07-08
Improve muxer docs

Demi-Marie:


core/src/muxing.rs
```diff
@@ -91,14 +91,19 @@ pub trait StreamMuxer {
     /// is ready to be polled, similar to the API of `Stream::poll()`.
     /// Only the latest task that was used to call this method may be notified.
     ///
+    /// Implementors may rely on this method being called soon after the current
+    /// task has been notified. It is thus safe to use this method to perform
```
tomaka, line 95:
```suggestion
    /// task has been notified. It is thus possible to use this method to perform
```

The word "safe" means something specific in Rust.

core/src/muxing.rs
```diff
@@ -91,14 +91,19 @@ pub trait StreamMuxer {
     /// is ready to be polled, similar to the API of `Stream::poll()`.
     /// Only the latest task that was used to call this method may be notified.
     ///
+    /// Implementors may rely on this method being called soon after the current
```
tomaka, line 94:
```suggestion
    /// Implementors may rely on this method being called soon after the registered
```

core/src/muxing.rs
```diff
@@ -91,14 +91,19 @@ pub trait StreamMuxer {
     /// is ready to be polled, similar to the API of `Stream::poll()`.
     /// Only the latest task that was used to call this method may be notified.
     ///
+    /// Implementors may rely on this method being called soon after the registered
+    /// task has been notified. It is thus possible to use this method to perform
+    /// background work, such as processing incoming packets and polling timers.
```
twittner, line 96: Why do we need to commit to this? The PR title is "Improve muxer docs", but this bit promises specific API guarantees that have not been discussed anywhere.

core/src/muxing.rs
```diff
@@ -91,14 +91,19 @@ pub trait StreamMuxer {
...
     /// An error can be generated if the connection has been closed.
     fn poll_event(&self, cx: &mut Context) -> Poll<Result<StreamMuxerEvent<Self::Substream>, Self::Error>>;
 
     /// Opens a new outgoing substream, and produces the equivalent to a future that will be
     /// resolved when it becomes available.
     ///
     /// The API of `OutboundSubstream` is totally opaque, and the object can only be interfaced
-    /// through the methods on the `StreamMuxer` trait.
+    /// through the methods on the `StreamMuxer` trait. Interacting with `OutboundSubstream`
+    /// directly has undefined results.
```
twittner, line 106: Again, what is the motivation for bringing up undefined behaviour here?

core/src/muxing.rs
DemiMarie-parity: @tomaka has told me this privately, and his work on QUIC support requires it.  More generally, any muxer that needs to have background work done needs this.

core/src/muxing.rs
DemiMarie-parity: It isn’t actually UB, but it might corrupt the internal state of the muxer and cause future misbehavior.

core/src/muxing.rs
DemiMarie-parity: My own implementation of QUIC support did not need it, but had to spawn a background task for each muxer, which is unfortunate.

core/src/muxing.rs
twittner: Since `OutboundSubstream` is an unconstrained type parameter it is up to the `StreamMuxer` impl to define what interactions are possible. The trait can not and should not anticipate what interactions lead to "misbehavior".

DemiMarie-parity: Fixed.

romanb:
Superseded by https://github.com/libp2p/rust-libp2p/commit/86d228122284cd678dc431361ae544e7bdba1750.

PR #1652, opened 2020-07-08
Updating github actions to use actions-rs/cargo for running cargo commands

leanthebean:
I noticed with the latest cargo (cargo 1.44.1) the latest version of libp2p (0.21.1) was not compiling due to this error:

Screen Shot 2020-07-08 at 2 25 10 PM

and yet the CI github action to run tests was not catching it (https://github.com/libp2p/rust-libp2p/actions) and gave a green checkmark to a bunch of commits that failed to compile locally. 

This PR isn't a fix for the compile error, but a fix to the github actions workflow so that these errors are caught in the future.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,9 +28,15 @@ jobs:
         path: target
         key: cargo-build-target-${{ hashFiles('Cargo.toml') }}
     - name: Run tests, with no feature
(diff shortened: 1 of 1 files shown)


leanthebean:
Actually, it looks like the existing CI is indeed failing like it's supposed to (https://github.com/libp2p/rust-libp2p/runs/851012006?check_suite_focus=true)

PR #1655, opened 2020-07-09
Update rustls dependency.

romanb:
The `async-tls` patch release `0.7.1` bumped its `rustls` dependency from `0.17` to `0.18`. `libp2p-websocket` has an explicit dependency on `rustls-0.17`, thus both versions get be pulled in resulting in problems like
```rust
error[E0277]: the trait bound `async_tls::connector::TlsConnector: std::convert::From<std::sync::Arc<rustls::client::ClientConfig>>` is not satisfied
  --> transports/websocket/src/tls.rs:73:47
   |
73 |             client: Arc::new(client_config()).into(),
   |                                               ^^^^ the trait `std::convert::From<std::sync::Arc<rustls::client::ClientConfig>>` is not implemented for `async_tls::connector::TlsConnector`
   |
   = help: the following implementations were found:
             <async_tls::connector::TlsConnector as std::convert::From<std::sync::Arc<rustls::client::ClientConfig>>>
   = note: required because of the requirements on the impl of `std::convert::Into<async_tls::connector::TlsConnector>` for `std::sync::Arc<rustls::client::ClientConfig>`
```
See e.g. [this build](https://github.com/libp2p/rust-libp2p/pull/1651/checks?check_run_id=851012006). To fix this, this PR updates `rustls` to `0.18` for `libp2p-websocket` as well. Given that `rustls`-types are at least in part exposed on the `async-tls` API (as witnessed above), it may have been more appropriate for the minor `rustls` version bump to be released as `async-tls-0.8` :man_shrugging:

### DIFF
--- a/transports/websocket/CHANGELOG.md
+++ b/transports/websocket/CHANGELOG.md
@@ -1,6 +1,6 @@
-# ?.?.? [????-??-??]
+# 0.21.1 [2020-07-09]
 
-- Update `rustls` dependency to `0.18`.
+- Update `async-tls` and `rustls` dependency.
 
 # 0.21.0 [2020-07-02]
 
--- a/transports/websocket/CHANGELOG.md
+++ b/transports/websocket/CHANGELOG.md
@@ -1,3 +1,7 @@
+# ?.?.? [????-??-??]
+
+- Update `rustls` dependency to `0.18`.
+
 # 0.21.0 [2020-07-02]
 
 - Update `libp2p-core`.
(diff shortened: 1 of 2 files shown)


romanb:
Does someone know whether `libp2p-websocket` itself fully encapsulates its use of `rustls`? That would determine whether this can be a patch release or not.

tomaka:
Yes we don't expose anything from `rustls` in the external API (which to me is a good programming practice), so this can be patch.

romanb:
Since `async-tls-0.7.1` has been yanked, this PR turns into a dependency upgrade for both `rustls` (0.18.0) and `async-tls` (0.8). Neither seems to be exposed on the `libp2p-websocket` API, so this will go out as a patch release upon merge.

Issue #1631, opened 2020-06-26
libp2p-noise incompatible with go-libp2p-noise


tomaka:
Our implementation of noise isn't compatible with go-libp2p.
While Noise messages do function properly, the problem lies in the protobuf message exchange inside of the Noise messages.

rust-libp2p expects the protobuf handshake messages to be prefixed with their length (as seen [here](https://github.com/libp2p/rust-libp2p/blob/bd5e81fedbc5aa9bf941a8ba5da02dee26f291e3/protocols/noise/src/io/handshake.rs#L362-L364) and [here](https://github.com/libp2p/rust-libp2p/blob/bd5e81fedbc5aa9bf941a8ba5da02dee26f291e3/protocols/noise/src/io/handshake.rs#L401-L402)) while go-libp2p doesn't send any length.

I have managed to successfully connect to go-libp2p after I removed [sending the prefix](https://github.com/libp2p/rust-libp2p/blob/bd5e81fedbc5aa9bf941a8ba5da02dee26f291e3/protocols/noise/src/io/handshake.rs#L401-L402) and temporarily replacing [these lines](https://github.com/libp2p/rust-libp2p/blob/bd5e81fedbc5aa9bf941a8ba5da02dee26f291e3/protocols/noise/src/io/handshake.rs#L362-L364) with a hard-coded length.

burdges:
https://eli.thegreenplace.net/2011/08/02/length-prefix-framing-for-protocol-buffers

tomaka:
In this situation we know that the first Noise messages exchanged consist entirely in a protobuf message and nothing else.
Noise messages themselves are length-delimited, so I suppose that it wasn't deemed necessary to indicate the length again inside of the encrypted messages themselves.

tomaka:
> In this situation we know that the first Noise messages exchanged consist entirely in a protobuf message and nothing else.

Although this isn't necessarily the case in the absolute as far as I understand, and length-prefixing the protobuf messages would probably have been a good idea.

pawanjay176:
Isn't the noise spec wire format saying that we _do_ require length prefix for handshake payloads as well?!
https://github.com/libp2p/specs/tree/master/noise#wire-format

sinkingsugar:
Well, the spec got changed. go follows it as now nim too:
https://github.com/status-im/nim-libp2p/commit/7852c6dd0f00f40463227a39a9ec6ea5782c5d36

tomaka:
[...]

PR #1658, opened 2020-07-09
Spec-compliant noise handshake payloads (Step 1).

romanb:
Context: https://github.com/libp2p/rust-libp2p/issues/1631.

This is the first step of a three-step process to address the above issue. In this step, support for reading noise handshake payloads without an additional length prefix is added, falling back to attempting a decoding with length prefix on failure. Length prefixes are still sent in this step. _Hence interoperability with other libp2p implementations is not yet achieved after with step._

To achieve a better separation of handshake and transport I/O, the `NoiseFramed` type (new module `io::framed`) has been extracted from `NoiseOutput`. `NoiseFramed` is a `Sink` and `Stream` of length-delimited Noise protocol messages. This type is used in the handshake phase. Once a handshake completes the underlying Noise session transitions to transport mode and the `NoiseFramed` is wrapped in the `NoiseOutput` which provides a regular `AsyncRead` / `AsyncWrite` I/O resource on top of the framed encoding. No new buffers are introduced, they are just split between `NoiseFramed` and `NoiseOutput`.

While this could technically be a patch release since code has mostly been moved and slightly adapted without radical changes, due to this reorganisation of the internal I/O I think a controlled upgrade is better, i.e. this should not just instantly be picked up as a patch by all downstream crates, so I propose a minor release here.

I tested the following:

  * When the length prefix is no longer sent (a follow-up PR and release), the `ipfs-kad` example successfully connects to the enabled bootstrap node using noise.
  * A simple `cargo run` in `substrate` with this PR looks good as well, i.e. connections are established and blocks are being sync'ed.

The next two steps after this PR has been released are the following, each requiring their own release:

  2. Removing the sending of the length prefix.
  3. Removing the support for reading length-prefixed protobuf payloads.


romanb:
See the code comment [here](https://github.com/libp2p/rust-libp2p/pull/1658/files#diff-ceb1fb7eeb5632b4d3754a9212037f0dR369-R385) on why this approach to fallback-decoding (suggested by @tomaka in https://github.com/libp2p/rust-libp2p/issues/1631) should be unproblematic for the upgrade path. At least I couldn't come up with something cleaner that does not involve significantly more effort / temporary code.

protocols/noise/src/io/handshake.rs
```diff
@@ -26,14 +26,13 @@ mod payload_proto {
 
 use crate::error::NoiseError;
 use crate::protocol::{Protocol, PublicKey, KeypairIdentity};
-use crate::io::SnowState;
+use crate::io::{NoiseOutput, framed::NoiseFramed};
 use libp2p_core::identity;
 use futures::prelude::*;
 use futures::task;
-use futures::io::AsyncReadExt;
+// use futures::io::AsyncReadExt;
```
tomaka, line 33: Commented out code

protocols/noise/src/io/framed.rs
```diff
@@ -0,0 +1,429 @@
...
+                    trace!("read: {}/{} bytes", *off + n, len);
+                    if n == 0 {
+                        trace!("read: eof");
+                        this.read_state = ReadState::Eof(Err(()));
+                        return Poll::Ready(Some(Err(io::ErrorKind::UnexpectedEof.into())))
+                    }
+                    *off += n;
+                    if len == *off {
+                        trace!("read: decrypting {} bytes", len);
+                        let mut frame = vec![0u8; len];
```
twittner, line 213: This does introduce a new buffer contrary to the PR's description, no? Previously we reused our internal buffers and copied the data to/from the provided `Async{Read,Write}` argument slices. Now we create a fresh decryption buffer on every message.

protocols/noise/src/io.rs
```diff
@@ -20,448 +20,117 @@
...
-    Eof,
-    /// encryption error (terminal state)
-    EncErr
-}
-
 impl<T: AsyncRead + Unpin> AsyncRead for NoiseOutput<T> {
-    fn poll_read(mut self: Pin<&mut Self>, cx: &mut Context, buf: &mut [u8]) -> Poll<io::Result<usize>> {
-        let mut this = self.deref_mut();
+    fn poll_read(self: Pin<&mut Self>, cx: &mut Context, buf: &mut [u8]) -> Poll<io::Result<usize>> {
+        let this = Pin::into_inner(self);
```
twittner, line 63: I think within `poll_read`, one could directly use `self`.

protocols/noise/src/io/framed.rs
romanb: That is true, it essentially comes with the separation of framed (`NoiseFramed`) and streaming (`NoiseOutput`) I/O. At this point when `NoiseFramed` is wrapped in the `NoiseOutput` we would ideally like to decrypt directly into the frame buffer of `NoiseOutput` instead of exchanging the current frame buffer with the new frame. Though I don't know how to avoid this or what compiler optimisations may make of it. Do you have an idea for a different approach? The fundamental problem is that using `NoiseOutput` during the handshake phase is no longer desired, because it "hides" the noise framing already during the handshake phase. Separating these two stages seemed like the obvious route to take and as a small positive side-effect removed some code duplication between the `AsyncWrite::poll_write` and `AsyncWrite::poll_flush` of `NoiseOutput`.

twittner: > Though I don't know how to avoid this or what compiler optimisations may make of it.

I do not think the compiler can optimise away the vector allocations.

> Do you have an idea for a different approach? The fundamental problem is that using `NoiseOutput` during the handshake phase is no longer desired, because it "hides" the noise framing already during the handshake phase. 

`NoiseOutput` could have inherent private methods — thus visible to the child module `handshake` — that support reading and returning the frame length which the handshake state uses.

romanb: What do you think of https://github.com/libp2p/rust-libp2p/pull/1658/commits/6a4fc139ea1a81e83332e11cadbae4b7c6bad9b6? `bytes` seems to be cut out for this use-case.

protocols/noise/src/io.rs
romanb: Done in https://github.com/libp2p/rust-libp2p/pull/1658/commits/062205f741fa151f5eed0d17036efe3528fd16a2.

protocols/noise/src/io/framed.rs
twittner: Using `BytesMut` is better, but I am not sure if we really need to split `NoiseOutput` and add another `NoiseFramed` type. For reference this is what I had in mind: https://github.com/libp2p/rust-libp2p/compare/master...twittner:noise-read-frame.

romanb: Thanks. I made the split not only to address the problem though, but at the same time saw an opportunity to improve the code - two birds with one stone so to speak. I'm sad to hear if you don't share that view. But before I discard this approach, let me try to motivate more clearly why I perceive the split as an improvement:

  * It removes some code duplication in `AsyncWrite::poll_write` and `AsyncWrite::poll_flush` by design, i.e. turns [these two blocks](https://github.com/libp2p/rust-libp2p/blob/master/protocols/noise/src/io.rs#L237-L395) into one in the `Sink` impl of `NoiseFramed`.
  * The structure of the code reflects more clearly the structure of the I/O, i.e. the noise framing on top of which an encrypted stream is provided more clearly stands out and is reflected in the `NoiseOutput(NoiseFramed)` layering, instead of being "built-in" to `AsyncRead` / `AsyncWrite`. I.e. one larger block of code is split into two simpler ones, each a bit easier to understand on its own.
  * In a similar vain, the handshake phase and the transport phase can explicitly operate on distinct I/O types, using `NoiseFramed` with the convenient `Stream` / `Sink` API in handshake mode and wrapping it in the encrypted `NoiseOutput` `AsyncRead` / `AsyncWrite` stream when changing to transport mode.

In your diff, there is `NoiseOutput::read_frame`, returning the frame length, which is used in tandem with the `AsyncRead` API to write frame-aware code during the handshake, and writing frames remains by means of `poll_write` + `flush` (knowing that this will send exactly one frame). I still perceive this like a workaround for letting one peek into the abstraction over the noise framing done by the `AsyncRead` + `AsyncWrite` impl, indicating (to me) that this abstraction comes too early and should be applied only when the handshake is complete, using a frame-based API during the handshake.

If this is still unconvincing and you don't see the changes overall as an improvement (or even as a step backward), I would somewhat reluctantly give in. Let me know.

twittner: > * It removes some code duplication in `AsyncWrite::poll_write` and `AsyncWrite::poll_flush` by design, i.e. turns [these two blocks](https://github.com/libp2p/rust-libp2p/blob/master/protocols/noise/src/io.rs#L237-L395) into one in the `Sink` impl of `NoiseFramed`.

I did not look at the writing part at all. Would it not be possible to use an inherent method for `NoiseOutput` which is shared between writing and flushing?

> I made the split not only to address the problem though, but at the same time saw an opportunity to improve the code - two birds with one stone so to speak. I'm sad to hear if you don't share that view.  [...] If this is still unconvincing and you don't see the changes overall as an improvement (or even as a step backward), I would somewhat reluctantly give in and follow the road shown in your diff (or maybe you want to submit it as a PR?). Let me know.

The primary complication to me is the explicit state machine handling which is still there. Adding inherent methods to provide access to frames does not seem that different to me compared to doing so via separate `Sink`/`Stream` impls which then require `BytesMut` to mitigate the overhead that comes with it. But regardless, the code I linked to was merely an illustration and not meant to replace this PR.

romanb: > I did not look at the writing part at all. Would it not be possible to use an inherent method for NoiseOutput which is shared between writing and flushing?

Maybe, I didn't try because that part "came naturally" with the split of `NoiseOutput` and `NoiseFramed`. It wasn't really my intention to make refactorings in this PR, like removing duplicated code, that are not in direct connection to the changes for frame-based handshake I/O.

> The primary complication to me is the explicit state machine handling which is still there.

True, but I think it is an improvement nevertheless.

> Adding inherent methods to provide access to frames does not seem that different to me compared to doing so via separate Sink/Stream impls [..]

I take it then that you really see no value in actually having clearly different types and APIs for handshake and transport mode and don't share my feeling that combinations of `read_frame`, `poll_write`, `poll_read` and `flush` to read and write frames during handshake mode, instead of inherently frame-based `send` and `next`, seem like an unclean approach to writing frame-aware code during the handshake phase. :white_flag:.

> [..] which then require BytesMut to mitigate the overhead that comes with it.

I don't think of using `BytesMut` as an unwelcome requirement to avoid presumed overhead that would come with an implementation that does not use it, but rather as an inherent part of a good implementation that does such a split. However, I only ran some unscientific "benchmarks" using the existing smoke tests with various fixed payload sizes, in which these changes even came out a bit on top but at least showed no regression.

> But regardless, the code I linked to was merely an illustration and not meant to replace this PR

Since the approaches are very different, I don't see much point in adapting this PR, however. So closing it in favor of a new one seems like the best way to go, regardless of who does it.

twittner: > [...] presumed overhead [...]

I do hope that Rust allows us to spot allocations from reading source code and therefore *conclude* rather than *presume* if there is overhead or not ;-)

> Since the approaches are very different, I don't see much point in adapting this PR, however. So closing it in favor of a new one seems like the best way to go, regardless of who does it.

Why close it? As I wrote above: "the code I linked to was [...] not meant to replace this PR".

Issue #1650, opened 2020-07-08
[request-response] Allow for a way to close the substream


thomaseizinger:
Closing the substream in a request-response protocol is a convenient way of signaling an error to the other peer.

Unfortunately, the only way I've been able to achieve that at the moment is by creating a response enum:

https://github.com/comit-network/comit-rs/blob/2cca5fef69b49337279c16f4f7421b8052e4cbd0/comit/src/network/protocols/announce.rs#L352-L374

This has the downside that the sender side has to deal with the enum as well although the response never makes it over the wire.

https://github.com/comit-network/comit-rs/blob/2cca5fef69b49337279c16f4f7421b8052e4cbd0/comit/src/network/protocols/announce.rs#L205-L217

Would it be possible to add an interface to `RequestResponse` that allows to close a substream? Maybe dropping the `ResponseChannel` could do that automatically?

romanb:
> Maybe dropping the ResponseChannel could do that automatically?

If you drop the `ResponseChannel`, then the `InboundUpgrade` [should notice and finish](https://github.com/libp2p/rust-libp2p/blob/master/protocols/request-response/src/handler/protocol.rs#L111), resulting in the substream getting closed. This is what allows one-way protocols that don't have responses. Does that not work for you?

thomaseizinger:
> > Maybe dropping the ResponseChannel could do that automatically?
> 
> If you drop the `ResponseChannel`, then the `InboundUpgrade` [should notice and finish](https://github.com/libp2p/rust-libp2p/blob/master/protocols/request-response/src/handler/protocol.rs?rgh-link-date=2020-07-09T09%3A58%3A52Z#L111), resulting in the substream getting closed. This is what allows one-way protocols that don't have responses. Does that not work for you?

I think what I would want is:

```rust
fn upgrade_inbound(mut self, mut io: NegotiatedSubstream, protocol: Self::Info) -> Self::Future {
    async move {
        let read = self.codec.read_request(&protocol, &mut io);
        let request = read.await?;
        if let Ok(()) = self.request_sender.send(request) {
            match self.response_receiver.await {
				Ok(response) => {
					let write = self.codec.write_response(&protocol, &mut io, response);
                    write.await?;
				}
				Err(_) => {
					io.close().await;
					}
                }
            }
        Ok(())
    }.boxed()
}
```

If the substream is not explicitely closed, the side waiting for the response will not receive `ConnectionClosed` but instead wait for the timeout that is configured on their end.

[...]

PR #1660, opened 2020-07-10
[request-response] Close substream after writing request/response.

romanb:
Since forgetting to properly `close()` a substream may result in subtle problems that can be difficult to diagnose properly (see e.g. https://github.com/libp2p/rust-libp2p/pull/1606) it may be better to avoid this pitfall for implementations of `RequestResponseCodec` by always `close()`ing the substream in the `InboundUpgrade` and `OutboundUpgrade` explicitly.

### DIFF
--- a/protocols/request-response/src/handler/protocol.rs
+++ b/protocols/request-response/src/handler/protocol.rs
@@ -111,9 +111,9 @@ where
                 if let Ok(response) = self.response_receiver.await {
                     let write = self.codec.write_response(&protocol, &mut io, response);
                     write.await?;
-                    io.close().await?;
                 }
             }
+            io.close().await?;
             Ok(())
         }.boxed()
     }
@@ -164,4 +164,3 @@ where
         }.boxed()
     }
 }
-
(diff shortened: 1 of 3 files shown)


protocols/request-response/src/handler/protocol.rs
```diff
@@ -111,6 +111,7 @@ where
                 if let Ok(response) = self.response_receiver.await {
                     let write = self.codec.write_response(&protocol, &mut io, response);
                     write.await?;
+                    io.close().await?;
                 }
             }
```
thomaseizinger, line 116: Is there a reason why this one isn't just at the end of the upgrade? Then it would be called under all circumstances, even if we fail to provide a response because the sender has been dropped.

```suggestion
                }
            }            
            io.close().await?;
```

romanb: If no response is sent, there is no possibility for problems like https://github.com/libp2p/rust-libp2p/pull/1606 to occur. Of course, there is no harm in explicitly closing the substream anyway.

thomaseizinger:
This closes #1650 now.

PR #1662, opened 2020-07-13
Automatic fixes by cargo-fix

Demi-Marie:


DemiMarie-parity:
Including `cargo fix` in CI would be great!

twittner:
Does `&mut Context<'_>` convey useful information? I can see that returning a type that borrows data and writing `-> Foo<'_>` instead of `-> Foo` is a helpful signal but what does it add in argument position other than noise?

DemiMarie-parity:
I have no idea, but I still don’t want to rely on deprecated features of rust.

romanb:
@Demi-Marie If you could merge master, resolving the conflicts, I can go ahead and merge this,

Demi-Marie:
@romanb I reset to master and reran `cargo fix --all --edition-idioms`.

PR #1664, opened 2020-07-16
Spec-compliant noise handshake payloads (Step 1, Take 2).

romanb:
This is an alternative for https://github.com/libp2p/rust-libp2p/pull/1658 with a simpler implementation strategy, based on a proposal by @twittner. See the previous PR for further details. The upgrade strategy remains the same.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,14 @@
 - [`parity-multiaddr` CHANGELOG](misc/multiaddr/CHANGELOG.md)
 - [`libp2p-core-derive` CHANGELOG](misc/core-derive/CHANGELOG.md)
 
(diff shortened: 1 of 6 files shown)


romanb:
Updated the PR description.

romanb:
Unless there are objections, I will go ahead with https://github.com/libp2p/rust-libp2p/pull/1658, but will keep this PR on ice for a while as a handy fallback if unforeseen regressions arise that are not easily resolved - for I cannot deny that despite being my preferred approach, https://github.com/libp2p/rust-libp2p/pull/1658 involves more changes than done here and thus has a higher likelihood for regressions.

romanb:
It seems the other upgrade path is successful, so this can be closed.

PR #1667, opened 2020-07-17
Set SO_REUSEPORT and remove port translation.

dvc94ch:
So with this patch I was able to get ipfs-embed to traverse the nat in netsim-embed [0]. ~The outgoing port is hard coded to 31333 which is the port used by ipfs-embed by default. Since there may be multiple listeners it's unclear which port should be used for outgoing connections.~ ~Note that the example only works sometimes, other times it deadlocks and times out.~

The first listener that works for the dialing address is used by the swarm when dialing. If no local_addr is provided the transport selects a suitable one.

- [0] https://github.com/ipfs-rust/netsim-embed/blob/master/examples/p2p.rs

### DIFF
--- a/core/src/connection.rs
+++ b/core/src/connection.rs
@@ -84,6 +84,8 @@ impl Endpoint {
 pub enum ConnectedPoint {
     /// We dialed the node.
     Dialer {
+        /// Local connection address.
+        local_addr: Option<Multiaddr>,
         /// Multiaddress that was successfully dialed.
         address: Multiaddr,
     },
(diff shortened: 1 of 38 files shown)


dvc94ch:
So this feature needs more work. While implementing a simple dial back protocol to test if nat traversal works I noticed that you need to disable port reuse when dialing, because otherwise it will fail. Any thoughts on how it should work?

dvc94ch:
@tomaka so what's the solution here? Wait for quic to land and use quick by default?

PR #1669, opened 2020-07-20
[libp2p-noise] Disable sending of legacy handshake payloads (by default).

romanb:
This is step 2 in resolving https://github.com/libp2p/rust-libp2p/issues/1631. With this PR, `rust-libp2p` is now compatible with the other `libp2p-noise` implementations. As a conservative option, if one does not wish to migrate to the spec-compliant handshakes with this release, a feature flag `send-legacy-handshake` is provided by `libp2p-noise`. The plan is to provide an analogous flag `recv-legacy-handshake` with a future release that removes support for receiving legacy handshakes _by default_. In this way the two feature flags allow delaying this transition in downstream projects, if necessary. Of course, eventually these flags should be removed.

As evidence and test that this release of `libp2p-noise` provides interoperability, the `build_development_transport` used by all examples now uses `libp2p-noise`, which includes the `ipfs-kad` example that connects to IPFS nodes.

### DIFF
--- a/protocols/noise/Cargo.toml
+++ b/protocols/noise/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "libp2p-noise"
 description = "Cryptographic handshake protocol using the noise framework."
-version = "0.21.0"
+version = "0.22.0"
 authors = ["Parity Technologies <admin@parity.io>"]
 license = "MIT"
 repository = "https://github.com/libp2p/rust-libp2p"
@@ -35,4 +35,8 @@ sodiumoxide = "0.2.5"
 
 [build-dependencies]
 prost-build = "0.6"
+
+[features]
+default = []
+send-legacy-handshake = []
 
--- a/protocols/noise/Cargo.toml
+++ b/protocols/noise/Cargo.toml
@@ -36,7 +36,3 @@ sodiumoxide = "0.2.5"
 [build-dependencies]
 prost-build = "0.6"
 
-[features]
-default = []
-send-legacy-handshake = []
-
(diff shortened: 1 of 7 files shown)


protocols/noise/Cargo.toml
```diff
@@ -36,3 +36,7 @@ sodiumoxide = "0.2.5"
 [build-dependencies]
 prost-build = "0.6"
 
+[features]
+default = []
```
twittner, line 40: Empty default features are the default ;-)
```suggestion

```

tomaka:
A feature flag is the wrong tool for the job. Feature flags are only meant to be use to add features to a code base and certainly not to modify its visible behaviour.

Imagine that for example a `rust-ipfs` library, which depends (indirectly) on `libp2p-noise` without the feature flag, and and a `sc-network` library, which in this hypothetical situation depends on `libp2p-noise` with the feature flag.
By depending on both `rust-ipfs` and `sc-network`, `rust-ipfs` would simply stop working because the handshake is modified by depending on `sc-network`.

romanb:
> By depending on both rust-ipfs and sc-network, rust-ipfs would simply stop working because the handshake is modified by depending on sc-network.

Good point, these features are not "additive" in the sense that enabling them can cause code to break that does not wish to use them. I guess that leaves only introducing and passing down configuration options, though I'd much prefer to just follow along the upgrade via the releases instead of delaying it.

romanb:
I replaced the feature flag with the introduction of a `LegacyConfig`. I'd like to go ahead and release this.

PR #1672, opened 2020-07-22
Autonat

dvc94ch:
Builds on top of #1667 to add a dial back protocol with unit tests. The wire format is the same as the libp2p autonat protocol, but not tested if it's compatible, as there isn't a published spec for it. Uses netsim-embed to simulate a traversable and an untraversable nat.


dvc94ch:
maybe useful for #2052

Issue #1671, opened 2020-07-22
Gossipsub interop issue with go-ipfs


rklaehn:
I have the following scenario: I create a libp2p based program using gossipsub and connect to a go-ipfs peer.

When I subscribe to a topic immediately at startup, *sometimes* I don't ever get any messages for that topic despite the remote nodes clearly publishing on that topic.

When adding a small delay (`tokio::time::delay_for(Duration::from_secs(2)).await`) to allow for some time to pass before subscribing, everything works as expected.

I tried to work around this issue by regularly unsubscribing and subscribing to all topics that somebody is interested in at the moment, but that did not work either.

Just calling subscribe again will not work, since subscribe will abort early with a `false` return value, indicating that the topic is already subscribed.

Calling unsubscribe immediately followed by subscribe did both return true, but after the first unsubscribe/subscribe I did not get any pubsub msgs anymore.

Am I using the API correctly? It seems that I should be able to call subscribe *once* for a topic and then rely on the topic being subscribed despite not yet having any peers...

rklaehn:
@AgeManning would really appreciate your help. I can get detailed info about this scenario.

tcharding:
We are doing this too, subscibing on startup. Interested to hear the result of this thread.

rklaehn:
I have implemented a debug instance for the Gossipsub behaviour ( https://github.com/Actyx/rust-libp2p/tree/rkl/gossipsub-debug-instances ), and discovered a few weird things. Here is the state of the gossipsub in the case where I added a delay fudge to make things work:

```
Gossipsub {
    config: GossipsubConfig {
        protocol_id: "/meshsub/1.0.0",
        history_length: 5,
        history_gossip: 3,
        mesh_n: 6,
        mesh_n_low: 4,
        mesh_n_high: 12,
        gossip_lazy: 6,
        heartbeat_initial_delay: 5s,
        heartbeat_interval: 1s,
        fanout_ttl: 60s,
        max_transmit_size: 262144,
        hash_topics: false,
        no_source_id: false,
        manual_propagation: false,
    },
    events: [],
    control_pool: {},
    local_peer_id: PeerId(
        "12D3KooWSm42gMt9XAp5pkaaYBDQfWmJvCvY3WfvV1cUBeCYnCeo",
    ),
    topic_peers: {
        TopicHash {
            hash: "/klk/prod/2020-01-09",
        }: [
            PeerId(
[...]

PR #1673, opened 2020-07-23
Debug instance for the gossipsub behaviour

rklaehn:
To assist looking into https://github.com/libp2p/rust-libp2p/issues/1671

Not sure if the nicer debug output is worth the extra dependency. For me it is, but YMMV. I can remove this if you see this differently.

Edit: I added truncation of the message bytes to ensure that the debug output for a gossipsub msg is not too big.

### DIFF
--- a/protocols/gossipsub/src/config.rs
+++ b/protocols/gossipsub/src/config.rs
@@ -231,8 +231,19 @@ impl GossipsubConfigBuilder {
 
 impl std::fmt::Debug for GossipsubConfig {
     fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        struct StringOrBytes<'a>(&'a [u8]);
+
+        impl<'a> std::fmt::Debug for StringOrBytes<'a> {
+            fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+                if let Ok(text) = std::str::from_utf8(self.0) {
+                    write!(f, "\"{}\"", text)
+                } else {
+                    write!(f, "{}", hex::encode(self.0))
+                }
+            }
+        }
         let mut builder = f.debug_struct("GossipsubConfig");
-        let _ = builder.field("protocol_id", &self.protocol_id);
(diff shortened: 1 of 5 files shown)


protocols/gossipsub/src/config.rs
```diff
@@ -231,8 +231,19 @@ impl GossipsubConfigBuilder {
...
+                if let Ok(text) = std::str::from_utf8(self.0) {
+                    write!(f, "\"{}\"", text)
+                } else {
+                    write!(f, "{}", hex_fmt::HexFmt(self.0))
+                }
+            }
+        }
         let mut builder = f.debug_struct("GossipsubConfig");
-        let _ = builder.field("protocol_id", &self.protocol_id);
+        let _ = builder.field("protocol_id", &StringOrBytes(&self.protocol_id));
```
tomaka, line 246: It feels more simple to me to just inline that code.

(I typed that blindly, might not compile)

```suggestion
        if let Ok(text) = std::str::from_utf8(self.protocol_id.0) {
            let _ = builder.field("protocol_id", &text));
        } else {
            let _ = builder.field("protocol_id", hex_fmt::HexFmt(self.protocol_id.0));
        }
```

protocols/gossipsub/src/config.rs
```diff
@@ -231,8 +231,19 @@ impl GossipsubConfigBuilder {
...
+
+        impl<'a> std::fmt::Debug for StringOrBytes<'a> {
+            fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+                if let Ok(text) = std::str::from_utf8(self.0) {
+                    write!(f, "\"{}\"", text)
+                } else {
+                    write!(f, "{}", hex_fmt::HexFmt(self.0))
+                }
+            }
+        }
```
tomaka, line 244:
```suggestion
```

protocols/gossipsub/src/config.rs
rklaehn: You are right. I had that thing laying around and copied it in, but here it is only used once.

romanb:
@rklaehn If you could update the gossipsub changelog and merge master, I can go ahead and merge this.

PR #1675, opened 2020-07-23
Gossipsub: Use sets instead of vectors inside behaviour state

rklaehn:
To completely eliminate the possibility of having duplicate peers in these sets

### DIFF
--- a/protocols/gossipsub/src/behaviour.rs
+++ b/protocols/gossipsub/src/behaviour.rs
@@ -374,7 +374,7 @@ impl Gossipsub {
             let mesh_peers = self
                 .mesh
                 .entry(topic_hash.clone())
-                .or_insert_with(|| Default::default());
+                .or_insert_with(Default::default);
             mesh_peers.extend(new_peers);
         }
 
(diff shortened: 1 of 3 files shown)


protocols/gossipsub/src/behaviour.rs
```diff
@@ -771,7 +766,9 @@ impl Gossipsub {
                     }
                 }
             }
-            peers.retain(|peer| to_remove_peers.contains(&peer));
```
rklaehn, line 774: This is different than in the old code, but I think the old code has the condition flipped.

AgeManning: Yes true. Nice catch

AgeManning:
This a nice addition. There are some extra checks (for duplicates in the Vectors) that we no longer need and can be removed. 

I'm going to try and finish up fixing all these bugs and hopefully address your issue in #1583. To this end, I've merged these into there.

rklaehn:
So the signing PR is basically your development branch? I might have a few DRY refactoring improvements that I will PR against that one then.

Anyway, I am going to close this PR. Thanks for the inclusion.

PR #1676, opened 2020-07-25
Change the gossipsub rpc protocol to use bytes for message ids

rklaehn:
And do the corresponding changes to make this work.

This aligns gossipsub with the changes in the spec https://github.com/libp2p/specs/commit/8e5da407893d6e340b7ec68d849cb57b19449671

Note that this is already included in https://github.com/libp2p/rust-libp2p/pull/1583 , but since the message signing PR might take some time to be merged, this is just the format change as an independent PR.


rklaehn:
Closing because #1583 is merged

PR #1680, opened 2020-07-27
protocols/kad: Expose kbucket distance range

mxinden:
~~Instead of only returning non-empty kbuckets via `Kademlia::kbuckets`,
return empty ones as well.~~

~~Doing so enables consumers to (1) reason about each and every bucket and
(2) depend on a stable order of the buckets.~~

Add `KBucketRef::range` exposing the minimum inclusive and maximum
inclusive `Distance` for the bucket

More specifically within substrate I would like to expose the following metric:

```
HELP Number of nodes per protocol and Kademlia kbucket
substrate_sub_libp2p_kademlia_num_nodes_per_kbucket{protocol="sub", bucket="255"} 20
```

### DIFF
--- a/protocols/kad/src/kbucket.rs
+++ b/protocols/kad/src/kbucket.rs
@@ -541,6 +541,23 @@ mod tests {
     }
 
     #[test]
+    fn buckets_are_non_overlapping_and_exhaustive() {
+        let local_key = Key::from(PeerId::random());
+        let timeout = Duration::from_secs(0);
+        let mut table = KBucketsTable::<KeyBytes, ()>::new(local_key.into(), timeout);
+
+        let mut prev_max = U256::from(0);
+
+        for bucket in table.iter() {
+            let (min, max) = bucket.range();
+            assert_eq!(Distance(prev_max + U256::from(1)), min);
+            prev_max = max.0;
+        }
+
+        assert_eq!(U256::MAX, prev_max);
+    }
+
+    #[test]
     fn bucket_contains_range() {
         fn prop(ix: u8) {
             let index = BucketIndex(ix as usize);
--- a/protocols/kad/src/kbucket.rs
+++ b/protocols/kad/src/kbucket.rs
@@ -119,10 +119,13 @@ impl BucketIndex {
     /// Returns the minimum inclusive and maximum inclusive [`Distance`]
     /// included in the bucket for this index.
     fn range(&self) -> (Distance, Distance) {
-        (
-            Distance(U256::pow(U256::from(2), U256::from(self.0))),
-            Distance(U256::pow(U256::from(2), U256::from(self.0 + 1)) - 1),
-        )
+        let min = Distance(U256::pow(U256::from(2), U256::from(self.0)));
+        let max = match U256::overflowing_pow(U256::from(2), U256::from(self.0 + 1)) {
+            (exp, false) => Distance(exp - 1),
+            (_, true) => Distance(U256::MAX),
+        };
+
+        (min, max)
     }
 
     /// Generates a random distance that falls into the bucket for this index.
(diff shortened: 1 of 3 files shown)


romanb:
The only small issue I have with this is that it leaks implementation details. Ideally, how the implementation strategy for the buckets should not leak in the public API, but that is what happens when not filtering out empty buckets - empty buckets only exist because of the 256-fixed-array implementation, e.g. I don't think they would be present in an implementation based on a dynamically growing "bucket-tree", what is actually suggested in the paper. Of course, one could artificially "fill up" the gaps in the iterator with empty buckets in that case, but even then there would no longer always be 256 buckets because the ranges covered by the buckets vary in that case, if I remember correctly.

Isn't what you're mainly interested in the distances covered by the buckets? The order should always be stable. Maybe exposing the ranges covered by each bucket would be sufficient? If you know which distance each non-empty bucket covers, you also know the "empty space" in-between.

mxinden:
I agree that my suggestion leaks implementation details. E.g. making [`NUM_BUCKETS`](https://github.com/libp2p/rust-libp2p/blob/master/protocols/kad/src/kbucket.rs#L81) configurable in the future would be a breaking change.

@romanb are you suggesting a change along the lines of the following?

```diff
diff --git a/protocols/kad/src/kbucket.rs b/protocols/kad/src/kbucket.rs
index 2ea7f362..715eedde 100644
--- a/protocols/kad/src/kbucket.rs
+++ b/protocols/kad/src/kbucket.rs
@@ -447,6 +447,14 @@ where
     TKey: Clone + AsRef<KeyBytes>,
     TVal: Clone
 {
+    pub fn min_distance(&self) -> Distance {
+        unimplemented!()
+    }
+
+    pub fn max_distance(&self) -> Distance {
+        unimplemented!()
+    }
+
```

romanb:
@mxinden Yes, something like that. Or just a single method along the lines of `fn range(&self) -> (Distance, Distance)`.

mxinden:
@romanb would you mind taking another look?

protocols/kad/src/kbucket.rs
```diff
@@ -525,6 +540,30 @@ mod tests {
         }
     }
 
+    #[test]
+    fn bucket_contains_range() {
```
romanb, line 544: Another property to check may be that the ranges of all buckets partition the keyspace, i.e. are non-overlapping and exhaustive. I leave it up to you whether you want to include such a test.

mxinden: Good idea. Added in https://github.com/libp2p/rust-libp2p/pull/1680/commits/501e80c5c73adbcceff4e01848874a560b560ee3 plus https://github.com/libp2p/rust-libp2p/pull/1680/commits/f8ba555ae4a75fd4d9f263ff86c882b66dd43b25 fixing the overflow panic it revealed. Thanks!

protocols/kad/src/kbucket.rs
```diff
@@ -119,10 +119,13 @@ impl BucketIndex {
...
     fn range(&self) -> (Distance, Distance) {
-        (
-            Distance(U256::pow(U256::from(2), U256::from(self.0))),
-            Distance(U256::pow(U256::from(2), U256::from(self.0 + 1)) - 1),
-        )
+        let min = Distance(U256::pow(U256::from(2), U256::from(self.0)));
+        let max = match U256::overflowing_pow(U256::from(2), U256::from(self.0 + 1)) {
+            (exp, false) => Distance(exp - 1),
+            (_, true) => Distance(U256::MAX),
+        };
```
romanb, line 126:
```suggestion
        if self.0 == u8::MAX {
            (min, U256::MAX)
        } else {
            let max = U256::pow(U256::from(2), U256::from(self.0 + 1)) - 1;
            (min, max)
        }
```

romanb:
Feel free to update the changelog and merge at your convenience.

Issue #1601, opened 2020-06-02
Our implementation of ping might not be conformant


tomaka:
I'm opening this issue because someone has been reporting this to me a long time ago, but I haven't verified the claim.

In our libp2p-ping implementation we open a new substream every time we want to ping the remote, while for js-libp2p and go-libp2p the substream seems to be kept open.

This needs be checked before actually making any change.

tcharding:
Confirming that the go-libp2p ping implementation does keep the stream open. See https://github.com/libp2p/go-libp2p/blob/master/p2p/protocol/ping/ping.go

For reference:
```
// Ping pings the remote peer until the context is canceled, returning a stream
// of RTTs or errors.
func Ping(ctx context.Context, h host.Host, p peer.ID) <-chan Result {
	s, err := h.NewStream(ctx, p, ID)
    ...
    for ctx.Err() == nil {
			var res Result
			res.RTT, res.Error = ping(s)
   ...
    select {
			case out <- res:
			case <-ctx.Done():
				return
			}
    }
}
```
And `ping` is the same as ours, record start time, write 32 bytes, read and check they are the same bytes written.

romanb:
Are the protocols actually incompatible in the sense that undesirable behaviour or errors result? Or is it just a matter of doing what others do? The [spec for the ping protocol](https://github.com/libp2p/specs/issues/183) is a bit short.

tomaka:
> Are the protocols actually incompatible in the sense that undesirable behaviour or errors result?

I believe that the consequence is that the Go or JS node thinks that the Rust node is unresponsive and shuts down the connection.

It's a tolerable situation considering that the connection will stay open for a few dozen seconds during which everything works fine, but it would still be great to fix.

romanb:
Should be fixed as of https://github.com/libp2p/rust-libp2p/pull/1692.

PR #1692, opened 2020-08-06
[ping] Refactor the ping protocol for conformity.

romanb:
This addresses https://github.com/libp2p/rust-libp2p/issues/1601. Pings are now sent over a single substream, as it is done in other libp2p implementations. A few points worth noting:

  * Since each peer sends its pings over a single, dedicated substream, every peer that participates in the protocol has effectively two open substreams. This seems to be in-line with other implementations.
  * I tested interoperability with the `ipfs ping` command against `rust-libp2p` nodes using the ping example.
  * To keep backward-compatibility and not unnecessarily close connections, with the default configuration of `max_failures` set to `1`, the first outbound ping failure is "free" and "silent". A node using the old protocol will close inbound substreams prematurely after responding to a ping, causing a subsequent ping failure for the node using the new protocol. This tweak allows nodes using the different versions to interoperate without killing connections prematurely, while at the same time never emitting more events for failures than before (i.e. `max_failures - 1` events are emitted for failures before closing the connection as before). I tested interoperability in this sense between the ping example on this branch and the one in `master`.

### DIFF
--- a/protocols/ping/src/protocol.rs
+++ b/protocols/ping/src/protocol.rs
@@ -18,23 +18,27 @@
 // FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
 // DEALINGS IN THE SOFTWARE.
 
-use futures::{future::BoxFuture, prelude::*};
+use futures::prelude::*;
 use libp2p_core::{InboundUpgrade, OutboundUpgrade, UpgradeInfo};
-use log::debug;
+use libp2p_swarm::NegotiatedSubstream;
 use rand::{distributions, prelude::*};
 use std::{io, iter, time::Duration};
+use void::Void;
 use wasm_timer::Instant;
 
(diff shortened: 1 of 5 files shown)


protocols/ping/src/protocol.rs
```diff
@@ -55,59 +59,60 @@ impl UpgradeInfo for Ping {
...
-                socket.write_all(&payload).await?;
-            }
-            socket.close().await?;
-            Ok(())
-        }.boxed()
+    let payload: [u8; PING_SIZE] = thread_rng().sample(distributions::Standard);
+    log::debug!("Preparing ping payload {:?}", payload);
+    stream.write_all(&payload).await?;
+    let started = Instant::now();
+    let mut recv_payload = [0u8; 32];
```
mxinden, line 91:
```suggestion
    let mut recv_payload = [0u8; PING_SIZE];
```

PR #1694, opened 2020-08-07
Expose TBehaviour in Toggle Struct

nannick:
For my use case it is useful to access the TBheaviour. For instance when Mdns is wrapped in a Toggle struct, I still want the ability to call Mdns functions such as has_node(). However that is currently not possible because there is no way to expose TBehaviour


romanb:
Is this not already covered by https://github.com/libp2p/rust-libp2p/pull/1684?

nannick:
> Is this not already covered by #1684?

Ahh yes it is. I didn't check earlier issues to see if someone already submitted a PR. Thank you for your help!

Issue #1687, opened 2020-08-04
In libp2p-request-responses, give the possiblity to specify a limit to the number of simultaneous requests per peer


tomaka:
Considering that there doesn't exist any back-pressure mechanism when it comes to opening substreams, we need way to protect against an attacker opening thousands of substreams on a connection.

While the multiplexer does enforce a limit to the total number of open substreams per connection, this value must be pretty high in order to avoid accidentally reaching it in legitimate scenarios.

Instead, it seems preferable to me to enforce, for each individual peer and protocol, the number of substreams that can be open simultaneously.

When it comes to request-response protocols, this can be done by enforcing a limit to the number of simultaneous requests that a remote is capable of making.
I'm opening this issue in libp2p because to me it is preferable to do this verification in the `ProtocolsHandler`.

tomaka:
I'm unsure whether we should check against this limit on the sending side too, or if we delegate this responsibility to the user.
Delegating this responsibility to the user makes this potentially very error-prone.

twittner:
As mentioned [elsewhere](https://github.com/paritytech/substrate/pull/6634#discussion_r455093596) I think it would be a good idea to send metadata such as error codes or back-pressure information together with the actual payload. Presumably the limit depends on the capacity of a node and is thus quite dynamic.

romanb:
Closed by https://github.com/libp2p/rust-libp2p/pull/1726, https://github.com/libp2p/rust-libp2p/pull/1731. For bugs, new tweaks or other change requests relating to this functionality, it is probably better to open new issues.

PR #1696, opened 2020-08-10
Add `Throttled` to libp2p-request-response.

twittner:
Wraps the existing `RequestResponse` behaviour and applies strict limits to the number of inbound and outbound requests per peer.

The wrapper is opt-in and if not used, the protocol behaviour of `RequestResponse` does not change. This PR also does not introduce an extra protocol, hence the limits applied need to be known a priori for all nodes which is not always possible or desirable. As mentioned in #1687 I think that we should eventually augment the protocol with metadata which allows a more dynamic exchange of requests and responses.

This PR also replaces the two oneshot channels with a single one from the [scambio](https://crates.io/crates/scambio) crate which saves one allocation per request/response. If not desirable because the crate has seen less testing the first commit could be reverted.

### DIFF
--- a/protocols/request-response/src/throttled.rs
+++ b/protocols/request-response/src/throttled.rs
@@ -172,7 +172,7 @@ impl<C: RequestResponseCodec + Clone> Throttled<C> {
     /// Answer an inbound request with a response.
     ///
     /// See [`RequestResponse::send_response`] for details.
-    pub fn send_response(&mut self, ch: ResponseChannel<C::Request, C::Response>, rs: C::Response) {
+    pub fn send_response(&mut self, ch: ResponseChannel<C::Response>, rs: C::Response) {
         if let Some(limit) = self.limits.get_mut(&ch.peer) {
             limit.recv_budget += 1;
             debug_assert!(limit.recv_budget <= limit.maximum.get())
--- a/protocols/request-response/src/throttled.rs
+++ b/protocols/request-response/src/throttled.rs
@@ -211,9 +211,7 @@ impl<C: RequestResponseCodec + Clone> Throttled<C> {
 
 impl<C> NetworkBehaviour for Throttled<C>
 where
-    C: RequestResponseCodec + Send + std::fmt::Debug + Clone + 'static,
-    C::Request: std::fmt::Debug,
-    C::Response: std::fmt::Debug
+    C: RequestResponseCodec + Send + Clone + 'static
 {
     type ProtocolsHandler = RequestResponseHandler<C>;
     type OutEvent = Event<C::Request, C::Response>;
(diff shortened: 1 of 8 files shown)


swarm/src/behaviour.rs
```diff
@@ -281,6 +281,44 @@ pub enum NetworkBehaviourAction<TInEvent, TOutEvent> {
...
+                    handler,
+                    event: f(event)
+                },
+            NetworkBehaviourAction::ReportObservedAddr { address } =>
+                NetworkBehaviourAction::ReportObservedAddr { address }
+        }
+    }
+
+    /// Map the event the swarm will return.
+    pub fn map_out<E>(self, f: impl FnOnce(TOutEvent) -> E) -> NetworkBehaviourAction<TInEvent, E> {
```
tomaka, line 306: :+1:

tomaka:
I'm not fundamentally opposed to `scambio`, but it feels like reinventing the wheel and having to maintain one additional crate just to merge two `Arc`s into one?

twittner:
It is up to you. I do not want to impose it on `rust-libp2p`. Since `scambio` is one of my personal crates I am maintaining it anyway. It was created to allow implementing an efficient `call` primitive so it allows successive send-receive operations. I also wanted something without `unsafe` code blocks and having 50% less allocations compared to using two channels is not that bad either ;-)

protocols/request-response/src/throttled.rs
```diff
@@ -0,0 +1,302 @@
...
+        // The limit may have been added by [`Throttled::send_request`] already.
+        if !self.limits.contains_key(peer) {
+            let limit = self.previous.pop(peer).unwrap_or_else(|| self.default.clone());
+            self.limits.insert(peer.clone(), limit);
+        }
+    }
+
+    fn inject_disconnected(&mut self, peer: &PeerId) {
+        log::trace!("{:08x}: disconnected from {}", self.id, peer);
+        self.behaviour.inject_connected(peer);
```
romanb, line 247:
```suggestion
        self.behaviour.inject_disconnected(peer);
```

protocols/request-response/tests/ping.rs
```diff
@@ -49,11 +49,11 @@ fn ping_protocol() {
     let cfg = RequestResponseConfig::default();
 
     let (peer1_id, trans) = mk_transport();
-    let ping_proto1 = RequestResponse::new(PingCodec(), protocols.clone(), cfg.clone());
+    let ping_proto1 = RequestResponse::new(PingCodec(), protocols.clone(), cfg.clone()).throttled();
```
romanb, line 52: I think it would be good to have a separate, dedicated test for the throttling.

twittner: Done in 52d03bd82f29262f76cb0891474722fa90983dbe.

PR #1697, opened 2020-08-11
core: remove duplicates when performing address translation

andresilva:
When doing address translation we want to keep the discovered external address but keep the local port. If we are listening on multiple addresses (with the same port) we end up with a bunch of duplicates.

### DIFF
--- a/core/src/network.rs
+++ b/core/src/network.rs
@@ -206,7 +206,7 @@ where
         THandler: 'a,
     {
         let mut addrs: Vec<_> = self.listen_addrs()
-            .flat_map(move |server| address_translation(server, observed_addr))
+            .filter_map(move |server| address_translation(server, observed_addr))
             .collect();
 
         // remove duplicates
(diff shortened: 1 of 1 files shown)


romanb (approved):
I think this is a good thing to do, not least because, as far as I understand it, these duplicates result in undesirable increases of an observed addresses "score" [here](https://github.com/libp2p/rust-libp2p/blob/master/swarm/src/lib.rs#L757).

core/src/network.rs
```diff
@@ -205,7 +205,15 @@ where
         TMuxer: 'a,
         THandler: 'a,
     {
-        self.listen_addrs().flat_map(move |server| address_translation(server, observed_addr))
+        let mut addrs: Vec<_> = self.listen_addrs()
+            .flat_map(move |server| address_translation(server, observed_addr))
```
romanb, line 209:
```suggestion
            .filter_map(move |server| address_translation(server, observed_addr))
```

andresilva: Yeah I guess that's more obvious.

PR #1698, opened 2020-08-12
swarm/one_shot: Initialize handler with KeepAlive::Until

mxinden:
The `OneShotHandler` `keep_alive` property is altered on incoming and
outgoing reqeusts. By default it is initialized in `KeepAlive::Yes`. In
case there are no incoming or outgoing requests happening, this state is
never changed and thus the handler keeps the underlying connection alive
indefinitely.

With this commit the handler is initialized with `KeepAlive::Until`. As
before the `keep_alive` timer is updated on incoming requests and set to
`KeepAlive::Yes` on outgoing requests.

---

Related to https://github.com/paritytech/polkadot/issues/1544.

### DIFF
--- a/swarm/src/protocols_handler/one_shot.rs
+++ b/swarm/src/protocols_handler/one_shot.rs
@@ -72,7 +72,7 @@ where
             dial_queue: SmallVec::new(),
             dial_negotiated: 0,
             max_dial_negotiated: 8,
-            keep_alive: KeepAlive::Yes,
+            keep_alive: KeepAlive::Until(Instant::now() + config.keep_alive_timeout),
             config,
         }
     }
--- a/swarm/src/protocols_handler/one_shot.rs
+++ b/swarm/src/protocols_handler/one_shot.rs
@@ -245,3 +245,30 @@ impl Default for OneShotHandlerConfig {
     }
 }
 
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    use futures::executor::block_on;
+    use futures::future::poll_fn;
+    use libp2p_core::upgrade::DeniedUpgrade;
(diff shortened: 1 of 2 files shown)


swarm/src/protocols_handler/one_shot.rs
```diff
@@ -72,7 +72,7 @@ where
             dial_queue: SmallVec::new(),
             dial_negotiated: 0,
             max_dial_negotiated: 8,
-            keep_alive: KeepAlive::Yes,
+            keep_alive: KeepAlive::Until(Instant::now() + config.keep_alive_timeout),
```
romanb, line 75: It may be better to just move this condition from `inject_fully_negotiated_outbound`
```rust
        if self.dial_negotiated == 0 && self.dial_queue.is_empty() {
            self.keep_alive = KeepAlive::Until(Instant::now() + self.config.keep_alive_timeout);
        }
```
into `poll()`
```rust
       ...
       if !self.dial_queue.is_empty() {
            ...
        } else {
            self.dial_queue.shrink_to_fit();
            if self.keep_alive.is_yes() && self.dial_negotiated == 0 {
                self.keep_alive = KeepAlive::Until(Instant::now() + self.config.keep_alive_timeout);
            }
        }
```
The reason being that starting timers on creation of connection handlers is mildly problematic, because at least right now, handlers are created before the connection is even established, whereas a connection handler is `poll()`ed as soon as it is established and it gives the handler a chance to "make its first step" before the timeout starts (if there is indeed nothing that the handler wishes to do, as determined by `poll()`).

mxinden:
Thanks @romanb for pointing out the early initialization issue! Could you take another look?

PR #1699, opened 2020-08-12
protocols/mdns: Make libp2p-mdns socket agnostic

mxinden:
Allow libp2p-mdns to use either async-std or tokio to drive required UDP
socket.

As far as I can tell the two features (`mdns-async-std`, `mdns-tokio`)
and the corresponding tests are additive.

Closes https://github.com/libp2p/rust-libp2p/issues/1318

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,7 @@ default = [
     "identify",
     "kad",
     "gossipsub",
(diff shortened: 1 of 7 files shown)


romanb:
@mxinden Do you want to merge this before the next release?

PR #1700, opened 2020-08-13
Fix connection keep-alive for request-response.

romanb:
Upon reviewing https://github.com/libp2p/rust-libp2p/pull/1698 I double-checked the analogous handling in `request-response` which seemed to lack a condition, resulting in an indefinite keep-alive. More generally and judging from prior experience from other handlers where the keep-alive handling has been tricky to get right, it seems this keep-alive handling may better be done in a more general manner in `libp2p-core`, for all connection handlers. Not sure yet how realistic that is though.

### DIFF
--- a/protocols/request-response/src/handler.rs
+++ b/protocols/request-response/src/handler.rs
@@ -313,7 +313,7 @@ where
             self.outbound.shrink_to_fit();
         }
 
-        if self.inbound.is_empty() {
+        if self.inbound.is_empty() && self.keep_alive.is_yes() {
             // No new inbound or outbound requests. However, we may just have
             // started the latest inbound or outbound upgrade(s), so make sure
             // the keep-alive timeout is preceded by the substream timeout.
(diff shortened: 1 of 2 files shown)


protocols/request-response/src/handler.rs
```diff
@@ -313,7 +313,7 @@ where
             self.outbound.shrink_to_fit();
         }
 
-        if self.inbound.is_empty() {
+        if self.inbound.is_empty() && self.keep_alive.is_yes() {
```
mxinden, line 316: Paraphrasing this to make sure I am on the same page: Without the additional `&& self.keep_alive.is_yes()` this condition would be triggered on each poll if the `inbound` queue is empty and thus the `self.keep_alive` timeout would be refreshed each time leaving it no time to expire. Is that correct?

romanb: That is correct.

Issue #1526, opened 2020-03-29
Kad-GetProviders doesn't work on two nodes


michaelvoronov:
It seems that getproviders in libp2p-kad doesn't work for two nodes. From code [here](https://github.com/libp2p/rust-libp2p/blob/92ce5d6179c059c52a6c58c6c55ad69544ac2e93/protocols/kad/src/behaviour.rs#L599) it could be obtained that `provider_peers` tries to find providers only in the buckets, but buckets aren't contain any info about a current node... So it works perfectly on >2 nodes, but doesn't on 2.

mxinden:
@michaelvoronov would you mind providing a unit test that reproduces the failure above?

BlackYoup:
@mxinden I would be happy to provide such tests as I just had the issue when trying to adapt the distributed-key-value-store example to put providers instead of putting values :)

I'm not very familiar with the code base, where would it make more sense? Would `protocols/kad/src/behaviour/test.rs` be ok?

michaelvoronov:
Sorry, I missed notification, we've switched to put/get functionality, but I think this issue could be obtained from source code.

mxinden:
@BlackYoup either in `protocols/kad/src/behaviour/test.rs` or an example along the lines of https://github.com/libp2p/rust-libp2p/blob/master/examples/distributed-key-value-store.rs. Thanks!

@michaelvoronov I am still having difficulties following your above reasoning.

BlackYoup:
@mxinden I opened https://github.com/libp2p/rust-libp2p/pull/1704 which updates the distributed-key-value-store example. With this, if you run it in two terminals:
- Terminal 1: `PUT_PROVIDER foo`
- Terminal 2: `GET_PROVIDERS foo` 

Nothing will be printed in terminal 2. If you add a log when the `QueryResult::GetProviders(Ok(ok))` is called, you'll see that the `providers` hashset is empty. It works fine if you use three terminals.

michaelvoronov:
@mxinden Let's consider two nodes (1 and 2) in a network that know about each other. 1 starts providing some key and then 2 calls get providers for this key that will trigger `provider_peers` on the 1. This functions looks like this:

```rust
    fn provider_peers(&mut self, key: &record::Key, source: &PeerId) -> Vec<KadPeer> {
        let kbuckets = &mut self.kbuckets;
        self.store.providers(key)
            .into_iter()
            .filter_map(move |p|
                if &p.provider != source {
[...]

PR #1704, opened 2020-08-13
Examples: adapt distributed-key-value-store to also GET / PUT providers

BlackYoup:
This PR is related to https://github.com/libp2p/rust-libp2p/issues/1526.

I adapted the distributed-key-value-store example to also support provide and get providers queries.

Let me know if you'd prefer this in a whole other example. I also didn't rename the example as I'm not sure how to call it.


romanb:
> With #1526 the example should also work for two nodes using provider records.

By "should" I mean that I actually tried it, thanks for updating the example to make that easy!

PR #1708, opened 2020-08-17
[kad] Store addresses of provider records.

romanb:
So far, provider records are stored without their addresses and the addresses of providers are obtained from the routing table on demand. I don't remember whether this was ever done on purpose or just an oversight as part of a larger amount of work. In any case, this approach has two shortcomings:

  1. We can only return provider records whose provider peers happen to currently be in the local routing table.

  2. The local node never returns itself as a provider for a key, even if it is indeed a provider.

These issues are addressed here by storing the addresses together with the provider records, falling back to
addresses from the routing table only for backward-compatibility with existing implementations of `RecordStore` that use persistent storage and may thus have records that persist across restarts.

Closes https://github.com/libp2p/rust-libp2p/issues/1526.

### DIFF
--- a/protocols/kad/src/behaviour.rs
+++ b/protocols/kad/src/behaviour.rs
@@ -92,6 +92,9 @@ pub struct Kademlia<TStore> {
     /// Queued events to return when the behaviour is being polled.
     queued_events: VecDeque<NetworkBehaviourAction<KademliaHandlerIn<QueryId>, KademliaEvent>>,
 
+    /// The currently known addresses of the local node.
+    local_addrs: HashSet<Multiaddr>,
+
     /// The record storage.
     store: TStore,
 }
@@ -358,6 +361,7 @@ where
             record_ttl: config.record_ttl,
             provider_record_ttl: config.provider_record_ttl,
             connection_idle_timeout: config.connection_idle_timeout,
+            local_addrs: HashSet::with_capacity(6)
         }
     }
 
@@ -708,7 +712,14 @@ where
     /// The results of the (repeated) provider announcements sent by this node are
     /// reported via [`KademliaEvent::QueryResult{QueryResult::StartProviding}`].
     pub fn start_providing(&mut self, key: record::Key) -> Result<QueryId, store::Error> {
-        let record = ProviderRecord::new(key.clone(), self.kbuckets.local_key().preimage().clone());
+        // Note: We store our own provider records locally without local addresses
+        // to avoid redundant storage and outdated addresses. Instead these are
+        // acquired on demand when returning a `ProviderRecord` for the local node.
+        let local_addrs = Vec::new();
+        let record = ProviderRecord::new(
+            key.clone(),
+            self.kbuckets.local_key().preimage().clone(),
+            local_addrs);
         self.store.add_provider(record)?;
         let target = kbucket::Key::new(key.clone());
         let peers = self.kbuckets.closest_keys(&target);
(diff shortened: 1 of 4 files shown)


protocols/kad/src/behaviour.rs
```diff
@@ -358,6 +361,7 @@ where
             record_ttl: config.record_ttl,
             provider_record_ttl: config.provider_record_ttl,
             connection_idle_timeout: config.connection_idle_timeout,
+            local_addrs: HashSet::with_capacity(6)
```
mxinden, line 364: Both `Addresses` in `addresses.rs` and `local_addrs` use the assumption that a usual node has < 6 addresses in the average case. Do you think this is worth unifying in a single constant?

protocols/kad/src/behaviour.rs
```diff
@@ -784,12 +795,42 @@ where
...
+                    } else {
+                        KadConnectionType::NotConnected
+                    };
+                    if multiaddrs.is_empty() {
+                        // The provider is either the local node and we fill in
+                        // the local addresses on demand, or it is a legacy
+                        // provider record without addresses, in which case we
+                        // try to find addresses in the routing table, as was
+                        // done before provider records were stored along with
+                        // their addresses.
```
mxinden, line 817: Do you want to reference this pull request here or purely depend on `git blame` for future us? I am fine either way. The former is just a suggestion.

protocols/kad/src/behaviour.rs
```diff
@@ -2497,3 +2553,8 @@ pub enum RoutingUpdate {
     /// peer ID).
     Failed,
 }
+
+/// The maximum number of local addresses as reported by
+/// other peers that the behaviour tracks. The behaviour
+/// always tracks all its listen addresses.
```
mxinden, line 2559:
```suggestion
/// The maximum number of local external addresses. When reached any further externally reported addresses are ignored. The behaviour always tracks all its listen addresses.
```

Nit pick: Wouldn't this be _more_ correct?

romanb:
> Good catch. I have missed this to be honest.
> 
> Wouldn't the `ProviderRecord`s need to be updated in `inject_addr_failure` and `inject_address_change`?

I don't think that is necessary as provider records must be regularly re-published. It may not even be desirable.

As for `inject_addr_failure`, just because one node cannot reach a certain address does not mean others cannot either. A provider may publish a bunch of addresses, different subsets being reachable by different nodes, depending on where they are located. This is not dissimilar to the addresses in the routing table where we do remove such addresses, except for the last one, which is a  trade-off in favour of faster routing lookups in the general case, because these addresses are used so much more frequently in the iterative routing. In general, however, it is the job of the provider to tell others which addresses to store and advertise to others.

As for `inject_address_change`, even if a provider is generally ok with its addresses being updated without his direct instruction in this manner, I think a push-based approach would be more desirable. Meaning, a provider could re-publish its own provider records whenever his listen addresses change, but given that this is already done in regular intervals anyway, it is at best a nice-to-have.

Does that sound reasonable or do you have more serious concerns?

protocols/kad/src/behaviour.rs
romanb: I can add a constant but note that this is really not any kind of important assumption, it is just an arbitrary low constant > 1, since a node rarely has just one listen address. I would even be fine with not using `with_capacity` at all here, given that this is just a single instance per behaviour and we're hence just talking about avoiding (or not) a single allocation per `Kademlia` instance.

protocols/kad/src/behaviour.rs
romanb: The commit message links to https://github.com/libp2p/rust-libp2p/issues/1526 and together with the git history, that seems sufficient? I'm not against links to PRs/issues in the code, but I prefer them in commit messages.

protocols/kad/src/behaviour.rs
mxinden: I am in favor of not using `with_capactiy` here at all.

romanb: Done in https://github.com/libp2p/rust-libp2p/pull/1708/commits/f40e8a68fd71eb9c38ec017ac33af77199127f4e.

Issue #1707, opened 2020-08-17
`ProtocolsHandler` methods need to be implemented in helpers.


twittner:
The `ProtocolsHandler` trait has several methods with default implementations, in particular

- `inject_address_change`
- `inject_listen_upgrade_error`

It also has some helper functionality like `map_in`, `map_out` or `select`. The traits impls from those do not implement these methods by forwarding the calls to their inner impls. `ProtocolsHandler` types which have been mapped or selected, will thus miss those callback.

romanb:
Closed by https://github.com/libp2p/rust-libp2p/issues/1707, though there may be more refinements later.

PR #1710, opened 2020-08-17
Implement ProtocolsHandler methods in wrappers.

twittner:
This PR forwards calls to some `ProtocolsHandler` methods that were previously not implemented in wrappers such as `MapInEvent`.

~~It is [unclear](https://github.com/twittner/rust-libp2p/blob/a8b990c5c11dee82477e243f77f0d9211b444228/swarm/src/protocols_handler/multi.rs#L160) though how this can be implemented in some handlers such as `MultiHandler` as the information at hand does not enable it to decide which handler to forward the call to.~~

NB that there is no guarantee at all that a `listen_protocol` call will eventually be followed by a call to either `inject_fully_negotiated_inbound` or `inject_listener_upgrade_err`.

For context see #1707.

### DIFF
--- a/swarm/src/protocols_handler.rs
+++ b/swarm/src/protocols_handler.rs
@@ -156,9 +156,7 @@ pub trait ProtocolsHandler: Send + 'static {
     /// Indicates to the handler that upgrading an inbound substream to the given protocol has failed.
     fn inject_listen_upgrade_error(
         &mut self,
-        _: ProtocolsHandlerUpgrErr<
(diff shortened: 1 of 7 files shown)


mxinden:
@twittner would you mind adding an entry to the swarm `CHANGELOG.md?

romanb:
@mxinden I already did so.

mxinden:
Oh, that was fast, sorry, missed the commit. Thanks @romanb.

PR #1714, opened 2020-08-18
Add `ProtocolsHandler::InboundOpenInfo`.

twittner:
See #1709 for details.

PR #1715, opened 2020-08-18
Restore `RequestResponse::throttled`.

### LINKED ISSUE #1706: request-response: Throttled can not count its receive budget properly.
In `libp2p-request-response` there is a `Throttled` type which internally maintains counters for inbound and outbound requests per peer and limits their number. For inbound requests, it decrements its receive budget by 1 when receiving a `RequestResponseHandlerEvent::Request` from its inner `RequestResponse` behaviour. This budget should increase again, either when the request has been answered via `Throttled::send_response`, or when the request processing has timed out. The problem is that `Throttled` can not perform this increment correctly.

If it increments on a `RequestResponseHandlerEvent::InboundTimeout` it can not know if this timeout corresponds to a previously seen `RequestResponseHandlerEvent::Request`, because the timeout can also happen when the remote has not finished sending the request which would not cause a `Request` event. If it does not increment here it relies on the application to eventually invoke `Throttled::send_response` which is not guaranteed.

If we would only increment the budget in `Throttled::send_response` if sending over the `oneshot` channel succeeds (indicating that the upgrade is still in progress because the `oneshot` receiver has not been dropped yet), and otherwise increment on a timeout event, we are subject to a race condition because even after a successful send, the next poll of the future may determine first that the time is up and cause a timeout which makes us increment twice.

The proper solution to this is to attach an ID to inbound requests and errors, such as timeouts. This would allow `Throttled` to maintain a set of current requests and increment only if the ID is a element of this set. Unfortunately the `ProtocolsHandler` trait does not provide a way to attach information to inbound upgrades like it does for outbound ones.

twittner:
Fixes #1706.
Continuation of #1714.

### DIFF
--- a/protocols/request-response/src/throttled.rs
+++ b/protocols/request-response/src/throttled.rs
@@ -48,15 +48,15 @@ pub struct Throttled<C: RequestResponseCodec> {
     /// The wrapped behaviour.
     behaviour: RequestResponse<C>,
     /// Information per peer.
-    limits: HashMap<PeerId, Info>,
+    limits: HashMap<PeerId, PeerInfo>,
     /// The default limits applied to all peers unless overriden.
-    default: Info,
+    default: PeerInfo,
     /// Pending events to report in `Throttled::poll`.
     events: VecDeque<Event<C::Request, C::Response>>,
 }
 
 #[derive(Clone, Debug)]
-struct Info {
+struct PeerInfo {
     /// The remaining number of outbound requests that can be send.
     send_budget: u16,
     /// The remaining number of inbound requests that can be received.
@@ -69,10 +69,10 @@ struct Info {
     inbound: HashSet<RequestId>
 }
 
-impl Default for Info {
+impl Default for PeerInfo {
     fn default() -> Self {
         let maximum = NonZeroU16::new(1).expect("1 > 0");
-        Info {
+        PeerInfo {
             send_budget: maximum.get(),
             recv_budget: maximum.get(),
             maximum,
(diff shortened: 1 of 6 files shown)


romanb:
Would you mind rebasing or merging `master` for ease of review, since https://github.com/libp2p/rust-libp2p/pull/1714 is now merged?

protocols/request-response/src/throttled.rs
```diff
@@ -46,35 +47,37 @@ pub struct Throttled<C: RequestResponseCodec> {
...
+    default: Info,
     /// Pending events to report in `Throttled::poll`.
-    events: VecDeque<Event<C::Request, C::Response>>
+    events: VecDeque<Event<C::Request, C::Response>>,
 }
 
-/// A `Limit` of inbound and outbound requests.
 #[derive(Clone, Debug)]
-struct Limit {
+struct Info {
```
romanb, line 59:
```suggestion
struct PeerInfo {
```

romanb (approved):

twittner:
Thinking a little more about the approach here I think we have a race condition w.r.t. timeouts and errors. If the sender of a request times out or errors earlier than the receiver it grants itself another request. Should that request reach the receiver before it has detected the timeout or error and likewise allowed another request from the sender, it would potentially generate a `TooManyInboundRequests` event.

Without a way to give credit to a remote that is independent of receiving a request, i.e. being able to send unidirectional and unsolicited credit grants (which allows the receiver to retry sending the credit after a timeout or error) there is always a need for both ends to adjust their credit views which leads to inconsistencies as described above.

romanb:
Ultimately superseded by https://github.com/libp2p/rust-libp2p/pull/1726.

PR #1719, opened 2020-08-25
protocols/kad: Implement log2 for Distance

mxinden:
In order to make use of the distances returned by `KBucketRef::range` in
a human readable format, one needs to be able to tranlate the 256 bit in
a `Distance` to a smaller space, e.g. the currently used 8 bit bucket
scheme. To be able to do so one needs to be able to extract the 256 bit
from a `Distance`.

This commit makes the `U256` within a `Distance` public allowing callers
of `KBucketRef::range` to work with the raw bits themselves.

---

Follow up on https://github.com/libp2p/rust-libp2p/pull/1680.

With this pull request the implementation internal struct `U256` is exposed on the public interface. That is not great. I have been unable to come up with a better way of exposing this information without tying the public interface to the internal 256-bit-fixed-array-implementation (see https://github.com/libp2p/rust-libp2p/pull/1680#issuecomment-664420438 for past discussion). Instead of exposing `U256` itself, one could as well offer a method on `Distance` returning a `[u8]` instead.

I don't think simply exposing a 256bit bucket identifier via `Distance`'s `Debug` implementation in logs or metrics is of much help for any human being.

Thoughts?


romanb:
> In order to make use of the distances returned by KBucketRef::range in
a human readable format, one needs to be able to tranlate the 256 bit in
a Distance to a smaller space, e.g. the currently used 8 bit bucket
scheme.

What format do you have in mind and what do you mean by translating it to an "8 bit bucket scheme"? External code shouldn't presume there to be 256 buckets either. Since you mention a human-readable format for logs and metrics, wouldn't it be appropriate to implement `Display` for `Distance`?

romanb:
> What format do you have in mind and what do you mean by translating it to an "8 bit bucket scheme"? External code shouldn't presume there to be 256 buckets either.

Oh, I guess you just want to show or distribute the distances on a (base 2) logarithmic scale, independent of how many buckets there are? I think I would have a slight preference to just offering functions like `Distance::log2()` over making `uint` crate types part of the public API.

protocols/kad/src/kbucket/key.rs
```diff
@@ -169,6 +169,13 @@ impl AsRef<KeyBytes> for KeyBytes {
 #[derive(Copy, Clone, PartialEq, Eq, Default, PartialOrd, Ord, Debug)]
 pub struct Distance(pub(super) U256);
 
+impl Distance {
+    /// Returns the base 2 logarithm of the [`Distance`].
+    pub fn log2(&self) -> Option<u32> {
+        (256 - self.0.leading_zeros()).checked_sub(1)
```
mxinden, line 175: Instead of the hard-coded `256` here, we could as well derive it. But we would then depend on the internal representation of a `U256` instead. Let me know what you think.

```suggestion
        (U256::MAX.bits() - self.0.leading_zeros()).checked_sub(1)
```

mxinden (commented):
We could now use `Distance::log2` in `BucketIndex::new`.

https://github.com/libp2p/rust-libp2p/blob/7b415d5e7040e45c541f76f2c409e63d4d3249c6/protocols/kad/src/kbucket.rs#L108-L112

As a downside that would blur the lines between the public `log2` and the internal bucket scheme. I am fine either way. Thoughts?

protocols/kad/src/kbucket/key.rs
```diff
@@ -167,7 +167,14 @@ impl AsRef<KeyBytes> for KeyBytes {
 
 /// A distance between two keys in the DHT keyspace.
 #[derive(Copy, Clone, PartialEq, Eq, Default, PartialOrd, Ord, Debug)]
-pub struct Distance(pub U256);
+pub struct Distance(pub(super) U256);
+
+impl Distance {
+    /// Returns the base 2 logarithm of the [`Distance`].
+    pub fn log2(&self) -> Option<u32> {
+        (256 - self.0.leading_zeros()).checked_sub(1)
```
romanb, line 175:
```suggestion
    /// Returns the integer part of the base 2 logarithm of the [`Distance`].
    pub fn ilog2(&self) -> Option<u32> {
        (256 - self.0.leading_zeros()).checked_sub(1)
```
Just for added clarity.

protocols/kad/src/kbucket/key.rs
romanb: Either way sounds fine to me.

romanb (approved):

romanb:
> We could now use Distance::log2 in BucketIndex::new.

I think we should do that.

protocols/kad/src/kbucket/key.rs
mxinden: Leaving as is then.

PR #1720, opened 2020-08-27
Gossipsub v1.1

AgeManning:
# Description

This PR upgrades the current gossipsub implementation to support the [v1.1 spec](https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/gossipsub-v1.1.md). 

It adds a number of features, bug fixes and performance improvements. 

Besides support for all new 1.1 features, other improvements that are of particular note: 

- Improved duplicate LRU-time cache (this was previously a severe bottleneck for large message throughput topics)
- Extended message validation configuration options
- Arbitrary topics (users can now implement their own hashing schemes)
- Improved message validation handling - Invalid messages are no longer dropped but sent to the behaviour for application-level processing (including scoring)
- Support for floodsub, gossipsub v1 and gossipsub v2 
- Protobuf encoding has been shifted into the behaviour. This has permitted two improvements:
     1. Message size verification during publishing (report to the user if the message is too large before attempting to send).
     2. Message fragmentation. If an RPC is too large it is fragmented into its sub components and sent in smaller chunks. 

### Additional Notes

The peer eXchange protocol defined in the v1.1 spec is inactive in its current form. The current implementation permits sending `PeerId` in `PRUNE` messages, however a `PeerId` is not sufficient to form a new connection to a peer. A `Signed Address Record` is required to safely transmit peer identity information. Once these are confirmed (https://github.com/libp2p/specs/pull/217) a future PR will implement these and make PX usable. 

##### Final Note

This is currently undergoing further testing and an audit and I may modify this slightly as bugs come up :)


mxinden:
> This is currently undergoing further testing and an audit and I may modify this slightly as bugs come up :)

@agemanning would you like this to be reviewed already, or wait till the mentioned testing is finished?

AgeManning:
The audit for this will start tomorrow and finish at the end of September. 

Throughout September, we'll also be ramping up our testing and analysis of how it performs at scale. I think it would also help getting some eyes on this from yourself. 

Depending on how you guys want to release this. Potentially we can review this as is, and then make changes in a seperate PR later with audit and testing modifications. 

Alternatively if people would rather wait for a more complete version, we can wait until the end of September and include all modifications in this PR. 

As of tomorrow, this should be a fairly stable version and I won't be changing it much (unless we find bugs) so it's ready for review :)

protocols/gossipsub/src/backoff.rs
```diff
@@ -0,0 +1,176 @@
...
+    hash_map::{Entry, HashMap},
+    HashSet,
+};
+use std::time::Duration;
+use wasm_timer::Instant;
+
+/// Stores backoffs in an efficient manner.
+pub struct BackoffStorage {
+    /// Stores backoffs and the index in backoffs_by_heartbeat per peer per topic.
+    backoffs: HashMap<TopicHash, HashMap<PeerId, (Instant, usize)>>,
```
mxinden, line 34: To gain more type safety you could wrap this `usize` into a new type (new-type-pattern), implementing [`Index`](https://doc.rust-lang.org/std/ops/trait.Index.html) for that new type.

Example: https://github.com/libp2p/rust-libp2p/pull/1473/files#diff-1c07a3e4093523a2fb794d47a23f587fR324

protocols/gossipsub/src/behaviour.rs
```diff
@@ -87,21 +102,51 @@ pub enum MessageAuthenticity {
...
-    fn is_anonymous(&self) -> bool {
+    pub fn is_anonymous(&self) -> bool {
         match self {
             MessageAuthenticity::Anonymous => true,
             _ => false,
         }
     }
 }
 
+/// Event that can happen on the gossipsub behaviour.
```
mxinden, line 120:
```suggestion
/// Event that can be emitted by the gossipsub behaviour.
```

protocols/gossipsub/src/behaviour.rs
```diff
@@ -87,21 +102,51 @@ pub enum MessageAuthenticity {
...
+        message: GossipsubMessage,
+    },
+    /// A remote subscribed to a topic.
+    Subscribed {
+        /// Remote that has subscribed.
+        peer_id: PeerId,
+        /// The topic it has subscribed to.
+        topic: TopicHash,
+    },
+
```
mxinden, line 140:
```suggestion
```
Either removing this one, or inserting a newline above.

protocols/gossipsub/src/behaviour.rs
```diff
@@ -163,14 +218,25 @@ pub struct Gossipsub {
 
     /// An LRU Time cache for storing seen messages (based on their ID). This cache prevents
     /// duplicates from being propagated to the application and on the network.
-    duplication_cache: LruCache<MessageId, ()>,
+    duplication_cache: DuplicateCache<MessageId>,
```
mxinden, line 221:
```suggestion
    duplicate_cache: DuplicateCache<MessageId>,
```

protocols/gossipsub/src/behaviour.rs
```diff
@@ -180,45 +246,89 @@ pub struct Gossipsub {
...
+    ///Storage for backoffs
+    backoffs: BackoffStorage,
+
     /// Message cache for the last few heartbeats.
     mcache: MessageCache,
 
     /// Heartbeat interval stream.
     heartbeat: Interval,
+
+    /// number of heartbeats since the beginning of time; this allows us to amortize some resource
```
mxinden, line 258:
```suggestion
    /// Number of heartbeats since the beginning of time; this allows us to amortize some resource
```

mxinden (commented):
Just a couple of small nits. I need to spend more time to fully understand Gossipsub v1.1 to give a thorough review.

protocols/gossipsub/src/behaviour/tests.rs
```diff
@@ -22,57 +22,307 @@
 
 #[cfg(test)]
 mod tests {
```
mxinden, line 24: How about not wrapping the entire content of the file in the `mod tests {}`, but instead move the `#[cfg(test)]` to `lib.rs` where the `tests.rs` module is introduced?

protocols/gossipsub/src/config.rs
```diff
@@ -48,61 +50,288 @@ pub enum ValidationMode {
...
     /// Duplicates are prevented by storing message id's of known messages in an LRU time cache.
     /// This settings sets the time period that messages are stored in the cache. Duplicates can be
     /// received if duplicate messages are sent at a time greater than this setting apart. The
     /// default is 1 minute.
-    pub duplicate_cache_time: Duration,
+    duplicate_cache_time: Duration,
+
+    /// When set to `true`, prevents automatic forwarding of all received messages. This setting
+    /// allows a user to validate the messages before propagating them to their peers. If set to
+    /// true, the user must manually call `validate_message()` on the behaviour to forward message
```
mxinden, line 120:
```suggestion
    /// true, the user must manually call [`Gossipsub::report_message_validation_result`] on the behaviour to forward message
```

protocols/gossipsub/src/config.rs
```diff
@@ -48,61 +50,288 @@ pub enum ValidationMode {
...
+    /// once validated (default is `false`). Furthermore, the application may optionally call
+    /// `invalidate_message()` on the behaviour to remove the message from the memcache. The
```
mxinden, line 122: I think this is outdated and now included in `report_message_validation_result`, right?

protocols/gossipsub/src/config.rs
```diff
@@ -48,61 +50,288 @@ pub enum ValidationMode {
...
+    /// The default value is to concatenate the source peer id with a sequence number. Setting this
+    /// parameter allows the user to address packets arbitrarily. One example is content based
+    /// addressing, where this function may be set to `hash(message)`. This would prevent messages
+    /// of the same content from being duplicated.
+    ///
+    /// The function takes a `GossipsubMessage` as input and outputs a String to be interpreted as
+    /// the message id.
+    message_id_fn: fn(&GossipsubMessage) -> MessageId,
+
+    /// By default, gossipsub will reject messages that are sent to us that has the same message
```
mxinden, line 140:
```suggestion
    /// By default, gossipsub will reject messages that are sent to us that have the same message
```

protocols/gossipsub/src/config.rs
```diff
@@ -48,61 +50,288 @@ pub enum ValidationMode {
...
+
+    /// Number of heartbeat slots considered as slack for backoffs. This gurantees that we wait
+    /// at least backoff_slack heartbeats after a backoff is over before we try to graft. This
+    /// solves problems occuring through high latencies. In particular if
+    /// `backoff_slack * heartbeat_interval` is longer than any latencies between processing
+    /// prunes on our side and processing prunes on the receiving side this guarantees that we
+    /// get not punished for too early grafting. The default is 1.
+    backoff_slack: u32,
+
+    /// Whether to do flood publishing or not. If enabled newly created messages will always be
```
mxinden, line 172:
```suggestion
    /// Whether to do flood publishing or not. If enabled newly created messages authored by the local node will always be
```

protocols/gossipsub/src/config.rs
```diff
@@ -48,61 +50,288 @@ pub enum ValidationMode {
...
+    iwant_followup_time: Duration,
+
+    /// Enable support for flooodsub peers. Default false.
+    support_floodsub: bool,
+}
+
+impl GossipsubConfig {
+    //all the getters
+
+    /// The protocol id prefix to negotiate this protocol. The protocol id is of the form
```
mxinden, line 225: Do we need to duplicate all the comments here? Given that all `GossipConfig` properties are private, I think it is enough to only have the comments on the `impl GossipsubConfig {}` block. This would also match the style Kademlia is using.

https://github.com/libp2p/rust-libp2p/blob/master/protocols/kad/src/behaviour.rs#L160

protocols/gossipsub/src/config.rs
```diff
@@ -302,33 +669,200 @@ impl GossipsubConfigBuilder {
...
+        if !(self.config.mesh_outbound_min * 2 <= self.config.mesh_n) {
+            return Err(
+                "The following inequality doesn't hold mesh_outbound_min <= self.config.mesh_n / 2",
+            );
+        }
+        Ok(self.config.clone())
     }
 }
 
 impl std::fmt::Debug for GossipsubConfig {
```
mxinden, line 818: As far as I can tell none of the fields are treated special. Why not derive `Debug`?

protocols/gossipsub/src/gossip_promises.rs
```diff
@@ -0,0 +1,98 @@
...
+            // If a promise for this message id and peer already exists we don't update expires!
+            self.promises
+                .entry(message_id.clone())
+                .or_insert_with(|| HashMap::new())
+                .entry(peer.clone())
+                .or_insert(expires);
+        }
+    }
+
+    pub fn deliver_message(&mut self, message_id: &MessageId) {
```
mxinden, line 56:
```suggestion
    pub fn message_delivered(&mut self, message_id: &MessageId) {
```

This message doesn't actively deliver a message, but instead tracks that a message was delivered. Wouldn't this name be more intuitive?

protocols/gossipsub/src/gossip_promises.rs
```diff
@@ -0,0 +1,98 @@
...
+        // Someone delivered a message, we can stop tracking all promises for it
```
mxinden, line 57:
```suggestion
        // Someone delivered a message, we can stop tracking all promises for it.
```

protocols/gossipsub/src/handler.rs
```diff
@@ -49,12 +76,34 @@ pub struct GossipsubHandler {
...
+
+    /// The type of peer this handler is associated to.
+    peer_kind: Option<PeerKind>,
+
+    /// Keeps track on whether we have sent the peer kind to the behaviour.
+    // NOTE: Use this flag rather than checking the substream count each poll.
+    peer_kind_sent: bool,
+
+    /// If the peer doesn't support the gossipsub protocol we do not immediately disconnect.
+    /// Rather, we disable the handler and prevent any ingoing or outgoing substreams from being
```
mxinden, line 99:
```suggestion
    /// Rather, we disable the handler and prevent any incoming or outgoing substreams from being
```

protocols/gossipsub/src/handler.rs
```diff
@@ -49,12 +76,34 @@ pub struct GossipsubHandler {
...
+    inbound_substreams_created: usize,
+
+    /// The type of peer this handler is associated to.
+    peer_kind: Option<PeerKind>,
+
+    /// Keeps track on whether we have sent the peer kind to the behaviour.
+    // NOTE: Use this flag rather than checking the substream count each poll.
+    peer_kind_sent: bool,
+
+    /// If the peer doesn't support the gossipsub protocol we do not immediately disconnect.
```
mxinden, line 98: Can you explain why the handler is kept alive even though gossipsub does not utilize the underlying connection?

mxinden (commented):
Still only small nits that caught my attention while understanding the greater picture.

AgeManning:
Sorry for the delay. There are some known issues we're fixing up and modifying. Will address these very soon.

romanb:
Am I understanding correctly that this will essentially supersede `libp2p-floodsub`? Given that they both share the same base protocol and signing options it probably makes sense. If you can confirm that, then we can already close https://github.com/libp2p/rust-libp2p/pull/1015 as redundant, as the message sizes are already configurable for gossipsub. This PR will then also close https://github.com/libp2p/rust-libp2p/issues/991 since we're going to deprecate the `libp2p-floodsub` crate.

AgeManning:
In principle yes. Tbh though the floodsub side of this has not been extensively tested. 

Currently we haven't set up a configuration parameter that converts this to only support floodsub. Currently we can handle floodsub peers connecting to us, but we always advertise the meshsub topics and handle gossipsub peers also. 

If we want to deprecate floodsub, we should probably allow a user to configure this protocol to only support the floodsub id and reject gossipsub nodes (for users wishing to only use floodsub). 

This PR is very close to being complete. Will be addressing max's comments this week :)

romanb:
> In principle yes. Tbh though the floodsub side of this has not been extensively tested.

Well, I think `libp2p-floodsub` currently has no tests at all, except maybe the chat example, so anything is an improvement.

> If we want to deprecate floodsub, we should probably allow a user to configure this protocol to only support the floodsub id and reject gossipsub nodes (for users wishing to only use floodsub).

I think that would be good. I started updating `libp2p-floodsub` only to realise that a lot of I would be doing has already been done in `libp2p-gossipsub`, since they share the same [pubsub base protocol](https://github.com/libp2p/specs/tree/master/pubsub). I would even go as far as saying that it would be nice to eventually rebrand `libp2p-gossipsub` as `libp2p-pubsub`, supporting both floodsub and gossipsub, reflecting the structure of the specifications.

protocols/gossipsub/src/config.rs
AgeManning: Correct.

protocols/gossipsub/src/config.rs
AgeManning: Yep, I'll shift them there.

protocols/gossipsub/src/config.rs
AgeManning: I think the derive cannot handle the `message_id_fn` and `fast_message_id_fn` types in general. So we exclude them in a custom implementation.

protocols/gossipsub/src/gossip_promises.rs
AgeManning: makes sense to me

protocols/gossipsub/src/handler.rs
AgeManning: IIRC, the previous functionality is such that if a peer does not support gossipsub or floodsub it would be immediately disconnected. 

This allows the handler to not disconnect but rather set the KeepAlive to false. 

I think this is important for more complex NetworkBehaviours/ProtocolHandlers that may allow for other protocols and may wish to keep connections with peers that don't support gossipsub but may support another behaviour. In this way, the global network behaviour can drop peers based on keep alive and have the option to drop peers based on their supported protocols.

AgeManning:
@mxinden I've addressed most of your comments. 

I think this is ready to be properly reviewed and attempt to merge it down. 

It doesn't have the necessary logic to replace floodsub yet, but we can get to that in a future PR.

Dev380:
Is this going to happen anytime soon?

mxinden:
@Dev380 in case you are referring to the review process itself, I will do an in-depth review this week.

@AgeManning thanks for the heads-up and sorry for the delay.

mxinden:
As promised I gave this another review, see https://github.com/sigp/rust-libp2p/pull/93.

> Given the size of libp2p#1720 I did my review directly inline. A lot of suggestions are just related to styling - I hope delivering my review as a pull request makes accepting those easier. Comments are prefixed with // mxinden:.
>
> Let me know if this format works for you. If not, I can transform this into a traditional Github review.
>
> Again, as libp2p#1720 is quite large, I have not yet fully reviewed all changes / additions.

AgeManning:
Thanks max. I think we've addressed all the comments. I've merged into your branch there. Let me know if anything is missed. 

If all good, I'll merge that branch down and then merge latest master.

jolestar:
It is possible to support a custom peer selector/filter strategy for publishing messages?

AgeManning:
> It is possible to support a custom peer selector/filter strategy for publishing messages?

We could add this in, but I'm not sure why you would want to. What is it's intended purpose? Do you want it to apply just to publishing messages or also to forwarding?

Lets say you want to publish to only half the peers in your mesh (based on some filter). Those peers would then forward it to peers in their mesh and the message would propagate regardless (with added latency), including to those you don't directly publish to. This filtering would also affect the scoring that is implemented in 1.1 as some nodes would look like they are censoring. 

Maybe if you could detail the reasons why you might want this I could provide more help.

jolestar:
> We could add this in, but I'm not sure why you would want to. What is it's intended purpose? Do you want it to apply just to publishing messages or also to forwarding?
> 
> Lets say you want to publish to only half the peers in your mesh (based on some filter). Those peers would then forward it to peers in their mesh and the message would propagate regardless (with added latency), including to those you don't directly publish to. This filtering would also affect the scoring that is implemented in 1.1 as some nodes would look like they are censoring.
> 
> Maybe if you could detail the reasons why you might want this I could provide more help.

Thanks. My usage scenario is blockchain. I want to publish messages to some peers base on the peer's state,  such as peer's type(lighting node/full node),  block_height, or total_difficulity. If the Gossipsub does not support a custom strategy, the application needs to implement a custom Gossip protocol, such as subtract(https://github.com/paritytech/substrat)' network/network-gossip.  But the hardest part of the custom strategy is how to maintain the custom state of the peer. if the custom state can not keep in Gossipsub's peer state, it will very hard to implement a peer filter strategy.

blacktemplar:
> > We could add this in, but I'm not sure why you would want to. What is it's intended purpose? Do you want it to apply just to publishing messages or also to forwarding?
> > Lets say you want to publish to only half the peers in your mesh (based on some filter). Those peers would then forward it to peers in their mesh and the message would propagate regardless (with added latency), including to those you don't directly publish to. This filtering would also affect the scoring that is implemented in 1.1 as some nodes would look like they are censoring.
> > Maybe if you could detail the reasons why you might want this I could provide more help.
> 
> Thanks. My usage scenario is blockchain. I want to publish messages to some peers base on the peer's state, such as peer's type(lighting node/full node), block_height, or total_difficulity. If the Gossipsub does not support a custom strategy, the application needs to implement a custom Gossip protocol, such as subtract(https://github.com/paritytech/substrat) network/network-gossip. But the hardest part of the custom strategy is how to maintain the custom state of the peer. if the custom state can not keep in Gossipsub's peer state, it will very hard to implement a peer filter strategy.

I think you should try to solve that with dividing your topics. If you really need dynamic strategies for distributing messages I am not sure if gossipsub is the right way to do it. You would have to think how to handle the different states in your IHAVE messages too, so it becomes very complicated very fast...

jolestar:
> > > We could add this in, but I'm not sure why you would want to. What is it's intended purpose? Do you want it to apply just to publishing messages or also to forwarding?
> > > Lets say you want to publish to only half the peers in your mesh (based on some filter). Those peers would then forward it to peers in their mesh and the message would propagate regardless (with added latency), including to those you don't directly publish to. This filtering would also affect the scoring that is implemented in 1.1 as some nodes would look like they are censoring.
> > > Maybe if you could detail the reasons why you might want this I could provide more help.
> > 
> > 
> > Thanks. My usage scenario is blockchain. I want to publish messages to some peers base on the peer's state, such as peer's type(lighting node/full node), block_height, or total_difficulity. If the Gossipsub does not support a custom strategy, the application needs to implement a custom Gossip protocol, such as subtract(https://github.com/paritytech/substrat) network/network-gossip. But the hardest part of the custom strategy is how to maintain the custom state of the peer. if the custom state can not keep in Gossipsub's peer state, it will very hard to implement a peer filter strategy.
> 
> I think you should try to solve that with dividing your topics. If you really need dynamic strategies for distributing messages I am not sure if gossipsub is the right way to do it. You would have to think how to handle the different states in your IHAVE messages too, so it becomes very complicated very fast.

The gossipsub protocol broadcast message base node's knowledge base, node exchange knowledge by IHAVE message. But IHAVE only support MessageID Vec, I think whether it is possible to support a weighted id, the weight may be message's index or other sortable property, it can help to filter peer and reduce repeat gossip messages.

blacktemplar:
> > > > We could add this in, but I'm not sure why you would want to. What is it's intended purpose? Do you want it to apply just to publishing messages or also to forwarding?
> > > > Lets say you want to publish to only half the peers in your mesh (based on some filter). Those peers would then forward it to peers in their mesh and the message would propagate regardless (with added latency), including to those you don't directly publish to. This filtering would also affect the scoring that is implemented in 1.1 as some nodes would look like they are censoring.
> > > > Maybe if you could detail the reasons why you might want this I could provide more help.
> > > 
> > > 
> > > Thanks. My usage scenario is blockchain. I want to publish messages to some peers base on the peer's state, such as peer's type(lighting node/full node), block_height, or total_difficulity. If the Gossipsub does not support a custom strategy, the application needs to implement a custom Gossip protocol, such as subtract(https://github.com/paritytech/substrat) network/network-gossip. But the hardest part of the custom strategy is how to maintain the custom state of the peer. if the custom state can not keep in Gossipsub's peer state, it will very hard to implement a peer filter strategy.
> > 
> > 
> > I think you should try to solve that with dividing your topics. If you really need dynamic strategies for distributing messages I am not sure if gossipsub is the right way to do it. You would have to think how to handle the different states in your IHAVE messages too, so it becomes very complicated very fast.
> 
> The gossipsub protocol broadcast message base node's knowledge base, node exchange knowledge by IHAVE message. But IHAVE only support MessageID Vec, I think whether it is possible to support a weighted id, the weight may be message's index or other sortable property, it can help to filter peer and reduce repeat gossip messages.

I am not sure if I understand you correctly, but if you propose additions to the gossipsub protocol you are probably better of to propose them in https://github.com/libp2p/specs/.

protocols/gossipsub/src/transform.rs
```diff
@@ -0,0 +1,58 @@
+//! This trait allows of extended user-level decoding that can apply to message-data before a
```
mxinden, line 1: Can you add a license header here?

mxinden (approved):
I would be in favor of merging this pull request into `master`. As far as I can tell, the only outstanding task would be to fix the intra-doc links. What do you think @AgeManning @blacktemplar?

protocols/gossipsub/src/transform.rs
AgeManning: Can do. 

I think its mostly stable now. Will fix the doc links.

AgeManning:
Fixed the doclinks and merged latest master.

mxinden:
Great. I will merge later today unless anyone objects till then.

mxinden:
:tada: Thank you for the extensive amount of work you put into this.

protocols/gossipsub/src/peer_score/params.rs
```diff
@@ -0,0 +1,399 @@
...
+}
+
+impl Default for PeerScoreThresholds {
+    fn default() -> Self {
+        PeerScoreThresholds {
+            gossip_threshold: -10.0,
+            publish_threshold: -50.0,
+            graylist_threshold: -80.0,
+            accept_px_threshold: 10.0,
+            opportunistic_graft_threshold: 20.0,
```
diegomrsantos, line 80: @AgeManning could you please share how those values were defined?

AgeManning: These essentially can be defined for whatever network you are running gossipsub on. Its been a few years, but I believe these were from the go-implementation at the time, which may have been used in some of the tests in the v1.1 gossipsub paper where they tested a few attack vectors. 

These numbers shouldn't be set independently, but based on the maximum and minimum values you can get from the score.

diegomrsantos: I found this PR that says the values were inspired by LH https://github.com/status-im/nimbus-eth2/pull/2091. These are the values used:

```
     gossipThreshold: -4000,
     publishThreshold: -8000,
     graylistThreshold: -16000
```

They are used here indeed https://gist.github.com/blacktemplar/5c1862cb3f0e32a1a7fb0b25e79e6e2c#file-51000-toml-L33, but LH doesn't use to configure GossipSub, but rather here https://github.com/libp2p/rust-libp2p/blob/25d14349cc32f7a8e3feffb931ba2e84cccb0013/protocols/gossipsub/src/metrics.rs#L76

AgeManning: Now I"m confused about what you are asking. 

The values in this repository i.e `gossip_threshold: -10` is not the same value as what we have chosen in the Ethereum consensus network i.e `-4000`. 

As I mentioned, these values need to be defined for whatever network you are running. For Ethereum's consensus network, we did a bunch of estimations and determined a bunch of numbers. That is entirely independent of the `gossipsub` protocol, which should run on any network. 

If you question is specific to lighthouse or the Ethereum network, then that is a different question and probably should be asked on the lighthouse repo as those values are for a specific network. 

You did point out that the metrics value uses the Ethereum consensus values as default. This probably should be using the same defaults as the rust-libp2p repo. At the time, Lighthouse was the only one consuming these metrics and so was built around our scoring values to get sane results in the histograms.

diegomrsantos: You're right, sorry for the confusion. It's clear now, thanks. I somehow misled myself looking at this PR that introduces params in the lighthouse rust-libp2p fork https://github.com/sigp/rust-libp2p/pull/556/files#diff-9f622a0085d0d96aef9e883b7afdcdd4a652e8801be7cfabf82819f5a74a70e0. Is there a lighthouse PR defining the values for lighthouse?

AgeManning: What specifically are you interested in?

Are you asking how we figured out what values to use in Lighthouse? There is an external document I can share with you if that's the case. The PR to lighthouse will likely just show values being added with no context. 

Feel free to DM me in the Lighthouse discord if you like

Issue #1721, opened 2020-08-28
Send-back address in MemoryTransport is wrong


tomaka:
When the memory transport receives an incoming connection, it erroneously reports its own port as the remote address: https://github.com/libp2p/rust-libp2p/blob/0136af50253ea8d608367d0bc5514a65fb7b083a/core/src/transport/memory.rs#L193

PR #1724, opened 2020-08-31
core/transport/memory: Return dialer address in Upgrade event

mxinden:
Previously a `Listener` would return its own address as the
`remote_addr` in the `ListenerEvent::Upgrade` event.

With this commit a `Listener` returns the dialer address as the
`remote_addr` in the `ListenerEvent::Upgrade` event. To do so a
`DialFuture` registers a port with the global `HUB` at construction
which is later on unregistered in the `Drop` implementation of the
dialer's `Chan`. The sending side of the `mpsc::channel` registered in
the `HUB` is dropped at `DialFuture` construction, thus one can not dial
the dialer port. This mimics the TCP transport behaviour preventing both
dialing and listening on the same TCP port.

Closes https://github.com/libp2p/rust-libp2p/issues/1721.

### DIFF
--- a/core/CHANGELOG.md
+++ b/core/CHANGELOG.md
@@ -16,6 +16,10 @@
   two peer IDs are equal if and only if they use the same hash algorithm and
   have the same hash digest. [PR 1608](https://github.com/libp2p/rust-libp2p/pull/1608).
 
+- Return dialer address instead of listener address as `remote_addr` in
+  `MemoryTransport` `Listener` `ListenerEvent::Upgrade`
+  [PR 1724](https://github.com/libp2p/rust-libp2p/pull/1724).
+
 # 0.21.0 [2020-08-18]
 
 - Remove duplicates when performing address translation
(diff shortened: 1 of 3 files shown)


mxinden:
> #1721 does not mention what problems the current behaviour causes. Is it simply that the ListenerEvent::Upgrade::remote_addr is the same for all connections with the memory transport?

That was my impression but maybe @tomaka has more information in that regard?

tomaka:
I opened #1721 after looking at Substrate logs and realizing that the local address and send back address were the same.
I admit that I didn't think this through too much. I immediately assumed that it was wrong and opened an issue to not forget about this problem. But it is in the end not obvious than it is a problem.

mxinden:
I think the behavior introduced in this pull request is at least more intuitive than the current behavior. Thus I am voting to merge this pull request. Objections?

romanb:
Please don't forget to update the changelog, either before merging or after.

mxinden:
> Please don't forget to update the changelog, either before merging or after.

Thanks for the reminder @romanb! It slipped my mind.

PR #1726, opened 2020-09-01
Restore `RequestResponse::throttled`.

twittner:
In contrast to the existing "throttled" approach this PR adds back-pressure to the protocol without requiring pre-existing knowledge of all node limits. It adds small, CBOR-encoded credit grant messages which communicate back to the sender how many more requests it is allowed to send.

Alternative to https://github.com/libp2p/rust-libp2p/pull/1715.


protocols/request-response/src/lib.rs
```diff
@@ -210,14 +217,23 @@ impl<TResponse> ResponseChannel<TResponse> {
     pub fn is_open(&self) -> bool {
         !self.sender.is_canceled()
     }
+
+    /// Get the request ID of this response.
```
romanb, line 221:
```suggestion
    /// Get the ID of the inbound request waiting for a response.
```

protocols/request-response/src/codec/header.rs
```diff
@@ -0,0 +1,244 @@
...
+/// A wrapper around a `ProtocolName` impl which augments the protocol name.
+///
+/// The type implements `ProtocolName` itself and creates a name for a
+/// request-response protocol based on the protocol name of the wrapped type.
+#[derive(Debug, Clone)]
+pub struct ProtocolWrapper<P>(P, Bytes);
+
+impl<P: ProtocolName> ProtocolWrapper<P> {
+    pub fn v1(p: P) -> Self {
+        let mut full = BytesMut::from(b"/request-response/header/1".as_ref());
```
romanb, line 113:
```suggestion
        let mut full = BytesMut::from(b"/request-response/throttled/1".as_ref());
```

protocols/request-response/src/codec/header.rs
```diff
@@ -0,0 +1,244 @@
...
+        (self.header, self.data)
+    }
+}
+
+/// A wrapper around a `ProtocolName` impl which augments the protocol name.
+///
+/// The type implements `ProtocolName` itself and creates a name for a
+/// request-response protocol based on the protocol name of the wrapped type.
+#[derive(Debug, Clone)]
+pub struct ProtocolWrapper<P>(P, Bytes);
```
romanb, line 109:
```suggestion
pub struct ThrottledProtocol<P>(P, Bytes);
```

protocols/request-response/src/throttled.rs
```diff
@@ -18,285 +18,495 @@
...
+            if let Some(info) = self.peer_info.get_mut(&p) {
+                log::trace!("{:08x}: response {} sent", self.id, r);
+                if info.recv_budget == 0 { // need to send more credit to the remote peer
+                    let crd = info.limit.switch();
+                    info.recv_budget = info.limit.max_recv.get();
+                    let cid = self.next_credit_id();
+                    let rid = self.behaviour.send_request(&p, Message::credit(crd, cid));
+                    log::trace!("{:08x}: sending {} as credit {} to {}", self.id, crd, cid, p);
+                    let credit = Credit { id: cid, request: rid, amount: crd };
+                    self.credit_messages.insert(p.clone(), credit);
```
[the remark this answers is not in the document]
twittner: `ResponseSent` was pretty much a leftover from a previous implementation approach and the correctness should not depend on the order of responses vs. credit grants. I have removed `ResponseSent` altogether and moved the grant sending into `send_response`.

protocols/request-response/src/codec/header.rs
twittner: I wonder if we should really use such long prefixes which are negotiated all the time. Since the actual protocol can be assumed to be fairly unique, I switched to a short `/t/1` prefix instead. Currently the multistream-select related data is sometimes larger than the actual request-response data.

protocols/request-response/src/codec/header.rs
twittner: I generalised `ProtocolWrapper` somewhat so it deserves that name.

protocols/request-response/src/codec/header.rs
romanb: Yes, I was thinking the same, though my first instinct was to just drop (or shorten to `/rr` or `/rqrs`) the `/request-response` prefix, also because in principle if there are multiple such wrapper protocols they may even be stacked on top of each other, which would repeat the `request-response` prefix (e.g. `/throttled/1/foobar/1` rather than `/request-response/throttled/1/request-response/foobar/1`).

romanb (approved):

mxinden:
As far as I know all of libp2p uses Protobuf for binary encoding. Can you explain what benefit is gained by encoding `Header` via cbor instead of Protobuf? Is this benefit worth the lost consistency across the project?

twittner:
> As far as I know all of libp2p uses Protobuf for binary encoding.

Mplex and yamux encode their headers in a custom format. Multi-addresses use another custom binary encoding. Peer IDs are multihashes, multistream-select uses length-prefixed and `\n`-terminated byte strings. And yes, other parts use protocol buffers.

> Can you explain what benefit is gained by encoding `Header` via cbor instead of Protobuf?

Contrary to protocol buffers, CBOR is an IETF specification ([RFC 7049](https://tools.ietf.org/html/rfc7049)) and not merely implementation defined. It is also rather lightweight w.r.t. tooling requirements. As you can see, there is no need for another compiler like `protoc`, no extra-language schema file, no build.rs. The full expressiveness of Rust can be used, e.g. generic types, sum types etc. without difficulties (but granted, `Header` in particular is not a complicated type). There is also no code generation that creates an extra set of types to match some schema language.

> Is this benefit worth the lost consistency across the project?

I do not agree with this consistency claim which rests on a particular way to look at the various libp2p bits and pieces. From my perspective encodings are everywhere and nothing is lost by embracing this variety.

mxinden:
> Mplex and yamux encode their headers in a custom format. Multi-addresses use another custom binary encoding. Peer IDs are multihashes, multistream-select uses length-prefixed and \n-terminated byte strings. And yes, other parts use protocol buffers.

I stand corrected! :slightly_smiling_face: 

Thanks for the detailed explanation @twittner. I would prefer to have a single generic binary encoding scheme just for the sake of reducing complexity. At the same time this opinion is not a strong one and should not block this pull request.

Issue #1622, opened 2020-06-19
protocols/kad: Default maximum message size not matching with Golang


mxinden:
The Kademlia Golang implementation [configures its `VarintReader`](https://github.com/libp2p/go-libp2p-kad-dht/blob/330b9beabaacd7e44aa8c80c19435b1bf1081212/dht_net.go#L78) with a maximum message size of [4 MB](https://github.com/libp2p/go-libp2p-core/blob/9cf214f5019d0aebb1f5dca184db276284e7639e/network/network.go#L24) whereas the the Kademlia Rust implementation [configures its `UviBytes`](https://github.com/libp2p/rust-libp2p/blob/e8e41f4efb129d556ac768ea84737ba54bd7ca28/protocols/kad/src/protocol.rs#L194) with a maximum message size of [4096 B](https://github.com/libp2p/rust-libp2p/blob/master/protocols/kad/src/protocol.rs#L170).

4096 B might be a bit conservative as a default. E.g. when connecting to the IPFS DHT with https://github.com/mxinden/kademlia-exporter it oftentimes runs into the issue:

```
[2020-04-11T22:45:24Z DEBUG libp2p_kad::behaviour] Request to PeerId("") in query QueryId(0) Failed with Io(Custom { kind: PermissionDenied, error: "len > max" })
```

This limit was introduced in https://github.com/libp2p/rust-libp2p/pull/920 and is motivated by https://github.com/paritytech/substrate/issues/1645.

I would suggest quadrupling the value (16384 B) as doubling it seems to work fine for discovery usage with the Golang implementation leaving some additional head-space.

Thoughts?

romanb:
Increasing the default for better interoperability sounds good to me. If the 4KiB limit is still desired for substrate, we can set it there, since this limit is already configurable.

romanb:
Closed by #1730.

PR #1730, opened 2020-09-07
[kad] Increase default payload limits.

romanb:
For better interoperability by default, picking up the suggestion of @mxinden  in https://github.com/libp2p/rust-libp2p/issues/1622.


mxinden (approved):
Sorry for letting https://github.com/libp2p/rust-libp2p/issues/1622 go stale. Thanks for the follow up.

PR #1731, opened 2020-09-07
Add back the LRU cache to `Throttled`.

twittner:
Peers may send all their requests, reconnect and send again all their requests, starting from a fresh budget. The LRU cache keeps the peer information around and reuses it when the peer reconnects, continuing with the previous remaining limit.

On top of #1726.


romanb (approved):
:+1:

PR #1732, opened 2020-09-07
[core] Simplify accepting incoming connections.

romanb:
This PR is a small simplification to the handling of pending incoming connections. Instead of handing out a mutable borrow to the connection pool in the `IncomingConnectionEvent`, so one can call `IncomingConnectionEvent::accept()`, just provide `Network::accept()`. That this is a simplification can also easily be seen from the LOC diff.


twittner (approved):

mxinden (approved):
:+1: for this approach and for not handing out a mutable borrow to the connection pool.

Issue #1725, opened 2020-09-01
[gossipsub] `multi_hop_propagation` occasionally fails.


romanb:
I don't know when it started, but the `multi_hop_propagation` integration test now occasionally fails with a timeout:

```rust
thread 'multi_hop_propagation' panicked at '[quickcheck] TEST FAILED (runtime error). Arguments: (74, 40)
Error: "called `Result::unwrap()` on an `Err` value: TimeoutError { _private: () }"', /home/roman/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/src/libstd/macros.rs:13:23
```

When it happens, it always seems to be because one message is not received, that is, `received_msgs` remains equal to `number_nodes - 2` while all Swarms are `Pending`. I didn't dig deeper as I'm not familiar with the protocol and there seem to be larger changes upcoming in https://github.com/libp2p/rust-libp2p/pull/1720.

romanb:
I can no longer reproduce this but it should be reopened with new information if it happens again.

PR #1737, opened 2020-09-08
protocols/gossipsub/tests/smoke: Improve wait_for reporting

mxinden:
Small improvement for `protocols/gossipsub/tests/smoke.rs` providing more information on failure, hopefully helping us to track down the CI failures.

I am able to reproduce https://github.com/libp2p/rust-libp2p/issues/1725 occasionally but haven't found the cause (yet).

### DIFF
--- a/protocols/gossipsub/src/behaviour.rs
+++ b/protocols/gossipsub/src/behaviour.rs
@@ -1454,7 +1454,7 @@ impl fmt::Debug for GossipsubRpc {
 }
 
 /// Event that can happen on the gossipsub behaviour.
-#[derive(Debug)]
+#[derive(Clone, Debug)]
 pub enum GossipsubEvent {
     /// A message has been received. This contains the PeerId that we received the message from,
     /// the message id (used if the application layer needs to propagate the message) and the
@@ -1534,4 +1534,3 @@ impl fmt::Debug for PublishConfig {
         }
     }
 }
-
--- a/protocols/gossipsub/tests/smoke.rs
+++ b/protocols/gossipsub/tests/smoke.rs
@@ -103,32 +103,23 @@ impl Graph {
         }
     }
 
-    /// Polls the graph and passes each event into the provided FnMut until it returns `true`.
-    fn wait_for<F>(self, mut f: F) -> Self
-    where
-        F: FnMut(GossipsubEvent) -> bool,
-    {
-        // The future below should return self. Given that it is a FnMut and not a FnOnce, one needs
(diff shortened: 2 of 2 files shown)


romanb (approved):

AgeManning:
oh. I tracked down the cause of this but only recall the basic idea. 

So, quickcheck creates many nodes and connects them to previously established nodes. The problem comes that you can have too many nodes be attached to one established node such that the newly established node doesn't get in the mesh (because the mesh already has enough peers). So when the first publish happens, it doesn't propagate to the lone peers not in the mesh. Also I think there was an issue that we publish before all grafts have been handled so the mesh isn't properly constructed sometimes. 

In these cases, we need to wait for a few heartbeats and handle IHAVE/IWANT messages before the lone peers will receive the message. It can be the case that one peer is not in the mesh and then future peers attach it it, so it requires a number of heartbeats before it reaches the outer peers. 

I recall it being rather involved, so I added some small hacks to minimize the failure rate, i.e the `graph.drain_poll()` and lowered the number of nodes that can be created. 

One solution would be to keep track of the number of peers attached to each node and ensure that we only connect to previous peers that don't have `D_low` already attached (assuming I recall the issue correctly)

AgeManning (approved):

protocols/gossipsub/src/behaviour.rs
```diff
@@ -1454,7 +1454,7 @@ impl fmt::Debug for GossipsubRpc {
 }
 
 /// Event that can happen on the gossipsub behaviour.
-#[derive(Debug)]
+#[derive(Clone, Debug)]
```
mxinden, line 1457: I won't add a changelog entry for this. In case you think it is necessary, please speak up.

mxinden:
> So, quickcheck creates many nodes and connects them to previously established nodes. The problem comes that you can have too many nodes be attached to one established node such that the newly established node doesn't get in the mesh (because the mesh already has enough peers). So when the first publish happens, it doesn't propagate to the lone peers not in the mesh.

While that sounds like one of the bugs our graph construction might have, I don't think it is the only one. Graph construction is happening deterministically via a seed. I would expect a previous failure to resurface when run with the same seed. When a failure happened on my machine, rerunning with the same Quickcheck input would not surface the bug each time.

I will need to dig deeper into this. Thanks for all the details!

AgeManning:
Oh right. The ones I found where with large nodes and were reproducible. 

If they are not reproducible sounds like gossip related as that is non-deterministic.

PR #1740, opened 2020-09-09
Prepare release of `libp2p-0.27.0`.

romanb:
This is step 3 of the `libp2p-yamux` upgrade process.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,7 +23,7 @@
 - [`parity-multiaddr` CHANGELOG](misc/multiaddr/CHANGELOG.md)
 - [`libp2p-core-derive` CHANGELOG](misc/core-derive/CHANGELOG.md)
 
-# Version 0.27.0 [unreleased]
+# Version 0.27.0 [2020-09-09]
 
 - Update `libp2p-yamux` to `0.24.0`. *Step 3 of 4 in a multi-release
   upgrade process.* See the `libp2p-yamux` CHANGELOG for details.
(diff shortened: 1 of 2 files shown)


twittner (approved):

mxinden:
With `v0.26.0` and `v0.27.0` already being released today, do I understand the following implications for Polkadot correctly?

- The next 4 (`v0.25.0` - `v0.28.0`) need to be rolled out in sequence.

  > If current code uses the default Yamux configuration, all upgrades need to be performed in order because each
  version is only compatible with its immediate predecessor.

- We need to wait for each version to be widely deployed across all active nodes before moving on to the next version.

- Given that the versions are already being released now, we can not add any new upcoming breaking changes by any `libp2p-xxx` crate in between, thus locking ourselves to the current libp2p status quo for the next 4 Polkadot releases.

romanb:
Published to `crates.io`.

mxinden:
> > The next 4 (v0.25.0 - v0.28.0) need to be rolled out in sequence.
> 
> No, because substrate does not use the default configuration, it sets [lazy_open](https://github.com/paritytech/substrate/blob/13b0650c45c4994f7fde3614fb8d09f5719eaeb1/client/network/src/transport.rs#L79) and hence never sends initial window updates and neither does it change the default receive window. That makes it safe for substrate to go straight to `libp2p-0.28` without any issues (hopefully).

I missed this. Thanks for clarifying!

 
> The reason both the yamux and the libp2p releases are made all at once is because:
> 
>     1. It avoids the cognitive overhead of an upgrade process that drags on for months.
> 
>     2. It allows all downstream dependencies who do use the default yamux configuration to perform the upgrades at their own pace, instead of us prescribing a bound for how quickly that can be done by our release intervals.

:+1:

romanb:
@mxinden See also https://github.com/paritytech/substrate/issues/6966.

PR #1742, opened 2020-09-10
 core/Cargo.toml: Require parity-multiaddr 0.9.2 for Ord bound

mxinden:
With b601ed4 libp2p-core requires the `Ord` bound on `Multiaddr`. This
bound is only implemented with parity-multiaddr `0.9.2`. Make sure
libp2p-core requires at least parity-multiaddr `0.9.2`.

I am not sure it is worth the patch release of both libp2p-core and libp2p. Let me know. Happy to do the release.


romanb (approved):
:+1: Feel free to do the patch release with this PR.

Issue #1518, opened 2020-03-25
mdns service operates on only one interface


eskimor:
Hi there!

I realized that mdns registers 0.0.0.0 for multicasts in the kernel, which translates to pick the default interface .. which was the wrong one on my machine. This could be fixed by registering all available interfaces instead and by opening multiple sockets for each interface. Interfaces could be gathered by means of the [pnet crate](https://docs.rs/pnet/0.25.0/pnet/datalink/fn.interfaces.html).

Alternatively we could just provide a way for the user to specify an IP or even better IPs (in case you in fact want to work with multiple interfaces).

A friend of mine also pointed out that on today's systems it is also not unlikely that interfaces appear/vanish at runtime .. so ideally we would also account for this possibility.

PR #1750, opened 2020-09-12
WIP: Join multicast group on all interfaces

eskimor:
Fix issue #1518.

We now join the multicast group on all interfaces, instead of just the default interface. This definitely fixes the node discovery for me. Initially I mentioned that there is also a problem with the `query_socket`, but I don't think this is true (it does not make sense and everything seems to work with this change). Nevertheless I will do some more testing on Monday and provide an additional PR for that problem if need be.

### DIFF
--- a/protocols/mdns/Cargo.toml
+++ b/protocols/mdns/Cargo.toml
@@ -25,6 +25,7 @@ smallvec = "1.0"
 tokio = { version = "0.2", default-features = false, features = ["udp"], optional = true }
 void = "1.0"
 wasm-timer = "0.2.4"
+pnet = "0.26.0"
 
 [dev-dependencies]
 get_if_addrs = "0.5.3"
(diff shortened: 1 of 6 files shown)


eskimor:
Turns out younger me was right, query socket needs fixing as well.

mxinden:
Depending on all of `pnet` just to get the hosts seems worrisome. See this discussion https://github.com/libp2p/rust-libp2p/pull/1514#issuecomment-603817280.

Are there any smaller crates out there offering the same functionality?

eskimor:
Thanks @mxinden! I believe so, I am currently investigating [multicast-socket](https://github.com/bltavares/multicast-socket), which actually does exactly what we need and does not rely on pnet. Looks promising so far.

romanb:
I believe this is already done now as part of https://github.com/libp2p/rust-libp2p/pull/1830. Feel free to open new issues if the approach taken there seems inadequate or incorrect to you.

eskimor:
Looks good to me :-)

PR #1754, opened 2020-09-14
CA support

juniuszhou:
Libp2p is most used network stack for public blockchain project now. For consortium chain, a requirement is to support CA exchange and verification. Then organization can issue the CA certificate and node just allow connection from the peer with correct certificate.
In the solution, we extend the noise protocol and use the data field to exchange CA.
The PR just modify the handshake message and introduce two more parameters as anchor CA and node CA as der format.


twittner:
> In the solution, we extend the noise protocol and use the data field to exchange CA.

Why is it required to modify noise instead of implementing a separate protocol for this exchange?

tomaka:
To give some context, @juniuszhou and me chatted in private before and as he was opening this PR.
I agree that this should be its own protocol, and since this is a feature designed for Substrate I also believe that this should live in the Substrate repository.

If a similar feature is needed in rust-libp2p in the future, we should rather [support TLS 1.3](https://github.com/libp2p/specs/tree/master/tls) and give the option to specify root certificate authorities.

juniuszhou:
closed according to the comment. will use a new layer to implement CA exchange instead of putting it in noise handshake.

PR #1755, opened 2020-09-15
Update some dependencies to the latest

koushiro:

### DIFF
--- a/protocols/floodsub/src/layer.rs
+++ b/protocols/floodsub/src/layer.rs
@@ -206,7 +206,7 @@ impl Floodsub {
 
         let self_subscribed = self.subscribed_topics.iter().any(|t| message.topics.iter().any(|u| t == u));
         if self_subscribed {
-            self.received.add(&message);
+            self.received.add(&message).expect("Not enough space to store this item, rebucketing failed.");
             if self.config.subscribe_local_messages {
                 self.events.push_back(
                     NetworkBehaviourAction::GenerateEvent(FloodsubEvent::Message(message.clone())));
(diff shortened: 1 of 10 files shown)


koushiro:
Seemingly simple dependency updates can be handed over to `dependabot ` now.
Nice!

protocols/floodsub/src/layer.rs
```diff
@@ -206,7 +206,7 @@ impl Floodsub {
 
         let self_subscribed = self.subscribed_topics.iter().any(|t| message.topics.iter().any(|u| t == u));
         if self_subscribed {
-            self.received.add(&message);
+            self.received.add(&message).expect("Not enough space to store this item, rebucketing failed.");
```
mxinden, line 209: I am not sure panicking is the right thing to do here. We should probably handle this more gracefully. Do you think this happening is so unlikely?

> When the filter returns NotEnoughSpace, the element given is actually added to the filter, but some random other element gets removed. This could be improved by implementing a single-item eviction cache for that removed item.

mxinden (commented):
Instead of a single large pull request updating all dependencies at once, I would prefer updating each individually through the pull requests opened by dependabot. E.g. the the `cuckoofilter` related changes could just be proposed on https://github.com/libp2p/rust-libp2p/pull/1763.

What do people think?

@koushiro is there a particular dependency that you need updated right away?

protocols/floodsub/src/layer.rs
koushiro: Actually, the cuckoofilter v0.3.2 will panic directly in the `add` function when map is full.

https://github.com/axiomhq/rust-cuckoofilter/blob/v0.3.2/src/lib.rs#L158

Issue #1760, opened 2020-09-15
protocols/plaintext: Incompatible UVI encoding with Go implementation


mxinden:
The Golang `/plaintext/2.0.0` implementation uses UVI encoding for the handshake messages only, not for the actual goodput of the protocol.

https://github.com/libp2p/go-libp2p-core/blob/master/sec/insecure/insecure.go

The Rust `/plaintext/2.0.0` implementation uses UVI encoding for both the handshake messages as well as all further messages exchanged over the protocol.

https://github.com/libp2p/rust-libp2p/blob/master/protocols/plaintext/src/lib.rs#L174

The `/plaintext/2.0.0` specification is somewhat vague, saying all messages should be UVI encoded, but not whether that includes messages send after the handshake (goodput):

> Message Framing
>
> All messages sent over the wire are prefixed with the message length in bytes, encoded as an unsigned variable length integer as defined by the multiformats unsigned-varint spec.

https://github.com/libp2p/specs/tree/master/plaintext

PR #1765, opened 2020-09-16
protocols/plaintext: Do not UVI encode goodput 

mxinden:
Only prefix handshake messages with the message length in bytes as an
unsigned varint. Return a plain socket once handshaking succeeded.

Tested compatibility with https://github.com/mxinden/libp2p-perf.

Fixes https://github.com/libp2p/rust-libp2p/issues/1760

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,9 +23,10 @@
 - [`parity-multiaddr` CHANGELOG](misc/multiaddr/CHANGELOG.md)
 - [`libp2p-core-derive` CHANGELOG](misc/core-derive/CHANGELOG.md)
 
-# Version 0.28.2 [unreleased]
+# Version 0.29.0 [unreleased]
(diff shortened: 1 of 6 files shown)


romanb (approved):
:+1:  Hopefully the specs can be clarified. It is curious that this hasn't come up before. I was probably naively assuming compatibility was tested earlier in the context of https://github.com/libp2p/rust-libp2p/pull/1236, or maybe the other implementation(s) actually changed afterwards?

mxinden:
> Hopefully the specs can be clarified.

I have opened https://github.com/libp2p/specs/pull/295.

> or maybe the other implementation(s) actually changed afterwards?

As far as I can tell from the initial pull request https://github.com/libp2p/go-libp2p-core/pull/37 the Go implementation only ever UVI encoded the handshake messages .

Issue #1758, opened 2020-09-15
Ping protocol tests fail when mplex is used instead of yamux. 


artemii235:
#### Some background

Our team is using Ping protocol as part of bigger combined behaviour that derives `NetworkBehaviour`. We had few issues with Yamux so switched to Mplex. We've noticed ping timeouts while there's no actual reason for this to happen so I started digging and discovered that Ping doesn't seem to work properly when Mplex is used.

#### How to reproduce:

1. Rust toolchain used: `nightly-2020-02-01-x86_64-apple-darwin`
1. Use the latest master state.
1. Use Mplex instead of Yamux at https://github.com/libp2p/rust-libp2p/blob/master/protocols/ping/tests/ping.rs#L209
1. Run Ping protocol tests.

#### Expected behavior:

Tests pass

#### Actual behavior

`max_failures` fails and `ping_pong` runs indefinitely.

romanb:
> We had few issues with Yamux so switched to Mplex.

It would be great if you could report any issues you encountered at https://github.com/paritytech/yamux, if you think there is a problem in the Yamux implementation.

artemii235:
> It would be great if you could report any issues you encountered at https://github.com/paritytech/yamux

We did our tests a couple of months ago, things might have changed as yamux version was upgraded. I'm going to retest the yamux today and check whether problems we encountered are solved. If not I will create an issue in yamux repo.

romanb:
This should now be fixed in `master` and there will probably be new releases this week. Thanks a lot for the report.

Issue #1629, opened 2020-06-25
Random upgrade timeouts between peers


BlackYoup:
Hello everyone,

I'm having a weird issue in which, sometimes, a substream will timeout during its upgrade. I have little knowledge in rust futures and in this project's internals so I'm sorry if I misinterpret some stuff. You can find a commit with my debug logs [here](https://github.com/BlackYoup/rust-libp2p/commit/e20c6af7fa7f074409896cca3e65f3ec16ce3e5a).

## Context

I'm running on my local machine 3 services (let's call them node A / B / C), all of them using rust-libp2p. The swarm created uses a custom NetworkBehaviour which has 3 behaviours:
- Kademlia
- Identify
- A custom behaviour I made

The transport used is plaintext TCP using the mplex streams multiplexer. I have some other issues with yamux which made me use mplex.

My mplex configuration:
```
let mut mplex = MplexConfig::default();
mplex.max_buffer_len_behaviour(MaxBufferBehaviour::Block);
mplex.split_send_size(1024 * 1024);
mplex.max_substreams(100000); // Yes, this is insane, it's to workaround another issue I had and I didn't take the time to lower it yet
```

I'm using rust-libp2p 0.20.1 but had the same issue on 0.19.1. The issue may be related to timing. It might not occur in debug mode with enough logs to "slow" the services but release mode almost always triggers it. Also, I started seeing it after a refactor of my code which seems to have brought a performance boost as well.

The 3 nodes connect to each other on my local machine using 127.0.0.1:port . Node B then advertise 1/2k providers records into the DHT. Node C will then receive requests from another external service and will do providers lookups into the DHT. The timeout randomly appears during those lookups. From time to time I get something like `Request to PeerId("12D3KooWAZ1gXB3ZokKzEnkXvWDK6RWj38rNkkeVTsqyevL5AYqV") in query QueryId(67) failed with Upgrade(Timeout)`. The timeout is the default one, 10 seconds.

## Upgrade timeout

I started using RUST_LOG=trace to get all the logs I could and started tracing the issue. It seems that node C successfully creates the substream and starts sending data to node B. On node B, I can see the created substream:

<details><summary>Substream 68 logs</summary>

```
2020-06-25T14:46:00.419Z TRACE libp2p_mplex                                  > Received message: Open { substream_id: 68 }
2020-06-25T14:46:00.419Z TRACE libp2p_mplex                                  > Filter pass for Open { substream_id: 68 }
[...]

Issue #508, opened 2018-09-21
Add a crate that provides a test suit for an implementation of StreamMuxer


tomaka:
Add a crate that provides:

```rust
pub fn test_muxer<T: StreamMuxer>(muxer: T) { ... }
```

Of course the implementation would check whether the muxer behaves as expected.

Then `mplex`, `yamux` and future implementations could import that crate and run the function.

tomaka:
We should do #493 first

tomaka:
cc @dvdplm

PR #1769, opened 2020-09-22
[mplex] Refactoring with Patches

romanb:
## Overview

This is a rather substantial refactoring of `libp2p-mplex`, though the control-flow skeleton remains the same, as dictated by the `StreamMuxer` API. In the context of looking into https://github.com/libp2p/rust-libp2p/issues/1758, besides a bug in `libp2p-ping` that only occurs with mplex and for which I will open a separate PR, the following two problems in `libp2p-mplex` became apparent and are intended to be addressed here:

  * Avoid stalls caused by a read operation on one substream reading (and buffering) frames for another substream without
    notifying the corresponding task. While testing https://github.com/libp2p/rust-libp2p/issues/1758, even when the `ping_pong` tests ran through, there were occasional stalls in between. It turned out that, when a read operation for one substream buffers frames for another, the corresponding task that may be waiting on these buffered frames is not notified. This led to situations in the `ping_pong` test where both sides sent each other the outbound ping simultaneously, each on its own substream, and after the read operations for the inbound substreams returned `Pending`, the read operation for the outbound substreams waiting for the responses would read and buffer the inbound ping of the remote for the other substream and, not finding any frames for themselves yet, again returning `Pending` without waking up the task(s) interested in the newly buffered frame(s). Only the ping timeout triggered the polling again. To avoid needless polling of the same task, it seemed necessary to me that read-interest for substreams is tracked for the particular substream IDs, instead of "globally".

  * Remove dropped substreams from the tracked set of open substreams, to avoid artificially running into substream limits.

> This PR alone is necessary but not sufficient to r-e-s-o-l-v-e https://github.com/libp2p/rust-libp2p/issues/1758, since there is also a small bug in `libp2p-ping` when used in combination with `libp2p-mplex`. This bug will receive a separate, follow-up PR.

## Related Context

Both of the above problems probably relate to https://github.com/libp2p/rust-libp2p/issues/1629 and https://github.com/libp2p/rust-libp2p/issues/1504 which report mplex symptoms that seem highly related (hitting unexpected substream limits and intermittent stalls). Though the latter is closed because panics were fixed, the problem with unexpectedly hitting the substream limit remained unsolved there, I think.

## Other Changes

While I was at it, I also went ahead with https://github.com/libp2p/rust-libp2p/issues/313 and thus another change here is:

  * Schedule sending of a `Reset` frame when an open substream gets dropped.

Furthermore, the semantics of the `max_substreams` configuration changed as follows:

  * Old behaviour: The connection is immediately closed with an error, regardless of whether the limit is reached by an attempt to open an outbound substream or by a new inbound substream.
  * New behaviour: Outbound substream attempts beyond the configured limit are delayed (`Poll::Pending`) with a wakeup once an existing substream closes, i.e. the limit results in back-pressure for new outbound substreams. New inbound substreams beyond the limit are immediately answered with a `Reset`, again a form of back-pressure. If too many (by some internal threshold) pending `Reset` frames accumulate, e.g. as a result of an aggressive number of inbound substreams being opened beyond the configured limit, the connection is closed ("DoS protection").

## Testing

While the multiplexers themselves still need more testing, also compatibility testing (https://github.com/libp2p/rust-libp2p/issues/508), and ideally comparative performance testing, I have so far done the following (besides passing the libp2p test suite where mplex is often used, of course):

  * A follow-up PR for `libp2p-ping` that fixes a small bug and depends on this PR will randomise the choice of multiplexer used for integration tests, expecting the same behaviour. This kind of testing essentially revealed all the issues mentioned here as a result of looking into https://github.com/libp2p/rust-libp2p/issues/1758.
  * A subtrate node seems to start and sync fine, configured only with mplex and using this branch.

### DIFF
--- a/muxers/mplex/src/io.rs
+++ b/muxers/mplex/src/io.rs
@@ -1,0 +1,687 @@
+// Copyright 2018 Parity Technologies (UK) Ltd.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a
+// copy of this software and associated documentation files (the "Software"),
+// to deal in the Software without restriction, including without limitation
+// the rights to use, copy, modify, merge, publish, distribute, sublicense,
+// and/or sell copies of the Software, and to permit persons to whom the
+// Software is furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS
+// OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+// FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+// DEALINGS IN THE SOFTWARE.
+
+use bytes::Bytes;
+use crate::{MplexConfig, MaxBufferBehaviour};
+use crate::codec::{Codec, Frame, LocalStreamId, RemoteStreamId};
+use log::{debug, trace};
+use fnv::FnvHashMap;
+use futures::{prelude::*, ready, stream::Fuse};
+use futures::task::{ArcWake, waker_ref, WakerRef};
+use futures_codec::Framed;
+use parking_lot::Mutex;
+use std::collections::hash_map::Entry;
+use std::{cmp, io, mem, sync::Arc, task::{Context, Poll, Waker}};
+
+pub use std::io::{Result, Error, ErrorKind};
+
+/// A multiplexed I/O stream.
+pub struct Multiplexed<C> {
+    /// The current operating status.
+    status: Status,
+    /// The underlying I/O stream.
+    io: Fuse<Framed<C, Codec>>,
+    /// The configuration.
+    config: MplexConfig,
+    /// Buffer of received frames that have not yet been consumed.
+    buffer: Vec<Frame<RemoteStreamId>>,
+    /// Whether a flush is pending before reading frames can proceed.
+    pending_flush: bool,
+    /// Streams for which a `Reset` frame should be sent.
+    pending_reset: Vec<LocalStreamId>,
+    /// The substreams that are considered at least half-open.
+    open_substreams: FnvHashMap<LocalStreamId, SubstreamState>,
+    /// The ID for the next outbound substream.
+    next_outbound_stream_id: LocalStreamId,
+    /// Registry of wakers for pending tasks interested in reading.
+    notifier_read: Arc<NotifierRead>,
+    /// Registry of wakers for pending tasks interested in writing.
+    notifier_write: Arc<NotifierWrite>,
+    /// Registry of wakers for pending tasks interested in opening
+    /// an outbound substream, when the configured limit is reached.
+    notifier_open: Arc<NotifierOpen>,
+}
+
+/// The operation status of a `Multiplexed` I/O stream.
+#[derive(Debug)]
+enum Status {
+    /// The stream is considered open and healthy.
+    Ok,
+    /// The stream has been actively closed.
+    Closed,
+    /// The stream has encountered a fatal error.
+    Err(io::Error),
+}
+
+impl<C> Multiplexed<C>
+where
+    C: AsyncRead + AsyncWrite + Unpin
+{
+    /// Creates a new multiplexed I/O stream.
+    pub fn new(io: C, config: MplexConfig) -> Self {
+        let max_buffer_len = config.max_buffer_len;
+        Multiplexed {
+            config,
+            status: Status::Ok,
+            io: Framed::new(io, Codec::new()).fuse(),
+            buffer: Vec::with_capacity(cmp::min(max_buffer_len, 512)),
+            open_substreams: Default::default(),
+            pending_flush: false,
+            pending_reset: Vec::new(),
+            next_outbound_stream_id: LocalStreamId::dialer(0),
+            notifier_read: Arc::new(NotifierRead {
+                pending: Mutex::new(Default::default()),
+            }),
+            notifier_write: Arc::new(NotifierWrite {
+                pending: Mutex::new(Default::default()),
+            }),
+            notifier_open: Arc::new(NotifierOpen {
+                pending: Mutex::new(Default::default())
+            })
+        }
+    }
+
+    /// Flushes the underlying I/O stream.
+    pub fn poll_flush(&mut self, cx: &mut Context<'_>) -> Poll<io::Result<()>> {
+        match &self.status {
+            Status::Closed => return Poll::Ready(Ok(())),
+            Status::Err(e) => return Poll::Ready(Err(io::Error::new(e.kind(), e.to_string()))),
+            Status::Ok => {}
+        }
+
+        // Send any pending reset frames.
+        ready!(self.send_pending_reset(cx))?;
+
+        // Flush the underlying I/O stream.
+        let waker = NotifierWrite::register(&self.notifier_write, cx.waker());
+        match ready!(self.io.poll_flush_unpin(&mut Context::from_waker(&waker))) {
+            Err(e) => self.on_error(e),
+            Ok(()) => {
+                self.pending_flush = false;
+                Poll::Ready(Ok(()))
+            }
+        }
+    }
+
+    /// Closes the underlying I/O stream.
+    ///
+    /// > **Note**: No `Close` or `Reset` frames are sent on open substreams
+    /// > before closing the underlying connection. However, the connection
+    /// > close implies a flush of any frames already sent.
+    pub fn poll_close(&mut self, cx: &mut Context<'_>) -> Poll<io::Result<()>> {
+        match &self.status {
+            Status::Closed => return Poll::Ready(Ok(())),
+            Status::Err(e) => return Poll::Ready(Err(io::Error::new(e.kind(), e.to_string()))),
+            Status::Ok => {}
+        }
+
+        // Note: We do not make the effort to send pending `Reset` frames
+        // here, we only close (and thus flush) the underlying I/O stream.
+
+        let waker = NotifierWrite::register(&self.notifier_write, cx.waker());
+        match self.io.poll_close_unpin(&mut Context::from_waker(&waker)) {
+            Poll::Pending => Poll::Pending,
+            Poll::Ready(Err(e)) => self.on_error(e),
+            Poll::Ready(Ok(())) => {
+                self.pending_reset = Vec::new();
+                // We do not support read-after-close on the underlying
+                // I/O stream, hence clearing the buffer and substreams.
+                self.buffer = Vec::new();
+                self.open_substreams = Default::default();
+                self.status = Status::Closed;
+                Poll::Ready(Ok(()))
+            }
+        }
+    }
+
+    /// Waits for a new inbound substream, returning the corresponding `LocalStreamId`.
+    pub fn poll_next_stream(&mut self, cx: &mut Context<'_>) -> Poll<io::Result<LocalStreamId>> {
+        self.guard_ok()?;
+
+        // Wait for the next inbound `Open` frame.
+        let stream_id = ready!(self.poll_recv_frame(cx, None, |frame| match frame {
+            Frame::Open { stream_id } => Some(stream_id.into_local()),
+            _ => None,
+        }))?.expect("Unexpected end of substream.");
+
+        debug!("New inbound substream: {:?}", stream_id);
+
+        Poll::Ready(Ok(stream_id))
+    }
+
+    /// Creates a new (outbound) substream, returning the allocated stream ID.
+    pub fn poll_open_stream(&mut self, cx: &mut Context<'_>) -> Poll<io::Result<LocalStreamId>> {
+        self.guard_ok()?;
+
+        // Check the stream limits.
+        if self.open_substreams.len() >= self.config.max_substreams {
+            debug!("Maximum number of substreams reached: {}", self.config.max_substreams);
+            let _ = NotifierOpen::register(&self.notifier_open, cx.waker());
+            return Poll::Pending
+        }
+
+        // Send the `Open` frame.
+        let waker = NotifierWrite::register(&self.notifier_write, cx.waker());
+        match ready!(self.io.poll_ready_unpin(&mut Context::from_waker(&waker))) {
+            Ok(()) => {
+                let stream_id = self.next_outbound_stream_id();
+                let frame = Frame::Open { stream_id };
+                match self.io.start_send_unpin(frame) {
+                    Ok(()) => {
+                        self.open_substreams.insert(stream_id, SubstreamState::Open);
+                        // The flush is delayed and the `Open` frame may be sent
+                        // together with other frames in the same transport packet.
+                        self.pending_flush = true;
+                        Poll::Ready(Ok(stream_id))
+                    }
+                    Err(e) => self.on_error(e),
+                }
+            },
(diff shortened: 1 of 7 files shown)


muxers/mplex/src/io.rs
```diff
@@ -0,0 +1,733 @@
...
+    fn can_read(&self, id: &LocalStreamId) -> bool {
+        match self.open_substreams.get(id) {
+            Some(SubstreamState::Open) | Some(SubstreamState::SendClosed) => true,
+            _ => false,
+        }
+    }
+
+    /// Sends pending frames, without flushing.
+    fn send_pending_frames(&mut self, cx: &mut Context<'_>) -> Poll<io::Result<()>> {
+        while let Some(frame) = self.pending_frames.pop() {
```
tomaka, line 579: I don't know if it was already like this before, but I find it strange that `pending_frames` is a FILO queue here. I would expect entries to be sent in the order they have been pushed.

muxers/mplex/src/io.rs
```diff
@@ -0,0 +1,733 @@
...
+            if self.poll_send_frame(cx, || frame.clone())?.is_pending() {
```
tomaka, line 580: Why not call `self.pending_frames.pop()` within the closure? As far as I understand, if the closure is called, the outcome cannot be `Pending`.

muxers/mplex/src/io.rs
```diff
@@ -0,0 +1,733 @@
...
+    SendClosed,
+    /// A `Close` frame has been received but the stream is still
+    /// open for writing (remote half-close).
+    RecvClosed
+}
+
+struct NotifierRead {
+    /// List of wakers to wake when read operations can proceed
+    /// on a substream (or in general, for the key `None`).
+    pending: Mutex<FnvHashMap<Option<LocalStreamId>, Waker>>,
```
tomaka, line 626: If I understand correctly, wakers whose key (the `Option<LocalStreamId>`) is `Some` come from calling `poll_read_stream`, and wakers whose key is `None` come from calling `poll_next_stream`.

If that's accurate, then it is correct to have a `Waker` here rather than a `Vec<Waker>`.

I find it quite hard to follow the logic, and it doesn't seem very fool-proof, but I also don't know how to make it clearer.

tomaka (changes requested):

muxers/mplex/src/io.rs
romanb: Yes, that is certainly better.

> I don't know if it was already like this before ...

There were no `pending_frames` before, since they were introduced primarily to send close/reset frames for dropped substreams as necessary, i.e. #313.

muxers/mplex/src/io.rs
romanb: I don't think that's permissible due to borrowing. Even if `poll_send_frame` were made a standalone function and thus not borrow all of `self`, it still calls `on_error` in case of an I/O error which also wants a mutable borrow for `pending_frames` to drop/clear it on error. I agree that the pop/push as well as the frame cloning is suboptimal, but I didn't consider it that important or inefficient. The frames sent here are only close/reset frames containing only a stream ID. We could of course make `poll_send_frame` return the frame in the `Pending` case with a custom return enum.

muxers/mplex/src/io.rs
romanb: > If that's accurate, then it is correct to have a Waker here rather than a Vec<Waker>.

Yes.

> I find it quite hard to follow the logic, and it doesn't seem very fool-proof, but I also don't know how to make it clearer.

Not sure what logic you mean, I guess in general the task wakeup logic for reading? Yes, that is, as is mostly the case, subtle. There are these two different "threads of control" for reading by design: 1) Awaiting the next inbound stream and b) Awaiting new data on a specific stream. Thereby 1) and each instance of 2) potentially read and buffer frames for other streams (or for new inbound streams) that need to result in wakeup attempts.

muxers/mplex/src/io.rs
```diff
@@ -0,0 +1,733 @@
...
+
+            trace!("Received: {:?}", frame);
+
+            // Handle Open/Close/Reset frames.
+            match frame {
+                Frame::Open { stream_id } => {
+                    let id = stream_id.into_local();
+
+                    if self.open_substreams.contains_key(&id) {
+                        debug!("Ignoring received `Open` Frame for open substream: {:?}", id);
```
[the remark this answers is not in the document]
romanb: I didn't think too much about it, this was [existing behaviour](https://github.com/libp2p/rust-libp2p/blob/master/muxers/mplex/src/lib.rs#L304). In general I do have the tendency to be lenient with harmless mistakes made by a remote. Would you prefer to reset the stream or even fail the entire multiplexed stream with an error?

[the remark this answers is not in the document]
mxinden: > Would you prefer to reset the stream or even fail the entire multiplexed stream with an error?

I think failing the entire multiplexed stream with an error will help us surface bugs in the future.

tomaka (approved):

muxers/mplex/src/io.rs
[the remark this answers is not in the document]
romanb: Done [here](https://github.com/libp2p/rust-libp2p/pull/1769/commits/dbe8337d01400389cf53ab59ab4a1b546218cf4e#diff-7dd5a5fe0de322516c6dd03671be9181R499-R501) and [here](https://github.com/libp2p/rust-libp2p/pull/1769/commits/dbe8337d01400389cf53ab59ab4a1b546218cf4e#diff-7dd5a5fe0de322516c6dd03671be9181R522-R524). Let me know if that looks ok to you.

romanb:
@mxinden I assume you're still taking your time to review. There is no rush from my side, I appreciate the time taken for the reviews.

muxers/mplex/src/io.rs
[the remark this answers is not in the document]
mxinden: Looks good, thank you for the change.

muxers/mplex/src/io.rs
```diff
@@ -0,0 +1,771 @@
...
+    ///
+    /// If `Some` stream ID is given, `Ok(None)` signals EOF for
+    /// that substream. If `None` is given for the stream ID,
+    /// `Ok(None)` is never returned. If the underlying connection
+    /// encounters EOF, a corresponding error is returned.
+    fn poll_read_frame<F, O>(
+        &mut self,
+        cx: &mut Context<'_>,
+        stream_id: Option<LocalStreamId>,
+        mut select: F,
```
mxinden, line 403: If I understand correctly `poll_read_frame` is only called by `poll_next_stream` and `poll_read_stream` with the former passing `None` as `stream_id` and the latter passing `Some`. Both seem to pass static closures via the `select` argument to `poll_read_frame`.

Accepting a function in `poll_read_frame / select` allows very diverse use-cases for the method, but as far as I can tell there are only the two mentioned above. Is this flexibility worth the complexity? Can `poll_read_frame` not imply the intention of its caller through `stream_id` only?

muxers/mplex/src/io.rs
```diff
@@ -0,0 +1,771 @@
...
+    /// The current operating status.
+    status: Status,
+    /// The underlying I/O stream.
+    io: Fuse<Framed<C, Codec>>,
+    /// The configuration.
+    config: MplexConfig,
+    /// Buffer of received frames that have not yet been consumed.
+    buffer: Vec<Frame<RemoteStreamId>>,
+    /// Whether a flush is pending before reading frames can proceed.
+    pending_flush: bool,
```
mxinden, line 46: Do I understand that this is only `true` when an `Open` frame is still to be flushed? If so, how about renaming it to `pending_open_frame_flush`. I was a bit surprised `poll_read_frame` yields to flushes in general, even though it only does for `Open` frames, which makes sense.

muxers/mplex/src/io.rs
```diff
@@ -0,0 +1,771 @@
...
+
+        let waker = NotifierWrite::register(&self.notifier_write, cx.waker());
+        match self.io.poll_close_unpin(&mut Context::from_waker(&waker)) {
+            Poll::Pending => Poll::Pending,
+            Poll::Ready(Err(e)) => Poll::Ready(self.on_error(e)),
+            Poll::Ready(Ok(())) => {
+                self.pending_frames = VecDeque::new();
+                // We do not support read-after-close on the underlying
+                // I/O stream, hence clearing the buffer and substreams.
+                self.buffer = Vec::new();
```
mxinden, line 152:
```suggestion
                self.buffer = Default::default();
```

Nit pick: To be consistent with `on_error`.

muxers/mplex/src/io.rs
```diff
@@ -0,0 +1,771 @@
...
+    /// Registry of wakers for pending tasks interested in opening
+    /// an outbound substream, when the configured limit is reached.
+    notifier_open: Arc<NotifierOpen>,
+}
+
+/// The operation status of a `Multiplexed` I/O stream.
+#[derive(Debug)]
+enum Status {
+    /// The stream is considered open and healthy.
+    Ok,
```
mxinden, line 72: Wouldn't `Open` be more descriptive, given that there is a `Closed`? Not a strong opinion.

mxinden (approved):
I am fine merging this pull request as is. I have a couple of questions suggestions, but none of them are important. For what my opinion is worth, I find this to be a well-conceived implementation.

muxers/mplex/src/io.rs
romanb: > Can poll_read_frame not imply the intention of its caller through stream_id only?

Not quite, because the `select` function also maps the `Frame` to some other type, in the same manner as a function passed to `filter_map`. However, I made an attempt at restructuring the read control flow for more clarity, avoiding the function parameter, making `poll_read_frame` return a `Frame` and making the control flow of the individual read operations clearer, albeit at a bit of repeating a bit of of the same control flow in two places. It does seem preferable to me though. See https://github.com/libp2p/rust-libp2p/pull/1769/commits/7bd149888467429897b2b2ff6596e12cb803d025. What do you think?

muxers/mplex/src/io.rs
romanb: > Do I understand that this is only true when an Open frame is still to be flushed?

Yes, when at least one new outbound stream has been opened (i.e. an `Open` frame was sent) but not yet flushed before trying to read. Would you be ok with the name `pending_flush_open`?

romanb:
> thread 'async-std/runtime' panicked at 'Peer1: Unexpected event: Event(OutboundFailure { peer: PeerId("12D3KooWPC56uLBNMVu3fLA5jDxA32uxpHUnUpZMEqLAQpjtRFZC"), request_id: RequestId(205), error: ConnectionClosed })', protocols/request-response/tests/ping.rs:164:22
test ping_protocol ... FAILED

These tests don't even use mplex, so this sudden error seems unrelated, but I can look into it separately, if it is somewhat reproducible.

muxers/mplex/src/io.rs
romanb: Done in https://github.com/libp2p/rust-libp2p/pull/1769/commits/adca14ffba41da1cf77b692c59790694bcef12e6.

muxers/mplex/src/io.rs
mxinden: `pending_flush_open` rolls off the tongue much better, thanks!

romanb: Done in https://github.com/libp2p/rust-libp2p/pull/1769/commits/37c464c7a950ccbc9e7aba5380c02887af41e79a.

muxers/mplex/src/io.rs
```diff
@@ -0,0 +1,793 @@
...
+impl ArcWake for NotifierWrite {
+    fn wake_by_ref(this: &Arc<Self>) {
+        let wakers = mem::replace(&mut *this.pending.lock(), Default::default());
+        for waker in wakers {
+            waker.wake();
+        }
+    }
+}
+
+impl NotifierRead {
```
mxinden, line 715: Any particular reason why the `NotifierRead` `impl`s are separated from the corresponding `struct ` definition?

muxers/mplex/src/io.rs
mxinden: I missed the `map`ing functionality, indeed. Thanks for clarifying.

I am in favor of https://github.com/libp2p/rust-libp2p/commit/7bd149888467429897b2b2ff6596e12cb803d025, thanks for the change.

muxers/mplex/src/io.rs
```diff
@@ -0,0 +1,793 @@
...
+/// A multiplexed I/O stream.
+pub struct Multiplexed<C> {
+    /// The current operating status.
+    status: Status,
+    /// The underlying I/O stream.
+    io: Fuse<Framed<C, Codec>>,
+    /// The configuration.
+    config: MplexConfig,
+    /// Buffer of received frames that have not yet been consumed.
+    buffer: Vec<Frame<RemoteStreamId>>,
```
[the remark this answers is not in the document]
romanb: In principle, yes, I'd prefer to have a read buffer per stream, i.e. a map. Not just because of potentially suboptimal performance if just a single stream is slow to consume its frames, but also because it would allow just `Reset`ing a single stream if the buffer limit for that stream is hit, instead of failing the entire connection when the shared `buffer` is full with the default configuration `MaxBufferBehaviour::CloseAll`. Essentially changing `MaxBufferBehaviour::CloseAll` to `MaxBufferBehaviour::ResetStream`, as this is more along the lines of what the mplex spec prescibes in the [implementation notes](https://github.com/libp2p/specs/tree/master/mplex#implementation-notes). The configured buffer limit would then apply to a single substream and together with the substream limit lets one still reason about resource usage bounds. But in any case, the shared buffer was pre-existing code and I'd really prefer to propose such a change in a separate PR. I didn't want to pile up too many semantical changes here.

muxers/mplex/src/io.rs
romanb: What do you mean by "separated"? That it is not directly below the `struct` definition? No reason for that, other than that I don't particularly mind about such ordering of definitions. I'm happy to change it.

muxers/mplex/src/io.rs
romanb: > Can one of you expand on why it is safe to only cache a single waker for Option<LocalStreamId>::None?

Because the key `None` is only used by `poll_next_stream`, which is part of the implementation for `StreamMuxer::poll_event` whose API contract explicitly states that "Only the latest task that was used to call this method may be notified.". The API contract for `StreamMuxer::read_substream`, `StreamMuxer::write_substream` etc. is similar. In general, and as far as I know, the reason for `StreamMuxer`s to be `Sync` is to permit using different substreams from different threads, not to use the same substream from different threads. Only waking the last task that `poll`s is generally common for most such APIs, e.g. [Future::poll](https://docs.rs/futures/0.3.5/futures/future/trait.Future.html#tymethod.poll) has an analogous contract. In practice, I would think the reason for this contract is a matter of practicality due ease and efficiency of implementation and the relatively rare usefulness for having to "remember" all tasks to wake, instead of just the last.

mxinden: Thanks for the in-depth explanation.

muxers/mplex/src/io.rs
mxinden: I am just a bit confused by the fact that all of `NotifierWrite` is interleaved. I am fine either way.

mxinden (approved):
Still looks good to me. Thanks for all the comments.

muxers/mplex/src/io.rs
romanb: I changed the ordering in https://github.com/libp2p/rust-libp2p/pull/1769/commits/77ea68b4f24ac77b0392d93a42c8cf4eb66d54ac.

muxers/mplex/src/io.rs
tomaka: > I would think the reason for this contract is a matter of practicality due to ease and efficiency of implementation and the relatively rare usefulness for having to "remember" all tasks to wake, instead of just the last.

One big problem with remember all tasks to wake up is that your container of wakers might grow a lot.

For instance, imagine that it takes 60 seconds before data is received on a substream and that, due to the nature of polling, we poll a substream 10 times per second. This results in 600 calls to `poll_read_substream` before the wakers actually get awakened. This means that we would insert 600 elements in the list of wakers.

The [`will_wake`](https://doc.rust-lang.org/nightly/std/task/struct.Waker.html#method.will_wake) method allows reducing the amount of duplicates, and also we're not really supposed to poll a substream 10 times per second. But both these mechanisms are unreliable.

Enforcing one waker per "thing to wake" guarantees a bound to the number of wakers, which is not a bad idea.

PR #1770, opened 2020-09-22
[ping] Add missing flush after write.

### LINKED ISSUE #1758: Ping protocol tests fail when mplex is used instead of yamux.

romanb:
After sending the ping and before awaiting the pong, the stream must be flushed to guarantee the data is sent before awaiting the reply. This is not necessary with `libp2p-yamux` where every sent frame is flushed, but with mplex it may remain sitting in a buffer otherwise. This problem was revealed by https://github.com/libp2p/rust-libp2p/issues/1758 together with some other mplex issues addressed in https://github.com/libp2p/rust-libp2p/pull/1769. This PR is on top of #1769 since the tests randomise usage of mplex and yamux and only pass consistently with these patches. For a proper diff (it is small), see [here](https://github.com/romanb/rust-libp2p/compare/patch-mplex...romanb:ping-mplex).

Together with #1769, closes #1758.

### DIFF
--- a/protocols/ping/CHANGELOG.md
+++ b/protocols/ping/CHANGELOG.md
@@ -1,3 +1,10 @@
+# 0.22.1 [unreleased]
+
+- Ensure the outbound ping is flushed before awaiting
+  the response. Otherwise the behaviour depends on
+  implementation details of the stream muxer used.
+  The current behaviour resulted in stalls with Mplex.
+
 # 0.22.0 [2020-09-09]
 
 - Update `libp2p-swarm` and `libp2p-core`.
--- a/protocols/ping/Cargo.toml
+++ b/protocols/ping/Cargo.toml
@@ -2,7 +2,7 @@
 name = "libp2p-ping"
 edition = "2018"
 description = "Ping protocol for libp2p"
-version = "0.22.0"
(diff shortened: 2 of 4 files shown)


mxinden (approved):
https://github.com/libp2p/rust-libp2p/pull/1770/commits/18711c5f6a9690d357e3e5873b31035383925c95 looks good to me. I will look into https://github.com/libp2p/rust-libp2p/pull/1769.

PR #1775, opened 2020-09-29
swarm/src/protocols_handler: Use FuturesUnordered in NodeHandlerWrapper

mxinden:
> Futures managed by FuturesUnordered will only be polled when they
generate wake-up notifications. This reduces the required amount of work
needed to poll large numbers of futures.

https://docs.rs/futures/0.3.5/futures/stream/struct.FuturesUnordered.html

Instead of iterating each inbound and outbound upgrade looking for one
to make progress, use a `FuturesUnordered` for both pending inbound and
pending outbound upgrades. As a result only those upgrades are polled
that are ready to progress.

### DIFF
--- a/swarm/Cargo.toml
+++ b/swarm/Cargo.toml
@@ -2,7 +2,7 @@
 name = "libp2p-swarm"
 edition = "2018"
 description = "The libp2p swarm"
-version = "0.23.0"
+version = "0.22.1"
 authors = ["Parity Technologies <admin@parity.io>"]
 license = "MIT"
 repository = "https://github.com/libp2p/rust-libp2p"
--- a/swarm/Cargo.toml
+++ b/swarm/Cargo.toml
@@ -2,7 +2,7 @@
 name = "libp2p-swarm"
 edition = "2018"
 description = "The libp2p swarm"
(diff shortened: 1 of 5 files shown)


romanb (approved):
Nice improvement :+1:

tomaka (approved):
While the change in isolation is good, it might reveal buggy `poll` implementations in `ProtocolsHandler`s that wouldn't wake up the task when they should.
For Substrate in particular, we should thoroughly test it when updating libp2p.

mxinden:
> While the change in isolation is good, it might reveal buggy poll implementations in ProtocolsHandlers that wouldn't wake up the task when they should.

Good point @tomaka . I will add a note in the changelog so we don't forget.

swarm/Cargo.toml
```diff
@@ -2,7 +2,7 @@
 name = "libp2p-swarm"
 edition = "2018"
 description = "The libp2p swarm"
-version = "0.22.0"
+version = "0.22.1"
```
mxinden, line 5: I am declaring this as a non-breaking change, thus bumping the patch version only. Let me know in case you prefer a minor version bump.

PR #1779, opened 2020-09-30
Fix #1778: Double mdns recv_buffer to 4096 bytes

### LINKED ISSUE #1778: mdns recv_buffer too small for some configurations
My PC has a couple of interfaces, with something like 16 IP addresses in total. This results in mdns responses of a size of approximately 3000 bytes. The problem is, [recv_buffer](https://github.com/libp2p/rust-libp2p/blob/aaa6e4add30c6a905c8a00ec85ef6912d1a89b49/protocols/mdns/src/service.rs#L125) is only 2048 bytes, which obviously causes parsing of responses to fail with an EOF error.

eskimor:
Also add warn if parsing of a packet fails, so problems with for example packet
sizes can more easily be detected.

Reasoning:

The previous 2048 bytes is already larger than the recommend size by
[rfc6762](https://tools.ietf.org/html/rfc6762#page-46), which recommends
a packet size that fits within the interface MTU.

Nevertheless the rfc also states, that if more data needs to be
transmitted one should rely on IP fragments, which would only pose
problems when interfacing with devices not understanding IP fragments.
The standard also imposes a strict upper limit of 9000 bytes.

Therefore raising `recv_buffer` to 4096 seems sensible, given that we
already exceed the recommend size and given that we are not restricting
the size when sending in any way. Meaning the implementation already
sends packets larger than recommended, just can't handle them on
reception.

### DIFF
--- a/protocols/mdns/src/service.rs
+++ b/protocols/mdns/src/service.rs
@@ -23,6 +23,7 @@ use dns_parser::{Packet, RData};
 use either::Either::{Left, Right};
 use futures::{future, prelude::*};
 use libp2p_core::{multiaddr::{Multiaddr, Protocol}, PeerId};
+use log::warn;
 use std::{convert::TryFrom as _, fmt, io, net::Ipv4Addr, net::SocketAddr, str, time::{Duration, Instant}};
 use wasm_timer::Interval;
 use lazy_static::lazy_static;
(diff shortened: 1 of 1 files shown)


eskimor:
In addition to this PR we might consider whether it is sensible in the first place to send the IP addresses of all interfaces in the response. If I am not missing something, there is very little value in sending IPs of other interfaces.

romanb (approved):
Looks reasonable to me.

PR #1780, opened 2020-09-30
Add another chat example (but it uses tokio instead of async_std)

zyansheep:
Basically, It took me a long time to figure out how to get libp2p swarm polling to work with tokio so I wanted to add an official example to help out any newcomers.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -123,3 +123,7 @@ members = [
     "transports/websocket",
     "transports/wasm-ext"
 ]
+
+[[example]]
+name = "chat-tokio"
+required-features = ["tcp-tokio", "mdns-tokio"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -95,7 +95,7 @@ libp2p-websocket = { version = "0.23.1", path = "transports/websocket", optional
 [dev-dependencies]
 async-std = "1.6.2"
 env_logger = "0.7.1"
-tokio = { version = "0.2", features = ["io-util", "io-std", "stream"] }
+tokio = { version = "0.2", features = ["io-util", "io-std", "stream", "macros"] }
 
 [workspace]
 members = [
--- a/examples/chat-tokio.rs
+++ b/examples/chat-tokio.rs
@@ -118,10 +118,11 @@ async fn main() -> Result<(), Box<dyn Error>> {
         };
 
         behaviour.floodsub.subscribe(floodsub_topic.clone());
+
+        // Create Swarm and Specify Tokio executor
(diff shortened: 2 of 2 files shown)


romanb:
I may be mistaken, but I think to actually use tokio exclusively in that example, you need to enable the `tcp-tokio` feature for that example, e.g. something like
```
[[example]]
name = "chat-tokio"
features = ["tcp-tokio"]
```
and configure the `Swarm` with a tokio-based executor, i.e.
```rust
    SwarmBuilder::new(transport, behaviour, local_peer_id)
        .executor(Box::new(|fut| { tokio::spawn(fut); }))
        .build()
```
otherwise `build_development_transport` will give you an `async-std`-based TCP transport and the connection background tasks will be spawned onto a [futures::executor::ThreadPool](https://docs.rs/futures/0.3.5/futures/executor/struct.ThreadPool.html).

zyansheep:
Huh, didn't think of that, I will add that.

romanb:
I still think we need to enable the `tcp-tokio` feature for that example, as I already mentioned, otherwise `build_development_transport()` [creates an async-std-based TCP transport](https://github.com/libp2p/rust-libp2p/blob/master/src/lib.rs#L289-L292). Alternatively the tokio transport could be constructed directly in the example. It would also be great if the example clearly documented at the top what changes it contains compared to the regular chat example in order to use tokio everywhere. That would make a nice reference.

zyansheep:
I opened up a separate project and enabled tcp-tokio, however, adding or removing the feature didn't recompile anything or remove async-std from the final compiled binary. I played around with it for awhile and eventually found out that doing something like this worked:
```toml
[dependencies.libp2p]
default-features = false
version = "0.28.1"
features = [ "tcp-tokio", "mdns-tokio", "floodsub", "identify", "yamux", "mplex", "noise", "websocket", "dns" ]
```
As well as changing the program to use `MdnsTokio` instead of `Mdns`.
I added a Cargo.toml file specifying the feature changes to the example directory, however it still compiled async-std and doesn't compile tcp-tokio or mdns-tokio.
I don't think specifying features for specific `[[example]]`s works.

romanb:
I see. I made a proposal for what we can do for the example in https://github.com/zyansheep/rust-libp2p/pull/1. If you are willing to merge that PR, I will follow up with merging this one.

zyansheep:
Ok, I have added it. Once [https://github.com/rust-lang/cargo/issues/1197](this feature) is merged into stable I might reopen this issue.

romanb (approved):
Thanks!

PR #1781, opened 2020-09-30
[multistream-select] Require remaining negotiation data to be flushed.

romanb:
There appears to still be an edge-case in `multistream-select` whereby the `remaining` data to send w.r.t. protocol  is successfully written before a `poll_read` on a `Negotiated` stream, but where the subsequent `poll_flush()` is pending. Now `remaining` is empty and the next `poll_read()` will go straight to reading from the underlying I/O stream, despite the flush not having happened
yet, which can lead to a form of deadlock during protocol negotiation. This seems to be the cause for the sporadic upgrade timeouts in https://github.com/libp2p/rust-libp2p/issues/1629.

It was a happy coincidence that I looked into https://github.com/libp2p/rust-libp2p/issues/1629 again just before `async-io-1.1.5` was released, because up to `1.1.4`, write operations would occasionally yield without really needing to, [like this](https://github.com/stjepang/async-io/blob/master/src/lib.rs#L1495-L1502). `1.1.5` removed that random yielding for write operations but it is exactly these random yields that provoke the issue in the setup provided by https://github.com/libp2p/rust-libp2p/issues/1629 (i.e. [this code](https://github.com/BlackYoup/libp2p-handler-timeout-reproducer/tree/mplex-fix)).

Rather than complicating the existing code further in order to accommodate for this case (e.g. by wrapping the `remaining` bytes in an `Option` to distinguish whether they have been flushed or not), it seems preferable to simplify the code by giving up on this optimisation that only affects the last negotiation protocol message sent by the "listener". So we give up on the ability to combine data sent by the "listener" immediately after protocol negotiation together with the final negotiation frame in the same transport-level frame/packet. That seems acceptable. The code simplifications should be convincing.

### DIFF
--- a/misc/multistream-select/CHANGELOG.md
+++ b/misc/multistream-select/CHANGELOG.md
@@ -1,5 +1,10 @@
 # 0.8.3 [unreleased]
 
+- Fix a potential deadlock during protocol negotiation due
+  to a missing flush, potentially resulting in sporadic protocol
+  upgrade timeouts.
+  [PR 1781](https://github.com/libp2p/rust-libp2p/pull/1781).
+
 - Update dependencies.
 
 # 0.8.2 [2020-06-22]
--- a/misc/multistream-select/src/dialer_select.rs
+++ b/misc/multistream-select/src/dialer_select.rs
@@ -241,8 +241,7 @@ where
                         }
                         Message::Protocol(ref p) if p.as_ref() == protocol.as_ref() => {
                             log::debug!("Dialer: Received confirmation for protocol: {}", p);
-                            let (io, remaining) = io.into_inner();
-                            let io = Negotiated::completed(io, remaining);
+                            let io = Negotiated::completed(io.into_inner());
                             return Poll::Ready(Ok((protocol, io)));
                         }
                         Message::NotAvailable => {
--- a/misc/multistream-select/src/length_delimited.rs
+++ b/misc/multistream-select/src/length_delimited.rs
@@ -76,22 +76,18 @@ impl<R> LengthDelimited<R> {
         }
     }
 
-    /// Drops the [`LengthDelimited`] resource, yielding the underlying I/O stream
-    /// together with the remaining write buffer containing the uvi-framed data
-    /// that has not yet been written to the underlying I/O stream.
-    ///
-    /// The returned remaining write buffer may be prepended to follow-up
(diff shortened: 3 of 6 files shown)


romanb:
> There appears to still be an edge-case in multistream-select whereby the remaining data to send w.r.t. protocol is successfully written before a poll_read on a Negotiated stream, but where the subsequent poll_flush() is pending. Now remaining is empty and the next poll_read() will go straight to reading from the underlying I/O stream, despite the flush not having happened
yet, which can lead to a form of deadlock during protocol negotiation. This seems to be the cause for the sporadic upgrade timeouts in #1629.

To make this a bit more concrete: If `Negotiated::poll_read` still has `remaining` data to send despite being in state `Completed`, it [calls Negotiated::poll()](https://github.com/libp2p/rust-libp2p/blob/master/misc/multistream-select/src/negotiated.rs#L211) to complete protocol negotiation. Then we [try to flush()](https://github.com/libp2p/rust-libp2p/blob/master/misc/multistream-select/src/negotiated.rs#L95) which [writes the remaining negotiation data](https://github.com/libp2p/rust-libp2p/blob/master/misc/multistream-select/src/negotiated.rs#L277-L285]). However, if the [final flush](https://github.com/libp2p/rust-libp2p/blob/master/misc/multistream-select/src/negotiated.rs#L285) is `Pending`, `remaining` is already empty and on the next `Negotiated::poll_read()` we go straight to reading from the underlying I/O stream, but the remote may still be waiting for the remaining negotiation data/frame before sending and that still sits around unflushed in the buffer. The sporadic, random yields by `async-io <= 1.1.4` for write operations, including `poll_flush()` is what made this more likely to occur every now and then.

mxinden (approved):
Great catch! Implementation looks good to me.

> So we give up on the ability to combine data sent by the "listener" immediately after protocol negotiation together with the final negotiation frame in the same transport-level frame/packet.

For e.g. TCP connections where `TCP_NODELAY` is not set, hence Nagle's algorithm still being active, final negotiation frames might still be bundled with upper layer payloads in the same transport-level frame, no?

romanb:
> For e.g. TCP connections where TCP_NODELAY is not set, hence Nagle's algorithm still being active, final negotiation frames might still be bundled with upper layer payloads in the same transport-level frame, no?

I think so, yes.

PR #1783, opened 2020-10-05
[multistream-select] Fix panic with `V1Lazy` (regression) and more convenient transport boxing.

romanb:
Fixes a panic when using the `V1Lazy` negotiation protocol, a regression introduced in [1484](https://github.com/libp2p/rust-libp2p/pull/1484/files#diff-6ce2135f99a1031d8c2df76296d2be52L122-R136)(Upgrade to stable futures). Thereby adds integration tests for a transport upgrade with both `V1` and `V1Lazy` to the `multistream-select` crate to prevent future regressions.

As a small refactoring on the side, I changed `Transport::boxed` to always also box the `StreamMuxer` and the transport errors, as this is how boxed transport are pretty much always used. I don't think it is necessary to support boxing of a transport at various levels of granularity. Now, when you want a boxed transport you just need to tack on `.boxed()` at the end of the transport builder and get a boxed transport with a boxed stream muxer and boxed transport errors, i.e. a simple boxed type for the whole transport.

### DIFF
--- a/misc/multistream-select/src/negotiated.rs
+++ b/misc/multistream-select/src/negotiated.rs
@@ -127,6 +127,7 @@ impl<TInner> Negotiated<TInner> {
 
                     if let Message::Header(v) = &msg {
                         if *v == version {
+                            *this.state = State::Expecting { io, protocol, version };
                             continue
                         }
                     }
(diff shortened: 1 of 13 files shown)


misc/multistream-select/src/negotiated.rs
```diff
@@ -127,6 +127,7 @@ impl<TInner> Negotiated<TInner> {
 
                     if let Message::Header(v) = &msg {
                         if *v == version {
+                            *this.state = State::Expecting { io, protocol, version };
```
romanb, line 130: This is the patch for the regression, where this line got removed.

tomaka (approved):
Looks good to me

PR #1784, opened 2020-10-05
[mplex] Split the receive buffer per substream.

romanb:
This PR picks up a discussion from https://github.com/libp2p/rust-libp2p/pull/1769#discussion_r494942944 about splitting the shared receive buffer per substream. This split allows more efficient reading from the buffer for a particular substream, to efficiently drop all still buffered frames for a particular stream when it is dropped, and to reset only the offending substream if it reaches its buffer limit with `MaxBufferBehaviour::ResetStream`. Previously this was implemented as `MaxBufferBehaviour::CloseAll` and resulted in the entire connection closing. The buffer split should generally be advantageous whenever not all substreams are read at the same pace and some temporarily fall behind in consuming their data.

### DIFF
--- a/muxers/mplex/src/io.rs
+++ b/muxers/mplex/src/io.rs
@@ -191,8 +191,8 @@ where
                     let id = stream_id.into_local();
                     if let Some(state) = self.substreams.get_mut(&id) {
                         if let Some(buf) = state.recv_buf_open() {
-                            trace!("Buffering {:?} for stream {} (total: {})", data, id, buf.len() + 1);
                             buf.push(data);
+                            trace!("Buffered {:?} for stream {} (total: {})", data, id, buf.len());
                             self.notifier_read.wake_read_stream(id);
                         } else {
                             trace!("Dropping data {:?} for closed or reset substream {}", data, id);
@@ -233,6 +233,8 @@ where
                         self.substreams.insert(stream_id, SubstreamState::Open {
                             buf: Default::default()
                         });
+                        log::debug!("New outbound substream: {} (total {})",
+                            stream_id, self.substreams.len());
                         // The flush is delayed and the `Open` frame may be sent
                         // together with other frames in the same transport packet.
                         self.pending_flush_open = true;
@@ -407,8 +409,8 @@ where
                     if let Some(state) = self.substreams.get_mut(&id) {
                         if let Some(buf) = state.recv_buf_open() {
                             debug_assert!(buf.len() <= self.config.max_buffer_len);
-                            trace!("Buffering {:?} for stream {} (total: {})", data, id, buf.len() + 1);
                             buf.push(data);
+                            trace!("Buffered {:?} for stream {} (total: {})", data, id, buf.len());
                             self.notifier_read.wake_read_stream(id);
                             if buf.len() > self.config.max_buffer_len {
                                 debug!("Frame buffer of stream {} is full.", id);
@@ -436,9 +438,8 @@ where
                 }
                 frame @ Frame::Open { .. } => {
                     if let Some(id) = self.on_open(frame.remote_id())? {
-                        let new_len = self.open_buffer.len() + 1;
-                        trace!("Buffering new inbound stream {} (total: {})", id, new_len);
                         self.open_buffer.push_front(id);
+                        trace!("Buffered new inbound stream {} (total: {})", id, self.open_buffer.len());
                         self.notifier_read.wake_next_stream();
                     }
                 }
@@ -600,11 +601,11 @@ where
             return Ok(None)
         }
 
-        log::debug!("New inbound stream: {}", id);
-
         self.substreams.insert(id, SubstreamState::Open {
             buf: Default::default()
         });
+
+        log::debug!("New inbound substream: {} (total {})", id, self.substreams.len());
 
         Ok(Some(id))
     }
--- a/muxers/mplex/src/io.rs
+++ b/muxers/mplex/src/io.rs
@@ -382,7 +382,7 @@ where
             if let Some(blocked_id) = &self.blocked_stream {
                 // We have a blocked stream and cannot continue reading
                 // new frames for any stream, until frames are taken from
-                // the blocked stream's buffer. Try to wake pending readers
+                // the blocked stream's buffer. Try to wake a pending reader
                 // of the blocked stream.
                 if !self.notifier_read.wake_read_stream(*blocked_id) {
                     // No task dedicated to the blocked stream woken, so schedule
--- a/muxers/mplex/src/io.rs
+++ b/muxers/mplex/src/io.rs
@@ -50,7 +50,7 @@ pub struct Multiplexed<C> {
     /// The stream that currently blocks reading for all streams
     /// due to a full buffer, if any. Only applicable for use
     /// with [`MaxBufferBehaviour::Block`].
-    blocked_stream: Option<LocalStreamId>,
+    blocking_stream: Option<LocalStreamId>,
     /// Pending frames to send at the next opportunity.
     ///
     /// An opportunity for sending pending frames is every flush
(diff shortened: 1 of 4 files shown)


romanb:
Naturally, the semantics of the configuration option `max_buffer_len` thus change to be per substream so we may want to consider changing the default according to the default of `max_substreams`. `max_substreams` together with `max_buffer_len` together still permit reasoning about bounded memory usage.

muxers/mplex/src/io.rs
```diff
@@ -40,11 +41,16 @@ pub struct Multiplexed<C> {
...
+    /// The buffer of new inbound substreams that have not yet
+    /// been drained by `poll_next_stream`.
+    open_buffer: VecDeque<LocalStreamId>,
     /// Whether a flush is pending due to one or more new outbound
     /// `Open` frames, before reading frames can proceed.
     pending_flush_open: bool,
+    /// The stream that currently blocks reading for all streams
+    /// due to a full buffer, if any. Only applicable for use
+    /// with [`MaxBufferBehaviour::Block`].
+    blocked_stream: Option<LocalStreamId>,
```
mxinden, line 53:
```suggestion
    blocking_stream: Option<LocalStreamId>,
```

Wouldn't `blocking` be more descriptive, given that this stream is blocking other streams which are thus blocked?

muxers/mplex/src/io.rs
```diff
@@ -346,37 +352,96 @@ where
...
+                // permits errors on "terminal" conditions, e.g. if the connection
+                // has been closed or on protocol misbehaviour.
                 return Poll::Ready(Ok(None))
             }
 
+            // Check if there is a blocked stream.
+            if let Some(blocked_id) = &self.blocked_stream {
+                // We have a blocked stream and cannot continue reading
+                // new frames for any stream, until frames are taken from
+                // the blocked stream's buffer. Try to wake pending readers
```
mxinden, line 385:
```suggestion
                // the blocked stream's buffer. Try to wake a pending reader
```

As far as I can tell we can only wake the most recent reader, right?

muxers/mplex/src/io.rs
```diff
@@ -346,37 +352,96 @@ where
...
+                // of the blocked stream.
+                if !self.notifier_read.wake_read_stream(*blocked_id) {
+                    // No task dedicated to the blocked stream woken, so schedule
+                    // this task again to have a chance at progress.
```
mxinden, line 389:
```suggestion
                    // this task again to have a chance to progress.
```

muxers/mplex/src/io.rs
```diff
@@ -346,37 +352,96 @@ where
...
+                    cx.waker().clone().wake();
+                } else if blocked_id != &id {
```
mxinden, line 391:
```suggestion
                } else {
                    if blocked_id == &id {
                        debug_assert!(false, "...");
                    }     
                }
```

What do you think of ensuring that a read on a blocking stream never makes it this far?

mxinden (approved):

muxers/mplex/src/io.rs
romanb: I don't think I could infer the behaviour w.r.t. all substreams from either name and would always need to look at the comment, but I'm fine with the name change.

muxers/mplex/src/io.rs
romanb: Yes, that comment is a left-over from the previous implementation. Thanks!

muxers/mplex/src/io.rs
romanb: I think "at" is grammatically correct as well. Do you see a specific reason to prefer "to" in this context?

mxinden: Was more of a gut feeling. Sorry for the noise.

muxers/mplex/src/io.rs
romanb: Why do you think it can never happen that the current task is woken at line `387`?

muxers/mplex/src/io.rs
romanb: Done in https://github.com/libp2p/rust-libp2p/pull/1784/commits/d689611611763c2b844afce13aab500204ab2d6f.

muxers/mplex/src/io.rs
mxinden: I do think it can happen that the current task is woken up in `387` and thus control flow reaches into the `else` block in `391`, but only when the current task is not trying to read from a stream that is currently blocking the connection.

My suggestion is confusing. I hope I am not missing something. Let me retry:

Say the read buffer of stream _S1_ is full and thus is blocking the entire connection. Say that the current task is trying to read from _S1_. In this case the current task would return `Poll::Ready(Ok(Some(data)))` in `364` already before the `loop`.

With the above in mind, one can not reach the loop when trying to read from the stream (in the case above _S1_) that is currently blocking the connection. To enforce this one could insert a `debug_assert` below `382` like such:

```diff
             if let Some(blocked_id) = &self.blocking_stream {
+               if blocked_id == id { debug_assert!(false, "..."); }
+
                // We have a blocked stream and cannot continue reading
                // new frames for any stream, until frames are taken from
                // the blocked stream's buffer. Try to wake a pending reader
                // of the blocked stream.
```

I don't think this is particularly important. Thus feel free to ignore this suggestion. What I would like to ensure though is that I am not misunderstanding something.

romanb: I see. Your comment made me aware though that, contrary to my own expectations, the check for the blocking stream is currently only in `poll_read_stream`. That means `poll_next_stream` is never blocked and can grow the substream buffers indefinitely, only limited by hitting the `max_substreams` limit. I corrected that [here](https://github.com/libp2p/rust-libp2p/pull/1784/commits/fde6f008d489abf259e2cc1978c7212167b4c37e) together with more commentary. I would appreciate if you could give that another critical look to see if it all makes sense now. I'll see if I can come up with some property tests w.r.t. the `MaxBufferBehaviour`.

romanb:
> Naturally, the semantics of the configuration option max_buffer_len thus change to be per substream so we may want to consider changing the default according to the default of max_substreams. max_substreams together with max_buffer_len still permit reasoning about bounded memory usage.

I went ahead and changed the default `max_buffer_len` accordingly [here](https://github.com/libp2p/rust-libp2p/pull/1784/commits/7f5a05ee50d7aaa0e37996445b407d905f0cbc1e).

mxinden (commented):
> I would appreciate if you could give that another critical look to see if it all makes sense now.

I like how you extracted buffering into `fn buffer` and handling blocking on max buffer size reach within `fn poll_read_frame`. Adjustments look good to me.

> I'll see if I can come up with some property tests w.r.t. the MaxBufferBehaviour.

That would be great. Curious what it might find.

muxers/mplex/src/io.rs
```diff
@@ -661,70 +723,158 @@ where
...
+    ///
+    /// If the new data frame exceeds the `max_buffer_len` for the buffer
+    /// of the substream, the behaviour depends on the configured
+    /// [`MaxBufferBehaviour`]. Note that the excess frame is still
+    /// buffered in that case (but no further frames will be).
+    ///
+    /// Fails the entire multiplexed stream if too many pending `Reset`
+    /// frames accumulate when using [`MaxBufferBehaviour::ResetStream`].
+    fn buffer(&mut self, id: LocalStreamId, data: Bytes) -> io::Result<()> {
+        if let Some(state) = self.substreams.get_mut(&id) {
```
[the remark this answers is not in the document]
romanb: I incorporated that suggestion [here](https://github.com/libp2p/rust-libp2p/pull/1784/commits/e5887b6798975b361f67bfb29761dd8cd78d15a3) together with some randomised testing for the `MaxBufferBehaviour`. It didn't reveal any new issues but increased confidence. There is a lot of potential for randomised (state machine) testing for this kind of code.

romanb:
@mxinden Thanks for your reviews.

mxinden (approved):

PR #1785, opened 2020-10-06
[mplex] Small enhancements.

romanb:
This PR builds on https://github.com/libp2p/rust-libp2p/pull/1784 (proper diff [here](https://github.com/romanb/rust-libp2p/compare/mplex-split-buffer...romanb:mplex-polish?expand=1)), making a few further small improvements:

  * Replace use of `fnv` with `nohash-hasher` to avoid unnecessary hashing of substream IDs.
  * Improve logging through randomly generated connection IDs in the same way as in Yamux.
  * More granular execution of pending flushes for newly opened outbound streams before reading. Otherwise there is a theoretical possibility of both ends being unnecessarily blocked on reading due to attempting to flush and both sides cannot flush for some reason. It is only necessary to flush before a read if the particular substream for which the read happens has just been opened, but not for other existing substreams.

### DIFF
--- a/muxers/mplex/src/io.rs
+++ b/muxers/mplex/src/io.rs
@@ -29,32 +29,35 @@ use nohash_hasher::{IntMap, IntSet};
 use parking_lot::Mutex;
 use smallvec::SmallVec;
 use std::collections::VecDeque;
-use std::{cmp, fmt, io, mem, sync::Arc, task::{Context, Poll, Waker}};
+use std::{cmp, fmt, io, mem, task::{Context, Poll, Waker}};
+use std::sync::{Arc, atomic::{AtomicU32, Ordering}};
 
 pub use std::io::{Result, Error, ErrorKind};
+
+const NEXT_CONNECTION_ID: AtomicU32 = AtomicU32::new(0);
 
 /// A connection identifier.
 ///
 /// Randomly generated and mainly intended to improve log output
 /// by scoping substream IDs to a connection.
 #[derive(Clone, Copy)]
-struct Id(u32);
+struct ConnectionId(u32);
 
-impl fmt::Debug for Id {
+impl fmt::Debug for ConnectionId {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        write!(f, "{:08x}", self.0)
+        write!(f, "{:04x}", self.0)
     }
 }
 
-impl fmt::Display for Id {
+impl fmt::Display for ConnectionId {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        write!(f, "{:08x}", self.0)
+        write!(f, "{:04x}", self.0)
     }
 }
 /// A multiplexed I/O stream.
 pub struct Multiplexed<C> {
     /// A unique ID for the multiplexed stream (i.e. connection).
-    id: Id,
+    id: ConnectionId,
     /// The current operating status of the multiplex stream.
     status: Status,
     /// The underlying multiplexed I/O stream.
@@ -113,7 +116,7 @@ where
 {
     /// Creates a new multiplexed I/O stream.
     pub fn new(io: C, config: MplexConfig) -> Self {
-        let id = Id(rand::random());
+        let id = ConnectionId(NEXT_CONNECTION_ID.fetch_add(1, Ordering::Relaxed));
         debug!("New multiplexed stream: {}", id);
         Multiplexed {
             id,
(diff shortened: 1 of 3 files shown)


muxers/mplex/src/io.rs
```diff
@@ -22,22 +22,42 @@ use bytes::Bytes;
...
+use std::{cmp, fmt, io, mem, sync::Arc, task::{Context, Poll, Waker}};
 
 pub use std::io::{Result, Error, ErrorKind};
 
+/// A connection identifier.
+///
+/// Randomly generated and mainly intended to improve log output
+/// by scoping substream IDs to a connection.
+#[derive(Clone, Copy)]
+struct Id(u32);
```
mxinden, line 41: I find it a bit confusing that `Id` is an identifier for a (multiplexed) stream whereas `LocalStreamId` is an identifier for a substream. Would it not be more descriptive if we had `StreamId` and `SubstreamId` ( / `LocalSubstreamId`)?

mxinden, line 41: If I am not mistaken with the birthday paradox in mind on a server with ~100_000 connections the chance for a collision is > 50%. While these identifiers are only used for logging (for now) is the expected performance win with a u32 instead of a u64 worth the risk of a collision?

romanb: It's not a matter of performance but of compact representation in the logs. Compare:
```rust
[2020-10-14T11:01:44Z DEBUG libp2p_mplex::io] f00c8822: New inbound substream: (18/receiver) (total 19)
[2020-10-14T11:01:44Z TRACE libp2p_mplex::io] f00c8822: Received Open { stream_id: RemoteStreamId { num: 19, role: Dialer } }
```
with
```rust
[2020-10-14T11:02:40Z TRACE libp2p_mplex::io] 19350c447f51a00a: Received Data { stream_id: RemoteStreamId { num: 0, role: Dialer }, data: b"Hello world" }
[2020-10-14T11:02:40Z TRACE libp2p_mplex::io] 19350c447f51a00a: Buffering b"Hello world" for stream (0/receiver) (total: 139)
```
I don't mind using a `u64` instead if you don't think the long representation in the logs is too distracting. We could also use a `u64` but case to `u32` for logging, but that may just add confusion.

romanb: I'm not really happy with either naming and struggled with it. "Stream" is certainly an appropriate and general term that can be used in all these cases and what kind of stream it is depends on the context. Making that context completely unambiguous in the name is tricky. The relative prefix "sub" only seems to help because in the context of libp2p we are mainly concerned with connections with one level of multiplexing. The API of this module currently uses the terminology `Multiplexed::poll_open_stream` etc. because it is clear from the context that these are "sub" streams of a multiplexed stream. Admittedly in the documentation and logs I've added the "sub" prefix for consistency with the public `StreamMuxer` API and terminology. The `Multiplexed` API on the other hand is only internal. All that said, would you go along for now with `Id` -> `ConnectionId`?

romanb: What do you think of https://github.com/libp2p/rust-libp2p/pull/1785/commits/f6bb3544128f961a23261a6fb71a56cd4504c901?

romanb: What do you think of https://github.com/libp2p/rust-libp2p/pull/1785/commits/f6bb3544128f961a23261a6fb71a56cd4504c901?

mxinden: > The Multiplexed API on the other hand is only internal. All that said, would you go along for now with Id -> ConnectionId?

Yes. I am in favor of `ConnectionId` as it mirrors the [`core` `ConnectionId`](https://github.com/libp2p/rust-libp2p/blob/master/core/src/connection/manager.rs#L77).

mxinden: I am not in favor of introducing a global variable. While still only used for logging, I would find it confusing that connection ids are monotonic e.g. when executing tests in serial, and possibly not monotonic (per test) when run concurrently.

Even though it is a bit verbose, I would prefer the u64, simply as there are no edge-cases. But I am fine with the u32 as well (maybe with a comment pointing out the probability of collisions).

romanb: > I am not in favor of introducing a global variable. While still only used for logging, I would find it confusing that connection ids are monotonic e.g. when executing tests in serial, and possibly not monotonic (per test) when run concurrently.

I don't quite understand your concerns with monotonic connection IDs or why you are referring to a module-scoped atomic counter as a global variable, especially in contrast to the use of a global random number generator, but I'm OK with the random `u64` for now.

romanb: See https://github.com/libp2p/rust-libp2p/pull/1785/commits/b98234b3dc6f1ae80cc071006dfaf9cca81e119c.

mxinden (approved):
Looks good to me.

PR #1788, opened 2020-10-08
Replace `get_if_addrs` with `if-addrs`

koushiro:

The crate [get_if_addrs](https://github.com/maidsafe-archive/get_if_addrs) is no longer maintained by its authors.
I think the [if-addrs](https://github.com/messense/if-addrs) (forked from [get_if_addrs](https://github.com/maidsafe-archive/get_if_addrs)) is a good alternative.

If we find a better library written in pure rust later, we can use it instead of `if-addrs` (see #1489)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,7 +27,7 @@
 
 - Update `libp2p-core`, `libp2p-floodsub`, `libp2p-gossipsub`, `libp2p-mplex`,
   `libp2p-noise`, `libp2p-plaintext`, `libp2p-pnet`, `libp2p-request-response`,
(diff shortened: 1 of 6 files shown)


koushiro:
@romanb @mxinden PTAL

romanb (approved):

mxinden (approved):
Thanks @koushiro!

mxinden:
@koushiro can you bump the version in `transports/tcp/Cargo.toml` to `v0.22.1` as done in the CHANGELOG.md?

koushiro:
@mxinden Done

messense:
if-addrs crate only uses C code to support old Android version.

PR #1792, opened 2020-10-13
feat: upgrade multihash

vmx:
rust-multihash changed a lot. It now uses only stack allocation, which
leads to different/sometimes more limited APIs. This commit updates
to the most recent version.

It also doesn't pull in unused hashing functions like e.g. BLAKE2.

I tried to make as little changes as possible. If you think bigger changes
should be made (perhaps the way rust-multihash works now can be
leveraged), please let me know.

I currently point to a branch of rust-multihash and not a release. The reason
is that I'd like to wait for a release until this PR is reviewed and ready to
be merged. In case there are any changes needed on the rust-multihash
side of things, in order to make it nicer to use with libp2p, we can still make
those easily (even breaking changes).


core/src/peer_id.rs
```diff
@@ -60,10 +60,9 @@ impl cmp::PartialOrd for PeerId {
 
 impl cmp::Ord for PeerId {
     fn cmp(&self, other: &Self) -> cmp::Ordering {
-        // must use borrow, because as_bytes is not consistent with equality
-        let lhs: &[u8] = self.borrow();
-        let rhs: &[u8] = other.borrow();
-        lhs.cmp(rhs)
+        let lhs = self.to_bytes();
```
mxinden, line 63: Given that `Multihash::to_bytes` allocates a `Vec` encoding the `Multihash` into it, would it not be quite expensive to sort a long list of `PeerId`s?

mxinden (commented):
I don't have much experience in this part of the codebase. Hope the comments below are of some help.

> I currently point to a branch of rust-multihash and not a release. The reason
is that I'd like to wait for a release until this PR is reviewed and ready to
be merged.

:+1:

core/src/peer_id.rs
```diff
@@ -122,29 +121,29 @@ impl PeerId {
     /// equality of peer IDs. That is, two peer IDs may be considered equal
     /// while having a different byte representation as per `into_bytes`.
     pub fn into_bytes(self) -> Vec<u8> {
```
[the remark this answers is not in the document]
vmx: Yes I think so. Though I wanted to keep changes quite minimal. But I'm happy to remove `into_bytes()`. I'll do that and we'll so how that will look like.

core/src/peer_id.rs
vmx: Would it be possible to just use the digest (without the multihash code and the length) instead? Then we wouldn't need allocations.

vmx:
Thanks for the review! I'm not sure about the ordering (just deriving it) works as expected (though I'd hope so).

core/Cargo.toml
```diff
@@ -12,7 +12,7 @@ categories = ["network-programming", "asynchronous"]
 [dependencies]
 asn1_der = "0.6.1"
 bs58 = "0.3.0"
-ed25519-dalek = "1.0.1"
+ed25519-dalek = "1.0.0-pre.4"
```
mxinden, line 15: Why downgrade this dependency?

https://crates.io/crates/ed25519-dalek

core/Cargo.toml
```diff
@@ -21,7 +21,7 @@ lazy_static = "1.2"
 libsecp256k1 = { version = "0.3.1", optional = true }
 log = "0.4"
 multiaddr = { package = "parity-multiaddr", version = "0.9.2", path = "../misc/multiaddr" }
-multihash = "0.11.0"
+multihash = { version = "0.12", default-features = false, features = ["std", "arb", "derive", "sha2"] }
```
mxinden, line 24:
```suggestion
multihash = { version = "0.12", default-features = false, features = ["std", "derive", "sha2"] }
```

I think we need `Arbitrary` in test code only, right? If so, can this be moved to `dev-dependencies`?

mxinden (commented):
Sorry @vmx for the slow review cycles.

core/Cargo.toml
vmx: That's a rebase failure from my side, I'll fix it.

core/Cargo.toml
vmx: Yes it can. At the moment it doesn't make any difference, but in future Rust versions it could. I'll move it down.

vmx:
@mxinden no worries about the slow review. It also gave me time to move slow and do other stuff :)

I've rebased and squashed a few commits for a cleaner history.

core/src/multihash.rs
```diff
@@ -0,0 +1,34 @@
...
+pub type Multihash = MultihashGeneric<U64>;
+
+#[derive(Copy, Clone, Debug, Eq, Multihash, PartialEq)]
+#[mh(alloc_size = U64)]
+pub enum Code {
+    #[mh(code = 0x00, hasher = multihash::IdentityHasher::<U64>, digest = multihash::IdentityDigest<U42>)]
+    Identity,
+    #[mh(code = 0x12, hasher = multihash::Sha2_256, digest = multihash::Sha2Digest<U32>)]
+    Sha2_256,
+}
```
tomaka, line 34: I don't understand why `multihash` has been changed to require this type of glue code?

All we want is a small crate that provides a glorified `struct Multihash(Code, Vec<u8>); enum Code { Identity, Sha2 }`, not (to exaggerate) an entreprise framework.

vmx: If you only want a glorified `struct Multihash(Code, Vec<u8>)`, you don't need to do that, you could just import `Multihash` instead. Though then you would lose the ability to do the hashing of data, it would need to be done separately. I can also do a PR to see what that would look like.

The reason to do it this way (with a custom table) is that with the default Multihash table, you would also pull in hash algorithm dependencies you never need. Also the default table doesn't support the Identity hash (long story short, Identity hash is too often not a good idea, hence it's not part of it by default (in libp2p it's OK though)).

vmx: I just had an idea that might work for libp2p to make this way simpler, let me have a try an I come back to you to see if that fits your needs better.

vmx: > I don't really understand why `multihash` has been changed to require this type of glue code?

`multihash` is now a bit more low level and more like a building block for your application. If you don't want to pull in several hashers you don't even need, you need to create your own table.

I've pushed a commit which makes things more explicit and is not using the macro. I hope that glue is small enough and also makes it clearer why it is needed.

I tried a simpler version, with just having constants for `IDENTITY` and `SHA2_256` and a `digest` method, but that didn't really end up nicely. The current code is way more type safe (though again more similar to the original code).

vmx: And yet another idea. In `multihash` the default table has a feature flag for each of is codec. I thought it wasn't useful as you would either want to experiment with `multihash`, hence just use the default table and don't care about the additional dependencies. I further thought that if you only need a subset, you'd use the derive macro or implement it yourself.

Looks like I was wrong.

Let me have yet another look to see if I can revive those feature flags in a good way and whether libp2p could leverage that feature.

tomaka: Don't take this wrong, but at some point I don't really see the point of having a dependency anymore. I have the feeling that the code would be more simple if we did the prefixing and hashing manually.

vmx: Thanks for the feedback, that's exactly what I was looking for. I don't want to create libraries that don't cover real world use cases. So let's see what I can come up with in `multihash` to make it really usable for libp2p.

vmx: @tomaka I pushed a new branch of rust-multihash. Does this look better now (I like it a lot better)?

protocols/kad/src/behaviour.rs
```diff
@@ -546,7 +546,7 @@ where
     pub fn kbucket<K>(&mut self, key: K)
         -> Option<kbucket::KBucketRef<'_, kbucket::Key<PeerId>, Addresses>>
     where
-        K: Borrow<[u8]> + Clone
```
romanb, line 549: The removal of `Borrow` constraints is very unfortunate because it weakens the semantics, i.e. `Into<Vec<u8>>` is too lax of a constraint. The generic code in `libp2p-kad` (and maybe elsewhere in libp2p where `Borrow` is used) wants a byte representation of some key type `K` whose `Eq`, `Ord` and `Hash` are consistent with those same impls for `K` in order for there to be no surprises (which we had in the past). That requirement is lost with `Into<Vec<u8>>` and in the absence of `Borrow` may probably only be recovered by a custom trait or excessive, repeated documentation that may more easily be overlooked compared to the more well-known semantics of `std::borrow::Borrow`.

core/src/multihash.rs
tomaka: :+1:

vmx:
Please excuse me for making a force-push. But it just made sense to start over fresh. It's again a bit of a different approach, but it now needs even less changes. Please have a look.

protocols/kad/src/behaviour.rs
vmx: Thanks for the feedback, I've changed my approach with the newest commit. Now the `Borrow` is preserved.

core/src/peer_id.rs
```diff
@@ -35,7 +35,7 @@ const MAX_INLINE_KEY_LENGTH: usize = 42;
 // TODO: maybe keep things in decoded version?
 #[derive(Clone, Eq)]
 pub struct PeerId {
-    multihash: Multihash,
+    multihash: Vec<u8>,
```
romanb, line 38:
```suggestion
    multihash: Bytes,
```
It would be good to retain the cheap cloning of `PeerId`s, as these are cloned a lot. See also https://github.com/libp2p/rust-libp2p/pull/1187.

protocols/kad/src/kbucket/key.rs
```diff
@@ -92,10 +91,10 @@ impl<T> Into<KeyBytes> for Key<T> {
     }
 }
 
-impl From<Multihash> for Key<Multihash> {
```
romanb, line 95: Is there a specific reason for replacing this instance with `From<Vec<u8>>`?

protocols/kad/src/kbucket/key.rs
```diff
@@ -182,22 +181,13 @@ impl Distance {
...
-    use multihash::{wrap, Code};
-    use rand::Rng;
 
     impl Arbitrary for Key<PeerId> {
         fn arbitrary<G: Gen>(_: &mut G) -> Key<PeerId> {
             Key::from(PeerId::random())
         }
     }
 
-    impl Arbitrary for Key<Multihash> {
```
romanb, line 194: Is there a specific reason for removing this instance?

protocols/kad/src/query/peers/closest.rs
```diff
@@ -487,7 +486,7 @@ mod tests {
             let known_closest_peers = random_peers(g.gen_range(1, 60), g)
                 .into_iter()
                 .map(Key::from);
-            let target = Key::from(Into::<Multihash>::into(PeerId::random()));
```
romanb, line 490: I know it is just a small technicality, but using a `Key<Multihash>` in these tests reflects common usage better. This should just be a `Key` based on a random `Multihash`, not sure why the current code goes through a `PeerId`.

src/lib.rs
```diff
@@ -161,8 +161,6 @@ pub use bytes;
 pub use futures;
 #[doc(inline)]
 pub use multiaddr;
-#[doc(inline)]
-pub use multihash;
```
romanb, line 165: What is the motivation for removing this re-export in favor of adding a re-export from `libp2p-core`? It would be good to avoid unnecessary API changes, unless you think there is a good reason to do so.

protocols/kad/src/kbucket/key.rs
```diff
@@ -92,10 +91,10 @@ impl<T> Into<KeyBytes> for Key<T> {
     }
 }
 
-impl From<Multihash> for Key<Multihash> {
-    fn from(m: Multihash) -> Self {
-        Key::new(m)
-    }
+impl From<Vec<u8>> for Key<Vec<u8>> {
```
romanb, line 94: Do we need this instance?

protocols/kad/src/kbucket/key.rs
vmx: The reason this is removed that it is not really possible to have a `Key<Multihash>` implementation. The problem is that `Key` needs to implement `Borrow<[u8]>`. But you cannot implement `Borrow<[u8]>` for Multihash as it doesn't own a continuous sequence of bytes containing the full multihash.

This relates to that comment https://github.com/libp2p/rust-libp2p/pull/1792#pullrequestreview-523440886 where I removed the `Borrow` to `Into<Vec<u8>>` in order to be able to implement `Key<Multihash>`.

protocols/kad/src/kbucket/key.rs
vmx: There is no `Key<Multihash>` instance as mentioned in the other comment.

protocols/kad/src/query/peers/closest.rs
vmx: Again the problem mentioned in the other comment that I don't find a way to get `Key<Multihash>` working.

src/lib.rs
vmx: The main motivation was, that rust-multihash makes heavy use of features and I thought having to get those right in a single `Cargo.toml` file is easier. That's the reason I re-exported in `libp2p-core`, so that other libp2p crates can use it easily (as the ones using Multihash all depend on `libp2p-core` anyway). I overlooked that I break the public API of `libp2p`.

I see several options, please let me know which one to pick:

 1. Keeping this change
 2. re-exporting `multihash` with its default features
 3. re-exporting `multihash` with the same features defined as in `libp2p-core`
 4. re-exporting `multihash` from `libp2p-core`.

protocols/kad/src/kbucket/key.rs
vmx: That is my workaround to convert a `Multihash` to a `Key` (as `Key<Multihash> is missing). You then call `Key::from(my_multihash.to_bytes()`.

protocols/kad/src/kbucket/key.rs
romanb: > The problem is that Key needs to implement Borrow<[u8]>.

Not quite, the type itself does not. The public constructors `Key::new` and `KeyBytes::new` need this constraint, but I think the `From<Multihash>` instance can bypass these (possibly extracting code from these public constructors to internal ones that can be used in both places), as it can ensure that the byte representation of the multihash that is used to create the `Key` is consistent with the original `Multihash` w.r.t. equality. Would that not work? I guess there is no reason not to also have a `From<Vec<u8>>` instance but it would be preferable not to settle for a `Key<Vec<u8>>` when one wants a `Key<Multihash>`.

core/src/peer_id.rs
vmx: Done.

protocols/kad/src/kbucket/key.rs
vmx: You are absolutely right. Sorry that I missed that. Thanks for the detailed review and helping me.

src/lib.rs
romanb: How about `3 + 4`, i.e. just doing
```rust
pub use core::multihash;
```
here?

vmx: Yes that is what I meant with `4.`. I'll do that.

core/Cargo.toml
```diff
@@ -12,6 +12,7 @@ categories = ["network-programming", "asynchronous"]
 [dependencies]
 asn1_der = "0.6.1"
 bs58 = "0.3.0"
+bytes = "0.6.0"
```
romanb, line 15: To avoid problems, I think it best to stick to `0.5` in this PR and leave the `0.6` update to https://github.com/libp2p/rust-libp2p/pull/1810.

vmx: As this PR needs changes on rust-multihash which aren't merged yet, it will take some time on that side. I'll take it into account it case that merging moves faster than the dependency upgrade.

vmx:
I've rebased once more and also downgraded `bytes` to 0.5. I squashed the commit as it now the upgrade is really pretty self-contained.

romanb (approved):

core/src/peer_id.rs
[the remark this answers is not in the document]
mxinden: With the recent change to `Bytes` to represent a `PeerId`, my comment is obsolete.

vmx:
Please let me know if that PR is close to be merged, I'm happy to rebase it again myself (or feel free to rebase it yourself).

romanb:
I think we will go ahead with merging early next week if there are no further comments until then.

vmx:
Thanks everyone involved. This review made rust-multihash easier to use and upgrade to.

Issue #1795, opened 2020-10-15
Response to `ls` message of multistream-select looks invalid


tomaka:
It seems that in rust-libp2p the response to `ls` messages starts with the number of protocols.

https://github.com/libp2p/rust-libp2p/blob/8cec457b5e1bdc3423fbb95e72a801986e76bcf8/misc/multistream-select/src/protocol.rs#L203

However the specs mention that it should start with the total size of the message: 
https://github.com/multiformats/multistream-select#ls-example-in-detail

romanb:
Our implementation probably predates https://github.com/multiformats/multistream-select/pull/19 and thus at some point was actually spec-compliant :innocent:.

romanb:
> However the specs mention that it should start with the total size of the message: [..]

Note that all messages that get `encode()`ed are sent down a `LengthDelimited` sink, so that size is prefixed there. The `ls` response you mention looks correct according to the old spec but apparently that was never the implementation, so the `<varint-number-of-protocols>` is what should apparently no longer be sent.

mxinden:
As far as I can tell `ls` is only used in the [`dialer_select_proto_parallel`](https://github.com/libp2p/rust-libp2p/blob/master/misc/multistream-select/src/dialer_select.rs) flow when negotiating with less than or equal 3 protocols. According to this [comment](https://github.com/multiformats/go-multistream/issues/41#issuecomment-525877132) go-libp2p is only using `ls` for debugging.

Does this imply that all rust-libp2p users interacting with go-libp2p users have at most 3 protocols to negotiate at a given point in time?

Any suggestions on an upgrade strategy to eventually be compliant with the spec while not breaking compatibility with the previous rust-libp2p version?

- We could (1) disable parallel negotiation, (2) change our `ls` message and (3) reenable parallel negotiation over multiple rust-libp2p versions.

- We could parse each `ls` response both in the old and the new format, hoping to not receive a message that would parse valid for both formats. Off the top of my head, given that protocols are both length delimited and newline delimited, this should work.

romanb:
> We could (1) disable parallel negotiation, (2) change our ls message and (3) reenable parallel negotiation over multiple rust-libp2p versions.

[...]

PR #1811, opened 2020-10-22
[multistream-select] Fix `ls` response encoding/decoding (spec-compliance).

romanb:
This is step 2 of addressing https://github.com/libp2p/rust-libp2p/issues/1795. It changes the `ls` response encoding/decoding according to the spec, that is (line breaks only for ease of reading):

```
<varint-ls-response-length>
<varint-protocol-length><protocol>\n
<varint-protocol-length><protocol>\n
# ...more protocols
\n
```

Incidentally, since `ls` responses are always terminated with a line feed and thus end with two line feeds (one for the last protocol and one for the entire message), we no longer need the artificial protocol length limit that was introduced earlier in order to ensure disambiguation of `ls` responses that contain a single protocol and singular protocol responses / confirmations as the latter can never end with a double line feed.


mxinden (approved):
This looks good to me.

Instead of merging already, I would delay the merge. In case we have an urgent feature but `rust-libp2p` `0.29` is not yet widely deployed, we can release right off of master. What do you think?

romanb:
Sounds ok to me.

romanb:
I would like to include this in the next release, if there are no objections.

mxinden:
> I would like to include this in the next release, if there are no objections.

No objections from my side.

Issue #1798, opened 2020-10-17
Reduce Generics - Hide implementation details.


semtexzv:
Hello, I've been lurking around libp2p for some time, doing some hacking, and have summarized my issues here.
If it interests anyone:

While the capabilities of current design are quite impressive, It seems to be overly complex for what it does. It should be possible to work at the application layer without having to interact with complex traits like `Transport` and `StreamMuxer`, and their associated types.

This makes it hard to modify the library (I've tried several times, and and just got discouraged by the insanity of having to specify 8 generic parameters in some places ).

Several ideas:

**1. Remove Substream type from `StreamMuxer`** 
  -  force every implementation to do its own bookkeeping for substreams, same way as the wrapper muxer does.
  - Or make each substream carry a reference to underlying stream as an `Arc<Mutex>>`
  - Generally make substreams an implementation detail. Protocols should only ever know about AsyncRead/Write

**2. Remove `ConnectionInfo` as a trait, replace it with `PeerId`**
  -  It seems that this is the only implementor of this trait
  -  In places, where it might be useful to have more data about peers, Introduce this in local way, don't pollute the network, Swarm, and all other traits if its not needed. 

**3. Use boxing in reasonable places.** 
  -  This is a general trend in rust, that nothing is ever boxed, everything resolved statically. While it might be slightly more performant in some cases, the use of complex traits and generics everywhere introduces tight coupling which means modifications are very hard to implement.
  - We already do a bunch of copying for each network message. One indirect call through trait object is nothing in comparison.

**5. Move away from custom built combinators for `Transport` and `Upgrade`** 
  - Try to design a simpler approach based on async-await futures.

**6. Make behaviors independent.** 
 -  Current approach of building a behavior out of arbitrary ones works on 1 level, (You can't reasonably compose IPFS behavior with others in your behaviour, since then you can't use `Kademlia` behaviour yourself). Structuring this as a hierarchy is not correct. This is a set.
 - Introduce a separate control flow construct for each behaviour, and run them in parallel. Maybe a design with future for each behavior, where the future will:
      a. Listen for network messages, 
      b. Listen for Internal messages
      c. Emit network messages
      d. Emit internal mesages to other behaviours

My goal is not to just be negative, but to provide feedback from working with the library, and some directions, in which I'd like to move. I'm going to try to move forward on some of these myself, but I'm not sure how far I'll be able to go.

[...]

PR #1808, opened 2020-10-20
core/Cargo.toml: Update to ed25519-dalek v1.0.1

mxinden:
See https://github.com/paritytech/substrate/pull/7341#discussion_r508516095 for motivation.

I guess dependabot didn't spot this given the `pre.4` suffix.

PR #1813, opened 2020-10-23
Remove ConnectionInfo trait

thomaseizinger:
In all cases, we pass the PeerId directly as the connection info.
The flexbility of doing something different here was originally
envisioned but turned out to be never needed.

For reference see: https://github.com/libp2p/rust-libp2p/issues/1798#issuecomment-714526056


core/src/network/event.rs
```diff
@@ -20,28 +20,24 @@
...
+    ConnectionId,
+    ConnectedPoint,
+    ConnectionError,
+    ConnectionHandler,
+    Connected,
+    EstablishedConnection,
+    IntoConnectionHandler,
+    ListenerId,
+    PendingConnectionError,
+}, transport::Transport, PeerId};
```
romanb, line 33: I don't usually comment on code formatting since I can work with almost anything, but this particular mixture of line breaks and no line breaks looks like its unduly giving prominence to the indented block of imports, with the others squashed between the opening and closing braces. Was this done automatically or manually? At the end of the day I'm sure I can work with this as well, I'm just curious.

thomaseizinger: I tried to not make any formatting changes but this one must have slipped through! Was definitely not on purpose but probably auto-formatted by my IDE :grimacing:

thomaseizinger: Reverted the formatting changes :)

swarm/src/lib.rs
```diff
@@ -123,18 +122,17 @@ use libp2p_core::{
...
 use smallvec::SmallVec;
-use std::{error, fmt, hash::Hash, io, ops::{Deref, DerefMut}, pin::Pin, task::{Context, Poll}};
+use std::{error, fmt, io, ops::{Deref, DerefMut}, pin::Pin, task::{Context, Poll}};
 use std::collections::HashSet;
 use std::num::{NonZeroU32, NonZeroUsize};
 use upgrade::UpgradeInfoSend as _;
 
 /// Contains the state of the network, plus the way it should behave.
-pub type Swarm<TBehaviour, TConnInfo = PeerId> = ExpandedSwarm<
+pub type Swarm<TBehaviour = PeerId> = ExpandedSwarm<
```
mxinden, line 131:
```suggestion
pub type Swarm<TBehaviour> = ExpandedSwarm<
```

mxinden (commented):
Thanks for doing this cleanup.

In case there are no objections to this change, would you mind including the corresponding `Cargo.toml` and `Changelog` changes for `core`, `swarm` and `/` in the pull request similar to https://github.com/libp2p/rust-libp2p/pull/1808/ ?

swarm/src/lib.rs
thomaseizinger: Upsi, the search and replace messed up this one :D

mxinden:
>  Which changes would you like to see in the Cargo.toml?

Would you mind cherry-picking the following commit? https://github.com/mxinden/rust-libp2p/commit/4ffde074b04b1bf97cda57983ddfdd5c74db968b

thomaseizinger:
> > Which changes would you like to see in the Cargo.toml?
> 
> Would you mind cherry-picking the following commit? [mxinden@4ffde07](https://github.com/mxinden/rust-libp2p/commit/4ffde074b04b1bf97cda57983ddfdd5c74db968b)

Done!

thomaseizinger:
@mxinden I had to update a couple of more versions for the code to actually compile. Please check if that is correct :)

romanb (approved):

mxinden (approved):
> @mxinden I had to update a couple of more versions for the code to actually compile. Please check if that is correct :)

Ah, yes. Sorry, I missed those. Thanks.

PR #1816, opened 2020-10-28
[mplex] Revert to being lenient with duplicate Close frames.

romanb:
Version  `0.23.0` started [treating duplicate `Close` frames for a substream as a protocol violation](https://github.com/libp2p/rust-libp2p/pull/1769#discussion_r493604381). As there seem to be libp2p implementations that occasionally send such frames (cf. https://github.com/sigp/lighthouse/issues/1832#issuecomment-717859849) and it is a harmless redundancy, I prefer and propose here again to be lenient about such redundant frames rather than requiring other implementations to change, especially since this may be considered a regression compared to previous versions of `libp2p-mplex`. This PR prepares release `libp2p-mplex-0.23.1`.

### DIFF
--- a/muxers/mplex/CHANGELOG.md
+++ b/muxers/mplex/CHANGELOG.md
@@ -1,3 +1,12 @@
+# 0.23.1 [2020-10-28]
+
+- Be lenient with duplicate `Close` frames received. Version
+  `0.23.0` started treating duplicate `Close` frames for a
+  substream as a protocol violation. As some libp2p implementations
+  seem to occasionally send such frames and it is a harmless
+  redundancy, this releases reverts back to the pre-0.23 behaviour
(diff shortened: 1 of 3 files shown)


mxinden (approved):
> I prefer and propose here again to be lenient about such redundant frames rather than requiring other implementations to change, especially since this may be considered a regression compared to previous versions of libp2p-mplex.

I agree. Thanks for the fix. Sorry for the trouble my suggestion has caused.

romanb:
> I agree. Thanks for the fix. Sorry for the trouble my suggestion has caused.

No worries, the suggestion was quite reasonable. This situation reminds me of the [robustness principle](https://en.wikipedia.org/wiki/Robustness_principle), which is not free of criticism.

Issue #1611, opened 2020-06-16
protocols/kad: Node with incompatible protocol name can be added to routing table


mxinden:
A Kademlia [`NetworkBehaviour`](https://github.com/libp2p/rust-libp2p/blob/master/protocols/kad/src/behaviour.rs#L1241) adds a node to its routing table via the `inject_connection_established` hook:

https://github.com/libp2p/rust-libp2p/blob/44236b5ba39cbd324b029cb9fb5ed77a75806706/protocols/kad/src/behaviour.rs#L1280-L1291

This hook is triggered before the protocol negotiation on substreams. Thereby a remote node with a protocol name other than the [set protocol name](https://github.com/libp2p/rust-libp2p/blob/master/protocols/kad/src/behaviour.rs#L132) could be added to the local node's routing table.

See https://github.com/libp2p/rust-libp2p/pull/1473#discussion_r440218337 for details.

romanb:
It should be noted though that, being disconnected shortly after, these peers don't have a high chance of _staying_ in the routing table, meaning peers that connect with the correct protocol name have a strong tendency to push these unwanted peers out of the buckets. Nevertheless, I think this issue shows that customising the protocol name on the negotiated substreams, while still useful, is not the right approach to clean segregation of DHTs. The options stated inhttps://github.com/libp2p/rust-libp2p/issues/1560 are more appropriate to achieve that in a cleaner way.

PR #1821, opened 2020-11-03
[kad] Update local routing table on new connections after protocol confirmation.

romanb:
This is a proposal for closing https://github.com/libp2p/rust-libp2p/issues/1611. In order to avoid adding peers to the local routing table which use a different protocol name and thus actually a different overlay network, only update the local routing table for newly established connections once the associated connection handler reports that the protocol has been confirmed. The configured protocol is confirmed on a connection once the first inbound or outbound substream is successfully negotiated.

### DIFF
--- a/protocols/kad/CHANGELOG.md
+++ b/protocols/kad/CHANGELOG.md
@@ -1,5 +1,14 @@
 # 0.25.0 [unreleased]
 
+- Upon newly established connections, delay routing table
+  updates until after the configured protocol name has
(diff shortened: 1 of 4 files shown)


mxinden (approved):
Great! Thanks for fixing this.

In Substrate we use manual insertion, but only check the protocol identifier before passing a node back to Kademlia. In case we never get around to deeper inspections of nodes, we might as well revert https://github.com/paritytech/substrate/pull/6549 once this is released.

PR #1823, opened 2020-11-05
misc/multistream-select: Differentiate interpretation of EOF

mxinden:
#### Summary / Commit Message

> When a listener rejects a protocol with [`Message::NotAvailable`] and
the dialer does not have alternative protocols to propose then the
dialer will stop the negotiation and drop the corresponding stream. As a
listener interpret an EOF after sending a [`Message::NotAvailable`] as a
failed negotiation. Interpret an EOF as an io error in all other cases.

#### Context

The above surfaced in the following way:

With https://github.com/paritytech/substrate/pull/7478 block-requests are send via `libp2p-request-response`. A full node supports both sending block requests and responding to block requests. A light client only supports sending block requests. When the full node asks the light client whether it supports the block request protocol, the light client replies with `Message::NotAvailable`. The full node thus drops the stream. Previously a dropped stream was interpreted as an io error on the light client side. Passed on to `libp2p-request-response` the connection is terminated (see [here](https://github.com/libp2p/rust-libp2p/blob/d988b05f8574edf9b92e096b483c48cddb9b79e3/protocols/request-response/src/handler.rs#L259)).

There are two solutions to this:

- Ignore io errors in `libp2p-request-responses`.

- Interpret an EOF after sending a `Message::NotAvailable` as a negotiation failure and not as an io failure.

This pull request proposes the latter.


mxinden:
Thanks to @tomaka who was a great help debugging this.

misc/multistream-select/src/tests.rs
```diff
@@ -85,9 +85,8 @@ fn no_protocol_found() {
             let protos = vec![b"/proto1", b"/proto2"];
             let io = match listener_select_proto(connec, protos).await {
                 Ok((_, io)) => io,
-                // We don't explicitly check for `Failed` because the client might close the connection when it
-                // realizes that we have no protocol in common.
```
tomaka, line 89: :eyes:

tomaka (approved):
Code changes look good to me.
The logic of the change itself might be a bit more debatable, but I agree with this solution

romanb:
> Interpret an EOF after sending a Message::NotAvailable as a negotiation failure and not as an io failure.

How about always interpreting EOF as a failed negotiation rather than only in this particular case? Essentially making explicit that dropping or closing an I/O stream is a permissible course of action to "gracefully" fail a negotiation. I'd prefer more uniform behavior over special cases.

tomaka:
The argument is that the multiplexing layer has no choice but to report protocol errors in the form of an `io::Error`.
By treating `io::Error` as "failed negotiation", we would silently ignore errors on the Yamux and Mplex protocols.

mxinden:
> The argument is that the multiplexing layer has no choice but to report protocol errors in the form of an `io::Error`.
> By treating `io::Error` as "failed negotiation", we would silently ignore errors on the Yamux and Mplex protocols.

If I understand correctly, @romanb is suggesting to treat EOF errors as negotiation errors, not to treat all `io::Error` as negotiation errors. More concretely to always treat `Poll::Ready(None)` as `NegotiationError::Failed`, not only after sending `NotAvailable`, and not to treat `Poll::Ready(Some(Err(_)))` as `NegotiationError::Failed`.

https://github.com/libp2p/rust-libp2p/blob/c812b8d56c2919bd00921c5ac06f6e3614c82356/misc/multistream-select/src/listener_select.rs#L152-L164

tomaka:
:+1: that is a sensible option as well.

romanb:
> If I understand correctly, @romanb is suggesting to treat EOF errors as negotiation errors [..]

Yes, I'm only referring to `io::ErrorKind::UnexpectedEof` respectively `Poll::Ready(None)` for `Stream` APIs. I see no reason not to treat unexpected EOF uniformly as `NegotiationFailed`, rather than only giving it special treatment when reading after sending `NotAvailable`.

tomaka (approved):

mxinden:
@romanb do my recent changes via https://github.com/libp2p/rust-libp2p/pull/1823/commits/01b54b891c648df23d8480fef3420239b3158316 look good to you? To make sure I did not misinterpret your suggestion I will wait for an approval from you. I don't mean to rush and appreciate the help.

romanb (approved):
Looks good to me, though I would have expected that the change is made for all cases of `Poll::Ready(None)` during negotiation, i.e. for all occurrences in `dialer_select.rs` and `listener_select.rs`, instead of only at one place for the listener.

mxinden:
> Looks good to me, though I would have expected that the change is made for all cases of `Poll::Ready(None)` during negotiation, i.e. for all occurrences in `dialer_select.rs` and `listener_select.rs`, instead of only at one place for the listener.

Agreed. https://github.com/libp2p/rust-libp2p/pull/1823/commits/adf31c8adda33fb48f40e36b79e27e85a7d9d9fa makes it consistent across all cases of `Poll::Ready(None)`. Thanks for the suggestion.

PR #1824, opened 2020-11-05
Fix the multiaddr parsing in the JS snippet

tomaka:
`parsed[4]` can contain for example `x-parity-ws//submit/`, and not just `x-parity-ws` alone


transports/wasm-ext/src/websockets.js
```diff
@@ -34,7 +34,7 @@ const multiaddr_to_ws = (addr) => {
 	let parsed = addr.match(/^\/(ip4|ip6|dns4|dns6|dns)\/(.*?)\/tcp\/(.*?)\/(ws|wss|x-parity-ws\/(.*)|x-parity-wss\/(.*))$/);
 	if (parsed != null) {
 		let proto = 'wss';
-		if (parsed[4] == 'ws' || parsed[4] == 'x-parity-ws') {
+		if (parsed[4] == 'ws' || parsed[4].startsWith('x-parity-ws/')) {
```
cecton, line 37:
```suggestion
		if (parsed[4].startsWith('ws') || parsed[4].startsWith('x-parity-ws/')) {
```

cecton: Oh actually in my case it was using x-parity-ws (I guess it gets translated somewhere).

I have tested and it works (for x-parity-ws so without my change proposal)

tomaka: The `/ws` and `/wss` suffixes are the standard libp2p multiaddresses protocols, but they have the drawback of not allowing specifying a path.
To circumvent this, we invented `x-parity-ws/<path>` and `x-parity-wss/<path>`.

It's always either `ws`, `wss`, `x-parity-ws/<path>`, or `x-parity-wss/<path>`. There's no `ws/<path>`.
In other words, your suggestion isn't correct/necessary.

romanb (approved):

transports/wasm-ext/src/websockets.js
cecton: I see. Thanks for the explanation

PR #802, opened 2018-12-20
Bump default Mplex split_send_size to 64Kbyte

dvdplm:
The default value of 1Kbyte is rather small and may impact throughput adversly. This PR proposes 64Kbyte. As a comparison, the [Yamux default `receive_window` size is 256Kbyte](https://github.com/hashicorp/yamux/blob/master/spec.md#flow-control).

PR #1831, opened 2020-11-10
[plaintext] Retain remaining read buffer after handshake.

romanb:
When the plaintext protocol handshake finishes and the `Framed` I/O used during the handshake is discarded in favor of the underlying I/O stream, the remaining read buffer of `Framed` must be retained, as it may have buffered data beyond the end of the handshake. This problem manifests itself in potential stalls / timeouts during protocol negotiation, because a peer does not necessarily wait for confirmation of a protocol before it sends the proposal for the next protocol upgrade. E.g. `/multistream/1.0.0 \n /mplex/6.7.0` for the mplex upgrade may already be received and buffered by the `Framed` of the plaintext handshake by a peer before the handshake is complete. That peer will read the final handshake frame but then discard the remaining read buffer of `Framed`, leaving the remote waiting for the protocol confirmation for the next upgrade, but that proposal got "lost". The problem first appeared in https://github.com/libp2p/rust-libp2p/pull/1765 where it was missed during review.

This surfaced, without looking for it, during some benchmark attempts with the TCP transport and plaintext protocol in order to settle https://github.com/libp2p/rust-libp2p/pull/802. However, because the symptoms are the same, I'm reasonably certain that this is the actual cause of the early connection negotiation stalls observed in https://github.com/libp2p/rust-libp2p/issues/1629#issuecomment-701396435, unrelated to the random upgrade timeouts (which should already be fixed), since that test setup intentionally uses the plaintext protocol. So hopefully this should resolve the remaining problem(s) mentioned in #1629 that I had trouble to reproduce earlier.

### DIFF
--- a/protocols/plaintext/src/handshake.rs
+++ b/protocols/plaintext/src/handshake.rs
@@ -22,7 +22,7 @@ use crate::PlainText2Config;
 use crate::error::PlainTextError;
 use crate::structs_proto::Exchange;
 
-use bytes::BytesMut;
+use bytes::{Bytes, BytesMut};
 use futures::prelude::*;
 use futures_codec::Framed;
 use libp2p_core::{PublicKey, PeerId};
@@ -111,7 +111,7 @@ impl HandshakeContext<Local> {
 }
 
 pub async fn handshake<S>(socket: S, config: PlainText2Config)
-    -> Result<(S, Remote), PlainTextError>
+    -> Result<(S, Remote, Bytes), PlainTextError>
 where
     S: AsyncRead + AsyncWrite + Send + Unpin,
 {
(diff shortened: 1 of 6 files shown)


protocols/plaintext/src/handshake.rs
```diff
@@ -134,6 +134,12 @@ where
         }
     };
 
+    // The `Framed` wrapper may have buffered additional data that
+    // was already received but is no longer part of the plaintext
+    // handshake. We need to capture that data before dropping
+    // the `Framed` wrapper via `Framed::into_inner()`.
+    let read_buffer = framed_socket.read_buffer().clone().freeze();
```
romanb, line 141: We may be able to avoid cloning here in the future, if something like https://github.com/matthunz/futures-codec/pull/54 is available.

mxinden (approved):
Well spotted. Thanks for fixing this!

romanb:
Thanks for the quick review. Any objections if I make a patch release straight away?

mxinden:
> Thanks for the quick review. Any objections if I make a patch release straight away?

@romanb patch release sounds good to me. Thanks!

romanb:
`libp2p-0.30.1` and `libp2p-plaintext-0.24.1` published.

PR #1833, opened 2020-11-11
Limit buffer capacity in multiaddr visit_seq

pawanjay176:
Buffer allocated in `visit_seq` does not have an upper limit. Limit buffer capacity to 4096.

### DIFF
--- a/misc/multiaddr/src/lib.rs
+++ b/misc/multiaddr/src/lib.rs
@@ -346,7 +346,7 @@ impl<'de> Deserialize<'de> for Multiaddr {
                 formatter.write_str("multiaddress")
             }
             fn visit_seq<A: de::SeqAccess<'de>>(self, mut seq: A) -> StdResult<Self::Value, A::Error> {
-                let mut buf: Vec<u8> = Vec::with_capacity(seq.size_hint().unwrap_or(0));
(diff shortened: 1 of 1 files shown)


romanb (approved):
Just to clarify, I'm not so familiar with the serialisation code: Since this is just for `Vec::with_capacity`, this limits only the _initial_ capacity of the buffer, so that you cannot make the deserialisation code allocate an arbitrarily large `Vec` without actually "sending" that much data (e.g. if the `Multiaddr` is received over the network), since the `Vec` will still grow beyond its capacity as necessary. Correct?

pawanjay176:
@romanb yes that is correct. Without a limit, you can send malformed input such that `seq.size_hint()` returns a very large value and crashes the program even if the following data is pretty small.

mxinden (approved):

mxinden:
Thanks @pawanjay176!

mxinden:
I pushed https://github.com/libp2p/rust-libp2p/commit/2e6b0ec1fac596aec03a32250655f00495be98fd to update patch version and changelog. Should we do a release already, or wait till the next larger release?

romanb:
No reason to wait, a multiaddr patch release seems appropriate.

mxinden:
For the record `parity-multiaddr` `v0.9.5` is now [released](https://github.com/libp2p/rust-libp2p/releases/tag/parity-multiaddr-0.9.5) and [published](https://crates.io/crates/parity-multiaddr/0.9.5).

Issue #725, opened 2018-12-04
Add back the relay


tomaka:
#723 removed the implementation of `relay` which was outdated and unusable.
It should be put back.

johan-bjareholt:
What's the progress on this? I have a use-case for this on a Rust project and would like to try prototyping it.

tomaka:
There needs to be some API design because we need to "inject" relayed connections inside the `RawSwarm`.

tomaka:
An approach could be to:

- Add a method to `RawSwarm` that allows injecting an "already-dialed" connection (which would in practice be a relayed substream). In other words, a method that accepts a `TTransport::Output`.

- On top of that, the `Swarm` would wrap around the `Transport` before passing it to the `RawSwarm`, so that its `Output` becomes some sort of `Either` type. All regular connections are wrapped inside `Either::First`, and all relayed substreams are wrapped inside `Either::Second`. This way, we can use the aforementioned method to directly inject a substream.

- Add an associated type to `NetworkBehaviour` and a variant to `NetworkBehaviourEvent` in order to request the `Swarm` that it injects a substream. Adding an associated type is quite a large disruption to the API, but I guess it's ok.

- The user would have to pass their encryption layer upgrade and multiplexing layer upgrade to the `RelayConfig`, so that the relay can upgrade inner substreams. Alternatively, we could modify the `Swarm` in order to pass these upgrades at initialization.

The part that's still unclear to me is how do we handle requests for dialing addresses as `/relay/...`. Should they really be handled transparently? That is, do we allow addresses containing `/relay/` to travel around the wire the same way as regular addresses and handle them the same way?

tomaka:
Alternatively, we could somehow improve the `Transport` trait so that it's capable of plugging the relay mechanism by itself.

Instead of having the `Swarm` and `RawSwam` require `Transport<Output = (PeerId, impl StreamMuxer)>`, we could instead require `Transport<Output = impl SomeTrait>` , where `SomeTrait` is a trait that has some logic attached to it.

tomaka:
> Add an associated type to NetworkBehaviour and a variant to NetworkBehaviourEvent in order t allow the behaviour to request the Swarm that it injects a substream. Adding an associated type is quite a large disruption to the API, but I guess it's ok.

That cannot work.

[...]

PR #1134, opened 2019-05-19
Add back the relay

tomaka:
Fix #725 

Non-working work-in-progress.
Adds back the relay protocol.

Continuing this PR is basically blocked on resolving the questions in #725.

PR #1838, opened 2020-11-15
protocols/relay: Implement circuit relay specification

mxinden:
This pull request implements the [libp2p circuit relay](https://github.com/libp2p/specs/tree/master/relay) specification. It is based on previous work from https://github.com/libp2p/rust-libp2p/pull/1134.

Instead of altering the `Transport` trait, the approach taken in this pull request is to wrap an existing implementation of `Transport` allowing one to:

- Intercept `dial` requests with a relayed address.

- Inject incoming relayed connections with the local node being the destination.

- Intercept `listen_on` requests pointing to a relay, ensuring to keep a constant connection to the relay, waiting for incoming requests with the local node being the destination.

More concretely one would wrap an existing `Transport` implementation as seen below, allowing the `Relay` behaviour and the `RelayTransport` to communicate via channels.

### Example

```rust
let (relay_transport, relay_behaviour) = new_transport_and_behaviour(
    RelayConfig::default(),
    MemoryTransport::default(),
);

let transport = relay_transport
    .upgrade(upgrade::Version::V1)
    .authenticate(plaintext)
    .multiplex(YamuxConfig::default())
    .boxed();

let mut swarm = Swarm::new(transport, relay_behaviour, local_peer_id);

let relay_addr = Multiaddr::from_str("/memory/1234").unwrap()
    .with(Protocol::P2p(PeerId::random().into()))
    .with(Protocol::P2pCircuit);
let dst_addr = relay_addr.clone().with(Protocol::Memory(5678));

// Listen for incoming connections via relay node (1234).
Swarm::listen_on(&mut swarm, relay_addr).unwrap();

// Dial node (5678) via relay node (1234).
Swarm::dial_addr(&mut swarm, dst_addr).unwrap();
```

### Status Quo

This pull request is ready to be reviewed and tested. See https://github.com/libp2p/rust-libp2p/pull/1838#issuecomment-782114226 for details.

Closes https://github.com/libp2p/rust-libp2p/issues/725.


dvc94ch:
Why not return the Relay from the RelayTransportWrapper directly? Do you anticipate needing to pass additional arguments like some sort of config?

dvc94ch:
So one issue is that the transport wrapper doesn't emit `NewAddress` and `AddressExpired` events for the relayed addresses when connection to a relay is established or lost. This means that the identify protocol will not report the relayed addresses and the dht will not be able to make requests via relays.

mxinden:
> Why not return the Relay from the RelayTransportWrapper directly? Do you anticipate needing to pass additional arguments like some sort of config?

Yes, I expect there to be a config of some kind for the `Relay` behaviour in the future, e.g. to configure the amount of allowed concurrent relayed connections. I would suggest delaying this discussion until we know more about the config. I agree that the current user experience of having to pass channels is not ideal.

dvc94ch:
So seems to be that the network is confused when receiving a ConnectionEstablished that doesn't match the requested PeerId, but the relay PeerId and marks the dialing attempt as failed. It's a bit tricky to write a test for Swarm::dial(peer_id) as there is no way to `add_address` without adding the request-response or kademlia behaviour, and it's quite a bit of boilerplate to create a behaviour that returns a vec of addresses on addresses_of_peer. Maybe there should be a StaticAddressBehaviour?

dvc94ch:
Nope, that isn't it either :/ looks like it works fine to find a block from behind a relay, but not to provide one. Not sure why yet.

mxinden:
This pull request is still not ready for a review, though there not that many things missing (see `TODO` lines). Stay tuned.

In the meantime, as the basic functionality is implemented, I was able to run a [libp2p-perf](https://github.com/mxinden/libp2p-perf) benchmark via a rudimentary rust relay server (see `protocols/relay/examples/relay.rs`).

```
libp2p-perf-client --> libp2p-relay-server --> libp2p-perf-server
```

```
cargo run --bin client --release -- --server-address  /ip6/::1/tcp/42419/p2p/RELAY_PEER_ID/p2p-circuit/p2p/LIBP2P_PERF_SERVER_PEER_ID
                                 
           
Interval        Transfer        Bandwidth
0 s - 10.06 s   39 MBytes       31.00 MBit/s
```

31 MBit/s on localhost (notably fully encrypted via noise) is at most a good start. I am posting this to show that it works, not to show that my implementation is performant.

mxinden:
This pull request is ready for review.


### Spec compliance

While not yet fully spec-compliant, it does implement all major features (dialing, listening, accepting, ...).


### Interoperability

I have tested interoperability with the [Golang circuit relay implementation](https://github.com/libp2p/go-libp2p-circuit/) via the basic [relay example](https://github.com/libp2p/go-libp2p-examples/tree/master/relay). More specifically I have tested `rust -> golang -> rust` and `golang -> rust -> golang`.


### Review and testing


I would very much appreciate code reviews, especially from @romanb. There is no time pressure, thus please don't feel rushed.

In addition, assuming you (@nothingismagick, @thomaseizinger, @foltik, @dvc94ch, @akhilman and @leesmet) are interested in this effort based on your emoji reactions, I would be thankful for any voluntary testers. The examples in `/protocols/relay/src/lib.rs` as well as `/protocols/relay/tests/lib.rs` should help you getting started.

romanb:
> I would very much appreciate code reviews, especially from @romanb. There is no time pressure, thus please don't feel rushed.

:+1:  I began a first review that I plan to finish tomorrow.

protocols/relay/src/transport.rs
```diff
@@ -0,0 +1,432 @@
...
+/// Allows the local node to:
+///
+/// 1. Establish relayed connections by dialing `/p2p-circuit` addresses.
+/// 2. Accept incoming relayed connections.
+/// 3. Connecting to relay nodes in order for them to listen for incoming
+///    connections for the local node.
+#[derive(Clone)]
+pub struct RelayTransportWrapper<T: Clone> {
+    to_behaviour: mpsc::Sender<TransportToBehaviourMsg>,
+    from_behaviour: Arc<Mutex<mpsc::Receiver<BehaviourToTransportMsg>>>,
```
romanb, line 66: Does it not matter which transport instance emits the `ListenerEvent::Upgrade` received through this channel? If I understand correctly, then through sharing the channel receiver in this manner, if I create a transport and clone it to "listen on" two different relay nodes, I can get events for "inbound connections" from one reported by the other? At least I would expect to get only those from a listener that actually came from the relay that I "listened on" with this particular transport instance. Do I misunderstand the implementation? It even looks like I can get events for inbound relayed connections on a listener that did not even "listen on" a relay address.

protocols/relay/src/transport.rs
```diff
@@ -0,0 +1,432 @@
...
+                src_peer_id,
+                relay_addr,
+            })) => {
+                return Poll::Ready(Some(Ok(ListenerEvent::Upgrade {
+                    upgrade: RelayedListenerUpgrade::Relayed(Some(stream)),
+                    local_addr: relay_addr.with(Protocol::P2pCircuit),
+                    remote_addr: Protocol::P2p(src_peer_id.into()).into(),
+                })));
+            }
+            Poll::Ready(None) => unimplemented!(),
```
romanb, line 316: Do I understand correctly that this can only happen when the behaviour is dropped? Since there are similar panics in the behaviour when the transport gets dropped, I wonder if this can not cause trouble, as one of them is always destined to be dropped first?

protocols/relay/src/transport.rs
```diff
@@ -0,0 +1,432 @@
...
+            Poll::Pending => {}
+        }
+
+        // TODO: This should likely only be reported once the relay connection is established. But
+        // how should one know that?
```
romanb, line 321: Sounds like one would need a dedicated channel between the behaviour and this particular listener? I'm wondering if the `to_transport` sender of such a fresh channel could not be sent to the behaviour in the `TransportToBehaviourMsg::ListenReq` for a `listen_on()`, thereby removing the `Arc<Mutex>` in the `RelayTransportWrapper` and maybe solving the problem of emitting `IncomingDstReqSuccess` to the right listener? Apologies if this is a naive suggestion, I'm still wrapping my head around the details.

protocols/relay/src/transport.rs
```diff
@@ -0,0 +1,432 @@
...
+    >;
+    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        match self.project() {
+            RelayedListenerUpgradeProj::Inner(upgrade) => match upgrade.poll(cx) {
+                Poll::Ready(Ok(out)) => return Poll::Ready(Ok(EitherOutput::First(out))),
+                Poll::Ready(Err(err)) => return Poll::Ready(Err(EitherError::A(err))),
+                Poll::Pending => {}
+            },
+            RelayedListenerUpgradeProj::Relayed(substream) => {
+                return Poll::Ready(Ok(EitherOutput::Second(substream.take().unwrap())))
```
romanb, line 351:
```suggestion
                return Poll::Ready(Ok(EitherOutput::Second(substream.take().expect("Future polled after completion"))))
```

romanb (commented):
Thanks for putting in the work. There are still some parts I didn't look at yet in detail, but I think it better to already submit some comments and questions, many of them just minor suggestions but some may be more interesting. Something I wasn't quite sure about yet from the implementation, but this does not cover the ability for "multi hop relaying" as specified under "future work" in the spec, or does it?

protocols/relay/src/behaviour.rs
```diff
@@ -0,0 +1,595 @@
...
+        poll_parameters: &mut impl PollParameters,
+    ) -> Poll<NetworkBehaviourAction<RelayHandlerIn, Self::OutEvent>> {
+        if !self.outbox_to_transport.is_empty() {
+            match self.to_transport.poll_ready(cx) {
+                Poll::Ready(Ok(())) => {
+                    self.to_transport
+                        .start_send(self.outbox_to_transport.pop().unwrap())
+                        .unwrap();
+                }
+                Poll::Ready(Err(_)) => unimplemented!(),
```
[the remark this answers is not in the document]
mxinden: This is an oversight, thanks for the catch.

`Poll::Ready(Err(mpsc::SendError))` can only happen when the `Relay` `NetworkBehaviour` is polled after the `RelayTransportWrapper` is dropped. `ExpandedSwarm` does not provide a guarantee on the drop order (e.g. `behaviour` before `network`). Though for this case I assume we can expect that `ExpandedSwarm` would not poll the `behaviour` after having dropped the `network`.

I will put a descriptive `panic` in place. Let me know in case my assumption above is incorrect.

protocols/relay/src/handler.rs
```diff
@@ -0,0 +1,550 @@
...
+/// Event that can be sent to the relay handler.
+pub enum RelayHandlerIn {
+    /// Tell the handler whether it is handling a connection used to listen for incoming relayed
+    /// connections.
+    UsedForListening(bool),
+    /// Denies a relay request sent by the node we talk to acting as a source.
+    DenyIncomingRelayReq(protocol::IncomingRelayReq),
+
+    /// Denies a destination request sent by the node we talk to.
+    DenyDstReq(PeerId, RequestId),
```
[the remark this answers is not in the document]
mxinden: Added a TODO note (yet to be pushed). Thank you for the hint. I am not quite sure whether there should be a separate mechanism for users to deny incoming destination requests and thus I am not sure whether this variant is needed in the first place. I can imagine that, as a relay, one would want to enforce limits on the number of incoming and outgoing relay connections. But as a destination node, the already existing connection limit mechanism should suffice. Happy for comments in case you have thoughts on this.

protocols/relay/src/behaviour.rs
```diff
@@ -0,0 +1,595 @@
...
+                        });
+                }
+            }
+            // Remote wants us to become a destination.
+            RelayHandlerEvent::IncomingDstReq(src, request_id) => {
+                let send_back = RelayHandlerIn::AcceptDstReq(src, request_id);
+                self.outbox_to_swarm
+                    .push_back(NetworkBehaviourAction::NotifyHandler {
+                        peer_id: event_source,
+                        handler: NotifyHandler::Any,
```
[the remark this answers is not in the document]
mxinden: Oh, good catch. Thanks.

protocols/relay/src/handler.rs
```diff
@@ -0,0 +1,550 @@
...
+    /// [`AcceptDstReq`](RelayHandlerIn::AcceptDstReq) or
+    /// [`DenyDstReq`](RelayHandlerIn::DenyDstReq).
+    IncomingDstReq(PeerId, RequestId),
+
+    /// A `RelayReq` that has previously been sent has been accepted by the remote. Contains
+    /// a substream that communicates with the requested destination.
+    ///
+    /// > **Note**: There is no proof that we are actually communicating with the destination. An
+    /// >           encryption handshake has to be performed on top of this substream in order to
+    /// >           avoid MITM attacks.
```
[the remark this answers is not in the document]
mxinden: First of, this note is from previous implementation attempts (https://github.com/libp2p/rust-libp2p/pull/1134/files#diff-3e891c993e81968e6593c3e9dd2806b688f4b0b30d30c246ac63f016a86e2df2R87), which this work is based on, thus I might have a wrong understanding myself.

> The relay spec says "Relayed connections are end-to-end encrypted just like regular connections." but I don't see how that is achieved with this protocol, as the relay node can read all traffic.

The relay protocol itself does not secure (authenticate / encrypt / ...) relayed connections. But given that relayed connections are treated just like any other connection by the source and destination node, they too go through the configured encryption and multiplexing upgrade.

> Well, by nature of this protocol, one usually trusts the relay node

With the above in mind and under the circumstance that the source is aware of the destinations `PeerId` (i.e. public key), not just its `Multiaddr`, I would claim that the source does not need to trust the relay.

@romanb does the above make sense? If so, how could we improve the comment to avoid future confusions? I might be missing something here.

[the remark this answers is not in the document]
romanb: > But given that relayed connections are treated just like any other connection by the source and destination node, they too go through the configured encryption and multiplexing upgrade.

It may just have been a misunderstanding on my part. So, to clarify, a "relayed connection" between A and B through R consists of two substreams, one between A and R and one between R and B. These substreams exist on fully upgraded, authenticated, multiplexed (TCP) connections between A and R, as well as R and B, respectively. These two substreams form the relayed "connection", and they are the only substreams on the respective connections. This relayed "connection" comprising these two substreams then goes through all transport upgrades, authentication and multiplexing. So we have "nested" encryption and multiplexing. Is that right? Then indeed the relayed "connection" would be end-to-end encrypted (and authenticated) and I simply did not consider the second, nested transport upgrade happening through the negotiated substreams.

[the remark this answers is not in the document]
mxinden: > Is that right?

Yes, all of the above is correct, except:

> These substreams exist on fully upgraded, **authenticated**, multiplexed (TCP) connections between A and R, as well as R and B, respectively.

I guess one could as well use `libp2p-plaintext` for (non-) _authentication_ for the outer (TCP) connection in case leaking meta-data is ok. Though I have not seen any mention of this idea before and in addition `libp2p-plaintext` is explicitly designed for testing only.

> These two substreams form the relayed "connection", and **they are the only substreams on the respective connections**.

I am not aware of the constraint that the two TCP connections between A-R and R-B can at most carry a single relay substream. This implementation would potentially re-use the TCP connections in case additional relay requests could make use of either or both TCP connections. One could enforce the single TCP connection per relay connection constrained e.g. to minimize the impact of nested multiplexing. Enforcing this constrained would be difficult for destination nodes behind NATs, as a relay node would not be able to establish a new TCP connection for an incoming relay request to the destination. In addition the `Relay` behaviour might be used in conjunction with other `NetworkBehaviour`s, which themselves could potentially request substreams for the given connection.

Let me know whether the above makes sense @romanb.

mxinden:
_Very much appreciate the help here, @romanb! Sorry for being rather slow to reply to each of your comment._

protocols/relay/src/protocol.rs
```diff
@@ -0,0 +1,181 @@
...
+use libp2p_swarm::NegotiatedSubstream;
+use smallvec::SmallVec;
+use std::io::{Error, IoSlice};
+use std::pin::Pin;
+use std::task::{Context, Poll};
+use std::{convert::TryFrom, error, fmt};
+
+/// Any message received on the wire whose length is superior to that will be refused and will
+/// trigger an error.
+const MAX_ACCEPTED_MESSAGE_LEN: usize = 1024;
```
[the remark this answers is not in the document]
mxinden: I don't know to be honest. I adopted this constant from the previous implementation attempt. Off the top of my head enforcing a limit seems reasonable. I will have to put more thoughts into what a good value would be.

protocols/relay/src/transport.rs
mxinden: > Do I misunderstand the implementation? It even looks like I can get events for inbound relayed connections on a listener that did not even "listen on" a relay address.

Your understanding above is correct.

One needs to support two kinds of incoming relayed connections:
1. Incoming relayed connections relayed via a known relay node, for which the local node previously explicitly called `Swarm::listen_on(&mut swarm, RELAY_ADDR + CIRCUIT_RELAY_POSTFIX).unwrap();` i.e. explicitly asked the relay node to act as a relay for the local node.
2. Incoming relayed connections via some (unspecified) relay node, via which the local node is not explicitly listening for incoming connections.

While I would consider use-case (1) the most important one, e.g. to overcome NATs, I still think the use-case (2) is relevant. Say e.g. that node _A_ and _B_ want to connect. Both are publicly routable, but don't have a encryption protocol in common. In case there is a relay node R which has an encryption protocol in common with both _A_ and _B_, the relay node can act as a mediator between the two nodes _A_ and _B_.

Thus far, I did not come up with a clean interface which would support both (1) and (2). As you correctly noticed, for now I simply (/naively) pass an incoming relayed connection to the first listener polling the channel to the `Relay` behaviour, ignoring whether the listener is actually listening on a relay connection. I agree that this is not ideal.

Properly supporting use-case (1) and always passing an incoming relayed connection to the listener listening via that specific relay node should be easily doable via additional channels routing an incoming relayed connection to the correct listener.

I am not yet sure how to support use-case (2). First thing that comes to my mind is to ask users, that want to receive incoming relayed connections via any relay node, to register a _catch-all_ listener, e.g. via `Swarm::listen_on(&mut swarm, "/p2p-circuit".try_into().unwrap()).unwrap();` which would then yield all incoming relayed connections that are relayed via a relay node via which the local node is not explicitly listening.

Does the above make sense? If so, can you think of a better interface to support both use-cases?

protocols/relay/src/transport.rs
mxinden: > Apologies if this is a naive suggestion, I'm still wrapping my head around the details.

No, I agree that this is the way to go, only issue I have today is how to support use-case (2) (an incoming relayed connection via a relay node one is not explicitly listening via) (see https://github.com/libp2p/rust-libp2p/pull/1838#discussion_r586447824).

protocols/relay/src/transport.rs
romanb: > First thing that comes to my mind is to ask users, that want to receive incoming relayed connections via any relay node, to register a catch-all listener, e.g. via Swarm::listen_on(&mut swarm, "/p2p-circuit".try_into().unwrap()).unwrap(); which would then yield all incoming relayed connections that are relayed via a relay node via which the local node is not explicitly listening.

That sounds good to me!

protocols/relay/src/handler.rs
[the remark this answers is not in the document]
romanb: To avoid confusion, I would probably remove this constructor for now, making it clear in the code that all such requests are accepted.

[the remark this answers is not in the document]
mxinden: With the reworked listener logic, yielding an incoming relayed connection via the correct listener (see https://github.com/libp2p/rust-libp2p/pull/1838#discussion_r582262518), this code path will no longer be unused. When receiving an incoming relayed connection but not having a listener to pass the request to, the request is denied instead.

protocols/relay/src/transport.rs
mxinden: Absolutely correct. This is problematic, especially with the reworked listener logic (https://github.com/libp2p/rust-libp2p/pull/1838#discussion_r582262518). I will change the code to terminate the listener instead. Thanks for highlighting this.

mxinden:
> Something I wasn't quite sure about yet from the implementation: This does not cover the ability for "multi hop relaying" as specified under "future work" in the spec, or does it?

Correct. It does not. There is also a proposal for circuit relay v2 (https://github.com/libp2p/go-libp2p-circuit/pull/125) which would be worth considering in the future.

mxinden:
I am reasonably sure I addressed all the comments above, most notably the re-work of the listener logic (https://github.com/libp2p/rust-libp2p/pull/1838#discussion_r582262518).

In case you have time for another review @romanb, that would be terrific.

protocols/relay/src/behaviour.rs
```diff
@@ -0,0 +1,769 @@
...
+                        self.listeners.remove(&relay_peer_id);
+                    }
+                    Poll::Pending => {}
+                },
+                None => {
+                    // No listener to send request to, thus dropping it. This
+                    // case should be rare, as we check whether we have a
+                    // listener before accepting an incoming destination
+                    // request.
+                    self.outbox_to_listeners.pop_front();
```
romanb, line 587:
```suggestion
                    let event = self.outbox_to_listeners.pop_front();
                    log::trace!("Dropping event for unknown listener: {:?}", event);
```

protocols/relay/src/handler.rs
```diff
@@ -0,0 +1,755 @@
...
+                    ));
+            }
+            RelayOutboundOpenInfo::Destination {
+                src_connection_id,
+                incoming_relay_req,
+                ..
+            } => {
+                let err_code = match error {
+                    ProtocolsHandlerUpgrErr::Timeout => {
+                        self.pending_error = Some(ProtocolsHandlerUpgrErr::Timeout);
```
romanb, line 513: Do we really want to terminate the connection on a single relay substream negotiation timeout? Other protocols may use the connection (successfully) as well, right?

romanb (approved):
Looks good to me! Glad to see the `Arc<Mutex>` gone. I only left two more minor comments.

protocols/relay/src/handler.rs
mxinden: If I understand correctly, given that `libp2p-relay` uses the default timeout, this will only happen after 10 seconds. I would deem a node that is not able to respond to the light-weight circuit relay requests within 10 seconds as either misbehaving or overloaded. For the former, disconnecting seems to be the way to go. For the latter I would say disconnecting is beneficial for both sides. The local node might be able to succeed in whatever it is up to via another relay or destination node. In addition I would guess other protocols running on the same connection are not making progress either if the light-weight circuit relay negotiation does not make progress. The overloaded remote node would receive less traffic (both through the relay protocol and any other protocol out there) and thus become less overloaded overall

The above said, I don't know what the correct behaviour is and I don't know whether it should be consistent across protocols. E.g. `libp2p-request-response` does not terminate the whole connection on `Timeout`, which makes sense to me as the payloads being exchanged could potentially be very large.

https://github.com/libp2p/rust-libp2p/blob/f48bb15fa7749187570d4e092dbc2346cee188a0/protocols/request-response/src/handler.rs#L252-L260

With my reasoning above in mind, what do you think @romanb we should do in case of a `ProtocolsHandlerUpgrErr::Timeout`?

mxinden:
I want to draw attention to the most recent commit https://github.com/libp2p/rust-libp2p/pull/1838/commits/7956f0b57550992d210c68482dac8a4ade168ce9 which is not based on any of the above review comments.

> protocols/relay: Disable active relay behaviour by default
> 
> Add `actively_connect_to_dst_nodes` configuration option. Configures
> whether to actively establish an outgoing connection to a destination
> node, when being asked by a source node to relay a connection to said
> destination node.
> 
> For security reasons this behaviour is disabled by default, thus a relay
> node will not actively establish an outgoing connection to a destination
> node in case it is not yet connected to said destination node. Instead a
> destination node should establish a connection to a relay node before
> advertising their relayed address via that relay node to a source node.

For comparison, [here is the same configuration option of the Golang implementation](https://github.com/libp2p/go-libp2p-circuit/blob/c0edc67235d7852fdb9ea429c4e2e13dde8dc3bc/relay.go#L67-L70).

mxinden:
_Continuing on the discussion in https://github.com/libp2p/rust-libp2p/pull/1838#discussion_r591641777 above._

> > In addition I would guess other protocols running on the same connection are not making progress either if the light-weight circuit relay negotiation does not make progress.
> 
> I think conceptually, different protocols using different substreams on the same connection should be considered independently, while necessarily sharing the same network resource. Making it a requirement that all protocols used on a connection work well (e.g. without timeouts) all the time seems a bit problematic. Especially when it comes to timeouts on (outbound) substreams, in my mind, these should be reported to client code which can then a) decide whether to retry and if so how often and with what kind of back-off strategy and b) whether the entire connection should be closed after x timeouts. The concrete application knows what protocols are used on a particular connection, whereas a single protocol does not know with which others it must share the connections. These were also the considerations for `libp2p-request-response`. Making the fixed choice within a particular protocol that the connection is killed if a particular substream protocol upgrade does not complete within 10 seconds seems very rigid. It is probably fine in this particular instance, so I'm not opposed, but in general I think any errors other than protocol violations, (outbound) timeouts in particular, should just be reported on the API and left to client code to handle.

The above reasoning makes sense to me. Thank you for the detailed write-up.

**In general** this seems like something worth striving for in a consistent manner across `ProtocolHandler` implementations. E.g. `libp2p-gossipsub` currently [terminates the connection](https://github.com/libp2p/rust-libp2p/blob/master/protocols/gossipsub/src/handler.rs#L279) whereas, as mentioned above, `libp2p-request-response` doesn't.

**In the specific** case here https://github.com/libp2p/rust-libp2p/pull/1838/commits/d55bd99ca775f49025add4814bfa9564b1d51766 makes `libp2p-relay` not close connections on `dial` and `incoming` upgrade `Timeout` errors. Instead connections would be eventually closed through the keep alive mechanism, in case both `libp2p-relay` as well as all other protocols have no more use for the connection. In the future one could explore actively setting `KeepAlive::No` on `Timeout` errors, only switching back to `KeepAlive::Yes` or `KeepAlive::Until` in case of any future success. That would speed up the connection garbage collection process.

mxinden:
> Glad to see the `Arc<Mutex>` gone.

Very much agreed!

mxinden:
After 133 commits this is finally merged. :tada:

Thanks goes to @tomaka for the initial version (https://github.com/libp2p/rust-libp2p/pull/1134), @romanb for the reviews and @dvc94ch for the initial testing.

I will publish a release of `libp2p-relay` `v0.1.0` soon.

PR #1842, opened 2020-11-16
[swarm] Configurable and "infinite" scores for external addresses.

romanb:
## Motivation
It is currently not possible to add external addresses to a `Swarm` that are retained forever. All added or observed external addresses are added to an ordered list with a finite score (`1`) for every new addition and thus subject to being purged eventually. It is desirable to be able to add addresses to this collection that are retained until explicitly removed and always sorted at the beginning, not influencing the existing semantics for the additional finite scores and expiring addresses. It also seems desirable to let the score of a particular address report / observation be chosen freely, instead of being fixed to `1`. The motivation relates to https://github.com/paritytech/substrate/issues/7518.

To that end, this PR extends the address score arithmetic used for external addresses with an infinite cardinal, permitting addresses with such a score to be retained "forever" or until explicitly removed, thereby exposing address scores on the public API for more insight and configurability.

## Implementation

  * The score of an (external) address is now not just a `u32` but an `AddressScore`, i.e. either `AddressScore::Finite(u32)` or `AddressScore::Infinite`.
  * Only finitely scored addresses compete among each other for retention in the ordered address list, the same way as before. "Infinitely scored" addresses are always retained and sorted at the beginning, i.e. higher than any finite score.
  * `Swarm::add_external_address` now takes both a `Multiaddr` and an `AddressScore`.
  * Analogously to the above, `NetworkBehaviourAction::ReportObservedAddr` now also takes an `AddressScore` in addition to the `Multiaddr`. The `Identify` behaviour reports all addresses with a finite score of `1`, i.e. the same semantics as before.
  * A new method `Swarm::remove_external_address` for explicit removal, regardless of score. This allows for explicit curation of the external address list.
  * Iterating over the external addresses now also provides access to the score via iterating over `AddressRecord`s.
  * The external addresses given to the `NetworkBehaviour::poll` are also accompanied by their score (again via `AddressRecord`s).

### DIFF
--- a/swarm/src/registry.rs
+++ b/swarm/src/registry.rs
@@ -62,10 +62,10 @@ pub struct Addresses {
 
 /// An record in a prioritised list of addresses.
 #[derive(Clone, Debug, PartialEq, Eq)]
+#[non_exhaustive]
 pub struct AddressRecord {
     pub addr: Multiaddr,
     pub score: AddressScore,
-    _priv: (),
 }
 
 /// A report tracked for a finitely scored address.
--- a/swarm/src/registry.rs
+++ b/swarm/src/registry.rs
@@ -20,29 +20,143 @@
 
 use libp2p_core::Multiaddr;
 use smallvec::SmallVec;
-use std::{collections::VecDeque, num::NonZeroUsize};
+use std::{collections::VecDeque, cmp::Ordering, num::NonZeroUsize};
+use std::ops::{Add, Sub};
 
-/// Hold a ranked collection of [`Multiaddr`] values.
+/// A ranked collection of [`Multiaddr`] values.
 ///
-/// Every address has an associated score and iterating over addresses will return them
-/// in order from highest to lowest. When reaching the limit, addresses with the lowest
-/// score will be dropped first.
+/// Every address has an associated [score](`AddressScore`) and iterating
+/// over the addresses will return them in order from highest to lowest score.
+///
+/// In addition to the currently held addresses and their score, the collection
+/// keeps track of a limited history of the most-recently added addresses.
+/// This history determines how address scores are reduced over time as old
+/// scores expire in the context of new addresses being added:
+///
+///   * An address's score is increased by a given amount whenever it is
+///     [(re-)added](Addresses::add) to the collection.
+///   * An address's score is decreased by the same amount used when it
+///     was added when the least-recently seen addition is (as per the
+///     limited history) for this address in the context of [`Addresses::add`].
+///   * If an address's score reaches 0 in the context of [`Addresses::add`],
+///     it is removed from the collection.
+///
 #[derive(Debug, Clone)]
 pub struct Addresses {
-    /// The ranked sequence of addresses.
-    registry: SmallVec<[Record; 8]>,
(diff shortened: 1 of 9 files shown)


tomaka:
Thanks for tackling this!

In my opinion, this external addresses scoring system isn't great. It is a bit opaque, and multiple `ReportObservedAddr` events reported by multiple different `NetworkBehaviour`s will compete with each other without necessarily being aware of it.

For example, the `identity` behaviour will generate a `ReportObservedAddr` every 60 seconds. If I add another behaviour that reports a `ReportObservedAddr` every 30 seconds, this other behaviour will implicitly have double the weight of the one generated by `identify`.

But if we go under the constraint that we keep this scoring system, which is the best short-term solution, then this PR does make sense.

swarm/src/registry.rs
```diff
@@ -20,29 +20,143 @@
...
 
-// An address record associates a score to a Multiaddr.
+/// An record in a prioritised list of addresses.
 #[derive(Clone, Debug, PartialEq, Eq)]
-struct Record {
+pub struct AddressRecord {
+    pub addr: Multiaddr,
+    pub score: AddressScore,
+    _priv: (),
+}
```
tomaka, line 69:
```suggestion
#[non_exhaustive]
pub struct AddressRecord {
    pub addr: Multiaddr,
    pub score: AddressScore,
}
```

tomaka (approved):

romanb:
> In my opinion, this external addresses scoring system isn't great. It is a bit opaque, and multiple ReportObservedAddr events reported by multiple different NetworkBehaviours will compete with each other without necessarily being aware of it.

I tried to improve the documentation, but I'm not sure what you mean by different behaviours competing. If they both issue a report for the same address, the effect is cumulative. For different addresses the ones that are reported with the higher cumulative score (e.g. either more frequently or with higher scores) have the higher retention and priority. That seems to make sense to me, even under composition of behaviors. Of course behaviours / protocols should make an document to document how they go about reporting addresses, like I did for identify, to better reason about the composition. Nevertheless I'm open to alternative proposals on how to handle external addresses, i.e. feel free to open a new issue describing the problems with the current approach and possible alternatives, if you already have something in mind.

> But if we go under the constraint that we keep this scoring system [..]

That was my current understanding at least, so I only looked into extending it, rather than replacing it. Unless we already have a clear idea of how a replacement should look like, I would also suggest to go with this for now.

tomaka:
> Of course behaviours / protocols should make an attempt at documenting how they go about reporting addresses, like I did for identify, to better reason about the composition.

That's indeed the only solution to this problem. When I say "It is very difficult to understand the behaviour of external addresses without knowing how the entirety of libp2p works.", what I mean is that you need to look at the precise documentation of every `NetworkBehaviour` that you use.

It would be preferable to me, although probably difficult to achieve, that adding a network behaviour to the stack doesn't have any negative consequence on the others.

> That was my current understanding at least, so I only looked into extending it, rather than replacing it. Unless we already have a clear idea of how a replacement should look like, I would also suggest to go with this for now.

I didn't mention anything about keeping or replacing the current system because I also don't know how we could replace this system.

mxinden (approved):

PR #1845, opened 2020-11-17
Tcp

dvc94ch:
Superceeds #1789 .

Simplifies the tcp transport by using async-io instead of macros. This also helps people who can't update to the latest libp2p because it uses tokio 0.3 . Address changes are handled using `if-watch`, the same crate used in #1830, instead of polling "randomly". It also supports reusing ports for outgoing connections and makes the address_translation the resposibility of the transport. While the `if_watch` module is currently in the tcp crate, when adding the quic transport it might make sense to move it into a shared transport utility crate.


transports/tcp/src/if_task.rs
```diff
@@ -0,0 +1,151 @@
...
+        // channel is unbounded so that `register_listener` doesn't need to be async.
+        let (tx, rx) = mpsc::unbounded();
+        let task = Box::pin(Task::new(rx).await?.spawn());
+        #[cfg(feature = "async-std")]
+        async_std::task::spawn(task);
+        #[cfg(all(not(feature = "async-std"), feature = "tokio"))]
+        tokio::task::spawn(task);
+        #[cfg(all(not(feature = "async-std"), not(feature = "tokio")))]
+        std::thread::spawn(move || futures::executor::block_on(task));
+        Ok(Self { tx })
```
dvc94ch, line 37: This is a bit of a hack, not sure how to spawn tasks in an executor independent way. Because we are sharing a background task between listeners for polling interface changes, `TcpConfig::new()` returns a future, which is also not great.

protocols/request-response/tests/ping.rs
```diff
@@ -64,22 +64,19 @@ fn ping_protocol() {
...
-        while let Some(_) = swarm1.next().now_or_never() {}
-
-        let l = Swarm::listeners(&swarm1).next().unwrap();
-        tx.send(l.clone()).await.unwrap();
-
         loop {
-            match swarm1.next().await {
-                RequestResponseEvent::Message {
+            match swarm1.next_event().await {
+                SwarmEvent::NewListenAddr(addr) => tx.send(addr).await.unwrap(),
```
dvc94ch, line 69: since we don't use get-if-addrs which I think blocks the thread, it's not guaranteed that the event is emitted immediately

core/src/transport.rs
```diff
@@ -128,6 +128,9 @@ pub trait Transport {
     where
         Self: Sized;
 
+    /// Perform transport specific multiaddr translation.
+    fn address_translation(&self, _server: &Multiaddr, observed: &Multiaddr) -> Option<Multiaddr>;
+
```
dvc94ch, line 133: This is what was discussed with @romanb

transports/tcp/src/if_task.rs
dvc94ch: Fixed using agnostik, still waiting for https://github.com/bastion-rs/agnostik/pull/28

dvc94ch: giving up on this one, there is no adequate solution. new needs to take a spawner as an argument :/

mxinden (commented):
I haven't fully followed the whole effort, thus please bear with me.

As far as I can tell this pull request does 3 things:

- Move `libp2p-tcp` to `async-io`.

- Watch for and react to address changes via `if-watch`.

- Allow transport specific address translation.

If I am not mistaken these changes are not directly interdependent and could be proposed as separate pull requests, correct? If that is the case, splitting this pull request would help me as a reviewer and would overall simplify the process. 

What do others think?

transports/tcp/src/lib.rs
```diff
@@ -90,297 +96,198 @@ impl $tcp_config {
...
-                port,
-                addrs,
-                pending,
-                config: cfg
-            };
+    fn do_listen(self, socket_addr: SocketAddr) -> io::Result<TcpListenStream> {
+        let socket = self.create_socket(&socket_addr)?;
+        socket.bind(&socket_addr.into())?;
+        socket.listen(self.backlog as _)?;
+        let listener = Async::new(socket.into_tcp_listener())?;
```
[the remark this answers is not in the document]
dvc94ch: It is required because futures from async-net take &self, making them impossible to box without unsafe code. Also I don't really see why another library is required, when async-io is so easy to use.

romanb:
For what it is worth, I'm currently experimenting with this branch, mainly to explore options without a background task and thread, and without the async contructor for `TcpConfig`, as these aspects bother me a bit. If I manage to come up with something, I will report back here for further discussion.

dvc94ch:
I think there is a good reason for spawning a background task and it's a shame that there isn't much interest in a "simple" solution (https://github.com/async-rs/async-std/issues/912).

dvc94ch:
@romanb I assume you gave up on finding a solution?

mdns has an async constructor too, so it really doesn't hurt at this point to make tcp also be an async constructor. I think a background task/thread is required in this case, the only alternative would be for each listener to open it's own netlink socket, which would be wasteful (instead of sharing a netlink socket per transport).

romanb:
> @romanb I assume you gave up on finding a solution?

No, I haven't given up, I'm quite happy with the approach I'm currently pursuing and should have something to share by the end of next Monday.

> I think a background task/thread is required in this case, the only alternative would be for each listener to open it's own netlink socket, which would be wasteful (instead of sharing a netlink socket per transport).

It's one background task & thread per `TcpConfig::new()` though. I just don't think `TcpConfig::new()` should be such an expensive, thread-spawning operation, even if the standard use case involves just creating one of these. I currently think it is indeed sufficient for a `TcpListenStream` to directly poll its own `IfWatcher`, especially because I think an `IfWatcher` is only needed if the `TcpListenStream` listens on a wildcard address, not if the listener is listening on a single fixed address. For in the latter case, if the interface of that address goes down, the `TcpListenStream` will error on `accept()` pretty much immediately without the need for notifications. So in practice there will usually be at most one (or maybe two for both `Ipv4` and `Ipv6`) `TcpListenStream`s whose listen sockets are listening on all interfaces and require `IfWatcher` notifications. Furthermore, since in production configurations it is not even advisable to listen on wildcard addresses, there are neither extra background tasks nor threads nor `IfWatcher`s used when you just bind a node to one or more specific addresses to listen on, as I would want and expect. In any case, this is part of what I will propose next week. If I hit road blocks that cause me to give up, I will certainly post here, too.

romanb:
I opened my proposal as https://github.com/libp2p/rust-libp2p/pull/1887, which builds on this branch.

dvc94ch:
Closing in favor of #1887

Issue #1839, opened 2020-11-16
Add a way to limit the number of total inbound connections


tomaka:
The `SwarmBuilder` has a way to limit the number of connections per peer, and a way to limit the number of handshaking connections, but no way to limit the total number of incoming connections.

At the moment, someone can pre-generate thousands of PeerIds, then open thousands of TCP connections to a single node, and completely clog its file descriptors.

PR #1848, opened 2020-11-20
[core/swarm] Refactor and extend configurable connection limits.

romanb:
The initial motivation for the work in this PR is https://github.com/libp2p/rust-libp2p/issues/1839. However, the PR goes a bit further than to just add a single counter to have a more streamlined API. Thus, in order to better track different connection counts and permit configurable limits for these counts and make these available for inspection in an efficient manner (i.e. without linear counting operations), I introduced dedicated connection counters via a `ConnectionCounters` structure maintained by the internal connection pool but exposed on the API via the `NetworkInfo`. All connection states that are counted in this way can also have effective configurable limits.

All limits are now configured through a `ConnectionLimits` structure. The new configurable limits are

  * `ConnectionLimits::max_established_incoming` for limiting the number of concurrent established connections that were incoming.
  * `ConnectionLimits::max_established_outgoing` for limiting the number of concurrent established connections that were outgoing.

A limit that has been removed because tracking the related counts efficiently would be a bit laborious and I don't think it is particularly useful is

  * `NetworkConfig::set_outgoing_per_peer_limit()`

Note that there are still the pre-existing `ConnectionLimits::max_incoming` and `ConnectionLimits::max_outgoing` which correspond to the two new configuration options above, just for pending connections. I also took the liberty to change the connection limits to use `u32` instead of `usize` for their types, which seems more appropriate.

### Fixed panic in the `Peer` API

The `Network::peer()` API allows to iterate mutably over the ongoing connection attempts via `DialingPeer::attempts()`, where each `DialingAttempt` can be `abort()`ed during iteration. However, the iterator did not take into account the changes to the internal iteration offsets when elements are removed. This was discovered while extending the connection limit tests a bit and has been fixed now as a result.

### DIFF
--- a/core/CHANGELOG.md
+++ b/core/CHANGELOG.md
@@ -1,6 +1,7 @@
 # 0.25.0 [unreleased]
 
 - The `NetworkConfig` API is now a builder that moves `self`.
+  [PR 1848](https://github.com/libp2p/rust-libp2p/pull/1848/).
 
 - New configurable connection limits for established connections and
   dedicated connection counters. Removed the connection limit dedicated
--- a/core/CHANGELOG.md
+++ b/core/CHANGELOG.md
@@ -6,6 +6,7 @@
   dedicated connection counters. Removed the connection limit dedicated
   to outgoing pending connection _per peer_. Connection limits are now
   represented by `u32` intead of `usize` types.
+  [PR 1848](https://github.com/libp2p/rust-libp2p/pull/1848/).
 
 - Update `multihash`.
 
--- a/core/CHANGELOG.md
+++ b/core/CHANGELOG.md
@@ -1,5 +1,8 @@
 # 0.25.0 [unreleased]
 
+- New configurable connection limits for established connections and
+  dedicated connection counters.
+
 - Update `multihash`.
 
 - Update `multistream-select`.
--- a/core/CHANGELOG.md
+++ b/core/CHANGELOG.md
@@ -1,7 +1,9 @@
 # 0.25.0 [unreleased]
 
 - New configurable connection limits for established connections and
-  dedicated connection counters.
+  dedicated connection counters. Removed the connection limit dedicated
+  to outgoing pending connection _per peer_. Connection limits are now
+  represented by `u32` intead of `usize` types.
 
 - Update `multihash`.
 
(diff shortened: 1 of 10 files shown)


mxinden (commented):
Good catch on the iterator state mismatch!

core/src/connection/pool.rs
```diff
@@ -894,62 +887,169 @@ where
...
-    pub max_outgoing_per_peer: Option<usize>,
+/// Network connection information.
+#[derive(Debug, Clone)]
+pub struct ConnectionCounters {
+    /// The effective connection limits.
+    limits: ConnectionLimits,
+    /// The current number of incoming connections.
+    incoming: u32,
+    /// The current number of outgoing connections.
+    outgoing: u32,
```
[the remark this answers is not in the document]
romanb: Yes, sounds fine to me.

tomaka (approved):
Looks good to me with the same remarks as Max on the naming.

romanb:
I incorporated the naming suggestion(s) in https://github.com/libp2p/rust-libp2p/pull/1848/commits/4e447e99a0ec9f740732a6935a639b7c57af23e6 together with a bit refinement for the configuration APIs:

  * `ConnectionLimits` now has private fields for future extensions and offers a builder API that moves `self`.
  * `NetworkConfig` now has a builder API that moves `self`.

The `SwarmBuilder` already used this style, so there is some gain in consistency. Let me know if that last commit looks acceptable.

core/CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # 0.25.0 [unreleased]
 
+- The `NetworkConfig` API is now a builder that moves `self`.
```
mxinden, line 3:
```suggestion
- The `NetworkConfig` API is now a builder that moves `self`.
  [PR 1848](https://github.com/libp2p/rust-libp2p/pull/1848/).
```

core/CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
...
+
+- New configurable connection limits for established connections and
+  dedicated connection counters. Removed the connection limit dedicated
+  to outgoing pending connection _per peer_. Connection limits are now
+  represented by `u32` intead of `usize` types.
```
mxinden, line 8:
```suggestion
  represented by `u32` intead of `usize` types.
  [PR 1848](https://github.com/libp2p/rust-libp2p/pull/1848/).
```

mxinden (approved):
Both the connection limiting as well as the general API improvements look good to me.

PR #1850, opened 2020-11-22
*: Apply clippy suggestions and enable clippy on CI

mxinden:
This pull request fixes _deny_ level and some _warning_ level [clippy](https://github.com/rust-lang/rust-clippy) lints and enables clippy to run on each pull request and each commit on `master`.

In my eyes clippy is a helpful tool. Given that (for now) it is set to fail on _deny_ level lints only, I doubt it will be too annoying.

Let me know what you think.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -109,6 +109,42 @@ jobs:
     - name: Check rustdoc links
       run: RUSTDOCFLAGS="--deny broken_intra_doc_links" cargo doc --verbose --workspace --no-deps --document-private-items
 
+  check-clippy:
+    runs-on: ubuntu-latest
+    steps:
+
+    - name: Cancel Previous Runs
+      uses: styfle/cancel-workflow-action@0.6.0
+      with:
+        access_token: ${{ github.token }}
+
+    - uses: actions/checkout@v2
+
+    - uses: actions-rs/toolchain@v1
+      with:
+        profile: minimal
+        toolchain: stable
+        override: true
+
+    - run: rustup component add clippy
+
+    - name: Cache CARGO_HOME
+      uses: actions/cache@v2
+      with:
+        path: ~/.cargo
+        key: cargo-home-${{ hashFiles('Cargo.toml') }}
+
+    - name: Cache cargo build
+      uses: actions/cache@v2
+      with:
+        path: target
+        key: cargo-build-target-${{ hashFiles('Cargo.toml') }}
+
+    - uses: actions-rs/cargo@v1
+      with:
+        command: clippy
+        args: -- -W clippy::mutable_key_type
+
   integration-test:
     name: Integration tests
     runs-on: ubuntu-latest
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -125,8 +125,7 @@ jobs:
         profile: minimal
         toolchain: stable
         override: true
-
-    - run: rustup component add clippy
+        components: clippy
 
     - name: Cache CARGO_HOME
       uses: actions/cache@v2
@@ -140,7 +139,8 @@ jobs:
         path: target
         key: cargo-build-target-${{ hashFiles('Cargo.toml') }}
 
-    - uses: actions-rs/cargo@v1
+    - name: Run cargo clippy
+      uses: actions-rs/cargo@v1
       with:
         command: clippy
         args: -- -A clippy::mutable_key_type -A clippy::type_complexity
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -143,7 +143,7 @@ jobs:
     - uses: actions-rs/cargo@v1
       with:
         command: clippy
-        args: -- -W clippy::mutable_key_type
+        args: -- -A clippy::mutable_key_type -A clippy::type_complexity
 
   integration-test:
     name: Integration tests
--- a/misc/multiaddr/src/lib.rs
+++ b/misc/multiaddr/src/lib.rs
@@ -40,6 +40,7 @@ static_assertions::const_assert! {
 }
 
 /// Representation of a Multiaddr.
+#[allow(clippy::rc_buffer)]
 #[derive(PartialEq, Eq, PartialOrd, Ord, Clone, Hash)]
 pub struct Multiaddr { bytes: Arc<Vec<u8>> }
 
--- a/misc/multiaddr/src/lib.rs
+++ b/misc/multiaddr/src/lib.rs
@@ -59,6 +59,11 @@ impl Multiaddr {
         self.bytes.len()
     }
 
+    /// Returns true if the length of this multiaddress is 0.
+    pub fn is_empty(&self) -> bool {
+        self.bytes.len() == 0
+    }
+
     /// Return a copy of this [`Multiaddr`]'s byte representation.
     pub fn to_vec(&self) -> Vec<u8> {
         Vec::from(&self.bytes[..])
--- a/protocols/kad/src/record.rs
+++ b/protocols/kad/src/record.rs
@@ -130,7 +130,7 @@ impl PartialEq for ProviderRecord {
     fn eq(&self, other: &Self) -> bool {
         // In the [`Hash`] implementation of [`ProviderRecord`] all but the
         // `key` and `provider` fields are ignored. The following property has
-        // to hold for all inputs: `k1 == k2 ⇒ hash(k1) == hash(k2)`. Thereby
+        // to hold for all inputs: `k1 == k2 <=> hash(k1) == hash(k2)`. Thereby
         // the [`PartialEq`] implementation has to ignore `key` and `provider`
         // as well.
         //
--- a/protocols/kad/src/record.rs
+++ b/protocols/kad/src/record.rs
@@ -103,6 +103,10 @@ impl Record {
 
 /// A record stored in the DHT whose value is the ID of a peer
 /// who can provide the value on-demand.
+///
+/// Note: Two [`ProviderRecord`]s as well as their corresponding hashes are
+/// equal iff their `key` and `provider` fields are equal. See the [`Hash`] and
+/// [`PartialEq`] implementations.
 #[derive(Clone, Debug)]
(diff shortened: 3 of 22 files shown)


protocols/kad/src/record.rs
```diff
@@ -122,6 +122,21 @@ impl Hash for ProviderRecord {
     }
 }
 
+impl PartialEq for ProviderRecord {
```
mxinden, line 125: Drawing attention to this non-trivial change.

As far as I can tell `libp2p-kad` does not depend on `ProviderRecord` equality when it comes to the `expires`, or `addresses` field.

In case this looks reasonable to you, let me know in case you would like a changelog entry.

.github/workflows/ci.yml
```diff
@@ -109,6 +109,42 @@ jobs:
     - name: Check rustdoc links
       run: RUSTDOCFLAGS="--deny broken_intra_doc_links" cargo doc --verbose --workspace --no-deps --document-private-items
 
+  check-clippy:
```
mxinden, line 112: @TriplEight does this look sane to you? Did I get the `actions/cache@v2` order right? Your feedback would be very much appreciated.

.github/workflows/ci.yml
```diff
@@ -109,6 +109,42 @@ jobs:
...
+      with:
+        access_token: ${{ github.token }}
+
+    - uses: actions/checkout@v2
+
+    - uses: actions-rs/toolchain@v1
+      with:
+        profile: minimal
+        toolchain: stable
+        override: true
```
TriplEight, line 127:
```suggestion
        override: true
        components: clippy
```

TriplEight (changes requested):
I do not know if `actions-rs/cargo` is any better than just running a script. except the former takes more lines.
Other than that, it's better to have key jobs named, that looks better in the CI output.

.github/workflows/ci.yml
```diff
@@ -109,6 +109,42 @@ jobs:
...
+
```
TriplEight, line 128:
```suggestion
        components: clippy
```

TriplEight: hope I've guessed with the padding here

misc/multiaddr/src/lib.rs
```diff
@@ -59,6 +60,11 @@ impl Multiaddr {
         self.bytes.len()
     }
 
+    /// Returns true if the length of this multiaddress is 0.
+    pub fn is_empty(&self) -> bool {
```
romanb, line 64: Was this a clippy suggestion? Is it used anywhere?

protocols/kad/src/record.rs
romanb: Is this change connected in any way to clippy suggestions? I don't remember exactly, but I do think the derived `PartialEq` was deliberate at some point, but maybe the code changed afterwards. It is certainly important to know and understand these equality semantics. E.g. if a `RecordStore` or any code path discards records given because it thinks it already has that record based on a comparison with this new `PartialEq`, that would not go well, since these records are regularly re-published on purpose, and if a new record has either new addresses or a new expiry, it should be stored and replace the existing one. So we need to be very clear on when two provider records are considered equal. I'm not sure they should be considered equal if they have different addresses or a different expiry, but if you think that makes more sense, I'm fine with it, but it may be worthwhile documenting it more clearly in that case.

protocols/kad/src/record.rs
```diff
@@ -122,6 +122,21 @@ impl Hash for ProviderRecord {
...
+    fn eq(&self, other: &Self) -> bool {
+        // In the [`Hash`] implementation of [`ProviderRecord`] all but the
+        // `key` and `provider` fields are ignored. The following property has
+        // to hold for all inputs: `k1 == k2 ⇒ hash(k1) == hash(k2)`. Thereby
+        // the [`PartialEq`] implementation has to ignore `key` and `provider`
+        // as well.
```
romanb, line 131: I don't see how that follows since `k1 == k2 ⇒ hash(k1) == hash(k2)` certainly holds with the derived `PartialEq`/`Eq`. The required implication only goes one way. Am I overlooking something?

romanb (commented):
Most of the changes seem to be purely cosmetical. If clippy fails CI because of such suggestions, that would be mildly annoying.

protocols/kad/src/record.rs
romanb: Actually, this change to equality may be a real problem, because the `MemoryStore` has a `provided: HashSet<ProviderRecord>`.

romanb: Or, looking at the `MemoryStore` implementation, it may even be a _fix_. :eyes:.

romanb: My preliminary conclusion after reviewing that code is that it makes no difference, but the new equality semantics would be nevertheless be preferable, as I don't think that in principle `MemoryStore::provided` should ever contain two records for the same key and provider but with different expiry or addresses. What is your verdict after consideration of my comments?

misc/multiaddr/src/lib.rs
tomaka: Clippy warns if a type has a `len()` method but no `is_empty()` method.

protocols/kad/src/record.rs
mxinden: > My preliminary conclusion after reviewing that code is that it makes no difference

:+1: 

> What is your verdict after consideration of my comments?

In my eyes this newer equality semantic is more intuitive, especially given that it is on-par with the `Hash` implementation.

> but it may be worthwhile documenting it more clearly in that case.

I have pushed 1d28e21f94c84c552ef10f5011d393330bad381f. What do you think?

protocols/kad/src/record.rs
mxinden: In my eyes this should go both ways, thus I should have written `<=>`. I would expect two equal structs to derive to the same hash, and I would expect two equal hashes to be derived from two equal structs (ignoring the probability of hash collisions).

I have pushed d248c2934e14bba9059d1f8887ab7b493801996b. Do you agree with my intuition?

protocols/kad/src/record.rs
```diff
@@ -122,6 +126,21 @@ impl Hash for ProviderRecord {
     }
 }
 
+impl PartialEq for ProviderRecord {
+    fn eq(&self, other: &Self) -> bool {
+        // In the [`Hash`] implementation of [`ProviderRecord`] all but the
+        // `key` and `provider` fields are ignored. The following property has
+        // to hold for all inputs: `k1 == k2 <=> hash(k1) == hash(k2)`. Thereby
+        // the [`PartialEq`] implementation has to ignore `key` and `provider`
+        // as well.
```
romanb, line 135: While the change for equality seems fine, I think this comment is still misleading. This property does not need to hold in both directions and in general usually can't, and saying it must hold "modulo hash collisions" is rendering the property meaningless.

protocols/kad/src/record.rs
romanb: > [..] and I would expect two equal hashes to be derived from two equal structs

Since these are not cryptographic hashes, that is not a reasonable expectation. That hash collisions are totally fine is part of  `std::hash::Hash`es semantics for use with hash-based data structures.

protocols/kad/src/record.rs
mxinden: Thanks for the input. I indeed followed a wrong reasoning here.

I have pushed 5f1c8bec93998ad2f5400150f2e6f18c0b00c087 which removes the comment, now only depending on the `Note` section added in https://github.com/libp2p/rust-libp2p/pull/1850/commits/1d28e21f94c84c552ef10f5011d393330bad381f.

Please insist in case I am still missing something.

mxinden:
> Most of the changes seem to be purely cosmetical. If clippy fails CI because of such suggestions, that would be mildly annoying.

@romanb the clippy CI step will only fail if a `deny` level lint is introduced in the pull request. An example for a helpful `deny` level lint is `unused_io_amount`, which fired for `libp2p-pnet` and is fixed in this pull request. See [list of lints](https://rust-lang.github.io/rust-clippy/master/index.html) for details. `warn` level lints will show up inlined in the code changes, see warnings inlined in this pull request but will not fail the CI step.

Personally I find the clippy suggestions helpful. What do you think?

romanb (approved):
:+1:

mxinden:
@TriplEight thanks for your review. I addressed all your comments. Would you mind taking another look?

TriplEight (approved):
Looks good now @mxinden !

I was thinking maybe it would make sense to introduce generic levels of `clippy` and `fmt`, but in a way that wouldn't fail the CI every time, hence wouldn't be that annoying.
So just a proposal: make these new jobs, make them conditional and depending on, say, `ready to merge` label, and mark them as required statuses for merging. Or is it too much of management?

mxinden:
> So just a proposal: make these new jobs, make them conditional and depending on, say, ready to merge label, and mark them as required statuses for merging.

To me a failing CI step (i.e. clippy) on a not-yet-read-to-merge pull request is not annoying. I would much rather keep the status quo and not have to add a label to each pull request before all CI steps are run. Of course this is just one opinion. Happy to hear opinions of others as well.

TriplEight:
If it's not annoying, let's add them at the end of the pipeline so testing will show the real problems. Or do you prefer to fail fast?

Either way, how do you like it, just `cargo fmt` and `cargo clippy` or with some sophisticated rules?

mxinden:
> If it's not annoying, let's add them at the end of the pipeline so testing will show the real problems. Or do you prefer to fail fast?

I am not sure I follow. As far as I can tell the clippy step is already running as its own job, and thus in parallel to all other jobs. Am I missing something?

> Either way, how do you like it, just cargo fmt and cargo clippy or with some sophisticated rules?

I think the state introduced with this pull request is good enough. I don't think there are any plans nor needs to introduce `cargo fmt`.

TriplEight:
> As far as I can tell the clippy step is already running as its own job

Thought it runs just 2 checks, not a whole set.

> I don't think there are any plans nor needs to introduce cargo fmt.

OK, np.

Issue #1853, opened 2020-11-23
Change wire protocol name of the "multistream-lazy" variant


tomaka:
The rust-libp2p implementation of `multistream-select`, in addition to the standard `V1` protocol version, also supports a non-standard `V1Lazy` variant.
[Link to the documentation](https://docs.rs/multistream-select/0.8.5/multistream_select/enum.Version.html#variant.V1Lazy).

While this variant removes a round-trip-time in many circumstances, its drawback is that it isn't compatible with the non-Rust versions of libp2p. For that reason, nobody uses it.

I suggest here that we:

- Change [the wire name](https://github.com/libp2p/rust-libp2p/blob/1bd013c8432b88690b845594cf656a8e6756a28a/misc/multistream-select/src/protocol.rs#L41) of `V1Lazy` to be the same as `V1`, therefore making indistinguishable from the standard one.
- Remove the fact that [the listener delays the response](https://github.com/libp2p/rust-libp2p/blob/1bd013c8432b88690b845594cf656a8e6756a28a/misc/multistream-select/src/protocol.rs#L68-L69), as it is, as far as I understand, the reason why this protocol isn't the same as the standard one. The listener should immediately send back the response to the protocol request, even though it might result in two packets.

PR #1855, opened 2020-11-24
[multistream] Make the lazy variant more interoperable.

romanb:
Closes https://github.com/libp2p/rust-libp2p/issues/1853. The remaining optimisation for `V1Lazy` for a listener in the negotiation, whereby the listener delays flushing of the multistream version header, is hereby removed. The only remaining effect of `V1Lazy` as compared to `V1` is on the side of the dialer, which delays flushing of its singular protocol proposal (i.e. if it has no alternatives) in order to send it together with the first application data, e.g. a request, or an attempt is made to read from the negotiated stream, which similarly triggers a flush of the pending protocol proposal, as before. This permits `V1Lazy` dialers to be interoperable with `V1` listeners. The remaining theoretical pitfall whereby  application data gets misinterpreted as another protocol proposal by a listener remains, however unlikely.

`V1` remains the default, but after sufficient experience we may eventually be bold enough to make this lazy dialer flush a part of the default `V1` implementation, removing the dedicated `V1Lazy` version identifier. As an experiment I temporarily changed the default upgrade strategy to `V1Lazy` for all substreams and at least all the tests and the `ipfs-kad` example worked fine.

### DIFF
--- a/misc/multistream-select/CHANGELOG.md
+++ b/misc/multistream-select/CHANGELOG.md
@@ -1,6 +1,14 @@
 # 0.9.0 [unreleased]
 
-- Make the `V1Lazy` upgrade strategy more interoperable with `V1`.
+- Make the `V1Lazy` upgrade strategy more interoperable with `V1`. Specifically,
+  the listener now behaves identically with `V1` and `V1Lazy`. Furthermore, the
+  multistream-select protocol header is now also identical, making `V1` and `V1Lazy`
+  indistinguishable on the wire. The remaining central effect of `V1Lazy` is that the dialer,
+  if it only supports a single protocol in a negotiation, optimistically settles on that
+  protocol without immediately flushing the negotiation data (i.e. protocol proposal)
+  and without waiting for the corresponding confirmation before it is able to start
+  sending application data, expecting the used protocol to be confirmed with
+  the response.
 
 - Fix the encoding and decoding of `ls` responses to
   be spec-compliant and interoperable with other implementations.
--- a/misc/multistream-select/CHANGELOG.md
+++ b/misc/multistream-select/CHANGELOG.md
@@ -1,5 +1,7 @@
 # 0.9.0 [unreleased]
 
+- Make the `V1Lazy` upgrade strategy more interoperable with `V1`.
+
 - Fix the encoding and decoding of `ls` responses to
   be spec-compliant and interoperable with other implementations.
   For a clean upgrade, `0.8.4` must already be deployed.
--- a/misc/multistream-select/src/negotiated.rs
+++ b/misc/multistream-select/src/negotiated.rs
@@ -18,7 +18,7 @@
 // FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
 // DEALINGS IN THE SOFTWARE.
 
-use crate::protocol::{Protocol, MessageReader, Message, Version, ProtocolError};
+use crate::protocol::{Protocol, MessageReader, Message, ProtocolError, HeaderLine};
 
 use futures::{prelude::*, io::{IoSlice, IoSliceMut}, ready};
 use pin_project::pin_project;
@@ -83,9 +83,9 @@ impl<TInner> Negotiated<TInner> {
     pub(crate) fn expecting(
         io: MessageReader<TInner>,
         protocol: Protocol,
-        version: Option<Version>
+        header: Option<HeaderLine>
     ) -> Self {
-        Negotiated { state: State::Expecting { io, protocol, version } }
+        Negotiated { state: State::Expecting { io, protocol, header } }
     }
 
     /// Polls the `Negotiated` for completion.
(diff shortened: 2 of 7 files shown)


misc/multistream-select/src/protocol.rs
```diff
@@ -52,45 +50,32 @@ pub enum Version {
...
-    ///      the listener may not recognize upgrade boundaries and erroneously
-    ///      process a request despite not supporting one of the intermediate
-    ///      protocols that the dialer committed to. See [1] and [2].
+    /// > **Note**: `V1Lazy` is specific to `rust-libp2p`. The wire protocol is identical to `V1`
+    /// > and generally interoperable with peers only supporting `V1`. Nevertheless, there is a
+    /// > pitfall that is rarely encountered: When nesting multiple protocol negotiations, the
+    /// > listener should either be known to support all of the dialer's optimistically chosen
+    /// > protocols or there is must be no intermediate protocol without a payload and none of
+    /// > the protocol payloads must have the potential for being mistaken for a multistream-select
+    /// > protocol message. This avoids rare edge-cases whereby the listener may not recognize
```
tomaka, line 76: Let me know if I understand correctly:

If I create a protocol named `/foo` and the first message sent after opening a `/foo` substream is `/bar\nhello world`, then, if the listener doesn't support protocol `/foo`, it will misinterpret the payload as the dialer opening the protocol `/bar` and sending `hello world`.

misc/multistream-select/src/negotiated.rs
```diff
@@ -125,11 +129,12 @@ impl<TInner> Negotiated<TInner> {
                         }
                     };
 
-                    if let Message::Header(v) = &msg {
-                        if *v == version {
-                            *this.state = State::Expecting { io, protocol, version };
+                    match (&msg, version) {
+                        (Message::Header(Version::V1), Some(Version::V1)) => {
```
tomaka, line 133: The fact that we assume that `version` can never be `V1Lazy` looks a bit like a hack to me.

In an ideal world, we would differentiate the `Version` enum (containing `V1` and `V1Lazy`) from the `VersionHandshake` (containing only `V1`), no?

tomaka (approved):

misc/multistream-select/src/protocol.rs
romanb: Yes, see https://github.com/multiformats/go-multistream/issues/20 for more details (that link is also in the code).

misc/multistream-select/src/negotiated.rs
romanb: > The fact that we assume that version can never be V1Lazy looks a bit like a hack to me.

By definition of these versions, they're now indistinguishable on the wire and it is always expected to receive `V1`. So receiving `V1Lazy` is considered an error (in the decoder really).

> In an ideal world, we would differentiate the Version enum (containing V1 and V1Lazy) from the VersionHandshake (containing only V1), no?

Since `V1Lazy` is now really the same negotiation protocol as `V1`, only differing subtly in the behaviour of the dialer, it may be cleaner to separate this "lazy flush yes/no" setting from the `Version`, making it a separate configuration option for the dialer only, since `V1Lazy` used for a listener now has no effect at all. However, for transport and substream upgrades we currently only have this choice of `Version` without a concept of separate parameters for the roles of dialer and/or listener. Maybe `Version::V1 { dialer_lazy_flush: bool }` would already be an improvement, though be another API change. It would always be decoded from the wire with `dialer_lazy_flush: false` since that option is local. Should I give that a try?  So due to the fact that this would require larger API changes, together with my hope that `V1Lazy` will eventually disappear altogether, merging its behaviour into `V1`, led me to the current approach, but I'm open to alternatives.

tomaka: > It would always be decoded from the wire with dialer_lazy_flush: false since that option is local.

What I had in mind is that you wouldn't decode a `Version`, but a new enum called for example `VersionHandshake` that would only contain `V1` without any parameter.

I don't know if that makes sense, and I'm also not strongly opinionated. I'm fine with these changes.

romanb: > What I had in mind is that you wouldn't decode a Version, but a new enum called for example VersionHandshake that would only contain V1 without any parameter.

I see, separating these two kinds of versions. I will give that a try.

romanb: My proposal is https://github.com/libp2p/rust-libp2p/pull/1855/commits/70f041b8a79fb0f6e21048577e71188345dc2c25.

tomaka (approved):

misc/multistream-select/CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 # 0.9.0 [unreleased]
 
+- Make the `V1Lazy` upgrade strategy more interoperable with `V1`.
```
mxinden, line 3: Could you extend this comment? E.g. that the listener side changes are removed and that the `V1` identifier is used instead of `V1Lazy`.

mxinden (approved):

misc/multistream-select/CHANGELOG.md
romanb, line 3: How about this:
```suggestion
- Make the `V1Lazy` upgrade strategy more interoperable with `V1`. Specifically,
  the listener now behaves identically with `V1` and `V1Lazy`. Furthermore, the
  multistream-select protocol header is now also identical, making `V1` and `V1Lazy`
  indistinguishable on the wire. The remaining central effect of `V1Lazy` is that the dialer,
  if it only supports a single protocol in a negotiation, optimistically settles on that
  protocol without immediately flushing the negotiation data (i.e. protocol proposal)
  and without waiting for the corresponding confirmation before it is able to start
  sending application data, expecting for the used protocol to be confirmed with
  the response.
```

mxinden: Yes, perfect. I think that is very helpful. 

In "expecting for the used protocol to be confirmed with the response." I am not sure the "for" is needed, but I might just be mistaken.

PR #1857, opened 2020-11-25
Prepare release of `libp2p-v0.31` and all unreleased sub-crates.

romanb:
Prepare release of `libp2p-v0.31` and all unreleased sub-crates.

PR #1858, opened 2020-11-25
[swarm] Permit configuration override for the substream upgrade protocol to use.

romanb:
It is desirable to configure the substream upgrade protocol (i.e. multistream-select protocol) to use for substreams in general. Since introducing configuration options for the substream upgrade protocol to all libp2p protocol crates seems to be noisy and likely unnecessary, this PR proposes a configuration option for the `Swarm` that will apply to all (outbound) substreams. In this way, one can configure the upgrade protocols in a single place, once for the transport and once for the substreams.

### DIFF
--- a/swarm/CHANGELOG.md
+++ b/swarm/CHANGELOG.md
@@ -1,5 +1,9 @@
 # 0.25.0 [unreleased]
 
+- Permit a configuration override for the substream upgrade protocol
+  to use for all (outbound) substreams.
+  [PR 1858](https://github.com/libp2p/rust-libp2p/pull/1858).
+
 - Changed parameters for connection limits from `usize` to `u32`.
   Connection limits are now configured via `SwarmBuilder::connection_limits()`.
 
--- a/swarm/src/lib.rs
+++ b/swarm/src/lib.rs
@@ -990,7 +990,7 @@ where TBehaviour: NetworkBehaviour,
             transport: transport,
             behaviour,
             network_config: Default::default(),
-            substream_protocol_override: None,
+            substream_upgrade_protocol_override: None,
         }
     }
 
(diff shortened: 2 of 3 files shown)


romanb:
This is intended for inclusion in https://github.com/libp2p/rust-libp2p/pull/1857.

mxinden (approved):

tomaka (approved):
Before this PR, my opinion was that each protocol should individually decide whether `V1` or `V1Lazy` should be used, considering that the safety guarantees of `V1Lazy` depend on the details of the higher-level protocol.

However I've now realized that this would be a bad idea if multistream-select 2.0 gets released. It should indeed be the `Swarm` that decides which version of multistream-select is in use, and not individual protocols.

romanb:
> However I've now realized that this would be a bad idea if multistream-select 2.0 gets released. It should indeed be the Swarm that decides which version of multistream-select is in use, and not individual protocols.

Yes, we should probably remove the ability to configure the upgrade protocol on a `SubstreamProtocol`. The reason it got introduced was to allow very selective use of `V1Lazy` for specific protocols, but that was before `V1Lazy` was pretty much interoperable with `V1`.

mxinden:
> > However I've now realized that this would be a bad idea if multistream-select 2.0 gets released. It should indeed be the Swarm that decides which version of multistream-select is in use, and not individual protocols.
> 
> Yes, we should probably remove the ability to configure the upgrade protocol on a `SubstreamProtocol`. The reason it got introduced was to allow very selective use of `V1Lazy` for specific protocols, but that was before `V1Lazy` was pretty much interoperable with `V1`.

I created https://github.com/libp2p/rust-libp2p/issues/1859 to continue this discussion as well as to keep track of it.

PR #1862, opened 2020-11-26
swarm: Add ExpandedSwarm::is_connected

mxinden:
Commit 335e55e6 removed the `ConnectionInfo` trait in favor of `PeerId`s. Commit
1bd013c8 removed `ExpandedSwarm::connection_info` as it would only return the
`PeerId` that the caller is already aware of.

One could use `ExpandedSwarm::connection_info` not only to retrieve the
`ConnectionInfo` for a given peer, but also to check whether the underlying
`Network` has a connection to the peer.

This commit exposes the `is_connected` method on `Network` via `ExpandedSwarm`
to check whether the `Network` has an established connection to a given peer.

---

One could as well expose other read-only methods on `Network` like `is_dialing`
and `is_disconnected`. For now this pull request only restores the status-quo.
Let me know what you prefer.

### DIFF
--- a/swarm/CHANGELOG.md
+++ b/swarm/CHANGELOG.md
@@ -1,3 +1,8 @@
+# 0.25.1 [2020-11-26]
+
+- Add `ExpandedSwarm::is_connected`.
+  [PR 1862](https://github.com/libp2p/rust-libp2p/pull/1862).
+
 # 0.25.0 [2020-11-25]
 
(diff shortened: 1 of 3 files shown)


romanb (approved):

mxinden:
Release [tagged](https://github.com/libp2p/rust-libp2p/tree/libp2p-swarm-0.25.1) and [published](https://crates.io/crates/libp2p-swarm/0.25.1).

PR #1865, opened 2020-11-28
Use Arc<Vec<u8>> to replace Bytes in PeerId

jolestar:
For avoiding cargo clippy raise `mutable_key_type` error when using PeerId as HashSet/HashMap's key.

### DIFF
--- a/core/src/peer_id.rs
+++ b/core/src/peer_id.rs
@@ -19,11 +19,11 @@
 // DEALINGS IN THE SOFTWARE.
 
 use crate::PublicKey;
-use bytes::Bytes;
 use thiserror::Error;
 use multihash::{Code, Multihash, MultihashDigest};
 use rand::Rng;
 use std::{convert::TryFrom, borrow::Borrow, fmt, hash, str::FromStr, cmp};
+use std::sync::Arc;
 
 /// Public keys with byte-lengths smaller than `MAX_INLINE_KEY_LENGTH` will be
 /// automatically used as the peer id using an identity multihash.
@@ -35,7 +35,7 @@ const MAX_INLINE_KEY_LENGTH: usize = 42;
 // TODO: maybe keep things in decoded version?
 #[derive(Clone, Eq)]
 pub struct PeerId {
-    multihash: Bytes,
+    multihash: Arc<Vec<u8>>,
 }
 
 impl fmt::Debug for PeerId {
@@ -78,9 +78,9 @@ impl PeerId {
             Code::Sha2_256
         };
 
-        let multihash = hash_algorithm.digest(&key_enc).to_bytes().into();
+        let multihash = hash_algorithm.digest(&key_enc).to_bytes();
 
-        PeerId { multihash }
+        PeerId { multihash: Arc::new(multihash) }
     }
 
     /// Checks whether `data` is a valid `PeerId`. If so, returns the `PeerId`. If not, returns
@@ -99,9 +99,9 @@ impl PeerId {
     /// peer ID, it is returned as an `Err`.
     pub fn from_multihash(multihash: Multihash) -> Result<PeerId, Multihash> {
         match Code::try_from(multihash.code()) {
-            Ok(Code::Sha2_256) => Ok(PeerId { multihash: multihash.to_bytes().into() }),
+            Ok(Code::Sha2_256) => Ok(PeerId { multihash: Arc::new(multihash.to_bytes()) }),
             Ok(Code::Identity) if multihash.digest().len() <= MAX_INLINE_KEY_LENGTH
-                => Ok(PeerId { multihash: multihash.to_bytes().into() }),
+                => Ok(PeerId { multihash: Arc::new(multihash.to_bytes()) }),
             _ => Err(multihash)
         }
     }
@@ -112,8 +112,8 @@ impl PeerId {
     pub fn random() -> PeerId {
         let peer_id = rand::thread_rng().gen::<[u8; 32]>();
         PeerId {
-            multihash: Multihash::wrap(Code::Identity.into(), &peer_id)
-                .expect("The digest size is never too large").to_bytes().into()
+            multihash: Arc::new(Multihash::wrap(Code::Identity.into(), &peer_id)
+                .expect("The digest size is never too large").to_bytes())
         }
     }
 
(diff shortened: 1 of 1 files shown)


romanb:
From https://rust-lang.github.io/rust-clippy/master/index.html#mutable_key_type (emphasis mine):

> Known problems

> It’s correct to use a struct, that contains interior mutability as a key, when its Hash implementation doesn’t access any of the interior mutable types. However, this lint is unable to recognize this, so it causes a false positive in theses cases. __The bytes crate is a great example of this.__

See also https://github.com/rust-lang/rust-clippy/issues/5812.

mxinden:
For the reason mentioned by Roman above, we silence the warning in our CI:

https://github.com/libp2p/rust-libp2p/blob/dae07b075b421fb0a33bbf85d91ad802e37845f9/.github/workflows/ci.yml#L146

dvc94ch:
Well this affects all downstream users too, and it's not clear why using `Bytes` is more efficient or necessary in this case.

jolestar:
> Well this affects all downstream users too, and it's not clear why using `Bytes` is more efficient or necessary in this case. Perhaps you mean to use an `Arc<[u8]>`?

The Bytes in PeerId is converted from Vec<u8>,  I think using Vec<u8> at here is more efficient, for avoiding call into().

romanb:
> Well this affects all downstream users too, and it's not clear why using Bytes is more efficient or necessary in this case. Perhaps you mean to use an Arc<[u8]>?

The situation is simply that we have a `Vec<u8>` that we get from `multihash` and want cheap cloning for `PeerId`s. `Bytes` is just `Arc<Vec<u8>>` on steroids. If we don't need any of the additional features of `Bytes` for the implementation of `PeerId`, we can certainly go with an `Arc<Vec<u8>>`. I don't see what the conversion to `Arc<[u8]>` would buy us, given that it [does not even look like it can reuse the vector allocation and needs to copy the data again](https://doc.rust-lang.org/src/alloc/sync.rs.html#2200). In any case, just using the `Vec<u8>` is I think not a good idea, as one has to assume that every clone allocates a new small vector.

> The Bytes in PeerId is converted from Vec, I think using Vec at here is more efficient, for avoiding call into().

The important part is cheap cloning.

dvc94ch:
I just looked at how bytes works, as I assumed it would require an additional allocation (as Arc<Vec<u8>> would). It delays constructing the `Arc<Vec<u8>>` until the first clone. I guess an `Arc<[u8]>` may help prevent invalidating cache lines, since it's only a single allocation, but that's probably a stretch. I guess wrapping the `Vec<u8>` in an `Arc` could work?

jolestar:
> I just looked at how bytes works, as I assumed it would require an additional allocation (as Arc<Vec> would). It delays constructing the `Arc<Vec<u8>>` until the first clone. I guess an `Arc<[u8]>` may improve data cache utilization, since it's only a single allocation, but that's probably a stretch. I guess wrapping the `Vec<u8>` in an `Arc` could work?

I try to using Arc<Vec<u8>> to replace Bytes.

@romanb

core/src/peer_id.rs
```diff
@@ -123,7 +123,7 @@ impl PeerId {
     /// equality of peer IDs. That is, two peer IDs may be considered equal
     /// while having a different byte representation as per `into_bytes`.
     pub fn into_bytes(self) -> Vec<u8> {
-        self.multihash.to_vec()
+        self.multihash.as_ref().clone()
     }
```
dvc94ch, line 127: I guess this should be replaced with an `AsRef<[u8]>` or `fn as_bytes(&self) -> &[u8]` and defer cloning to the call site if necessary

dvc94ch: and an even better solution would be to use `Multihash` internally, since it is stack allocated and use an `Arc<PeerId>` if efficient cloning is required. For cid we just implement `Copy`. This also has the advantage that `Vec<PeerId>` only requires a single allocation.

dvc94ch: Seems like the reason for this is that the kademlia buckets require something that implements `Borrow<[u8]>`  so I guess this is the best that can be done for now

vorot93:
What is the problem with continuing to use `Bytes`?

dvc94ch:
It's an annoyance for all projects that use clippy and the only advantage you get is you defer allocating the arc to the first clone. Something like peeid should be stack allocated and implement the copy trait so that it can be used in collections or boxes or arcs, and a lot of work was done on multihash to make it stack allocated so that it composes well with other rust types.

vorot93:
I agree that if it makes semantic sense, types should be stack allocated. It's just that replacing `Bytes` with `Arc` is not the step in this direction (and the right one at all).

dvc94ch:
> It's just that replacing Bytes with Arc is not the step in this direction (and the right one at all).

I don't see it being a worse solution. As I explained I think it should be stack allocated, but since Bytes is already creating an Arc<Vec<u8>> internally it barely makes a difference. Do you have a reason to believe it will lead to a measurable performance degradation in your codebase?

dvc94ch:
Opened a PR for stack allocating PeerId which also makes clippy happy in #1874

vorot93:
1. `Bytes` does not enforce a certain storage representation (like `Arc<Vec<u8>>`). For example, you can actually create a `Bytes` based on a `'static` slice - and it will not cause allocation (see [Bytes::from_static](https://docs.rs/bytes/0.6.0/bytes/struct.Bytes.html#method.from_static)).
2. `Bytes` directly stores the data point in it, and even updates it when you slice it. With `Arc<Vec<u8>>`, the pointer to the **start** of the data is stored inside `Vec` - which means already one indirection away. And then you have to add the offset to it. Which means `Bytes` should be a little bit more efficient to access.

dvc94ch:
@vorot93 good to know, I wasn't aware of the second point. However I think #1874 is something everyone could get behind. Wdyt?

jolestar:
> 1. `Bytes` does not enforce a certain storage representation (like `Arc<Vec<u8>>`). For example, you can actually create a `Bytes` based on a `'static` slice - and it will not cause allocation (see [Bytes::from_static](https://docs.rs/bytes/0.6.0/bytes/struct.Bytes.html#method.from_static)).
> 2. `Bytes` directly stores the data point in it, and even updates it when you slice it. With `Arc<Vec<u8>>`, the pointer to the **start** of the data is stored inside `Vec` - which means already one indirection away. And then you have to add the offset to it. Which means `Bytes` should be a little bit more efficient to access.

But in PeerId, we do not support convert `static slice to PeerId, and also does not support update, so the two points are not a good reason to keep Bytes. I think #1874 is a good solution.

jolestar:
#1874 is approved, so I close this pr.

PR #1871, opened 2020-12-02
[multistream-select] Listener conformity for failed negotiations.

romanb:
Context: https://github.com/paritytech/substrate/pull/7478#issuecomment-736803765

When `V1Lazy` is used and the listener does not support the optimistic (and singular) proposal of the dialer, it currently happens that dialer and listener get a different outcome of the negotiation. The dialer eventually detects the failed negotiation as soon as it tries to read from the stream, but the listener either encounters an invalid message or unexpected premature EOF, depending on the payload that the dialer sent prematurely after its protocol proposal. In these cases the listener must be lenient and fail the negotiation "normally", i.e. not with a protocol violation or an I/O error. The tests have been expanded to make sure both versions behave uniformly in this manner w.r.t. failed negotiations. Nevertheless it would be good to get final confirmation by testing https://github.com/paritytech/substrate/pull/7478 with this branch. This qualifies for a patch release of `multistream-select`.

### DIFF
--- a/misc/multistream-select/src/listener_select.rs
+++ b/misc/multistream-select/src/listener_select.rs
@@ -57,7 +57,8 @@ where
         protocols: SmallVec::from_iter(protocols),
         state: State::RecvHeader {
             io: MessageIO::new(inner)
-        }
+        },
+        rejected_protocol: false,
     }
 }
 
@@ -72,7 +73,14 @@ where
     // TODO: It would be nice if eventually N = Protocol, which has a
     // few more implications on the API.
     protocols: SmallVec<[(N, Protocol); 8]>,
-    state: State<R, N>
+    state: State<R, N>,
+    /// Whether the listener rejected a proposed protocol.
+    ///
+    /// If the listener reads garbage or EOF after having rejected
+    /// a protocol, the dialer is likely using `V1Lazy` and
+    /// negotiation must be considered failed, but not with a
+    /// protocol violation or I/O error.
+    rejected_protocol: bool,
 }
 
 enum State<R, N>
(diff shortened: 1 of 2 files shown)


misc/multistream-select/src/tests.rs
```diff
@@ -112,8 +122,59 @@ fn no_protocol_found() {
...
+    // when `V1Lazy` is used.
+    //
+    // The choices here cover the specific situations that can arise with
+    // `V1Lazy` and which must nevertheless behave identically to `V1` w.r.t.
+    // the outcome of the negotiation.
+    let payloads = vec!{
+        // No payload, in which case all versions should behave identically
+        // in any case, i.e. the baseline test.
+        vec![],
+        // With this payload and `V1Lazy`, the listener interpets the first
```
mxinden, line 154:
```suggestion
        // With this payload and `V1Lazy`, the listener interprets the first
```

misc/multistream-select/src/listener_select.rs
```diff
@@ -72,7 +73,14 @@ where
...
     protocols: SmallVec<[(N, Protocol); 8]>,
-    state: State<R, N>
+    state: State<R, N>,
+    /// Whether the listener rejected a proposed protocol.
+    ///
+    /// If the listener reads garbage or EOF after having rejected
+    /// a protocol, the dialer is likely using `V1Lazy` and
+    /// negotiation must be considered failed, but not with a
+    /// protocol violation or I/O error.
+    rejected_protocol: bool,
```
mxinden, line 83: Instead of tracking whether one has ever rejected a protocol, would it make more sense to track whether the previous message send was a `NotAvailable`?

Sample scenario:

1. The listener rejects a protocol by sending `NotAvailabel` thus setting `rejected_protocol` to `true`.

2. The dialer then sends a `ListProtocols`.

3. The listener returns a `Protocols` message.

4. The dialer sends garbage down the line.

5. The listener terminates with a `NegotiationErr::Failed` given that `rejected_protocol` is still set to `true` instead of reporting the protocol violation.

mxinden (commented):
Thanks for the patch!

> Nevertheless it would be good to get final confirmation by testing paritytech/substrate#7478 with this branch.

Working on this right now.

mxinden:
> Nevertheless it would be good to get final confirmation by testing paritytech/substrate#7478 with this branch.

The issue described in https://github.com/paritytech/substrate/pull/7478#issuecomment-736803765 is resolved when using this branch.

misc/multistream-select/src/listener_select.rs
romanb: Sounds like a sensible refinement. I will try to do that.

romanb: How about https://github.com/libp2p/rust-libp2p/pull/1871/commits/de3f0478013d629cbc77189f710be83dd326aa3f?

misc/multistream-select/src/listener_select.rs
```diff
@@ -72,7 +73,14 @@ where
     // TODO: It would be nice if eventually N = Protocol, which has a
     // few more implications on the API.
     protocols: SmallVec<[(N, Protocol); 8]>,
-    state: State<R, N>
+    state: State<R, N>,
+    /// Whether the last message sent was a protocol rejection (i.e. `n/a`).
```
mxinden, line 77: I am not sure what is meant with `n/a`, did you mean to write `na\n`?

```suggestion
    /// Whether the last message sent was a protocol rejection (i.e. `na\n`).
```

mxinden (approved):

romanb:
I will publish this as `multistream-select-0.9.1` with changelog updates upon merging.

romanb:
Published, also via `libp2p-0.31.2` for convenience.

mxinden:
> Published, also via `libp2p-0.31.2` for convenience.

Thanks. I can do the update on Substrate later today. Will do this as a separate pull request given that https://github.com/paritytech/substrate/pull/7478/ might not be merged any time soon and given that others might run into this issue as well.

PR #1874, opened 2020-12-06
Stack allocated PeerId

dvc94ch:
Allocates PeerId's on the stack, incidentally also making clippy happy. For more info see #1865. Superceeds #1865.

### DIFF
--- a/core/src/peer_id.rs
+++ b/core/src/peer_id.rs
@@ -22,7 +22,7 @@ use crate::PublicKey;
 use thiserror::Error;
 use multihash::{Code, Multihash, MultihashDigest};
 use rand::Rng;
-use std::{convert::TryFrom, borrow::Borrow, fmt, hash, str::FromStr, cmp};
+use std::{convert::TryFrom, fmt, str::FromStr};
 
 /// Public keys with byte-lengths smaller than `MAX_INLINE_KEY_LENGTH` will be
 /// automatically used as the peer id using an identity multihash.
@@ -31,7 +31,7 @@ const MAX_INLINE_KEY_LENGTH: usize = 42;
 /// Identifier of a peer of the network.
 ///
 /// The data is a multihash of the public key of the peer.
-#[derive(Clone, Copy, Eq)]
+#[derive(Clone, Copy, Eq, Hash, Ord, PartialEq, PartialOrd)]
 pub struct PeerId {
     multihash: Multihash,
 }
@@ -47,21 +47,6 @@ impl fmt::Debug for PeerId {
 impl fmt::Display for PeerId {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         self.to_base58().fmt(f)
-    }
-}
-
-impl cmp::PartialOrd for PeerId {
-    fn partial_cmp(&self, other: &Self) -> Option<cmp::Ordering> {
-        Some(Ord::cmp(self, other))
-    }
-}
-
-impl cmp::Ord for PeerId {
-    fn cmp(&self, other: &Self) -> cmp::Ordering {
-        // must use borrow, because as_bytes is not consistent with equality
-        let lhs: &Multihash = self.borrow();
-        let rhs: &Multihash = other.borrow();
-        lhs.cmp(rhs)
     }
 }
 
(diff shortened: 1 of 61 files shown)


core/src/peer_id.rs
```diff
@@ -32,10 +31,9 @@ const MAX_INLINE_KEY_LENGTH: usize = 42;
 /// Identifier of a peer of the network.
 ///
 /// The data is a multihash of the public key of the peer.
-// TODO: maybe keep things in decoded version?
-#[derive(Clone, Eq)]
+#[derive(Clone, Copy, Eq, Hash, Ord, PartialEq, PartialOrd)]
```
thomaseizinger, line 34: `Copy` :partying_face:

examples/ipfs-kad.rs
```diff
@@ -91,7 +91,7 @@ fn main() -> Result<(), Box<dyn Error>> {
     };
 
     println!("Searching for the closest peers to {:?}", to_search);
-    swarm.get_closest_peers(to_search);
+    swarm.get_closest_peers(to_search.to_bytes());
```
romanb, line 94: No longer being able to pass `PeerId`s or `Multihash`es to the Kademlia API, as in this case to `get_closest_peers()`, kind of defeats the purpose of the generic `key: K` where `K: Borrow<[u8]>` argument. Could we change these arguments, which I think exist for `Kademlia::kbucket()` and `Kademlia::get_closest_peers()` to be `key: K` where `K: Into<kbucket::Key<K>>`?

protocols/kad/src/behaviour/test.rs
```diff
@@ -239,13 +239,13 @@ fn query_iter() {
         // Ask the first peer in the list to search a random peer. The search should
         // propagate forwards through the list of peers.
         let search_target = PeerId::random();
-        let search_target_key = kbucket::Key::new(search_target.clone());
-        let qid = swarms[0].get_closest_peers(search_target.clone());
+        let search_target_key = kbucket::Key::from(search_target);
+        let qid = swarms[0].get_closest_peers(search_target.to_bytes());
```
romanb, line 243: See the other comment about retaining the ability to pass `PeerId`s and `Multihash`es to this API.

transports/websocket/Cargo.toml
```diff
@@ -16,7 +16,7 @@ futures = "0.3.1"
 libp2p-core = { version = "0.25.0", path = "../../core" }
 log = "0.4.8"
 quicksink = "0.1"
-rustls = "0.18.0"
+rustls = "0.19.0"
```
romanb, line 19: I don't think this upgrade should be in this PR.

romanb (commented):
The direction looks good to me. Contrary to the PR description, I don't think this closes #1810, as we use `bytes` in a few other places throughout `libp2p`. It is also better not to address #1852 here, as that is unrelated. The build should be fixed separately either by upgrading `rustls` (https://github.com/libp2p/rust-libp2p/pull/1852) or by `async-tls-0.10.2` being yanked. See https://github.com/async-rs/async-tls/issues/42.

romanb (approved):
Looks good to me. I'd just like to leave this PR open for a bit longer as the next `libp2p` release is just to update `libp2p-mdns` and `libp2p-request-response`, and this PR will essentially trigger new releases of most crates. It may also give more people the chance to comment here. If there are no objections we can probably merge and release this next week.

mxinden (commented):
I am in favor of this patch. Thanks @dvc94ch.

Given that with this pull request `PeerId` implements `Copy` I would suggest removing all unnecessary `clone()` calls. What do you think?

I have run the benchmarks added in https://github.com/libp2p/rust-libp2p/pull/1875 plus the additional `sort_vec` benchmark. See the result below.

``` bash
$ git checkout master
$ cargo bench --bench peer_id -- --save-baseline master
$ git checkout peer-id-stack
$ cargo bench --bench peer_id -- --save-baseline peer-id-stack
$ critcmp master peer-id-stack

group         master                                 peer-id-stack                                                    
-----         ------                                 -------------                                                    
clone         1.00     15.6±0.27ns        ? B/sec    3.07     47.8±1.08ns        ? B/sec                              
from_bytes    1.59    179.1±5.96ns        ? B/sec    1.00    112.3±6.03ns        ? B/sec                              
sort_vec      1.07      7.5±0.27µs        ? B/sec    1.00      7.0±0.31µs        ? B/sec
```

romanb:
I plan to merge this later today. @mxinden Would a follow-up getting-started ticket for removing redundant `.clone()`s be acceptable? Since it is purely a syntactic simplification, I don't think we need to necessarily do that in this PR already.

PR #1875, opened 2020-12-07
core/benches: Add rudimentary benchmark for PeerId::from_bytes and clone

mxinden:
I hope these somewhat simple benchmarks can back up discussions in https://github.com/libp2p/rust-libp2p/pull/1874 and https://github.com/libp2p/rust-libp2p/pull/1865 with data.

```
$ cd core/
$ cargo bench
[...]
from_bytes              time:   [187.15 ns 187.88 ns 188.79 ns]                                                       
                        change: [-11.163% -10.359% -8.9828%] (p = 0.00 < 0.05)                                        
                        Performance has improved.
Found 9 outliers among 100 measurements (9.00%)
  1 (1.00%) low mild
  5 (5.00%) high mild
  3 (3.00%) high severe

clone                   time:   [14.811 ns 14.868 ns 14.940 ns]                                                       
                        change: [-14.744% -14.274% -13.812%] (p = 0.00 < 0.05)                                        
                        Performance has improved.
Found 20 outliers among 100 measurements (20.00%)
  1 (1.00%) low severe
  1 (1.00%) low mild
  6 (6.00%) high mild
  12 (12.00%) high severe


$ cat /proc/cpuinfo | grep "model name" | head -1
model name      : Intel(R) Core(TM) i7-8550U CPU @ 1.80GHz
```

In addition this pull request adds a CI step to ensure benchmarks continue to compile and run.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -145,6 +145,41 @@ jobs:
         command: clippy
         args: -- -A clippy::mutable_key_type -A clippy::type_complexity
 
+  run-benchmarks:
+    runs-on: ubuntu-latest
+    steps:
+
+    - name: Cancel Previous Runs
+      uses: styfle/cancel-workflow-action@0.6.0
+      with:
+        access_token: ${{ github.token }}
+
+    - uses: actions/checkout@v2
+
+    - uses: actions-rs/toolchain@v1
+      with:
+        profile: minimal
+        toolchain: stable
+        override: true
+
+    - name: Cache CARGO_HOME
+      uses: actions/cache@v2
+      with:
+        path: ~/.cargo
+        key: cargo-home-${{ hashFiles('Cargo.toml') }}
+
+    - name: Cache cargo build
+      uses: actions/cache@v2
+      with:
+        path: target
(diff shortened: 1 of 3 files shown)


dvc94ch:
Yes, clone is slower as it needs to copy about double the amount of bytes compared to the `Arc`, but faster if you put `Arc<PeerId>`, compared to just using `Bytes` (by about 14% may vary from run to run slightly).

However you can easily concoct a benchmark depending on what it is you want to prove. I concocted one here:

```rust
fn sort_vec(c: &mut Criterion) {
    let mut peer_ids = Vec::with_capacity(100);
    for _ in 0..peer_ids.capacity() {
        peer_ids.push(identity::Keypair::generate_ed25519()
            .public()
            .into_peer_id());
    }

    c.bench_function("clone", |b| {
        b.iter(|| {
            let mut peer_ids = peer_ids.clone();
            peer_ids.sort_unstable();
            black_box(peer_ids);
        })
    });
}
```

dvc94ch:
Let's merge this then?

romanb (approved):

Issue #1872, opened 2020-12-03
mDNS shouldn't panic if response packet becomes too large


tomaka:
I'm getting this bug report:

```
   Alice ERR: 
   Alice ERR: ====================
   Alice ERR: 
   Alice ERR: Version: 0.8.26-065b2e076-x86_64-linux-gnu
   Alice ERR: 
   Alice ERR:    0: sp_panic_handler::set::{{closure}}
   Alice ERR:    1: std::panicking::rust_panic_with_hook
   Alice ERR:              at /rustc/b7ebc6b0c1ba3c27ebb17c0b496ece778ef11e18/library/std/src/panicking.rs:595
   Alice ERR:    2: std::panicking::begin_panic_handler::{{closure}}
   Alice ERR:              at /rustc/b7ebc6b0c1ba3c27ebb17c0b496ece778ef11e18/library/std/src/panicking.rs:497
   Alice ERR:    3: std::sys_common::backtrace::__rust_end_short_backtrace
   Alice ERR:              at /rustc/b7ebc6b0c1ba3c27ebb17c0b496ece778ef11e18/library/std/src/sys_common/backtrace.rs:141
   Alice ERR:    4: rust_begin_unwind
   Alice ERR:              at /rustc/b7ebc6b0c1ba3c27ebb17c0b496ece778ef11e18/library/std/src/panicking.rs:493
   Alice ERR:    5: core::panicking::panic_fmt
   Alice ERR:              at /rustc/b7ebc6b0c1ba3c27ebb17c0b496ece778ef11e18/library/core/src/panicking.rs:92
   Alice ERR:    6: core::option::expect_none_failed
   Alice ERR:              at /rustc/b7ebc6b0c1ba3c27ebb17c0b496ece778ef11e18/library/core/src/option.rs:1268
   Alice ERR:    7: <libp2p_mdns::behaviour::Mdns as libp2p_swarm::behaviour::NetworkBehaviour>::poll
   Alice ERR:    8: <sc_network::discovery::DiscoveryBehaviour as libp2p_swarm::behaviour::NetworkBehaviour>::poll
   Alice ERR:    9: libp2p_swarm::ExpandedSwarm<TBehaviour,TInEvent,TOutEvent,THandler>::poll_next_event
   Alice ERR:   10: <sc_network::service::NetworkWorker<B,H> as core::future::future::Future>::poll
[...]

PR #1877, opened 2020-12-07
[mdns] Split response packets if necessary.

romanb:
Currently, when an MDNS response gets larger than 9000 bytes due to the number of addresses being encoded, a panic is triggered. See https://github.com/libp2p/rust-libp2p/issues/1872. There are primarily two approaches for dealing with the `9000` bytes MDNS packet size limit: Creating multiple packets or trimming the packet by dropping addresses. Since I didn't find any restriction in the MDNS specifications w.r.t. the number of multicast response packets that can be sent (we don't support or send unicast response packets, I think), I opted for the former. So we now prevent MDNS response packets becoming too large (> 9000 bytes) by using conservative upper bounds and creating multi-packet responses if necessary (effectively when there are more than 29 addresses to encode, which effectively means when the local peer has more than 29 listening addresses, as a peer only responds with its own addresses). I'm only just starting out with studying the MDNS specifications, so feel free to set me straight if I'm doing anything nonsensical here.

Additionally, I opted for just skipping addresses that don't fit into a TXT record or contain invalid characters, instead of failing to send a response (which may contain other valid records), thereby logging a warning for the skipped addresses.

As such, closes https://github.com/libp2p/rust-libp2p/issues/1872.

### DIFF
--- a/protocols/mdns/src/dns.rs
+++ b/protocols/mdns/src/dns.rs
@@ -119,38 +119,35 @@ pub fn build_query_response(
     let peer_id_bytes = encode_peer_id(&peer_id);
     debug_assert!(peer_id_bytes.len() <= 0xffff);
 
+    // The accumulated response packets.
     let mut packets = Vec::new();
 
-    'next_packet: loop {
-        let mut records = Vec::with_capacity(addresses.len() * MAX_TXT_RECORD_SIZE);
+    // The records accumulated per response packet.
+    let mut records = Vec::with_capacity(addresses.len() * MAX_TXT_RECORD_SIZE);
 
-        // Try to pack as many records into an MDNS response packet
-        // as are known to fit comfortably.
-        for _ in 0 .. MAX_RECORDS_PER_PACKET {
-            if let Some(addr) = addresses.next() {
-                let txt_to_send = format!("dnsaddr={}/p2p/{}", addr.to_string(), peer_id.to_base58());
-                let mut txt_record = Vec::with_capacity(txt_to_send.len());
-                match append_txt_record(&mut txt_record, &peer_id_bytes, ttl, &txt_to_send) {
-                    Ok(()) => {
-                        records.push(txt_record);
-                        continue
-                    }
-                    Err(e) => {
-                        log::warn!("Excluding address {} from response: {:?}", addr, e);
-                    }
-                }
+    // Encode the addresses as TXT records, and multiple TXT records into a
+    // response packet.
+    while let Some(addr) = addresses.next() {
+        let txt_to_send = format!("dnsaddr={}/p2p/{}", addr.to_string(), peer_id.to_base58());
+        let mut txt_record = Vec::with_capacity(txt_to_send.len());
+        match append_txt_record(&mut txt_record, &peer_id_bytes, ttl, &txt_to_send) {
+            Ok(()) => {
+                records.push(txt_record);
             }
-
-            // Push the final packet, if there are still unpacked records.
-            if !records.is_empty() {
-                packets.push(query_response_packet(id, &peer_id_bytes, &records, ttl));
+            Err(e) => {
+                log::warn!("Excluding address {} from response: {:?}", addr, e);
             }
-
-            break 'next_packet
         }
 
-        // Reached MAX_RECORDS_PER_PACKET. Finish up the current packet and
-        // begin a new one.
+        if records.len() == MAX_RECORDS_PER_PACKET {
+            packets.push(query_response_packet(id, &peer_id_bytes, &records, ttl));
+            records.clear();
+        }
+    }
+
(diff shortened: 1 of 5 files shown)


protocols/mdns/src/dns.rs
```diff
@@ -80,68 +101,70 @@ pub fn build_query() -> Vec<u8> {
...
-    append_u16(&mut out, addresses.len() as u16);
+    let peer_id_bytes = encode_peer_id(&peer_id);
+    debug_assert!(peer_id_bytes.len() <= 0xffff);
 
-    // Our single answer.
-    // The name.
-    append_qname(&mut out, SERVICE_NAME);
+    let mut packets = Vec::new();
+
+    'next_packet: loop {
```
mxinden, line 124: It took me a bit to understand this loop. The following control flow would be easier for me to understand. What do you think?

```diff
diff --git a/protocols/mdns/src/dns.rs b/protocols/mdns/src/dns.rs
index efb5a11a..fcc62fc2 100644
--- a/protocols/mdns/src/dns.rs
+++ b/protocols/mdns/src/dns.rs
@@ -120,38 +120,23 @@ pub fn build_query_response(
     debug_assert!(peer_id_bytes.len() <= 0xffff);
 
     let mut packets = Vec::new();
-
-    'next_packet: loop {
-        let mut records = Vec::with_capacity(addresses.len() * MAX_TXT_RECORD_SIZE);
-
-        // Try to pack as many records into an MDNS response packet
-        // as are known to fit comfortably.
-        for _ in 0 .. MAX_RECORDS_PER_PACKET {
-            if let Some(addr) = addresses.next() {
-                let txt_to_send = format!("dnsaddr={}/p2p/{}", addr.to_string(), peer_id.to_base58());
-                let mut txt_record = Vec::with_capacity(txt_to_send.len());
-                match append_txt_record(&mut txt_record, &peer_id_bytes, ttl, &txt_to_send) {
-                    Ok(()) => {
-                        records.push(txt_record);
-                        continue
-                    }
-                    Err(e) => {
-                        log::warn!("Excluding address {} from response: {:?}", addr, e);
-                    }
-                }
+    let mut records = Vec::with_capacity(addresses.len() * MAX_TXT_RECORD_SIZE);
+
+    while let Some(addr) = addresses.next() {
+        let txt_to_send = format!("dnsaddr={}/p2p/{}", addr.to_string(), peer_id.to_base58());
+        let mut txt_record = Vec::with_capacity(txt_to_send.len());
+        match append_txt_record(&mut txt_record, &peer_id_bytes, ttl, &txt_to_send) {
+            Ok(()) => {
+                records.push(txt_record);
+                continue
             }
-
-            // Push the final packet, if there are still unpacked records.
-            if !records.is_empty() {
-                packets.push(query_response_packet(id, &peer_id_bytes, &records, ttl));
+            Err(e) => {
+                log::warn!("Excluding address {} from response: {:?}", addr, e);
             }
-
-            break 'next_packet
         }
-
-        // Reached MAX_RECORDS_PER_PACKET. Finish up the current packet and
-        // begin a new one.
-        packets.push(query_response_packet(id, &peer_id_bytes, &records, ttl));
+        if records.len() == MAX_RECORDS_PER_PACKET || addresses.len() == 0 {
+            packets.push(query_response_packet(id, &peer_id_bytes, &records, ttl));
+        }
     }
```

protocols/mdns/src/dns.rs
```diff
@@ -349,11 +394,8 @@ impl fmt::Display for MdnsResponseError {
             MdnsResponseError::TxtRecordTooLong => {
                 write!(f, "TXT record invalid because it is too long")
             }
-            MdnsResponseError::NonAsciiMultiaddr => write!(
-                f,
-                "A multiaddr contains non-ASCII characters when serializd"
-            ),
-            MdnsResponseError::ResponseTooLong => write!(f, "DNS response is too long"),
+            MdnsResponseError::NonAsciiMultiaddr =>
+                write!( f, "A multiaddr contains non-ASCII characters when serialized"),
```
mxinden, line 398:
```suggestion
                write!(f, "A multiaddr contains non-ASCII characters when serialized"),
```

mxinden (approved):
Thanks for the bug-fix as well as the overall improvements! From what I can tell, splitting responses into multiple packets is fine.

protocols/mdns/src/dns.rs
romanb: What do you think about https://github.com/libp2p/rust-libp2p/pull/1877/commits/397b25026db0ea6103790c2b0873c4b36c0d5d40? With your suggestion as-is, I think the `continue` in the `Ok` branch would effectively render `MAX_RECORDS_PER_PEER` ineffective.

mxinden: Thanks, looks good to me. Indeed I missed the `continue`.

mxinden (approved):

PR #1880, opened 2020-12-08
swarm: Remove NotifyHandler::All

mxinden:
### Suggestion

With this pull request I would like to suggest removing the ability of a `NetworkBehaviour` to notify all `ProtocolHandler`s for a given peer with a single `NetworkBehaviourActionEvent`. More concretely I am suggesting to remove `NotifyHandler::All`. An
implementor of `NetworkBehaviour` can still notify all `ProtocolHandler`s for a given peer by emitting one `NotifyHandler` event per connection to that peer.

### Reasoning

The general ability to notify all handlers for a peer forces events send from the behaviour to the handler to be `Clone`. Reason is that the swarm needs to forward a clone of the event to each handler. A behaviour that always sends events to a single peer only still has to guarantee that that event is `Clone`.

### Example 

As a concrete example https://github.com/libp2p/rust-libp2p/pull/1838/ implementing the libp2p relay spec needs to send substreams between `ProtocolHandler`s. Substreams are always send to concrete `ProtocolHandler`s, never to all `ProtocolHandler`s. Today a substream needs to be wrapped in an `Arc<Mutex<_>>` (or similar) in order to be `Clone` before one can send it to another `ProtocolHandler`. Removing the ability to send to all `ProtocolHandler`s would relief the relay implementation of having to wrap each substream before sending.

### Alternative

Today the `ExpandedSwarm` requires `TInEvent` to be `Clone` in order to be able to notify all handlers for a given peer. Instead of removing the feature one could as well introduce a second `TInEvent2` which would not require clone, using the former to notify all handlers and the latter to notify specific handlers. A behaviour only requiring the latter could set the former to `()`. In my eyes this alternative would not be worth the complexity it introduces.

### Users

As far as I can tell `NotifyHandler::All` is not used in rust-libp2p nor Substrate. Does anyone know of any users of `NotifyHandler::All`?

**Is anyone objecting to this change or has alternative suggestions?**

### DIFF
--- a/protocols/kad/src/handler.rs
+++ b/protocols/kad/src/handler.rs
@@ -329,7 +329,7 @@ impl From<ProtocolsHandlerUpgrErr<io::Error>> for KademliaHandlerQueryErr {
 }
 
 /// Event to send to the handler.
-#[derive(Debug, Clone)]
+#[derive(Debug)]
 pub enum KademliaHandlerIn<TUserData> {
     /// Resets the (sub)stream associated with the given request ID,
     /// thus signaling an error to the remote.
@@ -429,7 +429,7 @@ pub enum KademliaHandlerIn<TUserData> {
 
 /// Unique identifier for a request. Must be passed back in order to answer a request from
 /// the remote.
-#[derive(Debug, Clone, PartialEq, Eq)]
+#[derive(Debug, PartialEq, Eq)]
 pub struct KademliaRequestId {
     /// Unique identifier for an incoming connection.
     connec_unique_id: UniqueConnecId,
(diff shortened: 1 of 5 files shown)


romanb (approved):
I am in favor as this is a good simplification. For the record here, `NotifyHandler::All` was [explicitly introduced for substrate](https://github.com/libp2p/rust-libp2p/pull/1440#issuecomment-590810941). So if it is no longer used there, that is a good thing. I just hope no-one else relies on it for its protocols and hence needs to adapt the code accordingly.

romanb:
@mxinden Feel free to go ahead with this PR at your convenience.

mxinden:
@tomaka any objections to this change? Are you aware of any (Polkadot) protocols that plan to use `NotifyHandler::All`?

tomaka:
Go ahead, this is a positive change to me.

Issue #1882, opened 2020-12-09
protocols/request-response: Consider introdcuing InboundFailure::ConnectionClosed


mxinden:
Today the request response `NetworkBehaviour` might not always emit an `InboundFailure` for all inbound requests that failed to be answered. More specifically in case the connection to the peer closes after `RequestResponse::send_response` has been called the response will be dropped but no `InboundFailure` event will be emitted.

https://github.com/libp2p/rust-libp2p/blob/7985c81911866f0bb002eba562152f0ed00c759a/protocols/request-response/src/lib.rs#L143-L152

Also see https://github.com/paritytech/substrate/pull/7696#discussion_r539462082.

Consider tracking inbound requests in the `NetworkBehaviour` in order to be able to detect the above scenario and emit an `InboundFailure::ConnectionClosed` event.

PR #1886, opened 2020-12-11
protocols/request-response: Emit InboundFailure::ConnectionClosed

mxinden:
A user of libp2p-request-response is guaranteed to receive an additional event
after receiving a request via `RequestResponseEvent::Message`.

After receiving the request:

- If the user responds in time and the connection is still alive, the
user can expect a `ResponseSent`.

- If the user drops the response channel, the user can expect an
`InboundFailure::ResponseOmission`.

- If the user does not respond in time, the user can expect an
`InboundFailure::Timeout`.

Thus far the user did not receive an event when the connection to the
peer closes. With this commit:

- If the connection to the peer closes before the users calls
`send_response` or after the user calls `send_response` but before the
response can be send on the network, the user can expect an
`InboundFailure::ConnectionClosed`.

---

Closes https://github.com/libp2p/rust-libp2p/issues/1882.

Opening as draft pull request for now. I want to go over all the edge-cases once more with a fresh mind.

### DIFF
--- a/protocols/request-response/src/lib.rs
+++ b/protocols/request-response/src/lib.rs
@@ -758,7 +758,7 @@ struct Connection {
     /// been received on this connection and emitted via `poll` but have not yet
     /// been answered.
     pending_outbound_response: HashSet<RequestId>,
-    /// Pending inbound responses for previously send requests on this
+    /// Pending inbound responses for previously sent requests on this
     /// connection.
     pending_inbound_responses: HashSet<RequestId>
 }
--- a/protocols/request-response/src/lib.rs
+++ b/protocols/request-response/src/lib.rs
@@ -382,19 +382,19 @@ where
 
     /// Initiates sending a response to an inbound request.
     ///
-    /// Once the response has been successfully sent on the corresponding
-    /// connection, [`RequestResponseEvent::ResponseSent`] is emitted.
+    /// If the [`ResponseChannel`] is already closed due to a timeout or the
+    /// connection being closed, the response is returned as an `Err` for
+    /// further handling. Once the response has been successfully sent on the
+    /// corresponding connection, [`RequestResponseEvent::ResponseSent`] is
+    /// emitted. In all other cases [`RequestResponseEvent::InboundFailure`]
+    /// will be or has been emitted.
     ///
     /// The provided `ResponseChannel` is obtained from an inbound
     /// [`RequestResponseMessage::Request`].
-    pub fn send_response(&mut self, ch: ResponseChannel<TCodec::Response>, rs: TCodec::Response) {
-        // Fails either if the inbound upgrade timed out waiting for the
-        // response in which case the handler emits
-        // `RequestResponseHandlerEvent::InboundTimeout` which in turn results
-        // in `RequestResponseEvent::InboundFailure`, or the connection itself
-        // closed which results in a `RequestResponseEvent::InboundFailure` as
-        // well.
-        let _ = ch.sender.send(rs);
+    pub fn send_response(&mut self, ch: ResponseChannel<TCodec::Response>, rs: TCodec::Response)
+        -> Result<(), TCodec::Response>
+    {
+        ch.sender.send(rs)
     }
 
     /// Adds a known address for a peer that can be used for
--- a/protocols/request-response/src/lib.rs
+++ b/protocols/request-response/src/lib.rs
@@ -478,40 +478,46 @@ where
 
     /// Remove pending outbound response for the given peer and connection.
     ///
-    /// If `expected` is true function panics in debug mode iff connection is still established but
-    /// request is unknown.
+    /// Returns `true` if the provided connection to the given peer is still
+    /// alive and the [`RequestId`] was previously present and is now removed.
+    /// Returns `false` otherwise.
     fn remove_pending_outbound_response(
         &mut self,
         peer: &PeerId,
         connection: ConnectionId,
         request: RequestId,
-        expected: bool,
-    ) {
-        if let Some(connection) = self.connected.get_mut(peer)
-            .and_then(|cs| cs.iter_mut().find(|c| c.id == connection))
-        {
-            let removed = connection.pending_outbound_responses.remove(&request);
-            if expected {
-                debug_assert!(removed, "Expected request id to be present.")
-            }
-        }
+    ) -> bool {
+        self.get_connection_mut(peer, connection)
+            .map(|c| c.pending_outbound_responses.remove(&request))
+            .unwrap_or(false)
     }
 
     /// Remove pending inbound response for the given peer and connection.
     ///
-    /// Function panics in debug mode iff connection is still established but request is unknown.
+    /// Returns `true` if the provided connection to the given peer is still
+    /// alive and the [`RequestId`] was previously present and is now removed.
+    /// Returns `false` otherwise.
     fn remove_pending_inbound_response(
         &mut self,
         peer: &PeerId,
         connection: ConnectionId,
         request: &RequestId,
-    ) {
-        if let Some(connection) = self.connected.get_mut(peer)
-            .and_then(|connections| connections.iter_mut().find(|c| c.id == connection))
-        {
-            let removed = connection.pending_inbound_responses.remove(request);
-            debug_assert!(removed, "Expected request id to be present.")
-        }
+    ) -> bool {
+        self.get_connection_mut(peer, connection)
+            .map(|c| c.pending_inbound_responses.remove(request))
+            .unwrap_or(false)
+    }
+
+    /// Returns a mutable reference to the connection in `self.connected`
+    /// corresponding to the given [`PeerId`] and [`ConnectionId`].
+    fn get_connection_mut(
+        &mut self,
+        peer: &PeerId,
+        connection: ConnectionId,
+    ) -> Option<&mut Connection> {
+        self.connected.get_mut(peer).and_then(|connections| {
+            connections.iter_mut().find(|c| c.id == connection)
+        })
     }
 }
 
@@ -629,7 +635,11 @@ where
     ) {
         match event {
             RequestResponseHandlerEvent::Response { request_id, response } => {
-                self.remove_pending_inbound_response(&peer, connection, &request_id);
+                let removed = self.remove_pending_inbound_response(&peer, connection, &request_id);
+                debug_assert!(
+                    removed,
+                    "Expect request_id to be pending before receiving response.",
+                );
 
                 let message = RequestResponseMessage::Response { request_id, response };
                 self.pending_events.push_back(
@@ -643,9 +653,7 @@ where
                     RequestResponseEvent::Message { peer: peer.clone(), message }
                 ));
 
-                match self.connected.get_mut(&peer)
-                    .and_then(|cs| cs.iter_mut().find(|c| c.id == connection))
-                {
+                match self.get_connection_mut(&peer, connection) {
                     Some(connection) => {
                         let inserted = connection.pending_outbound_responses.insert(request_id);
                         debug_assert!(inserted, "Expect id of new request to be unknown.");
@@ -663,14 +671,19 @@ where
                 }
             }
             RequestResponseHandlerEvent::ResponseSent(request_id) => {
-                self.remove_pending_outbound_response(&peer, connection, request_id, true);
+                let removed = self.remove_pending_outbound_response(&peer, connection, request_id);
+                debug_assert!(removed, "Expect request_id to be pending before response is sent.");
 
                 self.pending_events.push_back(
                     NetworkBehaviourAction::GenerateEvent(
                         RequestResponseEvent::ResponseSent { peer, request_id }));
             }
             RequestResponseHandlerEvent::ResponseOmission(request_id) => {
-                self.remove_pending_outbound_response(&peer, connection, request_id, true);
+                let removed = self.remove_pending_outbound_response(&peer, connection, request_id);
+                debug_assert!(
+                    removed,
+                    "Expect request_id to be pending before response is omitted.",
+                );
 
                 self.pending_events.push_back(
                     NetworkBehaviourAction::GenerateEvent(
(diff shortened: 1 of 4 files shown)


mxinden:
This pull request is ready for a review.

protocols/request-response/src/lib.rs
```diff
@@ -382,26 +374,27 @@ where
...
-    /// for further handling. Once the response has been successfully sent
-    /// on the corresponding connection, [`RequestResponseEvent::ResponseSent`]
-    /// is emitted.
+    /// Once the response has been successfully sent on the corresponding
+    /// connection, [`RequestResponseEvent::ResponseSent`] is emitted.
     ///
     /// The provided `ResponseChannel` is obtained from an inbound
     /// [`RequestResponseMessage::Request`].
-    pub fn send_response(&mut self, ch: ResponseChannel<TCodec::Response>, rs: TCodec::Response)
-        -> Result<(), TCodec::Response>
```
romanb, line 402: There is no harm in keeping this signature, is there? It can still be useful to get early feedback here (and to get the response back).

protocols/request-response/src/lib.rs
```diff
@@ -437,8 +430,9 @@ where
     /// Checks whether an outbound request initiated by
     /// [`RequestResponse::send_request`] is still pending, i.e. waiting
     /// for a response.
-    pub fn is_pending_outbound(&self, req_id: &RequestId) -> bool {
-        self.pending_responses.contains_key(req_id)
+    pub fn is_pending_outbound(&self, request_id: &RequestId) -> bool {
+        self.connected.iter()
+            .any(|(_, cs)| cs.iter().any(|c| c.pending_inbound_responses.contains(request_id)))
```
romanb, line 435: I don't think that is a practical option, as going through all connections is very inefficient. The method was only provided on the API because it could be implemented efficiently. Short of removing it, we could, for example, require a `PeerId` to be given together with the `RequestId`. Then we could offer symmetrically `is_pending_inbound()`. What do you think?

protocols/request-response/src/lib.rs
```diff
@@ -454,23 +448,63 @@ where
...
+    fn remove_pending_outbound_response(
+        &mut self,
+        peer: &PeerId,
+        connection: ConnectionId,
+        request: RequestId,
+        expected: bool,
+    ) {
+        if let Some(connection) = self.connected.get_mut(peer)
+            .map(|cs| cs.iter_mut().find(|c| c.id == connection))
+            .flatten()
```
romanb, line 482:
```suggestion
            .and_then(|cs| cs.iter_mut().find(|c| c.id == connection))
```

protocols/request-response/src/lib.rs
```diff
@@ -454,23 +448,63 @@ where
...
+    /// Function panics in debug mode iff connection is still established but request is unknown.
+    fn remove_pending_inbound_response(
+        &mut self,
+        peer: &PeerId,
+        connection: ConnectionId,
+        request: &RequestId,
+    ) {
+        if let Some(connection) = self.connected.get_mut(peer)
+            .map(|connections| connections.iter_mut().find(|c| c.id == connection))
+            .flatten()
```
romanb, line 502:
```suggestion
            .and_then(|connections| connections.iter_mut().find(|c| c.id == connection))
```

protocols/request-response/src/lib.rs
```diff
@@ -515,33 +549,61 @@ where
...
-                connections.remove(pos);
+    fn inject_connection_closed(&mut self, peer_id: &PeerId, conn: &ConnectionId, _: &ConnectedPoint) {
+        let connections = match self.connected.get_mut(peer_id) {
+            Some(peer) => peer,
+            None => {
+                debug_assert!(
+                    false,
+                    "Expected some established connection to peer before closing.",
+                );
+                return
```
romanb, line 565: For the sake of simplicity, I think I would just prefer an `.expect(...)` then, which also panics in release mode.

protocols/request-response/src/lib.rs
```diff
@@ -586,15 +649,39 @@ where
                 let channel = ResponseChannel { request_id, peer: peer.clone(), sender };
                 let message = RequestResponseMessage::Request { request_id, request, channel };
                 self.pending_events.push_back(NetworkBehaviourAction::GenerateEvent(
-                    RequestResponseEvent::Message { peer, message }
+                    RequestResponseEvent::Message { peer: peer.clone(), message }
                 ));
+
+                match self.connected.get_mut(&peer)
+                    .map(|cs| cs.iter_mut().find(|c| c.id == connection))
+                    .flatten()
```
romanb, line 657:
```suggestion
                    .and_then(|cs| cs.iter_mut().find(|c| c.id == connection))
```

protocols/request-response/src/lib.rs
```diff
@@ -670,4 +764,22 @@ const EMPTY_QUEUE_SHRINK_THRESHOLD: usize = 100;
 struct Connection {
     id: ConnectionId,
     address: Option<Multiaddr>,
+    /// Pending outbound responses where corresponding inbound requests have
+    /// been received on this connection and emitted via `poll` but have not yet
+    /// been answered.
+    pending_outbound_response: HashSet<RequestId>,
+    /// Pending inbound responses for previously send requests on this
```
romanb, line 771:
```suggestion
    /// Pending inbound responses for previously sent requests on this
```

protocols/request-response/src/lib.rs
mxinden: > There is no harm in keeping this signature, is there?

True. At the same time I find the error handling of `libp2p-request-response` with `send_response` returning a `Result` to be quite complex. E.g. when `send_response` returns `Ok(())`, one can not assume that sending the response succeeded, but instead only assume that the corresponding connection upgrade is still alive. While one can document this edge case I would prefer usage of `libp2p-request-response` not to require any knowledge about it internals. By not returning a `Result` there is only a single way of error reporting i.e. through the events emitted by the behaviour.

> It can still be useful to get early feedback here (and to get the response back).

To be honest I did not properly consider this use-case. Thinking about it some more I don't think returning the failed response payload would be helpful for Substrate. If a response fails, I would expect the remote to initiate a new request and I think keeping the response around until the remote retries by sending a new request introduces unnecessary complexity. This of course is just one data point. Can you think of a use-case where keeping the response around would be helpful?

The above said, this is not a strong opinion. Off the top of my head I would favor simplicity over complexity, especially as one can still change `send_response` to return a `Result` later on.

What do you think in regards to the above reasoning @romanb?

protocols/request-response/src/lib.rs
mxinden: :bulb: thanks!

protocols/request-response/src/lib.rs
mxinden: :+1: for the sake of consistency I will replace the `debug_assert` both for expecting to be connected to the peer and expecting an established connection for the given connection id right below it.

protocols/request-response/src/lib.rs
romanb: >  Can you think of a use-case where keeping the response around would be helpful?

Any situation where computing parts of the response may have been expensive and one may want to store or send these parts to some other peer, without having to reconstruct them. This may be contrived, it just seems to be a shame to throw away information (both the response and the knowledge about whether the inbound request already failed) at a point where it could be given. Note that initially, I actually did the same thing you've done here, likely for very similar reasons. The `Result` return value is a rather recent addition of mine which would be reverted again here. I am fine either way, I don't think I can make a strong case for keeping it at the moment.

protocols/request-response/src/lib.rs
```diff
@@ -571,12 +624,13 @@ where
...
         &mut self,
         peer: PeerId,
-        _: ConnectionId,
+        connection: ConnectionId,
         event: RequestResponseHandlerEvent<TCodec>,
     ) {
         match event {
             RequestResponseHandlerEvent::Response { request_id, response } => {
-                self.pending_responses.remove(&request_id);
+                self.remove_pending_inbound_response(&peer, connection, &request_id);
```
romanb, line 632: So the assumption that justifies panicking in `remove_pending_inbound_response` and `remove_pending_outbound_response` if the `request_id` is absent is that `inject_event` is never expected to be called with a `ConnectionId` after `inject_connection_closed` has already been called for that same connection, right? Wouldn't it be a bit clearer to move the `debug_assert!`s to the call-sites here, clarifying the expectations there and letting these `remove` methods return a `bool`, avoiding the extra `bool` parameter for `remove_pending_outbound_response`? Telling a method whether it should make a debug assertion or not seems a bit weird.

protocols/request-response/src/lib.rs
```diff
@@ -670,4 +754,22 @@ const EMPTY_QUEUE_SHRINK_THRESHOLD: usize = 100;
 struct Connection {
     id: ConnectionId,
     address: Option<Multiaddr>,
+    /// Pending outbound responses where corresponding inbound requests have
+    /// been received on this connection and emitted via `poll` but have not yet
+    /// been answered.
+    pending_outbound_response: HashSet<RequestId>,
```
romanb, line 760:
```suggestion
    pending_outbound_responses: HashSet<RequestId>,
```

protocols/request-response/src/lib.rs
```diff
@@ -604,17 +681,19 @@ where
...
+                self.pending_events.push_back(
+                    NetworkBehaviourAction::GenerateEvent(
+                        RequestResponseEvent::OutboundFailure {
+                            peer,
+                            request_id,
+                            error: OutboundFailure::Timeout,
+                        }));
             }
             RequestResponseHandlerEvent::InboundTimeout(request_id) => {
+                self.remove_pending_outbound_response(&peer, connection, request_id, false);
```
romanb, line 695: What justifies passing `false` here and hence expecting that `remove_pending_outbound_response` may have already been called earlier?

romanb (commented):
Generally looks good to me, thanks. I just left a few last questions/comments regarding the use of the debug assertions.

protocols/request-response/src/lib.rs
mxinden: `remove_pending_inbound_response` and `remove_pending_outbound_response` only panic in debug mode if the `request_id` is absent AND the connection is still alive. If the connection is not alive anymore, i.e. `inject_connection_closed` was called before `inject_event`, `remove_pending_inbound_response` and `remove_pending_outbound_response` do not panic when the `request_id` is absent.

> `inject_event` is never expected to be called with a `ConnectionId` after `inject_connection_closed` has already been called for that same connection, right?

At the time of writing the two functions I was not sure this is a guarantee the `Network` provides. Taking another look it does seem to be the case. Is this a guarantee that we want or can uphold in the future @romanb? If so I think your suggestion above of moving the `debug_assert` to the caller side simplifies the logic.

protocols/request-response/src/lib.rs
mxinden: If I am not mistaken `RequestResponseHandlerEvent::InboundTimeout` can be emitted for two scenarios:

1. An inbound request times out while reading the request. An `RequestResponseHandlerEvent::InboundTimeout` is emitted. In this case a `RequestResponseHandlerEvent::Request` is never emitted and thus the request is never added to the `connection.pending_outbound_response`.

2. Reading an inbound request succeeded. The request was passed up via `RequestResponseHandlerEvent::Request` and added to `connection.pending_outbound_response`. The response is not provided in time and thus a `RequestResponseHandlerEvent::InboundTimeout` is emitted.

Given that in the first case the request was never added to `connection.pending_outbound_response`, one can not expect it to be present when receiving an `RequestResponseHandlerEvent::InboundTimeout`. For that reason `false` is passed to `remove_pending_outbound_response`.

Is the above reasoning sound?

protocols/request-response/src/lib.rs
romanb: > Taking another look it does seem to be the case. Is this a guarantee that we want or can uphold in the future @romanb?

Yes, I think that `ConnectionClosed` is the last event emitted for a connection (and hence `inject_connection_closed` being invoked) is an elementary property.

protocols/request-response/src/lib.rs
romanb: Makes sense, it may become a lot clearer if the expectation of the debug assertion is stated here.

protocols/request-response/src/lib.rs
mxinden: https://github.com/libp2p/rust-libp2p/pull/1886/commits/324b2d7f9779026840d84d8023f19a621829d4dd reverts the change to the signature of `send_response` and in the doc comment stresses the point that a response can still fail after `send_response` returned `Ok`.

Thanks for bringing this up!

protocols/request-response/src/lib.rs
mxinden: Great. As said above this allows some simplifications done with https://github.com/libp2p/rust-libp2p/pull/1886/commits/bb310990a7bd3cbc46a8d8e1f67450d69c6fc899. Thanks for thinking this through!

As an aside I am not 100 % happy with the use of `debug_assert` in some and `expect` in other places. I would be fine with replacing these `debug_assert` calls with `expect` calls, but I am not sure which _panic-policy_ you have been following within rust-libp2p.

protocols/request-response/src/lib.rs
mxinden: Given that one can not assert the request id to be present before removing it, there is no `debug_assert` following this line. I did add a comment though describing why one can not assert the request id to be present before removing it (see https://github.com/libp2p/rust-libp2p/commit/bb310990a7bd3cbc46a8d8e1f67450d69c6fc899).

romanb (approved):

protocols/request-response/src/lib.rs
romanb: My only panic-policy is "Don't panic" :lotus_position:. I think `expect` are stronger invariants where execution cannot continue if they are violated, whereas with `debug_assert!`s that is not necessarily the case, even though they almost always indicate a bug when violated, of course.

mxinden:
Thanks for the thorough review @romanb.

PR #1908, opened 2021-01-06
*: Switch futures_codec to asynchronous-codec

mxinden:
`futures-codec` has not been updated in the recent months. It still
depends on `bytes` `v0.5` preventing all downstream dependencies to
upgrade to `bytes` `v1.0`.

This commit replaces `futures_codec` in favor of `asynchronous-codec`
The latter is a fully upgraded fork of the former.

Closes https://github.com/libp2p/rust-libp2p/pull/1900. Depends on https://github.com/paritytech/unsigned-varint/pull/47.

See also https://github.com/matthunz/futures-codec/pull/51 and https://github.com/matthunz/futures-codec/pull/57.

In addition to the above, this pull request upgrades `rust-libp2p` to `bytes` `v1`.

### DIFF
--- a/protocols/identify/Cargo.toml
+++ b/protocols/identify/Cargo.toml
@@ -2,7 +2,7 @@
 name = "libp2p-identify"
 edition = "2018"
 description = "Nodes identifcation protocol for libp2p"
-version = "0.26.0"
+version = "0.27.0"
 authors = ["Parity Technologies <admin@parity.io>"]
 license = "MIT"
 repository = "https://github.com/libp2p/rust-libp2p"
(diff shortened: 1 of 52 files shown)


protocols/identify/Cargo.toml
```diff
@@ -11,10 +11,10 @@ categories = ["network-programming", "asynchronous"]
 
 [dependencies]
 futures = "0.3.1"
-libp2p-core = { version = "0.26.0", path = "../../core" }
-libp2p-swarm = { version = "0.26.0", path = "../../swarm" }
+libp2p-core = { version = "0.27.0", path = "../../core" }
+libp2p-swarm = { version = "0.27.0", path = "../../swarm" }
 log = "0.4.1"
-prost = "0.6.1"
+prost = "0.7"
```
romanb, line 17: So this supersedes https://github.com/libp2p/rust-libp2p/pull/1911 right?

romanb (approved):
Thanks for doing this. We should merge this before #1887, since you already made the version and changelog updates here.

protocols/identify/Cargo.toml
mxinden: Correct. I had to upgrade `prost` in this pull request as well, as otherwise `prost` and `unsigned-varint` / `asynchronous-codec` would use different versions of `bytes`, which in turn fails to compile.

Issue #1898, opened 2020-12-18
New get_closest_peers signature prevents me to use arbitrary key (as was available in 0.31 version)


izderadicka:
Hi, 
in 0.33 version kad method `get_closest_peers` changed signature:
```rust
pub fn get_closest_peers<K>(&mut self, key: K) -> QueryId
    where
        K: Into<kbucket::Key<K>> + Into<Vec<u8>> + Clone
```

With this signature I'm struggling how to submit arbitrary key -   like record::Key or Vec<8>.   From for kbucket::Key is implemented only for PeerId and Multihash.   Is this intentional?   How to use now with arbitrary key?

mxinden:
First off @izderadicka do I understand correctly that you are using `get_closest_peers` to find a node closest to a given key, but retrieve the actual value corresponding to the key out-of-band i.e. with something other than `libp2p-kad`?

For the record this change was introduced in https://github.com/libp2p/rust-libp2p/pull/1874, you can find more details in https://github.com/libp2p/rust-libp2p/pull/1874#discussion_r537415361.

Off the top of my head I would suggest adding a `From<Vec<u8>> for Key<Vec<u8>>`:

```diff
diff --git a/protocols/kad/src/kbucket/key.rs b/protocols/kad/src/kbucket/key.rs
index e8d1af3e..7768d99c 100644
--- a/protocols/kad/src/kbucket/key.rs
+++ b/protocols/kad/src/kbucket/key.rs
@@ -111,6 +111,16 @@ impl From<PeerId> for Key<PeerId> {
     }
 }
 
+impl From<Vec<u8>> for Key<Vec<u8>> {
+    fn from(v: Vec<u8>) -> Self {
+        let bytes = KeyBytes(Sha256::digest(v.as_ref()));
+        Key {
+            preimage: v,
+            bytes
+        }
+    }
+}
```

That would allow you to call `get_closest_peers` with a `Vec<u8>`:

```diff
diff --git a/protocols/kad/src/behaviour.rs b/protocols/kad/src/behaviour.rs
index 5cfd12ca..97105e78 100644
--- a/protocols/kad/src/behaviour.rs
+++ b/protocols/kad/src/behaviour.rs
@@ -560,6 +560,15 @@ where
     ///
     /// The result of the query is delivered in a
     /// [`KademliaEvent::QueryResult{QueryResult::GetClosestPeers}`].
+    ///
+    /// ```
[...]

PR #1909, opened 2021-01-06
Add From impl for kbucket::Key

izderadicka:
- From<record::Key>
- From<Vec<u8>
- this will enable to use additional types in kad.get_closest_peers as it was possible in pre 0.33 version

### DIFF
--- a/protocols/kad/src/kbucket/key.rs
+++ b/protocols/kad/src/kbucket/key.rs
@@ -25,6 +25,8 @@ use sha2::digest::generic_array::{GenericArray, typenum::U32};
 use std::borrow::Borrow;
 use std::hash::{Hash, Hasher};
 
+use crate::record;
+
 construct_uint! {
     /// 256-bit unsigned integer.
     pub(super) struct U256(4);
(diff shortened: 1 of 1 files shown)


izderadicka:
PR for issue #1898

protocols/kad/src/kbucket/key.rs
```diff
@@ -25,6 +25,8 @@ use sha2::digest::generic_array::{GenericArray, typenum::U32};
 use std::borrow::Borrow;
 use std::hash::{Hash, Hasher};
```
mxinden, line 27:
```suggestion
```

I don't think we use the convention of separating crate level imports and external imports with a newline.

mxinden (approved):
Looks good to me. Thanks for the patch.

protocols/kad/src/kbucket/key.rs
izderadicka: NP - this is how rust-analyzer plugin in VS Code added it

romanb (approved):

mxinden:
For the record `libp2p-kad` `v0.27.1` [tagged](https://github.com/libp2p/rust-libp2p/releases/tag/libp2p-kad-v0.27.1) and [published](https://crates.io/crates/libp2p-kad/0.27.1).

PR #1910, opened 2021-01-09
protocols/gossipsub: remove RPC message size check

xu-cheng:
Since df7e73ec47b717c6167985486701962880e2d778, a too large sized RPC message will be fragmented. Therefore, the old message size check should be removed.

### DIFF
--- a/protocols/gossipsub/src/behaviour.rs
+++ b/protocols/gossipsub/src/behaviour.rs
@@ -589,15 +589,6 @@ where
             .into_protobuf(),
         );
 
-        // check that the size doesn't exceed the max transmission size
-        if event.encoded_len() > self.config.max_transmit_size() {
-            // NOTE: The size limit can be reached by excessive topics or an excessive message.
-            // This is an estimate that should be within 10% of the true encoded value. It is
-            // possible to have a message that exceeds the RPC limit and is not caught here. A
-            // warning log will be emitted in this case.
-            return Err(PublishError::MessageTooLarge);
-        }
-
         // Add published message to the duplicate cache.
         if !self.duplicate_cache.insert(msg_id.clone()) {
             // This message has already been seen. We don't re-publish messages that have already


mxinden (commented):
If I recall correctly this length check is still needed, given that one can not fragment a single message, but only a set of messages wrapped in a single rpc type.

This might be related to https://github.com/sigp/rust-libp2p/pull/93/files/73b4d9f21209214b0a93fff6f5c7b4b8295f9716#r528471904

@AgeManning would you mind taking a look?

AgeManning:
Yep @mxinden is correct.

A gossipsub RPC can contain many messages, control messages and subscriptions. If the total message is larger than the max transmit size the fragmentation splits these into to separate gossipsub RPCs. 

If the encoded message to be sent is larger than the maximum transmit size, then we cannot send that message and we can't fragment it. The check here is still relevant to catch these. 

@mxinden - Sorry even though you pinged me in that chain in the PR I seem to have missed it. The comments that say it is within 10% of the true value are old. They were related to a custom `encoded_len()` function I built to prevent doing the entire protobuf encoding in the behaviour. This has been changed and the comments should get removed in the next update.

xu-cheng:
Thanks for the explanation.

xu-cheng:
BTW, any chance to publish a new release of libp2p for gossipsub 1.1?

mxinden:
Sure. I will cut a new release of `unsigned-varint` (https://github.com/paritytech/unsigned-varint/pull/49), get that into `rust-libp2p` along with an upgrade to `bytes` and `tokio` and then prepare `libp2p` `v0.34.0`. In case it is merged in time, https://github.com/libp2p/rust-libp2p/pull/1887 would be included in the release as well. Sounds good @xu-cheng?

mxinden:
@xu-cheng I did a full release, including `libp2p-gossipsub`. See [`v0.34.0`](https://crates.io/crates/libp2p/0.34.0). Let us know in case anything else comes up.

PR #1912, opened 2021-01-12
Fix compilation with serde 1.0.119

tomaka:
I have no idea why, but the person who introduced this code decided to use `serde::export::Formatter`, which is a hidden export not supposed to be used, instead of `std::fmt::Formatter`, even though they're the same thing.

In serde 1.0.119, the `export` module is no longer public, which broke the build.

### DIFF
--- a/misc/multiaddr/src/onion_addr.rs
+++ b/misc/multiaddr/src/onion_addr.rs
@@ -1,7 +1,4 @@
-use std::borrow::Cow;
-use std::fmt::Debug;
-use serde::export::Formatter;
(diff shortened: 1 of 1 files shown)


romanb (approved):

romanb:
Published as `parity-multiaddr-0.10.1`.

sdbondi:
Thanks for the fix @tomaka, the auto-import (rust-analyser?) feature in my IDE doesn't prefer stdlib over re-exports - so no 'real' reason the serde re-export was used 😄

PR #1923, opened 2021-01-13
Backported serde build fix in #1912 to multiadd 0.7.3

sdbondi:
In serde 1.0.119, the export module is no longer public which breaks the build.

This PR is open to track changes within the backported branch and can be **closed without merging** once we 
are able to use the `parity-multiaddr` crate


sdbondi:
Oops - sorry wrong repo

Issue #1885, opened 2020-12-11
protocols/request-response: is_pending_outbound return true right after sending


mxinden:
Calling `is_pending_outbound` right after sending a request will return `false` if node is not yet connected to peer. `is_pending_outbound` only checks for requests already initiated on a connection, not requests where a connection still needs to be established.

Diff to reproduce:

```diff
diff --git a/protocols/request-response/tests/ping.rs b/protocols/request-response/tests/ping.rs
index de4d1983..257c9fc9 100644
--- a/protocols/request-response/tests/ping.rs
+++ b/protocols/request-response/tests/ping.rs
@@ -96,6 +96,7 @@ fn ping_protocol() {
         let addr = rx.next().await.unwrap();
         swarm2.add_address(&peer1_id, addr.clone());
         let mut req_id = swarm2.send_request(&peer1_id, ping.clone());
+        assert!(swarm2.is_pending_outbound(&req_id));
 
         loop {
             match swarm2.next().await {
@@ -302,4 +303,3 @@ impl RequestResponseCodec for PingCodec {
         write_one(io, data).await
     }
 }
```

whereistejas:
Hi,
As requested in the previous issue, I'm sharing my findings on how to resolve the issue before actually starting to code.

The method we intend to fix is the following:
```rust
    pub fn is_pending_outbound(&self, peer: &PeerId, request_id: &RequestId) -> bool {
        self.connected.get(peer)
            .map(|cs| cs.iter().any(|c| c.pending_inbound_responses.contains(request_id)))
            .unwrap_or(false)
    }
```

The problem is that this method only searches the `pending_inbound_responses` list for the provided the request ID. This fails to look for request IDs that may not have been established, as of, yet. Such request IDs can found in the `pending_outbound_requests` list. So, to resolve to the issue we also need to look for the provided request ID in the second list.

mxinden:
Exactly right @whereistejas. Could you create a pull request with the suggested fix, including the small amendment to the unit test, suggested in the initial description?

whereistejas:
Sure. In the mean, would you like to assign the issue to me?

whereistejas:
Hi,
While writing the code, I noticed that unlike in the case of `pending_inbound_responses` where we can select the particular request ID, this is not possible in the case of `pending_outbound_requests`.

Which led me to just check if the mentioned peer has a pending outbound reqeust.

[...]

PR #1928, opened 2021-01-15
is_pending_outbound fixes

whereistejas:
`is_pending_outbound` returns true if the connection
to the mentioned peer hasn't been established, yet.

Closes issue: #1885

Signed-off-by: Tejas Sanap <sanap.tejas@gmail.com>

### DIFF
--- a/protocols/request-response/src/lib.rs
+++ b/protocols/request-response/src/lib.rs
@@ -431,9 +431,14 @@ where
     /// [`PeerId`] initiated by [`RequestResponse::send_request`] is still
     /// pending, i.e. waiting for a response.
     pub fn is_pending_outbound(&self, peer: &PeerId, request_id: &RequestId) -> bool {
-        self.connected.get(peer)
+        // check if request id is exists in already established connections
+        let est_conn = self.connected.get(peer)
             .map(|cs| cs.iter().any(|c| c.pending_inbound_responses.contains(request_id)))
-            .unwrap_or(false)
+            .unwrap_or(false);
+        // check if peer exists in pending connections
+        let pen_conn = self.pending_outbound_requests.contains_key(peer);
+
+        est_conn || pen_conn
     }
 
     /// Checks whether an inbound request from the peer with the provided
--- a/protocols/request-response/src/lib.rs
+++ b/protocols/request-response/src/lib.rs
@@ -437,7 +437,7 @@ where
             .unwrap_or(false);
         // Check if request is still pending to be sent.
         let pen_conn = self.pending_outbound_requests.get(peer)
-            .map(|cs| cs.iter().any(|c| {c.request_id == *request_id}))
+            .map(|rps| rps.iter().any(|rp| {rp.request_id == *request_id}))
             .unwrap_or(false);
 
         est_conn || pen_conn
--- a/protocols/request-response/src/lib.rs
+++ b/protocols/request-response/src/lib.rs
@@ -431,12 +431,14 @@ where
     /// [`PeerId`] initiated by [`RequestResponse::send_request`] is still
     /// pending, i.e. waiting for a response.
     pub fn is_pending_outbound(&self, peer: &PeerId, request_id: &RequestId) -> bool {
-        // check if request id is exists in already established connections
+        // Check if request is already sent on established connection.
         let est_conn = self.connected.get(peer)
             .map(|cs| cs.iter().any(|c| c.pending_inbound_responses.contains(request_id)))
             .unwrap_or(false);
-        // check if peer exists in pending connections
-        let pen_conn = self.pending_outbound_requests.contains_key(peer);
+        // Check if request is still pending to be sent.
+        let pen_conn = self.pending_outbound_requests.get(peer)
+            .map(|cs| cs.iter().any(|c| {c.request_id == *request_id}))
+            .unwrap_or(false);
 
         est_conn || pen_conn
     }
--- a/protocols/request-response/tests/ping.rs
+++ b/protocols/request-response/tests/ping.rs
@@ -93,6 +93,7 @@ fn ping_protocol() {
         let addr = rx.next().await.unwrap();
         swarm2.add_address(&peer1_id, addr.clone());
         let mut req_id = swarm2.send_request(&peer1_id, ping.clone());
+        assert!(swarm2.is_pending_outbound(&peer2_id, &req_id));
 
         loop {
             match swarm2.next().await {
--- a/protocols/request-response/tests/ping.rs
+++ b/protocols/request-response/tests/ping.rs
@@ -93,7 +93,7 @@ fn ping_protocol() {
         let addr = rx.next().await.unwrap();
         swarm2.add_address(&peer1_id, addr.clone());
         let mut req_id = swarm2.send_request(&peer1_id, ping.clone());
-        assert!(swarm2.is_pending_outbound(&peer2_id, &req_id));
+        assert!(swarm2.is_pending_outbound(&peer1_id, &req_id));
 
         loop {
             match swarm2.next().await {
--- a/protocols/request-response/tests/ping.rs
+++ b/protocols/request-response/tests/ping.rs
@@ -106,7 +106,7 @@ fn ping_protocol() {
                     assert_eq!(&peer, &peer1_id);
                     assert_eq!(req_id, request_id);
                     if count >= num_pings {
-                       break 
+                        return
                     } else {
                         req_id = swarm2.send_request(&peer1_id, ping.clone());
                     }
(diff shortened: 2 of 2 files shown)


whereistejas:
Hi @mxinden , this PR is ready to be merged. I would really appreciate a review. Thanks!

protocols/request-response/src/lib.rs
```diff
@@ -431,9 +431,14 @@ where
     /// [`PeerId`] initiated by [`RequestResponse::send_request`] is still
     /// pending, i.e. waiting for a response.
     pub fn is_pending_outbound(&self, peer: &PeerId, request_id: &RequestId) -> bool {
-        self.connected.get(peer)
+        // check if request id is exists in already established connections
```
mxinden, line 434:
```suggestion
        // Check if request is already sent on established connection.
```

protocols/request-response/src/lib.rs
```diff
@@ -431,9 +431,14 @@ where
...
+        let est_conn = self.connected.get(peer)
             .map(|cs| cs.iter().any(|c| c.pending_inbound_responses.contains(request_id)))
-            .unwrap_or(false)
+            .unwrap_or(false);
+        // check if peer exists in pending connections
+        let pen_conn = self.pending_outbound_requests.contains_key(peer);
```
mxinden, line 439: Say you have two requests R1 and R2 all targeting peer P1. Say R1 is done and a result is returned to the upper layer. Say R2 is still in `self.pending_outbound_requests`.

If you call `is_pending_outbound(P1, R1)` you will get `true` even though R1 is already finished. You need to check the `request_id` in the `RequestProtocol` in `pending_outbound_request`.

protocols/request-response/src/lib.rs
```diff
@@ -431,9 +431,14 @@ where
     /// [`PeerId`] initiated by [`RequestResponse::send_request`] is still
     /// pending, i.e. waiting for a response.
     pub fn is_pending_outbound(&self, peer: &PeerId, request_id: &RequestId) -> bool {
-        self.connected.get(peer)
+        // check if request id is exists in already established connections
+        let est_conn = self.connected.get(peer)
             .map(|cs| cs.iter().any(|c| c.pending_inbound_responses.contains(request_id)))
-            .unwrap_or(false)
+            .unwrap_or(false);
+        // check if peer exists in pending connections
```
mxinden, line 438:
```suggestion
        // Check if request is still pending to be send.
```

mxinden (commented):
I think you missed a small edge case here, right?

protocols/request-response/src/lib.rs
whereistejas: Should we include this scenario in the test program, too? I think, we should.

whereistejas: I added a few more assert statements in ping.rs to cover this scenario. Thank you for bringing up this. :)

whereistejas: I'm trying to replicate this scenario. For some reason, I'm not able to do so.

I added a few print statements, to log stuff to the console, in the `is_pending_outbound` method.
```rust
let est_conn = self.connected.get(peer)
    .map(|cs| cs.iter().any(|c| c.pending_inbound_responses.contains(request_id)))
    .unwrap_or(false);

let pen_conn = self.pending_outbound_requests.contains_key(peer);

println!("request_id (from is_pending_outbound): {}", *request_id);
println!("est_conn && pen_conn: {} || {}", est_conn, pen_conn);
// print all the request ids in pending_outbound_requests for a given peer id
for value in self.pending_outbound_requests.get(peer) {
    for item in value {
        println!("request_id (from pending_outbound_requests): {}", item.request_id);
    }
}

est_conn || pen_conn
```

And, then I called the `is_pending_outbound` method in 'ping_protocol` method in `ping.rs` as follows:
```rust
let mut req_id = swarm2.send_request(&peer1_id, ping.clone());
let mut req_id_org = req_id.clone();
assert!(swarm2.is_pending_outbound(&peer1_id, &req_id));

let peer2 = async move {
    let mut count = 0;
    let addr = rx.next().await.unwrap();
    swarm2.add_address(&peer1_id, addr.clone());
    let mut req_id = swarm2.send_request(&peer1_id, ping.clone());
    let mut req_id_org = req_id.clone();
    assert!(swarm2.is_pending_outbound(&peer1_id, &req_id));

    //let mut req_id2 = swarm2.send_request(&peer1_id, ping.clone());
    loop {
        match swarm2.next().await {
            RequestResponseEvent::Message {
                peer,
                message: RequestResponseMessage::Response { request_id, response }
            } => {
                count += 1;
                assert_eq!(&response, &expected_pong);
                assert_eq!(&peer, &peer1_id);
                assert_eq!(req_id, request_id);

                // check if is_pending_outbound returns false for a finished request when we have already issued a new one
                if req_id != req_id_org {
                    assert!(swarm2.is_pending_outbound(&peer1_id, &req_id_org) == false);
                }
                
                req_id_org = req_id.clone();
                if count >= num_pings {
                   break 
                } else {
                    req_id = swarm2.send_request(&peer1_id, ping.clone());
                    assert!(swarm2.is_pending_outbound(&peer1_id, &req_id));
                }
            },
            e => panic!("Peer2: Unexpected event: {:?}", e)
        }
    }
```

On running this code, I get the following output:
```
request_id (from is_pending_outbound): 2
est_conn && pen_conn: true || false
request_id (from is_pending_outbound): 1
est_conn && pen_conn: false || false
request_id (from is_pending_outbound): 3
est_conn && pen_conn: true || false
request_id (from is_pending_outbound): 2
est_conn && pen_conn: false || false
request_id (from is_pending_outbound): 4
est_conn && pen_conn: true || false
request_id (from is_pending_outbound): 3
```

So, in every iteration of the loop, I'm checking if the previous request is outbound and it **always returns a false statement**. So, I guess, the edge case you mentioned doesn't occur? I highly suspect that I am doing something wrong.

protocols/request-response/tests/ping.rs
```diff
@@ -110,17 +110,15 @@ fn ping_protocol() {
...
                         req_id = swarm2.send_request(&peer1_id, ping.clone());
                     }
-                    println!("hi, {}", req_id);
                 },
                 e => panic!("Peer2: Unexpected event: {:?}", e)
             }
         }
 
-        // check if is_pending_outbound returns false for a finished request when we have already issued a new one
         let req_id2 = swarm2.send_request(&peer1_id, ping.clone());
```
whereistejas, line 118: hi @mxinden, could you give me some pointers on how to come up with a test for this scenario?

whereistejas:
Hi @mxinden ,
I have removed the incorrect tests I had added to `ping.rs`. I think this PR is ready to be merged. I will create a new issue for the tests and work on it, separately.

romanb (approved):
Looks good to me. Thanks!

mxinden (approved):
Thanks for fixing this. :+1: for tackling the additional unit test in a follow up pull request.

whereistejas:
> Thanks for fixing this.  for tackling the additional unit test in a follow up pull request.

Hi @mxinden, so we are done? Will the PR get merged on its own?

mxinden:
For the record, I pushed https://github.com/libp2p/rust-libp2p/commit/25ea8b67ff6b337b2c66ab9fcd044f6d7966698c updating changelogs and cargo tomls.

mxinden:
> hi @mxinden, could you give me some pointers on how to come up with a test for this scenario?

Instead of  extending an existing test, I would suggest creating a new one, only including a very minimal setup. Is the pseudo code below of some help @whereistejas?


```diff
diff --git a/protocols/request-response/tests/ping.rs b/protocols/request-response/tests/ping.rs
index b03931cd..5622fb06 100644
--- a/protocols/request-response/tests/ping.rs
+++ b/protocols/request-response/tests/ping.rs
@@ -274,6 +274,28 @@ fn ping_protocol_throttled() {
     pool.run_until(peer2);
 }
 
+#[test]
+fn is_pending_outbound() {
+    let ping = Ping("ping".to_string().into_bytes());
+    let offline_peer = PeerId::random();
+
+    let protocols = iter::once((PingProtocol(), ProtocolSupport::Full));
+    let cfg = RequestResponseConfig::default();
+
+    let (peer1_id, trans) = mk_transport();
+    let ping_proto1 = RequestResponse::throttled(PingCodec(), protocols.clone(), cfg.clone());
+    let mut swarm1 = Swarm::new(trans, ping_proto1, peer1_id.clone());
+
+    let request_id_1 = swarm1.send_request(offline_peer, ping.clone());
+
+    // Poll swarm until request 1 is reported as failed.
+
+    let request_id_2 = swarm1.send_request(offline_peer, ping.clone());
+
+    assert!(!swarm1.is_pending_outbound(offline_peers, request_id_1));
+    assert!(swarm1.is_pending_outbound(offline_peers, request_id_2));
+}
+
```

Issue #1926, opened 2021-01-15
Gossipsub: unable to re-publish message after failed attempts.


xu-cheng:
Sometimes, publishing message may fail in `Gossipsub` due to `PublishError::InsufficientPeers`. In this case, the application may want to retry the publish later. However, the subsequent re-publish would fail due to `PublishError::Duplicate` when a content-based hash is used to compute message id.

I think if a message fails to be published, it should be removed from `duplicate_cache` also maybe `mcache`.

mxinden:
@AgeManning @blacktemplar would you mind taking a look?

AgeManning:
Thanks for raising this edge case. 

This is an issue with content-addressed messages. I've updated the gossipsub protocol in #1930 to only add messages to the caches if they have been sent to at least one peer.

romanb:
Closed by https://github.com/libp2p/rust-libp2p/pull/1930.

PR #1930, opened 2021-01-19
[Gossipsub] Prevent non-published messages being added to caches

AgeManning:
This addresses issue #1926. 

Currently, Gossipsub messages may not be published on a topic if there are no subscribed peers on that topic, however they are being added to a duplicate cache. If the messages are content-addressed, this prevents the user from attempting to republish the message in the future. 

This PR modifies the logic such that messages are only added to the caches (duplicate, memcache and published_messages) if the message could be published to at least one peer. 

Previously, a failed publish would be added the memcache. This meant that the router would gossip about a failed publish, such that if a peer joined the topic after the failed publish (but within the gossip history time window) the message could still be propagated to the newly connected peers. With this PR, a failed publish now definitively indicates the message will not be published on the network and the user is required to re-submit at a later time, once sufficient peers exist on the topic.

### DIFF
--- a/protocols/gossipsub/src/behaviour.rs
+++ b/protocols/gossipsub/src/behaviour.rs
@@ -591,15 +591,11 @@ where
 
         // check that the size doesn't exceed the max transmission size
         if event.encoded_len() > self.config.max_transmit_size() {
-            // NOTE: The size limit can be reached by excessive topics or an excessive message.
-            // This is an estimate that should be within 10% of the true encoded value. It is
-            // possible to have a message that exceeds the RPC limit and is not caught here. A
-            // warning log will be emitted in this case.
             return Err(PublishError::MessageTooLarge);
         }
 
         // Add published message to the duplicate cache.
-        if !self.duplicate_cache.insert(msg_id.clone()) {
+        if self.duplicate_cache.contains(&msg_id) {
             // This message has already been seen. We don't re-publish messages that have already
             // been published on the network.
             warn!(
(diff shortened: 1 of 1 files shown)


protocols/gossipsub/src/behaviour.rs
```diff
@@ -591,15 +591,11 @@ where
...
         // check that the size doesn't exceed the max transmission size
         if event.encoded_len() > self.config.max_transmit_size() {
-            // NOTE: The size limit can be reached by excessive topics or an excessive message.
-            // This is an estimate that should be within 10% of the true encoded value. It is
-            // possible to have a message that exceeds the RPC limit and is not caught here. A
-            // warning log will be emitted in this case.
             return Err(PublishError::MessageTooLarge);
         }
 
         // Add published message to the duplicate cache.
```
xu-cheng, line 597: I think this comment should be updated accordingly.

xu-cheng:
Thanks for the fix.

protocols/gossipsub/src/behaviour.rs
AgeManning: Agreed

romanb (approved):
Looks good to me, thanks!

mxinden (approved):
Change looks good to me. Thanks.

> Previously, a failed publish would be added the memcache. This meant that the router would gossip about a failed publish, such that if a peer joined the topic after the failed publish (but within the gossip history time window) the message could still be propagated to the newly connected peers. With this PR, a failed publish now definitively indicates the message will not be published on the network and the user is required to re-submit at a later time, once sufficient peers exist on the topic.

That sounds reasonable to me.

mxinden:
For the record, I pushed https://github.com/libp2p/rust-libp2p/commit/bbd3df64ed4cc99a5ddac47ab53673161c70e6da to update changelogs and cargo tomls. Let us know in case you need a release with this patch.

Issue #1920, opened 2021-01-12
Integrate `/dnsaddr` support in `libp2p-dns`


romanb:
Follow-up to https://github.com/libp2p/rust-libp2p/issues/967. Initial draft at https://github.com/libp2p/rust-libp2p/pull/1450.

Issue #1462, opened 2020-02-18
Allow dialing multiaddr containing a peer id


rklaehn:
One minor difference that makes interop between rust-libp2p and go-libp2p difficult is that while in go-libp2p it is possible and default to dial to a multiaddr that contains a peer id, in rust-libp2p you can only dial to a multiaddr that contains just host/port.

Allow rust-libp2p to optionally dial full multiaddrs like e.g. `/ip4/192.168.178.20/tcp/4001/ipfs/QmaBUojUuCTMXTRw7yYqNscKZ7caAjFqGFJP3hqv3UH6Br`.

Related:
https://github.com/libp2p/rust-libp2p/issues/471

rklaehn:
I looked around a bit how this could be implemented. The place where a multiaddr is being parsed and split into individual components is e.g. in the tcp transport, multiaddr_to_socketaddr. It would be easy enough to extend this method to extract an optional peer id.

You could also add this peer id to the TcpConfig struct that is generated in that macro.

But then what? The place where we get the peer id from the remote is higher in the transport stack. Any recommendations how to get the peer id from the multiaddr to the place it can be validated against without having to extend the transport trait in an ugly way?

tomaka:
It's not the role of the TCP transport to care about what `PeerId` is expected.

The problem to me is that I see this feature as strictly inferior to what we have now, where you explicitly pass a `PeerId` by value at various places.
Putting the `PeerId` inside of the `Multiaddr` opens the gate to all sorts of weird corner cases.

Is there a situation in which multiaddresses with peerids are used in the other libp2p implementations apart from CLI purposes?

rklaehn:
Not sure about go-libp2p, but ipfs can not even dial a multiaddr that does *not* contain a peer id...

> It's not the role of the TCP transport to care about what PeerId is expected.

Well, that is basically my problem. I could hack the peer id into the TcpConfig and then grab it from a higher layer where you learn the peer id of the remote end, but it seems weird because it does not really seem to belong there. Why should the TcpTransport even know what a PeerId is...

Now what you would probably want to do if you wanted to support more complex multiaddrs would be that the tcp transport just grabs the front of the multiaddr and leaves the rest (could be a peer id, or a circuit relay address, or whatever) for the higher layers to figure out.

E.g. you got
```
[...]

PR #1931, opened 2021-01-21
[libp2p-dns] Implement `/dnsaddr` resolution.

romanb:
Closes https://github.com/libp2p/rust-libp2p/issues/1920, as well as, by necessity, https://github.com/libp2p/rust-libp2p/issues/1462.

This PR is based on https://github.com/libp2p/rust-libp2p/pull/1927 - a proper diff until that PR is merged can be [seen here](https://github.com/romanb/rust-libp2p/compare/trust-dns...romanb:dnsaddr).

In order to correctly resolve `/dnsaddr` addresses, the DNS transport needs to be given "fully qualified" multiaddresses when dialing, i.e. those that end with the `/p2p/...` protocol, since it chooses from the TXT records such addresses with a matching suffix of the address being dialed. So to pick the addresses for the correct peer, the `/p2p` protocol needs to be at the end of the address being dialed. This essentially goes back to #1462. So to address #1462 as well as implement `/dnsaddr` resolution,  we now make sure that dialing always uses such fully qualified addresses by appending the `/p2p` protocol as necessary in the `Network` of `libp2p-core` before giving the address for dialing to the transport. This is mostly transparent for the `PeerId`-based APIs, i.e. it is of course still possible to use a `PeerId` together with an "unqualified" address for dialing. It is just that at the end of the day, `Transport::dial()` always gets a fully-qualified address (when called from the `Network` anyway) and hence transports now  permit a trailing `/p2p` suffix (and in the case of the DNS transport, even make essential use of it).

As a side-effect, `Network::dial(addr)` and `Swarm::dial_addr(addr)` now also handle "fully-qualified" `/p2p` addresses, since the `Network` will then take the peer ID from the end of the address.

A related change as part of implementing `/dnsaddr` resolution is that `libp2p-dns` now actually tries to use all resolved addresses (well, up to a fixed maximum of dialing attempts).

The net result of these changes can be seen in the now cleaned up [`ipfs-kad` example](https://github.com/romanb/rust-libp2p/blob/dnsaddr/examples/ipfs-kad.rs), which can finally make use of `/dnsaddr/bootstrap.libp2p.io`.

### DIFF
--- a/protocols/relay/src/transport.rs
+++ b/protocols/relay/src/transport.rs
@@ -401,14 +401,12 @@ impl<T: Transport> Stream for RelayListener<T> {
                     Poll::Ready(Some(BehaviourToListenerMsg::IncomingRelayedConnection {
                         stream,
                         src_peer_id,
-                        relay_peer_id,
                         relay_addr,
+                        relay_peer_id: _
                     })) => {
                         return Poll::Ready(Some(Ok(ListenerEvent::Upgrade {
                             upgrade: RelayedListenerUpgrade::Relayed(Some(stream)),
-                            local_addr: relay_addr
-                                .with(Protocol::P2p(relay_peer_id.into()))
-                                .with(Protocol::P2pCircuit),
+                            local_addr: relay_addr.with(Protocol::P2pCircuit),
                             remote_addr: Protocol::P2p(src_peer_id.into()).into(),
                         })));
                     }
--- a/transports/dns/src/lib.rs
+++ b/transports/dns/src/lib.rs
@@ -301,7 +301,7 @@ where
             // attempt, return that error. Otherwise there were no valid DNS records
             // for the given address to begin with (i.e. DNS lookups succeeded but
             // produced no records relevant for the given `addr`).
-            Err(last_err.unwrap_or(
+            Err(last_err.unwrap_or_else(||
                 DnsErr::ResolveError(
                     ResolveErrorKind::Message("No matching records found.").into())))
         }.boxed().right_future())
@@ -483,7 +483,7 @@ fn invalid_data(e: impl Into<Box<dyn std::error::Error + Send + Sync>>) -> io::E
 }
 
 fn fqdn(name: &Cow<'_, str>) -> String {
-    if name.ends_with(".") {
+    if name.ends_with('.') {
         name.to_string()
     } else {
         format!("{}.", name)
--- a/transports/dns/src/lib.rs
+++ b/transports/dns/src/lib.rs
@@ -24,10 +24,11 @@
 //! [`DnsConfig`] and `TokioDnsConfig` for use with `async-std` and `tokio`,
 //! respectively.
 //!
-//! A [`GenDnsConfig`] is a [`Transport`] wrapper that is created around
+//! A [`GenDnsConfig`] is an address-rewriting [`Transport`] wrapper around
 //! an inner `Transport`. The composed transport behaves like the inner
-//! transport, except that [`Transport::dial`] resolves `/dns`, `/dns4/` and
-//! `/dns6/` components of a given `Multiaddr` through a DNS.
+//! transport, except that [`Transport::dial`] resolves `/dns/...`, `/dns4/...`,
+//! `/dns6/...` and `/dnsaddr/...` components of the given `Multiaddr` through
+//! a DNS, replacing them with the resolved protocols (typically TCP/IP).
 //!
 //! The `async-std` feature and hence the `DnsConfig` are
 //! enabled by default. Tokio users can furthermore opt-in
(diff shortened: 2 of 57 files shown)


romanb:
It looks like looking up TXT records from `_dnsaddr.bootstrap.libp2p.io` does not work on the build machine with the system DNS configuration.

mxinden:
> It looks like looking up TXT records from _dnsaddr.bootstrap.libp2p.io does not work on the build machine with the system DNS configuration.

This seems odd to me. Any suspicion why that might be the case?

As a data point, though I doubt a very helpful one, the `basic_resolve` unit test succeeds on my machine.

romanb:
> This seems odd to me. Any suspicion why that might be the case?

I don't know yet. I may have to do some noisy debugging on this branch for CI, for which I apologise in advance.

transports/dns/src/lib.rs
```diff
@@ -107,134 +171,150 @@ where
...
+                        Ok(Resolved::Many(ips)) => {
+                            for ip in ips {
+                                log::trace!("Resolved {} -> {}", name, ip);
+                                let addr = addr.replace(i, |_| Some(ip)).expect("`i` is a valid index");
+                                unresolved.push(addr);
+                            }
+                        }
+                        Ok(Resolved::Addrs(addrs)) => {
+                            if dnsaddr_lookups < MAX_DNSADDR_LOOKUPS {
+                                dnsaddr_lookups += 1;
```
mxinden, line 236: Say that a single `/dnsaddr` record contains a very large amount of bogus `dns6` addresses. Given that `dnsaddr_lookups` is only increased on successful `dnsaddr` lookups, the node would try to resolve each bogus `dns6` address possibly consuming a lot of resources.

Would it not be safer to keep count of the overall number of calls to `resolve()` and terminate the loop after some constant X?

romanb:
I changed the tests to be explicit about the name server(s) to use ([quad9](https://www.quad9.net/faq/) in this case). Besides solving the problem with the request timeouts for TXT lookups, I think it is better to avoid the indirection via the opaque system configuration for the tests.

transports/dns/src/lib.rs
romanb: > Say that a single /dnsaddr record contains a very large amount of bogus dns6 addresses.

I assume you mean when a single `/dnsaddr/...` lookup returns a lot of such TXT records. A single TXT record only contains a single address as per [the spec](https://github.com/multiformats/multiaddr/blob/master/protocols/DNSADDR.md) and thus we only process the first `<character-string>` of each TXT record, which can only be up to 255 bytes in length. There can be many TXT records in a reply, of course, but then there are also the size limits on DNS responses (anything > 512 bytes is already undesirable because, in the absence of [EDNS(0)](https://en.wikipedia.org/wiki/Extension_mechanisms_for_DNS) one has too fall back to TCP).

> Given that dnsaddr_lookups is only increased on successful dnsaddr lookups, the node would try to resolve each bogus dns6 address possibly consuming a lot of resources.

Not sure about whether that would consume a lot of resources, but I think counting the `resolve()` calls would be fine. Even beyond that, it may be preferable if we have a fixed limit on the number of TXT records that we process from a `/dnsaddr` lookup response, just to be on the safe side.

romanb: What do you think about https://github.com/libp2p/rust-libp2p/pull/1931/commits/5b99f128d412fb805929d4853bd7c85dee3450a5?

romanb: A small further refinement for better troubleshooting in case one runs into the limits: https://github.com/libp2p/rust-libp2p/pull/1931/commits/4e8d8f2d311808b82ca32f8d520f22efb3a99a8d.

mxinden: > I assume you mean when a single /dnsaddr/... lookup returns a lot of such TXT records.

Yes, sorry for the confusion and thanks for the detailed reply.

> There can be many TXT records in a reply, of course, but then there are also the size limits on DNS responses (anything 512 bytes is already undesirable because, in the absence of EDNS(0) one has too fall back to TCP).

More out of curiosity, even if the DNS server has to fall back to TCP to deliver the large DNS response, that does not defend against the attack of the response containing a large amount of TXT records, right? If I understand correctly trust-dns handles both UDP and TCP by default.

Limit enforcing via `MAX_DNS_LOOKUPS` and `MAX_TXT_RECORDS` look good to me.

mxinden (approved):
Changes on top of https://github.com/libp2p/rust-libp2p/pull/1927 look sound to me.

In my eyes your changes to `Swarm::dial_addr` and `Network::dial` prevent a lot of confusion (thinking back to my early days with rust-libp2p). Very much appreciated.

transports/dns/src/lib.rs
romanb: > More out of curiosity, even if the DNS server has to fall back to TCP to deliver the large DNS response, that does not defend against the attack of the response containing a large amount of TXT records, right? If I understand correctly trust-dns handles both UDP and TCP by default.

Yes, that is correct. It is just that due to the "expensive" TCP fallback, "honest" DNS responses rather tend to be small to avoid the TCP fallback, whilst for an attack returning large responses is probably a lot less interesting than returning false records, i.e. performing DNS spoofing / cache poisoning.

> Limit enforcing via MAX_DNS_LOOKUPS and MAX_TXT_RECORDS look good to me.

:+1: Thanks for reviewing.

romanb:
Still waiting on https://github.com/libp2p/rust-libp2p/pull/1927.

protocols/relay/src/transport.rs
```diff
@@ -401,14 +401,12 @@ impl<T: Transport> Stream for RelayListener<T> {
...
                         stream,
                         src_peer_id,
-                        relay_peer_id,
                         relay_addr,
+                        relay_peer_id: _
                     })) => {
                         return Poll::Ready(Some(Ok(ListenerEvent::Upgrade {
                             upgrade: RelayedListenerUpgrade::Relayed(Some(stream)),
-                            local_addr: relay_addr
-                                .with(Protocol::P2p(relay_peer_id.into()))
```
mxinden, line 410: :+1:

mxinden (approved):
New changes look good to me. Thanks for the additional `assert_eq`s.

Issue #1847, opened 2020-11-18
[multistream-select] Re-enable use of `ls` for the dialer aka "parallel negotiation".


romanb:
This is just a follow-up ticket for https://github.com/libp2p/rust-libp2p/issues/1795 to not forget about it. The wire format has been fixed and `ls` responses are now interoperable, but we still need to eventually re-enable the use of `ls` on the side of the dialer, i.e. make active use of it as before. Its use was temporarily disabled to permit an upgrade path. This should not be done before https://github.com/libp2p/rust-libp2p/pull/1811 has already been released.

PR #1934, opened 2021-01-25
[multistream-select] Re-enable 'parallel negotiation'.

romanb:
This PR re-enables the "parallel negotiation" strategy for the dialer in `multistream-select` based on an estimate of the number of alternative protocols the dialer supports. In this strategy the dialer sends an `ls` request straight away to avoid repeated protocol proposals and roundtrips. Closes https://github.com/libp2p/rust-libp2p/issues/1847.


mxinden (approved):
Change looks good to me.

As far as I can tell the last patch of the series (https://github.com/libp2p/rust-libp2p/pull/1811) is deployed with the recent Polkadot release [v0.8.27](https://github.com/paritytech/polkadot/releases/tag/v0.8.27).

I would expect it to be too early to break compatibility with versions < v0.8.27. What do you think? If so, I would suggest not yet merging this pull request until Polkadot v0.8.28 is released.

romanb:
> If so, I would suggest not yet merging this pull request until Polkadot v0.8.28 is released.

Sounds good to me.

mxinden:
With [Polkadot v0.8.28-1](https://github.com/paritytech/polkadot/releases/tag/v0.8.28-1) released, I think we can move forward here. Any objections @romanb @tomaka?

romanb:
I will go ahead with merging this today if there are no objections.

romanb:
Published in `multistream-select-0.10.2`.

PR #1936, opened 2021-01-26
Allow OneShotHandler's `max_dial_negotiate` limit to be configurable.

wngr:
This simply moves the `max_dial_negotiated` field of `OneShotHandler` into its config (and thus making it configurable). The default is set to the formerly hard-coded value of 8.

### DIFF
--- a/swarm/CHANGELOG.md
+++ b/swarm/CHANGELOG.md
@@ -1,3 +1,8 @@
+# 0.27.1 [unreleased]
+
+- Make `OneShotHandler`s `max_dial_negotiate` limit configurable.
+  [PR 1936](https://github.com/libp2p/rust-libp2p/pull/1936).
+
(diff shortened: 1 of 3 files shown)


romanb (approved):
Looks good to me, thanks!

mxinden (approved):

romanb:
Released as `libp2p-swarm-0.27.1`. Though I realised afterwards that the `OneShotHandlerConfig` struct is not yet `#[non_exhaustive]`, contrary to what I thought, so theoretically there may be code that needs to be updated to compile with this patch if it makes direct use of the `OneShotHandlerConfig` constructor, due to the new field. I don't think it is worth yanking for that reason though, as I think most code that uses the `OneShotHandler` either uses the default configuration or obtains it via the `Default` impl for subsequent customisation. Sorry for the oversight.

PR #1937, opened 2021-01-26
Fix DialPeerCondition::Always handling

koivunej:
Happened to test this `DialPeerCondition::Always` and found that it never works.

### DIFF
--- a/swarm/src/lib.rs
+++ b/swarm/src/lib.rs
@@ -691,12 +691,9 @@ where TBehaviour: NetworkBehaviour<ProtocolsHandler = THandler>,
                         this.behaviour.inject_dial_failure(&peer_id);
                     } else {
                         let condition_matched = match condition {
-                            DialPeerCondition::Disconnected
-                                if this.network.is_disconnected(&peer_id) => true,
-                            DialPeerCondition::NotDialing
-                                if !this.network.is_dialing(&peer_id) => true,
+                            DialPeerCondition::Disconnected => this.network.is_disconnected(&peer_id),
+                            DialPeerCondition::NotDialing => !this.network.is_dialing(&peer_id),
                             DialPeerCondition::Always => true,
-                            _ => false,
                         };
                         if condition_matched {
                             if ExpandedSwarm::dial(this, &peer_id).is_ok() {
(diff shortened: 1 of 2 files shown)


koivunej:
Not sure why the github UI is telling me my branch is out of date already, even if I cannot see any new commits after 6400719a. 

Looked around, there doesn't seem to be any tests to extend with the variant so didn't add them here either.

swarm/src/lib.rs
```diff
@@ -695,7 +695,8 @@ where TBehaviour: NetworkBehaviour<ProtocolsHandler = THandler>,
                                 if this.network.is_disconnected(&peer_id) => true,
                             DialPeerCondition::NotDialing
                                 if !this.network.is_dialing(&peer_id) => true,
-                            _ => false
+                            DialPeerCondition::Always => true,
+                            _ => false,
```
romanb, line 699: It may be a good idea to make the match exhaustive, which should be possible since `DialPeerCondition` is defined in the same crate and would have prevented this bug in the first place, e.g. something like
```rust
let condition_matched = match condition {
    DialPeerCondition::Disconnected => this.network.is_disconnected(&peer_id),
    DialPeerCondition::NotDialing => !this.network.is_dialing(&peer_id),
    DialPeerCondition::Always => true,
};
```

romanb (commented):
Thanks.

swarm/src/lib.rs
koivunej: Agreed, just a second.

koivunej:
Force pushed for having mixed up the parent of the first commit probably, which caused the branch out of date notification I was wondering.

romanb (approved):
We can probably put this out in a `libp2p-swarm` patch release straight away together with #1936. @mxinden What do you think?

mxinden (approved):
Thanks @koivunej for the fix!

> We can probably put this out in a libp2p-swarm patch release straight away together with #1936. @mxinden What do you think?

Sounds good to me. @romanb do you have time to prepare a new release? If not, I can do so later tonight or early tomorrow.

romanb:
> [..] do you have time to prepare a new release?

Sure, I will take care of it.

romanb:
Released as `libp2p-swarm-0.27.1`.

PR #1941, opened 2021-01-28
Move some crates.

romanb:
This PR is on top of https://github.com/libp2p/rust-libp2p/pull/1940 ([diff here](https://github.com/romanb/rust-libp2p/compare/rm-secio...romanb:mv-protocols)) and moves some crates in the directory structure. Specifically:

  * `protocols/plaintest` -> `transports/plaintext`
  * `protocols/deflate` -> `transports/deflate`
  * `protocols/noise` -> `transports/noise`
  * `protocols/pnet` -> `transports/pnet`

The reasoning being an attempt at clearer separation, whereby `protocols/` contains "application protocols" implemented as `NetworkBehaviour`s via `libp2p-swarm` whilst `transports/` contains transport protocols as well as upgrades for these that only depend on `libp2p-core` (if at all). All crates in `protocols/` depend on `libp2p-swarm` whereas none of the crates in `transports/` do. All of the crates in `transports/` can be used in the `Transport` construction and upgrade process. I made an attempt at documenting this structure in the updated `README` and also reflect it in the structure of the top-level `CHANGELOG`.


CHANGELOG.md
```diff
@@ -1,112 +1,130 @@
+
```
mxinden, line 1:
```suggestion
```

mxinden (approved):
Both restructurings as well as additional documentation looks good to me. Thanks!

PR #1945, opened 2021-02-04
swarm/src/toggle: Ignore listen upgr errors when disabled

mxinden:
A disabled `ToggleProtoHandler` can receive listen upgrade errors in the
following two cases:

1. Protocol negotiation on an incoming stream failed with no protocol being
   agreed on.

2. When combining `ProtocolsHandler` implementations a single `ProtocolsHandler`
   might be notified of an inbound upgrade error unrelated to its own upgrade
   logic. For example when nesting a `ToggleProtoHandler` in a
   `ProtocolsHandlerSelect` the former might receive an inbound upgrade error
   even when disabled.

`ToggleProtoHandler` should ignore the error in both of these cases.

Fixes https://github.com/paritytech/polkadot/issues/2373.


mxinden:
In case this patch finds approval, I would like to relase `libp2p-swarm` `v0.27.2` unless anyone objects.

romanb (approved):
:+1:  Feel free to cut releases after merging.

mxinden:
> releases

I would publish a new `libp2p-swarm` patch release. Is there some other crate you would like to be released in conjunction?

romanb:
> I planned to only publish a new libp2p-swarm patch release. Is there some other crate you would like to be released in conjunction?

Not necessarily. I only used plural in case you think it would be desirable to have a `libp2p` crate release that has the minimum bound for the `libp2p-swarm` patch version introduced with this PR.

tomaka (approved):
:+1:

mxinden:
`libp2p-swarm` [tagged](https://github.com/libp2p/rust-libp2p/tree/libp2p-swarm-0.27.2) and  [published](https://crates.io/crates/libp2p-swarm/0.27.2).

Issue #1942, opened 2021-01-28
support streaming responses in request-response protocol


dvc94ch:
currently if a request results in multiple responses you can't use the request-response protocol. I wonder how hard it would be to add a `final_response` method to the request response behaviour and keeping the channel open until the final response.

romanb:
My first impulse would be to have a dedicated crate for generic streaming-oriented protocols, where any number of "chunks" can be sent in either direction. `libp2p-request-response` is really designed for  RPC-like "1-request, 1-response" protocols and bending it to serve different needs may be too cumbersome and restrictive on what you can do. But this is just my "gut feeling" without having made a proper analysis or experiment with the necessary changes. This may be a duplicate of https://github.com/libp2p/rust-libp2p/issues/1727.

dvc94ch:
Closing as a duplicate for now, but many rpc api's support some kind of pubsub and it would be nice to reuse the underlying stream instead of opening a new one for each notification.

Issue #1727, opened 2020-09-02
Support for stream-oriented protocols


mzabaluev:
The trait infrastructure of `NetworkBehaviour`, `ProtocolsHandler` etc. is very generic, so stream-oriented protocols should be possible to implement. Streamed processing is useful for efficiently implementing transfers of bulky sequences of messages, where the receiving party can cancel the transfer early or apply back pressure to avoid buffering large amounts of data.
However, the utilities available in the library are geared towards unary request-response exchanges:

* The protocol handler in the `request_response` module encodes and decodes one message per substream. The substream I/O object is passed to the codec by reference, so it's not possible to take ownership of it to return a stream object as a decoded message.
* `OneShotHandler` could be a better fit, but its `connection_keep_alive` implementation only regards outbound substreams under negotiation when deciding whether to keep the connection alive. When stream or sink objects corresponding to protocol requests or responses are given to the `NetworkBehaviour` impl or to the application to use, the keep-alive logic should keep track of the substream handles that are in use.

It would be nice to have a stream-oriented equivalent, or possibly a generalization of, `RequestResponse`, so that the library users would not need to implement substream management on their own.

wngr:
Wouldn't it be possible to refactor the `request_response` behaviour in such a way, that a codec can read (respectively write) >= 1 message from the I/O object? This probably would mean though, that the current design using the `upgrade` mechanism to exchange messages won't work anymore, and some book-keeping would need to be done inside the `ProtocolsHandler`.

Anway, I very much agree with the need for a stream-oriented `RequestResponse` protocol, comparable to channels.

thomaseizinger:
Closing in favor of https://github.com/libp2p/rust-libp2p/issues/2657.

PR #1947, opened 2021-02-05
protocols/streaming-response

wngr:
this adds a `streaming-response` crate, which provides a `NetworkBehaviour` similar
to the `request-response` protocol to have a generic messaging mechanism
between two peers. The most notable difference to `request-response` and
the main motivation for its implementation is to provide an open-ended
streaming mechanism for the responses: a consumer's request can result
in any number of individual response frames, until said stream is
finalized. Conceptually, it is somewhat similar to grpc streaming.

Posting this here as a draft PR for discussion, whether you would be willing
to accept such a dedicated crate, or whether it would make sense to fold this into
`request-response`.

Related discussions #1727 #1942

### DIFF
--- a/protocols/streaming/src/handler.rs
+++ b/protocols/streaming/src/handler.rs
@@ -1,0 +1,306 @@
+use futures::future;
+use libp2p_core::{
+    upgrade::NegotiationError, InboundUpgrade, OutboundUpgrade, UpgradeError, UpgradeInfo,
+};
+use libp2p_swarm::{
+    protocols_handler::InboundUpgradeSend, KeepAlive, NegotiatedSubstream, ProtocolsHandler,
+    ProtocolsHandlerEvent, ProtocolsHandlerUpgrErr, SubstreamProtocol,
+};
+use smallvec::SmallVec;
+use std::{
+    collections::VecDeque,
+    convert::Infallible,
+    marker::PhantomData,
+    sync::{
+        atomic::{AtomicU64, Ordering},
+        Arc,
+    },
+    task::Poll,
+    time::Instant,
+};
+
+use crate::{
+    InboundStreamId, OutboundStreamId, StreamHandle, StreamingCodec, StreamingConfig,
+    EMPTY_QUEUE_SHRINK_THRESHOLD,
+};
+
+pub(crate) type RefCount = Arc<PhantomData<u8>>;
+#[derive(Debug)]
+pub enum StreamId {
+    Inbound(InboundStreamId),
+    Outbound(OutboundStreamId),
+}
+pub struct StreamingProtocolsHandler<T: StreamingCodec> {
+    config: StreamingConfig,
+    /// The current connection keep-alive.
+    keep_alive: KeepAlive,
+    /// Queue of events to emit in `poll()`.
+    pending_events: VecDeque<StreamingProtocolsHandlerEvent<T>>,
+    /// Outbound upgrades waiting to be emitted as an `OutboundSubstreamRequest`.
+    outbound: VecDeque<OutboundStreamId>,
+    /// Live streams.
+    open_streams: SmallVec<[(StreamId, RefCount); 2]>,
+    /// Counter for inbound stream ids.
+    inbound_stream_id: Arc<AtomicU64>,
+    /// A pending fatal error that results in the connection being closed.
+    pending_error: Option<ProtocolsHandlerUpgrErr<std::convert::Infallible>>,
+    _codec: PhantomData<T>,
+}
+
+impl<T: StreamingCodec> StreamingProtocolsHandler<T> {
+    pub(crate) fn new(inbound_stream_id: Arc<AtomicU64>, config: StreamingConfig) -> Self {
+        Self {
+            config,
+            keep_alive: KeepAlive::Yes,
+            pending_events: VecDeque::default(),
+            outbound: VecDeque::default(),
+            inbound_stream_id,
+            open_streams: Default::default(),
+            pending_error: None,
+            _codec: Default::default(),
+        }
+    }
+}
+
+#[derive(Debug)]
+pub enum StreamingProtocolsHandlerEvent<T: StreamingCodec> {
+    NewIncoming {
+        id: InboundStreamId,
+        stream: StreamHandle<T::Upgrade>,
+    },
+    StreamOpened {
+        id: OutboundStreamId,
+        stream: StreamHandle<T::Upgrade>,
+    },
+    InboundStreamClosed {
+        id: InboundStreamId,
+    },
+    OutboundStreamClosed {
+        id: OutboundStreamId,
+    },
+    OutboundTimeout(OutboundStreamId),
+    OutboundUnsupportedProtocols(OutboundStreamId),
+    InboundTimeout(InboundStreamId),
+    InboundUnsupportedProtocols(InboundStreamId),
+}
+
+#[derive(Debug, Default)]
+pub struct StreamingProtocol<T: StreamingCodec> {
+    _codec: PhantomData<T>,
+}
+impl<T: StreamingCodec> StreamingProtocol<T> {
+    fn new() -> Self {
+        Self {
+            _codec: Default::default(),
+        }
+    }
+}
+
+impl<T: StreamingCodec> UpgradeInfo for StreamingProtocol<T> {
+    type Info = T::Protocol;
+
+    type InfoIter = std::iter::Once<Self::Info>;
+
+    fn protocol_info(&self) -> Self::InfoIter {
+        // must start with '/'
+        std::iter::once(T::protocol_name())
+    }
+}
+
+impl<T: StreamingCodec> InboundUpgrade<NegotiatedSubstream> for StreamingProtocol<T> {
+    type Output = T::Upgrade;
+
+    type Error = std::convert::Infallible;
+
+    type Future = future::Ready<Result<Self::Output, Self::Error>>;
+
+    fn upgrade_inbound(self, io: NegotiatedSubstream, _: Self::Info) -> Self::Future {
+        future::ok(T::upgrade(io))
+    }
+}
+
+impl<T: StreamingCodec> OutboundUpgrade<NegotiatedSubstream> for StreamingProtocol<T> {
+    type Output = T::Upgrade;
+
+    type Error = std::convert::Infallible;
+
+    type Future = future::Ready<Result<Self::Output, Self::Error>>;
+
+    fn upgrade_outbound(self, io: NegotiatedSubstream, _: Self::Info) -> Self::Future {
+        future::ok(T::upgrade(io))
+    }
+}
+
+impl<T: StreamingCodec + Send + 'static> ProtocolsHandler for StreamingProtocolsHandler<T> {
+    type InEvent = OutboundStreamId;
+
+    type OutEvent = StreamingProtocolsHandlerEvent<T>;
+
+    type Error = ProtocolsHandlerUpgrErr<Infallible>;
+
+    type InboundProtocol = StreamingProtocol<T>;
+
+    type OutboundProtocol = StreamingProtocol<T>;
+
+    type InboundOpenInfo = InboundStreamId;
+
+    type OutboundOpenInfo = OutboundStreamId;
+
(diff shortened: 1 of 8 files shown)


mxinden (commented):
Thank you for the pull request @wngr.

I am in favor of adding a stream-oriented protocol abstraction to rust-libp2p.

> whether you would be willing to accept such a dedicated crate, or whether it would make sense to fold this into request-response.

Opposite [to @romanb gut feeling](https://github.com/libp2p/rust-libp2p/issues/1942#issuecomment-773230444), with the simple API interface (`request()`, `respond()`, `finish_response()`) you are proposing in this pull request in mind, I would expect adding stream-oriented responses to `libp2p-request-response` directly not to be too intrusive, put simply, replacing the `oneshot` between behaviour and handler with an mpsc channel. That said, my gut feeling also tells me that Roamn's gut feeling is usually right.

I will need to try adding stream-oriented responses to `libp2p-request-response` before giving a better review here. I might get around doing so in the next couple of weeks. In case someone else has time before that, I would appreciate a pull request with a proof-of-concept.

wngr:
IMHO the biggest design mismatch is the double use of the `SubstreamProtocol`'s timeout for both upgrading to the proper protocol and waiting for a response. For streaming based protocols, a response might take an arbitrary amount of time to be generated. So my first instinct would be to decouple those two. Once that is done, bolting the API I proposed in d716a43 should be straight-forward.
To be backwards compatible, another layer of timeout handling would need to be added though..

Thinking about this, I tend to come back to the same point: The inherent abstraction of the current `request-response` protocol is an RPC call. The caller expects a timely answer from a server (.. eerm peer). Whereas the `streaming-response` could be more considered to be a typed subscription to some service, with no protocol-level expectation about timely responses etc (apart from establishing the connection channel). So it's more directed than pubsub, but less rigid than an RPC call.

Not sure to which conclusion that leads, just leaving my thoughts here. Maybe it's better to have some duplication than try to come up with a silver bullet..

mxinden:
_Sorry for the slow response time here._

>  the biggest design mismatch is the double use of the SubstreamProtocol's timeout for both upgrading to the proper protocol and waiting for a response.

> Maybe it's better to have some duplication than try to come up with a silver bullet..

I tried patching `libp2p-request-response` to allow streamed responses to a single request which turned out harder than expected, thus I agree with the above. Sorry for leading the discussion astray.

In case there are multiple projects that could make use of streaming-response, I am happy to include it in rust-libp2p (//CC @thomaseizinger). In case there is only a single project, I would prefer for it to be maintained outside of the rust-libp2p mono-repo.

It might as well be worth considering making streaming-response more generic, in the sense of allowing streaming on both the requesting and the responding side. See @romanb's suggestion in https://github.com/libp2p/rust-libp2p/issues/1942#issuecomment-773230444:

> My first impulse would be to have a dedicated crate for generic streaming-oriented protocols, where any number of "chunks" can be sent in either direction on a single substream. 

This would allow both the communication pattern suggested in this pull request, as well as various other streaming based patterns.

wngr:
Thanks for your response nonetheless!

You have an idea of an API sketch for a generic duplex streaming protocol? Wouldn't that just be an opinionated behaviour that wraps some book-keeping into a common package:
* Connect to Peer, if not connected
* Keep alive connection (and provide some API to figure out which connections to keep open)
* Send `[u8]` to Peer with some metadata
* Receive `[u8]` from Peer with some metadata
* ..
And higher level APIs (like for example the API proposed in this PR) is layered on top?

Let me know if there is more interest in this direction. If not, I will be publishing the `streaming-response` behaviour as a separate crate.

mxinden:
> You have an idea of an API sketch for a generic duplex streaming protocol? Wouldn't that just be an opinionated behaviour that wraps some book-keeping into a common package:

Pretty much, yes. I would deem this useful as (A) a building block to build e.g. `streaming-response` or more complex streaming based protocols and (B) a way to prototype protocols without having to bother with the verbose `NetworkBehaviour` trait.

>  * Connect to Peer, if not connected
>
>  * Keep alive connection (and provide some API to figure out which connections to keep open)

:+1: 

>  * Send `[u8]` to Peer with some metadata
>
>  * Receive `[u8]` from Peer with some metadata

I was thinking of exposing an `AsyncRead + AsyncWrite` stream directly.

```rust
impl StreamingBehaviour {
  fn new_stream(&mut self, peer_id: PeerId, addr: Option<Multiaddr>) -> RequestId {
    // ...
  }
}

impl NetworkBehaviour for StreamingBehaviour {
    type OutEvent = StreamingEvent;
    // ...
}

enum StreamingEvent {
  NewStream {
    id: RequestId,
    stream: // something that implements AsyncRead + AsyncWrite and allows to be tracked to close an idle connection
  }
}

```


>  * And higher level APIs (like for example the API proposed in this PR) is layered on top?

Off the top of my head yes, though there might be many caveats that I am missing right now.

Take the above with a grain of salt. After all, I am not currently in need of a streaming based protocol abstraction myself, thus I am having difficulties coming up with a useful abstraction. In case a more generic abstraction like the one suggested above would be useful for multiple real-world protocols, I think this is worth pursuing. If not, I think your more specific `streaming-response` abstraction is the way to go.

What do you think @wngr?

wngr:
Yup, I think I'm gonna give this a try.

mxinden:
@wngr let me know once you would like another review here.

wngr:
Thanks, will do! The force push was accidental, nothing to see here ..

wngr:
I **finally** came around spending some time on this. It's pretty barebones, but I think convenient enough to be of broader use. I borrowed some things from the `request-response` protocol (dialing of peers, some house-keeping).
Handing out a _raw_ handle implementing `Async{Read,Write}` was a bit too untyped for me, that's why I added the `StreamingCodec` -- not sure whether it adds much though.
Ping @mxinden -- let me know if this goes into the direction you imagined.

protocols/streaming/tests/smoke.rs
```diff
@@ -0,0 +1,85 @@
...
+                }
+                SwarmEvent::ListenerError { error, .. } => panic!("{}", error),
+                SwarmEvent::Behaviour(libp2p_streaming::StreamingEvent::NewIncoming {
+                    peer_id,
+                    mut stream,
+                    ..
+                }) => {
+                    assert_eq!(peer_id, peer2_id);
+                    stream.write_all(b"Hello").await.unwrap();
+                    stream.flush().await.unwrap();
```
wngr, line 38: Smoke test. Direct usage of `Async{Read,Write}`.

protocols/streaming/tests/codec.rs
```diff
@@ -0,0 +1,120 @@
...
+async fn codec() -> anyhow::Result<()> {
+    setup_logger();
+
+    let (peer1_id, trans) = mk_transport();
+
+    let mut swarm1 = Swarm::new(trans, Streaming::<PingCodec>::default(), peer1_id);
+
+    let (peer2_id, trans) = mk_transport();
+    let mut swarm2 = Swarm::new(trans, Streaming::<PongCodec>::default(), peer2_id);
+
```
wngr, line 59: Example using a typed codec.

protocols/streaming/tests/streaming.rs
```diff
@@ -0,0 +1,149 @@
...
+                        stream.send(ResponseTicker(start.elapsed())).await.unwrap();
+                        sleep(Duration::from_millis(50)).await;
+                    }
+                    stream.close().await.unwrap();
+                });
+            }
+            SwarmEvent::ConnectionClosed { peer_id, .. } => {
+                assert_eq!(peer_id, peer1_id);
+                assert_eq!(rx.next().await.unwrap(), 10);
+                break;
```
wngr, line 139: Example demonstrating connection keep-alive.

mxinden:
Thanks @wngr! Review is in progress. Hope to finish sometime later today or next week.

protocols/streaming/src/handler.rs
```diff
@@ -0,0 +1,306 @@
...
+    task::Poll,
+    time::Instant,
+};
+
+use crate::{
+    InboundStreamId, OutboundStreamId, StreamHandle, StreamingCodec, StreamingConfig,
+    EMPTY_QUEUE_SHRINK_THRESHOLD,
+};
+
+pub(crate) type RefCount = Arc<PhantomData<u8>>;
```
mxinden, line 27: Do I understand correctly that the task of the handler is not woken up, when a `StreamHandle` with a `RefCount` is dropped? If so, a handler, if not polled due to some other event, might stay in `KeepAlive::Yes` forever, even though all its streams are closed, correct?

In `libp2p-relay` I have solved this with the help of a `oneshot`. Would that be an option here as well?

https://github.com/libp2p/rust-libp2p/blob/98bc5e64869d0d188384a91ba9287eff4eb27b0e/protocols/relay/src/handler.rs#L110-L119

protocols/streaming/src/lib.rs
```diff
@@ -0,0 +1,547 @@
...
+                        peer_id: *peer_id,
+                        id,
+                        error: OutboundFailure::ConnectionClosed,
+                    },
+                ));
+        }
+    }
+
+    fn inject_disconnected(&mut self, peer: &PeerId) {
+        self.connected.remove(peer);
```
mxinden, line 421: Given that `inject_connection_closed` above removes a peer from `connected` if no more connections exist and given that `inject_disconnected` is only ever called after `inject_connection_closed, this line should never be executed, correct? If so, how about at least adding a `debug_assert` here to make sure `inject_connection_closed` will properly clean up in the future.

https://github.com/libp2p/rust-libp2p/blob/98bc5e64869d0d188384a91ba9287eff4eb27b0e/swarm/src/lib.rs#L579-L583

protocols/streaming/src/handler.rs
```diff
@@ -0,0 +1,306 @@
...
+    pending_events: VecDeque<StreamingProtocolsHandlerEvent<T>>,
+    /// Outbound upgrades waiting to be emitted as an `OutboundSubstreamRequest`.
+    outbound: VecDeque<OutboundStreamId>,
+    /// Live streams.
+    open_streams: SmallVec<[(StreamId, RefCount); 2]>,
+    /// Counter for inbound stream ids.
+    inbound_stream_id: Arc<AtomicU64>,
+    /// A pending fatal error that results in the connection being closed.
+    pending_error: Option<ProtocolsHandlerUpgrErr<std::convert::Infallible>>,
+    _codec: PhantomData<T>,
```
mxinden, line 47:
```suggestion
```

Is this needed?

protocols/streaming/src/handler.rs
```diff
@@ -0,0 +1,306 @@
...
+        id: OutboundStreamId,
+    },
+    OutboundTimeout(OutboundStreamId),
+    OutboundUnsupportedProtocols(OutboundStreamId),
+    InboundTimeout(InboundStreamId),
+    InboundUnsupportedProtocols(InboundStreamId),
+}
+
+#[derive(Debug, Default)]
+pub struct StreamingProtocol<T: StreamingCodec> {
```
mxinden, line 88:
```suggestion
pub struct StreamingProtocol<T> {
```

Is this needed?

protocols/streaming/src/handler.rs
```diff
@@ -0,0 +1,306 @@
...
+    _codec: PhantomData<T>,
+}
+impl<T: StreamingCodec> StreamingProtocol<T> {
```
mxinden, line 91:
```suggestion
impl<T> StreamingProtocol<T> {
```

Same as above.

mxinden (commented):
Thanks @wngr!

Couple of suggestions below. I still need to take a deeper look at the test code.

wngr:
Thanks for your feedback @mxinden, appreciate it.

My focus in the past months has shifted quite a bit, so it's unlikely I will make much progress on this. Might get warmed up in the future, but no gain in keeping the PR open until then.

0x77dev:
Just to update the thread, the https://github.com/libp2p/rust-libp2p/pull/5027 has been merged into the master branch. You can find an easier method for establishing basic connections between peers in the [stream protocol example](https://github.com/libp2p/rust-libp2p/tree/master/examples/stream).

PR #1950, opened 2021-02-08
Make clippy "happy".

### LINKED ISSUE #1888: Remove redundant cloning of `PeerId`s
After https://github.com/libp2p/rust-libp2p/pull/1874 there are likely still unnecessary `.clone()` calls being made throughout the code base. These should be removed.

romanb:
Address all clippy complaints that are not purely stylistic (or even have corner cases with false positives). Ignore all "style" and "pedantic" lints. Closes #1888. There was quite a bit of unnecessary cloning, not just of peer IDs.

### DIFF
--- a/swarm/src/lib.rs
+++ b/swarm/src/lib.rs
@@ -383,7 +383,7 @@ where TBehaviour: NetworkBehaviour<ProtocolsHandler = THandler>,
                 let handler = me.behaviour.new_handler()
                     .into_node_handler_builder()
                     .with_substream_upgrade_protocol_override(me.substream_upgrade_protocol_override);
-                me.network.peer(peer_id.clone())
+                me.network.peer(*peer_id)
                     .dial(first, addrs, handler)
                     .map(|_| ())
                     .map_err(DialError::ConnectionLimit)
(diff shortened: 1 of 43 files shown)


swarm/src/lib.rs
```diff
@@ -407,8 +407,8 @@ where TBehaviour: NetworkBehaviour<ProtocolsHandler = THandler>,
     }
 
     /// Returns the peer ID of the swarm passed as parameter.
-    pub fn local_peer_id(me: &Self) -> &PeerId {
-        &me.network.local_peer_id()
+    pub fn local_peer_id(me: &Self) -> PeerId {
```
mxinden, line 410: Should we treat API changes like these as breaking changes?

mxinden (approved):
Thanks Roman for doing this!

swarm/src/lib.rs
romanb: Strictly speaking, yes, since it is a change of return type. We could of course keep having `Network::local_peer_id` and `Swarm::local_peer_id` return `&PeerId`, I may have done an unnecessary API change here.

romanb: I undid this unnecessary API change in https://github.com/libp2p/rust-libp2p/pull/1950/commits/db68488549e2f99e61919f2c8c93b17c68a8d41f. Thanks for bringing it up.

romanb:
Since none of the changes here affect the API or are otherwise noteworthy for a changelog, let alone require a release, I will merge this PR as-is.

PR #1953, opened 2021-02-12
Add Forest to users list

austinabell:
We've been on mainnet for a few weeks now, so seems reasonable to add here now


mxinden (approved):
> We've been on mainnet for a few weeks now, so seems reasonable to add here now

Congrats. Happy to see more users of rust-libp2p.

PR #1954, opened 2021-02-14
swarm/behaviour: Document inject_connected called for first only

mxinden:
`NetworkBehaviour::inject_connected` is called for the first established
connection to a peer only. See `swarm/src/lib.rs`:

```rust
this.behaviour.inject_connection_established(&peer_id, &connection.id(), &endpoint);
if num_established.get() == 1 {
    this.behaviour.inject_connected(&peer_id);
}
```

This commit adjusts the documentation accordingly.

### DIFF
--- a/swarm/src/behaviour.rs
+++ b/swarm/src/behaviour.rs
@@ -82,21 +82,21 @@ pub trait NetworkBehaviour: Send + 'static {
     /// address should be the most likely to be reachable.
     fn addresses_of_peer(&mut self, peer_id: &PeerId) -> Vec<Multiaddr>;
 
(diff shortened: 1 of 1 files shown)


romanb (approved):
Build errors seem to be due to an inappropriate `snow` patch release after upgrading `rand` whose traits are in the public API of `snow`.

mxinden:
> Build errors seem to be due to an inappropriate `snow` patch release after upgrading `rand` whose traits are in the public API of `snow`.

:+1: asked the author to yank the release and cut a minor release instead: https://github.com/mcginty/snow/pull/107#issuecomment-778838641

PR #1957, opened 2021-02-15
[release] Prepare v0.35

romanb:
This release mainly consolidates a couple of patch releases.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,11 +41,11 @@
 
 # `libp2p` facade crate
 
-## Version 0.35.0 [unreleased]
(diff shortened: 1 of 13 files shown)


mxinden (approved):
We could as well include yamux v0.9 (https://github.com/paritytech/yamux/pull/114). Not important. We might even be able to include it in a `libp2p-yamux` patch release.

romanb:
> We could as well include yamux v0.9 (paritytech/yamux#114). Not important. We might even be able to include it in a libp2p-yamux patch release.

I will happily make a `libp2p-yamux` and `libp2p` release once these PRs are in.

PR #1961, opened 2021-02-22
Link to the GitHub org instead of the project

thomaseizinger:
Numerous projects in our org use rust-libp2p. Link to the whole org instead of one specific repository.


mxinden (approved):
> Numerous projects in our org use rust-libp2p.

I am happy to hear that.

PR #1977, opened 2021-02-23
Don't poll network unnecessarily.

dvc94ch:
Currently we have customers with >100 nodes in their local network. We received reports that mdns was responsible for 25% to 75% of network traffic. As you can imagine our customers weren't very happy about that. This PR massively reduces the mdns bandwidth requirements.

Mdns needs two sockets, a send socket and a receive socket. The receive socket is listening on the mdns broadcast address. When you join a network and receive an `IfEvent::Up` event the send socket is used to send an mdns query to the broadcast address. Everyone listening on the broadcast address responds with their mdns records, including yourself, to the broadcast address, so everyone gets a fresh view of the network and the existing peers learn about your records.

To avoid the case where you join a network and your initial discovery message was lost I readded a timer that by default will send a mdns query to the broadcast address if there has been no incoming mdns queries in the last five minutes. This timeout is configurable.

~There is still one remaining issue and that is making `if-watch` pollable using manual futures. But that should minimally effect this PR other than requiring a version bump.~

~Companion PR https://github.com/ipfs-rust/if-watch/pull/7~


rkuhn (commented):
Looks good as far as I understand it!

romanb (commented):
Thanks for the PR. To summarise in my own words after a first review, this PR does the following:

  * It removes `service.rs` and the separate `MdnsService`, collapsing the (simplified) code into the `Mdns` behaviour while just moving other query types from `service.rs` to the new `query.rs` module.

  * It makes the MDNS record TTL for response records configurable, with a default of 5 minutes as was previously fixed in a constant.

  * It makes the interval at which multicast queries are sent out configurable with a default of 5 minutes, as opposed to the current frequency of 20 seconds, thereby additionally sending out a query whenever we join a new multicast group on some network interface to ensure timely discovery in most cases despite the long query interval. Lost queries are effectively compensated for by that regular query interval.

Does that sound about right as a summary?

I left some comments that probably need to be resolved but the direction looks good to me. I would prefer though if we keep the default query interval lower, say at 1 minute, documenting the fact that the larger your network the larger you may want to configure the query interval at the increased risk of delayed discoveries due to lost datagrams, but at the gain of less MDNS network traffic.

> This PR massively reduces the mdns bandwidth requirements.

It would be great for the record if you could put absolute numbers on these gains with these changes as observed on some networks, i.e. a rough before and after comparison.

protocols/mdns/src/behaviour.rs
```diff
@@ -149,138 +277,102 @@ impl NetworkBehaviour for Mdns {
...
-                        Poll::Ready((service, packet)) => {
-                            self.service = MdnsBusyWrapper::Free(service);
-                            packet
-                        },
-                        Poll::Pending => {
-                            self.service = MdnsBusyWrapper::Busy(fut);
-                            return Poll::Pending;
-                        }
+        // Poll receive socket.
+        while self.recv_socket.poll_readable(cx).is_ready() {
```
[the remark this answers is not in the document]
dvc94ch: > whether we can be informed about "read readiness" despite there not yet being a full datagram to read

I'm pretty sure that isn't possible. A linux datagram socket does not become readable until a datagram arrives and not when an ip packet arrives. I can try finding a reference.

> may result in missing wakeups (since now_or_never() does not poll with the cx context)

this is actually not possible, if now_or_never returns None, `poll_readable` is called again. the reason why it is done this way is because most `async` futures cancel and deregister the waker if they are dropped. `poll_readable` was added for exactly that reason, it prevents not getting wakeups.

> but it turns out the datagram we then try to send is too large for the available space in the socket send buffer

it's a datagram socket not a tcp socket. it is guaranteed that a 65k byte datagram (derived from a max udp payload len of u16 + the 8byte header) can be sent. I can also try to find a reference.  

TODO: find some references

examples/mdns-passive-discovery.rs
```diff
@@ -18,43 +18,45 @@
...
+    //
+    // For illustrative purposes, the ping protocol is configured to
+    // keep the connection alive, so a continuous sequence of pings
+    // can be observed.
+    let behaviour = Mdns::new(MdnsConfig::default()).await?;
+
+    // Create a Swarm that establishes connections through the given transport
+    // and applies the ping behaviour on each connection.
+    let mut swarm = Swarm::new(transport, behaviour, peer_id);
+    Swarm::listen_on(&mut swarm, "/ip4/0.0.0.0/tcp/0".parse()?)?;
```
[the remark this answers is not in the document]
dvc94ch: yes, you can't discover a listening address if you aren't listening on any sockets.

protocols/mdns/src/behaviour.rs
```diff
@@ -149,138 +277,102 @@ impl NetworkBehaviour for Mdns {
...
+                Ok(IfEvent::Down(inet)) => {
+                    if inet.addr().is_loopback() {
+                        continue;
+                    }
+                    if let IpAddr::V4(addr) = inet.addr() {
+                        log::trace!("leaving multicast on iface {}", addr);
+                        if let Err(err) = socket.leave_multicast_v4(&multicast, &addr) {
+                            log::error!("leave multicast failed: {}", err);
+                        } else {
+                            self.send_buffer.clear();
```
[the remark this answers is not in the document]
dvc94ch: maybe it shouldn't, but there was some send buffer clearing in the previous code.  I'll remove it thanks!

dvc94ch:
> It makes the interval at which multicast queries are sent out configurable with a default of 5 minutes, as opposed to the current frequency of 20 seconds, thereby additionally sending out a query whenever we join a new multicast group on some network interface to ensure timely discovery in most cases despite the long query interval. Lost queries are effectively compensated for by that regular query interval.

Well I'd actually make the default to never query the network. It is probably not required at all. Once we deploy this in production we can play with the settings but I suspect we could make the interval much larger.

The only reason why it may be needed is because udp is an unreliable transport so packets may be dropped without noticing it. I think the 20s interval was required before mainly because we didn't have if-watch and where therefore not notified of network changes.

protocols/mdns/src/behaviour.rs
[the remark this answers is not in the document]
romanb: > this is actually not possible, if now_or_never returns None, poll_readable is called again.

I see. I overlooked that the code is calling `poll_readable` within `while` :facepalm:. That makes it irrelevant whether there may even be false positives. Thanks for the clarification.

> it's a datagram socket not a tcp socket. it is guaranteed that a 65k byte datagram (derived from a max udp payload len of u16 + the 8byte header) can be sent. I can also try to find a reference.

I would be surprised if being notified of write readiness on a datagram socket would be so pessimistic, i.e. only signaling write readiness when a maximally-sized datagram can be sent. My understanding is that write readiness can be signaled whenever there is space in the socket send buffer, however much that may be, but I'm happy to be proven wrong.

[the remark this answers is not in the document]
rkuhn: I do recall that this was a contentious point on the LKML many years ago, but I stopped reading that list around 2012. I suspect that we must assume that the 64kiB send guarantee is not actually upheld by all systems (just like not all systems actually provide a monotonic clock even when they say so … ).

protocols/mdns/src/behaviour.rs
```diff
@@ -18,33 +18,80 @@
...
+
+pub struct MdnsConfig {
+    /// TTL to use for mdns records.
+    pub ttl: Duration,
+    /// Interval at which to poll the network for new peers. This isn't
+    /// necessary during normal operation but avoids the case that an
+    /// initial packet was lost and not discovering any peers until a new
+    /// peer joins the network. Receiving an mdns packet resets the timer
+    /// preventing unnecessary traffic.
+    pub interval: Duration,
```
mxinden, line 60:
```suggestion
    pub query_interval: Duration,
```

Would be more descriptive in my eyes.

romanb:
> I think the 20s interval was required before mainly because we didn't have if-watch and were therefore not notified of network changes.

I think so, too, and what is done here seems certainly more in the spirit of the libp2p MDNS spec, which states "When a peer starts (or detects a network change), it sends a query for all peers.". Nevertheless we currently have the notion of TTLs of the discovered peer records and that is why I would prefer to keep the query interval shorter than the default TTL - to avoid intermittent "expired" events that are immediately followed again by "discovered" events. Since the default TTL is 5 minutes, how about 4 minutes then for the default query TTL? Of course, since the TTL of a record is specified by the remote that sent it, and nodes are configured differently, the local query interval may still be larger than the received record TTL. If we feel like it, we could dynamically update the interval based on the shortest TTL received that did not yet expire, but I think that goes beyond the scope of this PR and we can leave that for another time, if there is interest.

dvc94ch:
My personal opinion is that ttl's should be ignored. The application should store all known addresses and discard them if a dial failure occurs. If the dial failure is temporarily, the address will be rediscovered at a later time. There is a need for most applications to have an address book of some sort (substrate implements it's own). I think we could consider adding a general behaviour for that to rust-libp2p that does the right thing for most cases. [0] is what ipfs-embed currently does, and users of ipfs-embed can add additional discovery mechanisms based on the gossipsub api (if you aren't using a dht you need a mechanism for peers to tell you about peers on their local subnet).

- [0] https://github.com/ipfs-rust/ipfs-embed/blob/master/net/src/peers.rs

protocols/mdns/src/behaviour.rs
```diff
@@ -107,6 +164,77 @@ impl Mdns {
     pub fn discovered_nodes(&self) -> impl ExactSizeIterator<Item = &PeerId> {
         self.discovered_nodes.iter().map(|(p, _, _)| p)
     }
+
+    fn inject_mdns_packet(&mut self, packet: MdnsPacket, params: &impl PollParameters) {
+        self.timeout.set_interval(self.interval);
```
mxinden, line 169: Say there is one very noisy node on the network broadcasting an mdns packet every 1 minute. With a timeout of e.g. 5 minutes configured locally, the timeout would never fire, thus the local node would never broadcast a query and thus addresses from other nodes would expire locally, correct?

In case I am not mistaken with the assumption above, I would see two ways forward:

- Do not reset the timeout, always sending out a query at each interval.
- Remove the notion of TTLs for address as suggested by David.

I have yet to put more thoughts into this, thus please feel free to ignore the comment.

mxinden (commented):
> It removes service.rs and the separate MdnsService, collapsing the (simplified) code into the Mdns behaviour while just moving other query types from service.rs to the new query.rs module.

I am in favor of merging the two.

protocols/mdns/src/behaviour.rs
dvc94ch: This isn't quite accurate. If a peer sends a query, all other peers respond with their addresses to the multicast address. So you really only need one peer to make the query, as everyone will get the updates from everyone. This means that they will not expire locally.

romanb:
> My personal opinion is that ttl's should be ignored. The application should store all known addresses and discard them if a dial failure occurs. If the dial failure is temporarily, the address will be rediscovered at a later time. There is a need for most applications to have an address book of some sort (substrate implements it's own). I think we could consider adding a general behaviour for that to rust-libp2p that does the right thing for most cases. [0] is what ipfs-embed currently does, and users of ipfs-embed can add additional discovery mechanisms based on the gossipsub api (if you aren't using a dht you need a mechanism for peers to tell you about peers on their local subnet).

That may be a good thing to do, I'd just prefer if we leave it to a possible follow-up PR, to not expand the scope here too much and to make reviewing of follow-up changes that remove the TTL easier to review. From my perspective as long as this PR keeps the query interval a bit lower than the TTL, it is already good to go.

protocols/mdns/src/behaviour.rs
romanb: I would prefer to not reset the timeout for now and leave the removal of the TTL to another PR.

dvc94ch: if the timeout is smaller than the ttl, it should still be safe to reset the timeout. as I said it uses multicast so everyone gets all queries and all responses.

romanb (approved):

protocols/mdns/src/behaviour.rs
mxinden: Thanks for clarifying @dvc94ch!

mxinden (approved):

romanb:
I updated the CHANGELOG and will merge soon. Thanks for the PR @dvc94ch and the second review @mxinden.

PR #1982, opened 2021-03-02
[core] Shorten and unify Debug impls of public keys.

### LINKED ISSUE #1939: Debug implementation of `PublicKey` is too verbose
Printing the `PublicKey` enum is in my opinion way to verbose when using the `Ed25519` variant, which leads to a lot of pollution in logs.

It uses this underlying `Debug` implementation: https://github.com/dalek-cryptography/ed25519-dalek/blob/925eb9ea56192053c9eb93b9d30d1b9419eee128/src/public.rs#L44
Which in turn uses this `Debug` implementation: https://github.com/dalek-cryptography/curve25519-dalek/blob/b307c155c966c16b81f63a5872a1ec1977829ddf/src/edwards.rs#L928-L929

Simply printing the hexadecimal public key is in my opinion more than enough.

romanb:
This PR is just a small update to the `Debug` impls of public key types. Specifically, ed25519 `PublicKey` and secp256k1 `PublicKey` get a new, shorter `Debug` impl, rather than a derived one (which dumps details that are usually not relevant for a libp2p user), while RSAs `PublicKey` is just aligned with the others for uniformity of the syntax. Closes #1939.

### DIFF
--- a/core/src/identity/ed25519.rs
+++ b/core/src/identity/ed25519.rs
@@ -102,8 +102,18 @@ impl From<SecretKey> for Keypair {
 }
 
 /// An Ed25519 public key.
-#[derive(PartialEq, Eq, Debug, Clone)]
+#[derive(PartialEq, Eq, Clone)]
 pub struct PublicKey(ed25519::PublicKey);
+
+impl fmt::Debug for PublicKey {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.write_str("PublicKey(compressed): ")?;
+        for byte in self.0.as_bytes() {
+            write!(f, "{:x}", byte)?;
+        }
+        Ok(())
+    }
+}
 
 impl PublicKey {
     /// Verify the Ed25519 signature on a message using the public key.
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -61,7 +61,7 @@ atomic = "0.5.0"
 bytes = "1"
 futures = "0.3.1"
 lazy_static = "1.2"
-libp2p-core = { version = "0.27.1", path = "core",  default-features = false }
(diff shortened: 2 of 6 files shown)


core/src/identity/ed25519.rs
```diff
@@ -102,9 +102,19 @@ impl From<SecretKey> for Keypair {
...
 
 /// An Ed25519 public key.
-#[derive(PartialEq, Eq, Debug, Clone)]
+#[derive(PartialEq, Eq, Clone)]
 pub struct PublicKey(ed25519::PublicKey);
 
+impl fmt::Debug for PublicKey {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.write_str("PublicKey(compressed): ")?;
+        for byte in self.0.as_bytes() {
```
mxinden, line 111: Not sure whether this is relevant in the first place, but would it not be a bit more efficient to write the bytes slice to the formatter in one go, instead of one byte at a time?

mxinden (approved):
:+1: Change looks good to me.

core/src/identity/ed25519.rs
romanb: If you mean formatting the entire `&[u8]` slice with "{:x?}", that works, it is just more noisy as you get "[2a, 1b, ..., 9f]" rather than "2a1b9f", i.e. it is roughly twice as long, but I don't mind if you think it is preferable. Or did you have something else in mind?

mxinden: I somehow thought there is a formatter in `std` that would output `2a1b...9f` directly, which turns out to be a wrong assumption. Thanks for clarifying. I am in favor of merging this pull request as-is.

PR #1986, opened 2021-03-03
Don't leak default features

quininer:
If you directly rely on libp2p crate, you cannot turn off some default-features even if you use `default-features = false`.


romanb (approved):
Thanks!

mxinden (approved):

PR #1987, opened 2021-03-03
Close response stream even if response cannot be sent

thomaseizinger:
This was already fixed at some point (https://github.com/libp2p/rust-libp2p/pull/1660) but it seems that recent changes unfortunately undid this again.

I added a test that verifies that the connection is closed for the dialing party.

### DIFF
--- a/protocols/request-response/src/handler/protocol.rs
+++ b/protocols/request-response/src/handler/protocol.rs
@@ -106,6 +106,7 @@ where
                     let write = self.codec.write_response(&protocol, &mut io, response);
                     write.await?;
                 } else {
+                    io.close().await?;
                     return Ok(false)
                 }
             }
--- a/protocols/request-response/tests/ping.rs
+++ b/protocols/request-response/tests/ping.rs
@@ -47,8 +47,8 @@ fn is_response_outbound() {
     let cfg = RequestResponseConfig::default();
 
     let (peer1_id, trans) = mk_transport();
-    let ping_proto1 = RequestResponse::new(PingCodec(), protocols.clone(), cfg.clone());
-    let mut swarm1 = Swarm::new(trans, ping_proto1, peer1_id.clone());
+    let ping_proto1 = RequestResponse::new(PingCodec(), protocols, cfg);
+    let mut swarm1 = Swarm::new(trans, ping_proto1, peer1_id);
 
     let request_id1 = swarm1.send_request(&offline_peer, ping.clone());
 
@@ -60,7 +60,7 @@ fn is_response_outbound() {
         e => panic!("Peer: Unexpected event: {:?}", e),
     }
 
-    let request_id2 = swarm1.send_request(&offline_peer, ping.clone());
+    let request_id2 = swarm1.send_request(&offline_peer, ping);
 
     assert!(!swarm1.is_pending_outbound(&offline_peer, &request_id1));
     assert!(swarm1.is_pending_outbound(&offline_peer, &request_id2));
@@ -77,11 +77,11 @@ fn ping_protocol() {
 
     let (peer1_id, trans) = mk_transport();
     let ping_proto1 = RequestResponse::new(PingCodec(), protocols.clone(), cfg.clone());
-    let mut swarm1 = Swarm::new(trans, ping_proto1, peer1_id.clone());
+    let mut swarm1 = Swarm::new(trans, ping_proto1, peer1_id);
 
     let (peer2_id, trans) = mk_transport();
     let ping_proto2 = RequestResponse::new(PingCodec(), protocols, cfg);
-    let mut swarm2 = Swarm::new(trans, ping_proto2, peer2_id.clone());
+    let mut swarm2 = Swarm::new(trans, ping_proto2, peer2_id);
 
     let (mut tx, mut rx) = mpsc::channel::<Multiaddr>(1);
 
@@ -157,17 +157,17 @@ fn emits_inbound_connection_closed_failure() {
 
     let (peer1_id, trans) = mk_transport();
     let ping_proto1 = RequestResponse::new(PingCodec(), protocols.clone(), cfg.clone());
-    let mut swarm1 = Swarm::new(trans, ping_proto1, peer1_id.clone());
+    let mut swarm1 = Swarm::new(trans, ping_proto1, peer1_id);
 
     let (peer2_id, trans) = mk_transport();
     let ping_proto2 = RequestResponse::new(PingCodec(), protocols, cfg);
-    let mut swarm2 = Swarm::new(trans, ping_proto2, peer2_id.clone());
+    let mut swarm2 = Swarm::new(trans, ping_proto2, peer2_id);
 
     let addr = "/ip4/127.0.0.1/tcp/0".parse().unwrap();
     Swarm::listen_on(&mut swarm1, addr).unwrap();
 
     futures::executor::block_on(async move {
-        while let Some(_) = swarm1.next().now_or_never() {}
+        while swarm1.next().now_or_never().is_some() {}
         let addr1 = Swarm::listeners(&swarm1).next().unwrap();
 
         swarm2.add_address(&peer1_id, addr1.clone());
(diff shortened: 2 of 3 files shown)


protocols/request-response/tests/ping.rs
```diff
@@ -201,6 +201,62 @@ fn emits_inbound_connection_closed_failure() {
...
+                },
+            )
+        };
+
+        let error = match event {
+            RequestResponseEvent::OutboundFailure { error, .. } => error,
+            e => panic!("unexpected event from peer 2: {:?}", e)
+        };
+
+        assert_eq!(error, OutboundFailure::ConnectionClosed);
```
mxinden, line 256:
```suggestion
        match event {
            RequestResponseEvent::OutboundFailure { OutboundFailure::ConnectionClosed, .. } => {},
            e => panic!("unexpected event from peer 2: {:?}", e)
        }
```

Small nit pick: Would that not be easier to read?

protocols/request-response/src/handler/protocol.rs
```diff
@@ -106,6 +106,7 @@ where
                     let write = self.codec.write_response(&protocol, &mut io, response);
                     write.await?;
                 } else {
+                    io.close().await?;
                     return Ok(false)
                 }
             }
```
mxinden, line 112: Not directly related to this pull request: I wonder whether we should handle the case where `self.request_sender.send` fails. As far as I can tell, this can only fail in case our internal tracking in `RequestResponseHandler.inbound` looses track of the receiving side. I would suggest adding a `debug_assert` here to catch future regressions. What do people think? (Happy to move this to a follow-up pull request.)

```diff
diff --git a/protocols/request-response/src/handler/protocol.rs b/protocols/request-response/src/handler/protocol.rs
index bff57058..6cee3469 100644
--- a/protocols/request-response/src/handler/protocol.rs
+++ b/protocols/request-response/src/handler/protocol.rs
@@ -101,16 +101,23 @@ where
         async move {
             let read = self.codec.read_request(&protocol, &mut io);
             let request = read.await?;
-            if let Ok(()) = self.request_sender.send((self.request_id, request)) {
-                if let Ok(response) = self.response_receiver.await {
-                    let write = self.codec.write_response(&protocol, &mut io, response);
-                    write.await?;
-                } else {
-                    return Ok(false)
-                }
+            match self.request_sender.send((self.request_id, request)) {
+                Ok(()) => {},
+                Err(_) => debug_assert!(true, "Expect request receiver to be alive."),
+            }
+
+            if let Ok(response) = self.response_receiver.await {
+                let write = self.codec.write_response(&protocol, &mut io, response);
+                write.await?;
+
+                io.close().await?;
+                // Response was sent. Indicate to handler to emit a `ResponseSent` event.
+                Ok(true)
+            } else {
+                io.close().await?;
+                // No response was sent. Indicate to handler to emit a `ResponseOmission` event.
+                Ok(false)
             }
-            io.close().await?;
-            Ok(true)
         }.boxed()
     }
 }

```

mxinden (approved):
Pull request looks good to me, just left 4 comments which I am fine with being ignored.

Thanks @thomaseizinger for putting in the work, especially with the test.

In case this is not urgent I would suggest to give @romanb a chance to review as well as Roman is most familiar with this code path.

protocols/request-response/tests/ping.rs
```diff
@@ -201,6 +201,62 @@ fn emits_inbound_connection_closed_failure() {
     });
 }
 
+#[test]
+fn emits_inbound_connection_closed_if_channel_is_dropped() {
```
romanb, line 205:
```suggestion
/// We expect the substream to be properly closed when response channel is dropped.
/// Since the ping protocol used here expects a response, the sender considers this
/// early close as a protocol violation which results in the connection being closed.
/// If the substream were not properly closed when dropped, the sender would instead
/// run into a timeout waiting for the response.
#[test]
fn emits_inbound_connection_closed_if_channel_is_dropped() {
```
Just to clarify, since it can be misleading to think that it is the dropping of the response channel itself that results in the connection being closed. It is rather the protocol violation (received substream close instead of response) observed by the sender of the request.

protocols/request-response/src/handler/protocol.rs
romanb: My reasoning for silently ignoring this case is that it simply means the handler has been dropped and the connection closed, so there is nothing more to do. If we want to put an assertion there to check that the upgrade is never polled without the handler still being alive and holding the receiver, that sounds fine to me. But I think a separate PR would indeed be better, as you suggested yourself.

romanb (approved):
I only left a small suggestion beside those already given by @mxinden. Otherwise looks good to me, thanks for catching this (partial) regression for the case where no response is sent!

protocols/request-response/tests/ping.rs
thomaseizinger: I was optimizing for the error message. If I extract the `OutboundFailure`, the error message reads:

Expected: ConnectionClosed
Got: Timeout

Thoughts?

protocols/request-response/src/handler/protocol.rs
thomaseizinger: If the handler has already been dropped, whatever we return here will not be delivered anyway correct?

Could we also return an error? Because that would make the handling more ergonomic.

protocols/request-response/tests/ping.rs
```diff
@@ -201,6 +201,62 @@ fn emits_inbound_connection_closed_failure() {
...
+    let ping = Ping("ping".to_string().into_bytes());
+
+    let protocols = iter::once((PingProtocol(), ProtocolSupport::Full));
+    let cfg = RequestResponseConfig::default();
+
+    let (peer1_id, trans) = mk_transport();
+    let ping_proto1 = RequestResponse::new(PingCodec(), protocols.clone(), cfg.clone());
+    let mut swarm1 = Swarm::new(trans, ping_proto1, peer1_id.clone());
```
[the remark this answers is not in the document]
thomaseizinger: I ran clippy on `ping.rs` to catch even more of these!

protocols/request-response/src/handler/protocol.rs
thomaseizinger: I was actually bothered by the control flow of this function because it is not that straight-forward to follow. Your snippet gave me an idea. What do you think of 75450be?

thomaseizinger: I've opted for returning an `io::ErrorKind::ConnectionAborted`. As far as I understand, this should never show up anywhere because the handler has already been dropped and it would be the one to receive this error in `ProtocolsHandler::inject_dial_upgrade_error` but I found the error variant to be descriptive of when this happens.

mxinden: > Could we also return an error?

I would prefer the assertion above, given that it would surface the underlying issue (wrong assumption) compared to the error which would be silently ignored. I do agree that https://github.com/libp2p/rust-libp2p/commit/9f97d7d150e6bede6a812710b41e99f5fb9983b2 improves the control flow over the status-quo, but I would argue that the diff I suggested above is even simpler, given that there is no additional function required.

I would suggest reverting https://github.com/libp2p/rust-libp2p/pull/1987/commits/9f97d7d150e6bede6a812710b41e99f5fb9983b2 in this pull request and applying the diff above in a follow-up pull request.

@thomaseizinger would the above work for you? Sorry for the lengthy review.

thomaseizinger: > but I would argue that the diff I suggested above is even simpler, given that there is no additional function required.

The reason I introduced the function was because I wanted a clear separation between the control flow of handling the upgrade and closing the substream.

This bug has only been re-introduced because we early return from the code that handles the upgrade. However, I guess one could argue that the introduced test should now cover this.

I'll drop the last commit and leave it up to you if you want to do any refactoring :)

mxinden:
> Just to avoid misunderstandings: This looks ready to be merged now, right?

Yes. I will merge now and push a commit updating cargo tomls and changelogs. I will update https://github.com/libp2p/rust-libp2p/pull/1988 shortly after.

mxinden:
For the record, commit updating changelog: https://github.com/libp2p/rust-libp2p/commit/67c465c72644f95e88084224cf90a0a0bc24aa9d

Thanks @thomaseizinger! I will follow up with another pull request as discussed.

Issue #1577, opened 2020-05-15
protocols/kad: Why run caching procedure for queries with Quorum::One only?


mxinden:
When querying for a key, Kademlia remembers the closest node that did not have the key to later on store the record on that node.

[Paper](https://pdos.csail.mit.edu/~petar/papers/maymounkov-kademlia-lncs.pdf) page 7:

> For caching purposes, once a lookup suceeds, the requesting node stores the <key, value> pair at the closest node it observed to the key that did not return the value.

This is implemented here:

https://github.com/libp2p/rust-libp2p/blob/c271f6f56b7b5c33b1ced135e01291b1d48e8eb8/protocols/kad/src/behaviour.rs#L1306-L1320

**Why is this procedure only triggered for queries with `Quorum::One`? Why not do this for all queries?**

//CC @romanb

romanb:
I have some difficulties remembering but I thin it really came down to the semantics just being a bit unclear. In standard Kademlia the record is cached at the nearest peer to the key that did not return the value _on a successful lookup_. A successful lookup in standard Kademlia is with quorum 1. If this is done for quorums > 1 as well, then the record will be cached even if, from a user's perspective, the lookup wasn't actually successful (e.g. quorum not satisfied). That said, I don't currently see any harm in doing that either, so if there was a very strong reason, I must have forgotten about it by now.

romanb:
I think I remember now. The question is: With a quorum > 1, which of the returned records do you cache? A choice would be implied there, which we cannot really make.

romanb:
@mxinden Did my last comment answer your question and/or do you have specific suggestions for what we could do?

mxinden:
> I think I remember now. The question is: With a quorum > 1, which of the returned records do you cache? They may not be the same, and the reason someone uses a quorum > 1 to begin with is to make a choice among the obtained results, usually. That is a choice that we cannot make internally in libp2p.

Thanks for the follow up. I agree that this is problematic. I can think of two improvements (not mutually exclusive):
- Once a query with quorum > 1 is finished AND was successful AND all records are the same then cache the record at the closest node.
- Once a query with quorum > 1 is finished but is returned to the user with a diverging set of records, allow the user to choose one to be cached at the closest node via e.g. a function like `cache_record_at_closest_node`.

What do you think? Do you have alternative ideas how to make this integral part of Kademlia work for queries with quorum > 1?

romanb:
[...]

PR #1988, opened 2021-03-04
[kad] Provide a targeted store operation.

romanb:
This PR adds `Kademlia::put_record_to` for storing a record at specific nodes, e.g. for write-back caching after a successful read. In that context, peers that were queried in a successful `Kademlia::get_record` operation but did not return a record are now returned in the `GetRecordOk::no_record` list of peer IDs as additional information which may be used to choose which of these peers to send (one of) the found records to.

The motivation for this PR originates in the discussion at https://github.com/libp2p/rust-libp2p/issues/1577, i.e. the inability for libp2p-kad itself to do write-back caching of records after successful lookups when a quorum > 1 is used. It is deemed best and the most flexible to provide the means to do so on the API. There may be other use-cases for targeted store operations as well. That is to say, `Kademlia::put_record_to` is intended as to augment the use of `Kademlia::put_record` with occasional targeted store operations, not as a replacement. Though if someone wishes to only use `Kademlia::put_record_to`, that can obviously be done, the resulting DHT then just no longer has the properties attributed to Kademlia.

I'm not sure if it would be worthwhile to be able to configure (i.e. enable/disable) the default behaviour of write-back caching upon lookups with a quorum of 1. Maybe some users would like full control over that behaviour by disabling it and doing any such writes themselves via `Kademlia::put_record_to`. Feedback appreciated.

### DIFF
--- a/protocols/kad/CHANGELOG.md
+++ b/protocols/kad/CHANGELOG.md
@@ -1,10 +1,17 @@
 # 0.29.0 [unreleased]
 
+- Add `KademliaCaching` and `KademliaConfig::set_caching` to configure
+  whether Kademlia should track, in lookups, the closest nodes to a key
+  that did not return a record, via `GetRecordOk::cache_candidates`.
+  As before, if a lookup used a quorum of 1, these candidates will
+  automatically be sent the found record. Otherwise, with a lookup
+  quorum of > 1, the candidates can be used with `Kademlia::put_record_to`
+  after selecting one of the return records to cache. As is the current
+  behaviour, caching is enabled by default with a `max_peers` of 1, i.e.
+  it only tracks the closest node to the key that did not return a record.
+
 - Add `Kademlia::put_record_to` for storing a record at specific nodes,
-  e.g. for write-back caching after a successful read. In that context,
-  peers that were queried in a successful `Kademlia::get_record` operation but
-  did not return a record are now returned in the `GetRecordOk::cache_candidates`
-  list of peer IDs.
+  e.g. for write-back caching after a successful read with quorum > 1.
 
 - Update `libp2p-swarm`.
 
--- a/protocols/kad/CHANGELOG.md
+++ b/protocols/kad/CHANGELOG.md
@@ -1,5 +1,11 @@
 # 0.29.0 [unreleased]
 
+- Add `Kademlia::put_record_to` for storing a record at specific nodes,
+  e.g. for write-back caching after a successful read. In that context,
+  peers that were queried in a successful `Kademlia::get_record` operation but
+  did not return a record are now returned in the `GetRecordOk::no_record`
+  list of peer IDs.
+
 - Update `libp2p-swarm`.
 
 # 0.28.1 [2021-02-15]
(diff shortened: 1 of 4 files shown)


protocols/kad/src/behaviour.rs
```diff
@@ -1744,19 +1799,22 @@ where
...
-                            // closest node to the key that did *not* return the
-                            // value is tracked in order to cache the record on
-                            // that node if the query turns out to be successful.
-                            let source_key = kbucket::Key::from(source);
-                            if let Some(cache_key) = cache_at {
-                                let key = kbucket::Key::new(key.clone());
-                                if source_key.distance(&key) < cache_key.distance(&key) {
+                        } else {
+                            no_record.push(source);
+                            if quorum.get() == 1 {
```
mxinden, line 1804: > I'm not sure if it would be worthwhile to be able to configure (i.e. enable/disable) the default behaviour of write-back caching upon lookups with a quorum of 1. Maybe some users would like full control over that behaviour by disabling it and doing any such writes themselves via Kademlia::put_record_to. Feedback appreciated.

I am fine with either solution. That said I do tend towards not treating _quorum of 1_ as a special case. Instead I suggest to always require the user to call `put_record_to` in order for _write back caching_ to close nodes without the record to happen. Not having  a special case for _quorum of 1_ seems more consistent to me, despite the fact that it is not perfectly inline with the paper.

I would deem an additional note on the `GetRecordOk` structure, informing the user of their responsibility, as well as an entry in the changelog enough. We could as well enforce this at the type level, requiring the collection of peers returned in `GetRecordOk::no_record` to be used, just like `std::result::Result`.

What do you think @romanb?

protocols/kad/src/behaviour.rs
```diff
@@ -2389,6 +2466,8 @@ pub enum QueryInfo {
         records: Vec<PeerRecord>,
         /// The number of records to look for.
         quorum: NonZeroUsize,
+        /// The peers what were queried but did not return a record for the `key`.
```
mxinden, line 2469:
```suggestion
        /// The peers that were queried but did not return a record for the `key`.
```

protocols/kad/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # 0.29.0 [unreleased]
 
+- Add `Kademlia::put_record_to` for storing a record at specific nodes,
+  e.g. for write-back caching after a successful read. In that context,
+  peers that were queried in a successful `Kademlia::get_record` operation but
+  did not return a record are now returned in the `GetRecordOk::no_record`
+  list of peer IDs.
```
mxinden, line 7: Might be worth mentioning `KademliaCaching` here. What do you think @romanb?

protocols/kad/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # 0.29.0 [unreleased]
 
+- Add `Kademlia::put_record_to` for storing a record at specific nodes,
+  e.g. for write-back caching after a successful read. In that context,
+  peers that were queried in a successful `Kademlia::get_record` operation but
+  did not return a record are now returned in the `GetRecordOk::no_record`
```
mxinden, line 6:
```suggestion
  did not return a record are now returned in the `GetRecordOk::cache_candidates`
```

protocols/kad/CHANGELOG.md
romanb: :+1: I forgot about the changelog. Updated in https://github.com/libp2p/rust-libp2p/pull/1988/commits/5bada7afbff87335f6bb3991126089ee5502a7e7.

PR #1989, opened 2021-03-07
protocols/mdns: IPv6 implementation closes #873

### LINKED ISSUE #873: mDNS does not support IPv6
## Summary

The MdnsService only joins and sends queries to IPv4 mDNS multicast address. It should also participate via IPv6. Per [RFC 6762]

> A dual-stack (v4/v6) host can participate in both ".local." zones, and should register its name(s) and perform its lookups both using IPv4 and IPv6.  This enables it to reach, and be reached by, both IPv4-only and IPv6-only hosts.  In effect, this acts like a multihomed host, with one connection to the logical "IPv4 Ethernet segment", and a connection to the logical "IPv6 Ethernet segment".

## Possible solution

I think a reasonable way to do this would be to keep the single `query_send_buffer` responsible for advertising oneself, but split out the `MdnsService::send_buffer` and `MdnsService::socket` into separate `v4` and `v6` instances. Depending on whether the sender is `v6` or `v4` we can give the `MdnsPacket` the correct `send_buffer`.

## Open Questions

 - Does this need to be configurable? Perhaps the `MdnsService` constructors should take a parameter that conditionally enables IPv6 and IPv4 support.
 - [RFC 6762] states that a host responding via IPv6 and IPv4 should result in a single host with two records (`A` and `AAAA`). Is there anything in the `mdns` behaviour that needs to be changed to ensure we don't end up with two separate hosts?

[RFC 6762]: https://tools.ietf.org/html/rfc6762

majchrzamemil:
Hi,
I got interest in libp2p and Rust, this PR presents implementation of IPv6 for mDNS protocol. I'd be glad to learn something and contribute to this project
Closes #873
157995850_1006568313203807_3724658030105501916_n
158730277_860597381166273_939988636688228741_n
Above I attached print screens form mdns example running in IPv6 mode and from wireshark sniffing packet on one of interfaces.

### DIFF
--- a/protocols/mdns/src/behaviour.rs
+++ b/protocols/mdns/src/behaviour.rs
@@ -45,10 +45,9 @@ use std::{
 };
 
 lazy_static! {
-    static ref IPV4_MDNS_MULTICAST_ADDRESS: SocketAddr =
-        SocketAddr::from((Ipv4Addr::new(224, 0, 0, 251), 5353));
-    static ref IPV6_MDNS_MULTICAST_ADDRESS: SocketAddr =
-        SocketAddr::from((Ipv6Addr::new(0xFF02, 0, 0, 0, 0, 0, 0, 0xFB), 5353));
+    static ref IPV4_MDNS_MULTICAST_ADDRESS: IpAddr = IpAddr::V4(Ipv4Addr::new(224, 0, 0, 251));
+    static ref IPV6_MDNS_MULTICAST_ADDRESS: IpAddr =
+        IpAddr::V6(Ipv6Addr::new(0xFF02, 0, 0, 0, 0, 0, 0, 0xFB));
 }
 
 pub struct MdnsConfig {
@@ -60,8 +59,8 @@ pub struct MdnsConfig {
     /// peer joins the network. Receiving an mdns packet resets the timer
     /// preventing unnecessary traffic.
     pub query_interval: Duration,
-    /// IP version type.
-    pub domain: IPType,
+    /// IP address for multicast.
+    pub ip_addr: IpAddr,
 }
 
 impl Default for MdnsConfig {
@@ -69,17 +68,11 @@ impl Default for MdnsConfig {
         Self {
             ttl: Duration::from_secs(6 * 60),
             query_interval: Duration::from_secs(5 * 60),
-            domain: IPType::IPv4,
+            ip_addr: *IPV4_MDNS_MULTICAST_ADDRESS,
         }
     }
 }
 
-/// Enum for IP type.
-#[derive(Debug)]
-pub enum IPType {
-    IPv4,
-    IPv6,
-}
 /// A `NetworkBehaviour` for mDNS. Automatically discovers peers on the local network and adds
 /// them to the topology.
 #[derive(Debug)]
@@ -127,15 +120,18 @@ pub struct Mdns {
     /// Discovery timer.
     timeout: Timer,
 
-    /// IP Type.
-    domain: IPType,
+    // Mutlicast address.
+    multicast_addr: IpAddr,
+
+    // Socket address.
+    socket_addr: SocketAddr,
 }
 
 impl Mdns {
     /// Builds a new `Mdns` behaviour.
     pub async fn new(config: MdnsConfig) -> io::Result<Self> {
-        let recv_socket = match config.domain {
-            IPType::IPv4 => {
+        let recv_socket = match config.ip_addr {
+            IpAddr::V4(_) => {
                 let socket = Socket::new(
                     Domain::ipv4(),
                     Type::dgram(),
@@ -150,7 +146,7 @@ impl Mdns {
                 socket.set_multicast_ttl_v4(255)?;
                 Async::new(socket)?
             }
-            IPType::IPv6 => {
+            IpAddr::V6(_) => {
                 let socket = Socket::new(
                     Domain::ipv6(),
                     Type::dgram(),
@@ -165,17 +161,17 @@ impl Mdns {
                 Async::new(socket)?
             }
         };
-        let send_socket = match config.domain {
-            IPType::IPv4 => {
-                let socket = std::net::UdpSocket::bind((Ipv4Addr::UNSPECIFIED, 0))?;
-                Async::new(socket)?
-            }
-            IPType::IPv6 => {
-                let socket = std::net::UdpSocket::bind((Ipv6Addr::UNSPECIFIED, 0))?;
-                Async::new(socket)?
-            }
+        let send_socket = {
+            let addrs = [
+                SocketAddr::new(IpAddr::V4(Ipv4Addr::UNSPECIFIED), 0),
+                SocketAddr::new(IpAddr::V6(Ipv6Addr::UNSPECIFIED), 0),
+            ];
+
+            let socket = std::net::UdpSocket::bind(&addrs[..])?;
+            Async::new(socket)?
         };
         let if_watch = if_watch::IfWatcher::new().await?;
+        let socket_addr = SocketAddr::new(config.ip_addr, 5353);
         Ok(Self {
             recv_socket,
             send_socket,
@@ -188,7 +184,8 @@ impl Mdns {
             query_interval: config.query_interval,
             ttl: config.ttl,
             timeout: Timer::interval(config.query_interval),
-            domain: config.domain,
+            multicast_addr: config.ip_addr,
+            socket_addr: socket_addr,
         })
     }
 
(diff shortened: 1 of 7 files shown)


protocols/mdns/src/behaviour.rs
```diff
@@ -279,29 +316,57 @@ impl NetworkBehaviour for Mdns {
...
+                                    log::error!("join multicast failed: {}", err);
+                                } else {
+                                    self.send_buffer.push_back(build_query());
+                                }
+                            }
+                        }
+                        IPType::IPv6 => {
+                            if let IpAddr::V6(addr) = inet.addr() {
+                                log::trace!("joining multicast on iface {}", addr);
+                                //TODO: determine interface ID now it generates errors.
```
majchrzamemil, line 340: One TODO left from my side to cleanup joining to multicast interface for IPv6

protocols/mdns/src/behaviour.rs
```diff
@@ -58,17 +60,26 @@ pub struct MdnsConfig {
...
+    /// IP version type.
+    pub domain: IPType,
 }
 
 impl Default for MdnsConfig {
     fn default() -> Self {
         Self {
             ttl: Duration::from_secs(6 * 60),
             query_interval: Duration::from_secs(5 * 60),
+            domain: IPType::IPv4,
```
dvc94ch, line 72: I don't think we need this config, we can just work with ip6 in addition to ip4

protocols/mdns/src/behaviour.rs
```diff
@@ -115,29 +126,54 @@ pub struct Mdns {
...
+                    Some(socket2::Protocol::udp()),
+                )?;
+                socket.set_reuse_address(true)?;
+                #[cfg(unix)]
+                socket.set_reuse_port(true)?;
+                socket.bind(&SocketAddr::new(Ipv4Addr::UNSPECIFIED.into(), 5353).into())?;
+                let socket = socket.into_udp_socket();
+                socket.set_multicast_loop_v4(true)?;
+                socket.set_multicast_ttl_v4(255)?;
+                Async::new(socket)?
```
dvc94ch, line 151: I think you can remove the v4 socket. Unless the `IPV6_V6ONLY` socket option is enabled a v6 socket should support both ip protocols.

protocols/mdns/src/behaviour.rs
```diff
@@ -115,29 +126,54 @@ pub struct Mdns {
...
-            let socket = UdpSocket::bind((Ipv4Addr::UNSPECIFIED, 0))?;
-            Async::new(socket)?
+        let send_socket = match config.domain {
+            IPType::IPv4 => {
+                let socket = std::net::UdpSocket::bind((Ipv4Addr::UNSPECIFIED, 0))?;
+                Async::new(socket)?
+            }
+            IPType::IPv6 => {
+                let socket = std::net::UdpSocket::bind((Ipv6Addr::UNSPECIFIED, 0))?;
+                Async::new(socket)?
```
dvc94ch, line 175: same here, binding to `Ipv6Addr::UNSPECIFIED` should allow sending ipv4 packets. please test that this is indeed the case

protocols/mdns/src/behaviour.rs
```diff
@@ -279,29 +316,57 @@ impl NetworkBehaviour for Mdns {
...
                     }
-                    if let IpAddr::V4(addr) = inet.addr() {
-                        log::trace!("joining multicast on iface {}", addr);
-                        if let Err(err) = socket.join_multicast_v4(&multicast, &addr) {
-                            log::error!("join multicast failed: {}", err);
-                        } else {
-                            self.send_buffer.push_back(build_query());
+                    match self.domain {
+                        IPType::IPv4 => {
+                            if let IpAddr::V4(addr) = inet.addr() {
```
dvc94ch, line 328: match on address type instead of self.domain

protocols/mdns/src/behaviour.rs
```diff
@@ -279,29 +316,57 @@ impl NetworkBehaviour for Mdns {
...
                 }
                 Ok(IfEvent::Down(inet)) => {
                     if inet.addr().is_loopback() {
                         continue;
                     }
-                    if let IpAddr::V4(addr) = inet.addr() {
-                        log::trace!("leaving multicast on iface {}", addr);
-                        if let Err(err) = socket.leave_multicast_v4(&multicast, &addr) {
-                            log::error!("leave multicast failed: {}", err);
+                    match self.domain {
```
dvc94ch, line 354: same here

protocols/mdns/src/behaviour.rs
```diff
@@ -332,14 +397,29 @@ impl NetworkBehaviour for Mdns {
...
                 if let Some(packet) = self.send_buffer.pop_front() {
-                    match self
-                        .send_socket
-                        .send_to(&packet, *IPV4_MDNS_MULTICAST_ADDRESS)
-                        .now_or_never()
-                    {
-                        Some(Ok(_)) => {}
-                        Some(Err(err)) => log::error!("{}", err),
-                        None => self.send_buffer.push_front(packet),
+                    match self.domain {
```
dvc94ch, line 400: not sure about this, it seems wasteful to do mdns on ip4 and ip6 at the same time, but I guess we could make the `MULTICAST_ADDRESS` configurable and default to `IPV4_MDNS_MULTICAST_ADDRESS`.

protocols/mdns/src/behaviour.rs
dvc94ch: You can add a multicast_addr: std::net::IpAddr instead and just combine that with a port to get a socketaddr

protocols/mdns/src/behaviour.rs
```diff
@@ -115,31 +119,59 @@ pub struct Mdns {
 
     /// Discovery timer.
     timeout: Timer,
+
+    // Mutlicast address.
+    multicast_addr: IpAddr,
+
+    // Socket address.
+    socket_addr: SocketAddr,
```
dvc94ch, line 127: do we need the socket_addr for anything?

protocols/mdns/src/behaviour.rs
```diff
@@ -115,31 +119,59 @@ pub struct Mdns {
 
     /// Discovery timer.
     timeout: Timer,
+
+    // Mutlicast address.
+    multicast_addr: IpAddr,
```
dvc94ch, line 124: make it a doc string `///`

protocols/mdns/src/behaviour.rs
dvc94ch: so this is actually needed?

protocols/mdns/src/behaviour.rs
majchrzamemil: It works on Linux, macOS has problem with joining multicast on IPv6, while switching to IPv6Addr on macOS I got os error 22. Should this be compatible with macOS?
If not I will go with your solution Ippv6Addr::UNSPECIFIED, but if it should i propose:
let send_socket = {
            let addrs = [
                SocketAddr::new(IpAddr::V4(Ipv4Addr::UNSPECIFIED), 0),
                SocketAddr::new(IpAddr::V6(Ipv6Addr::UNSPECIFIED), 0),
            ];

            let socket = std::net::UdpSocket::bind(&addrs[..])?;
            Async::new(socket)?
        };

protocols/mdns/src/behaviour.rs
majchrzamemil: done

protocols/mdns/src/behaviour.rs
majchrzamemil: done

protocols/mdns/src/behaviour.rs
majchrzamemil: done

protocols/mdns/src/behaviour.rs
majchrzamemil: done

protocols/mdns/src/behaviour.rs
```diff
@@ -58,13 +59,16 @@ pub struct MdnsConfig {
     /// peer joins the network. Receiving an mdns packet resets the timer
     /// preventing unnecessary traffic.
     pub query_interval: Duration,
+    /// IP address for multicast.
+    pub ip_addr: IpAddr,
```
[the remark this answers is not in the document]
majchrzamemil: done

protocols/mdns/src/behaviour.rs
majchrzamemil: done

protocols/mdns/src/behaviour.rs
majchrzamemil: it is used in line 397, might be created there but wanted avoid not needed object creation

majchrzamemil: and yes this is not needed in case of Linux, but in that case mdns will discover also IPv6 addresses

protocols/mdns/src/behaviour.rs
majchrzamemil: it is used in line 397, might be created there but wanted avoid not needed object creation

mxinden:
> Should Mdns work in IPv4 and IPv6 mode simultaneously?

For what my input is worth, the [golang implementation](https://github.com/whyrusleeping/mdns/blob/master/client.go) seems to operate on both IPv4 and IPv6 simultaneously. I am not sure what the default behaviour should be, but supporting the simultaneous behaviour seems to me like a good option to have. What do you think?

majchrzamemil:
If you consider this useful I can make implementation compliant with golang.

dvc94ch:
Depends if you only announce ip4 addresses in mdns4 and ip6 addresses in mdns6 it makes sense. Otherwise I don't think it's a good idea

mxinden:
> Depends if you only announce ip4 addresses in mdns4 and ip6 addresses in mdns6 it makes sense. Otherwise I don't think it's a good idea

Could you extend on this @dvc94ch? In case an ipv4 only node would receive ipv6 addresses of some remote peer, it would simply fail dialing them later on, no?

dvc94ch:
well it's multicast. so you are doubling the traffic (request sent to n peers and n responses  to n peers is n + n2 packets. if you do the same thing on ip6 it's 2(n + n2) packets). if you are only listening on ip4 addresses, then it doesn't hurt as much as no one will respond to ipv6 mdns requests.

mxinden:
Closing here since #2161 is merged. Thanks @majchrzamemil and @dvc94ch.

PR #1990, opened 2021-03-10
protocols/req-resp: Panic in inbound upgrade when handler dropped

mxinden:
Continuation of https://github.com/libp2p/rust-libp2p/pull/1987#discussion_r586205722.

An inbound upgrade should not be polled after the corresponding handler
has been dropped. Enforce this assumption by panicing in such case.

I am fine with this suggestion being rejected.

//CC @thomaseizinger @romanb

### DIFF
--- a/protocols/request-response/src/handler/protocol.rs
+++ b/protocols/request-response/src/handler/protocol.rs
@@ -101,17 +101,25 @@ where
         async move {
             let read = self.codec.read_request(&protocol, &mut io);
             let request = read.await?;
-            if let Ok(()) = self.request_sender.send((self.request_id, request)) {
-                if let Ok(response) = self.response_receiver.await {
-                    let write = self.codec.write_response(&protocol, &mut io, response);
-                    write.await?;
-                } else {
-                    io.close().await?;
-                    return Ok(false)
-                }
+            match self.request_sender.send((self.request_id, request)) {
+                Ok(()) => {},
+                Err(_) => panic!(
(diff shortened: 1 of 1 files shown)


romanb (approved):

protocols/request-response/src/handler/protocol.rs
```diff
@@ -101,17 +101,25 @@ where
...
+                write.await?;
+
+                io.close().await?;
+                // Response was sent. Indicate to handler to emit a `ResponseSent` event.
+                Ok(true)
+            } else {
+                io.close().await?;
+                // No response was sent. Indicate to handler to emit a `ResponseOmission` event.
+                Ok(false)
             }
```
thomaseizinger, line 122: Previously, the bug was introduced because we called `io.close()` in several places. Maybe we can prevent another regression by factoring the closing out?

```suggestion
            let response_sent = if let Ok(response) = self.response_receiver.await {
                let write = self.codec.write_response(&protocol, &mut io, response);
                write.await?;

                true // Response was sent. Indicate to handler to emit a `ResponseSent` event.
            } else {

				false // No response was sent. Indicate to handler to emit a `ResponseOmission` event.
            };

            io.close().await?;
            Ok(response_sent)
```

thomaseizinger: Maybe it is actually better to call it in two places because the regression was introduced through an early return which can happen more likely if we try to close at the end.

PR #1995, opened 2021-03-13
swarm/src/lib: Remove Deref and DerefMut impls on Swarm

mxinden:
Remove `Deref` and `DerefMut` implementations previously dereferencing
to the `NetworkBehaviour` on `Swarm`. Instead one can access the
`NetworkBehaviour` via `Swarm::behaviour` and `Swarm::behaviour_mut`.
Methods on `Swarm` can now be accessed directly, e.g. via
`my_swarm.local_peer_id()`.

Reasoning: Accessing the `NetworkBehaviour` of a `Swarm` through `Deref`
and `DerefMut` instead of a method call is an unnecessary complication,
especially for newcomers. In addition, `Swarm` is not a smart-pointer
and should thus not make use of `Deref` and `DerefMut`, see documentation
from the standard library below.

> Deref should only be implemented for smart pointers to avoid
confusion.

https://doc.rust-lang.org/std/ops/trait.Deref.html

---

For the reasons mentioned above I would consider this breaking change worth doing. This has been shortly discussed in https://github.com/libp2p/rust-libp2p/pull/1073 in the past. What do people think?


thomaseizinger:
FWIW, I am in favor of this change. I always found `Swarm::dial_addr` hard to use because I know that it was there and it always takes me a while to remember that I have to call it as an associated funcion.

romanb (approved):
No objections from my side.

mxinden:
I would deem 5 days enough as a review period, thus merging now. Thanks for the reviews. Happy for future comments / suggestions.

dvc94ch:
Sorry I didn't see this sooner and comment. But I wonder if #2006 isn't the better approach. Keep the deref/deref_mut and instead removing the swarm methods/events in favor of more `inject_*event*` on `NetworkBehaviour` and more `NetworkBehaviourAction`s. One wart of libp2p is having multiple ways of doing things. Users can register some channels or something if they really need to handle events outside of the swarm.

thomaseizinger:
@dvc94ch In my understanding, that is a orthogonal issue.

With this patch, we simply force users to explicitely type `swarm.behaviour.foobar` instead of just `swarm.foobar`. That is slightly more verbose but also more idiomatic Rust because `Deref` is only meant for smart-pointers.

As a result, we were able to remove the unintuitve associated functions for calling others things on `Swarm` like `local_peer_id` etc

Improving the NetworkBehaviour and all the things we talked about in the other thread are still possible as far as I can see :)

dvc94ch:
I think there are other cases where using `Deref` is valid, for example when creating a new type wrapper to implement a foreign trait on a foreign type.

But yes, I think it's orthogonal, I was just worried about too many api changes for users.

MOZGIII:
> What do people think?

Thx so much for doing this!

PR #1999, opened 2021-03-15
[identify] Implement /ipfs/id/push/1.0.0 alongside some refactoring.

### LINKED ISSUE #1136: Implement identify/push
https://github.com/libp2p/specs/tree/master/identify#identifypush

romanb:
* Implement the `/ipfs/id/push/1.0.0` protocol, i.e. the ability to actively push information of the local peer to specific remotes.
  * Make the initial delay as well as the recurring delay for the periodic identification requests configurable, introducing `IdentifyConfig`.
  * Small cleanups and refactorings along the way.

Closes #1136.


mxinden:
Sorry for the delay here @romanb. I will take a look later today.

protocols/identify/src/identify.rs
```diff
@@ -185,31 +269,57 @@ impl NetworkBehaviour for Identify {
...
+                    .clone();
+
+                let listen_addrs = listen_addrs(params);
+                let protocols = supported_protocols(params);
+
+                let info = IdentifyInfo {
+                    public_key: self.config.local_public_key.clone(),
+                    protocol_version: self.config.protocol_version.clone(),
+                    agent_version: self.config.agent_version.clone(),
+                    listen_addrs: listen_addrs.clone(),
```
mxinden, line 288:
```suggestion
                    listen_addrs,
```
Clone not needed according to clippy. Seems right given `fn listen_addrs`, haven't confirmed locally.

protocols/identify/src/identify.rs
```diff
@@ -185,31 +269,57 @@ impl NetworkBehaviour for Identify {
...
+                    protocols: protocols.clone(),
```
mxinden, line 289:
```suggestion
                    protocols,
```
Same here.

mxinden (approved):
Changes look good to me. Thanks for the clean patch.

I wonder whether `libp2p-identify` should (by default?) automatically initiate pushes to its peers on `inject_new_listen_addr`, `inject_expired_listen_addr` and `inject_new_external_addr`.

If I understand correctly, this would be in line with the specification.

> When a peer's basic information changes, for example, because they've obtained a new public listen address, they can use identify/push to inform others about the new information.

romanb:
> I wonder whether libp2p-identify should (by default?) automatically initiate pushes to its peers on inject_new_listen_addr, inject_expired_listen_addr and inject_new_external_addr.

I can give that a try in a follow-up PR. It didn't occur to me to do that here. Thanks for the suggestion.

mxinden:
> > I wonder whether libp2p-identify should (by default?) automatically initiate pushes to its peers on inject_new_listen_addr, inject_expired_listen_addr and inject_new_external_addr.
> 
> I can give that a try in a follow-up PR. It didn't occur to me to do that here. Thanks for the suggestion.

Follow up pull request sounds good. Thanks for giving it a shot! Feel free to merge this pull request.

protocols/identify/src/handler.rs
```diff
@@ -34,89 +47,119 @@ use libp2p_swarm::{
...
-        self.events.push(IdentifyHandlerEvent::Identified(protocol));
-        self.keep_alive = KeepAlive::No;
+        match output {
+            EitherOutput::First(remote_info) => {
+                self.events.push(
+                    ProtocolsHandlerEvent::Custom(
+                        IdentifyHandlerEvent::Identified(remote_info)));
+                self.keep_alive = KeepAlive::No;
+            }
+            EitherOutput::Second(()) => {}
```
mxinden, line 152: Something that came to my mind: How about emitting an event along the lines of `IdentifyEvent::Pushed` once a push succeeded? I would expect this to be useful for people tracking these events in a monitoring system like Prometheus. This event would serve a similar purpose to the `RequestSent` event in `libp2p-request-response`:

https://github.com/libp2p/rust-libp2p/blob/63512e5f160c058471e59e9e7c48a40cb503c1d0/protocols/request-response/src/lib.rs#L152-L161

In case you think this is useful @romanb I can prepare a pull request.

romanb: Sounds good to me.

Issue #2006, opened 2021-03-20
Review of NetworkBehaviour


dvc94ch:
I actually kind of like the `NetworkBehaviour` in general. Some things that may help would be:

`inject_new_external_addr` needs a `inject_expired_external_addr` when `Swarm::remove_external_address` is called.

`inject_new_listen_addr` and `inject_expired_listen_addr` should pass a `ListenerId`.

`SwarmEvent` and `NetworkBehaviour` inject methods seem to be two ways of doing the same thing (in many cases). Maybe deprecate `SwarmEvent` and adding some missing methods to `NetworkBehaviour` would be sensible?

For writing tests (where I think `SwarmEvent` is probably most useful), a `NetworkBehaviour` can be added that allows registering a channel to get something like a `SwarmEvent`.

Another idea would be remove swarm methods completely and make `NetworkBehaviourAction` more powerful instead.

romanb:
> SwarmEvent and NetworkBehaviour inject methods seem to be two ways of doing the same thing (in many cases). Maybe deprecate SwarmEvent and adding some missing methods to NetworkBehaviour would be sensible?

The `SwarmEvent` is actually the "newer" of these, at least in its current form, see e.g. https://github.com/libp2p/rust-libp2p/pull/1515. I agree that the current situation, where there are `NetworkEvent`s, `SwarmEvent`s (an almost 1-1 mirror of `NetworkEvent` by now) as well as all the callbacks on a `NetworkBehaviour`, makes for quite redundant APIs. Personally, I would actually prefer going in the direction of removing `libp2p-swarm` (and `libp2p-swarm-derive`) and providing a much more minimal interface and API for composing protocols by means of composing event handlers that directly consume `NetworkEvent`s as part of `libp2p-core`. A kind of more radical variant of https://github.com/libp2p/rust-libp2p/issues/1522. The main problems I see with the granular callback-based APIs of the `NetworkBehaviour` are:

  1. It is often unclear, especially when getting started, in which order the callbacks are invoked and what guarantees are given in terms of when / if they are invoked and what the pre- and post-conditions of each callback method are. But precisely understanding the semantics of each callback is paramount to implementing a `NetworkBehaviour` that works correctly and e.g. does not leak memory in local data structures.
[...]

PR #2007, opened 2021-03-20
mdns: MdnsConfig derives Clone and Debug.

dvc94ch:
Towards #2006 unblocks a new ipfs-embed release.

### DIFF
--- a/protocols/mdns/src/behaviour.rs
+++ b/protocols/mdns/src/behaviour.rs
@@ -49,6 +49,8 @@ lazy_static! {
         SocketAddr::from((Ipv4Addr::new(224, 0, 0, 251), 5353));
 }
 
+/// Configuration for mDNS.
+#[derive(Clone, Debug)]
 pub struct MdnsConfig {
     /// TTL to use for mdns records.
     pub ttl: Duration,


swarm/src/lib.rs
```diff
@@ -412,6 +414,7 @@ where TBehaviour: NetworkBehaviour<ProtocolsHandler = THandler>,
     /// [`NetworkBehaviourAction::ReportObservedAddr`] or explicitly
     /// through this method.
     pub fn add_external_address(&mut self, a: Multiaddr, s: AddressScore) -> AddAddressResult {
+        self.behaviour.inject_new_external_addr(&a);
         self.external_addrs.add(a, s)
```
romanb, line 418:
```suggestion
        let result = self.external_addrs.add(a, s);
        if let AddAddressResult::Inserted = &result {
            self.behaviour.inject_new_external_addr(&a);
        }
        result
```

swarm/src/lib.rs
```diff
@@ -422,6 +425,7 @@ where TBehaviour: NetworkBehaviour<ProtocolsHandler = THandler>,
     /// Returns `true` if the address existed and was removed, `false`
     /// otherwise.
     pub fn remove_external_address(&mut self, addr: &Multiaddr) -> bool {
+        self.behaviour.inject_expired_external_addr(addr);
         self.external_addrs.remove(addr)
```
romanb, line 429:
```suggestion
        if self.external_addrs.remove(addr) {
            self.behaviour.inject_expired_external_addr(addr);
        }
```
However, note also that the most common way in which external addresses expire is as a result of new addresses being added. See https://github.com/libp2p/rust-libp2p/blob/master/swarm/src/registry.rs#L186-L196. So, to be able to properly invoke this callback whenever an external address expires, I think you would need to incorporate the expired/removed addresses within a call to `add()` in the `AddAddressResult`, so that the `Swarm` can then invoke the callback.

romanb:
The additions generally look good to me. We just need to make sure that the new callbacks for external addresses are called only (and always) when appropriate.

swarm/src/behaviour.rs
```diff
@@ -147,17 +147,17 @@ pub trait NetworkBehaviour: Send + 'static {
...
+    /// Indicates to the behaviour that a new listener was created.
+    fn inject_new_listener(&mut self, _id: ListenerId) {
+    }
+
     /// Indicates to the behaviour that we have started listening on a new multiaddr.
-    fn inject_new_listen_addr(&mut self, _addr: &Multiaddr) {
+    fn inject_new_listen_addr(&mut self, _id: ListenerId, _addr: &Multiaddr) {
     }
 
     /// Indicates to the behaviour that a new multiaddr we were listening on has expired,
```
mxinden, line 158:
```suggestion
    /// Indicates to the behaviour that a multiaddr we were listening on has expired,
```

protocols/mdns/src/behaviour.rs
```diff
@@ -49,6 +49,8 @@ lazy_static! {
         SocketAddr::from((Ipv4Addr::new(224, 0, 0, 251), 5353));
 }
 
+/// Configuration for mDNS.
+#[derive(Clone, Debug)]
```
mxinden, line 53: Could you extract this change to a separate pull request? Making only a single cohesive change per pull request makes maintaining this project a lot easier, especially around version and changelog management.

dvc94ch: I see you renamed the PR, I moved the api changes, as this one has the branch name `mdns-config`

Issue #1668, opened 2020-07-18
use rustfmt


dvc94ch:
Why use rustfmt? The main reason is so you don't annoy developers when they accidentally run it before committing their changes.

tcharding:
Yes please, it is a real drag to have to turn off format on save in your editor everytime one wants to work on rust-libp2p when ones daily work is on a rust code base that uses rustfmt. It also means I can't have a buffer open doing rust-libp2p work and a buffer open to other repos without have to reset all the formatting changes, while manageable its a drag to have to think about this stuff.

Can we configure rustfmt in a manner that is acceptable to the rust-libp2p project?

burdges:
Rust was designed to be expressive, which makes Rust far harder to fmt than languages explicitly designed to be inexpressive, and so rustfmt ends up being fairly crappy.

We could set rustfmt.toml  to `ignore = ["/"]` according to https://github.com/rust-lang/rustfmt/blob/master/Configurations.md except that never actually worked:  https://github.com/rust-lang/rustfmt/pull/4147#issuecomment-660793906 

As luck would have it, we could exploit rustfmt's bugginess for this by simply doing https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=856d116c793b74e6809441b4ccf7af68 in lib.rs or somewhere so that rustfmt panics https://github.com/rust-lang/rustfmt/issues/3678

rklaehn:
@burdges out of curiosity: could you point out examples where the rustfmt formatted code is so bad? I guess the huge lists of type constraints on some parts of the code will look bad. Maybe these sections can be manually excluded. In general I think fmt does an OK job.

burdges:
I only know my own code, not the libp2p crate, but..

It screws up any whitespace that communicates emphasis, especially with iterator chains like https://github.com/w3f/schnorrkel/blob/master/src/vrf.rs#L944 so you need far more temporaries, which becomes excessive.  There are many situations even post-NLL where Rust requires closures be handled in line, not as temporaries. 

Anyways Parity has their own style guide and afaik they ban rustfmt from their code bases.  I hope this becomes more commonplace..

dvc94ch:
Actually there are parity codebases that use rustfmt. Gav doesn't like it so it's not used in substrate or polkadot, but many smaller crates/repos it's up to the devs basically.

dvc94ch:
[...]

PR #2009, opened 2021-03-22
core/identity: Encode and decode for Ed25519 SecretKey.

xoreo:
It is odd how the ed25519 PublicKeys come with `.encode()` and `.decode()` methods, but the PrivateKeys do not. So, this PR adds similar functionality to the ed25519 SecretKey struct. The encode method returns a `&[u8; 32]` to view the SecretKey as bytes, and the `decode` method converts the output of `encode()` back into an ed25519 SecretKey. Also, wrote a passing test for this functionality.

### DIFF
--- a/core/src/identity/ed25519.rs
+++ b/core/src/identity/ed25519.rs
@@ -230,12 +230,7 @@ mod tests {
         fn prop() -> bool {
             let sk = SecretKey::generate();
             let sk_after = SecretKey::decode(sk.encode()).unwrap();
-            for i in 0..32 {
-                if sk.0.to_bytes()[i] != sk_after.0.to_bytes()[i] {
-                    return false;
-                }
-            }
-            true
+            sk.0.to_bytes() == sk_after.0.to_bytes()
         }
         QuickCheck::new().tests(10).quickcheck(prop as fn() -> _);
     }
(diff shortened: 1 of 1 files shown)


xoreo:
I also ran rustfmt, I hope that is not an issue.

core/src/identity/ed25519.rs
```diff
@@ -178,6 +191,18 @@ impl SecretKey {
...
         Ok(SecretKey(secret))
     }
+
+    /// View the secret key as bytes.
+    pub fn encode(&self) -> &[u8; 32] {
+        self.0.as_bytes()
+    }
+
+    /// Decode a private key from a byte array as produced by `encode`.
+    pub fn decode(k: &[u8]) -> Result<SecretKey, DecodingError> {
```
mxinden, line 201: How does `decode` differ from `from_bytes` other than not zeroizing the input? Why duplicate the functionality?

mxinden (commented):
:wave: hello @xoreo

I am a bit confused. Why not use the methods on `ed25519::KeyPair`, namely `encode` and `decode`?

> I also ran rustfmt, I hope that is not an issue.

While I don't hold a strong opinion personally, this has been discussed before, e.g. see https://github.com/libp2p/rust-libp2p/issues/1668, thus please restrict your pull request to the change itself.

core/src/identity/ed25519.rs
xoreo: Pardon my confusion. I now see that this PR is unnecessary.

PR #2011, opened 2021-03-22
swarm: Extend NetworkBehaviour callbacks.

dvc94ch:


core/src/network.rs
```diff
@@ -187,7 +187,7 @@ where
     ///
     /// The translation is transport-specific. See [`Transport::address_translation`].
     pub fn address_translation<'a>(&'a self, observed_addr: &'a Multiaddr)
-        -> impl Iterator<Item = Multiaddr> + 'a
+        -> Vec<Multiaddr>
```
romanb, line 190: Is this a necessary change for this PR?  I'm not opposed to returning a `Vec`, but if it is not necessary for this PR I think it would be better to propose it separately.

swarm/src/lib.rs
```diff
@@ -732,10 +750,7 @@ where TBehaviour: NetworkBehaviour<ProtocolsHandler = THandler>,
                 },
                 Poll::Ready(NetworkBehaviourAction::ReportObservedAddr { address, score }) => {
                     for addr in this.network.address_translation(&address) {
-                        if this.external_addrs.iter().all(|a| a.addr != addr) {
-                            this.behaviour.inject_new_external_addr(&addr);
-                        }
-                        this.external_addrs.add(addr, score);
+                        this.add_external_address(addr, score);
```
romanb, line 753: :+1:

core/src/network.rs
dvc94ch: This is to avoid cloning/collecting when calling add_external_addrs when the Report observed address event happens

romanb: I see, I missed that :+1: .

romanb (approved):

mxinden (approved):
Looks good to me. Thanks for following this all the way through @dvc94ch!

dvc94ch:
A release with these changes would be nice for ipfs-embed (as there have been quite a few changes due to using master with Deref/DerefMut being removed etc.)

mxinden:
For the record, commit updating the Swarm changelog: https://github.com/libp2p/rust-libp2p/commit/bb206bcfdb8f43b71a560cc6938080369b033d48

mxinden:
> A release with these changes would be nice for ipfs-embed (as there have been quite a few changes due to using master with Deref/DerefMut being removed etc.)

Do I understand correctly that you are already running off of `libp2p` `master`? If so, how would a release of `libp2p-swarm` help? In either case you would have to tackle the breaking changes, no?

(As an aside, the script [here](https://github.com/libp2p/rust-libp2p/blob/master/swarm/CHANGELOG.md#0290-unreleased) might help with the upgrade.)

I am not opposed to doing a new release, it is just a bit of work as it does not only require a new release of `libp2p-swarm` but all its dependents as well.

dvc94ch:
well, I mean that it helps stop needing master, and because there have been a few changes on master backporting stuff to the current release is a bit annoying.

tomaka:
Would it be possible to get better pull request descriptions in the future? :pray: 
Having to read a diff when updating to a new version of libp2p to figure out which methods have been added and changed is a bit annoying, to say the least.

mxinden:
:+1: for better pull request descriptions.

To better understand your process @tomaka, did you get here via a `git blame` or via the [CHANGELOG.md](https://github.com/libp2p/rust-libp2p/commit/bb206bcfdb8f43b71a560cc6938080369b033d48) entry?

PR #2019, opened 2021-03-27
tcp: Make send/recv buffer size configurable.

dvc94ch:


transports/tcp/src/lib.rs
```diff
@@ -84,6 +84,10 @@ pub struct GenTcpConfig<T> {
     backlog: u32,
     /// The configuration of port reuse when dialing.
     port_reuse: PortReuse,
+    /// `SO_RCV_BUF` to set the receive buffer size.
```
mxinden, line 87:
```suggestion
    /// `SO_RCV_BUF` to set the receive buffer size, or `None` to keep the default.
```

transports/tcp/src/lib.rs
```diff
@@ -84,6 +84,10 @@ pub struct GenTcpConfig<T> {
...
+    recv_buffer_size: Option<usize>,
+    /// `SO_SNDBUF` to set the send buffer size.
```
mxinden, line 89:
```suggestion
    /// `SO_SNDBUF` to set the send buffer size, or `None` to keep the default.
```

transports/tcp/src/lib.rs
```diff
@@ -180,6 +184,8 @@ where
             nodelay: None,
             backlog: 1024,
             port_reuse: PortReuse::Disabled,
+            send_buffer_size: None,
```
mxinden, line 187: The doc comment of `fn new` would need updating as well. Wording should be similar to `IP_TTL`.

transports/tcp/src/lib.rs
```diff
@@ -327,6 +345,16 @@ where
         if let Some(nodelay) = self.nodelay {
             socket.set_nodelay(nodelay)?;
         }
+        if let Some(size) = self.send_buffer_size {
+            socket.set_send_buffer_size(size)?;
+            let actual = socket.send_buffer_size()?;
```
mxinden, line 350: Not sure how expensive this call is. Might be worth only doing it when debug or higher is enabled.

mxinden (commented):
What is your use-case for setting these size limits?

dvc94ch:
I observed a massive reduction in performance when adding a delay to connections with `netsim-embed`. The window size is the limiting factor as you can't fill the channel with enough data. I agree it's not very useful to set the buffer sizes manually as the default settings limit the size substantially. I also didn't observe an improvement to performance by increasing the buffer sizes, although I was able to verify that `iperf` manages to reach the theoretical optimum when run in `netsim-embed`.

I rewrote quic support for libp2p this week as the previous one was littered with todo's and added unnecessary abstractions and complexity. I'll finish testing it soon and compare it to tcp/yamux and tcp/mplex performance.

zmerp:
Any reason why this PR was closed?

dvc94ch:
no longer interested in the patch, you're welcome to reopen if you find it useful

Issue #2022, opened 2021-03-29
`/dns/` is no longer resolving records that are actually resolvable


montekki:
Addr of bootnode in form `/dns/bootnode/tcp/30333/p2p/12D3KooWEyoppNCUx8Yx66oV9fJnriXwCcXwDDUA2kj6vnc6iDEp` can no longer be reached, however `curl bootnode:9944` there from the node's machine works perfectly fine, [here are the logs of the node](https://gist.github.com/montekki/59af216a353e624c05e896bd8933f32b).

tomaka:
Did you replace the actual domain name with `bootnode` in the logs, or is `bootnode` some sort of alias?
If it's an alias, we'll need more information.

montekki:
The logs are unmodified copypaste, the `bootnode` is the name within the k8s cluster enviroment that can be `nslookup`ed: 

```
kubectl exec -i -t dnsutils -n fedor-gurke -- nslookup bootnode
Server:         10.27.240.10
Address:        10.27.240.10#53

Name:   bootnode.fedor-gurke.svc.cluster.local
Address: 10.27.247.34
```

tomaka:
Ah, so it's not a full qualified domain name.

@mxinden Do you know why we do [this](https://github.com/libp2p/rust-libp2p/blob/bb206bcfdb8f43b71a560cc6938080369b033d48/transports/dns/src/lib.rs#L485-L491)?

romanb:
> Do you know why we do this?

Really just as a naive attempt at optimisation that I didn't think much about, i.e. [from the docs](https://docs.rs/trust-dns-resolver/0.20.1/trust_dns_resolver/struct.Resolver.html#method.lookup_ip):

> For the least expensive query a fully-qualified-domain-name, FQDN, which ends in a final ., e.g. www.example.com., will only issue one query. Anything else will always incur the cost of querying the ResolverConfig::domain and ResolverConfig::search.

This obviously does not work with these kinds of indirections that require taking into account the local domain or search domains from the resolver configuration. I didn't really consider that at the time (i.e. we're taking about [this](https://github.com/bluejekyll/trust-dns/blob/main/crates/resolver/src/async_resolver.rs#L304-L341)). In hindsight we certainly cannot assume to always be given a domain that is an FQDN if we just stick a `.` at the end.

montekki:
So, is there a chance for a fix, or should we just continue with FQDNs for now (a bit tricky for our case)?

[...]

PR #2027, opened 2021-04-01
Remove fqdn() function from libp2p-dns

tomaka:
Fix https://github.com/libp2p/rust-libp2p/issues/2022

See [this comment](https://github.com/libp2p/rust-libp2p/issues/2022#issuecomment-809733928)

### DIFF
--- a/transports/dns/src/lib.rs
+++ b/transports/dns/src/lib.rs
@@ -45,7 +45,7 @@ use libp2p_core::{
     transport::{TransportError, ListenerEvent}
 };
 use smallvec::SmallVec;
-use std::{borrow::Cow, convert::TryFrom, error, fmt, iter, net::IpAddr, str};
+use std::{convert::TryFrom, error, fmt, iter, net::IpAddr, str};
 #[cfg(any(feature = "async-std", feature = "tokio"))]
 use std::io;
 #[cfg(any(feature = "async-std", feature = "tokio"))]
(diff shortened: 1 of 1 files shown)


mxinden (commented):
Do I understand correctly that without `fqdn()`, i.e. without (falsely) making all domains fully-qualified domains, we do two queries for each domain, even though one might suffice for fully-qualified domains? Put differently, this change has a performance impact only, while restoring correctness across all cases?

tomaka:
As far as I understand, yes

mxinden:
@tomaka have you had the chance to test both the fully qualified and the not fully qualified case?

tomaka:
I did not test this, as the fix seems obvious to me.
The `trust-dns` library expects a domain name. We do some pre-transformation on the domain name passed by the user. This PR removes this transformation so that `trust-dns` gets passed the exact value that the user requested.
If this introduces a bug, it can only either be in `trust-dns` or was already present before this PR.

mxinden (approved):
Just tested this with a non fully qualified domain locally with `DnsConfig::system()` which works as expected.

I can cut a new patch release of `libp2p-dns`. Is that good enough for you @montekki?

montekki:
Yeah as soon as we can propagate it into the `polkadot` codebase

mxinden:
`libp2p-dns` `v0.28.1` is [published](https://crates.io/crates/libp2p-dns/0.28.1) and [tagged](https://github.com/libp2p/rust-libp2p/releases/tag/libp2p-dns-0.28.1). Thanks for the patch @tomaka.

PR #2029, opened 2021-04-01
Derive debug

dvc94ch:
I can split them into two or three different PR's if you like, but because of the size of the changes, I didn't think it was worth it.

### DIFF
--- a/protocols/identify/src/identify.rs
+++ b/protocols/identify/src/identify.rs
@@ -85,6 +85,7 @@ enum Reply {
 
 /// Configuration for the [`Identify`] [`NetworkBehaviour`].
 #[non_exhaustive]
+#[derive(Debug)]
 pub struct IdentifyConfig {
     /// Application-specific version of the protocol family used by the peer,
     /// e.g. `ipfs/1.0.0` or `polkadot/1.0.0`.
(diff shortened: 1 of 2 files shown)


protocols/identify/src/identify.rs
```diff
@@ -174,6 +175,11 @@ impl Identify {
         }
     }
 
+    /// Returns the `agent_version` of the node.
+    pub fn agent_version(&self) -> &str {
```
mxinden, line 179: Can you expand on why you need this function? After all the `agent_version` has been defined by the user already.

dvc94ch: Useful for debugging. Since it's an arbitrary string you can include it in log messages to identify the swarm they originated from. Peerids are less suited and since they are generated randomly also aren't the same across test runs

dvc94ch: Actually substrate uses it in that way too

mxinden: Right, but why offer a getter for a static value that the caller passed at construction time. Am I missing something?

dvc94ch: Why store multiple copies of the same string in different places for no reason? But sure it is possible to copy the string around which is what I've been doing so far.

dvc94ch: I guess I can live without

dvc94ch:
ping @mxinden

mxinden (approved):

PR #2035, opened 2021-04-07
src: update tutorial with correct static methods in ping example

therealdannzor:
Running the example in a separate project results in errors because of two missing method implementations. This small update makes the ping-pong example work 😄 .

### DIFF
--- a/src/tutorial.rs
+++ b/src/tutorial.rs
@@ -238,7 +238,7 @@
 //!
 //!     // Tell the swarm to listen on all interfaces and a random, OS-assigned
 //!     // port.
-//!     swarm.listen_on("/ip4/0.0.0.0/tcp/0".parse()?)?;
+//!     Swarm::listen_on(&mut swarm, "/ip4/0.0.0.0/tcp/0".parse()?)?;
 //!
 //!     // Dial the peer identified by the multi-address given as the second
 //!     // command-line argument, if any.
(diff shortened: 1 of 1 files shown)


mxinden (commented):
Thanks for testing out the tutorial @therealdannzor and also thanks for preparing a patch!

https://github.com/libp2p/rust-libp2p/pull/1995 has been merged since the last release, allowing calling methods on a `Swarm` directly. Thus with the next release this will be fixed. I will prepare a release soon to make sure the confusion is removed.

If you link to the current `master` branch directly, things should compile fine.

See https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-git-repositories

therealdannzor:
> Thanks for testing out the tutorial @therealdannzor and also thanks for preparing a patch!
> 
> #1995 has been merged since the last release, allowing calling methods on a `Swarm` directly. Thus with the next release this will be fixed. I will prepare a release soon to make sure the confusion is removed.
> 
> If you link to the current `master` branch directly, things should compile fine.
> 
> See https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-git-repositories

Thanks for letting me know and for the quick reply, I wasn't aware of this planned fix. I'll close this then :)

PR #2038, opened 2021-04-09
*: Drop generic wasm32-unknown-unknown support

mxinden:
With `rand` `v0.8.0` platform support changed [1] due to its upgrade to
`getrandom` `v0.2`. With `getrandom` `v0.2` `wasm32-unknown-unknown` is no
longer supported out of the box:

> This crate fully supports the wasm32-wasi and wasm32-unknown-emscripten
> targets. However, the wasm32-unknown-unknown target is not automatically
> supported since, from the target name alone, we cannot deduce which JavaScript
> interface is in use (or if JavaScript is available at all).
>
> Instead, if the "js" Cargo feature is enabled, this crate will assume that you
> are building for an environment containing JavaScript, and will call the
> appropriate methods. Both web browser (main window and Web Workers) and
> Node.js environments are supported, invoking the methods described above using
> the wasm-bindgen toolchain.
>
> This feature has no effect on targets other than wasm32-unknown-unknown.

This commit drops support for wasm32-unknown-unknown in favor of the two more
specific targets wasm32-wasi and wasm32-unknown-emscripten.

Note on `resolver = "2"`: The new resolver is required to prevent features
being mixed, more specifically to prevent libp2p-noise to build with the
`ring-resolver` feature. See [3] for details.

---

[1] https://github.com/rust-random/rand/blob/master/CHANGELOG.md#platform-support
[2] https://docs.rs/getrandom/0.2.2/getrandom/#webassembly-support
[3] https://doc.rust-lang.org/nightly/cargo/reference/features.html#feature-resolver-version-2

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -63,7 +63,7 @@ jobs:
       uses: actions-rs/toolchain@v1
       with:
         toolchain: stable
-        target: wasm32-wasmi
+        target: wasm32-wasi
 
     - name: Install a recent version of clang
       run: |
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,18 +53,6 @@ jobs:
 
     - uses: actions/checkout@v2
 
-    - name: Install Rust wasm32-unknown-unknown
-      uses: actions-rs/toolchain@v1
-      with:
-        toolchain: stable
-        target: wasm32-unknown-unknown
-
-    - name: Install Rust wasm32-wasi
-      uses: actions-rs/toolchain@v1
-      with:
-        toolchain: stable
-        target: wasm32-wasi
-
     - name: Install a recent version of clang
       run: |
         wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key | apt-key add -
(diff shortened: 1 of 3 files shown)


tomaka:
After reading into the details, `getrandom` v0.2 doesn't actually change anything. It renames the `wasm-bindgen` and `stdweb` features to `js`, but doesn't fundamentally stops supporting `wasm32-unknown-unknown`.

A user of `rust-libp2p` can add an explicit dependency to `getrandom` in their Cargo.toml in order to enable the feature. This is exactly the same as before.

Still, I believe it makes sense to me to drop support for this weird `unknown` variant that doesn't really make sense.

tomaka (approved):

mxinden:
Thanks @tomaka for the review.

> A user of rust-libp2p can add an explicit dependency to getrandom in their Cargo.toml in order to enable the feature. This is exactly the same as before.
>
> I believe it makes sense to no longer try to maintain this weird unknown variant that doesn't really make sense, so +1 for this PR. However, this PR doesn't make rust-libp2p stop working on that platform.

:+1:

PR #2051, opened 2021-04-14
Handle websocket CLOSE with reason code

dvdplm:
Pass websocket CLOSE reason to the app as an `IncomingData::Closed(soketto::CloseReason)`.

This PR is a companion to https://github.com/paritytech/soketto/pull/31 and lets applications know what the remote end claimed to be the reason for closing the connection (if any).

Will keep this PR as a draft until https://github.com/paritytech/soketto/pull/31 is merged&released.


transports/websocket/src/framed.rs
```diff
@@ -475,11 +477,18 @@ impl IncomingData {
...
+        if let IncomingData::Closed(_) = self { true } else { false }
+    }
+
     pub fn into_bytes(self) -> Vec<u8> {
         match self {
             IncomingData::Binary(d) => d,
             IncomingData::Text(d) => d,
-            IncomingData::Pong(d) => d
+            IncomingData::Pong(d) => d,
+            IncomingData::Closed(CloseReason { code, ..}) => {
```
mxinden, line 489: What would be the use-case where one would want to transform a `Closed` into a vector of bytes? If there is none, I would prefer changing the signature of `into_bytes` to an `Option<Vec<u8>>` where `Closed` returns `None`.

mxinden (commented):
Extending `IncomingData` with `Closed` is not intuitive to me, i.e. I would not consider a closing event as _data_. How about renaming the `enum` to `Event` or `ConnectionEvent`?

transports/websocket/src/framed.rs
dvdplm: This is a wart for sure. And I guess the same argument could be made for `Pong`: what does transforming a `PONG` to bytes even mean?
I'm not sure if changing the signature helps significantly though, `into_*` methods should transform `owned –> owned` and be cheap/free. As a user I'd be a little surprised by a `into_bytes()` that returns `None`.

All of that to sum up my sentiment: shoehorning in `CLOSE` here is ugly but it already was ugly, given `PONG`. I think the real fix is to refactor things such that `into_bytes` needn't be called at all; which at the time felt a bit out of scope.

transports/websocket/src/framed.rs
```diff
@@ -489,7 +498,8 @@ impl AsRef<[u8]> for IncomingData {
         match self {
             IncomingData::Binary(d) => d,
             IncomingData::Text(d) => d,
-            IncomingData::Pong(d) => d
+            IncomingData::Pong(d) => d,
+            IncomingData::Closed(_) => unimplemented!(),
```
dvdplm, line 502: @mxinden this is the exact same situation as above, except it's not even possible to pretend there are any inner bytes to reference.

dvdplm:
> Extending `IncomingData` with `Closed` is not intuitive to me, i.e. I would not consider a closing event as _data_. How about renaming the `enum` to `Event` or `ConnectionEvent`?

What do you think of just `Incoming`, like `soketto::Incoming` ([docs](https://docs.rs/soketto/0.5.0/soketto/data/enum.Incoming.html))?

transports/websocket/src/framed.rs
dvdplm: Ouch, I went to update my substrate branch with this and got to (my own) [comment explaining why this crap is needed](https://github.com/paritytech/substrate/blob/e8c75bc526984da44f22f5d2a84191f04e4209d9/client/telemetry/src/transport.rs#L63-L67).

PR #2053, opened 2021-04-21
Fix mdns taking a long time to discover peers.

dvc94ch:
If you start listening after mdns joined a multicast group, the peers may not discover eachother until the 5min timeout expires.

### DIFF
--- a/protocols/mdns/CHANGELOG.md
+++ b/protocols/mdns/CHANGELOG.md
@@ -1,6 +1,7 @@
 # 0.30.1 [2021-04-21]
 
 - Fix timely discovery of peers after listening on a new address.
+  [PR 2053](https://github.com/libp2p/rust-libp2p/pull/2053/)
 
 # 0.30.0 [2021-04-13]
 
--- a/protocols/mdns/CHANGELOG.md
+++ b/protocols/mdns/CHANGELOG.md
@@ -1,3 +1,7 @@
+# 0.30.1 [2021-04-21]
+
+- Fix timely discovery of peers after listening on a new address.
+
 # 0.30.0 [2021-04-13]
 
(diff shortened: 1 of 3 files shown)


mxinden (commented):
Looks good to me, thanks.

Could you bump the `libp2p-mdns` version in `protocols/mdns/Cargo.toml` and `Cargo.toml` and add a changelog entry in `protocols/mdns/CHANGELOG.md`?

dvc94ch:
I'm wondering now if we need if-watch in mdns at all. Since the transport already uses if-watch and detects address changes, we can probably move the join_multicast/leave_multicast to the inject_listen_addr and inject_expired_listen_addr. Not sure how well it will work on tokio-tcp.

dvc94ch:
but that would be a breaking change, I still think we can merge and release this, I'll try removing if-watch and see what happens.

dvc94ch:
ah won't work. even if you don't have a listening address you still want to discover other peers.

protocols/mdns/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# 0.30.1 [2021-04-21]
+
+- Fix timely discovery of peers after listening on a new address.
```
mxinden, line 3:
```suggestion
- Fix timely discovery of peers after listening on a new address.
  [PR 2053](https://github.com/libp2p/rust-libp2p/pull/2053/)
```

mxinden (approved):

mxinden:
[Tagged](https://github.com/libp2p/rust-libp2p/releases/tag/libp2p-mdns-0.30.1) and [published](https://crates.io/crates/libp2p-mdns/0.30.1).

Issue #2013, opened 2021-03-24
Provide Open Metrics (/Prometheus) Wrapper


mxinden:
By design rust-libp2p does not expose Prometheus style metrics. Instead, given its event-based architecture, one should record the events emitted by rust-libp2p as a user.

Consider providing a wrapper around `libp2p-swarm` which records emitted events before bubbling them upwards. In case a generic wrapper is not feasible, offer a tutorial on how to instrument ones rust-libp2p instance.

The Prometheus format has been standardized, in a backwards compatible way, under the [Open Metrics](https://github.com/OpenObservability/OpenMetrics/) name, thus supporting Open Metrics instead of the old Prometheus style format is both more future-proof and likely better supported by monitoring systems.

Libraries:
- `open-metrics-client`: https://github.com/mxinden/rust-open-metrics-client
- `prometheus`: https://github.com/tikv/rust-prometheus/

PR #2063, opened 2021-04-30
misc/metrics: Add auxiliary crate to record events as OpenMetrics

mxinden:
## Motivation

Metrics (e.g. Prometheus or [OpenMetrics]) enable one to gain a deep understanding of a system, both while developing and operating it.

I find myself re-implementing metric instrumentation for rust-libp2p over and over again. The most advanced metric setup likely lives in [Substrate](https://github.com/paritytech/substrate/blob/master/client/network/src/service/metrics.rs). Other smaller setups can e.g. be found in the [Kademlia exporter](https://github.com/mxinden/kademlia-exporter). Instead of duplicating most of this logic once more for [rust-libp2p-relay-server](https://github.com/mxinden/rust-libp2p-relay-server) I would like to propose collaborating on a set of base metrics via a new crate here in rust-libp2p - namely `libp2p-metrics` - providing extensive instrumentation to rust-libp2p users out-of-the-box.

In the long term, I would like to extend this effort beyond the definition and exposition of metrics, all the way to Grafana dashboard definitions and alerting rules. See [Substrate/.maintain](https://github.com/paritytech/substrate/tree/master/.maintain/monitoring) as an example.

## Implementation

As said in https://github.com/libp2p/rust-libp2p/issues/2013 I don't plan to introduce [OpenMetrics] in each rust-libp2p crate itself, but instead I want to provide an optional auxiliary crate that can live on-the-side and can be used on-demand by recording events emitted by the Swarm and various protocols.

I have not yet found a great abstraction that is both intuitive and extensible. For now I am proposing the below, though I am more than happy for alternative suggestions.

```rust
let mut swarm = Swarm::new(
    block_on(libp2p::development_transport(local_key))?,
    Ping::new(PingConfig::new().with_keep_alive(true)),
    local_peer_id,
);

let mut metric_registry = Registry::default();
let metrics = Metrics::new(&mut metric_registry);
thread::spawn(move || block_on(metrics_server(metric_registry)));

block_on(async {
    loop {
        match swarm.next_event().await {
            SwarmEvent::Behaviour(ping_event) => {
                metrics.record(&ping_event);
            }
            swarm_event => {
                metrics.record(&swarm_event);
            }
        }
    }
})
```

You can find the full example in `misc/metrics/examples/metrics.rs` in this pull request.

Note, this pull request is using the [`open-metrics-client`](https://github.com/mxinden/rust-open-metrics-client) crate (authored by me) instead of the more popular [`prometheus`](https://github.com/tikv/rust-prometheus/) crate. You can find a high-level comparison on https://github.com/tikv/rust-prometheus/issues/392.

[OpenMetrics]: https://github.com/OpenObservability/OpenMetrics/

Fixes https://github.com/libp2p/rust-libp2p/issues/2013


mxinden:
The new crate proposed here - `libp2p-metrics` - is already used by all deployments of https://github.com/mxinden/rust-libp2p-relay-server. To find an abstraction that works for many projects instead of just one, I would like to collaborate with someone, preferably with a large code-base, helping them introduce `libp2p-metrics` into their project. (It sounds a lot more complicated than it is.)

Any volunteers? @ec2 would this be interesting for you and https://github.com/ChainSafe/forest?

ec2:
This is very cool!  @olibearo from our team will be leading the metrics and instrumentation efforts. I think we would be very down. Let me direct Jorge's attention to this thread so he can chime in.

dvc94ch:
ipfs-embed uses something prometheus based. while I probably won't change it for the time being, I will eventually rewrite it and make the core abstraction provided to applications a stream instead of a block and remove all the complex garbage collection etc. this crate does seem useful, thanks!

mxinden:
Great to hear @dvc94ch, let me know in case you every end up using it.

PR #2065, opened 2021-05-04
Fix discovered event emission.

dvc94ch:
mdns keeps rediscovering nodes. this PR changes that to only emit events for new nodes it discovered. In addition we make sure to only send a query if it is really needed. Some logging is added for debugging purposes.

### DIFF
--- a/protocols/mdns/src/behaviour.rs
+++ b/protocols/mdns/src/behaviour.rs
@@ -85,9 +85,6 @@ pub struct Mdns {
     /// Iface watcher.
     if_watch: IfWatcher,
 
-    /// Send a query.
-    query: bool,
-
     /// Buffer used for receiving data from the main socket.
     /// RFC6762 discourages packets larger than the interface MTU, but allows sizes of up to 9000
     /// bytes, if it can be ensured that all participating devices can handle such large packets.
@@ -146,7 +143,6 @@ impl Mdns {
             recv_socket,
             send_socket,
             if_watch,
-            query: false,
             recv_buffer: [0; 4096],
             send_buffer: Default::default(),
             discovered_nodes: SmallVec::new(),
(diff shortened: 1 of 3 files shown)


dvc94ch:
@mxinden thoughts?

mxinden:
Could you please describe the issue the status quo is facing, and the solution you are proposing, in the pull request description?

protocols/mdns/src/behaviour.rs
```diff
@@ -85,6 +85,9 @@ pub struct Mdns {
     /// Iface watcher.
     if_watch: IfWatcher,
 
+    /// Send a query.
+    query: bool,
```
mxinden, line 89: In case we do go with the extra boolean, this would need additional documentation.

protocols/mdns/src/behaviour.rs
```diff
@@ -271,7 +273,7 @@ impl NetworkBehaviour for Mdns {
     }
 
     fn inject_new_listen_addr(&mut self, _id: ListenerId, _addr: &Multiaddr) {
-        self.send_buffer.push_back(build_query());
+        self.query = true;
```
mxinden, line 276: Instead of the extra `bool` why not do:

```suggestion
        self.timer = Timer::set_interval(Instant::now(), self.query_interval);
```

That would make the interval fire at the next `poll_next`. As far as I can tell there would be no need for `self.query` at all, right?

mxinden (approved):
Great. Thanks for the additional commits.

Would you mind:
- Adding an entry in `protocols/mdns/CHANGELOG.md`.
- Bumping the patch version in `protocols/mdns/Cargo.toml`.

I can cut a patch release later today or tomorrow if you like.

dvc94ch:
Thanks

PR #2066, opened 2021-05-05
misc/multistream-select: Implement simultaneous open extension

mxinden:
From the multistream-select 1.0 simultaneous open protocol extension
specification:

> In order to support direct connections through NATs with hole
punching, we need to account for simultaneous open. In such cases, there
is no single initiator and responder, but instead both peers act as
initiators. This breaks protocol negotiation in multistream-select,
which assumes a single initator.
>
> This draft proposes a simple extension to the multistream protocol
negotiation in order to select a single initator when both peers are
acting as such.

See https://github.com/libp2p/specs/pull/196/ for details.

This commit implements the above specification, available via
`Version::V1SimultaneousOpen`.

---

Given that the extension is backward compatible, there is no gain for rust-libp2p to support it without supporting the larger effort of TCP hole punching.

To test this against the Golang implementation:

``` rust
#[test]
fn sim_open_server() {
    futures::executor::block_on(async {
        let listener = TcpListener::bind("127.0.0.1:5000").await.unwrap();
        let listener_addr = listener.local_addr().unwrap();
        println!("{:?}", listener_addr);

        let connec = listener.accept().await.unwrap().0;
        let protos = vec![b"/b", b"/a"];
        let (proto, io) = dialer_select_proto_serial(connec, protos, Version::V1SimOpen).await.unwrap();
        assert_eq!(proto, b"/a");
        io.complete().await.unwrap();
    });
}

#[test]
fn sim_open_client() {
    futures::executor::block_on(async {
        let connec = TcpStream::connect("127.0.0.1:5000").await.unwrap();
        let protos = vec![b"/b", b"/a"];
        let (proto, io) = dialer_select_proto_serial(connec, protos.into_iter(),Version::V1SimOpen)
            .await.unwrap();
        assert_eq!(proto, b"/a");
        io.complete().await.unwrap();
    });
}
```

``` golang
package main

import (
        // "fmt"
        "net"
        multistream "github.com/multiformats/go-multistream"
)

func sim_open_client() {
        conn, err := net.Dial("tcp", "127.0.0.1:5000")
        if err != nil {
                panic(err)
        }

        proto, server, err := multistream.SelectWithSimopenOrFail([]string{"/a"}, conn)
        if err != nil {
                panic(err)
        }

        if proto != "/a" {
                panic(proto)
        }

        if server {
                fmt.Println("Server")
        } else {
                fmt.Println("Client")
        }
}

func sim_open_server() {
        ln, err := net.Listen("tcp", "127.0.0.1:5000")
        if err != nil {
                panic(err)
        }
        conn, err := ln.Accept()
        if err != nil {
                panic(err)
        }

        mux := multistream.NewMultistreamMuxer()
	mux.AddHandler("/a", nil)

        proto, _, err := mux.Negotiate(conn)
        if err != nil {
                panic(err)
        }

        if proto != "/a" {
                panic(proto)
        }
}
  ```

### CODE REVIEW

**Comment by dvc94ch** [comment]
I'm seeing this in our ci logs:
ipfs_embed::net::peers: address reach failure Pending connection: Transport error: upgrade apply error: invalid data

seems like it happens when two peers try to dial eachother at the same time. am I correct in assuming this PR will fix that?

**Comment by mxinden** [comment]
> seems like it happens when two peers try to dial eachother at the same time. am I correct in assuming this PR will fix that?

Correct, though note that both peers need to run this patch.

See https://github.com/libp2p/specs/blob/master/connections/simopen.md for more details.

> Looks like I'll have to try this out on Monday.

:tada: please let me know how it goes.

**Comment by dvc94ch** [comment]
While trying to reproduce the first log and check that this PR fixes it I ran into a different error:
ipfs_embed::net::peers: address reach failure Pending connection: Transport error: Cannot assign requested address (os error 99)

File: misc/multistream-select/src/dialer_select.rs [hunk 1]
```diff
@@ -172,11 +190,123 @@ where
...
+                        Message::Protocol(p) if p == SIM_OPEN_ID => {
+                            let selection = SimOpenPhase {
+                                state: SimOpenState::SendNonce{ io },
+                            };
+                            *this.state = SeqState::SimOpenPhase { selection, protocol };
+                        }
+                        Message::NotAvailable => {
+                            *this.state = SeqState::AwaitProtocol { io, protocol }
+                        }
+                        _ => return Poll::Ready(Err(ProtocolError::InvalidMessage.into()))
```
rkuhn, line 276: so currently no fallback to old style in case the peer isn’t ready for V1SimOpen, right?

misc/multistream-select/src/dialer_select.rs
```diff
@@ -261,6 +400,168 @@ where
...
+    R: AsyncRead + AsyncWrite + Unpin,
+{
+    type Output = Result<(MessageIO<R>, SimOpenRole), NegotiationError>;
+
+    fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+
+        loop {
+            match mem::replace(&mut self.state, SimOpenState::Done) {
+                SimOpenState::SendNonce { mut io } => {
+                    match Pin::new(&mut io).poll_ready(cx)? {
```
rkuhn, line 435: shouldn’t this always succeed? (I’m not deep inside libp2p — the last relevant call looks to have been `poll_flush`, whose name seems to suggest that the socket should be ready)

dvc94ch: I think in libp2p-quic we're ignoring the flush. a substream is flushed on close, but there is no support in quinn to manually flush a stream.

- [0] https://github.com/ipfs-rust/libp2p-quic/blob/master/src/muxer.rs#L366

misc/multistream-select/src/dialer_select.rs
dvc94ch: I think it's supposed to be backwards compatible.

dvc94ch:
@mxinden this PR does not fix or fully fix the issue. I have a reproducible test case [0] which uses netsim-embed to simulate a simultaneous open. It's not possible to verify this without netsim-embed, because one swarm will dial first and the second swarm's dial will be rejected due to it having the same five tuple. However with netsim-embed we can simulate two different machines where they dial before receiving the first packet from the peer.

- [0] https://github.com/ipfs-rust/ipfs-embed/pull/80

misc/multistream-select/src/dialer_select.rs
mxinden: When using `Version::V1SimOpen` the dialer expects either:

- `/libp2p/simultaneous-connect`  in which case the remote indicates that it operates in dialing mode as well, i.e. a simultaneously dialed connection or
- `na` in which case the remote indicates that it is a listener, i.e. not a simultaneously dialed connection, and thus we fall back to the "old style" via `SeqState::AwaitProtocol`.

Any other message would be a protocol violation, both with and without the simultaneous open extension.

@rkuhn does that make sense?

misc/multistream-select/src/dialer_select.rs
mxinden: I am not quite sure I follow @rkuhn, `poll_flush` does not imply the same as `poll_ready` (see the [`Sink` docs](https://docs.rs/futures/0.3.15/futures/sink/trait.Sink.html#tymethod.poll_flush)), thus one can not call `start_send` right away. Does that make sense?

misc/multistream-select/src/dialer_select.rs
rkuhn: Yes, that `na` bit was the missing puzzle piece, thanks!

misc/multistream-select/src/dialer_select.rs
rkuhn: Re-reading the Sink docs I see what you mean, although the docs are still somewhat weird: I was tripped up by the documentation of `start_send` which basically implies that `poll_flush` is stronger than `poll_ready`. And I have a hard time imagining a useful implementation where `poll_ready` wouldn’t be immediately ready directly after a `poll_flush`.

dvc94ch:
Looks like just because multistream-select decided that machine #0 is the responder, noise still thinks it's an initiator.
```sh
Machine-#0 (stdout): Jun 08 12:29:21.575 TRACE multistream_select::protocol: Received message: Responder    
Machine-#1 (stdout): Jun 08 12:29:21.575 TRACE multistream_select::protocol: Received message: Initiator    
Machine-#0 (stdout): Jun 08 12:29:21.575 DEBUG multistream_select::dialer_select: Dialer: Proposed protocol: /noise    
Machine-#1 (stdout): Jun 08 12:29:21.778 TRACE multistream_select::protocol: Received message: Protocol(Protocol(b"/noise"))    
Machine-#1 (stdout): Jun 08 12:29:21.779 DEBUG multistream_select::listener_select: Listener: confirming protocol: /noise    
Machine-#1 (stdout): Jun 08 12:29:21.779 DEBUG multistream_select::listener_select: Listener: sent confirmed protocol: /noise    
Machine-#1 (stdout): Jun 08 12:29:21.780 TRACE libp2p_noise::io::framed: write state Ready    
Machine-#1 (stdout): Jun 08 12:29:21.781 TRACE libp2p_noise::io::framed: write: cipher text len = 32 bytes    
Machine-#1 (stdout): Jun 08 12:29:21.782 TRACE libp2p_noise::io::framed: write state WriteLen { len: 32, buf: [0, 32], off: 0 }    
Machine-#1 (stdout): Jun 08 12:29:21.782 TRACE libp2p_noise::io::framed: write: frame len (32, [0, 32], 0/2)    
Machine-#1 (stdout): Jun 08 12:29:21.782 TRACE libp2p_noise::io::framed: write state WriteData { len: 32, off: 0 }    
Machine-#1 (stdout): Jun 08 12:29:21.782 TRACE libp2p_noise::io::framed: write: 32/32 bytes written    
Machine-#1 (stdout): Jun 08 12:29:21.782 TRACE libp2p_noise::io::framed: write: finished with 32 bytes    
Machine-#1 (stdout): Jun 08 12:29:21.782 TRACE libp2p_noise::io::framed: write state Ready    
Machine-#1 (stdout): Jun 08 12:29:21.782 TRACE libp2p_noise::io::framed: read state: Ready    
Machine-#1 (stdout): Jun 08 12:29:21.782 TRACE libp2p_noise::io::framed: read state: ReadLen { buf: [0, 0], off: 0 }    
Machine-#0 (stdout): Jun 08 12:29:21.981 TRACE multistream_select::protocol: Received message: Protocol(Protocol(b"/noise"))    
Machine-#0 (stdout): Jun 08 12:29:21.981 DEBUG multistream_select::dialer_select: Dialer: Received confirmation for protocol: /noise    
Machine-#0 (stdout): Jun 08 12:29:21.982 TRACE libp2p_noise::io::framed: write state Ready    
Machine-#0 (stdout): Jun 08 12:29:21.983 TRACE libp2p_noise::io::framed: write: cipher text len = 32 bytes    
Machine-#0 (stdout): Jun 08 12:29:21.983 TRACE libp2p_noise::io::framed: write state WriteLen { len: 32, buf: [0, 32], off: 0 }    
Machine-#0 (stdout): Jun 08 12:29:21.983 TRACE libp2p_noise::io::framed: write: frame len (32, [0, 32], 0/2)    
Machine-#0 (stdout): Jun 08 12:29:21.983 TRACE libp2p_noise::io::framed: write state WriteData { len: 32, off: 0 }    
Machine-#0 (stdout): Jun 08 12:29:21.983 TRACE libp2p_noise::io::framed: write: 32/32 bytes written    
Machine-#0 (stdout): Jun 08 12:29:21.983 TRACE libp2p_noise::io::framed: write: finished with 32 bytes    
Machine-#0 (stdout): Jun 08 12:29:21.983 TRACE libp2p_noise::io::framed: write state Ready    
Machine-#0 (stdout): Jun 08 12:29:21.983 TRACE libp2p_noise::io::framed: read state: Ready    
Machine-#0 (stdout): Jun 08 12:29:21.983 TRACE libp2p_noise::io::framed: read state: ReadLen { buf: [0, 0], off: 0 }    
Machine-#0 (stdout): Jun 08 12:29:21.983 TRACE libp2p_noise::io::framed: read state: ReadLen { buf: [0, 0], off: 0 }    
Machine-#0 (stdout): Jun 08 12:29:21.984 TRACE libp2p_noise::io::framed: read: frame len = 32    
Machine-#0 (stdout): Jun 08 12:29:21.984 TRACE libp2p_noise::io::framed: read state: ReadData { len: 32, off: 0 }    
Machine-#0 (stdout): Jun 08 12:29:21.984 TRACE libp2p_noise::io::framed: read: 32/32 bytes    
Machine-#0 (stdout): Jun 08 12:29:21.984 TRACE libp2p_noise::io::framed: read: decrypting 32 bytes    
Machine-#0 (stdout): Jun 08 12:29:21.984 DEBUG libp2p_noise::io::framed: read: decryption error    
Machine-#0 (stdout): Jun 08 12:29:21.985 DEBUG libp2p_swarm: Connection attempt to PeerId("12D3KooWMm1c4pzeLPGkkCJMAgFbsfQ8xmVDusg272icWsaNHWzN") via "/ip4/192.168.39.3/tcp/30000/p2p/12D3KooWMm1c4pzeLPGkkCJMAgFbsfQ8xmVDusg272icWsaNHWzN" failed with Transport(Other(Custom { kind: Other, error: Transport(Custom { kind: Other, error: Other(A(B(Apply(Io(Kind(InvalidData)))))) }) })). Attempts remaining: 0.    
Machine-#0 (stdout): Jun 08 12:29:21.985 TRACE ipfs_embed::net::peers: address reach failure Pending connection: Transport error: upgrade apply error: invalid data
Machine-#0 (stdout): Jun 08 12:29:21.985 TRACE ipfs_embed::net::peers: removing address /ip4/192.168.39.3/tcp/30000/p2p/12D3KooWMm1c4pzeLPGkkCJMAgFbsfQ8xmVDusg272icWsaNHWzN for peer 12D3KooWMm1c4pzeLPGkkCJMAgFbsfQ8xmVDusg272icWsaNHWzN
Machine-#0 (stdout): Jun 08 12:29:21.985 TRACE ipfs_embed::net::peers: dial failure 12D3KooWMm1c4pzeLPGkkCJMAgFbsfQ8xmVDusg272icWsaNHWzN
Machine-#0 (stdout): Jun 08 12:29:21.985 TRACE ipfs_embed::net::peers: Unreachable(PeerId("12D3KooWMm1c4pzeLPGkkCJMAgFbsfQ8xmVDusg272icWsaNHWzN"))
Error: sim open failed
```

dvc94ch:
So this isn't that easy to fix. I don't think it makes sense to have this split between UpgradeOutbound/UpgradeInbound. Since an UpgradeOutbound also needs to be able to do what UpgradeInbound does. @mixden what are your thoughts on this?

dvc94ch:
anyway this is the approach I was pursuing, but giving up since the swarm types are too complicated and I'm not sure if this split makes sense. there should probably be a single `Upgrade` trait taking a role (Initiator/Responder) and sim-open can toggle the role.

```diff
+type Output<C, U> = <U as OutboundUpgrade<Negotiated<C>>>::Output;
+type Error<C, U> = <U as OutboundUpgrade<Negotiated<C>>>::Error;
+
 impl<C, U> Future for OutboundUpgradeApply<C, U>
 where
     C: AsyncRead + AsyncWrite + Unpin,
     U: OutboundUpgrade<Negotiated<C>>,
+    <U as OutboundUpgrade<Negotiated<C>>>::Output: 'static,
+    <U as OutboundUpgrade<Negotiated<C>>>::Error: 'static,
+    <U as OutboundUpgrade<Negotiated<C>>>::Future: Send + Unpin + 'static,
+    U: InboundUpgrade<Negotiated<C>,
+        Output = <U as OutboundUpgrade<Negotiated<C>>>::Output,
+        Error = <U as OutboundUpgrade<Negotiated<C>>>::Error>,
+
+    <U as InboundUpgrade<Negotiated<C>>>::Future: Send + Unpin + 'static,
 {
-    type Output = Result<U::Output, UpgradeError<U::Error>>;
+    type Output = Result<Output<C, U>, UpgradeError<Error<C, U>>>;
 
     fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
         loop {
             match mem::replace(&mut self.inner, OutboundUpgradeApplyState::Undefined) {
                 OutboundUpgradeApplyState::Init { mut future, upgrade } => {
-                    let (info, connection) = match Future::poll(Pin::new(&mut future), cx)? {
+                    let (info, role, connection) = match Future::poll(Pin::new(&mut future), cx)? {
                         Poll::Ready(x) => x,
                         Poll::Pending => {
                             self.inner = OutboundUpgradeApplyState::Init { future, upgrade };
                             return Poll::Pending
                         }
                     };
-                    self.inner = OutboundUpgradeApplyState::Upgrade {
-                        future: Box::pin(upgrade.upgrade_outbound(connection, info.0))
+                    let future: Pin<Box<dyn Future<Output = Result<Output<C, U>, Error<C, U>>> + Send + Unpin + 'static>> = match role {
+                        multistream_select::SimOpenRole::Initiator => Box::pin(upgrade.upgrade_outbound(connection, info.0)),
+                        multistream_select::SimOpenRole::Responder => Box::pin(upgrade.upgrade_inbound(connection, info.0)),
                     };
+                    self.inner = OutboundUpgradeApplyState::Upgrade { future };
                 }
                 OutboundUpgradeApplyState::Upgrade { mut future } => {
                     match Future::poll(Pin::new(&mut future), cx) {
@@ -230,4 +243,3 @@ impl<N: ProtocolName> AsRef<[u8]> for NameWrap<N> {
         self.0.protocol_name()
     }
 }
-
```

dvc94ch:
I think we need to keep the InboundUpgrade/OutboundUpgrade. Instead I propose simply adding a role arg to the OutboundUpgrade and handling it in noise.

mxinden:
> I think we need to keep the InboundUpgrade/OutboundUpgrade. Instead I propose simply adding a role arg to the OutboundUpgrade and handling it in noise.

_Thinking out loud_. Given that this only concerns connection upgrades and not substream upgrades and given that `Transport::upgrade` requires the upgrade to be both an `InboundUpgrade` and an `OutboundUpgrade`, would it not be possible to:

- Call `InboundUpgrade::upgrade_inbound` on inbound connections.
- Call `OutboundUpgrade::upgrade_outbound` or `InboundUpgrade::upgrade_inbound` on outbound connections depending on whether the _multistream select simultaneous open_ negotiation returned as an _Initiator_ or a _Responder_?

dvc94ch:
I think that would affect the identify protocol. It's ok to have a connection from two dialers, it's only the noise handshake that needs to change. Quic handles this internally so shouldn't be relevant.

dvc94ch:
Rereading the identify code, maybe it is not affected by this. I thought that the dialer asks the other party to identify themselves, but since this is a substream I guess either party can open an outbound substream.

You're welcome to try something else, but I think I'll deploy my fix at actyx until we have something that works and is upstream.

dvc94ch:
@mxinden I guess this PR can be merged, then we can continue discussion on how to hook it up on a different PR

mxinden:
> @mxinden I guess this PR can be merged, then we can continue discussion on how to hook it up on a different PR

What would be the benefit of having this logic in `master` without it being hooked up into `libp2p-core`?

dvc94ch:
Not much, but there isn't really any downside either. Making the diff between actyx libp2p and rust-libp2p smaller is an advantage for us. It's hard for us to estimate when and if a feature will be merged. We're releasing actyx 2.0.0 probably next week so ideally we would't rely on patching cargo.toml

mxinden:
Status update:

As mentioned above, while this pull request adds the core extension logic to `libp2p-multistream-select`, this pull request is still missing the glue code, allowing a dialing connection to be turned into a listening connection within `libp2p-core` or beyond.

I see multiple ways to do this, though I don't think any of them is ideal:

- Expose the simultaneous connection role to `OutboundUpgrade`. See https://github.com/libp2p/rust-libp2p/pull/2099/ as an example.
  ```diff
  - fn upgrade_outbound(self, socket: C, info: Self::Info) -> Self::Future;
  + fn upgrade_outbound(self, socket: C, info: Self::Info, role: SimOpenRole) -> Self::Future;
  ```
  
  While this does solve the issue itself, namely being able to turn an oubound upgrade into an inbound upgrade, it does so at the expense of (a) adding more complexity in general and (b) polluting the function signature for stream upgrades which will never need this functionality.

- Upgrade connections (not streams) based on the role returned by simultaneous connect. I.e. within `libp2p-core` choose to either upgrade a connection via `upgrade.upgrade.outbound` on `SimOpenRole::Initiator` or `upgrade.upgrade_inbound` on `SimOpenRole::Responder`.

  While this does require extensive changes within `libp2p-core`, it does not expose any of the complexity to users or `libp2p-swarm`.

I will need to put more thoughts into this. Alternative suggestions are very much appreciated.

In the long term I hope to solve this in a clean way within the ongoing  _Protocol Select_ efforts. More specifically I would like to find an abstraction that does not require a _hack_ for a single protocol peculiarity namely TCP's simultaneous open "feature".

mxinden:
> Not much, but there isn't really any downside either. Making the diff between actyx libp2p and rust-libp2p smaller is an advantage for us. It's hard for us to estimate when and if a feature will be merged. We're releasing actyx 2.0.0 probably next week so ideally we would't rely on patching cargo.toml

@dvc94ch would you mind expanding on why you need this feature? Are you seeing uncoordinated, as in not timed for hole punching, simultaneous TCP opens in the wild? If so, do you have metrics on how often this is happening in your environment?

dvc94ch:
we don't have metrics on how often this is happening. but it has happened and fixed it (even though arguably in a suboptimal way). and therefore isn't happening anymore. But once you account for roundtrip time, it is much more likely to happen in the real world than in a test environment. so if you have a path delay of 100ms there is quite a big time slot where this could happen.

I'm not sure why it depends on relay v2 or dcutr, I think this is an independent feature/issue.

dvc94ch:
Also did you see my comments about the relay in the ipfs-embed issue you commented on? Hope some were useful, but I think the relay v2 will take a while.

mxinden:
> But once you account for roundtrip time, it is much more likely to happen in the real world than in a test environment. so if you have a path delay of 100ms there is quite a big time slot where this could happen.

Can you expand on this? Yes, the longer the path delay the higher the chance that a simultaneous open happens. But even with a 100ms delay, I would judge the probability to be low for two nodes to accidentally dial each other in the same time window. I would need more data in order to justify the trade-off to merge a somewhat half-baked solution, unless we come up with a better design how to integrate it into rust-libp2p.

dvc94ch:
So we've released v2. I'm on vacation this week and we have team meetings next week. I'll see if I have some time to bake it a bit more after that.

rkuhn:
@mxinden Consider a scenario where gossipsub is used for peer discovery: node A sends its newest info to B and C, who are currently not connected since they didn’t know about each other. Following the received gossip B and C connect to each other, in which case we have observed it to be quite likely for a simultaneous open to occur.

mxinden:
Thanks @dvc94ch for offering additional help and thanks @rkuhn for the scenario.

https://github.com/libp2p/rust-libp2p/pull/2066/commits/2cfaec991746f190587b8735a9839e1b98bbc203 adds the missing integration into `libp2p-core`, thus making simultaneous open accessible to users. There is still lots to do (see `TODO`s), though to me this seems to be the right path. https://github.com/libp2p/rust-libp2p/pull/2066/commits/2cfaec991746f190587b8735a9839e1b98bbc203 does not touch `InboundUpgrade` nor `OutboundUpgrade`, thus not requiring Stream upgrades nor multiplexer upgrades to handle the simultaneous open edge case, which can only ever happen during security upgrades on a connection, not stream.

Feedback and help with the outstanding `TODO`s is welcome.

mxinden:
Despite the fact that https://github.com/libp2p/rust-libp2p/commit/2cfaec991746f190587b8735a9839e1b98bbc203 lays the ground-work needed to get this merged, it would be great to have more concrete data on how often simultaneous-open on TCP happens in the wild. This is especially relevant as we are currently designing [_Protocol Select_](https://github.com/protocol/web3-dev-team/pull/119) the successor to _Multistream Select_ where we might be able to optimize handling of simultaneous-open.

Any chance you could measure this on one of your nodes in your environment @dvc94ch @rkuhn?

rkuhn:
I think this depends greatly on the use-case: for uncorrelated connection attempts the probability might be very small or even negligible, while for correlated ones (as we produce) it is quite high. I don’t have exact numbers, but it is of the order of 10–20%. The main difficulty is to recognise correlations — in our case it is obvious, but in other cases it may be less obvious (e.g. an event in the network that leads many nodes to look up two certain cids, including the two nodes hosting the underlying blocks).

My opinion is that this scenario needs to be handled well in any case since there are uses that reliably trigger it and there are other uses where it may happen once in a blue moon — which is really difficult to debug when it occurs.

dvc94ch:
well, just tried this branch and it still doesn't work

dvc94ch:
looks like `apply_authentication` isn't called anywhere

mxinden:
> well, just tried this branch and it still doesn't work

You would need to be more specific than that.

> looks like apply_authentication isn't called anywhere

See `transport/upgrade.rs`:

https://github.com/libp2p/rust-libp2p/blob/9861a536981310a9e0f860b6fbab6bc981379eb6/core/src/transport/upgrade.rs#L132

dvc94ch:
ah looks like github collapsed some diffs, so my browser search didn't find it.  It isn't enabled by default, an `AuthenticationVersion::V1SimOpen` needs to be passed to `authenticate_with_version`. I guess that works for me, although possibly a little obscure to find for other users.

dvc94ch (approved):
I guess this will be in 0.39.1?

mxinden:
> I guess this will be in 0.39.1?

Given that his pull request introduces breaking changes, I can not release it as `v0.39.1`, though I plan to include it in the following minor (we are still below `v1.0`) release.

dvc94ch:
so can this be merged/released now that 0.39.1 is out?

rklaehn:
Any progress on this? At actyx we have had to fork rust-libp2p to make things work, and we would really like to be able to stop doing that...

mxinden:
> Any progress on this? At actyx we have had to fork rust-libp2p to make things work, and we would really like to be able to stop doing that...

Sorry for the delay / me not communicating well here and sorry for the trouble this is causing on your end.

I said earlier that I will include this pull request in `v0.40.0`, though as you can see I still haven't merged it. I am hesitating to do so because:

- It adds a lot of complexity to `libp2p-core`.
- (At least in theory) it is not necessary for hole punching (see https://github.com/libp2p/rust-libp2p/issues/2250 for details) and thus only required for uncoordinated TCP _Simultaneous Open_.
- We will replace all of _Multistream Select_ with a new protocol - _Protocol Select_ - in the future https://github.com/libp2p/specs/pull/349
- This is only relevant for TCP and not relevant for QUIC (https://github.com/libp2p/rust-libp2p/pull/2159).

I will test out https://github.com/libp2p/rust-libp2p/issues/2250 before making any decision here.

misc/multistream-select/src/protocol.rs
```diff
@@ -249,7 +303,7 @@ impl<R> MessageIO<R> {
         }
     }
 
-    /// Drops the [`MessageIO`] resource, yielding the underlying I/O stream.
+    /// Draops the [`MessageIO`] resource, yielding the underlying I/O stream.
```
mxinden:
```suggestion
    /// Drops the [`MessageIO`] resource, yielding the underlying I/O stream.
```

core/CHANGELOG.md
```diff
@@ -80,7 +80,43 @@
...
+         .authenticate(my_authentication)
+         .multiplex(my_multiplexer)
+       ```
+     - `Version::V1Lazy`
+
+       ```rust
+       my_transport.upgrade()
+         .authenticate_with_version(my_authentication, Version::V1Lazy)
+         .multiplex_with_version(my_multiplexer, Version::V1Lazy)
+       ```
```
wngr, line 110: I don't think it's worth adding the explicit `_with_version` methods. I'd just change the signatures of `authenticate` and `multiplex`. `Version` already provides a `Default` impl.

core/src/upgrade/apply.rs
```diff
@@ -243,14 +286,223 @@ where
...
+        upgrade: U,
+    },
+    Upgrade {
+        role: Role,
+        future: Either<
+            Pin<Box<<U as OutboundUpgrade<Negotiated<C>>>::Future>>,
+            Pin<Box<<U as InboundUpgrade<Negotiated<C>>>::Future>>,
+        >,
+    },
+    Undefined,
```
wngr, line 419: Why not make this an `Option` instead of the additional enum member?

misc/multistream-select/src/lib.rs
```diff
@@ -137,6 +137,18 @@ pub enum Version {
     /// [1]: https://github.com/multiformats/go-multistream/issues/20
     /// [2]: https://github.com/libp2p/rust-libp2p/pull/1212
     V1Lazy,
+    /// A variant of version 1 that selects a single initiator when both peers are acting as such,
+    /// in other words when both peers simultaneously open a connection.
+    ///
+    /// This multistream-select variant is specified in [1].
+    ///
+    /// Note: [`Version::V1SimultaneousOpen`] should only be used (a) on transports that allow
+    /// simultaneously opened connections, e.g. TCP with socket reuse and (2) during the first
```
wngr, line 146:
```suggestion
    /// simultaneously opened connections, e.g. TCP with socket reuse and (b) during the first
```

misc/multistream-select/src/lib.rs
```diff
@@ -137,6 +137,18 @@ pub enum Version {
...
+    /// negotiation on the connection, most likely the secure channel protocol negotiation. In all
```
wngr, line 147: WDYM with _first negotiation on the connection_?

misc/multistream-select/src/tests.rs
```diff
@@ -244,4 +242,37 @@ fn select_proto_serial() {
 
     async_std::task::block_on(run(Version::V1));
     async_std::task::block_on(run(Version::V1Lazy));
+    async_std::task::block_on(run(Version::V1SimultaneousOpen));
+}
+
+#[test]
+fn simultaneous_open() {
```
wngr, line 249: Would it make sense to add a "full stack test"?

wngr (commented):
Nice work! :100:

core/CHANGELOG.md
```diff
@@ -80,7 +80,43 @@
...
+       ```rust
+       my_transport.upgrade()
+         .authenticate(my_authentication)
+         .multiplex(my_multiplexer)
+       ```
+     - `Version::V1Lazy`
+
+       ```rust
+       my_transport.upgrade()
+         .authenticate_with_version(my_authentication, Version::V1Lazy)
```
thomaseizinger, line 108: This is a different type, isn't it?

```suggestion
         .authenticate_with_version(my_authentication, AuthenticationVersion::V1Lazy)
```

core/CHANGELOG.md
```diff
@@ -80,7 +80,43 @@
...
+
+     Users merely wanting to maintain the status quo should use the following call chain depending
+     on which `Version` they previously used:
+
+     - `Version::V1`
+
+       ```rust
+       my_transport.upgrade()
+         .authenticate(my_authentication)
+         .multiplex(my_multiplexer)
```
thomaseizinger, line 102: Have you considered adding a `.upgrade_sim_open()` API instead of changing the existing one? That would:

a) be backwards compatible
b) perhaps avoid the trickery around two `Version` enums?

misc/multistream-select/src/dialer_select.rs
```diff
@@ -141,10 +150,44 @@ pub struct DialerSelectSeq<R, I: Iterator> {
...
+        io: MessageIO<R>,
+        protocol: N,
+    },
+    SimOpenPhase {
+        selection: SimOpenPhase<R>,
+        protocol: N,
+    },
+    Responder {
+        responder: crate::ListenerSelectFuture<R, N>,
+    },
```
thomaseizinger, line 176: This feels a bit forced into here and could potentially be a bit more modularised but I think that would come at the cost of a lot of code duplication downstream because we would return different futures?

core/src/upgrade/apply.rs
```diff
@@ -74,7 +112,7 @@ where
         .protocol_info()
         .into_iter()
         .map(NameWrap as fn(_) -> NameWrap<_>);
-    let future = multistream_select::dialer_select_proto(conn, iter, v);
+    let future = multistream_select::dialer_select_proto(conn, iter, v.into());
```
thomaseizinger, line 115: I was considering suggesting to create a dedicated `dialer_sim_open_select_proto` function but I guess that idea is not particularly good because we rely on being able to name all these futures here?

misc/multistream-select/src/dialer_select.rs
```diff
@@ -268,6 +450,182 @@ where
...
+                        // that nonces need to be exchanged first, the attached
+                        // protocol by the remote needs to be ignored.
+                        Message::Protocol(_) => {
+                            self.state = SimOpenState::ReadNonce { io, local_nonce };
+                        }
+                        Message::Select(remote_nonce) => {
+                            match local_nonce.cmp(&remote_nonce) {
+                                Ordering::Equal => {
+                                    // Start over.
+                                    self.state = SimOpenState::SendNonce { io };
```
thomaseizinger, line 553: The [spec](
https://github.com/libp2p/specs/blob/master/connections/simopen.md#the-protocol) says we should fail in this case!

> In the unlikely case where both peers selected the same integer, connection establishment fails.

misc/multistream-select/src/protocol.rs
```diff
@@ -160,6 +178,22 @@ impl Message {
                 dest.put(MSG_PROTOCOL_NA);
                 Ok(())
             }
+            Message::Select(nonce) => {
+                dest.put(MSG_SELECT);
```
thomaseizinger, line 182: Any reason why we don't use `reserve` here?

misc/multistream-select/src/listener_select.rs
```diff
@@ -42,6 +42,37 @@ use std::{
...
+    I: IntoIterator,
+    I::Item: AsRef<[u8]>,
+{
+    listener_select_proto_with_state(State::RecvMessage { io }, protocols)
+}
+
+fn listener_select_proto_with_state<R, I>(
+    state: State<R, I::Item>,
+    protocols: I,
+) -> ListenerSelectFuture<R, I::Item>
```
thomaseizinger, line 75: I guess you are just following the style here but I always wondered why these aren't associated constructor functions on `ListenerSelectFuture`.

```rust
impl<R, N> ListenerSelectFuture<R, N> {
	pub fn new(inner: R, protocols: I) -> Self { }
	pub fn in_state(state: State<R, I::Item>, protocols: I) -> Self { } 
	pub fn assume_header_sent(io: MessageIO<R>, protocols: I) -> Self { }
}
```

misc/multistream-select/src/dialer_select.rs
```diff
@@ -268,6 +450,182 @@ where
...
+                        // stream as a permissible way to "gracefully" fail a negotiation.
+                        Poll::Ready(None) => return Poll::Ready(Err(NegotiationError::Failed)),
+                    };
+
+                    let result = match local_role {
+                        Role::Initiator if remote_msg == Message::Responder => Ok((io, local_role)),
+                        Role::Responder if remote_msg == Message::Initiator => Ok((io, local_role)),
+
+                        _ => Err(ProtocolError::InvalidMessage.into()),
+                    };
```
thomaseizinger, line 619: I find matching on a tuple easier to read in such circumstances.

```suggestion
                    let result = match (local_role, remote_msg) {
                        (Role::Initiator, Message::Responder) => Ok((io, local_role)),
                        (Role::Responder, Message::Initiator) => Ok((io, local_role)),
                        _ => Err(ProtocolError::InvalidMessage.into()),
                    };
```

misc/multistream-select/src/dialer_select.rs
```diff
@@ -268,6 +450,182 @@ where
...
+                                        io,
+                                        local_role: Role::Initiator,
+                                    };
+                                }
+                                Ordering::Less => {
+                                    self.state = SimOpenState::SendRole {
+                                        io,
+                                        local_role: Role::Responder,
+                                    };
+                                }
```
thomaseizinger, line 566: Could lift the assignment to `self.state` out of the `match`.

thomaseizinger: Together with actually failing on `Ordering::Equal`, you could write:

```rust
let local_role = match local_nonce.cmp(&remote_nonce) {
    Ordering::Equal => {
        return Poll::Ready(Err(...))
    }
    Ordering::Greater => Role::Initiator,
    Ordering::Less => Role::Responder
};

self.state = SimOpenState::SendRole { io, local_role };
```

misc/multistream-select/src/dialer_select.rs
```diff
@@ -268,6 +450,182 @@ where
...
+                    if let Err(err) = Pin::new(&mut io).start_send(msg) {
+                        return Poll::Ready(Err(From::from(err)));
+                    }
+
+                    self.state = SimOpenState::FlushNonce { io, local_nonce };
+                }
+                SimOpenState::FlushNonce {
+                    mut io,
+                    local_nonce,
+                } => match Pin::new(&mut io).poll_flush(cx)? {
```
thomaseizinger, line 510: Do you know about the [`_unpin` APIs](https://docs.rs/futures/0.3.18/futures/sink/trait.SinkExt.html#method.poll_flush_unpin) in `SinkExt`?

Maybe we can use them here to avoid a bit of boilerplate.

misc/multistream-select/src/dialer_select.rs
```diff
@@ -177,13 +220,141 @@ where
...
+                            return Poll::Pending;
+                        }
+                        // Treat EOF error as [`NegotiationError::Failed`], not as
+                        // [`NegotiationError::ProtocolError`], allowing dropping or closing an I/O
+                        // stream as a permissible way to "gracefully" fail a negotiation.
+                        Poll::Ready(None) => return Poll::Ready(Err(NegotiationError::Failed)),
+                    };
+
+                    match msg {
+                        Message::Header(v) if v == HeaderLine::from(*this.version) => {
```
thomaseizinger, line 297: This is equivalent I think? Given that `SimOpen` implies `V1`, I find it awkward to parse the version back into the headerline.

```suggestion
                        Message::Header(HeaderLine::V1) => {
```

misc/multistream-select/src/tests.rs
```diff
@@ -244,4 +242,37 @@ fn select_proto_serial() {
...
+            let protos = vec![b"/proto3", b"/proto2"];
+            let (proto, io, _) = dialer_select_proto_serial(connec, protos.into_iter(), version)
+                .await
+                .unwrap();
+            assert_eq!(proto, b"/proto2");
+            io.complete().await.unwrap();
+        };
+
+        futures::future::join(server, client).await;
+    }
```
thomaseizinger, line 275: This could just be an `async move` block inside `block_on` unless we want to add more tests?

thomaseizinger (commented):
I've left a few comments here and there. I guess my biggest question would be whether or not we need to "integrate" sim-open this tightly with the existing multistream-select implementation?

If we would offer a dedicated `upgrade_sim_open` function, we might get away with creating a dedicated `dialer_select_sim_open` that internally first does the sim-open dance and then continues with regular multistream-select (which we can already start in an arbitrary state).

However, I don't know how painful this would be in libp2p-core (i.e. would it introduce much duplication among the `Authenticated` etc structs?).

Side-note: `multistream-select` could really benefit from `async-await` but I guess as long as we want to name the return type, this is not going to happen :sweat_smile:

rkuhn:
Great to see movement here! I’m not married to this (or any other solution), but I’d like to get rid of our Actyx forks of libp2p, for which I’d need some way to express a fix to this bug (remember that sim-open does happen in our system, without hole punching, due to correlated peer discovery). I’m willing to help, what is the roadmap?

rkuhn:
FYI: I’m moving Actyx (and ipfs-embed) away from this patch. The alternative solution is to not use PortReuse::Enabled and instead manage remote addresses a bit more carefully — thankfully we have the Identify protocol which provides the needed information.

dvc94ch:
can you elaborate on that? you can never prevent two peers dialing simultaneously

rkuhn:
Without port reuse, these will be two completely separate connections, with no confusion over who is dialler and who is listener.

dvc94ch:
@mxinden: I guess this PR is abandoned? In that case it should probably be closed.

mxinden:
With https://github.com/libp2p/rust-libp2p/pull/2363 we won't need this pull request for hole punching.

I am still hesitant to support Multistream Select Sim Open in rust-libp2p. See reasoning in https://github.com/libp2p/rust-libp2p/pull/2066#issuecomment-937884928. With that in mind, I am closing here.

> @mxinden: I guess this PR is abandoned? In that case it should probably be closed.

Thanks for the ping @dvc94ch.

Sorry @wngr and @thomaseizinger to not address your PR feedback.

> Side-note: `multistream-select` could really benefit from `async-await` but I guess as long as we want to name the return type, this is not going to happen sweat_smile

Agreed. Related: I think we should make sure we can use `async-await` before starting [Protocol Select](https://github.com/libp2p/specs/pull/349).

PR #2067, opened 2021-05-06
misc/multistream-select: Cleanup trait bounds and remove unused file

mxinden:
Small cleanups found while implementing https://github.com/libp2p/rust-libp2p/pull/2066 namely:

- Remove unused file `error.rs`.
- Remove unnecessary trait bounds on structs.


misc/multistream-select/src/dialer_select.rs
```diff
@@ -128,23 +128,14 @@ where
 /// A `Future` returned by [`dialer_select_proto_serial`] which negotiates
 /// a protocol iteratively by considering one protocol after the other.
 #[pin_project::pin_project]
-pub struct DialerSelectSeq<R, I>
-where
-    R: AsyncRead + AsyncWrite,
-    I: Iterator,
-    I::Item: AsRef<[u8]>
-{
+pub struct DialerSelectSeq<R, I: Iterator> {
```
thomaseizinger, line 131: Did a quick exploration whether we could get rid of this trait-bound as well but I don't think it is worth it. Not if we want to keep exporting a named type from `multistream-select` and I believe switching to `impl Trait` for `dialer_select_proto` makes the situation in `rust-libp2p` actually worse because we'd have more type parameters bubbling up (or would need to start boxing things once we are in `libp2p-swarm`).

thomaseizinger (approved):

mxinden:
@vmx do you have a couple of minutes to give this pull request a review?

vmx (approved):

PR #2075, opened 2021-05-20
*: Use upstream multiaddr crate

mxinden:
> The history of the local multiaddr is that we just cloned the existing code in this repo for the sake of moving quickly.
>
> Now that things have stabilized, we should backport changes to the upstream repository.

Fixes https://github.com/libp2p/rust-libp2p/issues/758. Blocked on https://github.com/multiformats/rust-multiaddr/pull/40.


mxinden:
@vmx do you have some spare cycles to give this a review?

vmx (approved):
I've only checked the changes lines, so if there e.g. was a workspace internal package that wasn't updated properly i wouldn't catch it (though I guess the CI would).

mxinden:
> I've only checked the changes lines, so if there e.g. was a workspace internal package that wasn't updated properly i wouldn't catch it (though I guess the CI would).

:+1: thanks Volker!

PR #2079, opened 2021-05-26
[Gossipsub] Minor lint and spelling fixes

AgeManning:
This just contains a few lint fixes and some spelling mistakes which leak via logs. 

I was doing this updates locally, but figured should push upstream.


mxinden (approved):
Thanks Adrian!

PR #2080, opened 2021-05-26
protocols/relay: Simplify running the relay example

r-zig:
Add ability to run the relay example
with all peers run together via docker compose


protocols/relay/examples/relay.rs
```diff
@@ -82,3 +154,149 @@ fn main() -> Result<(), Box<dyn Error>> {
...
+#[derive(Debug, StructOpt)]
+#[structopt(name = "libp2p relay")]
+struct Opt {
+    /// The mode (relay, client-listen, client-dial)
+    #[structopt(long)]
+    mode: Mode,
+
+    /// Fixed value to generate deterministic peer id
+    #[structopt(short = "d")]
+    deterministic_peer_secret: u8,
```
mxinden, line 297:
```suggestion
    secret_key_seed: u8,
```

The word `seed` would be more intuitive to me here.

protocols/relay/examples/relay.rs
```diff
@@ -18,23 +18,70 @@
 // FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
 // DEALINGS IN THE SOFTWARE.
 
+//! A basic relay server and relay client implementation.
```
mxinden, line 21: Once you resolved the merge conflicts, these additions should be gone.

protocols/relay/examples/docker-compose.yml
```diff
@@ -0,0 +1,56 @@
...
+      - deterministic_peer_env=2
+      # - address_env=/ip4/172.16.238.10/tcp/4444/p2p/12D3KooWPjceQrSwdWXPyLLeABRXmuqt69Rg3sBYbU1Nft9HyQ6X/p2p-circuit
+      - address_env=/ip6/2001:3984:3989::10/tcp/4444/p2p/12D3KooWPjceQrSwdWXPyLLeABRXmuqt69Rg3sBYbU1Nft9HyQ6X/p2p-circuit
+    build:
+      context: ../../../.
+      dockerfile: ./protocols/relay/examples/Dockerfile
+    networks:
+      default:
+        ipv4_address: 172.16.238.11
+        ipv6_address: 2001:3984:3989::11
```
mxinden, line 31:
```suggestion
      - address_env=/dns6/relay/tcp/4444/p2p/12D3KooWPjceQrSwdWXPyLLeABRXmuqt69Rg3sBYbU1Nft9HyQ6X/p2p-circuit
    build:
      context: ../../../.
      dockerfile: ./protocols/relay/examples/Dockerfile
    networks:
      - default
```

Why not leverage DNS?

protocols/relay/examples/relay.rs
```diff
@@ -82,3 +154,149 @@ fn main() -> Result<(), Box<dyn Error>> {
...
+    }
+}
+
+/// Placeholder to publish the peer address based on given key
+fn set_shared_data(key: &str, value: &str) -> () {
+    println!(
+        "TODO publish to KAD or discovery service ... Set the key: {} data to:{}",
+        key, value
+    );
+}
```
mxinden, line 224:
```suggestion
```

I would prefer rust-libp2p not to contain any placeholder code. Could you remove everything around some external discovery mechanism?

protocols/relay/Cargo.toml
```diff
@@ -30,6 +30,8 @@ prost-build = "0.7"
 
 [dev-dependencies]
 env_logger = "0.8.3"
+hex = "0.4.2"
```
mxinden, line 33: Where is this used?

.dockerignore
```diff
@@ -0,0 +1,29 @@
+**/.classpath
```
mxinden, line 1: I would prefer to keep the repository reasonably clean. Is this file necessary? As far as I can tell, without the file the build context would just be a bit bigger.

protocols/relay/examples/Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+RUN apt-get update && \
+    apt-get dist-upgrade -y && \
+    apt-get install -y musl-tools && \
+    rustup target add x86_64-unknown-linux-musl
+# 1b: Download and compile Rust dependencies (and store as a separate Docker layer)
+RUN USER=root cargo new rust-libp2p
+WORKDIR /usr/src/libp2p
+
+COPY . /usr/src/libp2p
+RUN cargo build --target x86_64-unknown-linux-musl --example=relay --package=libp2p-relay
```
mxinden, line 14: Why is the copying required here? Afterall this is the build stage, right?

mxinden (commented):
Thanks for the work!

protocols/relay/Cargo.toml
r-zig: You right, I forgot this dependency after stop to use this crate.

mxinden:
Let me know @r-zig once this is ready for another review.

protocols/relay/examples/docker-compose.yml
r-zig: I didn't succeed to do that.
When specifying it in the relay mode - address_env=/dns6/relay/tcp/4444 
I received 
Error: MultiaddrNotSupported("/dns6/relay/tcp/4444")
When specify the relay as before with:
- address_env=/ip6/::/tcp/4444
The other peers weren't able to connect to it via dns6/relay/tcp/4444 address

protocols/relay/examples/Dockerfile
r-zig: Without this copy the cargo build will not work. It should refer the entire code base of libp2p.
(This is not the COPY of the build artifacts)
Can you provide other suggestion?

protocols/relay/examples/docker-compose.yml
mxinden: > When specifying it in the relay mode - address_env=/dns6/relay/tcp/4444 I received Error: MultiaddrNotSupported("/dns6/relay/tcp/4444")

Sorry, I forgot to mention that you will need to wrap the TCP transport with the DNS transport in order to resolve Multiaddresses with DNS addresses. The following diff should resolve your issues:


```diff
diff --git a/protocols/relay/examples/relay.rs b/protocols/relay/examples/relay.rs
index 79a3762d..a28c0d3b 100644
--- a/protocols/relay/examples/relay.rs
+++ b/protocols/relay/examples/relay.rs
@@ -54,6 +54,7 @@
 use futures::executor::block_on;
 use futures::stream::StreamExt;
 use libp2p::core::upgrade;
+use libp2p::dns::DnsConfig;
 use libp2p::ping::{Ping, PingConfig, PingEvent};
 use libp2p::plaintext;
 use libp2p::relay::{Relay, RelayConfig};
@@ -72,14 +73,14 @@ fn main() -> Result<(), Box<dyn Error>> {
     let local_peer_id = PeerId::from(local_key.public());
     println!("Local peer id: {:?}", local_peer_id);
 
-    let tcp_transport = TcpConfig::new();
+    let transport = block_on(DnsConfig::system(TcpConfig::new()))?;
 
     let relay_config = RelayConfig {
         connection_idle_timeout: Duration::from_secs(10 * 60),
         ..Default::default()
     };
     let (relay_wrapped_transport, relay_behaviour) =
-        libp2p_relay::new_transport_and_behaviour(relay_config, tcp_transport);
+        libp2p_relay::new_transport_and_behaviour(relay_config, transport);
 
     let behaviour = Behaviour {
         relay: relay_behaviour,
```

protocols/relay/examples/Dockerfile
```diff
@@ -0,0 +1,22 @@
+# 1: Build the exe
+FROM rust:latest as builder
```
mxinden, line 2:
```suggestion
FROM rust:1 as builder
```

To make sure that docker does not upgrade to a breaking change automatically.

protocols/relay/examples/Dockerfile
mxinden: As far as I can tell the `RUN` step is not needed:

```Dockerfile
# 1: Build the exe
FROM rust:latest as builder
LABEL Name=libp2p-relay Version=0.0.1
# 1a: Prepare for static linking
RUN apt-get update && \
    apt-get dist-upgrade -y && \
    apt-get install -y musl-tools && \
    rustup target add x86_64-unknown-linux-musl
# 1b: Download and compile Rust dependencies (and store as a separate Docker layer)
WORKDIR /usr/src/libp2p

COPY . .
RUN cargo build --target x86_64-unknown-linux-musl --example=relay --package=libp2p-relay

# 2: Copy the exe and extra files ("static") to an empty Docker image
FROM debian:buster-slim

COPY --from=builder /usr/src/libp2p/target/x86_64-unknown-linux-musl/debug/examples/relay .
USER 1000

ENTRYPOINT ./relay -d $deterministic_peer_env --mode $relay_command_arg_env --address $address_env

```

protocols/relay/examples/docker-compose.yml
r-zig: Your solution work. I added commit

protocols/relay/examples/Dockerfile
r-zig: Ok , but why you choose version 1 and not the current available latest version? (1.52.1)

mxinden: The Rust project follows [semantic versioning](https://semver.org/). In this particular case this means that e.g. an image with `v1.52.1` does not introduce any breaking changes compared to an image with `v1.52.0` or `v1.51.0`. As far as I am aware the `1` tag is overwritten each time a new image in the `v1.x.x` series is released. Thus pointing to `rust:1` instead of `rust:latest` will always give us the latest image from the `v1.x.x` series, but will not automatically update to the `v2.x.x` series given that the `v2.x.x` series might include breaking changes.

@r-zig does that make sense?

protocols/relay/examples/docker-compose.yml
```diff
@@ -0,0 +1,53 @@
+version: '3.4'
+services:
+  relay: 
+    image: libp2p-relay
+    environment:
+      - relay_command_arg_env=relay
+      - deterministic_peer_env=1
+      - address_env=/ip6/::/tcp/4444
```
mxinden, line 8:
```suggestion
    command:
      - "./relay"
      - "--mode=relay"
      - "-d=1"
      - "--address=/ip6/::/tcp/4444"
```


Why this additional indirection through environment variables? Why not pass these as commandline arguments to the binary directly?

protocols/relay/examples/docker-compose.yml
```diff
@@ -0,0 +1,53 @@
...
+  default:
+    driver: bridge
+    enable_ipv6: true
+    ipam:
+      driver: default
+      config:
+        - subnet: 172.16.238.0/24
+          gateway: 172.16.238.1
+        - subnet: 2001:3984:3989::/64
+          gateway: 2001:3984:3989::1
```
mxinden, line 53:
```suggestion
  - network-a
  - network-b
```

Are these configuration options needed? In addition, as you suggested via mail, how about placing the `client-dial` and `client-listen` into different networks and the `relay` into both?

protocols/relay/examples/relay.rs
```diff
@@ -173,3 +232,47 @@ impl From<()> for Event {
...
+
+#[derive(Debug, StructOpt)]
+#[structopt(name = "libp2p relay")]
+struct Opt {
+    /// The mode (relay, client-listen, client-dial)
+    #[structopt(long)]
+    mode: Mode,
+
+    /// Fixed value to generate deterministic peer id
+    #[structopt(short = "d")]
```
mxinden, line 272:
```suggestion
    #[structopt(long)]
```

I would suggest using the long command line flag here to avoid confusion. I am not sure how `d` would remind people of `secret_key_seed`.

protocols/relay/examples/relay.rs
```diff
@@ -24,62 +24,74 @@
 //! relay client listening via the relay server and (3) a dialing relay client
 //! dialing the listening relay client via the relay server.
 //!
-//! 1. To start the relay server, run `cargo run --example relay -- relay` which will print
-//!    something along the lines of:
+//! 1. To start the relay server, run `cargo run --example=relay --package=libp2p-relay -- -d 1 --mode relay --address /ip4/<ip address>/tcp/<port>`.
+//!    The -d help create static peer id using the given number argument
```
mxinden, line 28:
```suggestion
//!    The `-d` helps create a static peer id using the given number argument as  a seed.
```

protocols/relay/examples/relay.rs
```diff
@@ -24,62 +24,74 @@
...
+//!    The mode specify that the program run as a relay server
```
mxinden, line 29:
```suggestion
//!    The mode specifies whether the node should run as a relay server, a listening client or a dialing client.
```

protocols/relay/examples/relay.rs
```diff
@@ -24,62 +24,74 @@
...
+//!    The address specify a static address usually it will be some loop back address such as /ip4/0.0.0.0/tcp/4444
```
mxinden, line 30:
```suggestion
//!    The address specifies a static address. Usually it will be some loop back address such as `/ip4/0.0.0.0/tcp/4444`.
```

protocols/relay/examples/docker-compose.yml
r-zig: Without the ipam configuration the docker compose failed with
could not find an available, non-overlapping IPv6 address pool among the defaults to assign to the network.

protocols/relay/examples/docker-compose.yml
r-zig: Agree , please see the next change

protocols/relay/examples/relay.rs
r-zig: Fixed

protocols/relay/examples/relay.rs
r-zig: Fixed

protocols/relay/examples/relay.rs
r-zig: Fixed

protocols/relay/examples/relay.rs
r-zig: Fixed

protocols/relay/examples/docker-compose.yml
mxinden: How does this setup differ to the one below? Asked differently, why can the setup below work without these configuration options?

https://github.com/paritytech/polkadot/blame/1b82dc5b56e3970a0cfddaad17bbc7c487d5dbba/docker/sentry-docker-compose.yml#L101-L103

r-zig: The difference is the ipv6 vs ipv4.
When using ipv6 an IPv6 subnet must be defined.
See:
https://medium.com/@skleeschulte/how-to-enable-ipv6-for-docker-containers-on-ubuntu-18-04-c68394a219a2
I suggest to separate between those 2 options using docker-compose. ipv4.yml and docker-compose. ipv6.yml

mxinden: Tested this out locally. You are right. Sorry for the trouble! Two things:

- `enable_ipv6` does not seem to be available in docker-compose version `>=3.0` https://github.com/docker/docker.github.io/issues/3846. Unless I am missing something, would you mind downgrading the version to `2.1`?
    ```diff
    diff --git a/protocols/relay/examples/docker-compose.ipv6.yml b/protocols/relay/examples/docker-compose.ipv6.yml
    index 2d2b3c25..daf579bc 100644
    --- a/protocols/relay/examples/docker-compose.ipv6.yml
    +++ b/protocols/relay/examples/docker-compose.ipv6.yml
    @@ -1,7 +1,7 @@
    # This docker compose use the same example as the docker-compose.yml but with ipv6 instead of ipv4
    -version: '3.4'
    +version: '2.1'
    ```
- For the sake of simplicity, I would suggest removing the ipv4 docker-compose file.

mxinden (approved):
Other than https://github.com/libp2p/rust-libp2p/pull/2080#discussion_r659820692 this looks good to me. Thanks @r-zig for all the work.

protocols/relay/examples/docker-compose.yml
```diff
@@ -0,0 +1,62 @@
+# This docker compose use the same example as the docker-compose.yml but with ipv6 instead of ipv4
```
mxinden, line 1:
```suggestion
# Run `docker-compose up` to start the setup.

```

Issue #297, opened 2018-07-12
Implement the rendezvous protocol


tomaka:

thomaseizinger:
We are working on an implementation of the rendezvous protocol and plan on upstreaming it to here.

FelipeRosa:
I'm interested in using this in a personal project. Is there any tracking issues I could get a more detailed view on the current progress and, if needed, also help?

mxinden:
> Is there any tracking issues I could get a more detailed view on the current progress

@FelipeRosa see https://github.com/libp2p/rust-libp2p/pull/2107.

> and, if needed, also help?

@thomaseizinger need some extra hands?

thomaseizinger:
> @thomaseizinger need some extra hands?

Yes, could use some help actually (@FelipeRosa):

1. Feedback on the API would be great. We are still iterating on it while we are building https://github.com/comit-network/rendezvous-server and integrating it here https://github.com/comit-network/xmr-btc-swap/pull/593. If you have any feedback, feel free to comment on #2107.
2. The implementation also needs docs. Feel free to open PRs against the [`rendezvous`](https://github.com/comit-network/rust-libp2p/tree/rendezvous) branch of our fork.
3. Some of TODOs in the PR are also still open. If you want to tackle any of those, just ping me there!

FelipeRosa:
Cool! I'll see where I can help once I start working on this in my project.

Thanks for pointing directions @thomaseizinger.

mxinden:
Great work. And thanks for the reminder @thomaseizinger!

(rust-libp2p is the first libp2p implementation with a merged rendezvous implementation https://github.com/libp2p/website/pull/123/files.)

PR #2081, opened 2021-05-27
Signed Peer Records

thomaseizinger:
This is a very rough draft of implementing RFC0002 (signed envelopes) and RFC0003 (peer records).

It uses a quite different API from https://github.com/libp2p/rust-libp2p/pull/1803 which is why I am opening this in such an early stage to get feedback on which one is more likely to be accepted.

We are currently working on an implementation of the rendezvous protocol (https://github.com/libp2p/rust-libp2p/issues/297) which requires signed peer records.


mxinden:
@thomaseizinger let me know once this is ready for a review.

Pinging @blacktemplar. Maybe the two of you can collaborate on this effort.

thomaseizinger:
> @thomaseizinger let me know once this is ready for a review.
> 
> Pinging @blacktemplar. Maybe the two of you can collaborate on this effort.

Can you give it an initial pass that focusses on public APIs and overall modularisation / code layout?

The internals are dirty atm but that is easy to clean up once we settle on the APIs.

We are also currently dog-fooding this in our implementation of the rendezvous protocol but it would still be good to get your initial view in. For example, are you happy with those things in -core or would you want another crate?

AgeManning:
Just chiming in on @blacktemplar's behalf. The original version of this was written when the specs were not solidified and we potentially wanted to include them into the gossipsub protocol. 

We're not using them at the moment, and the old version in #1803 is likely obsolete. @blacktemplar is no longer working on this, but feel free to ping me if you need any input on this or discussion about the previous PR.

core/src/peer_record.rs
```diff
@@ -0,0 +1,254 @@
...
+            addresses: record
+                .addresses
+                .into_iter()
+                .map(|a| a.multiaddr.try_into())
+                .collect::<Result<Vec<_>, _>>()?,
+        })
+    }
+
+    // TODO: docs
+    pub fn warp_in_envelope(self, key: Keypair) -> SignedEnvelope {
```
mxinden, line 59: As a first thought I would prefer passing the data to some component owning the keypair instead of passing the keypair to the data. The former would reduce the exposure of the keypair minimizing the possibilities for mistakes / attacks.

That said, the notion of peer records is still in an early phase, thus any abstractions that we come up with today are likely premature.

mxinden (commented):
> We are also currently dog-fooding this in our implementation of the rendezvous protocol but it would still be good to get your initial view in. 

Perfect. Looking forward to seeing this in _action_.

Thank you for tackling this @thomaseizinger.

> For example, are you happy with those things in -core or would you want another crate?

I expect and hope for the concept of peer records to be widely adopted across the libp2p space across the various protocols. With that in mind, including it in `libp2p-core` makes sense to me.

core/src/peer_record.rs
```diff
@@ -0,0 +1,254 @@
...
+        SignedEnvelope::new(
+            key,
+            String::from(DOMAIN_SEP),
+            PAYLOAD_TYPE.as_bytes().to_vec(),
+            payload,
+        )
+        .unwrap() // TODO: Error handling
+    }
+
+    pub fn authenticate(self, key: Keypair) -> AuthenticatedPeerRecord {
```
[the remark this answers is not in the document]
thomaseizinger: Yes, agreed that the indirections are not good. I wasn't quite sure what the most natural and expressive way of constructing one of these would be.

Will see if we can make this work using From/TryFrom for the internals. I'd like to keep domain-specific functions for the actual public API because I consider them more expressive.

core/src/peer_record.rs
thomaseizinger: Interesting idea but I don't quite see how it would work. Your identity keypair is constructed as the very first thing and usually only used during the construction of `Network`.

The way I see this being used is that the user just clones the KeyPair at that stage and also passes it to the Behaviour (like Rendezvous) for example. Such a behaviour can then further use it to construct `SignedEnvelopes` or other things.

core/src/peer_record.rs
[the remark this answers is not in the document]
thomaseizinger: I removed `PeerRecord::authenticate`.

I think `PeerRecord::wrap_in_envelope` should stay as the information about the domain-separator should be tied to the `PeerRecord` because a `SignedEnvelope` is more abstract than a `PeerRecord`.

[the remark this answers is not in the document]
thomaseizinger: What might be debatable is whether or not we even need the concept of a normal `PeerRecord`. There is probably no point in having an "unauthenticated" peer record.

thomaseizinger:
@mxinden I incorporated some more learnings and feedback. Let me know what you think!

core/src/envelope.proto
```diff
@@ -0,0 +1,12 @@
+syntax = "proto2";
```
mxinden, line 1: Could you replace this file with https://github.com/libp2p/go-libp2p-core/blob/master/record/pb/envelope.proto ?

core/src/peer_record.proto
```diff
@@ -0,0 +1,22 @@
+syntax = "proto2";
```
mxinden, line 1: Could you replace this file with https://github.com/libp2p/go-libp2p-core/blob/master/peer/pb/peer_record.proto ?

mxinden (commented):
Looks good from a high level. Let me know once this is ready for a proper review.

Also, if you don't mind, I would suggest holding off merging here until we see the code in action, e.g. within the Rendezvous pull request.

>  Merge PeerRecord into AuthenticatedPeerRecord for simplicity 

:+1:

thomaseizinger:
> Also, if you don't mind, I would suggest holding off merging here until we see the code in action, e.g. within the Rendezvous pull request.

Sure, that makes total sense.
I only opened this a separate PR because it is functionally independent. If you want, we can also merge it together with the rendezvous protocol in the same PR?

mxinden:
> > Also, if you don't mind, I would suggest holding off merging here until we see the code in action, e.g. within the Rendezvous pull request.
> 
> Sure, that makes total sense.
> I only opened this a separate PR because it is functionally independent. If you want, we can also merge it together with the rendezvous protocol in the same PR?

I prefer separate PRs as they are "functionally independent", but merging them back-to-back.

thomaseizinger:
Included in #2107.

PR #2085, opened 2021-05-28
transports/websocket: Handle websocket CLOSE with reason code

dvdplm:
Pass websocket CLOSE reason to the app as an `Incoming::Closed(soketto::CloseReason)`.

This PR is a companion to https://github.com/paritytech/soketto/pull/31 and lets applications know what the remote end claimed to be the reason for closing the connection (if any).

`IncomingData` is now renamed to `Incoming` and the actual data (text or binary) is moved into its own `Data` enum, which in turn allows `into_bytes()` and the `AsRef` impl to apply more sanely to something that is actually data.


transports/websocket/src/framed.rs
```diff
@@ -442,55 +442,68 @@ fn location_to_multiaddr<T>(location: &str) -> Result<Multiaddr, Error<T>> {
...
-    /// UTF-8 encoded application data.
-    Text(Vec<u8>),
+pub enum Incoming {
+    /// Application data.
+    Data(Data),
     /// PONG control frame data.
-    Pong(Vec<u8>)
+    Pong(Vec<u8>),
+    /// Close reason.
+    Closed(CloseReason),
```
mxinden, line 458: To me, it is still a bit confusing why this is treated as an `Event` and thus returned via `Incoming` instead of being treated as an error and thus be returned via `Error::Closed`.

I am not familiar enough with soketto, thus I don't think this should block the effort.

Background for other readers: https://github.com/paritytech/soketto/pull/29#pullrequestreview-633290924.

transports/websocket/src/framed.rs
```diff
@@ -442,55 +442,68 @@ fn location_to_multiaddr<T>(location: &str) -> Result<Multiaddr, Error<T>> {
...
 
-/// Data received over the websocket connection.
+/// Data or control information received over the websocket connection.
 #[derive(Debug, Clone)]
-pub enum IncomingData {
-    /// Binary application data.
-    Binary(Vec<u8>),
-    /// UTF-8 encoded application data.
-    Text(Vec<u8>),
+pub enum Incoming {
```
mxinden, line 452: :+1:

transports/websocket/src/framed.rs
```diff
@@ -442,55 +442,68 @@ fn location_to_multiaddr<T>(location: &str) -> Result<Multiaddr, Error<T>> {
...
-    Pong(Vec<u8>)
+    Pong(Vec<u8>),
+    /// Close reason.
+    Closed(CloseReason),
+}
+
+/// Application data received over the websocket connection
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub enum Data {
+    /// Textual data.
```
mxinden, line 464:
```suggestion
    /// UTF-8 encoded textual data.
```

mxinden (approved):
Looks good to me overall.

> I have yet to test out how this works in application code (read: substrate), hence the draft status.

Let me know how this goes.

transports/websocket/src/framed.rs
dvdplm: It's not clear cut what the semantics should be. I went for the `Error::Closed` approach at first too, but @tomaka [pointed out](https://github.com/paritytech/soketto/pull/29#pullrequestreview-633290924) that the closing of the connection does carry information (a code in the simple case, but the specs allow for essentially any amount of "extra data").
The use case I have in mind for this is "collaborative backoff" where friendly peers can use the close reason to ask for respite, e.g. substrate telemetry asking nodes to send less data or go away for a while. In that context I think making the close reason an `Error` might be unfortunate.

mxinden:
Friendly ping @dvdplm.

dvdplm:
@mxinden Any chance of this getting in for v0.40?

transports/websocket/src/framed.rs
tomaka: The general consensus is that whenever a socket/connection produces an error, it means that the connection is no longer in a usable state. This isn't the case after receiving a WebSocket `Closed` frame. Receiving a closed frame is a normal situation, and the connection is still in a normal state and can be used to send more data or a `Closed` frame in return.

~~One drawback is: if `Closed` was instead an error, we would have a compile-time guarantee to not receive any other message after this closing. This guarantee can no longer be enforced by the compiler, but I think that this is fine.~~
EDIT: actually no, because the stream could continue to produce non-errors after having produced an error. We're not losing any compile-time guarantee.

tomaka (approved):
Adding `Closed` to the normal `Incoming` enum is what makes sense to me

mxinden:
> @mxinden Any chance of this getting in for v0.40?

Not into `libp2p` `v0.40.0` since I already published a release candidate, which in the ideal case, will just be promoted to a real release, or amended with bug / security fixes and then followed by another release candidate. Including a new feature in a release candidate would technically force us to test all over again.

That said, I am happy to release `libp2p-websocket` `v0.32.0` once `v0.31.0-rc.1` is promoted to `v0.31.0`.

---

@dvdplm would you mind including a changelog section for `v0.32.0` and an entry for this change?

kpp (approved):

dvdplm:
> @dvdplm would you mind including a changelog section for `v0.32.0` and an entry for this change?

Happy to do that. In this PR or separately?

mxinden:
This PR please.

mxinden:
Merged via https://github.com/libp2p/rust-libp2p/pull/2319.

PR #2087, opened 2021-05-31
protocols/kad: Extend emitted events with basic information

mxinden:
In order for users of `libp2p-kad` to gain deeper insights into the current state of the local DHT this pull request extends the events emitted by `libp2p-kad` with the information below. This is e.g. useful for https://github.com/libp2p/rust-libp2p/pull/2063.

- protocols/kad: Expose whether routing update is new peer

  Can e.g. be used to count the number of peers in the routing table.

- protocols/kad: Expose inbound request information

  Can e.g. be used to count the arrival rate of new records to be stored in the record store.

- protocols/kad: Expose kbucket range on RoutingUpdated

  Can e.g. be used to construct a graph of the current kbucket routing table state.


mxinden:
In combination with https://github.com/libp2p/rust-libp2p/pull/2063 this can be used to generate the following Grafana graphs.

image

More to come.

mxinden:
Marked as ready for review. Anyone out there got some time for a review? Maybe @thomaseizinger?

protocols/kad/src/behaviour.rs
```diff
@@ -2054,8 +2109,17 @@ pub struct PeerRecord {
...
 #[derive(Debug)]
 pub enum KademliaEvent {
-    /// A query has produced a result.
-    QueryResult {
+    /// An inbound request has been received and handled.
+    //
+    // Note on the difference between 'request' and 'query': A request is a
+    // single request-response style exchange with a single remote peer. A query
+    // is made of multiple requests across multiple remote peers.
+    InboundRequest {
```
thomaseizinger, line 2117: The name of this variant is not very "eventy" because it doesn't contain a verb and hence doesn't really express, what happened.

What about `InboundRequestServed`? That would also include the information you've packed into the docs.

protocols/kad/src/behaviour.rs
```diff
@@ -2054,8 +2109,17 @@ pub struct PeerRecord {
...
+        request: InboundRequest,
+    },
+
+    /// An outbound query has produced a result.
+    OutboundQueryResult {
```
thomaseizinger, line 2122: This one could be named `OutboundQueryCompleted` or `OutboundQueryFinished` in the same spirit.

thomaseizinger (approved):
Can't speak much on the functionality change but got two suggestions for improved naming. Hope it helps :)

mxinden:
>  but got two suggestions for improved naming

Changed.

Thanks for the input!

PR #2090, opened 2021-06-02
core/src/identity: Implement Keypair::from_protobuf_encoding for ed25519

mxinden:
Implement `Keypair` decoding from Protobuf according to [peer id
specification]. For now support ed25519 keys only. Future commits might
add RSA, secp256k1 and ecdsa.

[peer id specification]: https://github.com/libp2p/specs/blob/master/peer-ids/peer-ids.md#keys


mxinden:
@marten-seemann do you have some time to give this pull request a review, especially from the libp2p cryptography perspective?

@vmx  do you have some time to give this pull request a review, especially from the Rust perspective?

marten-seemann (commented):
I'm not sure if I can do a meaningful review here, considering my (lack of) Rust skills.

Why are you only implementing ed25519 though?

mxinden:
> Why are you only implementing ed25519 though?

Short story: Laziness.

Long story:

I have not found good support for parsing RSA private keys via PKCS1. Today we only support PKCS8 in rust-libp2p.

I have no easy way to test Secp256k1 parsing.

> Implementations MUST support Ed25519. Implementations SHOULD support RSA if they wish to interoperate with the mainline IPFS DHT and the default IPFS bootstrap nodes. Implementations MAY support Secp256k1 and ECDSA, but nodes using those keys may not be able to connect to all other nodes.

https://github.com/libp2p/specs/blob/master/peer-ids/peer-ids.md

Given that the libp2p ecosystem mostly bets on Ed25519 I am reluctant to put in the work to implement support for RSA, which might never be used, and reluctant to implement support for Secp256k1 which might be incorrect.

Note, rust-libp2p does fully support RSA and Secp256k1 public keys, thus is able to communicate with nodes using RSA and Secp256k1 private keys.

@marten-seemann does the above make sense? Am I missing something?

marten-seemann:
> Implementations SHOULD support RSA if they wish to interoperate with the mainline IPFS DHT and the default IPFS bootstrap nodes

How does Rust bootstrap if it doesn't support RSA? Do we have ed25519 bootstrappers now? If so, should we update the docs?

mxinden:
> How does Rust bootstrap if it doesn't support RSA?

rust-libp2p supports decoding RSA **public** keys from the protobuf definition defined in the [peer-id](https://github.com/libp2p/specs/blob/master/peer-ids/peer-ids.md) specification. Thus it can communicate with RSA based nodes. Up until now rust-libp2p did not support decoding **private** keys from the protobuf definition defined in the [peer-id](https://github.com/libp2p/specs/blob/master/peer-ids/peer-ids.md) specification. In order to load your private key you would have to pass the private key directly, e.g. PKCS8 encoded for an RSA private key and `[private key bytes][public key bytes]` "encoded" for an ed25519 private key.

This pull request adds support for decoding ed25519 private keys from the protobuf definition defined in the [peer-id](https://github.com/libp2p/specs/blob/master/peer-ids/peer-ids.md) specification. The motivation for this patch is to be able to load a ed25519 private key from an IPFS config files.

Sorry for the confusion here. Does the above make sense @marten-seemann?

> Do we have ed25519 bootstrappers now? If so, should we update the docs?

I don't know.

marten-seemann (approved):
Thank you for the detailed explanation. That makes a lot of sense to me.

core/src/identity.rs
```diff
@@ -115,6 +115,37 @@ impl Keypair {
...
+
+        let key_type = keys_proto::KeyType::from_i32(private_key.r#type)
+            .ok_or_else(|| DecodingError::new(format!("unknown key type: {}", private_key.r#type)))?;
+
+        match key_type {
+            keys_proto::KeyType::Ed25519 => {
+                ed25519::Keypair::decode(&mut private_key.data).map(Keypair::Ed25519)
+            },
+            keys_proto::KeyType::Rsa => {
+                Err(DecodingError::new("Unsupported"))
```
vmx, line 135: Would it make sense to add a more elaborate error message like "RSA keys are unsupported"? (same below).

vmx (approved):
I complement @marten-seemann's review with lack of libp2p knowledge, but Rust knowlege :)

Just a minor comment, rest looks good.

core/src/identity.rs
mxinden: Good idea. Added via https://github.com/libp2p/rust-libp2p/pull/2090/commits/42ff89ebc5402dd8a8aaa6f7a7dc9ef237c3491a.

Issue #2061, opened 2021-04-28
muxers/mplex: Allow up to 60 bit long stream IDs


mxinden:
According to the [mplex specification](https://github.com/libp2p/specs/blob/master/mplex/README.md) a stream ID embedded in an mplex header can be up to 60 bit long.

> The maximum header length is 9 bytes (per the unsigned-varint spec). With 9 continuation bits and 3 message flag bits the maximum stream ID is 60 bits (maximum value of 2^60 - 1).

Today, Rust `libp2p-mplex` restricts mplex headers to 32 bit.

https://github.com/libp2p/rust-libp2p/blob/29d969be340b610e19bf05661b731883f25e6821/muxers/mplex/src/codec.rs#L156

The rust `libp2p-mplex` implementation should support up 60 bit long stream IDs like specified in the mplex specification.

See https://github.com/libp2p/specs/pull/317 for details.

whereistejas:
Hi @mxinden ! If possible, I would love to take up this issue.

mxinden:
Closing here since https://github.com/libp2p/rust-libp2p/pull/2094 is merged.

PR #2094, opened 2021-06-05
muxers/mplex: Allow up to 60 bit long stream IDs

whereistejas:
hi @mxinden,
I have opened this PR for closing issue: #2061

Do we also need write a test to check if we can actually use 60 bit stream ID?

### DIFF
--- a/muxers/mplex/CHANGELOG.md
+++ b/muxers/mplex/CHANGELOG.md
@@ -1,10 +1,7 @@
-# 0.39.0 [unreleased]
-
-- Increased the lenght of Stream ID to 60 bits.
- 
 # 0.29.0 [unreleased]
 
 - Update dependencies.
+- Increased the length of Stream ID to 60 bits.
 
 # 0.28.0 [2021-03-17]
 
--- a/muxers/mplex/CHANGELOG.md
+++ b/muxers/mplex/CHANGELOG.md
@@ -1,7 +1,10 @@
 # 0.29.0 [unreleased]
 
 - Update dependencies.
-- Increased the length of Stream ID to 60 bits.
+
+- Support stream IDs of up to 60 bit length. See [PR 2094] for details.
+
+[PR 2094]: https://github.com/libp2p/rust-libp2p/pull/2094
 
 # 0.28.0 [2021-03-17]
 
--- a/muxers/mplex/CHANGELOG.md
+++ b/muxers/mplex/CHANGELOG.md
@@ -1,3 +1,7 @@
+# 0.39.0 [unreleased]
+
+- Increased the lenght of Stream ID to 60 bits.
+ 
 # 0.29.0 [unreleased]
 
 - Update dependencies.
--- a/muxers/mplex/src/codec.rs
+++ b/muxers/mplex/src/codec.rs
@@ -305,4 +305,33 @@ mod tests {
         let ok_msg = Frame::Data { stream_id: LocalStreamId { num: 123, role }, data };
         assert!(enc.encode(ok_msg, &mut out).is_ok());
     }
+
+    #[test]
+    fn test_60bit_stream_id() {
+        // create new codec object for encoding and decoding our frame
+        let mut codec = Codec::new();
+        // create a u64 stream ID
+        let id : u64 = u32::MAX as u64 + 1 ;
+        let stream_id = LocalStreamId { num: id, role: Endpoint::Dialer };
+
+        // open a new frame with that stream ID
+        let original_frame = Frame::Open { stream_id };
+
+        // encode that frame
+        let mut enc_frame = BytesMut::new();
+        match codec.encode(original_frame, &mut enc_frame) {
+            Err(e) => panic!("{}", e.to_string()),
+            _ => {}
+        };
+
+        // decode encoded frame and extract stream ID
+        let dec_string_id = match codec.decode(&mut enc_frame) {
+            Err(e) => panic!("{}", e.to_string()),
+            Ok(decoded_frame) => decoded_frame.unwrap().remote_id(),
+        };
+        
+        assert_eq!(dec_string_id.num, stream_id.num);
+        
+    }
+
 }
--- a/muxers/mplex/src/codec.rs
+++ b/muxers/mplex/src/codec.rs
@@ -47,7 +47,7 @@ pub(crate) const MAX_FRAME_SIZE: usize = 1024 * 1024;
 /// > the corresponding local ID has the role `Endpoint::Listener`.
 #[derive(Copy, Clone, PartialEq, Eq, Debug)]
 pub struct LocalStreamId {
-    num: u32,
+    num: u64,
     role: Endpoint,
 }
 
@@ -63,7 +63,7 @@ impl fmt::Display for LocalStreamId {
 impl Hash for LocalStreamId {
     #![allow(clippy::derive_hash_xor_eq)]
     fn hash<H: Hasher>(&self, state: &mut H) {
-        state.write_u32(self.num);
+        state.write_u64(self.num);
     }
 }
 
@@ -76,17 +76,17 @@ impl nohash_hasher::IsEnabled for LocalStreamId {}
 /// [`RemoteStreamId::into_local()`].
 #[derive(Copy, Clone, PartialEq, Eq, Hash, Debug)]
 pub struct RemoteStreamId {
-    num: u32,
+    num: u64,
     role: Endpoint,
 }
 
 impl LocalStreamId {
-    pub fn dialer(num: u32) -> Self {
+    pub fn dialer(num: u64) -> Self {
         Self { num, role: Endpoint::Dialer }
     }
 
     #[cfg(test)]
-    pub fn listener(num: u32) -> Self {
+    pub fn listener(num: u64) -> Self {
         Self { num, role: Endpoint::Listener }
     }
 
@@ -107,11 +107,11 @@ impl LocalStreamId {
 }
 
 impl RemoteStreamId {
-    fn dialer(num: u32) -> Self {
+    fn dialer(num: u64) -> Self {
         Self { num, role: Endpoint::Dialer }
     }
 
-    fn listener(num: u32) -> Self {
+    fn listener(num: u64) -> Self {
         Self { num, role: Endpoint::Listener }
     }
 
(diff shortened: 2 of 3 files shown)


mxinden:
> Do we also need write a test to check if we can actually use 60 bit stream ID?

Yes please, that would be great!

whereistejas:
Hi @mxinden,
I'm having a hard time figuring out how to write a test for this. I tried to go through the programs in the examples directory, looking for something that I could use. Can you please give me a little guidance about how I should proceed with writing this test?

mxinden:
`Codec` implements both `Encode` and `Decode`. To test your patch I would encode a `Frame::Open` with a  stream ID `> 2^32` into a `BytesMut` and then decode said `BytesMut` expecting the two stream IDs to equal.

Does that help @whereistejas?

whereistejas:
Hi @mxinden ,
I have finished writing the test. Your suggestion helped me a bunch, thanks! :)

muxers/mplex/src/codec.rs
```diff
@@ -305,4 +305,33 @@ mod tests {
         let ok_msg = Frame::Data { stream_id: LocalStreamId { num: 123, role }, data };
         assert!(enc.encode(ok_msg, &mut out).is_ok());
     }
+
+    #[test]
+    fn test_60bit_stream_id() {
+        // create new codec object for encoding and decoding our frame
```
mxinden, line 311:
```suggestion
        // Create new codec object for encoding and decoding our frame.
```

muxers/mplex/src/codec.rs
```diff
@@ -305,4 +305,33 @@ mod tests {
...
+        let mut codec = Codec::new();
+        // create a u64 stream ID
```
mxinden, line 313:
```suggestion
        // Create a u64 stream ID.
```

muxers/mplex/src/codec.rs
```diff
@@ -305,4 +305,33 @@ mod tests {
...
+        let id : u64 = u32::MAX as u64 + 1 ;
+        let stream_id = LocalStreamId { num: id, role: Endpoint::Dialer };
+
+        // open a new frame with that stream ID
+        let original_frame = Frame::Open { stream_id };
+
+        // encode that frame
```
mxinden, line 320:
```suggestion
        // Encode that frame.
```

muxers/mplex/src/codec.rs
```diff
@@ -305,4 +305,33 @@ mod tests {
...
+        let mut enc_frame = BytesMut::new();
+        match codec.encode(original_frame, &mut enc_frame) {
+            Err(e) => panic!("{}", e.to_string()),
+            _ => {}
+        };
```
mxinden, line 325:
```suggestion
        codec.encode(original_frame, &mut enc_frame).expect("Encoding to succeed.");
```

muxers/mplex/src/codec.rs
```diff
@@ -305,4 +305,33 @@ mod tests {
...
+
+        // decode encoded frame and extract stream ID
```
mxinden, line 327:
```suggestion
        // Cecode encoded frame and extract stream ID.
```

muxers/mplex/src/codec.rs
```diff
@@ -305,4 +305,33 @@ mod tests {
         let ok_msg = Frame::Data { stream_id: LocalStreamId { num: 123, role }, data };
         assert!(enc.encode(ok_msg, &mut out).is_ok());
     }
+
+    #[test]
+    fn test_60bit_stream_id() {
+        // create new codec object for encoding and decoding our frame
+        let mut codec = Codec::new();
+        // create a u64 stream ID
+        let id : u64 = u32::MAX as u64 + 1 ;
```
mxinden, line 314:
```suggestion
        let id: u64 = u32::MAX as u64 + 1 ;
```

muxers/mplex/src/codec.rs
```diff
@@ -305,4 +305,33 @@ mod tests {
...
+        match codec.encode(original_frame, &mut enc_frame) {
+            Err(e) => panic!("{}", e.to_string()),
+            _ => {}
+        };
+
+        // decode encoded frame and extract stream ID
+        let dec_string_id = match codec.decode(&mut enc_frame) {
+            Err(e) => panic!("{}", e.to_string()),
+            Ok(decoded_frame) => decoded_frame.unwrap().remote_id(),
+        };
```
mxinden, line 331:
```suggestion
        let dec_string_id = codec.decode(&mut enc_frame)
            .expect("Decoding to succeed.")
            .map(|f| f.remote_id())
            .unwrap();
```

muxers/mplex/src/codec.rs
```diff
@@ -305,4 +305,33 @@ mod tests {
...
+        
+        assert_eq!(dec_string_id.num, stream_id.num);
+
```
mxinden, line 334:
```suggestion
```

muxers/mplex/src/codec.rs
```diff
@@ -305,4 +305,33 @@ mod tests {
...
+        // decode encoded frame and extract stream ID
+        let dec_string_id = match codec.decode(&mut enc_frame) {
+            Err(e) => panic!("{}", e.to_string()),
+            Ok(decoded_frame) => decoded_frame.unwrap().remote_id(),
+        };
+        
+        assert_eq!(dec_string_id.num, stream_id.num);
+        
+    }
+
```
mxinden, line 336:
```suggestion
```

muxers/mplex/src/codec.rs
```diff
@@ -305,4 +305,33 @@ mod tests {
...
+    #[test]
+    fn test_60bit_stream_id() {
+        // create new codec object for encoding and decoding our frame
+        let mut codec = Codec::new();
+        // create a u64 stream ID
+        let id : u64 = u32::MAX as u64 + 1 ;
+        let stream_id = LocalStreamId { num: id, role: Endpoint::Dialer };
+
+        // open a new frame with that stream ID
```
mxinden, line 317:
```suggestion
        // Open a new frame with that stream ID.
```

mxinden (commented):
Test looks great. Just a couple of smaller suggestions.

Would you mind including a changelog entry in `muxers/mplex/CHANGELOG.md`?

whereistejas:
> Would you mind including a changelog entry in `muxers/mplex/CHANGELOG.md`?
Are we currently on version 0.39.0? I'm asking so that I may add the version in the change logs.

muxers/mplex/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# 0.39.0 [unreleased]
+
+- Increased the lenght of Stream ID to 60 bits.
```
mxinden, line 3: Please move this line under the `# 0.29.0 [...]` section below the `- Update dependencies` line.

muxers/mplex/src/codec.rs
```diff
@@ -18,14 +18,10 @@
 // FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
 // DEALINGS IN THE SOFTWARE.
 
-use asynchronous_codec::{Decoder, Encoder};
```
mxinden: Please revert the code formatting introduced in https://github.com/libp2p/rust-libp2p/pull/2094/commits/a01622283045025a33f6578082b5ed461c7e0372.

I am not opposed to using rustfmt, though I think, in case we do decide to use it across the repository, it should be a separate pull request. See https://github.com/libp2p/rust-libp2p/issues/1668 for details.

whereistejas: I had accidentally used `rustfmt` in the commit [ead5d97](https://github.com/libp2p/rust-libp2p/pull/2094/commits/ead5d97ae7d18dbc86abe7f5087d8e5deba00608). The commit [a016222](https://github.com/libp2p/rust-libp2p/pull/2094/commits/a01622283045025a33f6578082b5ed461c7e0372) reverted the formatting that was done by `rustfmt`.

whereistejas:
I think, we are pretty much finished. I would love to take up `rustfmt` issue in a seperate PR. The least I can do is open the PR. I will leave the decision of whether or not to merge it to the reviewers.

muxers/mplex/CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # 0.29.0 [unreleased]
 
 - Update dependencies.
+- Increased the length of Stream ID to 60 bits.
```
mxinden, line 4:
```suggestion

- Support stream IDs of up to 60 bit length. See [PR 2094] for details.

[PR 2094]: https://github.com/libp2p/rust-libp2p/pull/2094
```

mxinden (approved):
Thanks for the help @whereistejas.

whereistejas:
Thanks @mxinden! :)

senden9:
Hi!
I know that this pull request is already merged. But if I look at the changed files I see no restriction to 60 Bit. As far as I can see there is nothing that would stop me from creating a 61, 62, 63 or 64 Bit ID.

whereistejas:
@senden9 
I can be completely wrong here, but my understanding is that the first 3 bits are reserved for the message flag. Please refer to this document: https://github.com/libp2p/specs/blob/master/mplex/README.md

@mxinden can you please help out here? Also, I think it makes sense to add a test, that fails if we try to use a stream ID greater than 60bits.

senden9:
Right. It seams that the first three bits are reserved for flags. And the lowest three bits are shifted away in encoding. So `LocalStreamId::num` seams not to be the ID, but the Header (as called your linked document).

Anyway as I want introduce any more "noise" in the notification inboxes of the people of this PR: I will think a bit about that and then create a new Discussion or Issue if anything seems strange to me. Thanks for the fast reply @whereistejas

mxinden:
> I know that this pull request is already merged.

Feedback is always welcome, no matter what status the pull request is in ;)

> But if I look at the changed files I see no restriction to 60 Bit. As far as I can see there is nothing that would stop me from creating a 61, 62, 63 or 64 Bit ID.

Correct. You can still call `RemoteStreamId::dialer(u64::MAX)`. I agree that this is not ideal and that there should be something stopping you, e.g. through an `assert`.

On the other hand:

- I doubt anyone will ever reach stream IDs of more than 60 bit [given that it is monotonically increased by 1](https://github.com/libp2p/rust-libp2p/blob/c1ef4bffd225a78cefe8fa43dc8ee18e03ff4f93/muxers/mplex/src/codec.rs#L93-L98).
- `LocalStreamId` is not exposed by the crate, and thus an internal construct only.


All that said, the goal of this pull request is to support stream IDs of up to 60 bit coming from a remote peer. Using a `u64` instead of a `u32` does allow that.

This is the relevant excerpt from the specification:

> The maximum header length is 9 bytes (per the unsigned-varint spec). With 9 continuation bits and 3 message flag bits the maximum stream ID is 60 bits (maximum value of 2^60 - 1).

https://github.com/libp2p/specs/blob/master/mplex/README.md#message-format


@senden9 does the above make sense?

senden9:
@mxinden yes, thanks for the explanation. The "Internal construct only"-part is also a good argument for why it isn't too bad to not block the constructing or reaching of that invalid state.

PR #2099, opened 2021-06-09
sim open

dvc94ch:
This fixes the failing test case in #2066


misc/multistream-select/src/protocol.rs
```diff
@@ -238,7 +290,7 @@ impl<R> MessageIO<R> {
         MessageReader { inner: self.inner.into_reader() }
     }
 
-    /// Drops the [`MessageIO`] resource, yielding the underlying I/O stream.
+    /// Draops the [`MessageIO`] resource, yielding the underlying I/O stream.
```
rkuhn, line 293: oops

dvc94ch:
@mxinden do you have a timeline for when you expect this feature to be merged?

mxinden:
> @mxinden do you have a timeline for when you expect this feature to be merged?

I am not sure what you are referring to with "this feature". In case you are referring to the solution suggested in this pull request: I don't think we should merge it as (a) it is adding more complexity in general and (b) it is polluting the function signature for stream upgrades which will never need this functionality. In case you are referring to support for multistream-select simultaneous open in general: I don't know, especially as it depends on so many other efforts (e.g. circuit relay v2, direct connection upgrade through relay, ...).

See https://github.com/libp2p/rust-libp2p/pull/2066#issuecomment-867606365 for details.

PR #2103, opened 2021-06-14
Convert message_id_fn's to closures

ethDreamer:
This will allow capturing variables in these closures so that we can
make these functions aware of the forkId (necessary for altair).

### DIFF
--- a/protocols/gossipsub/CHANGELOG.md
+++ b/protocols/gossipsub/CHANGELOG.md
@@ -1,11 +1,9 @@
-# 0.34.0 [unreleased]
-
-- Allow message_id_fn's to accept closures that capture variables.
-  [PR 2103](https://github.com/libp2p/rust-libp2p/pull/2103)
-
 # 0.33.0 [unreleased]
 
 - Update dependencies.
+
+- Allow `message_id_fn`s to accept closures that capture variables.
+  [PR 2103](https://github.com/libp2p/rust-libp2p/pull/2103)
 
 # 0.32.0 [2021-07-12]
 
--- a/protocols/gossipsub/CHANGELOG.md
+++ b/protocols/gossipsub/CHANGELOG.md
@@ -1,3 +1,8 @@
+# 0.34.0 [unreleased]
+
+- Allow message_id_fn's to accept closures that capture variables.
+  [PR 2103](https://github.com/libp2p/rust-libp2p/pull/2103)
+
 # 0.33.0 [unreleased]
 
 - Update dependencies.
--- a/protocols/gossipsub/src/config.rs
+++ b/protocols/gossipsub/src/config.rs
@@ -20,6 +20,7 @@
 
 use std::borrow::Cow;
 use std::time::Duration;
+use std::sync::Arc;
 
 use libp2p_core::PeerId;
 
@@ -69,8 +70,8 @@ pub struct GossipsubConfig {
     duplicate_cache_time: Duration,
     validate_messages: bool,
     validation_mode: ValidationMode,
-    message_id_fn: fn(&GossipsubMessage) -> MessageId,
-    fast_message_id_fn: Option<fn(&RawGossipsubMessage) -> FastMessageId>,
+    message_id_fn: Arc<dyn Fn(&GossipsubMessage) -> MessageId + Send + Sync + 'static>,
+    fast_message_id_fn: Option<Arc<dyn Fn(&RawGossipsubMessage) -> FastMessageId + Send + Sync + 'static>>,
     allow_self_origin: bool,
     do_px: bool,
     prune_peers: usize,
@@ -234,6 +235,7 @@ impl GossipsubConfig {
     /// interpreted as the fast message id. Default is None.
     pub fn fast_message_id(&self, message: &RawGossipsubMessage) -> Option<FastMessageId> {
         self.fast_message_id_fn
+            .as_ref()
             .map(|fast_message_id_fn| fast_message_id_fn(message))
     }
 
@@ -396,7 +398,7 @@ impl Default for GossipsubConfigBuilder {
                 duplicate_cache_time: Duration::from_secs(60),
                 validate_messages: false,
                 validation_mode: ValidationMode::Strict,
-                message_id_fn: |message| {
+                message_id_fn: Arc::new(|message| {
                     // default message id is: source + sequence number
                     // NOTE: If either the peer_id or source is not provided, we set to 0;
                     let mut source_string = if let Some(peer_id) = message.source.as_ref() {
@@ -409,7 +411,7 @@ impl Default for GossipsubConfigBuilder {
                     source_string
                         .push_str(&message.sequence_number.unwrap_or_default().to_string());
                     MessageId::from(source_string)
-                },
+                }),
                 fast_message_id_fn: None,
                 allow_self_origin: false,
                 do_px: false,
@@ -574,8 +576,10 @@ impl GossipsubConfigBuilder {
     ///
     /// The function takes a [`GossipsubMessage`] as input and outputs a String to be
     /// interpreted as the message id.
-    pub fn message_id_fn(&mut self, id_fn: fn(&GossipsubMessage) -> MessageId) -> &mut Self {
-        self.config.message_id_fn = id_fn;
+    pub fn message_id_fn<F>(&mut self, id_fn: F) -> &mut Self
+    where F: Fn(&GossipsubMessage) -> MessageId + Send + Sync + 'static
+    {
+        self.config.message_id_fn = Arc::new(id_fn);
         self
     }
 
@@ -587,11 +591,13 @@ impl GossipsubConfigBuilder {
     ///
     /// The function takes a [`RawGossipsubMessage`] as input and outputs a String to be interpreted
     /// as the fast message id. Default is None.
-    pub fn fast_message_id_fn(
+    pub fn fast_message_id_fn<F>(
         &mut self,
-        fast_id_fn: fn(&RawGossipsubMessage) -> FastMessageId,
-    ) -> &mut Self {
-        self.config.fast_message_id_fn = Some(fast_id_fn);
+        fast_id_fn: F
+    ) -> &mut Self
+    where F: Fn(&RawGossipsubMessage) -> FastMessageId + Send + Sync + 'static,
+    {
+        self.config.fast_message_id_fn = Some(Arc::new(fast_id_fn));
         self
     }
 
--- a/protocols/gossipsub/src/config.rs
+++ b/protocols/gossipsub/src/config.rs
@@ -823,6 +823,8 @@ mod test {
     use super::*;
     use std::collections::hash_map::DefaultHasher;
     use std::hash::{Hash, Hasher};
+    use crate::Topic;
+    use crate::topic::IdentityHash;
 
     #[test]
     fn create_thing() {
@@ -832,6 +834,21 @@ mod test {
             .unwrap();
 
         dbg!(builder);
+    }
+
+    #[inline]
+    fn get_gossipsub_message() -> GossipsubMessage {
+        GossipsubMessage {
+            source: None,
+            data: vec![12, 34, 56],
+            sequence_number: None,
+            topic: Topic::<IdentityHash>::new("test").hash(),
+        }
+    }
+
+    #[inline]
+    fn get_expected_message_id() -> MessageId {
+        MessageId::from([49, 55, 56, 51, 56, 52, 49, 51, 52, 51, 52, 55, 51, 51, 53, 52, 54, 54, 52, 49, 101])
     }
 
     fn message_id_plain_function(message: &GossipsubMessage) -> MessageId {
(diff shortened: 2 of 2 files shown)


AgeManning:
@mxinden - This is a change we were considering, but perhaps need to flag other gossipsub users if they would like to go down this path. 

This is a breaking API change. 

We need to modify the message-id fn based on an external variable so this PR generalises the message-id fn with a trait object. In principle this has a performance hit for every message (we can get 1000/s). 

Potentially another approach would be to add a mutable function to the behaviour that allows users to dynamically change the message-id fn. Abruptly changing the function during operation can lead to scoring issues because IHAVE/IWANT and promises will not function correctly for a brief period of time, also all nodes would have to change the id fn simultaneously. 

Curious about anyone else's thoughts on this.

protocols/gossipsub/src/config.rs
```diff
@@ -69,8 +70,8 @@ pub struct GossipsubConfig {
     duplicate_cache_time: Duration,
     validate_messages: bool,
     validation_mode: ValidationMode,
-    message_id_fn: fn(&GossipsubMessage) -> MessageId,
-    fast_message_id_fn: Option<fn(&RawGossipsubMessage) -> FastMessageId>,
+    message_id_fn: Arc<dyn Fn(&GossipsubMessage) -> MessageId + Send + Sync + 'static>,
```
mxinden, line 73: Could you explain why an `Arc` instead of a `Box<dyn Fn ...>` is needed here?

mxinden (commented):
Given that navigating the space of function pointers (`fn`) and `Fn` is rather complex, especially when it comes to not introducing breaking changes in the future, would you mind including unit tests where `message_id` is set with:

- A plain function
- A closure not capturing anything from the environment
- A closure that does capture variables

> This is a breaking API change.

As long as it still enables all use-cases I am not opposed to a breaking API change.

> We need to modify the message-id fn based on an external variable so this PR generalises the message-id fn with a trait object. In principle this has a performance hit for every message (we can get 1000/s).

Would you mind expanding on the concrete use-case? Also I am not quite sure what the reason of the above mentioned performance hit is? The dynamic dispatching of the function pointer?

> Potentially another approach would be to add a mutable function to the behaviour that allows users to dynamically change the message-id fn. Abruptly changing the function during operation can lead to scoring issues because IHAVE/IWANT and promises will not function correctly for a brief period of time, also all nodes would have to change the id fn simultaneously.

Off the top of my head, this sounds like a footgun.

AgeManning:
> Would you mind expanding on the concrete use-case? Also I am not quite sure what the reason of the above mentioned performance hit is? The dynamic dispatching of the function pointer?

Yes, I was referring to the dynamic dispatch for every message when we need to calculate its id. Its probably marginal. The use case I think we were chasing was a closure that captures a variable. We need an external variable along with the gossipsub message to build our ids. 

I think this PR wasn't meant for this repo so early, but it might stand with the additional tests you've requested. I think we also need to do some further testing to make sure this change actually solves our use case. 

I'll let @ethDreamer comment on the other points.

protocols/gossipsub/src/config.rs
ethDreamer: Hi @mxinden 
From the derive above, the `GossipsubConfig` needs to be `Clone`. So if I used a `Box`, I would need to add an extra `Clone` trait bound inside the angle brackets. Because `Fn(&GossipsubMessage) -> MessageId` is already a non-auto trait, this requires us to create a `NewTrait` implementing `Fn<(&'r RawGossipsubMessage,)> + Clone` and then make this a `Box<NewTrait>`.

All of that makes the code less readable, and I'm not actually sure if any closure can be made `Clone` in the first place. But if you take a step back and consider the `Arc<T>` for a moment, the only time you need to pay the synchronization overhead is when the reference count changes (during `Arc<T>::new()`, `Arc<T>::clone()`, and `Arc<T>::drop()`). The [atomic counters aren't touched at all during Deref](https://doc.rust-lang.org/src/alloc/sync.rs.html#1299-1306).

So it should produce no performance hit compared to a `Box` and is actually probably faster than cloning the whole closure each time the `GossipsubConfig` is cloned.

ethDreamer:
Hi @mxinden,
I've updated this PR to include the 3 unit tests you requested.  They all compile just fine.  Note that if you try to compile these unit tests WITHOUT this change, the closure that captures variables will fail:

```
error[E0308]: mismatched types
   --> protocols/gossipsub/src/config.rs:901:28
    |
891 |           let closure = move |message: &GossipsubMessage| {
    |  _______________________-
892 | |             let mut s = DefaultHasher::new();
893 | |             message.data.hash(&mut s);
894 | |             let mut v = s.finish().to_string();
895 | |             v.push(captured);
896 | |             MessageId::from(v)
897 | |         };
    | |_________- the found closure
...
901 |               .message_id_fn(closure)
    |                              ^^^^^^^ expected fn pointer, found closure
    |
    = note: expected fn pointer `for<'r> fn(&'r types::GossipsubMessage) -> types::MessageId`
                  found closure `[closure@protocols/gossipsub/src/config.rs:891:23: 897:10]`
note: closures can only be coerced to `fn` types if they do not capture any variables
   --> protocols/gossipsub/src/config.rs:895:20
    |
895 |             v.push(captured);
    |                    ^^^^^^^^ `captured` captured here

For more information about this error, try `rustc --explain E0308`.
error: could not compile `libp2p-gossipsub` due to previous error
warning: build failed, waiting for other jobs to finish...
error: build failed
```

ethDreamer:
Hi @AgeManning / @mxinden 
I wanted to quantify the performance hit we might take from dynamic dispatch so I wrote some code to measure it (which I'll paste below). I profiled the code linking against the current `master` branch, and against this `message_id_closure` branch  (the only differences being the changes in this PR).

Here's an example of what the output looks like when timing multiple iterations that each call the closure `1024` times:
```
[mark@trost gossip-timing]$ cargo bench
    Finished bench [optimized] target(s) in 0.09s
     Running unittests (target/release/deps/gossip_timing-4edcb2389def7fe4)

running 1 test
test tests::bench_1024_message_ids ... bench:      71,697 ns/iter (+/- 1,740)

test result: ok. 0 passed; 0 failed; 0 ignored; 1 measured; 0 filtered out; finished in 3.62s
```

Provided I did the profiling correctly, the differences are negligible:
```
[mark@trost gossip-timing]$ ##### LINKING AGAINST `master` BRANCH #####
[mark@trost gossip-timing]$ cargo bench |& grep iter | awk '{ print $5 " " $6 " " $7 " " $8 }'
63,292 ns/iter (+/- 2,505)
[mark@trost gossip-timing]$ cargo bench |& grep iter | awk '{ print $5 " " $6 " " $7 " " $8 }'
81,328 ns/iter (+/- 15,014)
[mark@trost gossip-timing]$ cargo bench |& grep iter | awk '{ print $5 " " $6 " " $7 " " $8 }'
81,221 ns/iter (+/- 2,502)
[mark@trost gossip-timing]$ cargo bench |& grep iter | awk '{ print $5 " " $6 " " $7 " " $8 }'
78,773 ns/iter (+/- 2,165)
[mark@trost gossip-timing]$ cargo bench |& grep iter | awk '{ print $5 " " $6 " " $7 " " $8 }'
64,997 ns/iter (+/- 16,279)
[mark@trost gossip-timing]$ cargo bench |& grep iter | awk '{ print $5 " " $6 " " $7 " " $8 }'
65,066 ns/iter (+/- 4,996)
[mark@trost gossip-timing]$ ##### SWITCH TO LINKING AGAINST `message_id_closure` BRANCH #####
[mark@trost gossip-timing]$ cargo bench |& grep iter | awk '{ print $5 " " $6 " " $7 " " $8 }'
69,413 ns/iter (+/- 27,862)
[mark@trost gossip-timing]$ cargo bench |& grep iter | awk '{ print $5 " " $6 " " $7 " " $8 }'
80,668 ns/iter (+/- 11,902)
[mark@trost gossip-timing]$ cargo bench |& grep iter | awk '{ print $5 " " $6 " " $7 " " $8 }'
80,858 ns/iter (+/- 1,378)
[mark@trost gossip-timing]$ cargo bench |& grep iter | awk '{ print $5 " " $6 " " $7 " " $8 }'
80,177 ns/iter (+/- 1,963)
[mark@trost gossip-timing]$ cargo bench |& grep iter | awk '{ print $5 " " $6 " " $7 " " $8 }'
65,871 ns/iter (+/- 1,335)
[mark@trost gossip-timing]$ cargo bench |& grep iter | awk '{ print $5 " " $6 " " $7 " " $8 }'
69,045 ns/iter (+/- 3,616)
```

Going by the results that had small spreads, each call takes `~80±2 ns`, independent of which branch is linked against.

ethDreamer:
I used the `#[bench]` attribute and the unstable `test` feature to profile the code. So I had to use the nightly compiler. There doesn't appear to be a lot of documentation on this other than:

- [The Rust Book (section on benchmark tests)](https://doc.rust-lang.org/1.4.0/book/benchmark-tests.html)
- ["The Nightly Book" (section on the test crate)](https://doc.rust-lang.org/nightly/unstable-book/library-features/test.html)
- [test::Bencher docs](https://doc.rust-lang.org/1.1.0/test/struct.Bencher.html)

The tricky part is ensuring the compiler didn't optimize away things because it knows the result isn't being used. I tried to prevent by using `test::black_box` as indicated in the documentation. Anyway here's the source:

```
#![feature(test)]

use libp2p::gossipsub::MessageId;
use libp2p::gossipsub::{GossipsubConfig, GossipsubConfigBuilder, GossipsubMessage, TopicHash};
use std::collections::hash_map::DefaultHasher;

extern crate test;

fn message_id_plain_function(message: &GossipsubMessage) -> MessageId {
    let mut s = DefaultHasher::new();
    message.data.hash(&mut s);
    let mut v = s.finish().to_string();
    v.push('e');
    MessageId::from(v)
}

fn main() {
    let message = GossipsubMessage {
        source: None,
        data: vec![12, 34, 56],
        sequence_number: None,
        topic: TopicHash::from_raw("topichash01234"),
    };
    println!("message id: {:?}", message_id_plain_function(&message));
}

#[cfg(test)]
mod tests {
    use super::*;
    use test::Bencher;

    #[bench]
    fn bench_1024_message_ids(b: &mut Bencher) {
        let builder: GossipsubConfig = GossipsubConfigBuilder::default()
            .protocol_id_prefix("purple")
            .message_id_fn(message_id_plain_function)
            .build()
            .unwrap();

        b.iter(move || {
            let message = test::black_box(GossipsubMessage {
                source: None,
                data: vec![12, 34, 56],
                sequence_number: None,
                topic: TopicHash::from_raw("topichash01234"),
            });

            (0..255).fold(0, |a, b| {
                test::black_box(builder.message_id(&message).0[0]) + a + b
            });
            (0..255).fold(1, |a, b| {
                test::black_box(builder.message_id(&message).0[0]) + a + b
            });
            (0..255).fold(2, |a, b| {
                test::black_box(builder.message_id(&message).0[0]) + a + b
            });
            (0..255).fold(3, |a, b| {
                test::black_box(builder.message_id(&message).0[0]) + a + b
            });
        });
    }
}
```

AgeManning:
Nice! Thanks. Changes look fine to me

protocols/gossipsub/src/config.rs
```diff
@@ -825,4 +835,81 @@ mod test {
 
         dbg!(builder);
     }
+
+    #[inline]
```
mxinden, line 839:
```suggestion
```

Inlining does not always bring a performance win, e.g. the increased code size might actually hurt performance, and this is likely not a compile time win. What is the reasoning for suggesting to the compiler to inline this function? Does it really matter for testing code in the first place?

protocols/gossipsub/src/config.rs
```diff
@@ -825,4 +835,81 @@ mod test {
...
+        let builder: GossipsubConfig = GossipsubConfigBuilder::default()
+            .protocol_id_prefix("purple")
+            .message_id_fn(closure)
+            .build()
+            .unwrap();
+
+        let result = builder.message_id(&get_gossipsub_message());
+        assert_eq!(result, get_expected_message_id());
+    }
+
```
mxinden, line 914:
```suggestion
```

mxinden (commented):
Very much appreciate the benchmark driven approach @ethDreamer. Also thank you for the unit tests!

Just a couple of minor comments. Would you mind adding a changelog entry as well?

ethDreamer:
> Very much appreciate the benchmark driven approach @ethDreamer. Also thank you for the unit tests!
> 
> Just a couple of minor comments. Would you mind adding a changelog entry as well?

I've made a changelog entry, though I'm not sure if the versioning is correct. I've also made the other changes you requested. Let me know if you need anything else.

protocols/gossipsub/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+# 0.34.0 [unreleased]
+
+- Allow message_id_fn's to accept closures that capture variables.
+  [PR 2103](https://github.com/libp2p/rust-libp2p/pull/2103)
+
 # 0.33.0 [unreleased]
 
 - Update dependencies.
```
mxinden, line 8:
```suggestion
# 0.33.0 [unreleased]

- Update dependencies.

- Allow `message_id_fn`s to accept closures that capture variables.
  [PR 2103](https://github.com/libp2p/rust-libp2p/pull/2103)
```

mxinden (approved):
Thanks @ethDreamer!

PR #2104, opened 2021-06-15
NixOS flake build setup

Anderssorby:
With nix flakes installed you can with this reproducibly install dependencies and build the project with `nix build`. This installs all the dependencies even rust (nightly but that can be changed).


dvc94ch:
While nix and nix flakes are really cool, they are for binaries or dynamic libraries and don't add any value in this case, since rust-libp2p provides neither.

.gitignore
```diff
@@ -1,2 +1 @@
 target
-Cargo.lock
```
mxinden, line 2: Why would you want to check in the `Cargo.lock` file as a library?

mxinden (changes requested):
Appreciate the contribution, though as David noted, I don't see the benefit of this setup, given that rust-libp2p is a library and not a binary.

In addition, as a maintainer running on Debian I have no way / bandwidth to validate and maintain these going forward.

.gitignore
Anderssorby: This is needed by nix for reproducible builds.

Anderssorby:
It is not super important to merge this, but this is just work I needed to do to set up a development environment on NixOS. I needed to set up env parameters for protobuf to get it to working. This is more of a gesture to those preferring to use nix and/or NixOS with flakes to have a ready to go environment.

dvc94ch:
@Anderssorby I remember when I was using nixos a few years ago (and then running guixsd) also trying to go down this path of packaging crates individually. I think I've gained some perspective on this. nixos was designed around packages producing a binary written in c or a dynamic library written in c. And it's great as a c package manager, actually the best. But the needs of distro developers are different from the needs of application developers. This is why I think dolstra invented nixflakes, for application developers. But in rust it makes no sense to package the rlib of a crate (as it depends on the compiler version and the enabled features and the enabled features in any of its dependencies) and cargo is pretty good. So I'd recommend using nixflakes to install any system dependencies and just use cargo for building the rust part.

Anderssorby:
@dvc94ch Yes, this is only for the dev environment. Packaging the lib is not usefull. I think it should be standard for projects to have self contained declarative dev enviroments. Is it interesting if I remove all the package build stuff (naersk)?

dvc94ch:
It's been a really long time since I used nix.  So guix is more recent in my memory and I spent more time contributing to it (used arch linux for the last 4y), there we had a manifest that listed the packages required for a dev environment, in this case I it would be protobuf and openssl and would look something like the below. We don't track the oldest rustc we compile on, the latest cargo/rustc should always work (and that's generally throughout the rust ecosystem since we don't break backwards compatibility in an edition an the compiler retains support for old editions, even though sometimes that mean we have to keep weird behaviour for a very long time).

```scheme
(specifications->manifest
 '("protobuf" "openssl"))
```

mxinden:
> It is not super important to merge this, but this is just work I needed to do to set up a development environment on NixOS. I needed to set up env parameters for protobuf to get it to working. This is more of a gesture to those preferring to use nix and/or NixOS with flakes to have a ready to go environment.

:+1: hope this is useful for other NixOS users! 

Given that it is [discoverable through the search](https://github.com/libp2p/rust-libp2p/search?q=nix&type=issues) I would suggest closing here, but please don't delete the branch to allow others to fetch the code (and maybe update it in the future if need be).

Also, please add yourself to the [Notable Users](https://github.com/libp2p/rust-libp2p#notable-users) section, in case you want to show off what you are building.

Issue #2091, opened 2021-06-04
Use a HD wallet with libp2p


izolyomi:
Being aware of the libp2p project for a long time, I recently started experimenting with its current capabilities as a newbie. I planned to start with a hierarchical deterministic (HD) wallet that can generate both ed25519 and secp256k1 keys and use libp2p for safe authenticated communicate between the generated keys.

Unfortunately, I've found that the current `core::identity::{ed25519, secp256k1}` implementations do not seem to allow creating libp2p identities from an existing `ed25119_dalek::KeyPair` or a `PeerId` from a dalek or secp PublicKey, the preferred way of creating identities is always generating a random new one inside libp2p. In some cases, this might be worked around by hacking with serializing the keys to/from some byte formats, but not always and I don't think it's proper usage of the library.

Was disallowing construction of such identities intentional or just wasn't needed so far thus not implemented *yet*? If it wasn't a well-informed intentional decision, are you willing to implement this or accept related PRs to libp2p?

tomaka:
I'm answering to give the historical context.

> Was disallowing construction of such identities intentional 

Yes! The reason is to not expose any type from the `ed25519_dalek` library (or any other library for that matter) in the public API of libp2p, as `ed25519_dalek` is an implementation detail.
Exposing the types of `ed25519_dalek` would, in turn, make it annoying to use libp2p with, for example, `ed25519_zebra`, and would require libp2p to publish a breaking change version if it bumps its version of `ed25519_dalek`.

> this might be worked around by hacking with serializing the keys to/from some byte formats, but not always and I don't think it's proper usage of the library

That's indeed how you're supposed to do it!

izolyomi:
Thanks for the fast answer.

I understand the intention and concept by the explanation, but do not find this approach any clearer at first sight. What you suggest is essentially using an internal serialization byte format that is conceptually also a part of the library specs but is undocumented (at least I couldn't find any description), more volatile and harder to even detect changes in its format being unversioned.

Do I maybe misunderstand something here? Is the binary serialization format really supposed to provide a reasonable interface to connect a HD wallet to libp2p identities?

tomaka:
> an internal serialization byte format that is conceptually also a part of the library specs but is undocumented (at least I couldn't find any description)

I'm quite confident that it's this format described here: https://datatracker.ietf.org/doc/html/rfc8032#section-5.1.5

> The public key A is the encoding of the point [s]B.  First,
[...]

PR #2105, opened 2021-06-16
adds from conversions, borrows pubkey for encoding

sunny-g:


core/src/identity.rs
```diff
@@ -179,7 +179,7 @@ impl PublicKey {
 
     /// Encode the public key into a protobuf structure for storage or
     /// exchange with other nodes.
-    pub fn into_protobuf_encoding(self) -> Vec<u8> {
+    pub fn into_protobuf_encoding(&self) -> Vec<u8> {
```
thomaseizinger, line 182: Per Rust-convention, this would now have to be renamed to `to_protobuf_encoding`.

core/src/identity/ed25519.rs
```diff
@@ -101,6 +101,24 @@ impl From<SecretKey> for Keypair {
     }
 }
 
+impl From<ed25519::Keypair> for Keypair {
+    fn from(kp: ed25519::Keypair) -> Self {
+        Self(kp)
+    }
+}
```
thomaseizinger, line 108: I am not sure if exposing these is desirable. As far as I know, these newtypes exist to hide implementation details like the concrete version of the `ed25519` crate that is used.

If we expose conversion functions, any version bump of the underlying library ends up being a breaking change.

sunny-g: That makes sense - I'm pursuing this because the app I'm writing already uses `ed25519_dalek` and I wanted conversion to be simpler and infallible.

Maybe behind a feature flag?

mxinden: > Maybe behind a feature flag?

Even behind a feature flag a breaking change in `ed25519_dalek` would imply a breaking change in `libp2p-core`. Does that make sense @sunny-g?

For a similar discussion see https://github.com/libp2p/rust-libp2p/issues/2091#issuecomment-854498402.

sunny-g: The intent of putting it behind a feature flag would be to signal that it couldn't be relied on in-perpetuity, but could still be used in an opt-in basis. 

But otherwise yes, I totally understand not wanting to tie this library to a dependency. The only reason I see this is as somewhat of an issue is that `ed25519_dalek` is really popular, and is (or its dependencies are) used as a precursor for other higher-level crypto schemes, so having a zero-cost conversion would make `libp2p` Keypairs de-facto and cheaply interoperable with that ecosystem.

thomaseizinger: I understand the desire for efficiency but to put all of this into some perspective: There is usually only 1 keypair per libp2p application and that one is generated / converted on startup. Optimizing that at the expense of breaking encapsulation seems not worth it to me.

What exactly is your usecase that you need to construct the libp2p identity from an existing keypair?

In all of my applications, the network identity of a node is completely separated from any other cryptography that is happening. In particular, it is usually derived through deterministic derivation from an overall applications seed, together with the extended priv key of a Bitcoin wallet for example. I am not crypto expert but this kind of separation was sold as good practice to me so I am curious what your usecase looks like where you have an existing `ed25519_dalek` key pair and it is important that the network identity is using that.

sunny-g: @thomaseizinger Appreciate you asking; I'm not a crypto expert either. 

The gist of my use case is to use a libp2p keypair for network communication, as an IPNS identifier, and allow for [proxy re-encryption](https://crates.io/crates/recrypt) of messages signed to it without having to manage, cross-sign and distribute multiple keys, which is already a pain with one.

But my broader case is that `ed25519_dalek` and others like it are pretty popular, which would normally make it the de-facto type for me and others to use throughout an app that doesn't revolve around libp2p. And while I'm glad that libp2p is using it vs another vs rolling your own, Rust's strict typing + not allowing for cheap conversion kind of inhibits interoperability b/c now I have to do the conversions myself, make sure I `Zeroize` where I'm supposed to, make functions fallible when they shouldn't but can't-necessarily-guarantee-that-they-aren't, etc. 

Its not a huge deal, its just really nice to be able to share types (especially something as sensitive as crypto types) across my own and others' crates without having to re-invent the wheel.

thomaseizinger: Thanks for elaborating. Makes a lot of sense to me.

One point is the breaking change aspect of types, i.e. how likely we see a breaking change (and hence a version bump) in `ed25519_dalek` to happen. Given that the library is already at 1.0 that might be quite unlikely so from that perspective, exposing the dependency could be fine.

The other issue that was brought up in https://github.com/libp2p/rust-libp2p/issues/2091#issuecomment-854498402 is that `ed25519_dalek` is not the only `ed25519` implementation so if we are exposing conversions for the `ed25519_dalek` crate, why not also for the other crates? Where do you stop? :)

Is the complexity increase in `libp2p-core` really worth the added convenience of nice interop with all the different implementations? I tend to say no, especially because the serialization format of ed25519 keys is clearly specified in an RFC and therefore basically stable forever.

Not going down the rabbit hole of exposing these internals feels cleaner to me, even though it results in some not so nice conversions between these types on your end.

An `unstable` feature flag might address issue (1) but not the added complexity.

I am not a maintainer of rust-libp2p, just providing some thoughts to show that there aren't just benefits to adding these conversions. @mxinden will have to make the final call on how to move forward.

---

> The gist of my use case is to use a libp2p keypair for network communication and as an IPNS identifier without having to manage, cross-sign and distribute multiple keys, which is already a pain with one.

Regarding this: I assume the key you are using is still somewhat persistent and / or derived from a seed and you are not generating it randomly using `ed25519_dalek`? If you are deserializing the `ed25519_dalek` keypair from bytes, you can also just deserialize the libp2p one from the same bytes to get the same key. This sounds stupidly obvious I know but it may be helpful from a software design perspective because it means no single type is preferred over the other.

They are obviously still fallible but looking at it this way may make your code scream less "I want to be an infallible conversion" :)
Instead, the fact that you will be using the same keypair for multiple applications is a product of feeding the same bytes to two modules.

sunny-g: Understood, thanks for all the discussion and feedback :)

mxinden:
>  sunny-g closed this 14 hours ago 

```diff
- pub fn into_protobuf_encoding(self) -> Vec<u8> {
+ pub fn to_protobuf_encoding(&self) -> Vec<u8> {
```

@sunny-g I am still happy to accept this proposal though.

PR #2110, opened 2021-06-24
Add the ability for the behaviour to disconnect a peer

sergeyboyko0791:
Currently, we use a custom version of the [rust-libp2p](https://github.com/KomodoPlatform/rust-libp2p/tree/master) crate in the https://github.com/KomodoPlatform/atomicDEX-API project that allows us to disconnect peers from the network behaviour.
I would like to propose this functionality to allow the users to build their custom behaviors with the ability to disconnect peers.
Please look at the example of using `NetworkBehaviourAction::DisconnectPeer'`: https://github.com/KomodoPlatform/atomicDEX-API/blob/mm2.1/mm2src/mm2_libp2p/src/adex_ping.rs#L27

* Add 'NetworkBehaviourAction::DisconnectPeer'
* Add 'ExpandedSwarm::disconnect_peer_id'
* Add 'test_swarm_disconnect', 'test_behaviour_disconnect_all', 'test_behaviour_disconnect_one'

### DIFF
--- a/swarm/src/behaviour.rs
+++ b/swarm/src/behaviour.rs
@@ -296,11 +296,19 @@ pub enum NetworkBehaviourAction<TInEvent, TOutEvent> {
 
     /// Instructs the `Swarm` to initiate a graceful close of one or all connections
     /// with the given peer.
+    ///
+    /// Note: Closing a connection via
+    /// [`NetworkBehaviourAction::CloseConnection`] does not inform the
+    /// corresponding [`ProtocolsHandler`](s).
+    /// Closing a connection via a [`ProtocolsHandler`] can be done
+    /// either in a collaborative manner across [`ProtocolsHandler`]s
+    /// with [`ProtocolsHandler::connection_keep_alive`] or directly with
+    /// [`ProtocolsHandlerEvent::Close`].
     CloseConnection {
         /// The peer to disconnect.
         peer_id: PeerId,
-        /// The ID of the connection whose `ProtocolsHandler` to disconnect.
-        handler: CloseConnection,
+        /// Whether to close a specific or all connections to the given peer.
+        connection: CloseConnection,
     }
 }
 
@@ -322,9 +330,8 @@ impl<TInEvent, TOutEvent> NetworkBehaviourAction<TInEvent, TOutEvent> {
                 },
             NetworkBehaviourAction::ReportObservedAddr { address, score } =>
                 NetworkBehaviourAction::ReportObservedAddr { address, score },
-            NetworkBehaviourAction::CloseConnection { peer_id, handler } => {
-                NetworkBehaviourAction::CloseConnection { peer_id, handler }
-            }
+            NetworkBehaviourAction::CloseConnection { peer_id, connection } =>
+                NetworkBehaviourAction::CloseConnection { peer_id, connection }
         }
     }
 
@@ -341,8 +348,8 @@ impl<TInEvent, TOutEvent> NetworkBehaviourAction<TInEvent, TOutEvent> {
                 NetworkBehaviourAction::NotifyHandler { peer_id, handler, event },
             NetworkBehaviourAction::ReportObservedAddr { address, score } =>
                 NetworkBehaviourAction::ReportObservedAddr { address, score },
-            NetworkBehaviourAction::CloseConnection { peer_id, handler } =>
-                NetworkBehaviourAction::CloseConnection { peer_id, handler }
+            NetworkBehaviourAction::CloseConnection { peer_id, connection } =>
+                NetworkBehaviourAction::CloseConnection { peer_id, connection }
         }
     }
 }
@@ -388,7 +395,7 @@ impl Default for DialPeerCondition {
     }
 }
 
-/// The options which connection handlers to disconnect.
+/// The options which connections to close.
 #[derive(Debug, Clone)]
 pub enum CloseConnection {
     /// Disconnect a particular connection.
--- a/swarm/src/behaviour.rs
+++ b/swarm/src/behaviour.rs
@@ -294,13 +294,13 @@ pub enum NetworkBehaviourAction<TInEvent, TOutEvent> {
         score: AddressScore,
     },
 
-    /// Instructs the `Swarm` to initiate a graceful close of the connection
-    /// with a peer.
-    DisconnectPeer {
+    /// Instructs the `Swarm` to initiate a graceful close of one or all connections
+    /// with the given peer.
+    CloseConnection {
         /// The peer to disconnect.
         peer_id: PeerId,
         /// The ID of the connection whose `ProtocolsHandler` to disconnect.
-        handler: DisconnectPeerHandler,
+        handler: CloseConnection,
     }
 }
 
@@ -322,8 +322,8 @@ impl<TInEvent, TOutEvent> NetworkBehaviourAction<TInEvent, TOutEvent> {
                 },
             NetworkBehaviourAction::ReportObservedAddr { address, score } =>
                 NetworkBehaviourAction::ReportObservedAddr { address, score },
-            NetworkBehaviourAction::DisconnectPeer { peer_id, handler } => {
-                NetworkBehaviourAction::DisconnectPeer { peer_id, handler }
+            NetworkBehaviourAction::CloseConnection { peer_id, handler } => {
+                NetworkBehaviourAction::CloseConnection { peer_id, handler }
             }
         }
     }
@@ -341,8 +341,8 @@ impl<TInEvent, TOutEvent> NetworkBehaviourAction<TInEvent, TOutEvent> {
                 NetworkBehaviourAction::NotifyHandler { peer_id, handler, event },
             NetworkBehaviourAction::ReportObservedAddr { address, score } =>
                 NetworkBehaviourAction::ReportObservedAddr { address, score },
-            NetworkBehaviourAction::DisconnectPeer { peer_id, handler } =>
-                NetworkBehaviourAction::DisconnectPeer { peer_id, handler }
+            NetworkBehaviourAction::CloseConnection { peer_id, handler } =>
+                NetworkBehaviourAction::CloseConnection { peer_id, handler }
         }
     }
 }
@@ -390,15 +390,15 @@ impl Default for DialPeerCondition {
 
 /// The options which connection handlers to disconnect.
 #[derive(Debug, Clone)]
-pub enum DisconnectPeerHandler {
-    /// Disconnect a particular connection handler.
+pub enum CloseConnection {
+    /// Disconnect a particular connection.
     One(ConnectionId),
-    /// Disconnect all connection handlers.
+    /// Disconnect all connections.
     All,
 }
 
-impl Default for DisconnectPeerHandler {
+impl Default for CloseConnection {
     fn default() -> Self {
-        DisconnectPeerHandler::All
+        CloseConnection::All
     }
 }
--- a/swarm/src/behaviour.rs
+++ b/swarm/src/behaviour.rs
@@ -293,6 +293,15 @@ pub enum NetworkBehaviourAction<TInEvent, TOutEvent> {
         /// relative to other observed addresses.
         score: AddressScore,
     },
+
+    /// Instructs the `Swarm` to initiate a graceful close of the connection
+    /// with a peer.
+    DisconnectPeer {
+        /// The peer to disconnect.
+        peer_id: PeerId,
+        /// The ID of the connection whose `ProtocolsHandler` to disconnect.
+        handler: DisconnectPeerHandler,
+    }
 }
 
 impl<TInEvent, TOutEvent> NetworkBehaviourAction<TInEvent, TOutEvent> {
@@ -312,7 +321,10 @@ impl<TInEvent, TOutEvent> NetworkBehaviourAction<TInEvent, TOutEvent> {
                     event: f(event)
                 },
             NetworkBehaviourAction::ReportObservedAddr { address, score } =>
-                NetworkBehaviourAction::ReportObservedAddr { address, score }
+                NetworkBehaviourAction::ReportObservedAddr { address, score },
+            NetworkBehaviourAction::DisconnectPeer { peer_id, handler } => {
+                NetworkBehaviourAction::DisconnectPeer { peer_id, handler }
+            }
         }
     }
 
@@ -328,7 +340,9 @@ impl<TInEvent, TOutEvent> NetworkBehaviourAction<TInEvent, TOutEvent> {
             NetworkBehaviourAction::NotifyHandler { peer_id, handler, event } =>
                 NetworkBehaviourAction::NotifyHandler { peer_id, handler, event },
             NetworkBehaviourAction::ReportObservedAddr { address, score } =>
-                NetworkBehaviourAction::ReportObservedAddr { address, score }
+                NetworkBehaviourAction::ReportObservedAddr { address, score },
+            NetworkBehaviourAction::DisconnectPeer { peer_id, handler } =>
+                NetworkBehaviourAction::DisconnectPeer { peer_id, handler }
         }
     }
 }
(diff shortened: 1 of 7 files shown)


swarm/src/behaviour.rs
```diff
@@ -293,6 +293,15 @@ pub enum NetworkBehaviourAction<TInEvent, TOutEvent> {
         /// relative to other observed addresses.
         score: AddressScore,
     },
+
+    /// Instructs the `Swarm` to initiate a graceful close of the connection
+    /// with a peer.
+    DisconnectPeer {
+        /// The peer to disconnect.
+        peer_id: PeerId,
+        /// The ID of the connection whose `ProtocolsHandler` to disconnect.
```
thomaseizinger, line 302: These docs seems to be slightly off given that `DisconnectPeerHandler` is not just the ID.

swarm/src/behaviour.rs
```diff
@@ -373,3 +387,18 @@ impl Default for DialPeerCondition {
         DialPeerCondition::Disconnected
     }
 }
+
+/// The options which connection handlers to disconnect.
+#[derive(Debug, Clone)]
+pub enum DisconnectPeerHandler {
```
thomaseizinger, line 393: The docs mention the term `Options`, maybe `DisconnectOptions` is a better name for this?

I'd also agree to `DisconnectPeer` but `DisconnectPeerHandler` is a bit weird. In the end, we are disconnecting the actual peer, not the handler.

thomaseizinger (approved):
I had a need for `Swarm::disconnect_peer` the other day and was surprised that it doesn't exist. I am in favor of this!

Implementation looks good to me, left some minor comments!

swarm/src/behaviour.rs
sergeyboyko0791: I suppose someone would like to disconnect a specified connection by its `ConnectionId`. `DisconnectPeer` may confuse in this case.
> In the end, we are disconnecting the actual peer, not the handler.

I tried to follow the style of naming structures and enumerations that you follow :)
For example, [here](https://github.com/libp2p/rust-libp2p/blob/master/swarm/src/behaviour.rs#L338).
> The docs mention the term Options, maybe DisconnectOptions is a better name for this?

swarm/src/behaviour.rs
sergeyboyko0791: I see, thank you!

swarm/src/behaviour.rs
thomaseizinger: > I tried to follow the style of naming structures and enumerations that you follow :)

I did see that and thought you would have been inspired by `NotifyHandler` :)

What about `CloseConnection::{One,All}`? That naming would be inline with [`ProtocolsHandlerEvent::Close`](https://github.com/libp2p/rust-libp2p/blob/e8fed53598696a45a26866408534cfa186b23d4a/swarm/src/protocols_handler.rs#L322) although that rename would affect the overall terminology used in PR.

mxinden (commented):
First off, thanks @sergeyboyko0791 for the extensive pull request.

In general, I am not opposed to these changes.

I would like to use this opportunity to give users more guidance on which mechanism to use, now that we are adding yet another one. I guess guidance would be best given via additional documentation.

Today, as a user of `libp2p-swarm`, I see two ways to close a connection to a peer:

1. Via `ProtocolsHandler::keep_alive` in a collaborative manner across protocol handlers.
2. Via `ProtocolsHandlerEvent::Close` in a direct manner, instantly closing the connection whether used by other handlers or not.

Being able to close a connection from a `NetworkBehaviour` by emitting a `NetworkBehaviourAction` is a shortcut for (1). While I see this to be useful for people wanting to implement a _Connection Manager_ via a `NetworkBehaviour`, I don't think this non-collaborative style of closing connections is suitable for other protocols like `libp2p-kad` or `libp2p-gossipsub`.

Now, in order to be able to give users guidance on which mechanism to use when, I would like to understand your use-case @sergeyboyko0791. Would you mind expanding on why you need to close a connection from a `NetworkBehaviour`?

> I had a need for Swarm::disconnect_peer the other day and was surprised that it doesn't exist. I am in favor of this!

To get more data points, would you mind expanding on your use-case as well @thomaseizinger?

>  Please look at the example of using NetworkBehaviourAction::DisconnectPeer': https://github.com/KomodoPlatform/atomicDEX-API/blob/mm2.1/mm2src/mm2_libp2p/src/adex_ping.rs#L27

With the default `PingConfig`, `libp2p-ping` should close the connection on the second failure. Is that not the case for you? If so, would you mind either debugging this yourself or providing a reproducer?

thomaseizinger:
> To get more data points, would you mind expanding on your use-case as well @thomaseizinger?

My usecase were the examples for the `rendezvous` protocol. I wanted to showcase an application that:

- connects to a rendezvous point
- registers with the rendezvous point
- disconnects from it
- waits for incoming connections from other peers

To do this, I wanted an API on `ExpandedSwarm` analogous to `dial_{peer, address}` where I can disconnect from a specific peer.

For what it is worth, I think neither of the two existing approaches work very well here because the decision to disconnect is made on an application level. The individual `NetworkBehaviour`s don't have enough knowledge to handle this.

sergeyboyko0791:
Hi @mxinden, thanks for your reply and suggestions.

> With the default PingConfig, libp2p-ping should close the connection on the second failure. Is that not the case for you? If so, would you mind either debugging this yourself or providing a reproducer?

To be honest, I have not completely figured out how `Ping` works, and now I agree that it's not the use case to disconnect a peer manually from the behaviour.

> Would you mind expanding on why you need to close a connection from a NetworkBehaviour?

We have one more example where we disconnect a peer using `ExpandedSwarm::disconnect_peer_id` instead of `NetworkBehaviourAction::DisconnectPeer`, but this logic can be moved to a behaviour:

We maintain the max number of connected relays regarding the `GossipsubConfig::mesh_n_high` value on every swarm poll. In theory, if this value was decreased externally (for example, on SIGHUP), we have to disconnect the extra relays.

Please, look at [this](https://github.com/KomodoPlatform/atomicDEX-API/blob/beta-2.1.3731/mm2src/mm2_libp2p/src/atomicdex_behaviour.rs#L500).

swarm/src/lib.rs
```diff
@@ -463,6 +464,18 @@ where TBehaviour: NetworkBehaviour<ProtocolsHandler = THandler>,
         self.banned_peers.remove(&peer_id);
     }
 
+    /// Disconnects a peer by its peer ID.
```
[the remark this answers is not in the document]
sergeyboyko0791: Done

swarm/src/behaviour.rs
```diff
@@ -373,3 +387,18 @@ impl Default for DialPeerCondition {
...
+    /// Disconnect a particular connection handler.
+    One(ConnectionId),
+    /// Disconnect all connection handlers.
```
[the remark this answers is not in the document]
sergeyboyko0791: I see, thanks!

swarm/src/behaviour.rs
```diff
@@ -293,6 +293,15 @@ pub enum NetworkBehaviourAction<TInEvent, TOutEvent> {
         /// relative to other observed addresses.
         score: AddressScore,
     },
+
+    /// Instructs the `Swarm` to initiate a graceful close of the connection
+    /// with a peer.
+    DisconnectPeer {
```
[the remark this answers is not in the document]
sergeyboyko0791: Agree with you and @thomaseizinger

sergeyboyko0791:
I've pushed the changes you requested. Could you please review it again?
cc @thomaseizinger @mxinden

mxinden:
> I've pushed the changes you requested. Could you please review it again?
> cc @thomaseizinger @mxinden

Thanks. I will take a look likely later today.

swarm/src/behaviour.rs
```diff
@@ -293,6 +293,15 @@ pub enum NetworkBehaviourAction<TInEvent, TOutEvent> {
...
         score: AddressScore,
     },
+
+    /// Instructs the `Swarm` to initiate a graceful close of one or all connections
+    /// with the given peer.
+    CloseConnection {
+        /// The peer to disconnect.
+        peer_id: PeerId,
+        /// The ID of the connection whose `ProtocolsHandler` to disconnect.
+        handler: CloseConnection,
```
mxinden, line 303:
```suggestion
        connection: CloseConnection,
```

swarm/src/behaviour.rs
```diff
@@ -373,3 +387,18 @@ impl Default for DialPeerCondition {
         DialPeerCondition::Disconnected
     }
 }
+
+/// The options which connection handlers to disconnect.
```
mxinden, line 391:
```suggestion
/// The options which connections to close.
```

swarm/src/behaviour.rs
```diff
@@ -293,6 +293,15 @@ pub enum NetworkBehaviourAction<TInEvent, TOutEvent> {
         /// relative to other observed addresses.
         score: AddressScore,
     },
+
+    /// Instructs the `Swarm` to initiate a graceful close of one or all connections
+    /// with the given peer.
```
mxinden, line 298:
```suggestion
    /// with the given peer.
    ///
    /// Note: Closing a connection via
    /// [`NetworkBehaviourAction::CloseConnection`] does not inform the
    /// corresponding [`ProtocolsHandler`](s). Closing a connection via a
    /// [`ProtocolsHandler`] can be done either in a collaborative manner across
    /// [`ProtocolsHandler`]s with [`ProtocolsHandler::connection_keep_alive`]
    /// or directly with [`ProtocolsHandlerEvent::Close`].
```

swarm/src/behaviour.rs
mxinden:
```suggestion
        /// Whether to close a specific or all connections to the given peer.
```

swarm/src/lib.rs
```diff
@@ -463,6 +464,18 @@ where TBehaviour: NetworkBehaviour<ProtocolsHandler = THandler>,
         self.banned_peers.remove(&peer_id);
     }
 
+    /// Disconnects a peer by its peer ID, closing all connections to said peer.
+    ///
+    /// Returns `Ok(())` if a peer with this ID was in the list.
```
mxinden, line 469:
```suggestion
    /// Returns `Ok(())` if there was one or more established connections to the peer
```

swarm/src/lib.rs
```diff
@@ -463,6 +464,18 @@ where TBehaviour: NetworkBehaviour<ProtocolsHandler = THandler>,
...
+    pub fn disconnect_peer_id(&mut self, peer_id: PeerId) -> Result<(), ()> {
```
mxinden, line 470:
```suggestion
    ///
    /// Note: Closing a connection via [`ExpandedSwarm::disconnect_peer_id`] does
    /// not inform the corresponding [`ProtocolsHandler`](s). Closing a
    /// connection via a [`ProtocolsHandler`] can be done either in a
    /// collaborative manner across [`ProtocolsHandler`]s with
    /// [`ProtocolsHandler::connection_keep_alive`] or directly with
    /// [`ProtocolsHandlerEvent::Close`].
    pub fn disconnect_peer_id(&mut self, peer_id: PeerId) -> Result<(), ()> {
```

swarm/src/lib.rs
```diff
@@ -1242,7 +1293,155 @@ mod tests {
...
+                }
+
+                if poll1.is_pending() && poll2.is_pending() {
+                    return Poll::Pending
+                }
+            }
+        }))
+    }
+
+    /// Establishes a number of connections between two peers,
```
mxinden, line 1311:
```suggestion
    /// Establishes multiple connections between two peers,
```

swarm/src/lib.rs
```diff
@@ -1242,7 +1293,155 @@ mod tests {
...
```
mxinden, line 1381:
```suggestion
    /// Establishes multiple connections between two peers,
```

swarm/src/lib.rs
```diff
@@ -1242,7 +1293,155 @@ mod tests {
...
+    /// after which one peer disconnects the other
+    /// using [`NetworkBehaviourAction::CloseConnection`] thrown from a behaviour.
```
mxinden, line 1383:
```suggestion
    /// using [`NetworkBehaviourAction::CloseConnection`] returned by a [`NetworkBehaviour`].
```

swarm/src/lib.rs
```diff
@@ -1256,4 +1455,83 @@ mod tests {
             }
         }))
     }
+
+    /// Establishes a number of connections between two peers,
```
mxinden, line 1459:
```suggestion
    /// Establishes multiple connections between two peers,
```

swarm/src/lib.rs
```diff
@@ -1256,4 +1455,83 @@ mod tests {
...
+    /// after which one peer closes the only one connection
+    /// using [`NetworkBehaviourAction::CloseConnection`] thrown from a behaviour.
```
mxinden, line 1461:
```suggestion
    /// using [`NetworkBehaviourAction::CloseConnection`] returned by a [`NetworkBehaviour`].
```

swarm/src/lib.rs
```diff
@@ -1256,4 +1455,83 @@ mod tests {
             }
         }))
     }
+
+    /// Establishes a number of connections between two peers,
+    /// after which one peer closes the only one connection
```
mxinden, line 1460:
```suggestion
    /// after which one peer closes a single connection
```

mxinden (commented):
Thanks for writing test cases for each of the 3 new ways to disconnect a peer! Very much appreciated @sergeyboyko0791.

I have a couple smaller suggestions around documentation, in order to guide users. What do you think?

Also, could you please apply the diff below adding a changelog entry:

```diff
diff --git a/swarm/CHANGELOG.md b/swarm/CHANGELOG.md
index 103f88d2..0f01b314 100644
--- a/swarm/CHANGELOG.md
+++ b/swarm/CHANGELOG.md
@@ -15,7 +15,12 @@
 
   See [PR 2100] for details.
 
+- Add `ExpandedSwarm::disconnect_peer_id` and
+  `NetworkBehaviourAction::CloseConnection` to close connections to a specific
+  peer via an `ExpandedSwarm` or `NetworkBehaviour`. See [PR 2110] for details.
+
 [PR 2100]: https://github.com/libp2p/rust-libp2p/pull/2100
+[PR 2110]: https://github.com/libp2p/rust-libp2p/pull/2110/
```

sergeyboyko0791:
@mxinden, all done.
I just formatted the comments.
Thanks for the detailed review :upside_down_face:

swarm/src/behaviour.rs
```diff
@@ -293,6 +293,23 @@ pub enum NetworkBehaviourAction<TInEvent, TOutEvent> {
...
+    ///
+    /// Note: Closing a connection via
+    /// [`NetworkBehaviourAction::CloseConnection`] does not inform the
+    /// corresponding [`ProtocolsHandler`](s).
+    /// Closing a connection via a [`ProtocolsHandler`] can be done
+    /// either in a collaborative manner across [`ProtocolsHandler`]s
+    /// with [`ProtocolsHandler::connection_keep_alive`] or directly with
+    /// [`ProtocolsHandlerEvent::Close`].
```
mxinden, line 306:
```suggestion
    /// [`ProtocolsHandlerEvent::Close`](crate::ProtocolsHandlerEvent::Close).
```

swarm/src/behaviour.rs
```diff
@@ -293,6 +293,23 @@ pub enum NetworkBehaviourAction<TInEvent, TOutEvent> {
         /// relative to other observed addresses.
         score: AddressScore,
     },
+
+    /// Instructs the `Swarm` to initiate a graceful close of one or all connections
+    /// with the given peer.
+    ///
+    /// Note: Closing a connection via
+    /// [`NetworkBehaviourAction::CloseConnection`] does not inform the
+    /// corresponding [`ProtocolsHandler`](s).
```
mxinden, line 302:
```suggestion
    /// corresponding [`ProtocolsHandler`].
```

swarm/src/lib.rs
```diff
@@ -463,6 +464,25 @@ where TBehaviour: NetworkBehaviour<ProtocolsHandler = THandler>,
         self.banned_peers.remove(&peer_id);
     }
 
+    /// Disconnects a peer by its peer ID, closing all connections to said peer.
+    ///
+    /// Returns `Ok(())` if there was one or more established connections to the peer.
+    ///
+    /// Note: Closing a connection via [`ExpandedSwarm::disconnect_peer_id`] does
+    /// not inform the corresponding [`ProtocolsHandler`](s).
```
mxinden, line 472:
```suggestion
    /// not inform the corresponding [`ProtocolsHandler`].
```

sergeyboyko0791:
I almost prepared the fix :D

mxinden (approved):

sergeyboyko0791:
Thank you @mxinden @thomaseizinger for the review!

PR #2111, opened 2021-06-25
Redesign `upgrade::{write_one, write_with_len_prefix, read_one}`

thomaseizinger:
This PR re-designs the above three functions to:

- allow for better composition
- remove footguns
- make them easier to use

We achieve this by:

1. Deprecating the `write_one` function

Semantically, this function is a composition of `write_with_len_prefix` and `io.close()`. This represents a footgun because the `close` functionality is not obvious and only mentioned in the docs. Using this function multiple times on a single substream will produces hard to debug behaviour.

2. Deprecating `read_one` and `write_with_len_prefix` functions

3. Introducing `write_length_prefixed` and `read_length_prefixed`

- These functions are symmetric and do exactly what you would expect, just writing to the socket without closing
- They also have a symmetric interface (no more custom errors, just `io::Error`)

### DIFF
--- a/core/src/upgrade/transfer.rs
+++ b/core/src/upgrade/transfer.rs
@@ -28,7 +28,7 @@ use std::{error, fmt, io};
 /// Writes a message to the given socket with a length prefix appended to it. Also flushes the socket.
 ///
 /// > **Note**: Prepends a variable-length prefix indicate the length of the message. This is
-/// >           compatible with what `read_one` expects.
+/// >           compatible with what [`read_length_prefixed`] expects.
 pub async fn write_length_prefixed(socket: &mut (impl AsyncWrite + Unpin), data: impl AsRef<[u8]>)
     -> Result<(), io::Error>
 {
--- a/core/src/upgrade/transfer.rs
+++ b/core/src/upgrade/transfer.rs
@@ -25,29 +25,17 @@ use std::{error, fmt, io};
 
 // TODO: these methods could be on an Ext trait to AsyncWrite
 
-/// Send a message to the given socket, then shuts down the writing side.
+/// Writes a message to the given socket with a length prefix appended to it. Also flushes the socket.
 ///
 /// > **Note**: Prepends a variable-length prefix indicate the length of the message. This is
 /// >           compatible with what `read_one` expects.
-pub async fn write_and_close(socket: &mut (impl AsyncWrite + Unpin), data: impl AsRef<[u8]>)
-    -> Result<(), io::Error>
-{
-    write_varint(socket, data.as_ref().len()).await?;
-    socket.write_all(data.as_ref()).await?;
-    socket.close().await?;
-    Ok(())
-}
-
-/// Send a message to the given socket with a length prefix appended to it. Also flushes the socket.
-///
-/// > **Note**: Prepends a variable-length prefix indicate the length of the message. This is
-/// >           compatible with what `read_one` expects.
-pub async fn write_with_len_prefix(socket: &mut (impl AsyncWrite + Unpin), data: impl AsRef<[u8]>)
+pub async fn write_length_prefixed(socket: &mut (impl AsyncWrite + Unpin), data: impl AsRef<[u8]>)
     -> Result<(), io::Error>
 {
     write_varint(socket, data.as_ref().len()).await?;
     socket.write_all(data.as_ref()).await?;
     socket.flush().await?;
+
     Ok(())
 }
 
@@ -60,6 +48,7 @@ pub async fn write_varint(socket: &mut (impl AsyncWrite + Unpin), len: usize)
     let mut len_data = unsigned_varint::encode::usize_buffer();
     let encoded_len = unsigned_varint::encode::usize(len, &mut len_data).len();
     socket.write_all(&len_data[..encoded_len]).await?;
+
     Ok(())
 }
 
@@ -113,8 +102,8 @@ pub async fn read_varint(socket: &mut (impl AsyncRead + Unpin)) -> Result<usize,
 /// gigabytes.
 ///
 /// > **Note**: Assumes that a variable-length prefix indicates the length of the message. This is
-/// >           compatible with what `write_one` does.
-pub async fn read_one(socket: &mut (impl AsyncRead + Unpin), max_size: usize)
+/// >           compatible with what [`write_length_prefixed`] does.
+pub async fn read_length_prefixed(socket: &mut (impl AsyncRead + Unpin), max_size: usize)
     -> Result<Vec<u8>, ReadOneError>
 {
     let len = read_varint(socket).await?;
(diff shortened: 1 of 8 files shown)


mxinden (commented):
Would you mind including an entry in `core/CHANGELOG.md` as well?

core/src/upgrade/transfer.rs
```diff
@@ -29,7 +29,7 @@ use std::{error, fmt, io};
 ///
 /// > **Note**: Prepends a variable-length prefix indicate the length of the message. This is
 /// >           compatible with what `read_one` expects.
-pub async fn write_one(socket: &mut (impl AsyncWrite + Unpin), data: impl AsRef<[u8]>)
+pub async fn write_and_close(socket: &mut (impl AsyncWrite + Unpin), data: impl AsRef<[u8]>)
```
mxinden, line 32:
```suggestion
pub async fn write_with_len_prefix_and_close(socket: &mut (impl AsyncWrite + Unpin), data: impl AsRef<[u8]>)
```

For the sake of consistency with the below `fn` `write_with_len_prefix`, wouldn't this naming be better?

In the same spirit, what do you think of:

```diff
- pub async fn read_one(socket: &mut (impl AsyncRead + Unpin), max_size: usize)
+ pub async fn read_with_len_prefix(socket: &mut (impl AsyncRead + Unpin), max_size: usize)
```

thomaseizinger: So actually, I would want to remove the `with_len_prefix` from all functions because that piece of logic is integral to all of them and hence not really worth pointing out IMO.

The issue with that is that we would effectively change `write_one` for `write_with_len_prefix` assuming we also keep the rename to `write_one_and_close`.

This discussion makes me think though: Usually when it is hard to put multiple aspects into a single name, the composition of the function might not be ideal. Closing a substream is a single function call. What if we remove the functionality of closing entirely and only offer one function that just does the encoding?

thomaseizinger: Possible names: `encode_one` and `decode_one`.

mxinden: I find `encode_*` misleading.

Renaming `write_with_len_prefix` to `write_one` would mean a breaking change without a change at the type level. I would prefer to not do that.

How about removing the current `write_one` and retain `write_with_len_prefix`. Or maybe rename to `write_length_prefixed`?

thomaseizinger: > I find `encode_*` misleading.

Out of curiosity, can you expand on that?
What would be your expectation when you see a function signature of:

```rust
async fn encode_one(stream: &mut impl AsyncWrite, message: &impl AsRef<[u8]>);
```

And how is that expectation different from what the function will be doing?

My thinking was: Serializing to bytes and prefixing with the varint length _is_ a form of encoding the message. We could have the function return a Vec instead of trying to write to the stream directly but that is just unnecessary copying.


> Renaming `write_with_len_prefix` to `write_one` would mean a breaking change without a change at the type level. I would prefer to not do that.

Yep agreed. I tried to say that in my last comment but you expressed it much better 😅

> How about removing the current `write_one` and retain `write_with_len_prefix`. Or maybe rename to `write_length_prefixed`?

Happy with `write_length_prefixed`. Will adapt code accordingly!

thomaseizinger:
While we are at it with breaking the interface, I've always found it a bit annoying that `read_length_prefixed` returns a dedicated error and not just an `io::Error`. Do you think we could represent the `TooLarge` variant as one of the representations of `io::Error`?

core/src/upgrade/transfer.rs
mxinden: > And how is that expectation different from what the function will be doing?
>
> My thinking was: Serializing to bytes and prefixing with the varint length is a form of encoding the message. We could have the function return a Vec instead of trying to write to the stream directly but that is just unnecessary copying.

More of a gut feeling. _Encoding_ in my head changes the data itself. _Writing_ simply takes the bytes and shoves them down the wire. Neither seems to be a perfect fit here, though _writing_ in my eyes is closest, given that the length is only prefixed but the actual data is untouched.

mxinden (commented):
> While we are at it with breaking the interface, I've always found it a bit annoying that read_length_prefixed returns a dedicated error and not just an io::Error. Do you think we could represent the TooLarge variant as one of the representations of io::Error?

How about `ErrorKind::InvalidData`? If you decide to apply this change, can you document the `ErrorKind::InvalidData` return value?

thomaseizinger:
@mxinden While writing migration docs, I thought it would actually be much better to just not break the API right away and instead deprecate the old functions. The deprecation note should hopefully guide people towards the new functions.

I removed `ReadOneError` from the new `read_length_prefixed` function and returned a custom `io::Error` with a dedicated error message. Do you still want me to document that as part of the functions rustdoc? I think the error message is good enough that it should lead people towards the problem if they ever hit it.

mxinden (approved):
Looks good to me. Thanks for thorough patch.

I don't quite understand the benefit of deprecating the 3 methods, given that users will have to update at some point anyways and updating should be very low friction. That said, I am fine either way.

thomaseizinger:
> Looks good to me. Thanks for thorough patch.
> 
> I don't quite understand the benefit of deprecating the 3 methods, given that users will have to update at some point anyways and updating should be very low friction. That said, I am fine either way.

> Looks good to me. Thanks for thorough patch.
> 
> I don't quite understand the benefit of deprecating the 3 methods, given that users will have to update at some point anyways and updating should be very low friction. That said, I am fine either way.

My thinking was, if someone else made this change and I was on the receiving end of updating the library, how would I want to learn about the change?

Having code still work after an update is preferable, even if it exposes deprecated APIs. I can slap an `allow(deprecated)` on it if I don't want to bother with it or also live with the warning.

The other thing is, the deprecation warning is going to lead me straight to the new functionality without having dig through the changelog (in the case of rust-libp2p multiple changelogs 😅) for the relevant section.

protocols/floodsub/src/protocol.rs
```diff
@@ -95,15 +95,15 @@ where
 #[derive(Debug)]
 pub enum FloodsubDecodeError {
     /// Error when reading the packet from the socket.
-    ReadError(upgrade::ReadOneError),
+    ReadError(io::Error),
```
thomaseizinger, line 98: This is a breaking change that needs to be mentioned in the changelog and needs a changelog entry!

thomaseizinger (commented):
@mxinden Noticed another thing, please hold on with merging.

protocols/floodsub/CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 - Update dependencies.
 
+- Change `FloodsubDecodeError::ReadError` from a `upgrade::ReadOneError` to `std::io::Error`
```
mxinden, line 5:
```suggestion
- Change `FloodsubDecodeError::ReadError` from a `upgrade::ReadOneError` to
  `std::io::Error`. See [PR 2111].

[PR 2111]: https://github.com/libp2p/rust-libp2p/pull/2111
```

Issue #2064, opened 2021-05-04
DNS transport not working on Android


wngr:
#1927 changed the DNS transport to use the [trust_dns_resolver](https://docs.rs/trust-dns-resolver/0.20.2/trust_dns_resolver/) crate rather than the system's libc functions. This means, the DNS resolution is completely in process now. The [standard configuration](https://docs.rs/trust-dns-resolver/0.20.2/trust_dns_resolver/struct.Resolver.html#method.from_system_conf) tries to read `/etc/resolv.conf` to parse the host's DNS settings and use them.
This doesn't work on Android (see https://github.com/bluejekyll/trust-dns/issues/652). Supporting Android looks like a can of worms (https://github.com/termux/termux-packages/issues/1174), which is probably why that issue is also stalled.

I think rust-libp2p should offer a fallback to use the system resolver via libc and bring back the old code -- what do you say?

mxinden:
First off, rust-libp2p supporting Android would be neat! Mind sharing what you are up to?

> The standard configuration tries to read /etc/resolv.conf to parse the host's DNS settings and use them.

Would using a custom [`ResolverConfig`](https://docs.rs/trust-dns-resolver/0.20.2/trust_dns_resolver/config/struct.ResolverConfig.html) be an option for you?

wngr:
[Actyx](https://github.com/Actyx/) has been running rust-libp2p in production for quite some time now on different hosts, namely Linux, Windows, and Android. We're providing a p2p platform for building and running resilient apps in factories.

> Would using a custom ResolverConfig be an option for you?

No, as we want to rely on the system resolver (so `gethostbyname` et al), because the devices our software runs on are mostly administered by corporate IT, and they're usually a bit picky about their network setup ;-). On Linux that's fine, as the system dns config is picked up via `/etc/resolv.conf` (hopefully ..), but not Android..

mxinden:
Puuh, this (https://github.com/termux/termux-packages/issues/1174) indeed seems to be a can of worms.

> I think rust-libp2p should offer a fallback to use the system resolver via libc and bring back the old code -- what do you say?

I am a bit careful introducing yet another option to `libp2p-dns`. That said, I think it is worth exploring to see how intrusive offering that fallback would be.

wngr:
[...]

PR #2114, opened 2021-06-30
transports/dns: Document caveats of using the `trust-dns-resolver` crate

wngr:
Closes https://github.com/libp2p/rust-libp2p/issues/2064


mxinden (approved):
Very much appreciated. Thanks @wngr!

Issue #2092, opened 2021-06-04
protocols/kad: debug_assert panic - expecting disconnected peer not to be connected


mxinden:
I triggered the panic below with https://github.com/mxinden/rust-libp2p-server/. Unable to reproduce thus far.

```
thread 'main' panicked at 'assertion failed: !self.connected_peers.contains(disconnected.preimage())', /home/mxinden/.cargo/git/checkouts/rust-libp2p-800b9a7842b0accc/6b2873b/protocols/kad/src/behaviour.rs:1033:33
```

https://github.com/libp2p/rust-libp2p/blob/3b0f5a4f9630a6f51c277843bf669c1b30e2f74b/protocols/kad/src/behaviour.rs#L1024-L1025

Is the `debug_assert` to strict?

FelipeRosa:
I'm currently working on a personal project and this one pops quit often for me. I'm a bit new to libp2p in general, why is that `debug_assert` needed?

mxinden:
> I'm currently working on a personal project

:tada: 

Thanks for following up here @FelipeRosa.

> why is that debug_assert needed?

It is supposed to uphold that there is no state mismatch between routing table and connected peers tracked in the Kademlia `NetworkBehaviour`.

I haven't gotten around to further debugging this. In case you are able to reproduce this reliably, could you share a minimal reproducible example here @FelipeRosa?

FelipeRosa:
Hey @mxinden,

Thanks for the quick reply.

> I haven't gotten around to further debugging this. In case you are able to reproduce this reliably, could you share a minimal reproducible example here @FelipeRosa?

I'm not able to reproduce this issue consistently. I read a bit of the libp2p_kad code to try to understand where this might be coming from but I didn't quite wrap my head around it yet.

What I did try to do is create a little program to keep generating hashes that would fall into buckets with available space in them and then making `get_closest_peers` calls sequentially. 

Something like this but with some code to wait for queries to finish:
```rust
let rand_bytes: [u8; 32] = rng.gen();
let ma = libp2p::multihash::Sha2_256::digest(&rand_bytes).as_ref().to_vec();

if let Some(bucket) = swarm.behaviour_mut().kbucket(ma.clone()) {
    if !bucket.has_pending() {
        swarm.behaviour_mut().get_closest_peers(ma);
    }
}
```

My guess is that by doing this I would keep having peers connecting and disconnecting to generate as many routing table states as possible in order to reach the one which causes this bug.

[...]

PR #2120, opened 2021-07-05
protocols/kad/behaviour: Remove false assert on connected_peers

mxinden:
Given the following scenario:

1. Remote peer X connects and is added to `connected_peers`.
2. Remote peer X opens a Kademlia substream and thus confirms that it supports
   the Kademlia protocol.
3. Remote peer X is added to the routing table as `Connected`.
4. Remote peer X disconnects and is thus marked as `Disconnected` in the routing
   table.
5. Remote peer Y connects and is added to `connected_peers`.
6. Remote peer X re-connects and is added to `connected_peers`.
7. Remote peer Y opens a Kademlia substream and thus confirms that it supports
   the Kademlia protocol.
8. Remote peer Y is added to the routing table. Given that the bucket is already
   full the call to `entry.insert` returns `kbucket::InsertResult::Pending {
   disconnected }` where disconnected is peer X.

While peer X is in `connected_peers` it has not yet (re-) confirmed that it
supports the Kademlia routing protocol and thus is still tracked as
`Disconnected` in the routing table. The `debug_assert` removed in this pull
request does not capture this scenario.

Fixes https://github.com/libp2p/rust-libp2p/issues/2092

@FelipeRosa @izolyomi would you mind taking a look and confirm that this fixes your issues as well?

### DIFF
--- a/protocols/kad/CHANGELOG.md
+++ b/protocols/kad/CHANGELOG.md
@@ -11,7 +11,10 @@
 
 - Expose kbucket range on `KademliaEvent::RoutingUpdated` (see [PR 2087]).
 
+- Remove false `debug_assert` on `connected_peers` (see [PR 2120]).
+
 [PR 2087]: https://github.com/libp2p/rust-libp2p/pull/2087
+[PR 2120]: https://github.com/libp2p/rust-libp2p/pull/2120
 
 # 0.30.0 [2021-04-13]
 
--- a/protocols/kad/src/behaviour.rs
+++ b/protocols/kad/src/behaviour.rs
@@ -1046,10 +1046,19 @@ where
                                 self.queued_events.push_back(NetworkBehaviourAction::GenerateEvent(
                                     KademliaEvent::PendingRoutablePeer { peer, address }
                                 ));
-                                self.queued_events.push_back(NetworkBehaviourAction::DialPeer {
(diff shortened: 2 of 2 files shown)


romanb (approved):
Makes sense to me. I guess this was an oversight when implementing the delayed insertion into the routing tables (i.e. only after protocol confirmation).

izolyomi:
Using this new dependency version 0.39 seems to have a ton of breaking changes compared to 0.38. I'd have to understand all the changes and fix everything to test, I might stick with the old version for the moment.

FelipeRosa:
Thanks for the work @mxinden and @romanb 

Sure, I'll take a look at it later today.

However, I believe that should solve the problem because what I ended up doing to prevent triggering the `debug_assert` was setting Kademlia's behaviour kbuckets insert to `KademliaBucketInserts::Manual` and adding DHT peer addresses manually at a later stage (after the peer has sent Identify information and said it supports the kad protocol).

Since I've done that I have not seen the `debug_assert` being triggered and everything seems to be working fine when using the kad DHT.

izolyomi:
I've finally upgraded to the unreleased 0.39, it wasn't too hard to fix all breaking changes after all. I can confirm that I haven't encountered `debug_assert` failing even once.

mxinden:
Thanks everyone for the input and thanks @romanb for the review!

PR #2126, opened 2021-07-09
*: patch prost / prost-build version to v0.8

elenaf9:
Update `prost` and `prost-build` from v0.7 to v0.8. 
`prost-build` v0.7 depends on outdated version of `prost-types` with the issue https://github.com/tokio-rs/prost/issues/438:
>  Conversion from prost_types::Timestamp to SystemTime can cause an overflow and panic

cc: https://github.com/iotaledger/stronghold.rs/issues/228


mxinden (approved):
Thanks @elenaf9!

PR #2135, opened 2021-07-11
fix: Replace std::time::Instant by wasm_timer::Instant

### LINKED ISSUE #2134: WASM regression: Usage of std::time::Instant
While creating my custom node based on rust-libp2p, I rediscovered some uses of `std::time::Instant` in the Gossipsub and Kadmelia modules. As previously done by the PR #1071 (issue #900), the remaining `std::time::Instant` should again be replaced with `wasm_timer::Instant` to allow Wasm compatibility.

The spotted ones:
- [ ] https://github.com/libp2p/rust-libp2p/blob/ee11922d978834c0d5180eb8bc9e482f7945ecfa/protocols/gossipsub/src/time_cache.rs#L29
- [ ] https://github.com/libp2p/rust-libp2p/blob/ee11922d978834c0d5180eb8bc9e482f7945ecfa/protocols/gossipsub/src/peer_score.rs#L30
- [ ] https://github.com/libp2p/rust-libp2p/blob/ee11922d978834c0d5180eb8bc9e482f7945ecfa/protocols/mdns/src/behaviour.rs#L45
- [ ] https://github.com/libp2p/rust-libp2p/blob/ee11922d978834c0d5180eb8bc9e482f7945ecfa/protocols/kad/src/kbucket.rs#L80

fusetim:
Closes #2134
Needs https://github.com/tomaka/wasm-timer/pull/17

* Just replaced the imports of `std::time::Instant` with `wasm-timer::Instant` for WASM support.
* Checked the existing uses of Instant, there is no dependency outside of the crate itself. All uses were never publicly exposed.
* I can't run the tests on my PC, it became really unstable when I tried but it fully compiled (even on wasm32 target using the https://github.com/tomaka/wasm-timer/pull/17).

### DIFF
--- a/protocols/gossipsub/src/peer_score.rs
+++ b/protocols/gossipsub/src/peer_score.rs
@@ -27,7 +27,8 @@ use libp2p_core::PeerId;
 use log::{debug, trace, warn};
 use std::collections::{hash_map, HashMap, HashSet};
 use std::net::IpAddr;
-use std::time::{Duration, Instant};
+use std::time::Duration;
+use wasm_timer::Instant;
 
 mod params;
 use crate::error::ValidationError;
--- a/protocols/gossipsub/src/time_cache.rs
+++ b/protocols/gossipsub/src/time_cache.rs
@@ -26,7 +26,9 @@ use std::collections::hash_map::{
     Entry::{Occupied, Vacant},
 };
 use std::collections::VecDeque;
-use std::time::{Duration, Instant};
+use std::time::Duration;
+use wasm_timer::Instant;
+
 
 struct ExpiringElement<Element> {
     /// The element that expires
--- a/protocols/kad/src/kbucket.rs
+++ b/protocols/kad/src/kbucket.rs
@@ -77,7 +77,8 @@ pub use entry::*;
 use arrayvec::{self, ArrayVec};
 use bucket::KBucket;
 use std::collections::VecDeque;
-use std::time::{Duration, Instant};
+use std::time::Duration;
+use wasm_timer::Instant;
 
 /// Maximum number of k-buckets.
 const NUM_BUCKETS: usize = 256;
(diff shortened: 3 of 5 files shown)


dvc94ch:
mdns won't work in a browser as it requires access to a udp socket. can you tell us more about what platform you're trying to run on?

fusetim:
Yes you are right, in the case of mDNS it doesn't change anything, nevertheless we can think that some people could see the utility of using it with Node.js and in this case, Node.js has the capacity to open UDP sockets.

But in my case, it's about creating a custom node with a future protocol that I have to create, based on rust-libp2p and to be compiled with the target wasm32-unknown-unknown. It should run in the browser with some wasm-bindgen stuff.

I encountered an error with Gossipsub (when verifying messages) because it was using `std::time::Instant::now()` which is not implemented on the `wasm32-unknown-unknown` target today. In fact, I quickly went around to fix these imports and now it works perfectly following the changes made by this PR.

mxinden (commented):
In general, using `wasm_timer` across the project sounds good to me.

> Yes you are right, in the case of mDNS it doesn't change anything, nevertheless we can think that some people could see the utility of using it with Node.js and in this case, Node.js has the capacity to open UDP sockets.

I would expect there to be additional blockers when using `libp2p-mdns` on Node.js. That said, I don't think it hurts to use `wasm_timer` in `libp2p-mdns` as well.

I left a comment on your `wasm_timer` pull request. Once that pull request is merged and released we can continue here.

Thanks for creating the patch @fusetim.

dvc94ch:
believe this one was fixed by @wngr

thomaseizinger:
> believe this one was fixed by @wngr

Yep, thanks for the ping. Closing this.

fusetim:
For tracking, fixed by #2245. Thanks.

PR #2137, opened 2021-07-14
Add a total established connection limit

AgeManning:
I think it would be useful to be able to specify a total established connection limit, rather than just incoming and outgoing limits. 

If a user wants to set an incoming limit of 10 and an outgoing limit of 5, currently, this has a maximum connection limit of 15, but the ratio of incoming to outgoing is fixed. If a user wishes to specify the total and is happy to have a variable ratio of incoming/outgoing, I think this PR is useful to enable this at the `Network` level. 

This shouldn't break anything as its an optional addition to the `ConnectionLimit` struct.

### DIFF
--- a/core/CHANGELOG.md
+++ b/core/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 - Add `ConnectionLimit::with_max_established` (see [PR 2137]).
 
+[PR 2137]: https://github.com/libp2p/rust-libp2p/pull/2137/
+
 - Change `PublicKey::into_protobuf_encoding` to `PublicKey::to_protobuf_encoding` (see [PR 2145]).
 
 - Change `PublicKey::into_peer_id` to `PublicKey::to_peer_id` (see [PR 2145]).
--- a/core/CHANGELOG.md
+++ b/core/CHANGELOG.md
@@ -1,5 +1,7 @@
 # 0.30.0 [unreleased]
 
+- Add `ConnectionLimit::with_max_established` (see [PR 2137]).
+
 - Change `PublicKey::into_protobuf_encoding` to `PublicKey::to_protobuf_encoding` (see [PR 2145]).
 
 - Change `PublicKey::into_peer_id` to `PublicKey::to_peer_id` (see [PR 2145]).
--- a/core/src/connection/pool.rs
+++ b/core/src/connection/pool.rs
@@ -987,6 +987,9 @@ impl ConnectionCounters {
     fn check_max_established(&self, endpoint: &ConnectedPoint)
         -> Result<(), ConnectionLimit>
     {
+        // Check total connection limit
+        Self::check(self.num_established(), self.limits.max_established_total)?;
+        // Check incoming/outgoing connection limits
         match endpoint {
             ConnectedPoint::Dialer { .. } =>
                 Self::check(self.established_outgoing, self.limits.max_established_outgoing),
@@ -1031,6 +1034,7 @@ pub struct ConnectionLimits {
     max_established_incoming: Option<u32>,
     max_established_outgoing: Option<u32>,
     max_established_per_peer: Option<u32>,
+    max_established_total: Option<u32>,
 }
 
 impl ConnectionLimits {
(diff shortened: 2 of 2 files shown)


core/src/connection/pool.rs
```diff
@@ -1058,6 +1062,13 @@ impl ConnectionLimits {
         self
     }
 
+    /// Configures the maximum number of concurrent total established connections (inbound or
+    /// outbound).
+    pub fn with_max_established_total(mut self, limit: Option<u32>) -> Self {
```
mxinden, line 1067:
```suggestion
    /// Configures the maximum number of concurrent established connections (both inbound and
    /// outbound).
    pub fn with_max_established(mut self, limit: Option<u32>) -> Self {
```

I would find this naming more intuitive, given that `total` is relative and requires the context of `inbound` and `outbound`.

mxinden (commented):
In general, I am not opposed to this change, though I want to make sure it is not used in the wrong way.

Would you mind expanding on your use-case?

When running rust-libp2p in untrusted environments, I find the differentiation in incoming and outgoing connections valuable as an eclipse-attack defense mechanism.

AgeManning:
Yep agree inbound/outbound limits are useful for preventing eclipses. 

My use-case is that we allow the user essentially to specify an amount of peers (connections) they want to connect to. With this PR we can enforce this limit at the libp2p level, preventing us from having to establish connections and performing initial behaviour tasks before dropping them. 

Although the user specifies essentially a hard limit in connections, we enforce a percentage maximum of those to be inbound.  I think this PR allows a state where all of the total connections are outbound. We don't allow all the total connections to be inbound however. Essentially the ratio of inbound/outbound can vary for reaching the total connection limit. 

I think your point in important though. At the moment, users can choose not to define limits and be vulnerable to eclipses. Perhaps adding more info in the comment around setting a total limit might be useful.

mxinden:
Thanks for the additional context @AgeManning.

Would you mind:

- Adding a changelog entry.
- Documenting the importance of enforcing a limit to incoming connections to prevent eclipse attacks on either `with_max_established` or `with_max_established_outgoing`?

AgeManning:
Added your suggestions. Feel free to modify the added comment.

core/CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 # 0.30.0 [unreleased]
 
+- Add `ConnectionLimit::with_max_established` (see [PR 2137]).
```
mxinden, line 3:
```suggestion
- Add `ConnectionLimit::with_max_established` (see [PR 2137]).

[PR 2137]: https://github.com/libp2p/rust-libp2p/pull/2137/
```

core/src/connection/pool.rs
```diff
@@ -987,6 +987,9 @@ impl ConnectionCounters {
     fn check_max_established(&self, endpoint: &ConnectedPoint)
         -> Result<(), ConnectionLimit>
     {
+        // Check total connection limit
```
mxinden, line 990:
```suggestion
        // Check total connection limit.
```

core/src/connection/pool.rs
```diff
@@ -1058,6 +1062,16 @@ impl ConnectionLimits {
         self
     }
 
+    /// Configures the maximum number of concurrent established connections (both inbound and
+    /// outbound).
+    /// Note: This should be used in conjunction with
+    /// [`ConnectionLimits::with_max_established_incoming`] to prevent possible eclipse attacks (all connections being
+    /// inbound.
```
mxinden, line 1069:
```suggestion
    /// Configures the maximum number of concurrent established connections (both
    /// inbound and outbound).
    ///
    /// Note: This should be used in conjunction with
    /// [`ConnectionLimits::with_max_established_incoming`] to prevent possible
    /// eclipse attacks (all connections being inbound).
```

mxinden (approved):

mxinden:
> Feel free to modify the added comment.

I am unable to commit suggestions to your branch. Can you either give me permissions or apply the suggestions above yourself?

PR #2138, opened 2021-07-14
(suggestion) Make RecordStore associated types IntoIterator instead of Iterator

rubdos:
The idea behind this pull request:

```diff
--- a/protocols/kad/src/record/store.rs                               
+++ b/protocols/kad/src/record/store.rs                               
@@ -60,8 +60,8 @@ pub enum Error {                                    
 ///      to the closest nodes to the key.                            
 ///                                                                  
 pub trait RecordStore<'a> {                                          
-    type RecordsIter: Iterator<Item = Cow<'a, Record>>;              
-    type ProvidedIter: Iterator<Item = Cow<'a, ProviderRecord>>;     
+    type RecordsIter: IntoIterator<Item = Cow<'a, Record>>;          
+    type ProvidedIter: IntoIterator<Item = Cow<'a, ProviderRecord>>; 
```

This allows a RecordStore to return a lot more types, such as a `Vec` or slices, or anything listed under blanket implementation (https://doc.rust-lang.org/std/iter/trait.IntoIterator.html).

This is backwards compatible for implementers of the trait (because `impl IntoIterator for I where I: Iterator`) and gives the implementer a bit more freedom, but is a breaking change for callers that want to directly chain iterator calls (example in the `jobs.rs` patch); they have to add `into_iter()`. Notably, it is not a breaking change when used directly in a `for` loop.

### DIFF
--- a/protocols/kad/src/jobs.rs
+++ b/protocols/kad/src/jobs.rs
@@ -197,6 +197,7 @@ impl PutRecordJob {
         if self.inner.is_ready(cx, now) {
             let publish = self.next_publish.map_or(false, |t_pub| now >= t_pub);
             let records = store.records()
+                .into_iter()
                 .filter_map(|r| {
                     let is_publisher = r.publisher.as_ref() == Some(&self.local_id);
                     if self.skipped.contains(&r.key) || (!publish && is_publisher) {
@@ -290,6 +291,7 @@ impl AddProviderJob {
     {
         if self.inner.is_ready(cx, now) {
             let records = store.provided()
+                .into_iter()
                 .map(|r| r.into_owned())
                 .collect::<Vec<_>>()
                 .into_iter();
--- a/protocols/kad/src/record/store.rs
+++ b/protocols/kad/src/record/store.rs
@@ -60,8 +60,8 @@ pub enum Error {
 ///      to the closest nodes to the key.
 ///
 pub trait RecordStore<'a> {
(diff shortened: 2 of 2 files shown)


mxinden:
> and gives the implementer a bit more freedom

Could you expand on your use-case?

When working with the `RecordStore` trait only, I don't see the benefit of calling `into_iter` as the caller instead of as the callee. Are you using the methods of the concrete type and thus depend on the concrete type of `RecordsIter` and `ProvidedIter`?

rubdos:
> > and gives the implementer a bit more freedom
> 
> Could you expand on your use-case?

While working on my `impl RecordStore`, I would have used `Vec<Cow<'a, Record>>` as a placeholder while designing the iterators.  In the end, it didn't seem to matter; I used `Box<dyn Iterator>` in the end as placeholder.

> When working with the `RecordStore` trait only, I don't see the benefit of calling `into_iter` as the caller instead of as the callee.

Yes, that makes a lot of sense.

> Are you using the methods of the concrete type and thus depend on the concrete type of `RecordsIter` and `ProvidedIter`?

No, not at all!

---

I don't really see a big benefit in this suggestion, feel free to close if you don't see anything in it.

mxinden:
Closing as discussed above. Thanks @rubdos for the proposed patch. Please follow up once more in case there is anything we have been missing in favor of this patch.

PR #1334, opened 2019-12-06
QUIC support

Demi-Marie:
I am using the latest quinn git master.  While I could backport to the latest release, I would prefer to get the existing code working first.

Current approach:

* Use `quinn-proto` (the bare state machine)
* Each connection has a pair of `HashMap`s to store wakers, one for readers and one for writers.  When I/O on a stream becomes possible, the corresponding waker is awoken.
* The mutable state is protected with mutexes, so `libp2p-quic` is thread-safe.
* The state machine is only advanced on a background task.
* Channels are used for incoming connections (to an endpoint) and incoming streams (to a connection).  They are also used to send packets generated by a connection to the background task for transmission.
* Strict mutex aquisition ordering is upheld to prevent deadlocks.  Specifically, if both the mutex protecting the endpoint and the mutex protecting a connection need to be taken, the mutex protecting the endpoint must be taken first.

What I would like feedback on:

* Should I be using channels in places where I am currently using mutexes?
* To avoid potential memory exhaustion denial of service attacks, I am using a bounded channel for outgoing packets.  If it fills up, packets are dropped.  Since QUIC is a reliable protocol, I expect that these will be retransmitted, which is much better than having to buffer an unlimited amount of data.
* How should I handle the case where a connection arrives, but we are not ready for it?  We can’t buffer an unlimited number of them, so at some point we will need to impose backpressure.  The current plan is to rely on quinn-proto limiting the number of connections that it returns but have not yet been accepted.

### Edit

The code is now essentially complete and is ready for review.

PR #2144, opened 2021-07-16
Add libp2p-quic.

dvc94ch:
Would be nice to support tls too, the tls parts from #1334 could be extracted and supported in this crate. This is however left for future work.

cc @Demi-Marie

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,15 +14,16 @@ default = [
     "deflate",
     "dns-async-std",
     "floodsub",
+    "gossipsub",
     "identify",
     "kad",
-    "gossipsub",
     "mdns",
     "mplex",
     "noise",
     "ping",
     "plaintext",
     "pnet",
+    "quic",
     "relay",
     "request-response",
     "secp256k1",
@@ -36,15 +37,16 @@ deflate = ["libp2p-deflate"]
 dns-async-std = ["libp2p-dns", "libp2p-dns/async-std"]
 dns-tokio = ["libp2p-dns", "libp2p-dns/tokio"]
 floodsub = ["libp2p-floodsub"]
+gossipsub = ["libp2p-gossipsub"]
 identify = ["libp2p-identify"]
 kad = ["libp2p-kad"]
-gossipsub = ["libp2p-gossipsub"]
 mdns = ["libp2p-mdns"]
 mplex = ["libp2p-mplex"]
 noise = ["libp2p-noise"]
 ping = ["libp2p-ping"]
 plaintext = ["libp2p-plaintext"]
 pnet = ["libp2p-pnet"]
+quic = ["libp2p-quic"]
 relay = ["libp2p-relay"]
 request-response = ["libp2p-request-response"]
 tcp-async-io = ["libp2p-tcp", "libp2p-tcp/async-io"]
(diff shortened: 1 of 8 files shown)


DemiMarie:
Please CC this account instead; the @Demi-Marie account is no longer used.

mxinden:
Thanks for opening this up and thanks for offering your work to the rust-libp2p project.

As mentioned before I am not opposed to adding QUIC+Noise to rust-libp2p, though, before including it in rust-libp2p, there would need to be a corresponding specification in at least [_Recommended Draft_](https://github.com/libp2p/specs/blob/master/00-framework-01-spec-lifecycle.md#level-1-working-draft) stage, especially since I don't have enough experience to judge whether the construction is safe from a cryptographic standpoint.

We do have many unbounded buffers (e.g. various `Vec`s, `VecDequeue`s and `HashMap`s), which ideally I would like to get rid of in most cases to improve overall throughput, though most of these are within a single module, potentially removable without a breaking change. That said, I am very much opposed to an implementation using unbounded channels, especially as these oftentimes stretch across multiple modules.

Last but not least, unless we change our strategy of supporting both tokio and async-std, this crate would need to support both as well.

dvc94ch:
> As mentioned before I am not opposed to adding QUIC+Noise to rust-libp2p, though, before including it in rust-libp2p, there would need to be a corresponding specification in at least Recommended Draft stage, especially since I don't have enough experience to judge whether the construction is safe from a cryptographic standpoint.

Is that stage 1 or stage 2 with one reference implementation? stage 3 requires at least two implementations which I don't have the time or interest to build a go/js version.

> We do have many unbounded buffers (e.g. various Vecs, VecDequeues and HashMaps), which ideally I would like to get rid of in most cases to improve overall throughput, though most of these are within a single module, potentially removable without a breaking change. That said, I am very much opposed to an implementation using unbounded channels, especially as these oftentimes stretch across multiple modules.

I don't think this actually increases throughput in many cases. It is better for preventing denial of service attacks. I guess this can not be done as a follow-up?

> Last but not least, unless we change our strategy of supporting both tokio and async-std, this crate would need to support both as well.

Also this one, can't be done as a follow-up? It is quite a chunk of work to support an additional executor. The if-watch/udp-socket would need support and spawning a task would be annoying to say the least. Probably an executor would need to be passed as an argument when creating a new transport.

dvc94ch:
Anyway, when it comes to PL merging a spec, that isn't really up to me. I opened a PR, but they tend to spend years on their own specs, never mind a third party one.

dvc94ch:
Well, not that I'm surprised how this went. PL never fails expectation.

dvc94ch:
> As mentioned before I am not opposed to adding QUIC+Noise to rust-libp2p, though, before including it in rust-libp2p, there would need to be a corresponding specification in at least Recommended Draft stage

So there is a PR open to support tls in libp2p-quic [0]. Currently waiting on a new version of barebones-x509. Is supporting tls in addition to noise enough? We can expose the noise implementation behind an unstable-noise or experimental-noise or something like that. 

The tls version can't support 0-rtt and doesn't support keys other than ed25519 for compatibility with the noise implementation. Although with enough generics/traits we could change that.

- [0] https://github.com/ipfs-rust/libp2p-quic/pull/5

PR #2139, opened 2021-07-14
Deal with some clippy warnings

thomaseizinger:
Can drop individual commits if some of the changes are not wanted.

I purposely left everything in `relay` untouched to not interfere with https://github.com/libp2p/rust-libp2p/pull/2059 too much.

PR #2145, opened 2021-07-16
to/into consistency for PublicKey and PeerId

rubdos:
As per discussion in #2142:

- [x] Change `PublicKey::into_protobuf_encoding(self)` to `PublicKey::to_protobuf_encoding(&self)`
- [x] Change `PublicKey::into_peer_id(self)` to `PublicKey::to_peer_id(&self)`
  - [x] Yeet out a whole bunch of `clone()`'s.

The latter goes via a `From` implementation.

### DIFF
--- a/core/CHANGELOG.md
+++ b/core/CHANGELOG.md
@@ -1,4 +1,8 @@
-- Change `PublicKey::into_protobuf_encoding` to `PublicKey::to_protobuf_encoding` (see [PR 2145])
+- Change `PublicKey::into_protobuf_encoding` to `PublicKey::to_protobuf_encoding`.
+  Change `PublicKey::into_peer_id` to `PublicKey::to_peer_id`.
+  Change `PeerId::from_public_key(PublicKey)` to `PeerId::from_public_key(&PublicKey)`.
+  Add `From<&PublicKey> for PeerId`.
+  See [PR 2145]
 
 [PR 2145]: https://github.com/libp2p/rust-libp2p/pull/2145
 
--- a/core/CHANGELOG.md
+++ b/core/CHANGELOG.md
@@ -1,3 +1,7 @@
+- Change `PublicKey::into_protobuf_encoding` to `PublicKey::to_protobuf_encoding` (see [PR 2145])
+
+[PR 2145]: https://github.com/libp2p/rust-libp2p/pull/2145
+
 # 0.29.0 [2021-07-12]
 
 - Switch from `parity-multiaddr` to upstream `multiaddr`.
--- a/core/CHANGELOG.md
+++ b/core/CHANGELOG.md
@@ -1,8 +1,12 @@
-- Change `PublicKey::into_protobuf_encoding` to `PublicKey::to_protobuf_encoding`.
-  Change `PublicKey::into_peer_id` to `PublicKey::to_peer_id`.
-  Change `PeerId::from_public_key(PublicKey)` to `PeerId::from_public_key(&PublicKey)`.
-  Add `From<&PublicKey> for PeerId`.
-  See [PR 2145]
+# 0.30.0 [unreleased]
+
+- Change `PublicKey::into_protobuf_encoding` to `PublicKey::to_protobuf_encoding` (see [PR 2145]).
+
+- Change `PublicKey::into_peer_id` to `PublicKey::to_peer_id` (see [PR 2145]).
+
+- Change `PeerId::from_public_key(PublicKey)` to `PeerId::from_public_key(&PublicKey)` (see [PR 2145]).
+
+- Add `From<&PublicKey> for PeerId` (see [PR 2145]).
 
 [PR 2145]: https://github.com/libp2p/rust-libp2p/pull/2145
 
--- a/core/src/peer_id.rs
+++ b/core/src/peer_id.rs
@@ -114,7 +114,7 @@ impl PeerId {
     pub fn is_public_key(&self, public_key: &PublicKey) -> Option<bool> {
         let alg = Code::try_from(self.multihash.code())
             .expect("Internal multihash is always a valid `Code`");
-        let enc = public_key.clone().to_protobuf_encoding();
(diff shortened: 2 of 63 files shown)


rubdos:
Feel free to only pick the first commit.

Some comment on the `PeerId` conversion: there's a `From<PublicKey>` (and now also `From<&PublicKey>` `for PeerId`. Usually, `From` is not really advised for "expensive" stuff or allocating stuff. It feels a bit out-of-place and unexpected for me, here. Just my 2¢.

Anyway, this include the changelog entry in two separate steps, such that you can cherry-pick the first commit. Preferrably, #2142 is merged before this one, because that yields one less conflict (in core/CHANGELOG...).

core/src/peer_id.rs
```diff
@@ -114,13 +113,19 @@ impl PeerId {
     pub fn is_public_key(&self, public_key: &PublicKey) -> Option<bool> {
         let alg = Code::try_from(self.multihash.code())
             .expect("Internal multihash is always a valid `Code`");
-        let enc = public_key.clone().into_protobuf_encoding();
+        let enc = public_key.clone().to_protobuf_encoding();
```
thomaseizinger, line 116: This and other `clone`s are redundant now I believe.

core/src/peer_id.rs
```diff
@@ -179,32 +184,36 @@ impl FromStr for PeerId {
...
+            .public()
+            .to_peer_id();
         let second: PeerId = peer_id.to_base58().parse().unwrap();
         assert_eq!(peer_id, second);
     }
 
     #[test]
     fn random_peer_id_is_valid() {
-        for _ in 0 .. 5000 {
+        for _ in 0..5000 {
```
thomaseizinger, line 216: Unfortunately, we don't (yet) have rustfmt in here.

For this diff size, I think it is okay to leave it in but with larger diffs, it is best to not autoformat in rust-libp2p :)

See #1966.

core/CHANGELOG.md
```diff
@@ -1,3 +1,11 @@
+- Change `PublicKey::into_protobuf_encoding` to `PublicKey::to_protobuf_encoding`.
+  Change `PublicKey::into_peer_id` to `PublicKey::to_peer_id`.
+  Change `PeerId::from_public_key(PublicKey)` to `PeerId::from_public_key(PublicKey)`.
```
thomaseizinger, line 3: These signatures are identical 😁

thomaseizinger (changes requested):
A few nits!

core/src/peer_id.rs
rubdos: Damnit, I thought I had turned it off for every file. Will fix.

rubdos:
I probably didn't catch *all* of the `clone()`s, mostly because clippy also creates some noise in several other places. If wanted, I can squash that last commit into the `PeerId` commit. 7cbf674 should address your nits, and I'll gladly squash that into 99d0834  :-)

thomaseizinger (approved):
> I probably didn't catch all of the clone()s, mostly because clippy also creates some noise in several other places.

That should be fixed if you merge latest master into your PR. That will include https://github.com/libp2p/rust-libp2p/pull/2139.

> If wanted, I can squash that last commit into the PeerId commit. 7cbf674 should address your nits, and I'll gladly squash that into 99d0834 :-)

Appreciate the effort! Rust-libp2p uses squash-merge though so the individual patches of the PR will be gone anyway :)
Still nice to not force-push because then at least GitHub gives you a nice "changes since your last review" button.

LGTM!

rubdos:
>  Still nice to not force-push because then at least GitHub gives you a nice "changes since your last review" button.

I squashed as I said, I'll add more fixups for the clones now.

rubdos:
Still quite a bit of clippy noise going on, I fixed the `clone`'s that were still visible in my tmux buffer :rofl: 

When `rustfmt` gets default in this crate, I'll volunteer to take some more clippy stuff on me. Without `rustfmt` it's gonna be too much of a mess to submit a PR :cry:

rubdos:
I think this is ready for merge, if everyone is okay with the slightly breaking change.

mxinden (commented):
Thank you for the patch @rubdos. Appreciate the work you put into this and also very much in favor of being consistent with Rust's naming conventions.

Given that this is the first patch after the recent release, and also given that this is a breaking change in `libp2p-core`, would you mind cherry-picking the commit below into this branch?

https://github.com/mxinden/rust-libp2p/commit/c27758fd7bbde042dbd21aa6ed2cdc8435801bd1

rubdos:
> Thank you for the patch @rubdos. Appreciate the work you put into this and also very much in favor of being consistent with Rust's naming conventions.

My pleasure!

> Given that this is the first patch after the recent release, and also given that this is a breaking change in `libp2p-core`, would you mind cherry-picking the commit below into this branch?

Of course, done!

mxinden (approved):

rubdos:
Seems like there was later an incompatible commit. Here's a patch, @mxinden .

```diff+rust
diff --git a/core/src/identity.rs b/core/src/identity.rs                  
index 1569d4db..298797cd 100644                                           
--- a/core/src/identity.rs                                                
+++ b/core/src/identity.rs                                                
@@ -285,12 +285,12 @@ mod tests {                                         
     #[test]                                                              
     fn keypair_protobuf_roundtrip() {                                    
         let expected_keypair = Keypair::generate_ed25519();              
-        let expected_peer_id = expected_keypair.public().into_peer_id(); 
+        let expected_peer_id = expected_keypair.public().to_peer_id();   
                                                                          
         let encoded = expected_keypair.to_protobuf_encoding().unwrap();  
                                                                          
         let keypair = Keypair::from_protobuf_encoding(&encoded).unwrap();
-        let peer_id = keypair.public().into_peer_id();                   
+        let peer_id = keypair.public().to_peer_id();                     
                                                                          
         assert_eq!(expected_peer_id, peer_id);                           
     }                                                                    
```

PR #2150, opened 2021-07-20
Make it easier to define implementations of `NetworkBehaviour`

thomaseizinger:
See patches for details.

Not sure how controversial this proposal is :)

### DIFF
--- a/swarm/CHANGELOG.md
+++ b/swarm/CHANGELOG.md
@@ -1,8 +1,10 @@
 # Unreleased
 
 - Provide default implementations for all functions of `NetworkBehaviour`,
-  except for `new_handler` and `poll`.
-  This should make it easier to create new implementations.
+  except for `new_handler`, `inject_event` and `poll`.
+  This should make it easier to create new implementations. See [PR 2150].
+
+[PR 2150]: https://github.com/libp2p/rust-libp2p/pull/2150/
 
 # 0.30.0 [2021-07-12]
 
--- a/swarm/CHANGELOG.md
+++ b/swarm/CHANGELOG.md
@@ -1,3 +1,9 @@
+# Unreleased
+
+- Provide default implementations for all functions of `NetworkBehaviour`,
+  except for `new_handler` and `poll`.
+  This should make it easier to create new implementations.
+
 # 0.30.0 [2021-07-12]
 
 - Update dependencies.
--- a/swarm/src/behaviour.rs
+++ b/swarm/src/behaviour.rs
@@ -80,7 +80,9 @@ pub trait NetworkBehaviour: Send + 'static {
     /// The addresses will be tried in the order returned by this function, which means that they
     /// should be ordered by decreasing likelihood of reachability. In other words, the first
     /// address should be the most likely to be reachable.
-    fn addresses_of_peer(&mut self, peer_id: &PeerId) -> Vec<Multiaddr>;
+    fn addresses_of_peer(&mut self, _: &PeerId) -> Vec<Multiaddr> {
+        vec![]
+    }
 
     /// Indicate to the behaviour that we connected to the node with the given peer id.
     ///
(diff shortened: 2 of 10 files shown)


swarm/src/behaviour.rs
```diff
@@ -127,10 +129,10 @@ pub trait NetworkBehaviour: Send + 'static {
...
     fn inject_event(
         &mut self,
-        peer_id: PeerId,
-        connection: ConnectionId,
-        event: <<Self::ProtocolsHandler as IntoProtocolsHandler>::Handler as ProtocolsHandler>::OutEvent
-    );
+        _: PeerId,
+        _: ConnectionId,
+        _: <<Self::ProtocolsHandler as IntoProtocolsHandler>::Handler as ProtocolsHandler>::OutEvent
+    ) { }
```
mxinden, line 135: A `NetworkBehaviour` that does not receive any events from its `ProtocolsHandler` should use `void::Void` on their `ProtocolsHandler::OutEvent`, thus checked at compile time, no?

I would prefer the above over a default implementation. What do you think? If you agree, could you add a doc comment to `inject_event`?

swarm/CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+# Unreleased
+
+- Provide default implementations for all functions of `NetworkBehaviour`,
+  except for `new_handler` and `poll`.
+  This should make it easier to create new implementations.
```
mxinden, line 5:
```suggestion
  This should make it easier to create new implementations. See [PR 2150].
  
[PR 2150]: https://github.com/libp2p/rust-libp2p/pull/2150/
```

mxinden (commented):
Looks good to me other than `inject_event`. What do you think of my comment below @thomaseizinger?

swarm/src/behaviour.rs
thomaseizinger: We can do that yeah.

I approached this more from a prototyping perspective: If I want to define a new `NetworkBehaviour`, what is the least amount of stuff I am faced with?

I was also tempted to default `poll` actually. An `AddressBook` style `NetworkBehaviour` doesn't need `poll` for example :)

I am not sure `inject_event` with `match event { }` is actually cleaner. The only thing it protects you from is forgetting to deal with your events if you change the `ProtocolsHandler`.

For this to work, it implies that people actually override it with `match event { }` in the first place and not just leave it blank. At least rust-analyzer autocompletes trade-impls with `todo!()` though and the path of least resistance is to either leave those in or delete them 🤷‍♂️

In any case, defaulting the others is already an improvement so I'll do that.

mxinden (approved):
@thomaseizinger can you fix the failing tests below? Other than that, this looks good to me.

```
error[E0046]: not all trait items implemented, missing: `inject_event`
   --> swarm/src/test.rs:68:1
   ```
```
   ::: swarm/src/behaviour.rs:130:5
```

**Comment by thomaseizinger** [comment]
> @thomaseizinger can you fix the failing tests below? Other than that, this looks good to me.
> 
> ```
> error[E0046]: not all trait items implemented, missing: `inject_event`
>    --> swarm/src/test.rs:68:1
>    ```
> ```
>    ::: swarm/src/behaviour.rs:130:5
> ```

Ah damn. Still need to get used to my new RA based setup ...

**Comment by mxinden** [comment]
Can you take another look at the current failure @thomaseizinger?

```
error[E0223]: ambiguous associated type
  --> swarm/src/test.rs:89:12
   |
89 |         _: ProtocolsHandler::OutEvent
   |            ^^^^^^^^^^^^^^^^^^^^^^^^^^ help: use fully-qualified syntax: `<Type as protocols_handler::ProtocolsHandler>::OutEvent`

error: aborting due to previous error
```

thomaseizinger:
> Can you take another look at the current failure @thomaseizinger?
> 
> ```
> error[E0223]: ambiguous associated type
>   --> swarm/src/test.rs:89:12
>    |
> 89 |         _: ProtocolsHandler::OutEvent
>    |            ^^^^^^^^^^^^^^^^^^^^^^^^^^ help: use fully-qualified syntax: `<Type as protocols_handler::ProtocolsHandler>::OutEvent`
> 
> error: aborting due to previous error
> ```

Argh, I ran clippy without `--all-targets`.

thomaseizinger:
Hopefully everything is done now @mxinden !

Sorry about the mess, transitioning local setup at the moment and not quite used to it yet ...

mxinden:
Thanks @thomaseizinger!

PR #2159, opened 2021-07-26
Libp2p quic second attempt

### LINKED ISSUE #1334: Have the `decl_storage!` macro import rstd::prelude::*
When trying to use the decl_storage macro and build to wasm, the compiler complains that the Vec type has not been declared.

For example, this minimum runtime:
```
use srml_support::{StorageValue, dispatch::Result};
pub trait Trait: system::Trait {}

decl_module! {
  pub struct Module<T: Trait> for enum Call where origin: T::Origin {
    fn set_value(_origin, value: u32) -> Result {
      <Value<T>>::put(value);
      Ok(())
    }
  }
}

decl_storage! {
  trait Store for Module<T: Trait> as RuntimeExampleStorage {
    Value: u32;
  }
}
```
will give the following error:
```
error[E0412]: cannot find type `Vec` in this scope
  --> src/runtime_example.rs:13:1
   |
13 | / decl_storage! {
14 | |   trait Store for Module<T: Trait> as RuntimeExampleStorage {
15 | |     Value: u32;
16 | |   }
17 | | }
   | |_^ not found in this scope
help: possible candidates are found in other modules, you can import them into scope
   |
1  | use rstd::alloc::prelude::Vec;
   |
1  | use rstd::prelude::Vec;
   |
1  | use rstd::vec::Vec;
   |
1  | use srml_support::dispatch::Vec;
   |
and 1 other candidates

error: aborting due to previous error

For more information about this error, try `rustc --explain E0412`.
error: Could not compile `runtime-example-runtime`.
```

To solve this issue, you simply need to add: `use rstd::prelude::*;`

We should have the decl_storage macro automatically do this if it requires this import.

dvc94ch:
Adds support for tls, removes the `UnboundedSender`s from the endpoint. The endpoint still has `UnboundedReceiver`s, but that's not different than having a `VecDeque`. The transport dial api for example doesn't allow applying backpressure.

Closes paritytech/substrate#1334


mxinden:
Thanks @dvc94ch for giving this another shot.

As someone not directly involved in the many Rust QUIC efforts, could you compare this implementation on a high level to the many unfinished efforts out there, namely:

- https://github.com/libp2p/rust-libp2p/pull/1334
- https://github.com/tomaka/libp2p-rs/tree/quiccc-again
- https://github.com/paritytech/polkadot-sdk/issues/536

//CC @kpp who might be able to add some additional context as well.

dvc94ch:
well first demie wrote an implementation. then tomaka improved on it. it stalled mainly because of rustls/quinn-proto not having a new release for over a year. this implementation builds on previous work and resolves the open todo's. the tls implementation was updated to the latest spec and can tested with other implementations by kpp.

features it has that the old version didn't have is proper udp socket handling with ecn codes and gso (performance). support for plugging in your own crypto. support for 0rtt encryption (latency). other than correctness/performance/latency not really sure how a generic connection oriented transport can differentiate itself.

kpp:
@tomaka ^

dvc94ch:
anyone interested in performing a first round of review?

kpp:
Well, my issue is to finish @tomaka’s attempt.

dvc94ch:
Why do you think that is a better starting point?

dvc94ch:
Are you referring to https://github.com/kpp/rust-libp2p/tree/master/transports/tls-quic/src? It seems incomplete to me but the tls code does look quite neat.

kpp:
> Why do you think that is a better starting point?

That's an issue I was assigned.

> the tls code does look quite neat.

Thank you! I will merge Pierre's code from https://github.com/tomaka/libp2p-rs/tree/quiccc-again/transports/quic into it.

dvc94ch:
So parity decided to finish it's branch now. How convenient.

I assume that the most contentious part is that there is noise support in this branch and you probably think it'll make your code much simpler to support just TLS. It doesn't make much difference as you saw in the crypto trait. Although there isn't support for non ed25519 keys, it's probably not that hard to add.

Even in the quiccc branch there are lots of open todos and missing features. Are you planning on fixing those?

dvc94ch:
In that case I'll continue maintaining libp2p-quic. I assume @mxinden will wait a few months to see what happens before deciding anything.

> I'd strongly prefer for us to implement something that is already specified (i.e. TLS+QUIC) and not start creating parity-nquic that will get abandoned after 6 months.

Let's hope your branch doesn't get abandonned again in 6 months.

dvc94ch:
This branch supports TLS too, although @kpp seems to have done some improvements and cleanup.

kpp:
> So parity decided to finish it's branch now. How convenient.

I talked to Pierre. Let's unite on your PR.

transports/quic/src/tls/certificate.rs
```diff
@@ -0,0 +1,76 @@
...
+
+    // Generate the libp2p-specific extension.
+    let libp2p_extension: rcgen::CustomExtension = {
+        let extension_content = {
+            let serialized_pubkey = keypair.public().to_protobuf_encoding();
+            yasna::encode_der(&(serialized_pubkey, libp2p_ext_signature))
+        };
+
+        let mut ext = rcgen::CustomExtension::from_oid_content(LIBP2P_OID, extension_content);
+        ext.set_criticality(false);
```
kpp, line 62: Criticality can be enabled again according to https://github.com/libp2p/go-libp2p-tls/issues/87

transports/quic/Cargo.toml
```diff
@@ -0,0 +1,46 @@
...
+parking_lot = "0.11.1"
+quinn-noise = { version = "0.3.0", optional = true }
+quinn-proto = { version = "0.7.3", default-features = false }
+rand_core = "0.5.1"
+rcgen = { version = "0.8.11", optional = true }
+ring = { version = "0.16.20", optional = true }
+rustls = { version = "0.19.1", optional = true, features = ["dangerous_configuration"] }
+thiserror = "1.0.26"
+tracing = "0.1.26"
+udp-socket = "0.1.5"
```
kpp, line 35: This crate looks like a copypaste from https://github.com/quinn-rs/quinn/blob/main/quinn/src/platform/mod.rs. Wouldn't it be more correct to try to pick their code into `quinn_udp` crate in the root of quinn-rs?

dvc94ch: I guess that would be good, yes. Opened an issue for it https://github.com/quinn-rs/quinn/issues/1178

transports/quic/Cargo.toml
```diff
@@ -0,0 +1,46 @@
+[package]
+name = "libp2p-quic"
+version = "0.6.0"
+authors = ["David Craven <david@craven.ch>", "Parity Technologies <admin@parity.io>"]
+edition = "2018"
+description = "libp2p-quic is a noise based quic implementation for rust-libp2p."
```
kpp, line 6: Not only noise

transports/quic/Cargo.toml
```diff
@@ -0,0 +1,46 @@
...
+repository = "https://github.com/ipfs-rust/libp2p-quic"
```
kpp, line 7: Fix repo url

transports/quic/src/transport.rs
```diff
@@ -0,0 +1,350 @@
...
+            Ok((
+                SocketAddr::new(
+                    IpAddr::V6(Ipv6Addr::new(
+                        65535, 65535, 65535, 65535, 65535, 65535, 65535, 65535,
+                    )),
+                    8080,
+                ),
+                None
+            ))
+        );
```
kpp, line 348: No tests for `/ipV/ADDR/udp/PORt/quic/p2p/PEERID`.

transports/quic/tests/smoke.rs
```diff
@@ -0,0 +1,362 @@
...
+struct Pong(Vec<u8>);
+
+impl ProtocolName for PingProtocol {
+    fn protocol_name(&self) -> &[u8] {
+        "/ping/1".as_bytes()
+    }
+}
+
+#[async_trait]
+impl RequestResponseCodec for PingCodec {
```
kpp, line 205: Is there any reason not to reuse the existing `libp2p::ping`?

transports/quic/Cargo.toml
```diff
@@ -0,0 +1,46 @@
...
+license = "ISC"
```
kpp, line 8: This is completely up to you (and I am not asking you to change this) however all other subcrates in libp2p are licensed under a MIT license.

transports/quic/tests/smoke.rs
dvc94ch: for test purposes it's useful to be able to add large payloads otherwise the stream will only contain one packet

transports/quic/Cargo.toml
dvc94ch: can change, ISC and MIT are interchangeable. as long as it's not gpl, I don't really care much about the license

transports/quic/src/crypto.rs
```diff
@@ -0,0 +1,150 @@
+use ed25519_dalek::{Keypair, PublicKey};
+use libp2p_core::PeerId;
+use quinn_proto::crypto::Session;
+use quinn_proto::TransportConfig;
+use std::sync::Arc;
+
+pub struct CryptoConfig<K> {
+    pub keypair: Keypair,
```
kpp, line 8: For noise this should be ed25519::Keypair, however for tls this should be libp2p_core::identity::Keypair.

transports/quic/src/lib.rs
```diff
@@ -0,0 +1,131 @@
...
+    Connect(#[from] ConnectError),
+    #[error("{0}")]
+    Muxer(#[from] QuicMuxerError),
+    #[error("{0}")]
+    Io(#[from] std::io::Error),
+    #[error("a `StreamMuxerEvent` was generated before the handshake was complete.")]
+    UpgradeError,
+}
+
+pub trait ToLibp2p {
```
kpp, line 101: cfg feature noise

transports/quic/Cargo.toml
dvc94ch: Is this mandatory for merging? it may take a while. the rest should be easy to fix.

I have another project that will take me 1-2 weeks to complete (and since I'm getting payed for it it takes precedence), but I'll address your feedback afterwards when I have more idle time. Thanks!

kpp: I don't think this is mandatory for merging. I may even take this issue if you are busy.

dvc94ch: Feel free, that would be cool

kpp: I opened 2 PRs in the quinn repo extracting UDP code into a separate crate and applying your patches.

transports/quic/tests/smoke.rs
```diff
@@ -0,0 +1,362 @@
...
+    <C::Session as Session>::PacketKey: Unpin,
+{
+    let keypair = Keypair::generate(&mut rand_core::OsRng {});
+    let peer_id = keypair.to_peer_id();
+    let mut transport = QuicConfig::<C>::new(keypair);
+    if keylog {
+        transport.enable_keylogger();
+    }
+    let transport = transport
+        .listen_on("/ip4/127.0.0.1/udp/0/quic".parse()?)
```
dvc94ch, line 43: here we spawn the background task

transports/quic/tests/smoke.rs
```diff
@@ -0,0 +1,362 @@
...
+        .with_env_filter(tracing_subscriber::EnvFilter::from_default_env())
+        .try_init()
+        .ok();
+    log_panics::init();
+    let mut rng = rand::thread_rng();
+
+    let mut a = create_swarm::<C>(true).await?;
+    let mut b = create_swarm::<C>(false).await?;
+
+    Swarm::listen_on(&mut a, "/ip4/127.0.0.1/udp/0/quic".parse()?)?;
```
dvc94ch, line 70: here we tell the swarm about our listening address

dvc94ch (commented):
Not very nice, but I guess it works for now

transports/quic/src/transport.rs
```diff
@@ -0,0 +1,350 @@
...
+    let peer_id = if let Some(Protocol::P2p(peer_id)) = iter.peek() {
+        if peer_id.code() != multihash::Code::Identity.into() {
+            return Err(());
+        }
+        let public_key =
+            libp2p_core::PublicKey::from_protobuf_encoding(peer_id.digest()).map_err(|_| ())?;
+        let public_key = if let libp2p_core::PublicKey::Ed25519(public_key) = public_key {
+            public_key.encode()
+        } else {
+            return Err(());
```
kpp, line 259: This code does not support other peer ids than Ed25519

transports/quic/src/transport.rs
```diff
@@ -0,0 +1,350 @@
+use crate::crypto::Crypto;
+use crate::endpoint::{EndpointConfig, TransportChannel};
+use crate::muxer::QuicMuxer;
+use crate::{QuicConfig, QuicError};
+use ed25519_dalek::PublicKey;
```
kpp, line 5: Why did you use ed25519_dalek::PublicKey instead of libp2p::identity::ed25519::PublicKey?

transports/quic/Cargo.toml
dvc94ch: marking as resolved as @kpp and I decided out of band that we will keep `udp-socket`. reaching consensus with quinn developers was harder than expected.

transports/quic/Cargo.toml
dvc94ch: fixed in https://github.com/dvc94ch/rust-libp2p/pull/13

transports/quic/Cargo.toml
dvc94ch: fixed in https://github.com/dvc94ch/rust-libp2p/pull/13

transports/quic/Cargo.toml
dvc94ch: fixed in https://github.com/dvc94ch/rust-libp2p/pull/13

transports/quic/src/transport.rs
dvc94ch: fixed in https://github.com/dvc94ch/rust-libp2p/pull/13

transports/quic/src/lib.rs
dvc94ch: fixed in https://github.com/dvc94ch/rust-libp2p/pull/13

transports/quic/src/crypto.rs
dvc94ch: fixed in https://github.com/dvc94ch/rust-libp2p/pull/13

transports/quic/tests/smoke.rs
dvc94ch: leaving this open, although not a requirement for merging from my perspective

transports/quic/tests/smoke.rs
dvc94ch: also leaving this open, but not a requirement for merging

transports/quic/src/transport.rs
dvc94ch: Done

transports/quic/src/tls/certificate.rs
dvc94ch: Done

transports/quic/src/endpoint.rs
```diff
@@ -0,0 +1,459 @@
...
+    }
+}
+
+struct Endpoint<C: Crypto> {
+    channel: EndpointChannel<C>,
+    endpoint: QuinnEndpoint<C::Session>,
+    socket: UdpSocket,
+    crypto_config: Arc<CryptoConfig<C>>,
+    connections: FnvHashMap<ConnectionHandle, mpsc::Sender<ConnectionEvent>>,
+    outgoing: VecDeque<udp_socket::Transmit>,
```
DemiMarie, line 234: Is there a proof that this cannot get too large?  If it does get over a certain size, would it be better to drop packets, as e.g. the kernel network stack does?

transports/quic/src/endpoint.rs
```diff
@@ -0,0 +1,459 @@
...
+        transport
+    }
+}
+
+struct Endpoint<C: Crypto> {
+    channel: EndpointChannel<C>,
+    endpoint: QuinnEndpoint<C::Session>,
+    socket: UdpSocket,
+    crypto_config: Arc<CryptoConfig<C>>,
+    connections: FnvHashMap<ConnectionHandle, mpsc::Sender<ConnectionEvent>>,
```
DemiMarie, line 233: Could a HashDOS be caused by closing and opening connections until the hash table fills up with collisions?

transports/quic/src/endpoint.rs
```diff
@@ -0,0 +1,459 @@
...
+                    None => {
+                        me.endpoint.reject_new_connections();
+                        return Poll::Ready(());
+                    }
+                }
+            }
+        }
+
+        while !me.outgoing.is_empty() {
+            me.outgoing.make_contiguous();
```
DemiMarie, line 394: This is O(n)

transports/quic/src/endpoint.rs
```diff
@@ -0,0 +1,459 @@
...
+            me.recv_buf
+                .chunks_mut(me.recv_buf.len() / BATCH_SIZE)
+                .enumerate()
+                .for_each(|(i, buf)| unsafe {
+                    iovs.as_mut_ptr()
+                        .cast::<IoSliceMut>()
+                        .add(i)
+                        .write(IoSliceMut::new(buf));
+                });
+            let mut iovs = unsafe { iovs.assume_init() };
```
DemiMarie, line 416: Why the unsafe code here?

transports/quic/src/endpoint.rs
```diff
@@ -0,0 +1,459 @@
...
+        connection_id: ConnectionHandle,
+        event: EndpointEvent,
+    },
+    /// Instruct the endpoint to transmit a packet on its UDP socket.
+    Transmit(Transmit),
+}
+
+#[derive(Debug)]
+pub struct TransportChannel<C: Crypto> {
+    tx: mpsc::UnboundedSender<ToEndpoint<C>>,
```
DemiMarie, line 46: Is there a proof that this always remains bounded?

transports/quic/src/endpoint.rs
dvc94ch: These are outgoing, so you want to drop packets to prevent libp2p dos'ing itself? As far as I know this pattern is used throughout the codebase

transports/quic/src/endpoint.rs
dvc94ch: What's the mitigation? Using a btreemap?

transports/quic/src/endpoint.rs
dvc94ch: same here, this is the channel from libp2p not from the socket. other than erroring if the queue is full and trying to dial I don't see what we can do

transports/quic/src/endpoint.rs
dvc94ch: so `quinn` uses the same pattern. we allocate a vec at startup to store BATCH_SIZE udp packets. we then create a stack allocated array of pointers to pass to `recvmmsg`. we could heap allocate them in a vec instead so we don't have to declare uninitialized memory. however due to lifetime issues I don't think we can allocate the iovs at startup (at least not without more unsafe code).

transports/quic/src/endpoint.rs
dvc94ch: `quinn` does the same. a queue built from two vecs will require popping and pushing a value once more than if we just used a stack using a single vec. not sure much can be done here, or that it is a problem in practice?

transports/quic/src/tls/verifier.rs
```diff
@@ -0,0 +1,244 @@
...
+) -> Result<(barebones_x509::X509Certificate<'_>, Libp2pExtension<'_>), Error> {
+    let parsed = barebones_x509::parse_certificate(certificate)?;
+    let mut libp2p_extension = None;
+
+    parsed
+        .extensions()
+        .iterate(&mut |oid, critical, extension| {
+            match oid {
+                super::LIBP2P_OID_BYTES if libp2p_extension.is_some() => return Err(Error::BadDER),
+                super::LIBP2P_OID_BYTES if critical == true => {
```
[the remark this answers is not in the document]
kpp: You are right!

transports/quic/src/endpoint.rs
DemiMarie: Is it possible for a peer to create a situation where libp2p tries to send a lot of data, but then is not actually able to do so?  That said, I do not believe this should block merging if it is not a regression compared to the other parts of the codebase.

transports/quic/src/endpoint.rs
DemiMarie: That is what I would use, yes.

transports/quic/src/endpoint.rs
DemiMarie: No idea.  I think this is really a missing feature in the standard library, which should provide better APIs for this.

transports/quic/src/endpoint.rs
DemiMarie: I suggest factoring this into a separate function, with a proof of safety that holds even if `IoSliceMut:new()` panics.

transports/quic/src/endpoint.rs
DemiMarie: That makes sense.  FYI I believe much of the complexity in my original PR was from avoiding unbounded channels at all costs, which required me to use locking instead of message-passing.

transports/quic/src/tls/verifier.rs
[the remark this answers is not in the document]
DemiMarie: This also allows smuggling a non-critical libp2p extension alongside a critical one, which is incorrect.  I should probably document that `barebones_x509` does not check for duplicate extensions, as that would require heap allocation.  It truly is a barebones library.

DemiMarie (changes requested):
The libp2p extension may not be critical, and the unsafe code needs a proof of correctness.  Otherwise looks good.

transports/quic/src/endpoint.rs
kpp: We get UB if `chunks.len < BATCH_SIZE`: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=f53ebde1704cf8f6c70ade877ad17fc7

dvc94ch: yes, but it can't happen: `let recv_buf = vec![0; max_udp_payload_size * BATCH_SIZE].into_boxed_slice();`

dvc94ch: obviously if you assume uninitialized bytes are initialized that's UB

kpp: Yes, you are right. There is a corner case when `max_udp_payload_size=0` but we will panic on `.chunks_mut(0)`.

transports/quic/src/endpoint.rs
kpp: Or we can use `std::collections::HashMap` which is hashed with a default hash function - SipHasher13. 

> SipHash was designed in 2012 by Jean-Philippe Aumasson and Daniel J. Bernstein as a defense against hash-flooding DoS attacks.

dvc94ch: sounds good to me

transports/quic/src/endpoint.rs
dvc94ch: The first packet requires a large payload before the server will answer. This restriction is lifted after the ip address was validated to prevent amplification attacks. Later on I guess that would depend on the application protocol. If you are concerned about this it might be a good idea to bandwidth limit ip's using the bandwidth wrapper.

transports/quic/src/endpoint.rs
dvc94ch: since it's a queue, I don't think it can be done any more efficiently. as far as I know queues are usually built from two stacks, requiring pushing and popping the elements from one stack to the other. Make contiguous does just that.

transports/quic/src/endpoint.rs
kpp: > with a proof of safety that holds even if `IoSliceMut:new()` panics.

@Demi-Marie `IoSliceMut:new()` does not panic. Also:

> Note that dropping a MaybeUninit<T> will never call T’s drop code. It is your responsibility to make sure T gets dropped if it got initialized.

kpp: > I don't think we can allocate the iovs at startup (at least not without more unsafe code)

We allocate them but not initialize, because it's a self-reference issue.

transports/quic/src/crypto.rs
```diff
@@ -0,0 +1,215 @@
...
+
+    fn to_public(&self) -> libp2p_core::identity::PublicKey {
+        self.public.to_public()
+    }
+}
+
+#[cfg(feature = "noise")]
+impl ToLibp2p for ed25519_dalek::PublicKey {
+    fn to_keypair(&self) -> libp2p_core::identity::Keypair {
+        panic!("wtf?");
```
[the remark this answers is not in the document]
dvc94ch: Done, the code was removed

transports/quic/src/tls/verifier.rs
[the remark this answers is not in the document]
dvc94ch: @kpp was this resolved?

transports/quic/src/endpoint.rs
dvc94ch: Marking this as resolved

transports/quic/src/endpoint.rs
dvc94ch: Marking this as resolved

transports/quic/src/endpoint.rs
dvc94ch: I think this is done? @Demi-Marie are you satisfied?

transports/quic/src/endpoint.rs
dvc94ch: Marking as resolved

transports/quic/src/endpoint.rs
dvc94ch: Done

transports/quic/src/tls/verifier.rs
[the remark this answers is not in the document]
kpp: yes

transports/quic/src/endpoint.rs
kpp: Here is a PR https://github.com/dvc94ch/rust-libp2p/pull/17

kpp:
Let's wait https://github.com/libp2p/rust-libp2p/pull/2226, https://github.com/libp2p/rust-libp2p/pull/2227 and possibly https://github.com/libp2p/rust-libp2p/pull/2228 before merging master.

dvc94ch:
@kpp, @DemiMarie can you approve the PR? I think all issues have been addressed.

@mxinden what are the next steps to get this merged?

kpp (approved):
So far so good

DemiMarie (approved):
I recommend running the tests thousands of times in a tight loop, using netem to simulate packet loss and/or reordering.  This caught bugs I was working on my version, both in my code and in quinn-proto.

mxinden:
> @mxinden what are the next steps to get this merged?

I hope to find some time next week to give this an in-depth review. Thanks for the ping.

transports/quic/src/crypto.rs
```diff
@@ -0,0 +1,199 @@
+use libp2p_core::PeerId;
```
mxinden, line 1: For the sake of consistency, would you mind including license headers here?

transports/quic/src/endpoint.rs
```diff
@@ -0,0 +1,471 @@
...
+
+/// Message sent to the endpoint background task.
+#[derive(Debug)]
+enum ToEndpoint<C: Crypto> {
+    /// Instructs the endpoint to start connecting to the given address.
+    Dial {
+        /// UDP address to connect to.
+        addr: SocketAddr,
+        /// The remotes public key.
+        public_key: C::PublicKey,
```
mxinden, line 29: Do I understand correctly that one can thus not dial a remote without their `PeerId`?

transports/quic/src/endpoint.rs
mxinden: > same here, this is the channel from libp2p not from the socket. other than erroring if the queue is full and trying to dial I don't see what we can do

I don't understand why the direction (libp2p -> socket) matters in regards to bounded or unbounded. Can you expand on that @dvc94ch?

transports/quic/src/endpoint.rs
```diff
@@ -0,0 +1,471 @@
...
+        addr: SocketAddr,
+        public_key: C::PublicKey,
+    ) -> oneshot::Receiver<Result<QuicMuxer<C>, QuicError>> {
+        let (tx, rx) = oneshot::channel();
+        let msg = ToEndpoint::Dial {
+            addr,
+            public_key,
+            tx,
+        };
+        self.tx.unbounded_send(msg).expect("endpoint has crashed");
```
mxinden, line 63: Why can one not use a bounded channel here and await both the sending and the result through the oneshot?

transports/quic/src/endpoint.rs
```diff
@@ -0,0 +1,471 @@
...
+        rx
+    }
+
+    pub fn poll_incoming(
+        &mut self,
+        cx: &mut Context,
+    ) -> Poll<Option<Result<QuicMuxer<C>, QuicError>>> {
+        Pin::new(&mut self.rx).poll_next(cx)
+    }
```
mxinden, line 72: Why not implement `Stream` on `TransportChannel`?

transports/quic/src/endpoint.rs
```diff
@@ -0,0 +1,471 @@
...
+impl<C: Crypto> EndpointChannel<C> {
+    pub fn poll_next_event(&mut self, cx: &mut Context) -> Poll<Option<ToEndpoint<C>>> {
+        Pin::new(&mut self.rx).poll_next(cx)
+    }
+
+    pub fn create_connection(
+        &self,
+        id: ConnectionHandle,
+    ) -> (ConnectionChannel<C>, mpsc::Sender<ConnectionEvent>) {
+        let (tx, rx) = mpsc::channel(12);
```
mxinden, line 141: What is the reasoning for the constant `12`?

transports/quic/src/endpoint.rs
```diff
@@ -0,0 +1,471 @@
...
+            ty: self.socket.socket_type(),
+        };
+        let endpoint = EndpointChannel {
+            tx: tx2,
+            rx: rx1,
+            port: self.port,
+            max_datagrams: self.capabilities.max_gso_segments,
+            connection_tx: transport.tx.clone(),
+        };
+        async_global_executor::spawn(Endpoint::new(endpoint, self)).detach();
```
mxinden, line 222: All of rust-libp2p is generic over the executor to be used. Before we introduce a protocol locked onto a single executor, what is the reasoning for not being generic over the executor?

transports/quic/src/endpoint.rs
mxinden: How is this different to an unbounded channel? How is back-pressure to the upper layer enforced?

transports/quic/src/endpoint.rs
```diff
@@ -0,0 +1,471 @@
...
+                    me.outgoing.drain(..n);
+                }
+                Poll::Ready(Err(err)) => tracing::error!("send_to: {}", err),
+                Poll::Pending => break,
+            }
+        }
+
+        if me.event_slot.is_none() && me.incoming_slot.is_none() {
+            let mut metas = [RecvMeta::default(); BATCH_SIZE];
+            let mut iovs = MaybeUninit::<[IoSliceMut; BATCH_SIZE]>::uninit();
```
mxinden, line 405: I am assuming this is an optimization. If so, given the usage of unsafe code, what performance impact did this optimization show in benchmarks?

transports/quic/src/muxer.rs
```diff
@@ -0,0 +1,418 @@
...
+        cx: &mut Context,
+    ) -> Poll<Result<StreamMuxerEvent<Self::Substream>, Self::Error>> {
+        let mut inner = self.inner.lock();
+        let now = Instant::now();
+
+        while let Poll::Ready(event) = inner.endpoint.poll_channel_events(cx) {
+            inner.connection.handle_event(event);
+        }
+
+        let _max_datagrams = inner.endpoint.max_datagrams();
```
mxinden, line 125: Does this method call have any side effects?

transports/quic/src/muxer.rs
```diff
@@ -0,0 +1,418 @@
...
+        while let Some(event) = inner.connection.poll_endpoint_events() {
+            inner.endpoint.send_endpoint_event(event);
+        }
+
+        while let Some(event) = inner.connection.poll() {
+            match event {
+                Event::HandshakeDataReady => {}
+                Event::Connected => {
+                    // Break here so that the noise upgrade can finish.
+                    return Poll::Pending;
```
mxinden, line 155: Can you expand on this? Why does this Future have to pause? Where is the waker of the task registered?

transports/quic/src/muxer.rs
```diff
@@ -0,0 +1,418 @@
...
+                        waker.wake();
+                    }
+                    return Poll::Ready(Err(QuicMuxerError::ConnectionLost { reason }));
+                }
+                Event::Stream(StreamEvent::Opened { dir: Dir::Bi }) => {
+                    // handled at end.
+                }
+                Event::Stream(StreamEvent::Readable { id }) => {
+                    tracing::trace!("stream readable {}", id);
+                    if let Some(substream) = inner.substreams.get_mut(&id) {
```
mxinden, line 170: Is there a scenario where one would receive a `StreamEvent::Readable` but not have a reference to it in `inner.substreams`? If not, how about an `expect` here to enforce this assumption?

transports/quic/src/transport.rs
```diff
@@ -0,0 +1,385 @@
...
+                        peer_id,
+                        inner.muxer.take().expect("future polled after ready"),
+                    )))
+                } else {
+                    Poll::Pending
+                }
+            }
+            Poll::Ready(Err(err)) => Poll::Ready(Err(err.into())),
+            Poll::Ready(Ok(_)) => {
+                panic!("muxer.incoming is set to false so no events can be produced");
```
mxinden, line 236:
```suggestion
                unreachable!("muxer.incoming is set to false so no events can be produced");
```

Nit pick.

mxinden (commented):
I am sorry for the delay here.

I still need more time for a full review. I got a couple questions which I am posting ahead of time. Let me know in case you would rather wait for the full review.

transports/quic/src/endpoint.rs
dvc94ch: you want a bounded channel to prevent dos attacks. you can't prevent a user from dos'ing himself, he can just exit the application.

transports/quic/src/endpoint.rs
dvc94ch: I guess you can do that, but what's the point

transports/quic/src/endpoint.rs
dvc94ch: what's the point? this is an internal api, so there isn't much point in implementing a bunch of unnecessary traits

transports/quic/src/endpoint.rs
dvc94ch: empirical testing. I printlned when it was blocked and 12 seemed to work well for the test

transports/quic/src/endpoint.rs
dvc94ch: there is currently no api for a transport to use the swarm's executor

transports/quic/src/endpoint.rs
dvc94ch: queues are used throughout the codebase, don't know why we're making a fuss about it here

transports/quic/src/muxer.rs
dvc94ch: no, this is when using quinn-proto from git which changes the api

dvc94ch: https://github.com/quinn-rs/quinn/blob/main/quinn-proto/src/connection/mod.rs#L411

transports/quic/src/muxer.rs
dvc94ch: this is so the upgrade can complete
https://github.com/libp2p/rust-libp2p/pull/2159/files#diff-b1dcda9367774dd8a742457ff79bc528d030a8cc38d6791efefc78b30f3f2e03R222

dvc94ch:
I no longer work for actyx (they downsized to 3 employees). I'm currently working on a startup with @rklaehn and @wngr and as such my priorities have shifted and I won't be able to devote time to this effort for the time being.

kpp:
I am back next week. 🌴

transports/quic/src/endpoint.rs
mxinden: > can't prevent a user from dos'ing himself

How does an upper layer component know whether the lower layer is able to keep up. In other words, how does the upper layer know when to add more items and when to slow down? Backpressure through bounded channels enforces just that.

transports/quic/src/endpoint.rs
mxinden: The point would be to enable backpressure throughout the system. Through backpressure an upper layer component is paused when the lower layer is overwhelmed and woken up when the lower layer is ready to receive new items.

transports/quic/src/endpoint.rs
mxinden: I wonder whether the queue is needed at all. As a rule of thumb I follow the rule of [_Queues exist only to handle burstiness_](https://apenwarr.ca/log/20170814). Is _creating new connections_ a bursty task? If not, why introduce a queue?

transports/quic/src/endpoint.rs
mxinden: > queues are used throughout the codebase,

That is correct, yes, though I would argue that those places are either (a) not relevant (e.g. `libp2p-ping` with little to no bandwidth usage) or (b) a deficiency introduced in the past.

That said, lower layers of libp2p do enforce backpressure today, e.g. the TCP transport (`libp2p-tcp`) or the Yamux implementation..

mxinden (commented):
Thank you for the comments @dvc94ch!

A couple of follow ups on the backpressure discussions. I am sorry for being stubborn here. Still I think this is an important aspect of the implementation. I am happy to expand further if need be. Thanks for bearing with me.

mxinden:
> I no longer work for actyx (they downsized to 3 employees).

I am sorry to hear that. 

> I won't be able to devote time to this effort for the time being.

Understandable. I hope to be able to test this pull request sometime this or next week.

DemiMarie:
@mxinden @dvc94ch I wholeheartedly agree that the unbounded queues and the `VecDeque` need to go.

kpp:
> I no longer work for actyx (they downsized to 3 employees). I'm currently working on a startup with @rklaehn and @wngr and as such my priorities have shifted and I won't be able to devote time to this effort for the time being.

Should I open a new PR because I can push to this PR but to David'd branch through an intermediate PR?

@dvc94ch David, you didn't add license headers neither to https://github.com/ipfs-rust/libp2p-quic nor here. Would you please post license headers you prefer? I appreciate your work on this project, I can't add "Parity 2021". That would be inconvenient.

dvc94ch:
Feel free to open a new PR. I guess as license header you can add `Cloudpeer 2021`

kpp:
`Copyright 2021 Cloudpeer.` ?

dvc94ch:
not officially founded yet, hq will be in romania, but I guess we can start establishing it as a good open source player.

kpp:
Well, I believe the header should state the company you were working on while writing this code. Or your name if it was your pet project.

dvc94ch:
then David Craven if you prefer

PR #2161, opened 2021-07-27
IPv6 support for mDNS

dvc94ch:
> well it's multicast. so you are doubling the traffic (request sent to n peers and n responses to n peers is n + n2 packets. if you do the same thing on ip6 it's 2(n + n2) packets). if you are only listening on ip4 addresses, then it doesn't hurt as much as no one will respond to ipv6 mdns requests. (for the ip4 only case that would be 2n + n2)

cc @majchrzamemil I think the changes are reasonable, rebased your PR. But as I explained in #1989 enabling them both at the same time should remain optional and preferably only ip4 addresses should be broadcast on ip4 and ip6 addresses on ip6.

revived this effort as I have to do some work on mdns anyway. currently the timeout will expire on all machines at the same time flooding the network, but decided that that should be a separate PR.

### DIFF
--- a/protocols/mdns/CHANGELOG.md
+++ b/protocols/mdns/CHANGELOG.md
@@ -1,6 +1,8 @@
 # 0.32.0 [unreleased]
 
 - Update dependencies.
+- Add support for IPv6. To enable set the multicast address
+  in `MdnsConfig` to `IPV6_MDNS_MULTICAST_ADDRESS`.
 
 # 0.31.0 [2021-07-12]
 
--- a/protocols/mdns/CHANGELOG.md
+++ b/protocols/mdns/CHANGELOG.md
@@ -1,8 +1,12 @@
 # 0.32.0 [unreleased]
 
 - Update dependencies.
+
 - Add support for IPv6. To enable set the multicast address
   in `MdnsConfig` to `IPV6_MDNS_MULTICAST_ADDRESS`.
+  See [PR 2161] for details.
+  
+[PR 2161]: https://github.com/libp2p/rust-libp2p/pull/2161/
 
 # 0.31.0 [2021-07-12]
 
--- a/protocols/mdns/src/behaviour.rs
+++ b/protocols/mdns/src/behaviour.rs
@@ -38,7 +38,7 @@ use std::{
     cmp,
     collections::VecDeque,
     fmt, io, iter,
-    net::{IpAddr, Ipv4Addr, SocketAddr, UdpSocket},
+    net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr, UdpSocket},
     pin::Pin,
     task::Context,
     task::Poll,
@@ -46,8 +46,9 @@ use std::{
 };
 
 lazy_static! {
-    static ref IPV4_MDNS_MULTICAST_ADDRESS: SocketAddr =
-        SocketAddr::from((Ipv4Addr::new(224, 0, 0, 251), 5353));
+    static ref IPV4_MDNS_MULTICAST_ADDRESS: IpAddr = IpAddr::V4(Ipv4Addr::new(224, 0, 0, 251));
+    static ref IPV6_MDNS_MULTICAST_ADDRESS: IpAddr =
+        IpAddr::V6(Ipv6Addr::new(0xFF02, 0, 0, 0, 0, 0, 0, 0xFB));
 }
 
 /// Configuration for mDNS.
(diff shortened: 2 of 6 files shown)


protocols/mdns/src/behaviour.rs
```diff
@@ -118,24 +122,42 @@ pub struct Mdns {
 
     /// Discovery timer.
     timeout: Timer,
+
+    // Mutlicast address.
```
mxinden, line 126:
```suggestion
    // Multicast address.
```

mxinden (commented):
Looks good to me overall. Thanks for picking this back up!

I wonder whether it makes sense to add a simple test with two in-process nodes discovering each other. That should be portable enough to work on most Laptops and our CI, right? What do you think @dvc94ch? Would be great to have some basic tests.

protocols/mdns/tests/smoke.rs
```diff
@@ -0,0 +1,100 @@
...
+    let mut a = create_swarm(config.clone()).await?;
+    let mut b = create_swarm(config).await?;
+    let mut discovered = false;
+    loop {
+        futures::select! {
+            ev = a.select_next_some() => match ev {
+                SwarmEvent::Behaviour(MdnsEvent::Discovered(peers)) => {
+                    for (peer, _addr) in peers {
+                        if peer == *b.local_peer_id() {
+                            if discovered {
```
mxinden, line 50: What if `a` discovers `b` twice but `b` never discovers `a`. The test would still succeed, right?

protocols/mdns/tests/smoke.rs
```diff
@@ -0,0 +1,100 @@
...
+    }
+}
+
+#[async_std::test]
+async fn test_discovery_asyncstd_ipv4() -> Result<(), Box<dyn Error>> {
+    run_test(MdnsConfig::default()).await
+}
+
+#[async_std::test]
+async fn test_discovery_asyncstd_ipv6() -> Result<(), Box<dyn Error>> {
```
mxinden, line 84:
```suggestion
async fn test_discovery_async_std_ipv6() -> Result<(), Box<dyn Error>> {
```

protocols/mdns/tests/smoke.rs
```diff
@@ -0,0 +1,100 @@
...
+                    }
+                }
+                _ => {}
+            }
+        }
+    }
+}
+
+#[async_std::test]
+async fn test_discovery_asyncstd_ipv4() -> Result<(), Box<dyn Error>> {
```
mxinden, line 79:
```suggestion
async fn test_discovery_async_std_ipv4() -> Result<(), Box<dyn Error>> {
```

protocols/mdns/src/behaviour.rs
```diff
@@ -279,31 +296,58 @@ impl NetworkBehaviour for Mdns {
...
+                                } else {
+                                    self.timeout
+                                        .set_interval_at(Instant::now(), self.query_interval);
+                                }
+                            }
+                        }
+                        IpAddr::V6(multicast) => {
+                            if let IpAddr::V6(addr) = inet.addr() {
+                                log::trace!("joining multicast on iface {}", addr);
+                                // TODO: determine interface ID now it generates errors.
```
mxinden, line 320: I don't follow. Can you expand on this TODO?

mxinden (commented):
Wow, that was fast. Thanks @dvc94ch for adding the tests!

protocols/mdns/src/behaviour.rs
dvc94ch: not sure, this was written by @majchrzamemil

dvc94ch: I assume it's related to handling multiple network interfaces for which we have an issue open already, so I'll remove the TODO.

mxinden (commented):
Great. Only needs a changelog entry, other than that this is good to go.

protocols/mdns/CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 # 0.32.0 [unreleased]
 
 - Update dependencies.
+- Add support for IPv6. To enable set the multicast address
+  in `MdnsConfig` to `IPV6_MDNS_MULTICAST_ADDRESS`.
```
mxinden, line 5:
```suggestion
- Update dependencies.

- Add support for IPv6. To enable set the multicast address
  in `MdnsConfig` to `IPV6_MDNS_MULTICAST_ADDRESS`.
  See [PR 2161] for details.
  
[PR 2161]: https://github.com/libp2p/rust-libp2p/pull/2161/
```

mxinden (approved):

PR #2162, opened 2021-07-27
mdns: Refactor encoding/decoding

dvc94ch:
cc @rkuhn


rklaehn:
I reviewed the part where it randomizes the delay. Looks OK. The IPV6 stuff I am not sufficiently familiar with.

dvc94ch:
The ipv6 stuff is handled in a different PR #2161

rkuhn:
Is this related to the “devices not syncing” issue Oli showed you recently? (Will review tomorrow)

dvc94ch:
No, this is related to https://github.com/Actyx/Cosmos/issues/6645 .

I don't know why discovery is failing for oli. His problem was connecting mac and windows which was resolved after a restart. I told him how to debug it and posted it again in the banyan channel incase someone else has a similar issue. But as far as I know we don't support mac anyway?

I have not been able to reproduce an issue with mdns on linux so far and I have tested it on windows in the past, but I did discover a major problem with our discovery protocol which I haven't yet decided how to fix. Not sure if it is related to oli's issue, probably not.

rklaehn:
> The ipv6 stuff is handled in a different PR #2161

For the benefit of the reviewers, it would be then preferable to take it out of this PR, or make this a PR against the other branch.

rklaehn (approved):

protocols/mdns/src/behaviour.rs
```diff
@@ -118,27 +122,52 @@ pub struct Mdns {
...
-            let socket = UdpSocket::from(socket);
-            socket.set_multicast_loop_v4(true)?;
-            socket.set_multicast_ttl_v4(255)?;
-            Async::new(socket)?
+        let recv_socket = match config.multicast_addr {
+            IpAddr::V4(_) => {
+                let socket = Socket::new(Domain::IPV4, Type::DGRAM, Some(socket2::Protocol::UDP))?;
+                socket.set_reuse_address(true)?;
+                #[cfg(unix)]
+                socket.set_reuse_port(true)?;
```
rkuhn, line 138: If two processes using this approach run on the same host, either of them sees only a stochastic selection of half of the received datagrams — isn’t this going to break some assumptions?

protocols/mdns/src/behaviour.rs
```diff
@@ -164,10 +194,19 @@ impl Mdns {
...
+    fn fire_timer(&mut self) {
+        self.timeout
+            .set_interval_at(Instant::now(), self.query_interval);
+    }
+
     fn inject_mdns_packet(&mut self, packet: MdnsPacket, params: &impl PollParameters) {
         match packet {
             MdnsPacket::Query(query) => {
-                self.timeout.set_interval(self.query_interval);
+                self.reset_timer();
```
rkuhn, line 209: this removes the jitter again, doesn’t it?

rkuhn (commented):
What is the intended scope of this change, how much is it intended to fix?

protocols/mdns/src/behaviour.rs
dvc94ch: no, the jitter is computed during mdns initialization. I don't think it makes much difference if it's different every single time

protocols/mdns/src/behaviour.rs
dvc94ch: I think you're forgetting that it's multicast. So all sockets get all messages [0]

- [0] https://stackoverflow.com/questions/18443004/usage-of-so-reuseport-with-multicast-udp

protocols/mdns/src/behaviour.rs
```diff
@@ -150,7 +179,8 @@ impl Mdns {
             events: Default::default(),
             query_interval: config.query_interval,
```
rkuhn, line 180: This needs to have the jitter as well, otherwise reset_timer and fire_timer remove the jitter again.

protocols/mdns/src/behaviour.rs
rkuhn: computed but not stored in `self.query_interval`, as commented above

protocols/mdns/src/behaviour.rs
rkuhn: While what you say sounds plausible, I have been unable to find any normative reference that describes the interplay between IP_ADD_MEMBERSHIP and SO_REUSEPORT. The SO link you posted also contains the same question, unanswered.

protocols/mdns/src/behaviour.rs
dvc94ch: Nice catch, yes, that won't work

protocols/mdns/src/behaviour.rs
dvc94ch: I guess experimental validation of a hypothisis is sufficient (so we don't have to dive into the kernel sources)?

Running two instances of `export RUST_LOG=libp2p_mdns cargo run --example mdns-passive-discovery` on the same machine should convince you that both receive the request.

dvc94ch:
@mxinden what do you think about this one? I think the first commit is uncontroversial, the second one removes the dependency dns-parser that has not been updated in 4y. I think the new implementation is easier to read, supports utf8 (dns-parser erroneously errors if strings are not ascii) and at least for me easier to maintain/bugfix. I tested it with the previous rust-libp2p, so it should be compatible and if that one was compatible with go/js then it should still be compatible too.

protocols/mdns/src/behaviour.rs
```diff
@@ -155,6 +152,13 @@ impl Mdns {
             Async::new(socket)?
         };
         let if_watch = if_watch::IfWatcher::new().await?;
+        // randomize timer to prevent all converging and firing at the same time.
```
mxinden, line 155:
```suggestion
        // Randomize timer to prevent timers across nodes to converge and fire at the same time.
```

protocols/mdns/src/behaviour.rs
```diff
@@ -155,6 +152,13 @@ impl Mdns {
...
+        let query_interval = {
+            use rand::Rng;
+            let mut rng = rand::thread_rng();
+            let jitter = rng.gen_range(0..100);
+            config.query_interval + Duration::from_millis(jitter)
```
mxinden, line 160: Why increasing the interval itself instead of offsetting the initial trigger only and continuing on a fixed interval from there on? Wouldn't that suffice?

protocols/mdns/src/behaviour.rs
```diff
@@ -181,55 +185,74 @@ impl Mdns {
...
     }
 
-    fn inject_mdns_packet(&mut self, packet: MdnsPacket, params: &impl PollParameters) {
+    fn reset_timer(&mut self) {
+        self.timeout.set_interval(self.query_interval);
+    }
+
+    fn fire_timer(&mut self) {
+        self.timeout
+            .set_interval_at(Instant::now(), self.query_interval);
```
mxinden, line 194: Given that this function does not add any randomness, wouldn't that somewhat synchronize intervals across nodes again?

mxinden (commented):
Just a couple of small comments. I will need a bit more time for an in-depth review.

protocols/mdns/src/behaviour.rs
dvc94ch: They converge to firing at the same time if the interval is the same for all nodes, due to the conditions that cause the timers to reset being the same. We can add randomness every time we reset the timer or just give nodes a slightly different interval once. I don't think it makes a big difference, since fairness is not required. It doesn't matter if the same node always wins.

mxinden: I missed that `self.query_interval` instead of `config.query_interval` is used from here on.  Thanks for the clarification!

mxinden (commented):
> I think the first commit is uncontroversial, the second one removes the dependency dns-parser that has not been updated in 4y.

I am in favor of the first change. I am at least hesitant about the second. I would need to do more research on whether a self-written DNS parser is the right way forward for rust-libp2p.

Would you mind splitting the second change out into its own pull request?

dvc94ch:
@mxinden I'll close this PR for now

PR #2163, opened 2021-07-27
KademliaRecordFiltering and filter events

### LINKED ISSUE #2140: Kademlia: record validation
I'm toying with the Kademlia implementation, and noticed that the [record filtering and validation](https://github.com/libp2p/specs/tree/master/kad-dht#entry-validation) does not seem to have an API, currently.  I assume this is needed in, for instance, IPFS' IPNS, and this is something I also need in my application (validating signatures and clocks).

I currently solved this by implementing a `RecordStore` that filters records and returns `ValueTooLarge` when a PUT request is potentially harmful.

The [Entry Validation API specification](https://github.com/libp2p/specs/tree/master/kad-dht#entry-validation) (same link as above) prescribes/suggests a `select` and `validate` API when storing and merging records. I'm actually quite a fan of the filtering approach in `RecordStore` itself, although not by hijacking an existing error variant.

Does the `rust-libp2p` community see anything wrong with something like

```rust
#[derive(Debug)]                                                           
pub enum Error {                                                           
    /// The store is at capacity w.r.t. the total number of stored records.
    MaxRecords,                                                            
    /// The store is at capacity w.r.t. the total number of stored keys for
    /// provider records.                                                  
    MaxProvidedKeys,                                                       
    /// The value of a record to be stored is too large.                   
    ValueTooLarge,                                                         
    /// The store failed to validate the record.
    RecordValidationError(Option<String>),
}                                                                          
```

... potentially substituting the `Option<String>` by a "real" error of some kind? I think this could be a pretty clean way to comply with entry validation. I'd be glad to submit a PR for this, or to implement something completely different and make a PR for that.

rubdos:
This is a first version of a record filtering system, which filters both PutRecord and AddProvider, on top of the existing ttl and key filters.  First comments would be nice. Currently untested (waiting for my previous PR to be merged, then I can rebase and test :wink: )

- [ ] What about reporting storage failures? Document that the caller should issue `Reset` on the connection?

Fixes #2140

### DIFF
--- a/protocols/kad/CHANGELOG.md
+++ b/protocols/kad/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 - Update dependencies.
 
+- Introduce `KademliaStoreInserts` option, which allows to record filtering (see
+  [PR 2163]).
+
+[PR 2163]: https://github.com/libp2p/rust-libp2p/pull/2163
+
 # 0.31.0 [2021-07-12]
 
 - Update dependencies.
--- a/protocols/kad/CHANGELOG.md
+++ b/protocols/kad/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 - Update dependencies.
 
-- Introduce `KademliaStoreInserts` option, which allows to record filtering (see
+- Introduce `KademliaStoreInserts` option, which allows to filter records (see
   [PR 2163]).
 
 [PR 2163]: https://github.com/libp2p/rust-libp2p/pull/2163
--- a/protocols/kad/src/behaviour.rs
+++ b/protocols/kad/src/behaviour.rs
@@ -131,6 +131,32 @@ pub enum KademliaBucketInserts {
     Manual,
 }
 
+/// The configurable filtering strategies for the acceptance of
+/// incoming records.
+///
+/// This can be used for e.g. signature verification or validating
+/// the accompanying [`Key`].
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub enum KademliaRecordFiltering {
+    /// Whenever a (provider) record is received,
+    /// the record is forwarded immediately to the [`RecordStore`].
+    Unfiltered,
+    /// Whenever a record is received, an event of type [`KademliaEvent::InboundPutRecordRequest`] is emitted.
+    /// The event loop should then asynchronously validate this record,
+    /// and if deemed correct, should call [`Kademlia::put_record`].
+    ///
+    /// Provider records are forwarded directly to the [`RecordStore`].
+    FilterRecords,
+    /// Whenever a (provider) record is received, an event is emitted.
+    /// Provider records generate a [`KademliaEvent::InboundAddProviderRequest`],
+    /// normal records generate a [`KademliaEvent::InboundPutRecordRequest`].
+    ///
+    /// The event loop should then asynchronously validate this record,
+    /// and if deemed correct, should call [`Kademlia::put_record`] or
+    /// [`Kademlia::add_provider`], whichever is applicable.
+    FilterBoth,
+}
+
 /// The configuration for the `Kademlia` behaviour.
 ///
 /// The configuration is consumed by [`Kademlia::new`].
@@ -142,6 +168,7 @@ pub struct KademliaConfig {
     record_ttl: Option<Duration>,
     record_replication_interval: Option<Duration>,
     record_publication_interval: Option<Duration>,
+    record_filtering: KademliaRecordFiltering,
     provider_record_ttl: Option<Duration>,
     provider_publication_interval: Option<Duration>,
     connection_idle_timeout: Duration,
@@ -175,6 +202,7 @@ impl Default for KademliaConfig {
             record_ttl: Some(Duration::from_secs(36 * 60 * 60)),
             record_replication_interval: Some(Duration::from_secs(60 * 60)),
             record_publication_interval: Some(Duration::from_secs(24 * 60 * 60)),
+            record_filtering: KademliaRecordFiltering::Unfiltered,
             provider_publication_interval: Some(Duration::from_secs(12 * 60 * 60)),
             provider_record_ttl: Some(Duration::from_secs(24 * 60 * 60)),
             connection_idle_timeout: Duration::from_secs(10),
@@ -256,6 +284,15 @@ impl KademliaConfig {
     /// Does not apply to provider records.
     pub fn set_record_ttl(&mut self, record_ttl: Option<Duration>) -> &mut Self {
         self.record_ttl = record_ttl;
+        self
+    }
+
+    /// Sets whether or not records should be filtered before being stored.
+    ///
+    /// See [`KademliaRecordFiltering`] for the different values.
+    /// Defaults to [`KademliaRecordFiltering::Unfiltered`].
+    pub fn set_record_filtering(&mut self, filtering: KademliaRecordFiltering) -> &mut Self {
+        self.record_filtering = filtering;
         self
     }
 
--- a/protocols/kad/src/behaviour.rs
+++ b/protocols/kad/src/behaviour.rs
@@ -2384,7 +2384,7 @@ pub enum KademliaEvent {
 
     /// A peer sent a [`KademliaHandlerIn::PutRecord`] request and filtering is enabled.
     ///
-    /// Cfr. [`KademliaRecordFiltering`] and [`KademliaConfig::set_record_filtering`].
+    /// See [`KademliaRecordFiltering`] and [`KademliaConfig::set_record_filtering`].
     InboundPutRecordRequest {
         source: PeerId,
         connection: ConnectionId,
--- a/protocols/kad/src/behaviour.rs
+++ b/protocols/kad/src/behaviour.rs
@@ -156,9 +156,7 @@ pub enum KademliaRecordFiltering {
     /// Provider records generate a [`KademliaEvent::InboundAddProviderRequest`],
     /// normal records generate a [`KademliaEvent::InboundPutRecordRequest`].
     ///
-    /// The event loop should then asynchronously validate this record,
-    /// and if deemed correct, should call [`RecordStore::put`] or
-    /// [`RecordStore::add_provider`], whichever is applicable.
+    /// When deemed valid, a (provider) record needs to be explicitly stored in the [`RecordStore`] via [`RecordStore::put`] or [`RecordStore::add_provider`], whichever is applicable. A mutable reference to the [`RecordStore`] can be retrieved via [`Kademlia::store_mut`].
     FilterBoth,
 }
 
--- a/protocols/kad/src/behaviour.rs
+++ b/protocols/kad/src/behaviour.rs
@@ -69,6 +69,9 @@ pub struct Kademlia<TStore> {
     /// Configuration of the wire protocol.
     protocol_config: KademliaProtocolConfig,
 
+    /// Configuration of [`RecordStore`] filtering.
+    record_filtering: KademliaRecordFiltering,
+
     /// The currently active (i.e. in-progress) queries.
     queries: QueryPool<QueryInner>,
 
@@ -430,6 +433,7 @@ where
             kbuckets: KBucketsTable::new(local_key, config.kbucket_pending_timeout),
             kbucket_inserts: config.kbucket_inserts,
             protocol_config: config.protocol_config,
+            record_filtering: config.record_filtering,
             queued_events: VecDeque::with_capacity(config.query_config.replication_factor.get()),
             queries: QueryPool::new(config.query_config),
             connected_peers: Default::default(),
--- a/protocols/kad/src/behaviour.rs
+++ b/protocols/kad/src/behaviour.rs
@@ -1610,17 +1610,8 @@ where
             // The record is cloned because of the weird libp2p protocol
             // requirement to send back the value in the response, although this
             // is a waste of resources.
-            if self.record_filtering != KademliaStoreInserts::Unfiltered {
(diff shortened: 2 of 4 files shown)


rubdos:
I've tested it now, by adding
```rust
KademliaEvent::InboundPutRecordRequest { source, record, .. }
    log::info!(                                              
        "Inbound put record request from {}, record {:?}",   
        source,                                              
        record                                               
    );                                                       
    self.kademlia.store_mut().put(record).ok();              
}                                                            
```
to my `NetworkBehaviourEventProcess<KademliaEvent>` handler. Seems to work as intended.

protocols/kad/src/behaviour.rs
```diff
@@ -2192,7 +2251,23 @@ pub enum KademliaEvent {
...
+    /// Cfr. [`KademliaRecordFiltering`] and [`KademliaConfig::set_record_filtering`].
+    InboundPutRecordRequest {
+        source: PeerId,
+        connection: ConnectionId,
+        record: Record,
+    },
+
+    /// A peer sent a [`KademliaHandlerIn::AddProvider`] request and filtering [`KademliaRecordFiltering::FilterBoth`] is enabled.
+    ///
+    /// Cfr. [`KademliaRecordFiltering`] and [`KademliaConfig::set_record_filtering`].
```
mxinden, line 2267:
```suggestion
    /// See [`KademliaRecordFiltering`] and [`KademliaConfig::set_record_filtering`] for details..
```
Just for the sake of consistency.

protocols/kad/src/behaviour.rs
```diff
@@ -2192,7 +2251,23 @@ pub enum KademliaEvent {
     PendingRoutablePeer {
         peer: PeerId,
         address: Multiaddr,
-    }
+    },
+
+    /// A peer sent a [`KademliaHandlerIn::PutRecord`] request and filtering is enabled.
+    ///
+    /// Cfr. [`KademliaRecordFiltering`] and [`KademliaConfig::set_record_filtering`].
```
mxinden, line 2258:
```suggestion
    /// See [`KademliaRecordFiltering`] and [`KademliaConfig::set_record_filtering`].
```

protocols/kad/src/behaviour.rs
```diff
@@ -2192,7 +2251,23 @@ pub enum KademliaEvent {
...
+        connection: ConnectionId,
+        record: Record,
+    },
+
+    /// A peer sent a [`KademliaHandlerIn::AddProvider`] request and filtering [`KademliaRecordFiltering::FilterBoth`] is enabled.
+    ///
+    /// Cfr. [`KademliaRecordFiltering`] and [`KademliaConfig::set_record_filtering`].
+    InboundAddProviderRequest {
+        record: ProviderRecord,
+    },
```
mxinden, line 2270: How about moving these above `InboundRequestServed` as they are related to inbound requests and to routing updates?

mxinden (commented):
Just a couple of smaller suggestions. This looks great already, though it would need at least one unit test and changelog entry.

protocols/kad/src/behaviour.rs
```diff
@@ -131,6 +134,34 @@ pub enum KademliaBucketInserts {
...
+pub enum KademliaRecordFiltering {
+    /// Whenever a (provider) record is received,
+    /// the record is forwarded immediately to the [`RecordStore`].
+    Unfiltered,
+    /// Whenever a record is received, an event of type [`KademliaEvent::InboundPutRecordRequest`] is emitted.
+    /// The event loop should then asynchronously validate this record,
+    /// and if deemed correct, should call [`RecordStore::put`].
+    ///
+    /// Provider records are forwarded directly to the [`RecordStore`].
+    FilterRecords,
```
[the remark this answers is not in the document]
rubdos: Should the `FilterBoth` then be renamed to `Filter`?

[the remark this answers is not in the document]
rubdos: Looks like my `provider_received` change wasn't even correct... `if self.record_filtering != KademliaStoreInserts::Unfiltered`

rubdos:
Note to self wrt 55cc8d5 , Rust Analyzer "rename" doesn't take into account doc links yet.

I left one last question in your review.

Still remaining:

- [ ] What kind of unit test do we want here? Some kind of copy-paste from the `put` test, which uses the filtering setting instead?
- [ ] What with failed put requests, should we have the end user signal such failure to the requesting node? Currently we return OK, even before it's inserted good and well into the record store.

mxinden (commented):
> Should the FilterBoth then be renamed to Filter?

I am fine either way. I guess `FilterBoth` allows us to introduce `FilterRecord` and `FilterProvider` in the future with only a semi-breaking change.

>  What kind of unit test do we want here? Some kind of copy-paste from the put test, which uses the filtering setting instead?

:+1: 

>  What with failed put requests, should we have the end user signal such failure to the requesting node? Currently we return OK, even before it's inserted good and well into the record store.

I don't think this is necessary and might even open up an attack vector. That said, I am happy to revisit this in case someone has the need for it in the future.

protocols/kad/CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 - Update dependencies.
 
+- Introduce `KademliaStoreInserts` option, which allows to record filtering (see
```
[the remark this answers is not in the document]
rubdos: Absolutely right.

mxinden:
Thank you for the latest changes.

Note that we merged https://github.com/libp2p/rust-libp2p/pull/2188 recently. To reduce merge conflict size you might want to try first formatting `libp2p-kad` and then merging latest `master`.

Only thing missing from my side would be a simple unit test.

rubdos:
Awesome, rustfmt! Thanks for that :-)

Expect a test soon!

rubdos:
> Note that we merged #2188 recently. To reduce merge conflict size you might want to try first formatting `libp2p-kad` and then merging latest `master`.

I've instead rebased and formatted every commit, because `rustfmt`+merge was a daunting merge conflict...

rubdos:
I've added two tests, one that filters out the records and asserts that they cannot be found (this is only true on an honest network, which is the case in the test), and another that persists the record.

I'm not proud of the copy-pasting that's going on, but I feel that refactoring those tests is maybe something for another day?

protocols/kad/src/behaviour/test.rs
```diff
@@ -695,6 +695,407 @@ fn put_record() {
     QuickCheck::new().tests(3).quickcheck(prop as fn(_, _) -> _)
 }
 
+/// A node joining a fully connected network cannot store a record if the nodes ignore it through
```
mxinden, line 698: A bit late to suggest this, sorry about that. Still: How about integrating this test into `put_record` by adding something along the lines of:

```
        if rng.gen() {
             config.set_record_filtering(KademliaStoreInserts::FilterBoth);
        }
```

And then, in case one receives a `InboundPutRecordRequest`, double check that `config.record_filtering` was set to `FilterBoth`.

That would reduce the amount of code required, no?

protocols/kad/src/behaviour/test.rs
```diff
@@ -695,6 +695,407 @@ fn put_record() {
...
+            }
+
+            // All swarms are Pending and not enough results have been collected
+            // so far, thus wait to be polled again for further progress.
+            if results.len() != records.len() {
+                return Poll::Pending;
+            }
+
+            // Consume the results, checking that each record was replicated
+            // correctly to the closest peers to the key.
```
mxinden, line 1051: This is outdated, right?

protocols/kad/src/behaviour/test.rs
```diff
@@ -695,6 +695,407 @@ fn put_record() {
...
+                    kbucket::Key::from(*id1)
+                        .distance(&key)
+                        .cmp(&kbucket::Key::from(*id2).distance(&key))
+                });
+
+                let expected = expected
+                    .into_iter()
+                    .take(replication_factor.get())
+                    .collect::<HashSet<_>>();
+
```
mxinden, line 1077: Is this needed?

protocols/kad/src/behaviour/test.rs
rubdos: Do you have some coverage analysis in place to check that these code paths are hit? Using an rng for that sounds scary :'-)

I'll implement it like that though, sounds a lot cleaner. I'll test it using provider records too.

mxinden: You are right. Let's get rid of the RNG. How about the diff below + optionally setting `set_record_filtering` in the `TestConfig` `Arbitrary` implementation?

```diff
diff --git a/protocols/kad/src/behaviour/test.rs b/protocols/kad/src/behaviour/test.rs
index 110c1e22..b27fd076 100644
--- a/protocols/kad/src/behaviour/test.rs
+++ b/protocols/kad/src/behaviour/test.rs
@@ -488,24 +488,28 @@ fn get_record_not_found() {
 /// is equal to the configured replication factor.
 #[test]
 fn put_record() {
-    fn prop(records: Vec<Record>, seed: Seed) {
-        let mut rng = StdRng::from_seed(seed.0);
-        let replication_factor =
-            NonZeroUsize::new(rng.gen_range(1, (K_VALUE.get() / 2) + 1)).unwrap();
-        // At least 4 nodes, 1 under test + 3 bootnodes.
-        let num_total = usize::max(4, replication_factor.get() * 2);
-
-        let mut config = KademliaConfig::default();
-        config.set_replication_factor(replication_factor);
-        if rng.gen() {
-            config.disjoint_query_paths(true);
+    #[derive(Clone, Debug)]
+    struct TestConfig(KademliaConfig);
+
+    impl Arbitrary for TestConfig {
+        fn arbitrary<G: Gen>(g: &mut G) -> TestConfig {
+            let mut config = KademliaConfig::default();
+            config.set_replication_factor(NonZeroUsize::new(g.gen_range(1, (K_VALUE.get() / 2) + 1)).unwrap());
+            config.disjoint_query_paths(g.gen());
+            TestConfig(config)
         }
+    }
+
+    fn prop(records: Vec<Record>, config: TestConfig) {
+        let config = config.0;
+        // At least 4 nodes, 1 under test + 3 bootnodes.
+        let num_total = usize::max(4, config.query_config.replication_factor.get() * 2);
 
         let mut swarms = {
             let mut fully_connected_swarms =
                 build_fully_connected_nodes_with_config(num_total - 1, config.clone());
 
-            let mut single_swarm = build_node_with_config(config);
+            let mut single_swarm = build_node_with_config(config.clone());
             // Connect `single_swarm` to three bootnodes.
             for i in 0..3 {
                 single_swarm.1.behaviour_mut().add_address(
@@ -584,7 +588,7 @@ fn put_record() {
                         ))) => {
                             assert!(qids.is_empty() || qids.remove(&id));
                             assert!(stats.duration().is_some());
-                            assert!(stats.num_successes() >= replication_factor.get() as u32);
+                            assert!(stats.num_successes() >= config.query_config.replication_factor.get() as u32);
                             assert!(stats.num_requests() >= stats.num_successes());
                             assert_eq!(stats.num_failures(), 0);
                             match res {
@@ -635,7 +639,7 @@ fn put_record() {
 
                 let expected = expected
                     .into_iter()
-                    .take(replication_factor.get())
+                    .take(config.query_config.replication_factor.get())
                     .collect::<HashSet<_>>();
 
                 let actual = swarms
@@ -650,7 +654,7 @@ fn put_record() {
                     })
                     .collect::<HashSet<_>>();
 
-                assert_eq!(actual.len(), replication_factor.get());
+                assert_eq!(actual.len(), config.query_config.replication_factor.get());
 
                 let actual_not_expected = actual.difference(&expected).collect::<Vec<&PeerId>>();
                 assert!(
(END)

```

rubdos: With only 3 tests, I have a test case that doesn't get hit. With four, it does seem to hit all paths. I'll add two boolean parameters to the test properties instead, although I still have to increase the test count.

protocols/kad/src/behaviour/test.rs
rubdos: Not any more with the new approach, I think? Or do you mean to rewrite the comment itself?

mxinden:
> With only 3 tests, I have a test case that doesn't get hit. With four, it does seem to hit all paths. I'll add two boolean parameters to the test properties instead, although I still have to increase the test count.

Note that this one test (`put_record`) includes many more permutations than 3. The goal of quickcheck here is not exhaust all permutations on each execution, but instead probablistically exhaust all permutations across many runs (e.g. locally, on each pull request, on each merge, ...). It is a trait-off between running time and coverage. Just want to make sure there is no confusion. I am not opposed to increasing it to `4`.

I am fine with either the additional `bool` parameters or the `TestConfig`. As you wish.

rubdos:
> Note that this one test (`put_record`) includes many more permutations than 3. The goal of quickcheck here is not exhaust all permutations on each execution, but instead probablistically exhaust all permutations across many runs 

Aha. I got the feeling however that it used a deterministically seeded Rng behind the scenes, because I couldn't repro on 3 but it was stable on 4...

Are you fine with the way it turned out now? I've added two parameters and incremented the testing to 4...

mxinden (approved):
Thanks Ruben for the contributions and for bearing with me :)

rubdos:
> Thanks Ruben for the contributions and for bearing with me :)

Very glad to be of use. It's already been a pleasure to work with you folks!

supercmmetry:
I was thinking if we could refactor `InboundPutRecordRequest` and `InboundAddProviderRequest` into a `FilterRequest` or something similar inside `KademliaEvent`

PR #2169, opened 2021-07-31
Replace `Into` with `From`

thomaseizinger:
Unless restricted by orphan rules, implementing `From` is superior
because it implies `Into` but leaves the choice to the user, which
one to use. Especially for errors, `From` is convenient because that
is what `?` builds on.

Not sure if this needs a changelog entry?

### DIFF
--- a/muxers/yamux/CHANGELOG.md
+++ b/muxers/yamux/CHANGELOG.md
@@ -1,11 +1,9 @@
-# Unreleased
-
-- Implement `From<io::Error> for YamuxError` instead of `Into`.
-  [PR 2169](https://github.com/libp2p/rust-libp2p/pull/2169)
-
 # 0.34.0 [unreleased]
 
 - Update dependencies.
+
+- Implement `From<io::Error> for YamuxError` instead of `Into`.
+  [PR 2169](https://github.com/libp2p/rust-libp2p/pull/2169)
 
 # 0.33.0 [2021-07-12]
 
--- a/muxers/yamux/CHANGELOG.md
+++ b/muxers/yamux/CHANGELOG.md
@@ -1,3 +1,8 @@
+# Unreleased
+
+- Implement `From<io::Error> for YamuxError` instead of `Into`.
+  [PR 2169](https://github.com/libp2p/rust-libp2p/pull/2169)
+
 # 0.34.0 [unreleased]
 
 - Update dependencies.
--- a/misc/multistream-select/CHANGELOG.md
+++ b/misc/multistream-select/CHANGELOG.md
@@ -1,3 +1,8 @@
+# Unreleased
+
+- Implement `From<io::Error> for ProtocolError` instead of `Into`.
+  [PR 2169](https://github.com/libp2p/rust-libp2p/pull/2169)
+
 # 0.10.3 [2021-03-17]
 
 - Update dependencies.
--- a/misc/multistream-select/CHANGELOG.md
+++ b/misc/multistream-select/CHANGELOG.md
@@ -1,4 +1,4 @@
-# Unreleased
+# 0.10.4 [unreleased]
 
 - Implement `From<io::Error> for ProtocolError` instead of `Into`.
   [PR 2169](https://github.com/libp2p/rust-libp2p/pull/2169)
--- a/misc/multistream-select/Cargo.toml
+++ b/misc/multistream-select/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "multistream-select"
 description = "Multistream-select negotiation protocol for libp2p"
-version = "0.10.3"
+version = "0.10.4"
 authors = ["Parity Technologies <admin@parity.io>"]
 license = "MIT"
 repository = "https://github.com/libp2p/rust-libp2p"
--- a/misc/multistream-select/src/protocol.rs
+++ b/misc/multistream-select/src/protocol.rs
@@ -391,9 +391,9 @@ impl From<io::Error> for ProtocolError {
     }
 }
 
-impl Into<io::Error> for ProtocolError {
-    fn into(self) -> io::Error {
-        if let ProtocolError::IoError(e) = self {
+impl From<ProtocolError> for io::Error {
+    fn from(err: ProtocolError) -> Self {
+        if let ProtocolError::IoError(e) = err {
             return e
         }
         io::ErrorKind::InvalidData.into()
--- a/muxers/yamux/src/lib.rs
+++ b/muxers/yamux/src/lib.rs
@@ -360,9 +360,9 @@ where
 #[error("yamux error: {0}")]
 pub struct YamuxError(#[from] yamux::ConnectionError);
 
-impl Into<io::Error> for YamuxError {
-    fn into(self: YamuxError) -> io::Error {
-        match self.0 {
+impl From<YamuxError> for io::Error {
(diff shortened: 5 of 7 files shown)


mxinden (commented):
Cool, thanks!

> Not sure if this needs a changelog entry?

Yes please, as it adds `From` implementations. Also, please bump the patch version of `multistream-select` in `misc/multistream/Cargo.toml` and `./Cargo.toml`.

thomaseizinger:
> > Not sure if this needs a changelog entry?
> 
> Yes please, as it adds `From` implementations. Also, please bump the patch version of `multistream-select` in `misc/multistream/Cargo.toml` and `./Cargo.toml`.

Okay will add.

On the topic of version bumps: What is the reasoning to do these together with the actual change and not on release time? For all my projects, I usually track all changes in the changelog under an _unversioned_ `Unreleased` heading and at the time of a new release, decide based on the changes that were made whether or not it is a major, minor or patch release and update the versions accordingly.

A practical implication of this is that cargo's `[patch]` feature can be used to depend on unreleased versions because it enforces that the version mentioned in the `Cargo.toml` matches the one that is brought in via `[patch]`.

muxers/yamux/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+# Unreleased
+
+- Implement `From<io::Error> for YamuxError` instead of `Into`.
+  [PR 2169](https://github.com/libp2p/rust-libp2p/pull/2169)
+
 # 0.34.0 [unreleased]
 
 - Update dependencies.
```
mxinden, line 8:
```suggestion
# 0.34.0 [unreleased]

- Update dependencies.

- Implement `From<io::Error> for YamuxError` instead of `Into`.
  [PR 2169](https://github.com/libp2p/rust-libp2p/pull/2169)
```

mxinden:
> On the topic of version bumps: What is the reasoning to do these together with the actual change and not on release time? For all my projects, I usually track all changes in the changelog under an _unversioned_ `Unreleased` heading and at the time of a new release, decide based on the changes that were made whether or not it is a major, minor or patch release and update the versions accordingly.
> 
> A practical implication of this is that cargo's `[patch]` feature can be used to depend on unreleased versions because it enforces that the version mentioned in the `Cargo.toml` matches the one that is brought in via `[patch]`.

A couple of arguments that in my eyes speak for the current strategy:

- I am most familiar with a changeset at review / authoring time, not at release time. With that in mind, it is easier for me to determine whether something is a breaking change or not at review / authoring time.
- Doing things right away (at review / authoring time) instead of delayed at release time prevents loosing track of a change. For example, say that this pull request introduced a breaking change to `multistream-select`. If we don't bump the versions now, I might forget the change entirely at release time, thus pushing a faulty `libp2p-core` release. If we bump the version now, `cargo publish` of `libp2p-core` fails, given that the new version of `multistream-select` is not yet uploaded. With the fact in mind that each release cycle roughly contains 20 sub releases, this is a major bug source, e.g. see https://github.com/libp2p/rust-libp2p/pull/2136 and https://github.com/libp2p/rust-libp2p/commit/40c4287bc9a95461889e10612519960a16a2874f.

> A practical implication of this is that cargo's `[patch]` feature can be used to depend on unreleased versions because it enforces that the version mentioned in the `Cargo.toml` matches the one that is brought in via `[patch]`.

Indeed, this is a downside of the current approach, though in my eyes a cost worth paying given the extra safety at release time.

What do you think of the arguments above?

mxinden (commented):
Missing the `Cargo.toml` bumps. Other than that, this is good to go.

thomaseizinger:
> > On the topic of version bumps: What is the reasoning to do these together with the actual change and not on release time? For all my projects, I usually track all changes in the changelog under an _unversioned_ `Unreleased` heading and at the time of a new release, decide based on the changes that were made whether or not it is a major, minor or patch release and update the versions accordingly.
> > A practical implication of this is that cargo's `[patch]` feature can be used to depend on unreleased versions because it enforces that the version mentioned in the `Cargo.toml` matches the one that is brought in via `[patch]`.
> 
> A couple of arguments that in my eyes speak for the current strategy:
> 
>  * I am most familiar with a changeset at review / authoring time, not at release time. With that in mind, it is easier for me to determine whether something is a breaking change or not at review / authoring time.
> 
> * Doing things right away (at review / authoring time) instead of delayed at release time prevents loosing track of a change. For example, say that this pull request introduced a breaking change to `multistream-select`. If we don't bump the versions now, I might forget the change entirely at release time, thus pushing a faulty `libp2p-core` release. If we bump the version now, `cargo publish` of `libp2p-core` fails, given that the new version of `multistream-select` is not yet uploaded. With the fact in mind that each release cycle roughly contains 20 sub releases, this is a major bug source, e.g. see [swarm-derive: Update to v0.24.0 #2136](https://github.com/libp2p/rust-libp2p/pull/2136) and [40c4287](https://github.com/libp2p/rust-libp2p/commit/40c4287bc9a95461889e10612519960a16a2874f).

That does make sense! I am not particularly familiar with releasing within workspaces that have lots of crates, so I can see how this can be an issue.

thomaseizinger:
> Cool, thanks!
> 
> > Not sure if this needs a changelog entry?
> 
> Yes please, as it adds `From` implementations. Also, please bump the patch version of `multistream-select` in `misc/multistream/Cargo.toml` and `./Cargo.toml`.

I wasn't quite sure what you are referring to with `./Cargo.toml` as multistream-select is not mentioned in there. I bumped the version in `misc/multistream-select/Cargo.toml`.

mxinden:
> > Cool, thanks!
> > > Not sure if this needs a changelog entry?
> > 
> > 
> > Yes please, as it adds `From` implementations. Also, please bump the patch version of `multistream-select` in `misc/multistream/Cargo.toml` and `./Cargo.toml`.
> 
> I wasn't quite sure what you are referring to with `./Cargo.toml` as multistream-select is not mentioned in there. I bumped the version in `misc/multistream-select/Cargo.toml`.

Ah, sorry. Please bump the version in `core/Cargo.toml`. My bad.

thomaseizinger:
> > > Cool, thanks!
> > > > Not sure if this needs a changelog entry?
> > > 
> > > 
> > > Yes please, as it adds `From` implementations. Also, please bump the patch version of `multistream-select` in `misc/multistream/Cargo.toml` and `./Cargo.toml`.
> > 
> > 
> > I wasn't quite sure what you are referring to with `./Cargo.toml` as multistream-select is not mentioned in there. I bumped the version in `misc/multistream-select/Cargo.toml`.
> 
> Ah, sorry. Please bump the version in `core/Cargo.toml`. My bad.

Shouldn't be necessary, `core/Cargo.toml` only specifies its dependency using a minor version and as far as I know, cargo by default uses `^` as the semver matcher which takes the highest available patch version.

mxinden:
> Shouldn't be necessary, core/Cargo.toml only specifies its dependency using a minor version and as far as I know, cargo by default uses ^ as the semver matcher which takes the highest available patch version.

Correct. It is a reminder for me to make sure I release `multistream-select` with the next `libp2p` release. That said, it is more of a hack, especially as it unnecessarily restricts the number of versions compatible with `libp2p-core` and thus `libp2p`.

Let's leave it as is.

mxinden (approved):
Thanks!

PR #2175, opened 2021-08-02
[Gossipsub] Inconsistency in mesh peer tracking

AgeManning:
Peers that GRAFT to us without sending subscriptions for the grafted topic can be added to the mesh. These peers once disconnected, do then not get removed from the mesh. This creates a variety of inconsistencies. 

This PR resolves this by adding the topics that a peer send us via grafts, to the `peer_topics` mapping, indicating that they are in fact subscribed to that topic. There should never be a case that a peer grafts us, without locally being subscribed to the topic.

In a similar vein, this improves the consistency of blacklisted peers that the user can add/remove at arbitrary times.


protocols/gossipsub/src/behaviour.rs
```diff
@@ -2855,7 +2863,7 @@ where
                 Some(topics) => (topics),
                 None => {
                     if !self.blacklisted_peers.contains(peer_id) {
-                        debug!("Disconnected node, not in connected nodes");
+                        error!("Disconnected node, not in connected nodes");
```
mxinden, line 2866: How about an additional `debug_assert` here?

protocols/gossipsub/src/behaviour.rs
```diff
@@ -2913,11 +2921,6 @@ where
         connection_id: &ConnectionId,
         endpoint: &ConnectedPoint,
     ) {
-        // Ignore connections from blacklisted peers.
```
mxinden, line 2916: Can you expand on why we no longer ignore blacklisted peers?

AgeManning: I found that in many cases we are under the assumption that any connected peer exists in `connected_peers` and their topic subscriptions exist in `peer_topics`. When peers disconnect we use these mappings to correctly remove them from the mesh and other mappings in the behaviour. If there is any inconsistencies, gossipsub panics (this PR fixes on such instance). 

Although I haven't seen a panic from blacklisted peers (I'm not using them atm), i'm concerned about inconsistencies coming from states where users at arbitrary times blacklist and unblacklist (via the public functions) peers. 

This change represents a safer approach where we register the connections of all peers, add them to the mappings (so they exist regardless of when the user ad-hoc add/removes them) but we still block all messages for whatever peers are blacklisted at the time we receive them. On disconnects, as all peers connections are registered they are removed from all mappings correctly, regardless of the current state of blacklisting.

There probably is a way we could do this without registering the connection, but I think there a quite a few edge cases and this makes it simpler to reason about.

mxinden: Thanks @AgeManning for the details!

mxinden (approved):
Thanks @AgeManning for upstreaming the patch!

mxinden:
@AgeManning I don't have permissions to merge latest `master` via the UI. Could you either (a) grant me that permission or (b) merge yourself?

AgeManning:
Thanks @mxinden. In the future, I'll make these PRs fully accessible to you.

paulhauner:
We're on a bit of a tight timeline over at sigp/lighthouse with a hard-fork happening on a public testnet next week. This is the last blocker before we can get a release candidate out.

I appreciate that the maintainers have their own priorities, I just wanted to make that known 🙂 As always, thanks for your efforts and I appreciate that reviews can take time 🙏

paulhauner:
Thank you, much appreciated!

> do I understand correctly that you only need this patch to be merged into master, or would you need a new release as well?

We're fine with targeting `master` for the time being. No rush to push a release :)

PR #2180, opened 2021-08-04
Reexport `wasm-bindgen` feature from `parking_lot`

### LINKED ISSUE #2143: Enable `wasm-bindgen` feature for the `parking_lot` dependency
I am building `rust-libp2p` as part of our Wasm bindings. The produced WebAssembly module ends up containing this import and call to a `now` function:

```
(import "env" "now" (func $now (type 3)))
...
(func $instant::wasm::now::h339f51d64be03d28 (type 3) (result f64)
  call $now)
```

This import then ends up in the `wasm-bindgen`-generated JavaScript file where the `env` import is not found and an error is produced.

According to [this comment](https://github.com/rustwasm/wasm-bindgen/issues/2215#issuecomment-796244209), this is the result of the dependency on `parking_lot`, where the fix is to enable the `wasm-bindgen` feature on that crate.

If the feature is enabled in `rust-libp2p`'s `Cargo.toml`, then the import is no longer present and the issue is fixed.

I assume the best way to fix this, would be to enable the `wasm-bindgen` feature of the `parking_lot` crate when `rust-libp2p`'s `wasm-ext` feature is enabled. However, I'm not sure if I'm not actually missing something, as this seems like an issue that should have come up in the past when compiling to `wasm32-unknown-unknown`.

Please let me know if you would like a PR that adds this or if there's another way to work around it.

PhilippGackstatter:
Allow opting-in to the `wasm-bindgen` feature of `parking_lot` for better Wasm support. Bubbles up the feature as per the discussion in the corresponding issue.

Because this feature should be opt-in for Wasm targets, it is not enabled by default.

fixes #2143


mxinden (commented):
Thanks @PhilippGackstatter. While not directly intuitive I am in favor of being consistent with `parking_lot`.

Can you include an entry in the top level `CHANGELOG.md` in the "Version 0.40.0 [unreleased]" section?

PhilippGackstatter:
> Can you include an entry in the top level `CHANGELOG.md` in the "Version 0.40.0 [unreleased]" section?

Sure, all done.

mxinden (approved):
Thanks @PhilippGackstatter.

Issue #2173, opened 2021-08-02
Re-design feature sets


thomaseizinger:
Opening a new issue here to track the proposal discussed in https://github.com/libp2p/rust-libp2p/pull/2146.

cc @mxinden @dvc94ch 

## Summary

An opt-in based feature design is preferable over the current one due to how cargo's features work (once turned on, cannot be turned off in a dependency graph).

## Implementation plan

- [x] https://github.com/libp2p/rust-libp2p/pull/2918
- [x] https://github.com/libp2p/rust-libp2p/pull/2962
- [ ] Align naming of executor specific structs like for the tcp transport, likely by namespacing things into modules. This effort is tracked in https://github.com/libp2p/rust-libp2p/issues/2217.
- [ ] Only loosely related: Making the API of `Swarm::new` more mis-use resistent by forcing users to choose an executor, thus fixing https://github.com/libp2p/rust-libp2p/issues/2230.

## Minimal concrete changes

- No features on by default
- A `full` feature that activates everything, for ease of transition from the current design

## Extended changes (to be discussed)

1. Have a `development` feature that activates stuff like the `development` transport (and necessary crates).

The advantage over `full` with this one is that you can still cut down compile-times for tests but have the convenience of having most things available needed for writing tests.

2. Introduce dedicated executor features: I would like to see dedicated `tokio` and `async-std` features exposed by the top-level `libp2p` meta crate.

These features would:

- Forward to the dedicated protocol crates (i.e. `libp2p-tcp/tokio`)
- Automatically set the executor accordingly. At the moment, this [defaults to a futures-threadpool](https://github.com/libp2p/rust-libp2p/blob/841b09aa694a1eca6d2bde83f9c6ac8f247b4a50/swarm/src/lib.rs#L1058-L1072) which requires setting an executor explicitly if the one wants to use `tokio`, otherwise things are going to not work at runtime. `libp2p-swarm` could expose a feature per async-runtime that sets these executors automatically to the correct one.

As part of this re-design, we could introduce a more coherent naming. For example, at the moment `libp2p-tcp` exposes a `TcpConfig` and a `TokioTcpConfig`. If we expose executor features, it would be more coherent to name all respective configs after the executor-backend they are based on, i.e. `AsyncIoTcpConfig` and `TokioTcpConfig`.

3. Add an `ipfs` feature that activates ipfs-specific crates, f.e. `secp256k1` support in libp2p-core.

tomaka:
[...]

PR #2181, opened 2021-08-06
Make libp2p-core default features optional in all subcrates

rubdos:
It's impossible to use `libp2p` as dependency without having `libp2p-core` with all its default features (example use case: I don't need/want `secp256k1`). The reason is that `libp2p` requires (a.o.) `libp2p-swarm`, which in turn has `libp2p-core` with default features.

This is probably a backwards incompatible change. I would suggest that `libp2p` forwards its own `default` feature to `libp2p-*`.

Additionally, we probably need to discuss which subcrates require what features from libp2p-core, and how to make sure that CI still covers the common situations in terms of feature flags.


rubdos:
image

Ater the 888314e fixup, I got rid of secp :-)

thomaseizinger:
Relevant discussion: https://github.com/libp2p/rust-libp2p/pull/2146

rubdos:
> Relevant discussion: #2146

Ah, missed that. Seems like this PR then works towards #2173, and that default can soon be set to `[]` and a `full` feature flag can be added.

> Given lack of my own involvement on this front, I am more than happy to support?merging a subset of what is discussed in that PR.
>
> Generally, I think we should embrace opt-in more than opt-out. What do things look like if we just remove secp256k1 as a feature from libp2p-core?

Removing secp256k1 from core means that the default feature set is incompatible with things like IPFS, that might be not desirable.

Additional to `full`, you could consider a `ipfs` feature flag too, pulling in the standard ipfs-features... Just a suggestion though, not a thing that I'd take up.

thomaseizinger:
Good idea, I'll add the proposal to issue!

rubdos:
> I am in favor of this step towards #2173.

Does this mean "I am in favor of this PR as a step towards #2173", or "I am in favor of discussing this approach in context of #2173"? Not sure how to read you here.

> > This is probably a backwards incompatible change.
> 
> Yes. It would require a changelog entry.

Ack, will add after resolving the previous point :-)

> > I would suggest that libp2p forwards its own default feature to libp2p-*.
> 
> I am not quite sure I follow. Could you expand?

If the meta crate has it's `default` enabled, it could foward that to all the subcrates:

```toml
[features]
default = [
    "yamux",
    # ...
    "libp2p-core/default",
    "libp2p-identify/default",
    # ...
]
```

But probably, the intention is that the meta crate works on a higher level than that, so feel free to dismiss that idea.

> > Additionally, we probably need to discuss which subcrates require what features from libp2p-core
> 
> Given that `secp256k1` is the only feature on `libp2p-core` there is not much to discuss, no?

Yes, I went too abstract in my thinking too quickly :-)

> > and how to make sure that CI still covers the common situations in terms of feature flags.
> 
> We have a CI step testing with all features enabled. Given that features should always be designed in an additive way, this should be good enough. Am I missing something?

I would add some CI tests to check that certain features can be turned off, i.e., that they are in fact designed in an additive way.

mxinden:
> > I am in favor of this step towards #2173.
> 
> Does this mean "I am in favor of this PR as a step towards #2173", or "I am in favor of discussing this approach in context of #2173"? Not sure how to read you here.

The former. In other words, I am in favor of both this pull request and #2173 and I think we don't have to block this pull request on #2173 and can thus already merge this pull request.

---

> > > This is probably a backwards incompatible change.
> > 
> > 
> > Yes. It would require a changelog entry.
> 
> Ack, will add after resolving the previous point :-)

:pray: small caveat, for each crate touched in this pull request, one entry needs to be added to the corresponding CHANGELOG.md file. Sorry for the inconvenience :/


---

> 
> > > I would suggest that libp2p forwards its own default feature to libp2p-*.
> > 
> > 
> > I am not quite sure I follow. Could you expand?
> 
> If the meta crate has it's `default` enabled, it could foward that to all the subcrates:
> 
> ```toml
> [features]
> default = [
>     "yamux",
>     # ...
>     "libp2p-core/default",
>     "libp2p-identify/default",
>     # ...
> ]
> ```
> 
> But probably, the intention is that the meta crate works on a higher level than that, so feel free to dismiss that idea.

Ah, thanks for explaining. Yes, that would make sense to me, though I would guess we would  want to forward the `full` feature with #2173 in mind. Either way, I think this should be a separate pull request.

---
 
> > > and how to make sure that CI still covers the common situations in terms of feature flags.
> > 
> > 
> > We have a CI step testing with all features enabled. Given that features should always be designed in an additive way, this should be good enough. Am I missing something?
> 
> I would add some CI tests to check that certain features can be turned off, i.e., that they are in fact designed in an additive way.

Indeed, that would be nice to have. Would you have a test requiring said feature that is then expected to fail?

rubdos:
> > > I am in favor of this step towards #2173.
> > 
> > 
> > Does this mean "I am in favor of this PR as a step towards #2173", or "I am in favor of discussing this approach in context of #2173"? Not sure how to read you here.
> 
> The former. In other words, I am in favor of both this pull request and #2173 and I think we don't have to block this pull request on #2173 and can thus already merge this pull request.

Ack, I'll finish this up then.

> > > > This is probably a backwards incompatible change.
> > > 
> > > 
> > > Yes. It would require a changelog entry.
> > 
> > 
> > Ack, will add after resolving the previous point :-)
> 
> pray small caveat, for each crate touched in this pull request, one entry needs to be added to the corresponding CHANGELOG.md file. Sorry for the inconvenience :/

Don't worry. Vim-fu will help.

> > > > I would suggest that libp2p forwards its own default feature to libp2p-*.
> > > 
> > > 
> > > I am not quite sure I follow. Could you expand?
> > 
> > 
> > If the meta crate has it's `default` enabled, it could foward that to all the subcrates:
> > ```toml
> > [features]
> > default = [
> >     "yamux",
> >     # ...
> >     "libp2p-core/default",
> >     "libp2p-identify/default",
> >     # ...
> > ]
> > ```
> > But probably, the intention is that the meta crate works on a higher level than that, so feel free to dismiss that idea.
> 
> Ah, thanks for explaining. Yes, that would make sense to me, though I would guess we would want to forward the `full` feature with #2173 in mind. Either way, I think this should be a separate pull request.

Makes sense.

> > > > and how to make sure that CI still covers the common situations in terms of feature flags.
> > > 
> > > 
> > > We have a CI step testing with all features enabled. Given that features should always be designed in an additive way, this should be good enough. Am I missing something?
> > 
> > 
> > I would add some CI tests to check that certain features can be turned off, i.e., that they are in fact designed in an additive way.
> 
> Indeed, that would be nice to have. Would you have a test requiring said feature that is then expected to fail?

I think that just checking that everything compiles additively should be enough. Tests that require specific features will fail to compile anyway [citation needed]

rubdos:
Hmm, on the other hand, failure testing would be nice to have; as an example: test that secp256p1 doesn't work when the feature is disabled.

Since you already test `--no-default-features` and `--all-features`, I'm not entirely sure what other feature sets to test (apart from maybe the default feature set?)

CHANGELOG.md
```diff
@@ -43,6 +43,9 @@
 
 ## Version 0.40.0 [unreleased]
 
+- Make default features of `libp2p-core` optional in all individual crates.
+  [PR 2181](https://github.com/libp2p/rust-libp2p/pull/2181)
+
```
mxinden, line 48:
```suggestion
```

No need for this entry, given that this pull request does not touch the root level `Cargo.toml`.

misc/multistream-select/CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # 0.10.4 [unreleased]
 
+- Make default features of `libp2p-core` optional.
+  [PR 2181](https://github.com/libp2p/rust-libp2p/pull/2181)
+
```
mxinden, line 5:
```suggestion
```

No need for this entry, given that `libp2p-core` is only a dev-dependency to `multistream-select`.

mxinden (approved):
Looks good to me. Thank @rubdos for tackling this.

I think this is a good step forward, allowing people to use `libp2p` without `secp256k1`. Will leave it open for  a bit longer for others to comment as well.

PR #2182, opened 2021-08-06
swarm/src/lib: Refactor trait usage

mxinden:
- Removes the `Swarm` type alias, renaming `ExpandedSwarm` to `Swarm`.
- Remove `TInEvent`, `TOutEvent` and `THandler` trait parameters on
`Swarm`, instead deriving them through `TBehaviour`. Move derive logic
to separate type aliases.
- Simplify trait bounds on `Swarm` main `impl` and `Stream` `impl`.

`ExpandedSwarm` was introduced in https://github.com/libp2p/rust-libp2p/pull/1046.

> Turns `Swarm` into a `type Swarm<...> = ExpandedSwarm<...>` where `ExpandedSwarm` has a lot more template parameters.
> 
> Unfortunately Rust still has a lot of limitations when it comes to bounds. I got blocked in Substrate because of some combination of requirements that wasn't possible to satisfy in a struct declaration. This PR makes it possible to use the swarm in ways that weren't possible by removing the mandatory `where TBehaviour: NetworkBehaviour` requirement from `Swarm`.

//CC @kpp and @tomaka to make sure this is not breaking Substrate.


tomaka:
There wouldn't be any problem with Substrate. This was at a time where we were trying to give the users the possibility to plug their own `NetworkBehaviour` into Substrate, which is an idea we've abandoned.

As far as I remember, the problem is that you can write this:

```rust
struct Foo<T>;
impl<T, U> Foo<T> where T: NetworkBehaviour, T::Output: SomeTrait<U> {}
```

But you can't enforce the trait directly in the struct definition:

```rust
struct Foo<T, U> where T: NetworkBehaviour, T::Output: SomeTrait<U> {
    // would need to use U here, contrary to the first example where U isn't part of the struct definition
}
```

Unfortunately I can't remember the reason why we need to enforce `NetworkBehaviour` on the struct definition itself.

thomaseizinger (commented):
Very much in favour of this!

PR #2184, opened 2021-08-07
Support client mode in kademlia

### LINKED ISSUE #2032: Support client mode for Kademlia implementation
From what I understand from https://github.com/libp2p/rust-libp2p/discussions/2025#discussioncomment-570443 the Kademlia implementation does not support client mode. Client mode is quite useful as it allows lower quality (e.g. intermittent availability, high latency, low bandwidth, low CPU/RAM/Storage, etc.) nodes to utilize the DHT for discovery without decreasing the quality of the network.

This feature has been very helpful in go-libp2p and in the various consumers of it such as go-ipfs.

---

To implement client mode all that is really required is that the node does not advertise the DHT protocol (e.g. `/ipfs/kad/1.0.0`). Additionally, if the node is able to switch between client and server mode at runtime then updates to the supported protocols should be communicated via `identify-push`/`identify-delta`.

Once we have the basic client-server mode switching then as a follow up it might be a good idea to, as go-libp2p does, have an "automatic" mode that switches between client and server mode based on educated guessing about what is appropriate for the user (e.g. peers behind relays probably shouldn't be servers, and peers who haven't been online for a while and are stable probably shouldn't either).

whereistejas:
Fix #2032 

Feature description:

> To do this, we use libp2p's [AutoNAT](https://github.com/libp2p/go-libp2p-autonat), which acts as a distributed session traversal utility for NAT (STUN) layer, informing peers of their observed addresses and whether or not they appear to be publicly dialable. Only when peers detect that they are publicly dialable do they switch from client mode (where they can query the DHT but not respond to queries) to server mode (where they can both query and respond to queries). Similarly, if a server discovers that it is no longer publicly dialable, it will switch back into client mode.

From https://docs.ipfs.io/concepts/dht/#undialable-peers


protocols/kad/src/behaviour.rs
```diff
@@ -147,6 +147,8 @@ pub struct KademliaConfig {
     connection_idle_timeout: Duration,
     kbucket_inserts: KademliaBucketInserts,
     caching: KademliaCaching,
+    // Set the peer mode.
+    client: bool,
```
mxinden, line 151: Let's use a separate `enum` here, e.g. `enum Mode { Client, Server }` for the sake of type safety. `Mode` can then implement `Default`, which would default to `Server` (for now).

protocols/kad/src/behaviour.rs
```diff
@@ -352,6 +356,19 @@ impl KademliaConfig {
...
         self
     }
+
+    /// Enable or disable the `client` mode of a node.
+    ///
+    /// Client mode is disabled by default.
+    ///
+    /// In `client` mode, the peer would not be added to any other peer's 
+    /// routing table. The client peer can send queries over the Kademlia
+    /// network, but will not receive any queries from the other peer's in
```
mxinden, line 366:
```suggestion
    /// network, but will not receive any queries from the other peers in
```

protocols/kad/src/behaviour.rs
```diff
@@ -352,6 +356,19 @@ impl KademliaConfig {
...
+    /// the network.
+    pub fn set_client_mode(&mut self, t: bool) -> &mut Self {
```
mxinden, line 368:
```suggestion
    pub fn set_mode(&mut self, t: bool) -> &mut Self {
```

I would talk about the `Mode` in general, instead of the client mode in specific. The method would take a `Mode`, see comment above.

protocols/kad/src/dht.proto
```diff
@@ -26,6 +26,9 @@ message Record {
     // The remaining TTL of the record, in seconds.
     // Currently specific to rust-libp2p.
     uint32 ttl = 777;
+
+	// Client mode or server mode.
+	bool client = false;
```
mxinden, line 31: The fact that a node is in client or server mode is never transfered over the wire. Instead, one signals to be in server mode by offering the `/libp2p/kad/0.1.0` protocol on new substreams and via the identify protocol.
```suggestion
```

mxinden (commented):
There is still a lot missing here. Let me know if you need more details. Also see my comment  https://github.com/libp2p/rust-libp2p/issues/2032#issuecomment-884105888.

whereistejas:
Hi @mxinden,
I'm trying to write an integration test to test this enhancement. It's taking longer than anticipated.

whereistejas:
Hi @mxinden,
I have added an example file (sort of like an integration test) for our implementation. The idea is that you run multiple instances of the example in different terminals, and we can run one instance with `--client` flag. Then, we use the `list peer` command in the other (non-client) instances to check if the `client` peer is **not** present in their routing tables.

examples/kademlia-example.rs
```diff
@@ -0,0 +1,157 @@
+#![allow(dead_code, unused_variables)]
```
mxinden, line 1: Instead of an example, could you add a test to `kad/src/behaviour/tests.rs`?

protocols/kad/src/behaviour.rs
```diff
@@ -147,6 +147,8 @@ pub struct KademliaConfig {
     connection_idle_timeout: Duration,
     kbucket_inserts: KademliaBucketInserts,
     caching: KademliaCaching,
+    // Set the peer mode.
+    client: Mode,
```
mxinden, line 151:
```suggestion
    mode: Mode,
```

protocols/kad/src/behaviour.rs
```diff
@@ -180,6 +182,8 @@ impl Default for KademliaConfig {
             connection_idle_timeout: Duration::from_secs(10),
             kbucket_inserts: KademliaBucketInserts::OnConnected,
             caching: KademliaCaching::Enabled { max_peers: 1 },
+            // By default, a peer will not be in client mode.
+            client: Mode::default(),
```
mxinden, line 186:
```suggestion
            client: Mode::default(),
```

Just in case we ever update the default, we won't have an outdated comment.

protocols/kad/src/behaviour.rs
```diff
@@ -352,6 +356,19 @@ impl KademliaConfig {
...
     }
+
+    /// Enable or disable the `client` mode of a node.
+    ///
+    /// Client mode is disabled by default.
+    ///
+    /// In `client` mode, the peer would not be added to any other peers
+    /// routing table. The client peer can send queries over the Kademlia
+    /// network, but will not receive any queries from the other peers in
+    /// the network.
```
mxinden, line 367:
```suggestion
    /// Sets the [`Mode`] the node is operating in.
    ///
    /// The default is [`Mode::Server`].
    ///
    /// In [`Mode::Client`] the node does not advertise support for the Kademlia
    /// protocol, nor does it accept incoming Kademlia streams. Peers will thus not
    /// include the local node in their routing table. The node can still make
    /// outbound requests.
    ///
    /// Use [`Mode::Client`] when not publicly reachable.
```

protocols/kad/src/handler.rs
```diff
@@ -123,6 +128,9 @@ pub struct KademliaHandlerConfig {
 
     /// Time after which we close an idle connection.
     pub idle_timeout: Duration,
+
+    // If Mode::Client, node will act in `client` mode in the Kademlia network.
+    pub client: Mode,
```
mxinden, line 133:
```suggestion

    /// [`Mode`] the handler is operating in.
    pub mode: Mode,
```

mxinden, line 133: The `Mode` logic you are introducing in this pull request replaces the `KademliaHandlerConfig::allow_listening` logic. The latter can thus be replaced in favor of the former :tada:

protocols/kad/src/handler.rs
```diff
@@ -520,12 +537,16 @@ where
...
+        // TODO:
+        // Just because another peer is sending us Kademlia requests, doesn't necessarily
+        // mean that it will answer the Kademlia requests that we send to it.
+        // Thus, Commenting this code out for now.
+        // if let ProtocolStatus::Unconfirmed = self.protocol_status {
+        //     // Upon the first successfully negotiated substream, we know that the
+        //     // remote is configured with the same protocol name and we want
+        //     // the behaviour to add this peer to the routing table, if possible.
+        //     self.protocol_status = ProtocolStatus::Confirmed;
+        // }
```
mxinden, line 549: Correct. I am in favor of documenting this, though obviously both the `TODO` and the commented out code needs to be removed.

protocols/kad/src/protocol.rs
```diff
@@ -46,6 +46,18 @@ pub const DEFAULT_PROTO_NAME: &[u8] = b"/ipfs/kad/1.0.0";
 /// The default maximum size for a varint length-delimited packet.
 pub const DEFAULT_MAX_PACKET_SIZE: usize = 16 * 1024;
 
+#[derive(Debug, Clone)]
```
mxinden, line 49:
```suggestion
/// See [`crate::KademliaConfig::set_mode`].
#[derive(Debug, Clone)]
```

mxinden (commented):
Great steps forward @whereistejas!

In case you are interested in working on specifications as well: It would be great to have the _Client Mode_ feature described in the [Kademlia specification](https://github.com/libp2p/specs/blob/master/kad-dht/README.md).

examples/kademlia-example.rs
whereistejas: Writing this test in the `kademlia` module will be kind of difficult, as I'm using a lot of other things as well, for building the swarm or transport. Is it okay if I put this is in a test directory?

whereistejas:
> In case you are interested in working on specifications as well: It would be great to have the _Client Mode_ feature described in the [Kademlia specification](https://github.com/libp2p/specs/blob/master/kad-dht/README.md).

Sure! I can do this.

The code changes in the `kademlia` I have currently are not producing the desired results. I will take a look at it over the weekend. Do I need to make any other changes? I remember us discussing about another edge case in the `multistream` module (I maybe completely wrong here).

examples/kademlia-example.rs
mxinden: > Writing this test in the kademlia module will be kind of difficult, as I'm using a lot of other things as well, for building the swarm or transport. Is it okay if I put this is in a test directory?

No need for most of it. You can take the `manual_bucket_insert` and `bootstrap` test as an example.

> I can further enhance kademlia-example to create four peers and cross-check if the peer in client mode is actually, absent from the other's routing tables.

I don't think there is any need for an example. The "cross-check" sounds great, but should really be done in a test instead of an example.

mxinden:
> The current code changes in the `kademlia` submodule are not producing the desired results. I will take a look at it over the weekend. Do I need to make any other changes? I remember us discussing about another edge case in the `multistream` module (I maybe completely wrong here).

Nothing comes to my mind. What exactly is not working? I don't think this is related to `multistream-select`.

whereistejas:
libp2p kademlia client mode testing
Hi @mxinden, this is what I meant by not working. As you can see in the screenshot, the "other" peers can still see the "client" node.

whereistejas:
Okay, so something really weird is happening. I put some `println` statements in `listen_protocol` and `inject_fully_negotiated_inbound` to check if these methods are getting called when two peers connect to one another and I found that they are not getting called. I am using the code I wrote in `examples/kademlia-example.rs` to test this. Am I doing something wrong?

protocols/kad/src/behaviour/test.rs
```diff
@@ -1287,3 +1287,87 @@ fn network_behaviour_inject_address_change() {
...
+                                    .any(|p| p.node.key.preimage() == &swarm_ids[1]),
+                                "The client peer does not have the server node"
+                            );
+
+                            return Poll::Ready(());
+                        }
+
+                        if swarm.local_peer_id() == &swarm_ids[1] && server_listening == true {
+                            let bucket =
+                                swarm.behaviour_mut().kbucket(swarm_ids[0].clone()).unwrap();
```
whereistejas, line 1353: If the `server` peer doesn't have the `client` peer in its routing tables won't `bucket` be `None`?

protocols/kad/src/behaviour/test.rs
```diff
@@ -1287,3 +1287,87 @@ fn network_behaviour_inject_address_change() {
...
+        for swarm in swarms.iter_mut() {
+            loop {
+                match swarm.poll_next_unpin(ctx) {
+                    Poll::Ready(Some(SwarmEvent::NewListenAddr {
+                        listener_id: _,
+                        address,
+                    })) => {
+                        if address == swarm_addrs[0] {
+                            client_listening = true;
+                            return Poll::Ready(());
```
mxinden, line 1328: Why `return` in this case? The actual logic of the test is still to come, no?

protocols/kad/src/behaviour/test.rs
```diff
@@ -1287,3 +1287,87 @@ fn network_behaviour_inject_address_change() {
...
+                        }
+                        if address == swarm_addrs[1] {
+                            server_listening = true;
+                            return Poll::Ready(());
+                        }
+                    }
+                    Poll::Ready(_) => {
```
mxinden, line 1335: How about waiting for a `KademliaEvent::RoutingUpdated` from the client swarm with the server peer id?

protocols/kad/src/behaviour/test.rs
```diff
@@ -1287,3 +1287,87 @@ fn network_behaviour_inject_address_change() {
...
+    swarm_addrs.push(client_swarm.0);
+    swarm_addrs.push(server_swarm.0);
+
+    let swarm_ids: Vec<_> = swarms
+        .iter()
+        .map(|swarm| swarm.local_peer_id())
+        .cloned()
+        .collect();
+
+    block_on(poll_fn(move |ctx| {
```
mxinden, line 1314: Wouldn't you need to have the client dial the server before starting the test logic?

protocols/kad/src/behaviour/test.rs
mxinden: The bucket should not exist in the first place, yes.

protocols/kad/src/behaviour/test.rs
whereistejas: Yes, I agree. I'm still not very familiar with which events will be generated. During my initial testing, the test would get stuck in a never-ending loop, without the `return` statement.

protocols/kad/src/behaviour/test.rs
whereistejas: This is strange, indeed. I wonder how the test is still passing.

whereistejas:
Hi @mxinden, the test is working, now. I have incorporated all the changes you suggested. But, I wonder if we are missing any edge cases, do we need any more tests?

protocols/kad/src/behaviour/test.rs
```diff
@@ -1287,3 +1287,67 @@ fn network_behaviour_inject_address_change() {
...
+                match swarm.poll_next_unpin(ctx) {
+                    Poll::Ready(Some(SwarmEvent::Behaviour(KademliaEvent::RoutingUpdated {
+                        peer,
+                        ..
+                    }))) => {
+                        // Check if the server peer is present in the client peer's routing table.
+                        if swarm.local_peer_id().clone() == peers[0] {
+                            is_server_present = peer == peers[1];
+                        }
+                        return Poll::Ready(());
```
whereistejas, line 1328: There is one part of this code, I don't understand. I noticed that the test fails if we don't add `return Poll::Ready(())` in every `match` arm. I'm still a newbie at `async`, could you tell me why we need to return this and what its effect is?

whereistejas:
Screenshot from 2021-08-30 23-29-25

:cry:

whereistejas:
So, I guess the CI failure for commit [`54331e5`](https://github.com/libp2p/rust-libp2p/pull/2184/commits/54331e549aa4c9ad2641510409f5f85496b5554e) was just a random event.

protocols/kad/src/behaviour/test.rs
mxinden: I suggest reading the async book: https://rust-lang.github.io/async-book/

With the early return we might not check whether the server stores the client in its routing table.

protocols/kad/src/behaviour/test.rs
```diff
@@ -1317,3 +1317,67 @@ fn network_behaviour_inject_address_change() {
...
+
+    // Collect addresses and peer IDs.
+    let addrs: Vec<_> = swarms.iter().map(|(addr, _)| addr.clone()).collect();
+    let peers: Vec<_> = swarms
+        .iter()
+        .map(|(_, swarm)| swarm.local_peer_id().clone())
+        .collect();
+
+    swarms[0].1.dial_addr(addrs[1].clone()).unwrap();
+    swarms[1].1.dial_addr(addrs[0].clone()).unwrap();
```
mxinden, line 1340: Why does the server dial the client?

protocols/kad/src/behaviour/test.rs
```diff
@@ -1317,3 +1317,67 @@ fn network_behaviour_inject_address_change() {
...
+                        // Check if the server peer is present in the client peer's routing table.
+                        if swarm.local_peer_id().clone() == peers[0] {
+                            is_server_present = peer == peers[1];
+                        }
+                        return Poll::Ready(());
+                    }
+                    Poll::Ready(_) => {
+                        // Check if the client peer is NOT present in the server peer's routing table.
+                        if swarm.local_peer_id().clone() == peers[1] {
+                            is_client_absent = swarm.behaviour_mut().kbucket(peers[0]).is_none();
```
mxinden, line 1363: It is kind of hard to reliably check for absence of something, especially in an async setting.

How about:

1. The server stores a value `Kademlia::put_value`, waiting for a `KademliaEvent::OutboundQueryCompleted`.
2. The client connects to the server.
3. The client does a `Kademlia::get_record`, waiting for a `KademliaEvent::OutboundQueryCompleted`.
4. Check that the client has the server in its routing table and check that the server does not have the client in its routing table.

protocols/kad/src/behaviour/test.rs
whereistejas: It's there to guarantee that server tried to reach out to the client before, we check if the client is absent from the server's routing table.

protocols/kad/src/behaviour/test.rs
whereistejas: Sure. I will implement this. I will try to finish before the same time, tomorrow. Please keep an eye out for any notifications from this PR. I want to get this PR closed, as early as possible (ofcourse, without sacrificing code quality).

whereistejas:
It looks like our client peer is unable to connect to the server peers.

```bash
    1     Blocking waiting for file lock on build directory
    2    Compiling libp2p-kad v0.32.0 (/home/tsanap/repos/open-source-contrib/libp2p-related/rust-libp2p/protocols/kad)
    3     Finished test [unoptimized + debuginfo] target(s) in 6.79s
    4      Running unittests (/home/tsanap/repos/open-source-contrib/libp2p-related/rust-libp2p/target/debug/deps/libp2p_kad-71e8039cf6bd90c8)
    5
    6 running 1 test
    7 server peer: PeerId("12D3KooWDinU4DnAaGP1yKcXPqPHrM7YFzmHgixAyxVm792aWW3C")
    8 server peer: PeerId("12D3KooWSLnjd9jhs5GYPZEQ9eQrsa2PuVYcPhdh9ikwzkbD6HED")
    9 client peer: PeerId("12D3KooWShTygkUnBaJAAixUmoxe6Y912KfaM9amiVSjjXMDGB2K")
   10 Source: PeerId("12D3KooWDinU4DnAaGP1yKcXPqPHrM7YFzmHgixAyxVm792aWW3C"), Kademlia Event: NewListenAddr { listener_id: ListenerId(1), address: "/memory/4081114263267638787" }
   11 Source: PeerId("12D3KooWDinU4DnAaGP1yKcXPqPHrM7YFzmHgixAyxVm792aWW3C"), Kademlia Event: IncomingConnection { local_addr: "/memory/4081114263267638787", send_back_addr: "/memory/10260145968745791016" }
   12 Source: PeerId("12D3KooWDinU4DnAaGP1yKcXPqPHrM7YFzmHgixAyxVm792aWW3C"), Kademlia Event: Behaviour(RoutingUpdated { peer: PeerId("12D3KooWSLnjd9jhs5GYPZEQ9eQrsa2PuVYcPhdh9ikwzkbD6HED"), is_new_peer: true, add
   13 resses: ["/memory/10688807535701962771"], bucket_range: (Distance(57896044618658097711785492504343953926634992332820282019728792003956564819968), Distance(1157920892373161954235709850086879078532699846656405
   14 64039457584007913129639935)), old_peer: None })
   15 Source: PeerId("12D3KooWDinU4DnAaGP1yKcXPqPHrM7YFzmHgixAyxVm792aWW3C"), Kademlia Event: Dialing(PeerId("12D3KooWSLnjd9jhs5GYPZEQ9eQrsa2PuVYcPhdh9ikwzkbD6HED"))
   16 Source: PeerId("12D3KooWSLnjd9jhs5GYPZEQ9eQrsa2PuVYcPhdh9ikwzkbD6HED"), Kademlia Event: NewListenAddr { listener_id: ListenerId(1), address: "/memory/10688807535701962771" }
   17 Source: PeerId("12D3KooWSLnjd9jhs5GYPZEQ9eQrsa2PuVYcPhdh9ikwzkbD6HED"), Kademlia Event: IncomingConnection { local_addr: "/memory/10688807535701962771", send_back_addr: "/memory/15478558967721088777" }
   18 Source: PeerId("12D3KooWSLnjd9jhs5GYPZEQ9eQrsa2PuVYcPhdh9ikwzkbD6HED"), Kademlia Event: IncomingConnection { local_addr: "/memory/10688807535701962771", send_back_addr: "/memory/8542520709622284514" }
   19 Source: PeerId("12D3KooWSLnjd9jhs5GYPZEQ9eQrsa2PuVYcPhdh9ikwzkbD6HED"), Kademlia Event: Behaviour(RoutingUpdated { peer: PeerId("12D3KooWDinU4DnAaGP1yKcXPqPHrM7YFzmHgixAyxVm792aWW3C"), is_new_peer: true, add
   20 resses: ["/memory/4081114263267638787"], bucket_range: (Distance(57896044618658097711785492504343953926634992332820282019728792003956564819968), Distance(11579208923731619542357098500868790785326998466564056
   21 4039457584007913129639935)), old_peer: None })
   22 Source: PeerId("12D3KooWShTygkUnBaJAAixUmoxe6Y912KfaM9amiVSjjXMDGB2K"), Kademlia Event: NewListenAddr { listener_id: ListenerId(1), address: "/memory/5653567981690961167" }
   23 thread 'behaviour::test::client_mode' panicked at 'GET operation failed: NotFound { key: Key(b"123"), closest_peers: [] }', protocols/kad/src/behaviour/test.rs:1397:43
   24 note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
   25 test behaviour::test::client_mode ... FAILED
   26
   27 failures:
   28
   29 failures:
   30     behaviour::test::client_mode
   31
   32 test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 49 filtered out; finished in 0.01s
```

whereistejas:
Hi @mxinden, the CI tests are passing on my local machine. Could you take a look at the code, please?

whereistejas:
Hi @mxinden, this PR ready for review. Just a gentle reminder. :)

.gitignore
```diff
@@ -1,2 +1,3 @@
 target
 Cargo.lock
+.vscode/launch.json
```
mxinden, line 3: Please remove this line.

protocols/kad/src/behaviour/test.rs
```diff
@@ -1317,3 +1317,96 @@ fn network_behaviour_inject_address_change() {
...
+fn client_mode() {
+    // Create server peers.
+    let mut server_nodes = build_fully_connected_nodes_with_config(2, KademliaConfig::default());
+
+    // Create a client peer.
+    let mut cfg = KademliaConfig::default();
+    cfg.set_mode(Mode::Client);
+    let mut client = build_node_with_config(cfg);
+
+    // Fitler out peer Ids.
```
mxinden, line 1331:
```suggestion
    // Filter out peer Ids.
```

protocols/kad/src/behaviour/test.rs
```diff
@@ -1317,3 +1317,96 @@ fn network_behaviour_inject_address_change() {
...
+                                // Check if the server peer is not connected to the client peer.
+                                assert!(
+                                    swarm
+                                        .behaviour_mut()
+                                        .connected_peers
+                                        .iter()
+                                        .all(|p| *p != peers[2]),
+                                    "The server peer is connected to the client peer."
+                                );
+                            }
```
mxinden, line 1381: How about calling `swarm[2].behaviour_mut().get_record(..)` once the first server is done with the `PutRecord` query. That would prevent any races between the PUT and the GET.

protocols/kad/src/handler.rs
```diff
@@ -62,10 +62,15 @@ impl<T: Clone + fmt::Debug + Send + 'static> IntoProtocolsHandler for KademliaHa
...
 
     fn inbound_protocol(&self) -> <Self::Handler as ProtocolsHandler>::InboundProtocol {
-        if self.config.allow_listening {
-            upgrade::EitherUpgrade::A(self.config.protocol_config.clone())
-        } else {
-            upgrade::EitherUpgrade::B(upgrade::DeniedUpgrade)
+        match self.config.mode {
+            Mode::Client => upgrade::EitherUpgrade::B(upgrade::DeniedUpgrade),
+            Mode::Server => {
+                if self.config.allow_listening {
```
mxinden, line 68: Is the `allow_listening` logic still needed?

See https://github.com/libp2p/rust-libp2p/pull/2184/#discussion_r691909334

mxinden (commented):
Good progress here.

Just to make sure, does your test fail without the patch?

whereistejas:
The test is not failing without the patch, which is worrying me. I will try to take debug and find out why.

protocols/kad/src/behaviour/test.rs
whereistejas: I think, `swarm[2]` won't be available to us inside the `for` loop.

whereistejas: Hi, I have a doubt with the present test we have:

> It is kind of hard to reliably check for absence of something, especially in an async setting.
> 1. The server stores a value `Kademlia::put_value`, waiting for a `KademliaEvent::OutboundQueryCompleted`.
> 2. The client connects to the server.
> 3. The client does a `Kademlia::get_record`, waiting for a `KademliaEvent::OutboundQueryCompleted`.
> 4. Check that the client has the server in its routing table and check that the server does not have the client in its routing table.

Are we sure that using the GetRecord operation will update the routing tables of the server peer? I'm trying to find the code, which does the "update"-ing. But, I'm unable to find it. Could you please point me towards the appropriate code?

Update: (Please correct me if I'm wrong) I don't think Kademlia is bi-directional. For example, if peer A adds peer B to its routing table, then peer B will not add peer A to its routing table.

protocols/kad/src/behaviour/test.rs
```diff
@@ -1318,6 +1318,119 @@ fn network_behaviour_inject_address_change() {
...
+    addrs.push(server1.0);
+    let mut server1 = server1.1;
+
+    let server2 = build_node();
+    addrs.push(server2.0);
+    let mut server2 = server2.1;
+
+    // Create a client peer.
+    let mut cfg = KademliaConfig::default();
+    cfg.set_mode(Mode::Server);
```
mxinden, line 1337: I am a bit confused. Why is the client initialized with `Mode::Server`?

whereistejas: Oops, that's just a typing mistake on my side. Anyways, this test should fail when the client is in `Server` mode, which it is not. I have some doubts regarding the way `Kademlia` works, which I have expressed [here](https://github.com/libp2p/rust-libp2p/pull/2184#discussion_r720818215). Could you please help me out with those?

protocols/kad/src/behaviour/test.rs
mxinden: > Update: (Please correct me if I'm wrong) I don't think Kademlia is bi-directional. For example, if peer A adds peer B to its routing table, then peer B will not add peer A to its routing table.

Before this pull request both peers would likely add each other to their respective routing table. After this pull request, whether the two peers add each other to their routing table depends on the `Mode` the other peer is running in.

mxinden:
Closing here in favor of #2521.

Issue #2158, opened 2021-07-23
examples: Showcase how to use rust-libp2p in larger application


mxinden:
While we have many examples which showcase rust-libp2p in isolation, it would be great to have an example on how to integrate rust-libp2p into a larger application.

I was thinking of something along the lines of:

- A single task polling (a) the `Swarm` and (b) a channel of incoming commands.
  
  The commands coming through (b) would contain a `oneshot::Sender`, allowing the task to send back a response when ready.

- A _remote control_ providing `async` methods, each sending a command down to the task, awaiting the `oneshot::Receiver` side.

The example should derive its own `NetworkBehaviour` based on e.g. `libp2p-request-response` and `libp2p-identify`.

See the following resources for additional background:

- https://github.com/libp2p/rust-libp2p/discussions/2024
- https://github.com/iotaledger/stronghold.rs/pull/210#issuecomment-878976857

Happy to expand / discuss details in case someone wants to tackle this.

thomaseizinger:
> * The commands coming through (b) would contain a `oneshot::Sender`, allowing the task to send back a response when ready.

I've found [`bmrng`](https://docs.rs/bmrng/0.5.1/bmrng/index.html) to be helpful here.

mxinden:
> > * The commands coming through (b) would contain a `oneshot::Sender`, allowing the task to send back a response when ready.
> 
> I've found [`bmrng`](https://docs.rs/bmrng/0.5.1/bmrng/index.html) to be helpful here.

What would be the benefit of `bmrng` compared to a combination of `mpsc::channel` and `mpsc::oneshot` apart from being a bit less noisy?

thomaseizinger:
It is primarily less noisy but also comes with a built in timeout which is useful if you have other parts of the application that talk to the _remote control_ and don't want to wait forever on results coming back from the network layer.

But really, the main value is convenience through f.e. the [`send`](https://docs.rs/bmrng/0.5.1/bmrng/struct.RequestSender.html#method.send) function.

elenaf9:
I'd be happy to add such an example, similar to how it is done in the interface the iotaledger/stronghold.rs#210 PR you linked above.
But I probably won't get to it until end of next week, so in case anybody else would like to tackle this before, please feel free.

[...]

PR #2171, opened 2021-07-31
Add async chat example

thomaseizinger:
This showcases the use of an event-loop and channels to react to
events from and give commands to the swarm.

PR #2186, opened 2021-08-08
examples/: Add file sharing example

mxinden:
Basic file sharing application with peers either providing or locating
and getting files by name.

While obviously showcasing how to build a basic file sharing
application, the actual goal of this example is **to show how to
integrate rust-libp2p into a larger application**.

Architectural properties

- Clean clonable async/await interface ([`Client`]) to interact with the
network layer.

- Single task driving the network layer, no locks required.

Closes https://github.com/libp2p/rust-libp2p/issues/2158

Compared to https://github.com/libp2p/rust-libp2p/pull/2171/ this example would showcase the full idea of a remote control (`Client`) offering simple `async` methods with the help of `futures::channel::oneshot`. I would be curious what you think @thomaseizinger.

@elenaf9 What do you think of this example compared to the larger structure introduced in https://github.com/iotaledger/stronghold.rs/pull/210?

Also, I would be especially curious what people, who are not deeply familiar with `libp2p-swarm` and `libp2p-core` already, think about this example. Any suggestions from your side? (//CC @whereistejas @r-zig)


mxinden:
Also mentioning @umgefahren here, as you were involved on https://github.com/libp2p/rust-libp2p/issues/2158. Feedback very much appreciated.

whereistejas:
I think it is a very good example. I'm myself working on a [program](https://github.com/whereistejas/rust-libp2p/blob/kademlia-example/examples/kademlia-example.rs) where I have four peers in a kademlia network and one of those peers will be operating in client mode. I was having trouble getting the swarm to work, without using `async_std::block_on`. This program helped me out there. I think, we can have more comments in this example code and others, explaining the intricacies which are otherwise not easily apparent. For example, to use the `upgrade` method on a transport you need to import the `Transport` structure. I'm sure people with more experience can come up with better points, in this context.

Also, another thing that can be done, is we can compile a list of good sample projects (with links to specific files) where `libp2p` is used. I found the following sample program, which I really helped me out: https://github.com/zupzup/rust-peer-to-peer-example

examples/file-sharing.rs
```diff
@@ -0,0 +1,639 @@
...
+                    Default::default(),
+                ),
+            },
+            peer_id,
+        )
+        .build();
+
+        // In case a listen address was provided use it, otherwise listen on any
+        // address.
+        match listen_address {
```
elenaf9, line 261: I'd do this via an new method `Client::start_listening` & `Command::Start_Listening`, to be consistent.

examples/file-sharing.rs
```diff
@@ -0,0 +1,639 @@
...
+
+    /// Creates the network components, namely:
+    ///
+    /// - The network client to interact with the network layer from anywhere
+    ///   within your application.
+    ///
+    /// - The network event stream, e.g. for incoming requests.
+    ///
+    /// - The network task driving the network itself.
+    pub fn new(
```
elenaf9, line 225: This function does quit a lot. I'd split it up into to functions: one for creating the transport, network-behaviour,swarm, and then one for creating the Client and channels related to it. Maybe also move the spawning of the event_loop into that second function. This would highlight this step, and the logic related to it a bit more

elenaf9 (commented):
I think this showcases the usage of a remote-control & channels for swarm interaction quite well.
But I'd suggest to split the different parts a bit more into their own function each, for a better understanding and readability.

examples/file-sharing.rs
```diff
@@ -0,0 +1,639 @@
...
+    }
+
+    impl From<KademliaEvent> for ComposedEvent {
+        fn from(event: KademliaEvent) -> Self {
+            ComposedEvent::Kademlia(event)
+        }
+    }
+
+    #[derive(Debug)]
+    enum Command {
```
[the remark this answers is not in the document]
thomaseizinger: Agree with changing the signature to be fallible. I would suggest to refrain from returning Swarm errors directly though. If you ever want to implement things like retries, having that in the eventloop directly will result in a less chatty interface between Client and eventloop. A more high-level interface will likely have its own error type hence my suggestion to not use the error from the events directly.

You may want to go even one step further and remove Command::Dial and make it implicit as part of executing other commands. That would highlight that the eventloop is not just a "dumb" loop that forwards everything it gets to the client. (If we are on the same page here but in my experience, you want to push as much "networking" logic as possible down into the eventloop).

examples/file-sharing.rs
```diff
@@ -0,0 +1,639 @@
...
+    let opt = Opt::from_args();
+
+    // Creates the network components, namely:
+    //
+    // - The network client to interact with the network layer from anywhere
+    //   within your application.
+    //
+    // - The network event stream, e.g. for incoming requests.
+    //
+    // - The network task driving the network itself.
```
thomaseizinger, line 101: This is basically repeating the rustdoc, not sure if that is necessary?

examples/file-sharing.rs
```diff
@@ -0,0 +1,639 @@
...
+use async_std::task::spawn;
+use futures::prelude::*;
+use libp2p::core::{Multiaddr, PeerId};
+use libp2p::multiaddr::Protocol;
+use std::error::Error;
+use std::path::PathBuf;
+use structopt::StructOpt;
+
+#[async_std::main]
+async fn main() -> Result<(), Box<dyn Error>> {
```
thomaseizinger, line 89: Why not use `anyhow`?

examples/file-sharing.rs
```diff
@@ -0,0 +1,639 @@
...
+        Command::Provide { path, name } => {
+            // Advertise oneself as a provider of the file on the DHT.
+            network_client.start_providing(name.clone()).await;
+
+            loop {
+                match network_events.next().await {
+                    // Reply with the content of the file on incoming requests.
+                    Some(network::Event::InboundRequest { request, channel }) => {
+                        if request == name {
+                            let file_content = std::fs::read_to_string(&path)?;
```
thomaseizinger, line 128: Use async fs module instead?

examples/file-sharing.rs
```diff
@@ -0,0 +1,639 @@
...
+            let requests = providers.into_iter().map(|p| {
+                let mut network_client = network_client.clone();
+                let name = name.clone();
+                async move { network_client.request_file(p, name).await.ok_or(()) }.boxed()
+            });
+
+            // Await the requests, ignore the remaining once a single one succeeds.
+            let file = futures::future::select_ok(requests)
+                .await
+                .map_err(|_| Into::<Box<dyn Error>>::into("None of the providers returned file."))?
```
thomaseizinger, line 154: With `anyhow`, this would be a simple `.context`.

examples/file-sharing.rs
```diff
@@ -0,0 +1,639 @@
...
+
+            println!("Content of file {}: {}", name, file);
+        }
+    }
+
+    Ok(())
+}
+
+#[derive(Debug, StructOpt)]
+#[structopt(name = "libp2p relay")]
```
thomaseizinger, line 165: Copy-paste left over? :)

examples/file-sharing.rs
```diff
@@ -0,0 +1,639 @@
...
+    /// - The network client to interact with the network layer from anywhere
+    ///   within your application.
+    ///
+    /// - The network event stream, e.g. for incoming requests.
+    ///
+    /// - The network task driving the network itself.
+    pub fn new(
+        listen_address: Option<Multiaddr>,
+        secret_key_seed: Option<u8>,
+    ) -> Result<(Client, BoxStream<'static, Event>, BoxFuture<'static, ()>), Box<dyn Error>> {
```
thomaseizinger, line 228: This could use `impl Trait`.

examples/file-sharing.rs
```diff
@@ -0,0 +1,639 @@
...
+            let (sender, receiver) = oneshot::channel();
+            self.sender
+                .send(Command::Dial {
+                    peer_id,
+                    peer_addr,
+                    sender,
+                })
+                .await
+                .unwrap();
+            receiver.await.unwrap();
```
thomaseizinger, line 295: Document why we can unwrap these?

examples/file-sharing.rs
```diff
@@ -0,0 +1,639 @@
...
+    async fn event_loop(
+        mut swarm: Swarm<ComposedBehaviour>,
+        mut command_receiver: mpsc::Receiver<Command>,
+        mut event_sender: mpsc::Sender<Event>,
+    ) {
+        let mut outstanding_dial: HashMap<_, oneshot::Sender<()>> = HashMap::new();
+        let mut outstanding_start_providing = HashMap::new();
+        let mut outstanding_get_providers: HashMap<_, oneshot::Sender<HashSet<PeerId>>> =
+            HashMap::new();
+        let mut outstanding_request_file: HashMap<_, oneshot::Sender<String>> = HashMap::new();
```
thomaseizinger, line 350: The more common use of `outstanding` is to express that something is truly excellent. It can mean `pending` as well but I would probably just resort to `pending` to be unambiguous :)

thomaseizinger (approved):
Overall good, I does showcase the _client_ aspect a bit better. On the other hand, I think it lacks some emphasis that the `eventloop` should not be dumb but can actually perform significant logic involving its own state and incoming events.

Also added some nits throughout where I think we could write more idiomatic Rust code. I don't think examples should sacrifice on that :)

elenaf9:
> On the other hand, I think it lacks some emphasis that the eventloop should not be dumb but can actually perform significant logic involving its own state and incoming events.

I have a different view on this. In my opinion, the sole responsibility of the event loop should be to just "dumbly" call the `Swarm` methods according to the Commands it receives, return the result for each prior command once they are ready, and forward incoming events. 
I would consider the `Client` to be somewhat of a Proxy/ Stub of the actual Swarm (with the enhancement that result dont need to be polled, and are returned directly). So any relevant operation that can be performed on the actual swarm, should also be possible via the `Client`, and the event loop should only forward things.

thomaseizinger:
> > On the other hand, I think it lacks some emphasis that the eventloop should not be dumb but can actually perform significant logic involving its own state and incoming events.
> 
> @thomaseizinger I have a different view on this. In my opinion, the sole responsibility of the event loop should be to just "dumbly" call the `Swarm` methods according to the Commands it receives, return the result for each prior command once they are ready, and forward incoming events.
> I would consider the `Client` to be somewhat of a Proxy/ Stub of the actual Swarm (with the enhancement that result dont need to be polled, and are returned directly). So any relevant operation that can be performed on the actual swarm, should also be possible via the `Client`, and the event loop should only forward things.

That is fair! The reason I'd advocate against it is because it results in more code. Every interaction between client and eventloop requires a channel.

If the client acts as a proxy, it doesn't have any more knowledge about what is happening than the swarm. As a result, it should be possible to make every decision that you can make in the client also in the swarm but with less code needing to be written.

To illustrate what I mean, have a look at this snippet of code:

https://github.com/thomaseizinger/rust-libp2p/blob/cb0b5a7177ba2432b1f9b9c3fe0d78c21e2f3082/examples/chat-async.rs#L111-L122

If we would forward every mDNS event, the client would need to react to these generated events, only to tell the event loop to do something it could have achieved autonomously.

Given the controversy of this topic though, I am gonna retract my statement the we need to advocate for one or the other in the example :)

elenaf9:
> The reason I'd advocate against it is because it results in more code. Every interaction between client and eventloop requires a channel.
If the client acts as a proxy, it doesn't have any more knowledge about what is happening than the swarm. As a result, it should be possible to make every decision that you can make in the client also in the swarm but with less code needing to be written.
To illustrate what I mean, have a look at this snippet of code:
https://github.com/thomaseizinger/rust-libp2p/blob/cb0b5a7177ba2432b1f9b9c3fe0d78c21e2f3082/examples/chat-async.rs#L111-L122
If we would forward every mDNS event, the client would need to react to these generated events, only to tell the event loop to do something it could have achieved autonomously.

Good point, I agree that in some cases it makes sense to have more logic in the event loop, to reduce code and overhead.

Another thought:
If this is a lot of logic,  I would consider to implement it even one level lower as part of the NetworkBehaviour impl. Either with a custom poll method `#[behaviour(poll_method = "poll")]`, or by manually implementing the whole NetworkBehaviour trait to combine multiple protocols (like the proc_macro does it), and then add some logic on top.

But I agree that this very much depends on the concrete application and use-case.

examples/file-sharing.rs
```diff
@@ -0,0 +1,639 @@
...
+
+        loop {
+            futures::select! {
+                event = swarm.next() => match event.expect("Swarm stream to be infinite.") {
```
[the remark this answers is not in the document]
mxinden: I don't find `select_next_some` intuitive and usually find myself going back to the method's documentation to make sure I understand all its implications. While a bit more verbose, users should already be familiar with `next` and `expect`. I don't feel strongly about it. Please do insist in case you do.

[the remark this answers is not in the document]
whereistejas: I would agree with going for `futures::select!` and `swarm.next()`. This syntax is clearer and easier to understand, for someone who is not familiar with `futures`.

examples/file-sharing.rs
mxinden: Why? First I wouldn't expect an async version of `read_to_string` to lead to any performance gain, given that the actual context switching likely makes up only a small percentage of the overhead involved and second I don't think performance matters in the first place in this case.

thomaseizinger: That depends a lot on how big the file is, no?

But in any case, the main argument I'd bring forward is that IMO examples shouldn't contain shortcuts that can present footguns but rather should show idiomatic code that can be taken and extended.

Using a sync fs API within an async fn can go wrong if the constraints or requirements around it change.

I don't feel super strongly about it though, so feel free to leave as is! :)

examples/file-sharing.rs
mxinden: It is just the trait off between better error handling and yet another dependency increasing compile times / CI times. Given that the main purpose of this example is not showcasing great error handling, I went without `anyhow`. I cleaned up the `map_err` a bit now though.

examples/file-sharing.rs
mxinden: Good catch. Thanks!

examples/file-sharing.rs
```diff
@@ -0,0 +1,639 @@
...
+                            if let Some(sender) = outstanding_dial.remove(&peer_id) {
+                                sender.send(());
+                            }
+                        }
+                    }
+                    SwarmEvent::ConnectionClosed { .. } => {}
+                    SwarmEvent::UnreachableAddr { .. } => {}
+                    e => panic!("{:?}", e),
+                },
+                command = command_receiver.next() => match command {
```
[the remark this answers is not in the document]
mxinden: I am in favor of splitting the large `select!`, though moving it to separate functions would require passing the `pending_*` along with each call. With that in mind, I am not sure it would be an improvement.

examples/file-sharing.rs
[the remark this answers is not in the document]
wngr: In this particular case I consider it just to be syntactic sugar, which I mostly like very much given libp2p's verbosity :-)

examples/file-sharing.rs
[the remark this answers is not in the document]
wngr: Or one could use a custom struct wrapping all state providing custom functions to interact with said state. This might also yield the opportunity to DRY.

elenaf9 (approved):

examples/file-sharing.rs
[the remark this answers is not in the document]
mxinden: Thanks for the input @elenaf9 and @wngr. e05117d092a949bb5fa0e77e26a3c5994c91939f splits the `select!` into separate `async fn`s.

mxinden:
I very much appreciate all the input @thomaseizinger, @elenaf9, @wngr and @whereistejas! Great to see the many perspectives.

I addressed most but not all of your comments. I will merge this already as I find it to be in a useful state today. Follow-up improvements are always welcome.

Hope this pull request makes getting-started with rust-libp2p a little easier.

PR #2187, opened 2021-08-09
src/tutorial: Enhance docs on multiaddrs

elenaf9:
With #2100, the tutorial and examples were changed to print not only one listening address, but instead each one that is reported via a `SwarmEvent::NewListenAddr`. This resulted in the docs not fitting the actual stdout prints, as they implied that there is only one listening address.
This PR fixes the docs, please feel free to nitpick my wording.

### DIFF
--- a/examples/chat.rs
+++ b/examples/chat.rs
@@ -30,15 +30,15 @@
 //!
 //! # If they don't automatically connect
 //!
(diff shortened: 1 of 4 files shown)


mxinden (approved):
Thanks @elenaf9. Very much appreciated!

PR #2183, opened 2021-08-06
core/: Remove TInEvent and TOutEvent

mxinden:
TInEvent and TOutEvent are implied through THandler and thus
superflucious. Both are removed in favor of a derivation through
THandler.

Dependant on https://github.com/libp2p/rust-libp2p/pull/2182. Still some outstanding TODOs.

PR #2191, opened 2021-08-12
{core/,swarm/}: Dial with handler and return handler on error and closed

mxinden:
### Background

As promised in https://github.com/libp2p/rust-libp2p/discussions/2118#discussioncomment-1126577 I wanted to trial adding something similar to `ProtocolsHandler::OutboundOpenInfo` for connections instead of substreams to `NetworkBehaviourAction::{DialPeer,DialAddress}`.

https://github.com/libp2p/rust-libp2p/discussions/2118#discussioncomment-991633 details the benefits. Long story short, it allows one to attach state to a dial request, thus not having to keep track of it within a `NetworkBehaviour` implementation.

While trialing the above, I had the ideas for https://github.com/libp2p/rust-libp2p/pull/2182 and https://github.com/libp2p/rust-libp2p/pull/2183, both simplifying trait parameters, one in `libp2p-swarm` and one in `libp2p-core`. Passing an opaque `OpenInfo` type along with `NetworkBehaviourAction::{DialPeer,DialAddress}` would require yet another trait parameter. To circumvent this additional complexity, I came up with the below, in other words with the change proposed in this pull request.

### Proposal

- Require `NetworkBehaviourAction::{DialPeer,DialAddress}` to contain a `ProtocolsHandler` (see diff below). This allows a behaviour to attach custom state to its handler. The behaviour would no longer need to track this state separately during connection establishment, thus reducing state required in a behaviour. E.g. in the case of `libp2p-kad` the behaviour can include a `GetRecord` request in its handler, or e.g. in the case of `libp2p-request-response` the behaviour can include the first request in the handler.

    <details>
      <summary>Diff</summary>

    ```diff
    +pub enum NetworkBehaviourAction<TInEvent, TOutEvent, THandler> {
         /// Instructs the `Swarm` to return an event when it is being polled.
         GenerateEvent(TOutEvent),
     
    @@ -237,6 +238,7 @@ pub enum NetworkBehaviourAction<TInEvent, TOutEvent> {
         DialAddress {
             /// The address to dial.
             address: Multiaddr,
    +        handler: THandler,
         },
     
         /// Instructs the swarm to dial a known `PeerId`.
    @@ -254,6 +256,7 @@ pub enum NetworkBehaviourAction<TInEvent, TOutEvent> {
             peer_id: PeerId,
             /// The condition for initiating a new dialing attempt.
             condition: DialPeerCondition,
    +        handler: THandler,
         },
    ```

    </details>

- Return `ProtocolsHandler` on connection error and close (see diff below). This allows a behaviour to extract its custom state previously included in the handler on connection failure and connection closing. E.g. in the case of `libp2p-kad` the behaviour could extract the attached `GetRecord` from the handler of the failed connection and then start another connection attempt with a new handler with the same `GetRecord` or bubble up an error to the user.

    <details>
      <summary>Diff</summary>

    ```diff
    diff --git a/core/src/network/event.rs b/core/src/network/event.rs
    index 7b415826..82f12b13 100644
    --- a/core/src/network/event.rs
    +++ b/core/src/network/event.rs
    @@ -92,6 +92,7 @@ where
             send_back_addr: Multiaddr,
             /// The error that happened.
             error: PendingConnectionError<TTrans::Error>,
    +        handler: THandler,
         },
     
         /// A new connection to a peer has been established.
    @@ -124,12 +125,13 @@ where
             error: Option<ConnectionError<<THandler::Handler as ConnectionHandler>::Error>>,
             /// The remaining number of established connections to the same peer.
             num_established: u32,
    +        handler: THandler::Handler,
         },
     
         /// A dialing attempt to an address of a peer failed.
         DialError {
             /// The number of remaining dialing attempts.
    -        attempts_remaining: u32,
    +        attempts_remaining: DialAttemptsRemaining<THandler>,
     
             /// Id of the peer we were trying to dial.
             peer_id: PeerId,
    @@ -148,6 +150,8 @@ where
     
             /// The error that happened.
             error: PendingConnectionError<TTrans::Error>,
    +
    +        handler: THandler,
         },
     
         /// An established connection produced an event.
    @@ -169,6 +173,21 @@ where
         },
     }
     
    +pub enum DialAttemptsRemaining<THandler> {
    +    // TODO: Make this a NonZeroU32.
    +    Some(u32),
    +    None(THandler),
    +}
    ```

    </details>

What do people think? Do you see additional use-cases this would support or do you see use-cases that would not be but should be supported? Especially interested n @thomaseizinger opinion given your initial proposal on https://github.com/libp2p/rust-libp2p/discussions/2118.


thomaseizinger (commented):
Very nice idea. I think it is a good direction.

swarm/src/behaviour.rs
```diff
@@ -228,15 +242,24 @@ pub trait NetworkBehaviourEventProcess<TEvent> {
...
+// Note: `TInEvent` is needed to be able to implement
+// [`NetworkBehaviourAction::map_in`], mapping the handler `InEvent` leaving the
+// handler itself untouched.
 #[derive(Debug)]
-pub enum NetworkBehaviourAction<TInEvent, TOutEvent> {
+pub enum NetworkBehaviourAction<
+    TOutEvent,
+    THandler: IntoProtocolsHandler,
+    TInEvent = THandlerInEvent<THandler>,
+> {
```
thomaseizinger, line 254: Similar to `THandlerInEvent`, I think we could define `NetworkBehaviourAction` in terms of the `NetworkBehaviour` trait. That would reduce the amount of type parameters down to one.

It would also make it easier to refer to because you can just say `NetworkBehaviourAction<Self>` within the `NetworkBehaviour`.

Thoughts?

thomaseizinger: Ah we would have to give up our `map_XYZ` functions then.

swarm/src/lib.rs
```diff
@@ -684,17 +707,21 @@ where
                 }) => {
                     log::debug!(
                         "Connection attempt to {:?} via {:?} failed with {:?}. Attempts remaining: {}.",
-                        peer_id, multiaddr, error, attempts_remaining);
+                        // TODO: Can we do better on conversion?
```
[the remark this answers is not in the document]
mxinden: Fixed via https://github.com/libp2p/rust-libp2p/pull/2191/commits/a56980e21f597c344d94a5a62ffd98cb7abd45e8.

swarm/src/protocols_handler/node_handler.rs
```diff
@@ -65,6 +65,11 @@ where
         self.substream_upgrade_protocol_override = version;
         self
     }
+
+    // TODO: Rethink this method.
```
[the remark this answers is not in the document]
mxinden: That is an option worth exploring. Though I don't think we should do so in this pull request to separate concerns. Note that moving `ProtocolsHandler` into `libp2p-core` would require moving concepts like `KeepAlive` into `libp2p-core` as well.

swarm/src/behaviour.rs
mxinden: Good idea @thomaseizinger which I started to explore. See https://github.com/mxinden/rust-libp2p/pull/5

> Ah we would have to give up our map_XYZ functions then.

Not as far as I can tell with additional trait parameters (each defaulting to a derivative of `TBehaviour`). See PR above.

Unfortunately I don't see a way to _use_ `TBehaviour` in `NetworkBehaviourAction`. Unless I do so, compilation fails.

Let's move the discussion to https://github.com/mxinden/rust-libp2p/pull/5.

mxinden:
Reviews of all kinds are very much appreciated! (//CC @thomaseizinger in case you have time and interest.)

swarm/src/behaviour.rs
```diff
@@ -69,16 +69,20 @@ pub trait NetworkBehaviour: Send + 'static {
...
     ///
-    /// The network behaviour (ie. the implementation of this trait) and the handlers it has
-    /// spawned (ie. the objects returned by `new_handler`) can communicate by passing messages.
-    /// Messages sent from the handler to the behaviour are injected with `inject_event`, and
-    /// the behaviour can send a message to the handler by making `poll` return `SendEvent`.
+    /// The network behaviour (ie. the implementation of this trait) and the handlers it has spawned
+    /// (ie. the objects returned by `new_handler`) can communicate by passing messages. Messages
+    /// sent from the handler to the behaviour are injected with [`NetworkBehaviour::inject_event`],
+    /// and the behaviour can send a message to the handler by making [`NetworkBehaviour::poll`]
+    /// return [`NetworkBehaviourAction::SendEvent`].
```
thomaseizinger: Did you rename it? Otherwise I think this is actually called `NotifyHandler` :)

mxinden: :+1: It was previously `SendEvent`, I changed it to `['NetworkBehaviour::SendEvent']`. https://github.com/libp2p/rust-libp2p/pull/2191/commits/4d0faf9af81a5b1f3373b062e2a7cb498eb97dba changes it to `['NetworkBehaviour::NotifyHandler']`.

swarm/src/lib.rs
```diff
@@ -1704,4 +1704,9 @@ mod tests {
             }
         }))
     }
+
+    /// [`NetworkBehaviourAction::DialAddress`] and [`NetworkBehaviourAction::DialPeer`] require a
+    /// handler. This handler can be used to carry state. See corresponding doc comments.
+    #[test]
+    fn use_handler_to_carry_state() {}
```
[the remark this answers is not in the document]
mxinden: Ups. Thanks for the pointer. Removed in https://github.com/libp2p/rust-libp2p/pull/2191/commits/60c7261f665e3101e2c219104bcc3610c78cebe7.

mxinden:
Unless there are any objections or further suggestions, I will merge here sometime this week.

protocols/gossipsub/src/behaviour.rs
```diff
@@ -3169,47 +3174,17 @@ where
...
-                } => NetworkBehaviourAction::CloseConnection {
-                    peer_id,
-                    connection,
-                },
-            });
+            let event: NetworkBehaviourAction<
+                Self::OutEvent,
+                Self::ProtocolsHandler,
+                Arc<GossipsubHandlerIn>,
+            > = event;
```
thomaseizinger, line 3183: Based on my local testing, this is unnecessary.
```suggestion
```

thomaseizinger (approved):
LGTM

protocols/gossipsub/src/behaviour.rs
mxinden: Oh, a remnant of me debugging the `NetworkBehaviourAction::map_in` logic. As always, much appreciated @thomaseizinger :pray:

PR #2202, opened 2021-08-26
Wrap all comments at width 100

thomaseizinger:
Originated from this comment: https://github.com/libp2p/rust-libp2p/pull/2107#discussion_r673208592 cc @mxinden

In order to use this rustfmt setting, we need to define a custom alias
because the setting is not actually stable yet. Fortunately, we seem to
be able to use unstable settings when defined through the commandline!

The default value for wrapping comments is 80 which resulted in quite an
invasive diff. Most comments seem to be wrapped at around 100.


core/src/connection/handler.rs
```diff
@@ -28,8 +28,8 @@ use std::{fmt::Debug, task::Context, task::Poll};
 pub trait ConnectionHandler {
     /// The inbound type of events used to notify the handler through the `Network`.
     ///
-    /// See also [`EstablishedConnection::notify_handler`](super::EstablishedConnection::notify_handler)
-    /// and [`ConnectionHandler::inject_event`].
+    /// See also [`EstablishedConnection::notify_handler`](super::EstablishedConnection::
+    /// notify_handler) and [`ConnectionHandler::inject_event`].
```
mxinden, line 32: This change seems to be breaking CI, more specifically it seems to not be a valid intra doc link.

examples/file-sharing.rs
```diff
@@ -69,9 +68,9 @@
 //!
 //!    ```
 //!    cargo run --example file-sharing -- \
-//!              --peer /ip4/127.0.0.1/tcp/40837/p2p/12D3KooWPjceQrSwdWXPyLLeABRXmuqt69Rg3sBYbU1Nft9HyQ6X \
-//!              get \
-//!              --name <name-for-others-to-find-your-file>
+//!              --peer
+//! /ip4/127.0.0.1/tcp/40837/p2p/12D3KooWPjceQrSwdWXPyLLeABRXmuqt69Rg3sBYbU1Nft9HyQ6X \
```
mxinden, line 72: I am surprised that rustfmt formats code snippets. Is that the default?

protocols/relay/examples/relay.rs
```diff
@@ -24,34 +24,44 @@
...
-//!    The address specifies a static address. Usually it will be some loop back address such as `/ip4/0.0.0.0/tcp/4444`.
-//!    Example:
-//!    `cargo run --example=relay --package=libp2p-relay -- --mode relay --secret-key-seed 1 --address /ip4/0.0.0.0/tcp/4444`
-//!    `cargo run --example=relay --package=libp2p-relay -- --mode relay --secret-key-seed 1 --address /ip6/::/tcp/4444`
+//! 1. To start the relay server, run `cargo run --example=relay --package=libp2p-relay --mode relay
+//! --secret-key-seed 1 --address /ip4/<ip address>/tcp/<port>`.    The `-secret-key-seed` helps
+//! create a static peer id using the given number argument as  a seed.    The mode specifies
+//! whether the node should run as a relay server, a listening client or a dialing client.
+//!    The address specifies a static address. Usually it will be some loop back address such as
+//! `/ip4/0.0.0.0/tcp/4444`.    Example:
```
mxinden, line 32: While maybe not properly formatted before, I don't think this is either.

core/src/connection/handler.rs
thomaseizinger: Yes, I guess there is a reason why this setting is unstable :(

If I manage to change these docs in a way so the formatting doesn't break the link: Would this be acceptable to merge?

examples/file-sharing.rs
thomaseizinger: Uhm, I think in this case it is not about code snippets but just comments in general.

protocols/relay/examples/relay.rs
thomaseizinger: Fair enough, I'll look into it and see if I can make things work somehow.

protocols/gossipsub/src/peer_score/params.rs
```diff
@@ -117,12 +117,13 @@ pub struct PeerScoreParams {
...
-    ///  The weight of the parameter MUST be negative, unless you want to disable for testing.
-    ///  Note: In order to simulate many IPs in a manageable manner when testing, you can set the weight to 0
-    ///        thus disabling the IP colocation penalty.
+    ///  If the number of peers in the same IP exceeds `ip_colocation_factor_threshold, then the
+    /// value  is the square of the difference, ie `(peers_in_same_ip -
+    /// ip_colocation_threshold)^2`.  If the number of peers in the same IP is less than the
+    /// threshold, then the value is 0.  The weight of the parameter MUST be negative, unless
+    /// you want to disable for testing.  Note: In order to simulate many IPs in a manageable
+    /// manner when testing, you can set the weight to 0        thus disabling the IP
+    /// colocation penalty.
```
mxinden, line 126:
```suggestion
    /// If the number of peers in the same IP exceeds `ip_colocation_factor_threshold, then the
    /// value  is the square of the difference, ie `(peers_in_same_ip -
    /// ip_colocation_threshold)^2`. If the number of peers in the same IP is less than the
    /// threshold, then the value is 0. The weight of the parameter MUST be negative, unless
    /// you want to disable for testing.
    ///
    /// Note: In order to simulate many IPs in a manageable
    /// manner when testing, you can set the weight to 0 thus disabling the IP
    /// colocation penalty.
```

Multiple double spaces and the `Note: ` should be on a new line.

protocols/relay/examples/relay.rs
```diff
@@ -24,35 +24,53 @@
...
 //!
-//! 1. To start the relay server, run `cargo run --example=relay --package=libp2p-relay --mode relay --secret-key-seed 1 --address /ip4/<ip address>/tcp/<port>`.
-//!    The `-secret-key-seed` helps create a static peer id using the given number argument as  a seed.
-//!    The mode specifies whether the node should run as a relay server, a listening client or a dialing client.
-//!    The address specifies a static address. Usually it will be some loop back address such as `/ip4/0.0.0.0/tcp/4444`.
-//!    Example:
-//!    `cargo run --example=relay --package=libp2p-relay -- --mode relay --secret-key-seed 1 --address /ip4/0.0.0.0/tcp/4444`
-//!    `cargo run --example=relay --package=libp2p-relay -- --mode relay --secret-key-seed 1 --address /ip6/::/tcp/4444`
+//! 1. To start the relay server, run
+//! ```bash
```
mxinden, line 28: If I am not mistaken, this would not render correctly in markdown. While not relevant in this case, I wonder whether it would in future cases?

transports/plaintext/src/lib.rs
```diff
@@ -41,10 +41,10 @@ mod structs_proto {
 
 /// `PlainText1Config` is an insecure connection handshake for testing purposes only.
 ///
-/// > **Note**: Given that `PlainText1Config` has no notion of exchanging peer identity information it is not compatible
-/// > with the `libp2p_core::transport::upgrade::Builder` pattern. See
-/// > [`PlainText2Config`](struct.PlainText2Config.html) if compatibility is needed. Even though not compatible with the
-/// > Builder pattern one can still do an upgrade *manually*:
+/// > **Note**: Given that `PlainText1Config` has no notion of exchanging peer identity information
+/// it is not compatible > with the `libp2p_core::transport::upgrade::Builder` pattern. See
```
mxinden, line 45: This is weird no? I would expect either a single `>` at the first line, or a `>` on each line.

protocols/relay/examples/relay.rs
thomaseizinger: I don't understand what what would be incorrect about the rendering? This is what it looks like if I let CLion render it:

image

mxinden: Sorry to be a bit nit picky. The paragraph starting with "The `-secret-key-seed` helps create [...]" should be indented on the level of "1.", but in the rendered version above it is not indented at all.

PR #2212, opened 2021-08-31
Don't fire all timers at the same time.

dvc94ch:
here is the commit from the mdns-improvements branch.


mxinden:
Great. Thanks. Mind adding a changelog entry as well?

mxinden (approved):

PR #2214, opened 2021-09-02
Make event_process = false the default

### LINKED ISSUE #2190: Make `event_process = false` the default on `derive(NetworkBehaviour)`
With https://github.com/libp2p/rust-libp2p/pull/2186, we are adding the until this day most-comprehensive example on how to use `rust-libp2p`.

This example is using `event_process = false` to integrate nicely with an eventloop-based design.

Back when it was added, we wanted to stay backwards compatible but at least in my view, setting `event_process` to `false` deemed to be very useful and "idiomatic".

Not having it the default is causing [friction](https://github.com/libp2p/rust-libp2p/discussions/2177#discussioncomment-1160578) for users trying to follow our examples. (cc @whereistejas)

What are people's thoughts on making this the default with one of the next releases?

thomaseizinger:
Fixes #2190.

I've converted some parts of the codebase to use event_process = false now and left others. Let me know if you want any more changes :)


mxinden (approved):
Good idea. Thanks @thomaseizinger!

PR #2215, opened 2021-09-02
Revise naming of libp2p-ping symbols

thomaseizinger:
As discussed in #2174.


mxinden (approved):
Thanks @thomaseizinger.

Issue #2225, opened 2021-09-10
Build is broken on nightly


kpp:
commit fcb2f62752d465a0

```
rustc -vV
rustc 1.57.0-nightly (fdf65053e 2021-09-07)
binary: rustc
commit-hash: fdf65053e99e8966f9bd83b5a8491326cb33d638
commit-date: 2021-09-07
host: x86_64-unknown-linux-gnu
release: 1.57.0-nightly
LLVM version: 13.0.0
```

```
cargo b

error[E0034]: multiple applicable items in scope
   --> /home/kpp/.cargo/registry/src/github.com-1ecc6299db9ec823/prost-derive-0.7.0/src/lib.rs:109:14
    |
109 |             .intersperse(quote!(|));
    |              ^^^^^^^^^^^ multiple `intersperse` found
    |
    = note: candidate #1 is defined in an impl of the trait `Iterator` for the type `Map<I, F>`
    = note: candidate #2 is defined in an impl of the trait `Itertools` for the type `T`
help: disambiguate the associated function for candidate #1
    |
```

PR #2226, opened 2021-09-10
Update libp2p-rendezvous to fix build on nightly

kpp:
Fix https://github.com/libp2p/rust-libp2p/issues/2225


whereistejas:
haha, I was about to submit the same PR, myself.

thomaseizinger:
`libp2p-rendezvous` is not released yet so the version doesn't need to be bumped!

Otherwise LGTM.

We need a CI to check against failure on nightly.

mxinden (approved):
Thanks @kpp!

dvc94ch:
do we need three different PR's to fix the build on master? I'd suggest adding the clippy commit and updating clang to v11 in this PR. if it takes a week to review a PR it'll take three weeks to review three PR's.

PR #2229, opened 2021-09-11
protocols/rendezvous: Improve examples

wngr:
* Add support for the `Identify` protocol to the server, such that the
  `register_with_identify` example works as intended
* Add discovery loop to the `discovery` example and demonstrate cookie
  usage


protocols/rendezvous/examples/discover.rs
```diff
@@ -23,14 +23,14 @@ use libp2p::core::identity;
...
-use libp2p::swarm::Swarm;
 use libp2p::swarm::SwarmEvent;
+use libp2p::Swarm;
 use libp2p::{development_transport, rendezvous, Multiaddr};
 use std::time::Duration;
 
 const NAMESPACE: &str = "rendezvous";
 
-#[async_std::main]
+#[tokio::main]
```
thomaseizinger, line 33: Should probably drop the `async_std` dependency then.

thomaseizinger (approved):
LGTM!

dvc94ch:
is mxinden still the only one with write access to the repo? maybe it's time to add @thomaseizinger

thomaseizinger:
> is mxinden still the only one with write access to the repo? maybe it's time to add @thomaseizinger 

Thanks for the nomination! There is an ongoing discussion about becoming a maintainer, should be sorted once @mxinden comes back from vacation!

protocols/rendezvous/examples/discover.rs
```diff
@@ -53,62 +57,77 @@ async fn main() {
...
-            SwarmEvent::ConnectionEstablished { peer_id, .. } if peer_id == rendezvous_point => {
-                log::info!(
-                    "Connected to rendezvous point, discovering nodes in '{}' namespace ...",
-                    NAMESPACE
-                );
+    let mut discover_tick = tokio::time::interval(Duration::from_secs(30));
+    let mut cookie = None;
 
+    loop {
+        tokio::select! {
```
dvc94ch, line 64: `futures::select` would be more generic, presumably this is just a reexport

dvc94ch (approved):

protocols/rendezvous/examples/discover.rs
thomaseizinger: It is not, the implementation is in fact very different: https://github.com/tokio-rs/tokio/commit/8cf98d694663e8397bfc337e7a4676567287bfae

wngr: It's in my toolbox right now. Didn't bother trying to be runtime agnostic, as the example already used stuff from tokio.

mxinden (approved):
Thanks @wngr!

PR #2232, opened 2021-09-14
protocols/identify: Assist in peer discovery based on reported listen addresses from other peers

### LINKED ISSUE #2216: RFC: `Identify` should remember discovered peers and assist in peer discovery using `addresses_of_peer`
While implementing an address book [here](https://github.com/libp2p/rust-libp2p/pull/2133#issuecomment-881343154), I realized that we might be able to achieve the same functionality by adjusting how our existing protocols work. In particular, not all of our discovery-like protocols are currently assisting in peer discovery via `addresses_of_peer`. `Identify` for example [does not override](https://github.com/libp2p/rust-libp2p/blob/6924e5ef7a61688192ec1925fcb00d0c9da3ddba/protocols/identify/src/identify.rs#L189-L391) `addresses_of_peer`.

What do people think about changing that so that we remember the reported addresses of peers that we have discovered? I think the only issue here is that we potentially have an unbounded memory growth. We could however use a LRU cache for example and simply cap its storage at some number of entries.

Having `Identify` remember reported addresses is particularly useful if:

1. Node A knows about its own, externally reachable address
1. Node B does not know about external addresses of node A
1. Node A dials node B
1. Node B learns via `/identify` about A's external addresses
1. The Connection between node A and B gets closed

If `Identify` would remember the externally reachable addresses that have previously been identified, Node B would now be able to establish a connection to A by simply dialing its `PeerId`.
Currently, users of `rust-libp2p` have to actively handle `IdentifyEvent` and add all reported addresses to some other `NetworkBehaviour` that will report it as part of `addresses_of_peer` later.

thomaseizinger:
Resolves #2216.

### DIFF
--- a/protocols/identify/src/identify.rs
+++ b/protocols/identify/src/identify.rs
@@ -162,7 +162,10 @@ impl IdentifyConfig {
         self
     }
 
-    /// Configures the size of the LRU cache for discovered peers.
+    /// Configures the size of the LRU cache, caching addresses of discovered peers.
+    ///
+    /// The [`Swarm`] may extend the set of addresses of an outgoing connection attempt via
+    ///  [`<Identify as NetworkBehaviour>::addresses_of_peer`].
     pub fn with_cache_size(mut self, cache_size: usize) -> Self {
         self.cache_size = cache_size;
         self
@@ -272,7 +275,7 @@ impl NetworkBehaviour for Identify {
     ) {
         match event {
             IdentifyHandlerEvent::Identified(info) => {
-                // replace existing addresses to prevent other peer from filling up our memory
+                // Replace existing addresses to prevent other peer from filling up our memory.
                 self.discovered_peers
                     .put(peer_id, HashSet::from_iter(info.listen_addrs.clone()));
 
--- a/protocols/identify/src/identify.rs
+++ b/protocols/identify/src/identify.rs
@@ -272,15 +272,9 @@ impl NetworkBehaviour for Identify {
     ) {
         match event {
             IdentifyHandlerEvent::Identified(info) => {
-                match self.discovered_peers.get_mut(&peer_id) {
-                    Some(entry) => {
-                        entry.extend(info.listen_addrs.clone());
-                    }
-                    None => {
-                        self.discovered_peers
-                            .put(peer_id, HashSet::from_iter(info.listen_addrs.clone()));
-                    }
-                }
+                // replace existing addresses to prevent other peer from filling up our memory
+                self.discovered_peers
+                    .put(peer_id, HashSet::from_iter(info.listen_addrs.clone()));
 
                 let observed = info.observed_addr.clone();
                 self.events.push_back(NetworkBehaviourAction::GenerateEvent(
--- a/protocols/identify/src/identify.rs
+++ b/protocols/identify/src/identify.rs
@@ -31,9 +31,11 @@ use libp2p_swarm::{
     NetworkBehaviour, NetworkBehaviourAction, NotifyHandler, PollParameters, ProtocolsHandler,
     ProtocolsHandlerUpgrErr,
 };
+use lru::LruCache;
 use std::{
     collections::{HashMap, HashSet, VecDeque},
     io,
+    iter::FromIterator,
     pin::Pin,
     task::Context,
     task::Poll,
@@ -57,6 +59,8 @@ pub struct Identify {
     /// Peers to which an active push with current information about
     /// the local peer should be sent.
     pending_push: HashSet<PeerId>,
+    /// The addresses of all peers that we have discovered.
+    discovered_peers: LruCache<PeerId, HashSet<Multiaddr>>,
 }
 
 /// A pending reply to an inbound identification request.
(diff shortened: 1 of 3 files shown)


mxinden (commented):
Cross posting review here as well:

https://github.com/libp2p/rust-libp2p/issues/2216#issuecomment-932968891

protocols/identify/src/identify.rs
```diff
@@ -254,6 +272,16 @@ impl NetworkBehaviour for Identify {
     ) {
         match event {
             IdentifyHandlerEvent::Identified(info) => {
+                match self.discovered_peers.get_mut(&peer_id) {
+                    Some(entry) => {
+                        entry.extend(info.listen_addrs.clone());
```
mxinden: Do I understand correctly that the `HashSet` of `Multiaddr`s for a given peer could potentially grow endlessly, e.g. when the remote (malicious) peer sends a new set of addresses on each `identify` exchange? How about replacing the previous set with the new one instead? Would as well make sure outdated addresses are removed.

protocols/identify/src/identify.rs
```diff
@@ -254,6 +272,16 @@ impl NetworkBehaviour for Identify {
     ) {
         match event {
             IdentifyHandlerEvent::Identified(info) => {
```
mxinden, line 277: While we do have a limit of `4096 bytes` per identify message, what do you think of enforcing a limit on the number of addresses here anyways?

https://github.com/libp2p/rust-libp2p/blob/937b59de8946eb3773e292691dac4e5f3e081706/protocols/identify/src/protocol.rs#L204-L206

mxinden (commented):
Couple of comments. Overall I think this is a good idea.

protocols/identify/src/identify.rs
thomaseizinger: Replacing the set is a good idea!

protocols/identify/src/identify.rs
thomaseizinger: What do you think would be an appropriate limit?

Unless we can inform the other peer somehow, silently capping the number of addresses here could result in a nasty bug.

I think the overall message size is a good enough limit that will wake someone somewhere when it is hit! (Assuming the connection gets closed if the message is too large)

protocols/identify/src/identify.rs
```diff
@@ -152,17 +161,26 @@ impl IdentifyConfig {
         self.push_listen_addr_updates = b;
         self
     }
+
+    /// Configures the size of the LRU cache for discovered peers.
```
mxinden:
```suggestion
    /// Configures the size of the LRU cache, caching addresses of discovered peers.
    ///
    /// The [`Swarm`] may extend the set of addresses of an outgoing connection attempt via
    ///  [`<Identify as NetworkBehaviour>::addresses_of_peer`].
```

protocols/identify/src/identify.rs
```diff
@@ -254,6 +272,10 @@ impl NetworkBehaviour for Identify {
     ) {
         match event {
             IdentifyHandlerEvent::Identified(info) => {
+                // replace existing addresses to prevent other peer from filling up our memory
```
mxinden:
```suggestion
                // Replace existing addresses to prevent other peer from filling up our memory.
```

mxinden (commented):
Two small comments. Missing a changelog entry. Other than that, this looks good to me.

mxinden (approved):
Merci!

protocols/identify/src/identify.rs
```diff
@@ -152,17 +161,29 @@ impl IdentifyConfig {
         self.push_listen_addr_updates = b;
         self
     }
+
+    /// Configures the size of the LRU cache, caching addresses of discovered peers.
+    ///
+    /// The [`Swarm`] may extend the set of addresses of an outgoing connection attempt via
+    ///  [`<Identify as NetworkBehaviour>::addresses_of_peer`].
```
mxinden:
```suggestion
    ///  [`Identify::addresses_of_peer`].
```

Fixing 

```
error: unresolved link to `<Identify as NetworkBehaviour>::addresses_of_peer`
   --> protocols/identify/src/identify.rs:168:11
    |
168 |     ///  [`<Identify as NetworkBehaviour>::addresses_of_peer`].
    |           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ fully-qualified syntax is unsupported
    |
    = note: see https://github.com/rust-lang/rust/issues/74563 for more information
```

protocols/identify/src/identify.rs
```diff
@@ -152,17 +161,29 @@ impl IdentifyConfig {
         self.push_listen_addr_updates = b;
         self
     }
+
+    /// Configures the size of the LRU cache, caching addresses of discovered peers.
+    ///
+    /// The [`Swarm`] may extend the set of addresses of an outgoing connection attempt via
```
mxinden:
```suggestion
    /// The [`Swarm`](libp2p_swarm::Swarm) may extend the set of addresses of an outgoing connection attempt via
```

Fixing

```
 error: unresolved link to `Swarm`
   --> protocols/identify/src/identify.rs:167:14
    |
167 |     /// The [`Swarm`] may extend the set of addresses of an outgoing connection attempt via
    |              ^^^^^^^ no item named `Swarm` in scope
    |
    = note: `-D rustdoc::broken-intra-doc-links` implied by `-D broken-intra-doc-links`
    = help: to escape `[` and `]` characters, add '\' before them like `\[` or `\]`
```

PR #2240, opened 2021-09-20
core/src/transport: add CircuitBreaking wrapper

### LINKED ISSUE #1007: Add a Transport wrapper for nodes-hammering-prevention

mooori:
Close #1007

This pull request is heavily inspired by #1010.

As proposed in the conversation of #1010, the new type is called `CircuitBreaking` (instead of `HammerPrevention`).

The type `CircuitBreakerState` centralizes circuit handling. 

There are many ways in which circuit breaking could be made more sophisticated (e.g. as described [here](https://docs.microsoft.com/en-us/azure/architecture/patterns/circuit-breaker)). A central `CircuitBreakerState` aims at facilitating such enhancements.


core/src/transport/circuit_breaking.rs
```diff
@@ -0,0 +1,383 @@
...
+                        Poll::Pending => return Poll::Pending,
+                        Poll::Ready(()) => {
+                            let transport = transport
+                                .take()
+                                .expect("future called after being finished");
+                            match transport.dial(this.addr.clone()) {
+                                Ok(f) => this.inner.set(CircuitBreakingDialInner::Dialing(f)),
+                                Err(TransportError::MultiaddrNotSupported(_addr)) => {
+                                    // TODO do something other than panic?
+                                    panic!("Open circuit for unsupported Multiaddr")
```
mooori, line 230: I assumed this arm is not supposed to be executed and therefore it's ok to `panic!`. Not sure, though. Also there might be a more elegant solution.

mxinden: Instead of delaying the call to `TInner::dial`, one could do the call in `CircuitBreaking::dial` directly, only delaying the polling of the `TInner::Dial`, not the creation. That of course assumes that `Tinner::Dial` does nothing without being polled. I am not sure whether all implementations of `Transport` uphold this assumption.

core/src/transport/circuit_breaking.rs
```diff
@@ -0,0 +1,383 @@
...
+    cmp,
+    collections::hash_map::Entry,
+    pin::Pin,
+    sync::Arc,
+    task::{Context, Poll},
+    time::Duration,
+};
+
+/// Timeout set when opening CircuitBreaker.
+const INITIAL_TIMEOUT: Duration = Duration::from_secs(2);
```
mxinden: 2 seconds seems like a long time on first failure. What is the reasoning for this high value?

core/src/transport/circuit_breaking.rs
```diff
@@ -0,0 +1,383 @@
...
+    }
+
+    /// Closes the circuit of `addr` even if the circuit has not yet expired.
+    fn close_circuit(&self, addr: &Multiaddr) {
+        self.addresses.lock().remove(addr);
+    }
+
+    fn close_expired_circuits(&self) {
+        let mut addresses = self.addresses.lock();
+        addresses.retain(move |_, a| (&mut a.expiration).now_or_never().is_none());
```
mxinden: If I am not mistaken, `AddrInfo::expiration` could as well be a `std::time::Instant` and a simple comparison here (`a.expiration < now`) would suffice, right?

core/src/transport/circuit_breaking.rs
```diff
@@ -0,0 +1,383 @@
...
+    type Error = TInner::Error;
+    type Listener = TInner::Listener;
+    type ListenerUpgrade = TInner::ListenerUpgrade;
+    type Dial = CircuitBreakingDial<TInner>;
+
+    fn listen_on(self, addr: Multiaddr) -> Result<Self::Listener, TransportError<Self::Error>> {
+        self.inner.listen_on(addr)
+    }
+
+    fn dial(self, addr: Multiaddr) -> Result<Self::Dial, TransportError<Self::Error>> {
```
mxinden, line 197: `addr` might include a `PeerId`.

In large public networks (e.g. IPFS) many nodes advertise LAN local addresses. In most cases nodes don't share the same LAN and thus dialing these addresses fails.

Ignoring the `PeerId` in an address in `CircuitBreakerState` would prevent dialing say `192.168.2.108` multiple times in a row.

What do you think?

mxinden (commented):
Thank you for the pull request @mooori and sorry for the delayed review!

Overall this looks good to me. I like the simplicity of the implementation.

Do you have a concrete use-case for this? If so, out of curiosity, would you mind describing it here? I am undecided how useful the circuit breaking pattern in a p2p setting is.

Also, would you mind adding a couple of `log::debug` lines to ease debugging?

mooori:
Thank you @mxinden for the review and feedback!

I have to admit: I don't have a use-case for circuit breaking. I started working on it based on the issue and PR mentioned in the description.

Now I realize I should have gathered feedback _before_ working on this (rather dated) issue. My bad!

That said, do you think it makes sense to pursue this PR further?

If yes, I'm happy to tackle the improvements pointed out in your comments :)

mxinden:
> Now I realize I should have gathered feedback before working on this (rather dated) issue. My bad!

Yes, best to comment on an issue before working on it. Though also my fault as I am not doing a great job keeping those issues up-to-date - sorry.

> That said, do you think it makes sense to pursue this PR further?

I still think a `CircuitBreaker` implementation is good to have :+1:.

core/src/transport/circuit_breaking.rs
mooori: To verify it's probably required to check all `Transport` implementations (and assume they will continue to uphold this assumption in the future).

Could a `Multiaddr` which was initially supported when the circuit was opened become unspupported while waiting for the circuit to close?

core/src/transport/circuit_breaking.rs
mooori: The value was taken from the original PR (linked in the description). Would 1 second be more reasonable?

core/src/transport/circuit_breaking.rs
mooori: Done in 101a36d.

core/src/transport/circuit_breaking.rs
mooori: Good point! Done in f4c6f67.

mooori:
Sorry for the delay! I could resume working on this PR only recently.

I just pushed the changes and pointed out the corresponding commits in the comments above.

Debug logs were added in bce0ba9.

core/src/transport/circuit_breaking.rs
```diff
@@ -80,37 +85,80 @@ impl CircuitBreakerState {
...
+///
+/// Ignoring the `PeerId` in addresses prevents dialing e.g. `192.168.2.108`
+/// multiple times in a row.
+#[derive(PartialEq, Eq, Hash, Clone, Debug)]
+struct CircuitAddr(Multiaddr);
+
+impl From<Multiaddr> for CircuitAddr {
+    fn from(mut addr: Multiaddr) -> Self {
+        // Check if `addr` contains `Protocol::P2p(_)`.
+        let contained_p2p = addr.iter().find(|protocol| match protocol {
```
mxinden:
```suggestion
        let contained_p2p = addr.iter().any(|protocol| matches!(protocol, Protocol::P2p(_));
```

Might a bit simpler, right?

core/src/transport/circuit_breaking.rs
mxinden: Off the top of my head, I would suggest 200ms.

core/src/transport/circuit_breaking.rs
mxinden: > Could a Multiaddr which was initially supported when the circuit was opened become unspupported while waiting for the circuit to close?

In theory yes. If we ever decide this to be invalid, we should change the `TransportError` type to prevent his at type level.

For now, I would suggest keeping it as is, adding a comment on `fn dial` above documenting the potential panic.

mxinden (commented):
A couple of comments but looks good to me overall.

I would suggest we merge this soon and refine it once folks use this in larger setups.

core/src/transport/circuit_breaking.rs
mooori: Addressed in 25d5e86.

core/src/transport/circuit_breaking.rs
mooori: Addressed in 93fec82. (The formatting was cleaned-up as I removed the `TODO` comment.)

core/src/transport/circuit_breaking.rs
mooori: True :) Addressed in f64e6aa.

mooori:
I just pushed the changes :)

> I would suggest we merge this soon and refine it once folks use this in larger setups.

Feel free to ping me if I can help with refining it in the future.

core/src/transport/circuit_breaking.rs
```diff
@@ -0,0 +1,458 @@
...
+                    Err(TransportError::MultiaddrNotSupported(err)) => {
+                        // Avoid open circuits for unsupported addresses.
+                        return Err(TransportError::MultiaddrNotSupported(err));
+                    }
+                    Err(err) => {
+                        self.state.open_or_extend_circuit(addr.clone());
+                        Err(err)
+                    }
+                };
+                CircuitBreakingDialInner::Dialing(future?)
```
thomaseizinger: This `?` will directly return the error from line 212. Is this intended? If yes, I think we should return it directly to avoid confusion. It also means we can get rid of the `Ok` wrapping on line 205.

core/src/transport/circuit_breaking.rs
```diff
@@ -0,0 +1,458 @@
...
+    /// Panics if `addr` becomes unsupported while waiting for its circuit to
+    /// close. For an open circuit to exist, `addr` must have been supported
+    /// when it was dialed initially.
+    fn dial(self, addr: Multiaddr) -> Result<Self::Dial, TransportError<Self::Error>> {
+        // Purge expired addresses.
+        self.state.close_expired_circuits();
+
+        let inner = match self.state.get_wait_duration(&addr) {
+            None => {
+                debug!("Dialing address immediately (no open circuit): {:?}", addr);
```
thomaseizinger: Multiaddresses support formatting with `Display`!
```suggestion
                debug!("Dialing address immediately (no open circuit): {}", addr);
```

Also, I think rewording this might be a bit clearer:

```suggestion
                debug!("No open circuit for address {}, dialing immediately", addr);
```

core/src/transport/circuit_breaking.rs
```diff
@@ -0,0 +1,458 @@
...
+                    Err(err) => {
+                        self.state.open_or_extend_circuit(addr.clone());
+                        Err(err)
+                    }
+                };
+                CircuitBreakingDialInner::Dialing(future?)
+            }
+
+            Some(wait_duration) => {
+                debug!("Wait before dialing address (open circuit): {:?}", addr);
```
thomaseizinger:
```suggestion
                debug!("Circuit for address {} is open, dialing will be delayed by {}s", addr, wait_duration.as_secs());
```

core/src/transport/circuit_breaking.rs
```diff
@@ -0,0 +1,458 @@
...
+    fn poll(self: Pin<&mut Self>, cx: &mut Context) -> Poll<Self::Output> {
+        let mut this = self.project();
+
+        loop {
+            match this.inner.as_mut().project() {
+                CircuitBreakingDialInnerProj::Waiting(delay, transport) => match delay.poll(cx) {
+                    Poll::Pending => return Poll::Pending,
+                    Poll::Ready(()) => {
+                        debug!(
+                            "Dialing address after circuit breaking wait expired: {:?}",
```
thomaseizinger:
```suggestion
                            "Dialing address after circuit breaking wait expired: {}",
```

core/src/transport/circuit_breaking.rs
```diff
@@ -0,0 +1,458 @@
...
+                        }
+                    }
+                },
+
+                CircuitBreakingDialInnerProj::Dialing(dial) => {
+                    match dial.poll(cx) {
+                        Poll::Pending => return Poll::Pending,
+                        Poll::Ready(Ok(output)) => {
+                            // On success, remove the address.
+                            debug!("Successfully dialed addr: {:?}", this.addr);
```
thomaseizinger:
```suggestion
                            debug!("Successfully dialed addr: {}", this.addr);
```

thomaseizinger (dismissed):
Couple of suggestion, only one blocking comment about the result / `?`. Otherwise LGTM :)

core/src/transport/circuit_breaking.rs
mooori: Adressed in a3fed7e.

mooori:
Thank you for the review and suggestions @thomaseizinger. I've cleaned up the `Result`/`?` and commited your suggestions regarding the logs.

thomaseizinger:
Closing this as stale. Sorry that this got lost @mooori, in case you still have a usecase for this, feel free to comment here and we can pick it back up! :)

PR #2241, opened 2021-09-21
[Gossipsub] Handle unsupported peers

AgeManning:
Previously, peers that did not support gossipsub were removed from the connection-id mappings. 

This can cause the connection_id mappings to go out of sync with those managed by the swarm. This PR corrects this and adds an extra event that can inform the user that a peer that does not support the protocol has connected. The user can then optionally handle peers that don't support the protocol.


thomaseizinger (commented):
Thanks!

I think the change in behaviour here is worth pointing out in the changelog.
Do you mind adding an entry?

AgeManning:
Yep, done :)

mxinden (approved):
All in favor of the additional event. Thanks @AgeManning.

Issue #2166, opened 2021-07-29
WASM support


genderev:
Does rust-libp2p support the wasm32-unknown-unknown target? I tried to compile the [ping](https://github.com/libp2p/rust-libp2p/blob/master/examples/ping.rs) example with wasm-pack. 

I thought the issue would be that the example relies on `println!` which is not available to the browser, but I actually got this error:

```
Compiling getrandom v0.2.3

error: the wasm32-unknown-unknown target is not supported by default, you may need to enable the "js" feature. For more information see: https://docs.rs/getrandom/#webassembly-support

```

fusetim:
You just need to add this to your Cargo.toml:
`getrandom = { version = "0.2", features = ["js"] }`

More info there: https://docs.rs/getrandom/0.2.3/getrandom/#webassembly-support

mxinden:
Closing here. Let me know if you need any further help @genderev.

Thanks @fusetim for respoding!

PR #2245, opened 2021-09-22
"Just works" WASM (browser) experience

wngr:
These are the changes I needed to make in order to get libp2p to run via `wasm32-unknown-unknown` in the browser (both main thread and inside web workers).

* [x] The only occurrence left of `wasm-timer` is in `gossipsub`, as it uses an `Interval`. Haven't gotten around to refactor that (yet).
* [x] ~The other thing is that I'm waiting for a release of the instant create (see https://github.com/sebcrozet/instant/pull/36).~
* [x] Need to replace `SystemTime` in `core/src/peer_record.rs`, probably w/ https://github.com/sebcrozet/instant/pull/9 https://github.com/sebcrozet/instant/pull/42

Looking for comments, whether this is something that can get merged, or whether there are alternative approaches.

related #2134 #2166 #2143


mxinden:
@wngr very very very much appreciated! I am sorry for the delay here. I will give this a review. I just need a bit more time.

.github/workflows/ci.yml
```diff
@@ -80,6 +87,11 @@ jobs:
       # TODO: ideally we would build `--workspace`, but not all crates compile for WASM
       run: cargo build --target=wasm32-wasi
 
+    - name: Build on wasm32-wasi
```
```suggestion
    - name: Build on wasm32-unknown-unknown
```

Cargo.toml
```diff
@@ -101,11 +104,11 @@ libp2p-tcp = { version = "0.30.0", path = "transports/tcp", default-features = f
...
 
 [dev-dependencies]
-async-std = { version = "1.6.2", features = ["attributes"] }
-async-trait = "0.1"
+async-std = { version = "1.10.0", features = ["attributes"] }
+async-trait = "0.1.51"
 env_logger = "0.9.0"
-structopt = "0.3.21"
-tokio = { version = "1.0.1", features = ["io-util", "io-std", "macros", "rt", "rt-multi-thread"] }
+structopt = "0.3.23"
```
mxinden: Is this version bump needed?

Cargo.toml
```diff
@@ -65,9 +65,12 @@ all-features = true
 
 [dependencies]
 atomic = "0.5.0"
-bytes = "1"
-futures = "0.3.1"
-lazy_static = "1.2"
+bytes = "1.1.0"
+futures = "0.3.17"
+futures-timer = "3.0.2"
+getrandom = "0.2.3"
```
mxinden: Do I understand correctly that the crate needs to be listed here in order to enable the `js` feature? If so, would you mind adding a comment? Same for the other crates?

core/src/lib.rs
```diff
@@ -66,6 +66,7 @@ pub mod upgrade;
 
 pub use connection::{Connected, ConnectedPoint, Endpoint};
 pub use identity::PublicKey;
+pub use instant::{Duration, Instant};
```
mxinden: If I am not mistaken this is the first non-libp2p-related re-export in `libp2p-core`, correct?

If so, I wonder whether it is actually needed. Neither of these two types are exposed on the public API of `libp2p-core`, right? Do versions actually have to match across libp2p crates?

.github/workflows/ci.yml
```diff
@@ -80,6 +87,11 @@ jobs:
...
+      # TODO: also run `cargo test`
```
mxinden, line 80: Yes please, that would be great!

mxinden (commented):
Just a couple of comments for now. Let me know if you need help pinging folks for https://github.com/sebcrozet/instant/pull/9.

Again, thanks for this contribution!

.github/workflows/ci.yml
```diff
@@ -60,6 +60,13 @@ jobs:
         target: wasm32-wasi
         override: true
 
+    - name: Install Rust wasm32-unknown-unknown
+      uses: actions-rs/toolchain@v1.0.7
+      with:
+        toolchain: stable
+        target: wasm32-unknown-unknown
+        override: true
```
thomaseizinger: These CI jobs changed recently with https://github.com/libp2p/rust-libp2p/pull/2260. Can you merge latest master and adapt accordingly?

Cargo.toml
thomaseizinger: I don't think it is a version bump per-se. cargo would have already resolved anything > 0.3.21 for the target binary if needed. I usually tend to only specify these to the minor version for that reason, ie:

```toml
structopt = "0.3"
```

core/src/lib.rs
thomaseizinger: +1 for not re-exporting. We shouldn't pull this in unless necessary.

transports/wasm-ext/src/websockets.js
```diff
@@ -80,7 +80,18 @@ const dial = (addr) => {
 			read: (function*() { while(ws.readyState == 1) { yield reader.next(); } })(),
 			write: (data) => {
 				if (ws.readyState == 1) {
-					ws.send(data);
+          // The passed in `data` is an `ArrayBufferView` [0]. If the
```
thomaseizinger: This file seems to be tab-indented but the new code here is using spaces which I think is why this is rendering as un-aligned?

thomaseizinger (changes requested):
Thank you!

Some comments inline, nothing big :)

core/src/lib.rs
wngr: so you're saying that all relevant crates should depend on `instant` directly, and not go via the indirection through `core` proposed here -- right?

Cargo.toml
wngr: No, not needed. I guess I ran `cargo update` at some point. Should I revert the _unnecessary_ changes?

transports/wasm-ext/src/websockets.js
wngr: Right, now it should be aligned.

Cargo.toml
mxinden: > Should I revert the unnecessary changes?

Yes please.

---

Strategy for rust-libp2p as a library is: Bump minor version (patch version on < v1.0.0) only:

- on bug fixes
- on security fixes
- when a specific feature within the minor (patch) release is needed within rust-libp2p

This makes rust-libp2p compatible with a larger range of upstream crates and thus gives downstream users greater flexibility.

(Obviously I highly recommend downstream binaries to keep their upstream dependencies up-to-date.)

core/src/lib.rs
mxinden: Yes. That would be my preference, especially since, as far as I can tell, versions of `instant` don't have to match across crates, as `instant` is not exposed on the public `libp2p-core` interface.

.github/workflows/ci.yml
wngr: That's a bigger undertaking, which might be more geared towards `wasm32-wasi` rather than `wasm32-unknown-unknown`.
I guess for `wasm32-unknown-unknown` a wasm-bindgen based browser smoke test might be something, but I'll leave that to future-us :-)

core/src/lib.rs
wngr: :+1: done

Cargo.toml
wngr: Thanks for your detailled response, reverted all unrelated changes.

Cargo.toml
```diff
@@ -141,3 +149,6 @@ members = [
 [[example]]
 name = "chat-tokio"
 required-features = ["tcp-tokio", "mdns"]
+
+[patch.crates-io]
+instant = { git = "https://github.com/wngr/instant", branch = "systemtime" }
```
wngr: pending https://github.com/sebcrozet/instant/pull/42

wngr:
I think the test failure is unrelated to this PR.

wngr:
Should be good to go for another :eyes: @mxinden @thomaseizinger

mxinden:
> I think the test failure is unrelated to this PR.

Yes. Fixed once https://github.com/libp2p/rust-libp2p/pull/2295 is merged.

.github/workflows/ci.yml
```diff
@@ -74,10 +75,17 @@ jobs:
         key: ${{ matrix.toolchain }}
 
     - name: Build on ${{ matrix.toolchain }}
+      if: matrix.toolchain != 'wasm32-unknown-unknown'
       # TODO: also run `cargo test`
       # TODO: ideally we would build `--workspace`, but not all crates compile for WASM
       run: cargo build --target=${{ matrix.toolchain }}
 
+    - name: Build on ${{ matrix.toolchain }}
```
thomaseizinger: Instead of making two jobs that only differ in their features, we could add the matrix like this:

```yaml
matrix:
  toolchain: [
    wasm32-unknown-emscripten,
    wasm32-wasi
  ]
  include:
     - toolchain: wasm32-unknown-unknown
       args: "--features wasm-bindgen"
```

And then do:

```yaml
- name: Build on ${{ matrix.toolchain }}
  run: cargo build --target=${{ matrix.toolchain }} ${{ matrix.args }}
```

That will work because GitHub actions replaces non-existent matrix keys with an empty string.

Cargo.toml
```diff
@@ -67,6 +67,12 @@ all-features = true
 atomic = "0.5.0"
 bytes = "1"
 futures = "0.3.1"
+# Explicit dependency to be used in `wasm-bindgen` feature
+futures-timer = "3.0.2"
```
thomaseizinger: Can we have the comment _next_ to the dependency? That way it is a single line and the table can more easily be sorted alphabetically.

```suggestion
futures-timer = "3.0.2" # Explicit dependency to be used in `wasm-bindgen` feature
```

protocols/gossipsub/src/interval.rs
```diff
@@ -0,0 +1,210 @@
...
+/// This is useful to divide intervals.
+///
+/// While technically for large duration it's impossible to represent any
+/// duration as nanoseconds, the largest duration we can represent is about
+/// 427_000 years. Large enough for any interval we would use or calculate in
+/// tokio.
+fn duration_to_nanos(dur: Duration) -> Option<u64> {
+    dur.as_secs()
+        .checked_mul(1_000_000_000)
+        .and_then(|v| v.checked_add(dur.subsec_nanos() as u64))
```
thomaseizinger: Nit: This could use `?` to propagate the `None` case instead of combinators.

thomaseizinger (approved):
Thanks!

Made a few suggestions but nothing blocking.

PhilippGackstatter:
Great PR! Getting rid of `wasm-timer` in favor of `instant` is much appreciated. Do these changes rely on any browser-exclusive APIs (haven't checked what `futures-timer/wasm-bindgen` binds to) or would this also work in node.js?

wngr:
I think `futures-timer`(uses `gloo-timers` under the hood) indeed does not work inside nodejs, but I've never tried it.

mxinden:
Thanks for the pull request @wngr!

As you might have seen, we already cut a `v0.40.0` release candidate, thus this pull request would be included in `v0.41.0`. In case you need a release quickly, I am happy to prepare `v0.41.0` shortly after `v0.40.0` is released.

mxinden:
Merged via https://github.com/libp2p/rust-libp2p/pull/2320. Thanks @wngr for the work!

wngr:
> Thanks for the pull request @wngr!
> 
> As you might have seen, we already cut a `v0.40.0` release candidate, thus this pull request would be included in `v0.41.0`. In case you need a release quickly, I am happy to prepare `v0.41.0` shortly after `v0.40.0` is released.

Would appreciate if you could cut create a `v0.41.0` soon-ish!

PR #2246, opened 2021-09-24
Add starcoin to rust-libp2p users

jolestar:
Starcoin is a blockchain system and uses rust-libp2p to build the p2p network.


mxinden (approved):
Thanks @jolestar!

Issue #2242, opened 2021-09-21
Invalid Swarm behaviour


AgeManning:
I've been testing some code on the latest master and running into unexpected behaviour from the Swarm. 

I've found that Swarm is injecting a `ConnectionId` to the behaviour via `inject_connection_established()`. In some less-common circumstances (which I've not isolated) upon disconnection, via `inject_connection_closed()` the Swarm is inputting a different connection id. 

In gossipsub, this causes a panic as we are expecting the swarm to inform the behaviour of the past-injected connection, not a new one.

AgeManning:
Ok, I did a bit of digging. 

It looks like in https://github.com/libp2p/rust-libp2p/pull/2191 for new connections beyond a connection limit we close the connection. Previously we reported a PendingConnectionError as a PoolEvent. 

I think we then are injecting connection closed events to the behaviour with the `ConnectionLimit` error if a connection is attempted to being established beyond the connection limit. Ultimately a behaviour will see an `inject_connection_closed()` without a matching `inject_connection_established()` for the same id. 

If this is the intended behaviour, I'll have to update some assumptions in gossipsub. 

What do you guys think? @thomaseizinger @mxinden

AgeManning:
I've also seen this panic: https://github.com/libp2p/rust-libp2p/blob/master/protocols/identify/src/identify.rs#L277

So I assume we probably need to change the behaviour of how the connections are reported.

mxinden:
Thank you @AgeManning for the debugging work. Sorry for the trouble this has caused on your end.

> It looks like in #2191 for new connections beyond a connection limit we close the connection. Previously we reported a PendingConnectionError as a PoolEvent.
> 
> I think we then are injecting connection closed events to the behaviour with the `ConnectionLimit` error if a connection is attempted to being established beyond the connection limit. Ultimately a behaviour will see an `inject_connection_closed()` without a matching `inject_connection_established()` for the same id.

[...]

Issue #1896, opened 2020-12-18
core/: Parallel connection attemps - aka. happy eyeball


quininer:
I noticed that libp2p currently [only tries next address when one address fails](https://github.com/libp2p/rust-libp2p/blob/master/core/src/network.rs#L399), which causes some users to wait a long time when trying to connect to an unavailable ipv6 address.

Maybe we can try to use a fast fallback strategy like [happy eyeballs v2](https://tools.ietf.org/html/rfc8305) to quickly connect to available addresses.

I don't know if there are some special reasons that limit connection strategy of libp2p, if not, I will be happy to implement one.

quininer:
I checked the implementation of the connection pool, which may be a lot of work.

At present, we can use `dial_addr` to try concurrent connections.

mxinden:
I am re-opening this issue. While complex for sure, it would greatly benefit hole punching.

When two peers coordinate a hole punch via the [_Direct Connection Upgrade through Relay_](https://github.com/libp2p/specs/pull/173) protocol, each peer sends the remote a list of protocols to dial. Given that both sides need to **simultaneously** dial an address in order for a hole punch to suceed, one can not expect a high hole punch success rate when trying each address sequentially.

Given that rust-libp2p does not support parallel dialing today the [rust-libp2p DCUtR implementation](https://github.com/libp2p/rust-libp2p/pull/2076) only attempts to dial the first address provided by the remote peer. Though it could instead emit multiple `DialAddress` requests at once.

mxinden:
Implemented in https://github.com/libp2p/rust-libp2p/pull/2248.

mxinden:
Cross referencing https://github.com/libp2p/go-libp2p/issues/1785 by @marten-seemann here. I think long-term rust-libp2p should adopt an advanced address prioritization like go-libp2p.

thomaseizinger:
Onw property that I'd like to point out is that currently, we use transports in the order they are combined in. That is nice IMO.

mxinden:
On the matter of smart dialing mechanism, corresponding work on the Go side: https://github.com/libp2p/go-libp2p/pull/2260/

PR #2248, opened 2021-09-26
core/: Concurrent connection attempts

mxinden:
### Main feature

Concurrently dial candidates within a single dial attempt.

For the concrete interface changes, see the changelog entries in `core/CHANGELOG.md` and `swarm/CHANGELOG.md`.

Main motivation for this feature is to increase success rate on hole punching (see https://github.com/libp2p/rust-libp2p/issues/1896#issuecomment-885894496 for details). Though, as a nice side effect, as one would expect, it does improve connection establishment time.

#### 95th percentile duration of a Kademlia GetClosestPeers request:

image

#### 50th percentile duration of a Kademlia GetClosestPeers request:

image

You can explore this data and more yourself at the URL below. On the left you have a [Kademlia Exporter](https://github.com/mxinden/kademlia-exporter) running with this feature, on the right you have a Kademlia Exporter running without.

https://kademlia-exporter.max-inden.de/d/Pfr0Fj6Mk/rust-libp2p?orgId=1&var-data_source=Prometheus&var-instance=kademlia-exporter-ipfs-concurrent:8080&var-instance=kademlia-exporter-ipfs:8080&from=now-3h&to=now&refresh=10s

~~For now, the concurrency factor is not configurable for the user. I think a value of 5 is a sane default. In the future I am fine exposing this value as a configuration parameter, though I would suggest we stick with this value as a first iteration.~~

See https://github.com/libp2p/rust-libp2p/pull/2248#issuecomment-941174468.

### Cleanups and fixes done along the way

- Merge `pool.rs` and `manager.rs`.

- Instead of manually implementing state machines in `task.rs` use `async/await`.

- Fix bug where `NetworkBehaviour::inject_connection_closed` is called without a previous `NetworkBehaviour::inject_connection_established` (see https://github.com/libp2p/rust-libp2p/issues/2242).

- Return handler to behaviour on incoming connection limit error. I missed this case in https://github.com/libp2p/rust-libp2p/issues/2242.

### Additional notes

I am sorry this ended up as such a large change. In my eyes it simplifies many of the somewhat dated structures in `libp2p-core`.


protocols/rendezvous/tests/rendezvous.rs
```diff
@@ -173,11 +173,19 @@ async fn discover_allows_for_dial_by_peer_id() {
...
+        _ => todo!(),
+    }
+
+    // TODO: Clean up
+    match bob.select_next_some().await {
+        SwarmEvent::Behaviour(rendezvous::client::Event::Discovered { registrations, .. }) => {
+            assert!(!registrations.is_empty());
+        }
+        _ => todo!(),
+    }
```
mxinden: If I am not mistaken the feature of this pull request - concurrent dialing - uncovers a race condition in this test, namely that `bob` tries to discover `alice` before `alice` registered with the rendezvous server.

@thomaseizinger suggestion on how to proceed in a way that is idiomatic with the remainder of the test harness? Maybe a flag to `assert_behaviour_events` that allows awaiting the two events either in parallel or in sequence?

thomaseizinger: I guess we should have waited for `alice` to be registered from the get-go before `bob` goes and tries to discover.

I think the most-idiomatic (and moving forward, most generally useful) extension would be to allow `assert_behaviour_events` to use only a single swarm as well. Then we could just invoke them in sequence without the boilerplate of having to match on the event that is returned from `select_next_some`:

```
assert_behaviour_events! {
    alice: rendezvous::client::Event::Registered { .. },
    || { }
};

assert_behaviour_events! {
    bob: rendezvous::client::Event::Discovered { registrations, .. },
    || {
        assert!(!registrations.is_empty());
    }
};
```

mxinden: :+1: Done via https://github.com/libp2p/rust-libp2p/pull/2248/commits/dc3c46db06ac55474c9b9eeb3e540fb37b76391c. Let me know what you think.

mxinden:
This pull request is ready for review. Feedback from **everyone** is very much appreciated, especially on the public interface changes. Also //CC @thomaseizinger as I think your feedback would be very valuable here.

@AgeManning could run a version of lighthouse with this patch to make sure none of the errors reported in https://github.com/libp2p/rust-libp2p/issues/2242 still occur? It has been running on kademlia-exporter.max-inden.de/ for a couple days now without problems.

AgeManning:
Can do. Will report back.

mxinden:
> For now, the concurrency factor is not configurable for the user. I think a value of 5 is a sane default. In the future I am fine exposing this value as a configuration parameter, though I would suggest we stick with this value as a first iteration.

Thinking about this some more, providing a configuration option is quite easy (via `PoolConfig`, previously `ManagerConfig`) and allows us to first disable concurrent dialing by default, enabling it by default in some later release.

Disabling concurrent dialing is useful in scenarios where low connection churn is preferred over low connection establishment latency.

I will add the option sometime this week.

thomaseizinger:
> Merge pool.rs and manager.rs.

OMG yes! I wanted to do that at some point as I thought that it is an unnecessary indirection when I looked at libp2p-core. Thank you for tackling that!

core/src/connection.rs
```diff
@@ -260,9 +315,9 @@ where
                 }
                 Poll::Ready(Ok(SubstreamEvent::AddressChange(address))) => {
                     self.handler.inject_address_change(&address);
-                    return Poll::Ready(Ok(Event::AddressChange(address)));
+                    return Poll::Ready(Some(Ok(Event::AddressChange(address))));
                 }
-                Poll::Ready(Err(err)) => return Poll::Ready(Err(ConnectionError::IO(err))),
+                Poll::Ready(Err(err)) => return Poll::Ready(Some(Err(ConnectionError::IO(err)))),
```
thomaseizinger: Is it really worth implementing `Stream` if we always return `Some`?

core/src/connection/pool.rs
```diff
@@ -93,27 +165,36 @@ pub enum PoolEvent<'a, THandler: IntoConnectionHandler, TTransErr> {
...
         peer: Option<PeerId>,
-        /// A reference to the pool that managed the connection.
-        pool: &'a mut Pool<THandler, TTransErr>,
+    },
+
+    /// An inbound connection attempt failed.
+    PendingInboundConnectionError {
+        /// The ID of the failed connection.
+        id: ConnectionId,
+        /// Stack of protocols used to send back data to the remote.
```
thomaseizinger: This is either stale or I don't understand it.

core/src/connection/pool.rs
```diff
@@ -379,67 +377,80 @@ impl<THandler: IntoConnectionHandler, TTransErr> Pool<THandler, TTransErr> {
...
     /// by the pool effective immediately.
     pub fn disconnect(&mut self, peer: &PeerId) {
-        if let Some(conns) = self.established.get(peer) {
-            for (&id, _endpoint) in conns.iter() {
-                if let Some(manager::Entry::Established(e)) = self.manager.entry(id) {
-                    e.start_close(None);
-                }
+        if let Some(conns) = self.established.get_mut(peer) {
+            // Detour via `EstablishedConnection` is not ideal, but at least only one code path in
+            // order to start closing a connection exists.
```
thomaseizinger: Couldn't the functionality for closing be defined on `EstablishedConnectionInfo`? Then this would be a simple `iter_mut`.

core/src/connection/pool.rs
```diff
@@ -471,206 +491,495 @@ impl<THandler: IntoConnectionHandler, TTransErr> Pool<THandler, TTransErr> {
...
+                {
+                    hash_map::Entry::Occupied(entry) => {
+                        return Poll::Ready(PoolEvent::ConnectionEvent {
+                            connection: EstablishedConnection { entry },
+                            event,
+                        })
+                    }
+                    hash_map::Entry::Vacant(_) => {
+                        unreachable!("Receive `Notify` event from established connection")
                     }
```
thomaseizinger: This pattern comes up a lot. Perhaps an extension trait on `Entry` could simplify this?

```rust
trait EntryExt {
	fn expect_occupied(self, msg: &str) -> OccupiedEntry;
}

impl EntryExt for hash_map::Entry {
	fn expect_occupied(self, msg: &str) -> OccupiedEntry {
		todo!("return occupied or panic with `msg`")
	}
}
```

Then you could write here:

```rust
                let entry = self
                    .established
                    .get_mut(&peer_id)
                    .expect("Receive `Notify` event for established peer.")
                    .entry(id)
		    .expect_occupied("Receive `Notify` event from established connection");

		return Poll::Ready(PoolEvent::ConnectionEvent {
                     connection: EstablishedConnection { entry },
                     event,
                });
```

core/src/connection/pool/concurrent_dial.rs
```diff
@@ -0,0 +1,159 @@
...
+            .boxed(),
+        });
+
+        let dials = FuturesUnordered::new();
+        while let Some(dial) = pending_dials.next() {
+            dials.push(dial);
+            if dials.len() == CONCURRENCY_FACTOR {
+                break;
+            }
+        }
```
thomaseizinger: Could we get rid of this if we instead initialize `FuturesUnordered` empty here and just make sure we always have at most `CONCURRENCY_FACTOR` futures in it at the top of the `loop` in `poll`?

core/src/connection/pool/task.rs
```diff
@@ -0,0 +1,257 @@
...
+                })
+                .await;
+        }
+    }
+}
+
+pub async fn new_for_pending_incoming_connection<TFut, TTrans>(
+    connection_id: ConnectionId,
+    future: TFut,
+    drop_receiver: oneshot::Receiver<Void>,
```
thomaseizinger, line 144: This is an interesting pattern :)

core/src/network.rs
```diff
@@ -466,185 +460,14 @@ where
...
-            &mut self.dialing,
-            opts,
-        )
-    }
 }
 
 /// Options for a dialing attempt (i.e. repeated connection attempt
 /// via a list of address) to a peer.
-struct DialingOpts<PeerId, THandler> {
+struct DialingOpts<PeerId, THandler, I: Iterator<Item = Multiaddr>> {
```
thomaseizinger: We are not referring to an associated type in the field definitions. Do we need these trait bounds?

thomaseizinger: Also, the `PeerId` type parameter seems a bit weird?

thomaseizinger (approved):
Great work!

Couple of comments but nothing major. Please note that it is not an in-depth review of the connection management logic. I am trusting that the tests are covering this functionality wise!

AgeManning:
So far this hasn't had any of the issues I was seeing in #2242

core/src/connection.rs
```diff
@@ -295,26 +352,17 @@ pub struct IncomingInfo<'a> {
 
 impl<'a> IncomingInfo<'a> {
     /// Builds the `ConnectedPoint` corresponding to the incoming connection.
```
elenaf9:
```suggestion
    /// Builds the `PendingPoint` corresponding to the incoming connection.
```

core/src/connection/error.rs
```diff
@@ -60,16 +54,31 @@ where
...
+    PendingConnectionError<TransportError<TTransErr>>;
+
 /// Errors that can occur in the context of a pending `Connection`.
 #[derive(Debug)]
-pub enum PendingConnectionError<TTransErr> {
-    /// An error occurred while negotiating the transport protocol(s).
-    Transport(TransportError<TTransErr>),
+pub enum PendingConnectionError<TransportError> {
+    /// An error occurred while negotiating the transport protocol(s) on a connection.
+    Transport(TransportError),
```
elenaf9: Consider naming the generic here also `TTransErr`:

1. it would be consistent with `PendingInboundConnectionError` / `PendingOutboundConnectionError` above, and with the general naming convention for generic types in rust-libp2p
2. it's easier to identify the type as generic when reading the code. `TransportError` gives the impression that this is an actual error type.

core/src/network/peer.rs
```diff
@@ -469,162 +462,86 @@ where
...
-pub(super) struct DialingState {
-    /// The ID and (remote) address of the current connection attempt.
-    pub(super) current: (ConnectionId, Multiaddr),
-    /// Multiaddresses to attempt if the current one fails.
-    pub(super) remaining: Vec<Multiaddr>,
-}
-
 /// A `DialingAttempt` is an ongoing outgoing connection attempt to
 /// a known / expected remote peer ID and a list of alternative addresses
 /// to connect to, if the current connection attempt fails.
```
elenaf9:
```suggestion
/// A `DialingAttempt` is an ongoing outgoing connection attempt to
/// a known / expected remote peer ID.
```

core/src/network/peer.rs
```diff
@@ -469,162 +462,86 @@ where
...
-            ConnectedPoint::Listener { .. } => unreachable!("by definition of a `DialingAttempt`."),
-        }
+        &self.peer_id
     }
 
     /// Aborts the dialing attempt.
     ///
     /// Aborting a dialing attempt involves aborting the current connection
     /// attempt and dropping any remaining addresses given to [`Peer::dial()`]
     /// that have not yet been tried.
```
elenaf9:
```suggestion
    /// Aborts the dialing attempt.
```

swarm/src/lib.rs
```diff
@@ -141,6 +141,10 @@ pub enum SwarmEvent<TBehaviourOutEvent, THandlerErr> {
         /// Number of established connections to this peer, including the one that has just been
         /// opened.
         num_established: NonZeroU32,
+        /// [`Some`] when the new connection is an outgoing connection.
+        /// Addresses are dialed in parallel. Contains the addresses and errors
+        /// of dial attempts that failed before the one successful dial.
+        outgoing: Option<Vec<(Multiaddr, TransportError<io::Error>)>>,
```
elenaf9: Whether the connection is outgoing or incoming can be determined via `endpoint`; imo using`outgoing` here is a bit confusing, `dial_failures` or something similar would describe it a bit better. 
Maybe also consider to wrap the `(Multiaddr, TransportError<_>)` tuple here in a new struct.

elenaf9 (approved):

core/src/connection.rs
mxinden: I guess it is a trait off between ergonomics (e.g. being able to call `next` on `Connection` via `StreamExt`) and strict typing.

Thinking about it some more, I don't think the extra ergonomics are worth it. Reverted with 0fea620e963c3d30d56473c5782962eedf6b26a4.

core/src/connection.rs
mxinden: Fixed in ecdb812b9e154745b94397aaa7addf50857f7997.

core/src/connection/error.rs
mxinden: :+1: Done via 2a6dc5485edd35b14af2f0f6aa90295f8287b37a.

core/src/connection/pool.rs
mxinden: If I am not mistaken "Stack of protocols" is a `Multiaddr`. Changed all occurrences in d25c95beb994909e3cd7df976e4ac2104a468eec.

core/src/connection/pool.rs
mxinden: Great idea. Done via e7beec40c53320832765dc8ae750df718227cb98.

core/src/connection/pool/concurrent_dial.rs
mxinden: We could get rid of the loop here, yes.

3aca8d0d16cf01101de1eacaa95f89caa653994d makes the concurrency factor configurable. Thus, if not done here in `fn new` but in `fn poll` we would need to keep track of the provided concurrency factor beyond `fn new`.

If I am not missing something, this is a matter of style right? I am fine either way. Let me know whether you still want the delayed filling of `dials` with 3aca8d0d16cf01101de1eacaa95f89caa653994d in mind.

core/src/connection/pool/task.rs
mxinden: Interesting as in "worth keeping" or interesting as in "not intuitive"?

I find it to be simple but not particularly intuitive.

I use the same in https://github.com/libp2p/rust-libp2p/blob/7718d1de387c63786cffdcbfcfdf1c4df40bc21a/protocols/relay/src/handler.rs#L110-L119.

core/src/network.rs
mxinden: Oh :see_no_evil:, not sure how that came about. Fixed in 34daf905677ce29b1678eb675e923864f27c67aa.

core/src/network/peer.rs
mxinden: Good catch. Thanks! Fixed in d09910627a8ed4b07cbd484833fb7ce91b4dbcb2.

core/src/network/peer.rs
mxinden: :pray: Removed with d5e06b6d2b2abfc690cb6ae8402db36b00348b73.

swarm/src/lib.rs
mxinden: > Whether the connection is outgoing or incoming can be determined via endpoint; imo usingoutgoing here is a bit confusing, dial_failures or something similar would describe it a bit better.

Good idea!

Renamed to `concurrent_dial_errors`. `[...]_errors` as we are referring to _errors_ instead of _failures_ in the events below.

Done in 7331d90d318c36df6dfd241c8fc6e0832ad383e2.

>  Maybe also consider to wrap the (Multiaddr, TransportError<_>) tuple here in a new struct.

I went without the additional type, favoring the simplicity of a tuple. That said this  is not a strong opinion, thus please comment in case you want me to change it.

core/src/connection/pool.rs
mxinden: Good idea. Done in 9fc587053364c588bc437868aab794d6c1403469.

mxinden:
> So far this hasn't had any of the issues I was seeing in #2242

Thanks @AgeManning for testing! That is great to hear.

---

Thanks @thomaseizinger and @elenaf9 for the reviews. Much appreciated! Good to have additional input on these changes.

I addressed all your comments. Let me know in case you would like to see any additional changes.

core/src/connection/pool/task.rs
thomaseizinger: Interesting in "in need to keep this in mind" 😁

core/src/connection/pool.rs
thomaseizinger: Okay, I figured it is something like that, just never saw it being referred to as such!

thomaseizinger (approved):
Thank you!

core/CHANGELOG.md
```diff
@@ -39,6 +39,14 @@
 - Report `ListenersEvent::Closed` when dropping a listener in `ListenersStream::remove_listener`,
   return `bool` instead of `Result<(), ()>` (see [PR 2261]).
 
+- Concurrently dial up address candidates within a single dial attempt (see [PR 2248]) configured
```
```suggestion
- Concurrently dial address candidates within a single dial attempt (see [PR 2248]) configured
```

PR #2256, opened 2021-09-30
protocols/rendezvous: export `Cookie`

wngr:


thomaseizinger (approved):
Thank you!

I really wish Rust would finally get a lint that detects usage of non-exported types in public APIs.

mxinden (approved):

PR #2258, opened 2021-10-01
Upgrade to Soketto 0.7.0

jsdw:


mxinden (approved):
Thanks @jsdw!

Do I understand correctly that `soketto` is not exposed on the API of `libp2p-websocket`? In other words, that this is not a breaking change?

Either way, would you mind including a changelog entry in `transports/websocket/CHANGELOG.md`?

Other than that, this looks good to me.

Do I understand correctly that this enables https://github.com/libp2p/rust-libp2p/pull/2085/?

jsdw:
I'm not super familiar with this codebase yet, so I reckon @dvdplm would be better equipped to answer those questions 🙂

(though offhand it looks like soketto isn't exposed publically at all, and so this does not introduce any breaking changes)

I'll add an entry to the changelog when Im next at a computer :)

jsdw:
Looking at the changelog, there is already an entry for "update dependencies" for the current unreleased version, which I think this change falls into, so I don't believe there's a need to update the changelog.

mxinden:
> Looking at the changelog, there is already an entry for "update dependencies" for the current unreleased version, which I think this change falls into, so I don't believe there's a need to update the changelog.

:+1: right, sorry, I missed that.

---

I don't have permissions to merge latest _master_ branch to your branch. Could you do so @jsdw?

jsdw:
> I don't have permissions to merge latest master branch to your branch. Could you do so @jsdw?

Done!

mxinden:
Mhhh, GitHub actions didn't show until now. Merged via https://github.com/libp2p/rust-libp2p/pull/2271. Thanks @jsdw!

PR #2260, opened 2021-10-02
.github/workflows/ci: Improve cache effectiveness

thomaseizinger:
To have caches operate at its maximum usefulness, we need to have a
1-to-1 mapping between build-comamnd and cache key. Otherwise rustc
has to rebuild certain artefacts because they were not in the cache.

By using matrices, we make github parallelise some of our jobs. This
has a double positive impact on CI runtime. Not only are our caches
more effective now, several jobs are now run in parallel.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -92,6 +92,8 @@ jobs:
 
     - uses: actions/checkout@v2.3.4
 
+    - uses: Swatinem/rust-cache@v1.3.0
+
     - name: Check rustdoc links
       run: RUSTDOCFLAGS="--deny broken_intra_doc_links" cargo doc --verbose --workspace --no-deps --document-private-items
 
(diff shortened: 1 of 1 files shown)


mxinden (approved):
Neat! Wasn't aware of the caching trick. Great to have a faster CI. Thanks.

thomaseizinger:
The speed up isn't nearly as much as I had hoped it to be. I'll need to test some more theories here.

I am surprised the test runs still take 15 minutes!

thomaseizinger:
Hang on, that was without any caching. I retriggered the jobs, lets see :)

thomaseizinger:
I guess 9 minutes instead of 18 is still pretty good. Most of that time is spend building the workspace crates (~6 minutes).

thomaseizinger:
@mxinden You will have to change the branch protection settings before we can merge this one.

mxinden:
@thomaseizinger the branch protection rule for `master` still contained the old CI pipeline names from before this pull request. I adjusted it to the new names, thus now you should be able to merge pull requests as well.

(I am a bit surprised that I was not able to choose "all steps are required", but I guess Github could never know whether there is one more to come that hasn't reported its status.)

PR #2261, opened 2021-10-02
Create ListenersEvent::Closed on Swarm::remove_listener

elenaf9:
Discussion #2113.
Create a `ListenersEvent::Closed` when a listener is removed via `Swarm::remove_listener`.
This makes it more consistent with `Swarm::listen_on`, and also informs the Swarm about the associated expired addresses.

I initially tried to realize it via a `Stream` implementation on the `connection::Listener`, so that it would just return `Poll::Ready(None)` if the listener was removed (see [this commit](https://github.com/elenaf9/rust-libp2p/commit/6acc63831ee996c466e2fd9e54fb21a254edf397)). But due to Listener being wrapped in a `Pin` it is a bit more complicated, whereas this Draft now is only a very minimal change.

_Note: https://github.com/libp2p/rust-libp2p/pull/2261/commits/0efe82c3c5081a9761268b95fa1ca822b7b7b1ad  introduces a minor API change so that `remove_listener` returns a boolean instead of `Result<(), ()>`. This 1. more consistent with other methods like `remove_external_address`, and 2. follows the clippy lint of [result_unit_err](https://rust-lang.github.io/rust-clippy/master/index.html#result_unit_err)._

### DIFF
--- a/core/src/connection/listeners.rs
+++ b/core/src/connection/listeners.rs
@@ -217,19 +217,23 @@ where
 
     /// Remove the listener matching the given `ListenerId`.
     ///
-    /// Return `Ok(())` if a listener with this ID was in the list.
-    pub fn remove_listener(&mut self, id: ListenerId) -> Result<(), ()> {
+    /// Returns `true` if there was a listener with this ID, `false`
+    /// otherwise.
+    pub fn remove_listener(&mut self, id: ListenerId) -> bool {
         if let Some(i) = self.listeners.iter().position(|l| l.id == id) {
-            let mut listener = self.listeners.remove(i).ok_or(())?;
+            let mut listener = self
+                .listeners
+                .remove(i)
+                .expect("Index can not be out of bounds.");
             let listener_project = listener.as_mut().project();
             self.pending_events.push_back(ListenersEvent::Closed {
                 listener_id: *listener_project.id,
                 addresses: listener_project.addresses.drain(..).collect(),
                 reason: Ok(()),
             });
-            Ok(())
+            true
         } else {
-            Err(())
+            false
         }
     }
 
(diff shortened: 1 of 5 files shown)


core/src/connection/listeners.rs
```diff
@@ -90,6 +90,8 @@ where
     listeners: VecDeque<Pin<Box<Listener<TTrans>>>>,
     /// The next listener ID to assign.
     next_id: ListenerId,
+    /// Pending listeners events to return from `poll`.
```
mxinden:
```suggestion
    /// Pending listeners events to return from [`ListenersStream::poll`].
```

mxinden (commented):
I am in favor of this change.

core/src/connection/listeners.rs
```diff
@@ -216,7 +220,13 @@ where
     /// Return `Ok(())` if a listener with this ID was in the list.
     pub fn remove_listener(&mut self, id: ListenerId) -> Result<(), ()> {
         if let Some(i) = self.listeners.iter().position(|l| l.id == id) {
-            self.listeners.remove(i);
+            let mut listener = self.listeners.remove(i).ok_or(())?;
+            let listener_project = listener.as_mut().project();
+            self.pending_events.push_back(ListenersEvent::Closed {
+                listener_id: *listener_project.id,
+                addresses: listener_project.addresses.drain(..).collect(),
```
thomaseizinger: Would `std::mem::replace` also work here? Slightly more readable IMO.

thomaseizinger (commented):
Thank you! One question inline.

mxinden (approved):

mxinden:
Giving others a chance to comment on the latest changes. Will merge early next week in case there are no further comments.

mxinden:
Thanks @elenaf9!

PR #2262, opened 2021-10-03
protocols/: Add basic AutoNAT implementation

elenaf9:
This PR drafts a behaviour protocol that implements basic AutoNAT. It is based on #1672 (minus #1667) from @dvc94ch .
The Autonat Protocol implements a Codec for the Request-Response protocol, and wraps it in a new Network Behaviour with some additional functionality. The basic idea is:

_See https://github.com/libp2p/rust-libp2p/pull/2262#issuecomment-993068771 description of the most recent implementation._
<s>
- A peer **A** maintains a list of its own addresses: 
  - For each address the current score is set: Higher score means higher reachability.
  - Addresses are either listening addresses reported from Listeners (initial Score: `0`), or external addresses observed by other peers (initial Score: `1`).
- When a new connection is established to a Peer **B**, **A** sends a `DialRequest` to **B**, with all addresses from the above list (ordered by Address Score, descending)
- **B** tries to dial the addresses:
  - It adds all addresses of **A** to the address-book (part of the request-response protocol) 
  - Swarm::dial(Peer **A**) sequentially tries to dial all addresses until either one attempt succeeds, or all have failed
  - If success: **B** sends back an Ok with the address that it successfully dial, if error: **B** sends back Error
- If **A** receives an `Ok(address)` it increases the address score of this address by `1`
- Contrary to #1672, this implementation of the AutoNAT Protocol has no `OutEvent`. Instead, when increasing a peer's score a `NetworkBehaviourAction::ReportObservedAddr` with the current score is reported to to the Swarm.</s>


protocols/autonat/Cargo.toml
```diff
@@ -0,0 +1,20 @@
+[package]
+name = "libp2p-autonat"
+version = "0.20.0"
+authors = ["David Craven <david@craven.ch>"]
```
mxinden: Feel free to add yourself here ;)

protocols/autonat/src/behaviour.rs
```diff
@@ -0,0 +1,376 @@
...
+    NetworkBehaviourAction, PollParameters,
+};
+use std::{
+    collections::{HashMap, HashSet, VecDeque},
+    iter,
+    task::{Context, Poll},
+    time::Duration,
+};
+
+type FiniteAddrScore = u32;
```
mxinden: Why not use the _new type pattern_, catching wrong usage of the type at compile time?

protocols/autonat/src/behaviour.rs
```diff
@@ -0,0 +1,376 @@
...
+            pending_requests: VecDeque::default(),
+        }
+    }
+
+    /// Add a new address to the address list that is send to remote peers in a dial-request.
+    pub fn add_local_address(&mut self, address: Multiaddr) {
+        if self.addresses.get(&address).is_none() {
+            self.addresses.insert(address, 1);
+        }
+    }
```
mxinden: Why not use `PollParameters::external_addresses` instead of this additional mechanism? That would then as well the external addresses of the local node discovered through protocols like `libp2p-identify`.

You would need to call `PollParameters::external_addresses` continuously to make sure to not miss an update. Just a thought: We could as well replace the current mechanism (`PollParameters::external_addresses`) through a push-style mechanism, e.g. introducing `NetworkBehaviour::inject_external_address`, e.g. triggered by a `NetworkBehaviourAction::ReportObservedAddr` emitted by another behaviour.

Also, instead of referring to these addresses as _local addresses_, I would suggest referring to them as _external addresses_ to be consistent with the remainder of the codebase.

protocols/autonat/src/behaviour.rs
```diff
@@ -0,0 +1,376 @@
...
+        loop {
+            if let Some(peer_id) = self.pending_requests.pop_front() {
+                let mut scores: Vec<_> = self.addresses.clone().into_iter().collect();
+                // Sort so that the address with the highest score will be dialed first by the remote.
+                scores.sort_by(|(_, score_a), (_, score_b)| score_b.cmp(score_a));
+                let addrs = scores.into_iter().map(|(a, _)| a).collect();
+                self.inner
+                    .send_request(&peer_id, DialRequest { peer_id, addrs });
+            }
+            match self.inner.poll(cx, params) {
```
mxinden: Would any of the helper methods on `NetworkBehaviourAction`, e.g. `map_in` or `map_out` be able to make the match statement below simpler? Just an idea, I haven't thought this through.

protocols/autonat/src/behaviour.rs
```diff
@@ -0,0 +1,376 @@
...
+        conn: &ConnectionId,
+        endpoint: &ConnectedPoint,
+    ) {
+        self.inner
+            .inject_connection_established(peer, conn, endpoint);
+
+        // Initiate a new dial request if there is none pending.
+        if !self.ongoing_outbound.contains(peer) {
+            self.pending_requests.push_back(*peer);
+        }
```
mxinden: Just as a note, though this likely requires a lot more design work: We will need AutoNAT to do continuous probing, among other things, to keep mappings on NAT devices active.

mxinden (commented):
Wow, that was quick! Thanks.

As an aside, great to see `libp2p-request-response` being useful here.

Can you reference the new crate in `src/lib.rs`? Similar to:

https://github.com/libp2p/rust-libp2p/blob/c7abb6f70c97cdc75aed021870f5902ec2cb268e/src/lib.rs#L58-L61

Listing a couple of things needed beyond a basic version. Unfortunately the current draft of the specification is still _very_ basic, thus many of these likely need design work as well.

- Expose whether the local node is publicly reachable or not. (e.g. needed for `libp2p-kad` to decide to run in client or server mode https://github.com/libp2p/rust-libp2p/pull/2184).
- Implement various rate limiting to prevent attacks.
- Continuously trigger probes for local external addresses.
- ...

protocols/autonat/src/behaviour.rs
elenaf9: Okay, so I just noticed that `(Swarm)PollParameters::external_addresses`/ `registry::Addresses` already implements a majority of the logic that I was trying to achieve with my `local_address` map. I so far did not know about that, because the return type `AddressIntoIter` is not publicly exposed.  I assume that this is a mistake and that `AddressIntoIter` should be public; I can either change this as part of this PR, or do a separate one?  

But with this additional info, I would change quite a bit of my code, and I think I can remove the `addresses` HashMap completely, and just rely on the list provided in the `PollParameters::external_addresses` and send this list to the remote in a `DialRequest`.

> Just a thought: We could as well replace the current mechanism (PollParameters::external_addresses) through a push-style mechanism, e.g. introducing NetworkBehaviour::inject_external_address, e.g. triggered by a NetworkBehaviourAction::ReportObservedAddr emitted by another behaviour.

While I am in general more in favor of having such a push-style, I am not sure whether it makes sense in this case. Triggering `inject_external_address` would only be useful for the behaviour, if it independently maintains a map / list of its external addresses. I think in most cases this would just duplicate the logic that is already provided by `registry::Addresses`. Do you have specific use-cases in mind (maybe also from the other protocols) where this would make sense?
I think this is a general discussion, also kinda related to the discussion about the `AddressBook` in https://github.com/libp2p/rust-libp2p/discussions/2118: what should the Swarm implement in a central way, and what should be done independently by a Network Behaviour. I am a bit indecisive on this. 
But either way I think we need more documentation on the responsibilities/ concerns of the Swarm vs the ones of a Network Behaviour protocol.

src/lib.rs
```diff
@@ -40,6 +40,10 @@ pub use libp2p_core::multihash;
 #[doc(inline)]
 pub use multiaddr;
 
+#[cfg(feature = "autonat")]
+#[cfg_attr(docsrs, doc(cfg(feature = "autonat")))]
+#[doc(inline)]
+pub use libp2p_autonat as autonat;
```
elenaf9, line 46: > Can you reference the new crate in src/lib.rs

It is already referenced 😅

elenaf9:
> Expose whether the local node is publicly reachable or not. (e.g. needed for libp2p-kad to decide to run in client or server mode #2184).

Ok so I would propose some additional logic:
The behaviour has a list of public nodes, that have to be provided by the outside (similar how kademlia requires such a list of initial nodes; nodes can be added / removed from the list at any time).
Add a method `determine_nat_status()` that determines whether the current node is publicly available or not:
The method takes as input 
   1. The above list of publicly available nodes
   2. The number of requests that have to be successful in order for the address to count as "publicly available" (can be set in a config)
  
It will then send a `DialRequest` to each node, and collect the results to determine if one of the `external_addresses` can be reached by the required number of peers. It will then create a AutoNAT behaviour `OutEvent`, with either `NatStatus::Public(public_address)`, or `NatStatus::Private`, depending on whether one of the address was reachable by the required number of peers.
Depending on this result, the AutoNAT behaviour will report the address score, and also set a local status.
In a configurable frequency it will retriggger the `determine_nat_status` method to verify that its status did not change (or change its status in cases that it did change). This will:
1. keep the status up to date
2. keep mapping on NAT devices active

If we are in `NatStatus::Private`, we could also retrigger `determine_nat_status` directly if a new external address is observed. (Which would actually be an argument towards what @mxinden suggested above about changing `ReportObservedAddr` to a push-like behaviour).
I am not sure whether we would then even need to send a `DialRequest` to each peer that we establish a connection.

---

I think this is also a bit more aligned with what is described in the specs PR. Wdyt @mxinden?

mxinden:
Thanks for the follow-up @elenaf9!

>  The behaviour has a list of public nodes, that have to be provided by the outside (similar how kademlia requires such a list of initial nodes; nodes can be added / removed from the list at any time).

Is that a strict requirement? It would be great for `libp2p-autonat` to not require any configuration nor depend on any preconfigured nodes.

> The number of requests that have to be successful in order for the address to count as "publicly available" (can be set in a config)

Sounds good to me. We can likely learn a bunch from the Golang implementation, which is set to assume it is public on [3 successes](https://github.com/libp2p/go-libp2p-autonat/blob/1247ac6d9fa798e7032127878a6f3d0b9eb487c6/autonat.go#L283-L345).

> It will then create a AutoNAT behaviour OutEvent, with either NatStatus::Public(public_address), or NatStatus::Private, depending on whether one of the address was reachable by the required number of peers.

Reporting through `OutEvent` sounds good to me.

> In a configurable frequency it will retriggger the determine_nat_status method to verify that its status did not change (or change its status in cases that it did change). 

Sounds good to me. We can likely use the [Golang defaults at the beginning](https://github.com/libp2p/go-libp2p-autonat/blob/1247ac6d9fa798e7032127878a6f3d0b9eb487c6/options.go#L38-L42), right?

> If we are in NatStatus::Private, we could also retrigger determine_nat_status directly if a new external address is observed. 

:+1: good idea.

>  I am not sure whether we would then even need to send a DialRequest to each peer that we establish a connection.

How about keeping track of all the peers we are connected to. On an interval, we chose one of them at random and ask them to dial us. Results of those dials then influence whether we assume we are publicly reachable or not.

mxinden:
Crossreferencing https://github.com/libp2p/specs/pull/369 here which we will very much need in an initial version. Don't want libp2p to become the next amplification attack network :fearful:

protocols/autonat/src/behaviour.rs
```diff
@@ -0,0 +1,626 @@
...
+            // TODO: Add more filters
+            addr.iter()
+                .all(|proto| match proto {
+                    Protocol::P2pCircuit => false,
+                    Protocol::P2p(hash) => hash == peer.into(),
+                    _ => true,
+                })
+                .then(|| addr)
+        })
+        .collect()
```
elenaf9: > Crossreferencing libp2p/specs#369 here which we will very much need in an initial version. Don't want libp2p to become the next amplification attack network fearful

@mxinden do we already have in rust-libp2p any logic for filtering addresses for public reachable /valid ones/ replacing ip-addresses with the observed one / ... etc.? 
 I am not really sure how to implement this properly, so I'd be grateful for some help here.

elenaf9 (commented):
Sorry for the delay here.
I have now refactored a majority of the behaviour logic according to the above comments. It is now also more aligned with the go-autonat implementation, which btw helped a lot thanks! There is still quite a lot missing, especially regarding config parameters and the calculation of the retry-nat-status frequency, and tests of course.
I hope I'll find more time towards this weekend / next week to invest into this.

protocols/autonat/src/behaviour.rs
mxinden: We have this scattered across the code base, always specific to the concrete use-case. In case you find any good abstractions that would be worth extracting, we can move them into `rust-multiaddr`.

> any logic for filtering addresses for public reachable

I wrote this in the past: https://github.com/mxinden/kademlia-exporter/blob/8953e0b604a3597fec362e0816706a6f7ec83cb6/src/exporter/client/global_only.rs#L33-L54

> replacing ip-addresses with the observed one

This could be done with an `iter_mut` and a `Iterator::find` no?

We should as well drop all addresses with `Protocol::P2pCircuit`.

>  I am not really sure how to implement this properly, so I'd be grateful for some help here.

Let me know if the above is of some help. Also, I am happy to help out and push a commit to your branch adding the filtering logic.

elenaf9: Thanks @mxinden. I have added some logic in https://github.com/libp2p/rust-libp2p/pull/2262/commits/48ad51dd73489e75ef67bea63523088c42914c8e.
Unfortunately, `std::net::Ipv4Addr::is_global` and `std::net::Ipv6Addr::is_global` are behind the unstable `ip` feature. For `Ipv4Addr` I have copied most of the logic from `is_global` and added a note, because from dangling PRs it seemed like that logic would be ready for stable (see https://github.com/rust-lang/rust/pull/85585), but for `Ipv6Addr` it seemed a bit more complicated. I could also just copy the logic from `Ipv6Addr::is_global`, but this relies on some unstable methods itself that I would then have to copy as well, and I am not sure whether they are all already fixed.
What do you think about this (and the rest of the logic in `filter_valid_addrs`)?

> Also, I am happy to help out and push a commit to your branch adding the filtering logic.

If you have time and some specific filters/ adjustments in mind that are not implemented in `filter_valid_addrs` yet, I'd be happy to get some help.

protocols/autonat/src/behaviour.rs
```diff
@@ -0,0 +1,638 @@
...
+            required_success,
+            dismissed: 0,
+            errors: Vec::new(),
+            min_valid,
+        }
+    }
+
+    fn do_probe(&mut self, probe: Probe, mut addr_records: Vec<AddressRecord>) {
+        // Sort so that the address with the highest score will be dialed first by the remote.
+        addr_records.sort_by(|record_a, record_b| record_b.score.cmp(&record_a.score));
```
mxinden: Is the sort needed here?

https://github.com/libp2p/rust-libp2p/blob/220f84a97f931ba5cac752a9c3eedd4efe8bdb8a/swarm/src/lib.rs#L485-L488

protocols/autonat/src/behaviour.rs
```diff
@@ -0,0 +1,638 @@
...
+                    }
+                },
+                Poll::Ready(action) => return Poll::Ready(action.map_out(|_| unreachable!())),
+                Poll::Pending => return Poll::Pending,
+            }
+            if let Some(probe) = self.pending_probe.take() {
+                let records: Vec<_> = params.external_addresses().collect();
+                self.do_probe(probe, records)
+            } else if let Some(auto_retry) = self.auto_retry.as_ref() {
+                if Instant::now().duration_since(self.last_probe) > auto_retry.interval {
```
mxinden: Do I understand correctly that the underlying task is not woken up once `auto_retry.interval` fires? How about using `futures-timer` here, to ensure the timer triggers the underlying logic even if there is no other event calling `poll`?

protocols/autonat/src/behaviour.rs
```diff
@@ -0,0 +1,638 @@
...
+                    let records: Vec<_> = params.external_addresses().collect();
+                    self.do_probe(probe, records);
+                }
+            }
+        }
+    }
+}
+
+// Filter demanded dial addresses for validity, to prevent abuse.
+fn filter_valid_addrs(
```
mxinden: :rocket: 

Could you add a unit test to ensure all IP addresses in `demanded` are replaced with the IP address in `observed_remote_at`?

src/lib.rs
mxinden: Oh, my bad. Thanks for the pointer.

protocols/autonat/src/behaviour.rs
```diff
@@ -0,0 +1,638 @@
...
+
+    fn new_handler(&mut self) -> Self::ProtocolsHandler {
+        self.inner.new_handler()
+    }
+
+    fn addresses_of_peer(&mut self, peer: &PeerId) -> Vec<Multiaddr> {
+        if let Some(addrs) = self.ongoing_inbound.get(peer).map(|(a, _)| a.clone()) {
+            addrs
+        } else {
+            self.inner.addresses_of_peer(peer)
```
mxinden: If I am not mistaken `RequestResponse::add_address` is never called. With that in mind, is the delegation to `self.inner` needed here?

protocols/autonat/src/behaviour.rs
```diff
@@ -0,0 +1,638 @@
...
+                            request,
+                            channel,
+                        } => match self.handle_request(peer, request) {
+                            Some(addrs) => {
+                                self.ongoing_inbound.insert(peer, (addrs, channel));
+                                return Poll::Ready(NetworkBehaviourAction::DialPeer {
+                                    peer_id: peer,
+                                    handler: self.inner.new_handler(),
+                                    condition: DialPeerCondition::Always,
+                                });
```
mxinden: https://github.com/libp2p/rust-libp2p/pull/2317 just merged. I guess the tracking of `addrs` in `ongoing_inbound` is no longer needed?

mxinden (commented):
Thanks for the clean and comprehensible code!

From a features perspective, I think this is sufficient for a first version.

I haven't had a chance to test this out, which I would like to do before merging and releasing it.

Have you had a chance to test this (a) in the wild and (b) with the Golang implementation?

protocols/autonat/src/behaviour.rs
mxinden: Off the top of my head, not filtering out non-global IPv6 addresses is fine.

> If you have time and some specific filters/ adjustments in mind that are not implemented in filter_valid_addrs yet, I'd be happy to get some help.

I think the current implementation of `filter_valid_addrs` is sufficient.

canewsin:
@elenaf9 could you provide some basic example to use it ?

elenaf9:
@canewsin I will definitely add some test & code-docs that will exemplify how this can be used. But I currently don't have the capacity to add a full example, though I agree that it would be great to have one that combines AutoNAT with e.g. Kademlia and the Relay protocol. But imo that would be an issue on its own.


Btw if you are generally interested in AutoNAT, it would also be great to have your review on it. Since @mxinden stated that the features are sufficient for a first version, the only thing left for this for this PR is to resolve the review and add docs & tests. If you feel like there's anything else missing, feel free to comment on it.

protocols/autonat/src/behaviour.rs
elenaf9: With https://github.com/libp2p/rust-libp2p/pull/2262/commits/310b086423e243696b23b0afc3751d2242999b8c it now specifies a fixed set of addresses. But the addresses still need to be tracked in `ongoing_inbound` to verify on `inject_connection_established` that the address we connected to was actually one that we were supposed to dial.

protocols/autonat/src/behaviour.rs
elenaf9: `RequestResponse::addresses_of_peer` also includes the addresses of the currently connected peers, which we need if we do a probe with `extend_with_connected = true`.
Apart from that, with https://github.com/libp2p/rust-libp2p/pull/2262/commits/0a612c07cd1d7f985f4e20b44a8817bcb57defac now also the addresses of the "static servers" are added via the inner request-response protocol.

protocols/autonat/src/behaviour.rs
elenaf9: Should be solved now with https://github.com/libp2p/rust-libp2p/pull/2262/commits/99418f85f829a6a00d93c8a530d23e524c233ab0. 
Could you please double check this logic, as I am not too confident with `futures-timer`, and wakers in general?

protocols/autonat/src/behaviour.rs
```diff
@@ -0,0 +1,713 @@
...
+    collections::{HashMap, HashSet, VecDeque},
+    iter,
+    task::{Context, Poll},
+    time::Duration,
+};
+
+#[derive(Debug, Clone)]
+/// Config for the [`Behaviour`].
+pub struct Config {
+    // Timeout for requests.
```
mxinden:
```suggestion
    /// Timeout for requests.
```

Why not use doc comments here and below?

protocols/autonat/src/behaviour.rs
```diff
@@ -0,0 +1,713 @@
...
+            Some(_) => ProtocolSupport::Full,
+            None => ProtocolSupport::Outbound,
+        };
+        let protocols = iter::once((AutoNatProtocol, proto_support));
+        let mut cfg = RequestResponseConfig::default();
+        cfg.set_request_timeout(config.timeout);
+        let inner = RequestResponse::new(AutoNatCodec, protocols, cfg);
+        let auto_retry = config
+            .auto_retry
+            .map(|a| (a, Delay::new(Duration::new(0, 0))));
```
mxinden:
```suggestion
            .map(|a| (a, Delay::new(Duration::ZERO)));
```
Nit pick.

protocols/autonat/src/behaviour.rs
```diff
@@ -0,0 +1,713 @@
...
+        Some(addrs)
+    }
+
+    fn handle_response(
+        &mut self,
+        sender: PeerId,
+        response: Option<DialResponse>,
+    ) -> Option<NatStatus> {
+        let mut probe = self.ongoing_outbound.take()?;
+        if !probe.pending_servers.contains(&sender) {
```
mxinden:
```suggestion
        // If response is from a peer that is not part of the ongoing probe, ignore it.
        if !probe.pending_servers.contains(&sender) {
```

This took me a bit to understand. How about adding the comment above?

protocols/autonat/src/behaviour.rs
mxinden: Usage of `futures-timer::Delay` looks good to me. Thank you for the patch @elenaf9.

protocols/autonat/src/behaviour.rs
```diff
@@ -0,0 +1,713 @@
...
+};
+use std::{
+    collections::{HashMap, HashSet, VecDeque},
+    iter,
+    task::{Context, Poll},
+    time::Duration,
+};
+
+#[derive(Debug, Clone)]
+/// Config for the [`Behaviour`].
```
thomaseizinger: Nit: For all other derives, this is flipped.

```suggestion
/// Config for the [`Behaviour`].
#[derive(Debug, Clone)]
```

protocols/autonat/src/behaviour.rs
```diff
@@ -0,0 +1,713 @@
...
+            return None;
+        }
+        // Check that there is no ongoing dial to the remote.
+        if self.ongoing_inbound.contains_key(&sender) {
+            return None;
+        }
+        // Check if max simultaneous autonat dial-requests are reached.
+        if self.ongoing_inbound.len() >= config.max_ongoing {
+            return None;
+        }
```
thomaseizinger: These might deserve a `log::trace` or `log::debug` to not be completely silent about what is happening.

protocols/autonat/src/behaviour.rs
```diff
@@ -0,0 +1,713 @@
...
+                return Poll::Ready(NetworkBehaviourAction::GenerateEvent(event));
+            }
+            match self.inner.poll(cx, params) {
+                Poll::Ready(NetworkBehaviourAction::GenerateEvent(event)) => match event {
+                    RequestResponseEvent::Message { peer, message } => match message {
+                        RequestResponseMessage::Request {
+                            request_id: _,
+                            request,
+                            channel,
+                        } => match self.handle_request(peer, request) {
```
thomaseizinger: These _could_ be collapsed:

```suggestion
                Poll::Ready(NetworkBehaviourAction::GenerateEvent(RequestResponseEvent::Message { peer, message: RequestResponseMessage::Request {
                            request_id: _,
                            request,
                            channel,
                        } })) => match self.handle_request(peer, request) {
```

I don't know what rustfmt will make out of this but it should reduce the indentation level for all the following code a bit due to the removed nested `match`s.

protocols/autonat/src/protocol.rs
```diff
@@ -0,0 +1,310 @@
...
+        &mut self,
+        _: &AutoNatProtocol,
+        io: &mut T,
+    ) -> io::Result<Self::Request>
+    where
+        T: AsyncRead + Send + Unpin,
+    {
+        let bytes = upgrade::read_length_prefixed(io, 1024)
+            .await
+            .map_err(|e| io::Error::new(io::ErrorKind::InvalidData, e))?;
```
thomaseizinger: I don't think these should need `.map_err` anymore after https://github.com/libp2p/rust-libp2p/pull/2111.

protocols/autonat/src/protocol.rs
```diff
@@ -0,0 +1,310 @@
...
+        let request = DialRequest::from_bytes(&bytes)
+            .map_err(|e| io::Error::new(io::ErrorKind::InvalidData, e))?;
```
thomaseizinger: `from_bytes` returns `io::Error`. Why `.map_err` here?

thomaseizinger (commented):
Couple of ideas :)

I would like to see an integration test before we merge this.

elenaf9:
Thank you for the review @thomaseizinger & @mxinden.
> Have you had a chance to test this (a) in the wild and (b) with the Golang implementation?

Unfortunately I did not have the time for that yet. But I will do some testing and report back on that before I mark this PR as ready-for-review. Not sure when that will be though. (And I also still have it on my list to write down what is missing in this implementation compared to the Golang one).

> I would like to see an integration test before we merge this.

Yes I will definitely add some proper integration tests. I was originally planning to build upon the tests that were drafted in #1672 using `netsim_embed`. But the library version used there is much older and quite different from the most recent version and I am not sure whether it's even a suitable library for my use-case. Any other recommendation for rust libraries that allow the simulation of a network with different NAT configurations?

dvc94ch:
> Any other recommendation for rust libraries that allow the simulation of a network with different NAT configurations?

Not sure there are too many of those.

> But the library version used there is much older and quite different from the most recent version and I am not sure whether it's even a suitable library for my use-case.

In case you're wondering why it requires so much boilerplate these days, it's the only way to achieve proper namespace isolation. initially I tried hacking around it by isolating threads in different network namespaces, but had some trouble with async-io as networking happens on a single background thread, so the network namespace will be the same as the first thread that starts it. netsim-embed is used by ipfs-embed and actyx.

protocols/autonat/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+categories = ["network-programming", "asynchronous"]
+
+[build-dependencies]
+prost-build = "0.6"
+
+[dependencies]
+async-trait = "0.1"
+futures = "0.3"
+futures-timer = "3.0.2"
+libp2p-core = { version = "0.30.1", path = "../../core" }
```
mxinden:
```suggestion
libp2p-core = { version = "0.30.1", path = "../../core", default-features = false }
```

protocols/autonat/Cargo.toml
```diff
@@ -0,0 +1,21 @@
+[package]
+name = "libp2p-autonat"
+version = "0.20.0"
```
mxinden, line 5:
```suggestion
edition = "2021"
rust-version = "1.56.1"
version = "0.20.0"
```

protocols/autonat/src/behaviour.rs
```diff
@@ -0,0 +1,798 @@
...
+    /// Timeout for requests.
+    pub timeout: Duration,
+
+    /// Config if the peer should frequently re-determine its status.
+    pub auto_retry: Option<AutoRetry>,
+
+    /// Config if the current peer also serves as server for other peers.
+    /// In case of `None`, the local peer will never do dial-attempts for other peers.
+    pub server: Option<ServerConfig>,
+}
```
mxinden:
```suggestion
}

impl Default for Config {
    fn default() -> Self {
        Self {
            timeout: Duration::from_secs(60),
            auto_retry: None,
            server: None,
        }
    }
}
```

How about implementing `Default` for `Config`?

protocols/autonat/src/lib.rs
```diff
@@ -0,0 +1,30 @@
...
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+// FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+// DEALINGS IN THE SOFTWARE.
+
+//! Implementation of the AutoNAT protocol.
+
+pub use self::behaviour::{Behaviour, Config};
+
+mod behaviour;
```
mxinden:
```suggestion
pub mod behaviour;
```

Instead of exposing each type, I would suggest exposing the crate internal namespace structure, defining a types visibility at its definition. What do you think?

Also, in case I am not missing something, types like `AutoRetry` are unreachable as an external user.

protocols/autonat/src/behaviour.rs
```diff
@@ -0,0 +1,798 @@
...
+
+impl Probe {
+    // Evaluate current state to whether we already have enough results to derive the NAT status.
+    fn evaluate(self) -> Result<NatStatus, Self> {
+        // Find highest score of successful dials for an address.
+        let (address, highest) = self
+            .addresses
+            .iter()
+            .max_by(|(_, a), (_, b)| a.cmp(b))
+            .expect("At least one external address was added.");
```
mxinden: What if the user depends on discovering their external addresses at runtime? Would it not be better to handle this in some manner, e.g. by failing the probe?

protocols/autonat/src/protocol.rs
```diff
@@ -0,0 +1,304 @@
...
+                status_text: None,
+                addr: Some(addr),
+            }) => {
+                let addr = Multiaddr::try_from(addr)
+                    .map_err(|err| io::Error::new(io::ErrorKind::InvalidData, err))?;
+                Self::Ok(addr)
+            }
+            Some(structs_proto::message::DialResponse {
+                status: Some(status),
+                status_text: _,
```
mxinden: I think `status_text` is super helpful for debugging. How about either including it in `ResponseError` or at least logging it on `log::debug`?

protocols/autonat/src/behaviour.rs
```diff
@@ -0,0 +1,798 @@
...
+        self.next_probe_id.0 += 1;
+        probe_id
+    }
+
+    fn do_probe(&mut self, mut probe: Probe) {
+        for (peer_id, id) in probe.pending_servers.iter_mut() {
+            let request_id = self.inner.send_request(
+                peer_id,
+                DialRequest {
+                    peer_id: *peer_id,
```
mxinden: This should be the local peer ID and not the peer ID of some probing server, right?

Go servers return a `BadRequest` `peer id mismatch` to me.

https://github.com/libp2p/go-libp2p-autonat/blob/8ff057fa9a70c29be0a8eaf29c98a37d5e732767/svc.go#L97-L99

Changing this to the local peer ID solves the problem.

mxinden (commented):
Tested this out against an IPFS bootstrap node. Great to see it in action.

To not duplicate work, in case you would like to include an example client, see https://github.com/elenaf9/rust-libp2p/pull/1.

protocols/autonat/src/lib.rs
elenaf9: > Instead of exposing each type, I would suggest exposing the crate internal namespace structure, defining a types visibility at its definition. What do you think?

I guess its more a thing of personal preference. But since there are only few types, I prefer to have them on one layer instead of exposing the mod hierarchy inside of autonat; apart from that almost all types are in `behaviour` anyway. And other protocols mostly also directly expose the types, so it's more consistent with those as well.

elenaf9 (commented):
While writing the test I noticed I still had quite a few bugs; the ones you already mentioned @mxinden and also some edge cases I did not consider. Sorry if that costed you time, but they should be resolved now. Could you test the current version again if you have time?

protocols/autonat/src/behaviour.rs
elenaf9: Changed this now to instead abort the probe and report `Reachability::Unknown`.

protocols/autonat/src/behaviour.rs
elenaf9: > This should be the local peer ID and not the peer ID of some probing server, right?

Yes.
Afaik there is no way from within a `NetworkBehaviour` to get the own peer-id, so it now has to be entered in `Behaviour::new`. Not the best solution because the user has to enter it twice when creating the Swarm (since the swarm itself also needs it), but I don't think there's any other way to solve this.
Kinda related to what you wrote in https://github.com/libp2p/specs/pull/362#discussion_r739660755; do we even need the peer-id here at all? But I guess it would break go-autonat to remove it from the structure, so it think the current solution is fine.

protocols/autonat/src/lib.rs
mxinden: I am fine either way. The only thing I would like to prevent (which is already addressed in your code) is name repetition.

See https://github.com/libp2p/rust-libp2p/discussions/2174 and https://rust-lang.github.io/rust-clippy/master/index.html#module_name_repetitions

Also //CC @thomaseizinger as you were leading the discussion on naming conventions within rust-libp2p.

thomaseizinger: My personal preference also goes for private modules and explicit re-exporting of types.

Public modules far too easily suffer from accidential API breaking changes, plus anything that is reachable from `lib.rs` is automatically excluded from rustc's dead-code analysis.

Naming looks good here.

protocols/autonat/src/behaviour.rs
elenaf9: I just noticed: do we even need to filter non-global addresses? As the IP in "demanded" addresses is replaced with the observed one anyway, it should not matter whether the observed address is public or not right? I saw that the go implementation rejects requests where the observed IP is private, and filters all "demanded" addresses where the IP of this address is private. Does this really make sense? 
Following arguments would be against it:
- If the IP is private: 
  Maybe there are cases where the peers are all within a private network, but there are firewalls established and a peer would like to use autonat to figure out if it can be dialed by peers in the same private network?
- If the demanded address is private: 
  Maybe they still want to test the port because they have configured port-forwarding?

@mxinden @marten-seemann

protocols/autonat/src/behaviour.rs
elenaf9: Added in ebd9d2a4d8b6f606a8ee5f3192a03e153743b9d0.

protocols/autonat/src/behaviour.rs
elenaf9: Added in b70428e28c4d2a2315598deeb7ebd43838473ac8.

protocols/autonat/src/protocol.rs
elenaf9: Added with 2f23e4b871421dd4f40241e7b4df2f27862c7c77 and b70428e28c4d2a2315598deeb7ebd43838473ac8.
Since throughout the rest of the library afaik it's not common to return strings as error message, I just logged it but not included it it in the `NatStatus`. But I am rather neutral on this point, so I could also add it to `ResponseError` if you prefer that.

elenaf9:
Apparently there are some flaky tests; from what I am have seen they are not related to the changes in this PR?
- ebd9d2a4d8b6f606a8ee5f3192a03e153743b9d0 seems to fail because of #1725
- 44f684afd536e18e39377ba961c0fedb940fe816 was cancelled because `libp2p_swarm::tests::test_behaviour_disconnect_all` did not return

protocols/autonat/src/behaviour.rs
mxinden: >  Afaik there is no way from within a NetworkBehaviour to get the own peer-id, so it now has to be entered in Behaviour::new. Not the best solution because the user has to enter it twice when creating the Swarm (since the swarm itself also needs it), but I don't think there's any other way to solve this.

Agreed that having to pass the peer ID twice, once to `Swarm`, once to `AutoNAT` is silly. I have to do the same in the circuit relay v2 implementation https://github.com/libp2p/rust-libp2p/pull/2059. Unfortunately there is no other way today. Also I fail to come up with a good mechanism providing the local peer ID from the `Swarm`. Happy for suggestions.

>  Kinda related to what you wrote in libp2p/specs#362 (comment); do we even need the peer-id here at all? But I guess it would break go-autonat to remove it from the structure, so it think the current solution is fine.

Yes, I don't understand why we send it. I hope one day we will fix this in AutoNAT v2.

protocols/autonat/src/lib.rs
mxinden: > plus anything that is reachable from lib.rs is automatically excluded from rustc's dead-code analysis.

Oh, good point. Thanks!

mxinden:
> Sorry if that costed you time, but they should be resolved now.

Not at all. Thanks for fixing them quickly.

> Could you test the current version again if you have time?

Tested as a client on my laptop (not publicly reachable) and a cloud machine (publicly reachable) against a go-libp2p AutoNAT server. All working as expected. Haven't gotten around to testing your implementation as a server.

jnicholls:
This is exciting. Thanks for the hard work all!

elenaf9:
## Update
I did some refactoring of how probes  are implemented, it now mirrors more or less completely the go-implementation, namely:
- a probe only sends a dial-back request to a single server
- instead of overwriting the status according to a probe's result (and ignoring the old state), it now increases / decreases the confidence in a state, depending on whether the assumed status was confirmed or a different one is reported
- when confidence is 0 the status is flipped to the reported one
- only if the status is flipped, it is reported as an `OutEvent`
- `OutEvent` is now just the `NatStatus`

## go-libp2p autonat vs our implementation
Compared to go-autonat (or at least what I understand of it) our implementation differs as follows:

1. go-autonat skips dial peer if one of the addresses matches an address of the local peer, and skip addresses that are private:

See https://github.com/libp2p/rust-libp2p/pull/2262#discussion_r757955839 on why I would propose **not** to skip private addresses. I would also not skip addresses that we believe are ours, because it is **theoretically** possible that both, the remote peer and ourself, are in different networks, bu have the same IP in our perspective private network. This may be rather unlikely, but in this case following the go dial policy would mean we reject all requests from that peers. And following https://github.com/libp2p/specs/pull/369 replaces the demanded Ip with the observed one anyway, so I don't see a harm in keeping the addresses.

2. in go-autonat a server can theoretically do multiple simultaneous dialback attempts (with multiple addresses each) for a peer and resolve them independently

With rust-libp2p this is not possible because when initiating a dial, we don't know the new `ConnectionId`, which we would need if we want to map a `connection_established` or `dial_failure` back to the probe. **Therefore we reject a dial-back request if there is already an ongoing one for this peer**. But I can not really think of a case where a peer would send multiple dial-requests to the same server at the same time. Would only make sense if they want to test different addresses independently; but there isn't any method to initiate something like this in go-autonat either.

3. in go-autonat it always flipps to status `public` if one successful dial-back was received.

In the rust-implementation, we only flip the status if we have confidence 0. So if we have confidence 3 in a status Private, the go implementation would directly switch to `Public`, vs our implementation stays `Private` and reduces confidence. The reason for doing this differently is that a (malicious?) sever could "lie" about us being public; I can image that other protocols or the user might react to a flipped status, in which case this one false flip to "public" could trigger an unwanted reaction, e.g. we stop listening on a relay and can not be reached anymore. Apart from that, a single dial-back may generally succeed for various reasons, which may not mean that the local peer is reachable for everyone. 

4. go-autonat supports `AmbientAutoNAT` (which is more or less what this Pr implements in Rust) and `StaticAutoNAT` (which allows to just statically set a NAT status)

Setting a static NAT status is not possible with our implementation. I don't really see when/ why it is needed.

5. Our implementation allows to manually add peers that can be used as server. This can be e.g. a relay peer or any trusted server, which is useful if we usually don't keep-alive connections to other peers.

## Open questions
I am unsure about the following points, it would be great to get some additional opinions on it:

1. After switching back-and-forth between different options I have now implemented it so that only a `NatStatus` is reported when it is flipped. I don't really think there are use-cases in which a user would like to handle the result from every single probe, and it would open the conflict of: Should the event report the current (updated) status, or the actual result of a probe, which could differ from our assumed status? 
The downside of this is that a flipped status only has confidence 0 at the first iteration. User's may want to wait till a status reached a higher confidence before they take action. That's not possible right now, unless they frequently manually check the current confidence. An alternative would be to add a more complex output type, that would include:
  - the status reported by the probe
  - the current confidence
  - the assumed status (may be different from the current one)
We could extend this with a probe-id and then also enable the user to manually trigger a probe on a specific servers (= "Can this specific peer reach me?"), and read the result of this probe from the OutEvent. I added something like this in 2775041fa8e219672c5362eec89fb1397e2dc6b3, but eventually removed it again because I don't think something like this is really needed in practice. Wdyt?

2. If we don't have a server and / or all servers are throttled, should we reduce confidence or just keep current status? In what cases do we want to switch to status `Unknown`?

3. What addresses should be send to the remote in a dial request? Options:
   - only external addresses (what if there are none?)
   - external and listening addresses (current implementation; the remote will replace IP with the observed one and only dial distinct addresses anyway)
   - external addresses, if they are empty then listening addresses

4. From what I tested, it seems like when resetting the `futures_timer::Delay` with a new duration, the timer won't trigger the waker unless it was polled afterwards at least once. Which means that when resetting the timer outside of the `Behaviour::poll` method (e.g. in `inject_new_external_addr`) the timer becomes only active when `poll` is called the next time due to the waker being called from some other action. Do I have a bug somewhere or is this just how it is? An work-around to solve this could be to copy the waker into a property of `Behaviour` and then manually trigger it each time the timer is reset, but I feel like that's not really clean. Any other ideas?

## Examples
 In addition to the client example from @mxinden I also added one for a server. But this duplicates a lot from the client example, so maybe we should just combine them to a single one and make the input parameters for the client optional?

## Tests
### Integration Tests
I decided against using netsim-embed for the tests due to the lack of documentation and because the examples didn't run on my system (could probably be fixed in  my case, but I don't think tests should have any prerequisites for the host system).
I added some basic tests but the fact that most logic happens in the background and events are only reported if the status is flipped, makes it a bit difficult to test details. That being said, if there are specific cases that are not covered yet I am happy to add more tests.

### Tests in the wild
Unfortunately I don't really have the setup to properly test this in the wild, I only did some tests with the example client against an IPFS node, which seemed to work as expected. Does anybody have the setup and time to test this? Maybe @canewsin?

protocols/autonat/src/behaviour.rs
mxinden: _Late reply. Sorry._


>   * If the IP is private:
>       Maybe there are cases where the peers are all within a private network, but there are firewalls established and a peer would like to use autonat to figure out if it can be dialed by peers in the same private network?

I am fine with supporting this additional use-case. Though I wonder, for the sake of simplicity, whether we should support it early on, or wait till someone actually requests the feature. Don't really have an opinion.


>   * If the demanded address is private:
>       Maybe they still want to test the port because they have configured port-forwarding?

Could they then not just use their public IP address in combination with the port they want to test?

mxinden:
>   1. After switching back-and-forth between different options I have now implemented it so that only a `NatStatus` is reported when it is flipped. I don't really think there are use-cases in which a user would like to handle the result from every single probe

I want to add a use-case for a more detailed `NetworkBehaviour::OutEvent`: As an operator of a libp2p node running `libp2p-autonat` I would like to collect Prometheus metrics, especially when public and thus acting as an AutoNAT server. Exposing additional events via `OutEvent` would allow us to record these events via [`libp2p-metrics`](https://github.com/libp2p/rust-libp2p/tree/master/misc/metrics).

How about exposing outgoing and incoming probe requests and responses as `OutEvents` in addition to the current `NatStatus`, though making clear as a doc comment that the latter is the one to be relied on?

protocols/autonat/src/behaviour.rs
elenaf9: > Though I wonder, for the sake of simplicity, whether we should support it early on, or wait till someone actually requests the feature

It's actually the other way around. If we decide to support this and you don't seen an issue with it, we don't have to check if the IP is private or not, which would simplify the code. See this change https://github.com/elenaf9/rust-libp2p/commit/3dfaa1cfde148ad3cf61b37876a733b05303c808#r61977662 (takes a while to load), where I did just that. 
So the question is more: can this somehow be exploited/ is there a specific reason why we **should** reject/ filter private addresses?

mxinden:
>   2. in go-autonat a server can theoretically do multiple simultaneous dialback attempts (with multiple addresses each) for a peer and resolve them independently
> 
> 
> With rust-libp2p this is not possible because when initiating a dial, we don't know the new `ConnectionId`, which we would need if we want to map a `connection_established` or `dial_failure` back to the probe.

We could add a _probe ID_ to the handler passed when requesting the dial. On success (`ProtocolsHandler::poll`) the handler emits an event with the _probe ID_ to the `NetworkBehaviour`. On `dial_failure` we can extract the _probe ID_ from the returned handler.

> **Therefore we reject a dial-back request if there is already an ongoing one for this peer**.

Even though we could support multiple concurrent dial-back requests (see _probe ID_ suggestion above), I am not sure we should for the sake of resource management and attack resistance.

> But I can not really think of a case where a peer would send multiple dial-requests to the same server at the same time. 

The only thing that I can come up with is a peer that wants to test all addresses, thus sending a new request for each. Currently the server would return on the first successful one, not testing the remaining addresses. That said, I don't think we need to support this use-case.

Long story short, I am fine with the status quo.

mxinden:
>   3. What addresses should be send to the remote in a dial request? Options:
>        
>    * only external addresses (what if there are none?)
>    * external and listening addresses (current implementation; the remote will replace IP with the observed one and only dial distinct addresses anyway)
>    * external addresses, if they are empty then listening addresses


I am in favor of option (2) using external addresses and listening addresses as it is implemented in this pull request today. As you mention above, the remote replaces the IP address anyways. External addresses are the most promising, thus listed first. Listening addresses are the only ones available at startup and might include ports that our reported-address-to-external-address logic (`Transport::address_translation`) missed.

I would guess experience is the best teacher and that we might revisit this decision in the future.

protocols/autonat/src/behaviour.rs
mxinden: > So the question is more: can this somehow be exploited/ is there a specific reason why we **should** reject/ filter private addresses?

Sorry, I should have been more specific. With "simplicity" I meant a smaller potential attack surface, not less lines of code.

I think we are on the same page. I am not able to come up with an attack, thus I am fine either way. Your call :)

protocols/autonat/src/behaviour.rs
```diff
@@ -0,0 +1,853 @@
...
+    time::Duration,
+};
+
+/// Config for the [`Behaviour`].
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct Config {
+    /// Timeout for requests.
+    pub timeout: Duration,
+
+    // == Client Config
```
mxinden:
```suggestion
    /// Client Config
    
```

Nit pick: Do we do this (`==`) anywhere else? Why not just the above? As far as I can tell Rust fmt does not remove the newlines.

protocols/autonat/src/behaviour.rs
```diff
@@ -0,0 +1,853 @@
...
+    /// Delay on init before starting the fist probe.
+    pub boot_delay: Duration,
+    /// Interval in which the NAT should be tested again if max confidence was reached in a status.
+    pub refresh_interval: Duration,
+    /// Interval in which the NAT status should be re-tried if it is currently is unknown
```
mxinden:
```suggestion
    /// Interval in which the NAT status should be re-tried if it is currently unknown
```

protocols/autonat/src/behaviour.rs
```diff
@@ -0,0 +1,853 @@
...
+    fn from(result: Result<Multiaddr, ResponseError>) -> Self {
+        match result {
+            Ok(addr) => NatStatus::Public(addr),
+            Err(ResponseError::DialError) => NatStatus::Private,
+            _ => NatStatus::Unknown,
+        }
+    }
+}
+
+/// Network Behaviour for AutoNAT.
```
mxinden:
```suggestion
/// [`NetworkBehaviour`] for AutoNAT.
```
Just in case we ever rename `NetworkBehaviour` :innocent:

protocols/autonat/src/behaviour.rs
```diff
@@ -0,0 +1,853 @@
...
+        if self.config.use_connected {
+            servers.extend(self.connected.iter().map(|(id, _)| id));
+        }
+
+        servers.retain(|s| !self.throttled_servers.iter().any(|(id, _)| s == &id));
+
+        if servers.is_empty() {
+            return None;
+        }
+        let server = servers[rand::random::<usize>() % servers.len()];
```
mxinden: Why not use [`rand::SliceRandom::choose`](https://docs.rs/rand/0.8.4/rand/seq/trait.SliceRandom.html#tymethod.choose) here?

protocols/autonat/src/behaviour.rs
```diff
@@ -0,0 +1,853 @@
...
+        self.servers.retain(|p| p != peer);
+    }
+
+    // Select a random server for the probe.
+    fn random_server(&mut self) -> Option<PeerId> {
+        // Update list of throttled servers.
+        let i = self.throttled_servers.partition_point(|(_, time)| {
+            *time + self.config.throttle_server_period < Instant::now()
+        });
+        self.throttled_servers.drain(..i);
```
mxinden: Neat!

protocols/autonat/src/behaviour.rs
```diff
@@ -0,0 +1,853 @@
...
+                status_text
+            );
+            let response = DialResponse {
+                result: Err(ResponseError::DialRefused),
+                status_text: Some(status_text),
+            };
+            return Err(response);
+        }
+
+        let ongoing_for_client = self
```
mxinden: Is this really the number of ongoing requests? Is it not  the number of requests within `throttle_clients_period`?

protocols/autonat/src/behaviour.rs
```diff
@@ -0,0 +1,853 @@
...
+                result: Err(ResponseError::DialError),
+                status_text: Some(status_text),
+            };
+            return Err(response);
+        }
+
+        Ok(addrs)
+    }
+
+    // Set the delay to the next probe based on the time of our last prove
```
mxinden:
```suggestion
    // Set the delay to the next probe based on the time of our last probe
```

protocols/autonat/src/behaviour.rs
```diff
@@ -0,0 +1,853 @@
...
+                let mut addresses: Vec<_> = params.external_addresses().map(|r| r.addr).collect();
+                addresses.extend(params.listened_addresses());
+
+                match self.do_probe(addresses) {
+                    Some(_) => {}
+                    None => {
+                        let nat_status = NatStatus::Unknown;
+                        let has_flipped = self.handle_reported_status(nat_status);
+                        if has_flipped {
+                            self.pending_out_events.push_back(self.nat_status.clone());
```
mxinden: Why not return the event directly? `self.schedule_probe` should have already been `poll`ed again and thus the `Waker` registered due to the earlier `continue`.

mxinden:
> 4\. From what I tested, it seems like when resetting the `futures_timer::Delay` with a new duration, the timer won't trigger the waker unless it was polled afterwards at least once.

Yes, correct. This has been a common source of bugs.

> Which means that when resetting the timer outside of the `Behaviour::poll` method (e.g. in `inject_new_external_addr`) the timer becomes only active when `poll` is called the next time due to the waker being called from some other action. Do I have a bug somewhere or is this just how it is?

Not a bug, but intended behaviour which we have to work around.

> An work-around to solve this could be to copy the waker into a property of `Behaviour` and then manually trigger it each time the timer is reset, but I feel like that's not really clean.

Agreed that this explicit handling of the `Waker` is not clean / ideal.

> Any other ideas?

Will include suggestions in the next review.

protocols/autonat/src/behaviour.rs
```diff
@@ -0,0 +1,853 @@
...
+        cx: &mut Context<'_>,
+        params: &mut impl PollParameters,
+    ) -> Poll<NetworkBehaviourAction<Self::OutEvent, Self::ProtocolsHandler>> {
+        let mut is_probe_ready = false;
+        loop {
+            if self.schedule_probe.poll_unpin(cx).is_ready() {
+                is_probe_ready = true;
+                self.schedule_probe.reset(self.config.retry_interval);
+                continue;
+            }
```
mxinden: In this concrete case (in `NetworkBehaviour::poll`) we need to make sure to never return `Poll::Pending` without previously having registered the current `Waker` with the `Delay` in `self.schedule_probe`. I am guessing that this is the intention behind `is_probe_ready` and the `continue`, correct?

What do you think of https://github.com/libp2p/rust-libp2p/commit/cb797abd96e2032b98a9f5307215bb8be18d3cd7 instead? Would that solve the same issue? I would find it easier to understand.

protocols/autonat/src/behaviour.rs
```diff
@@ -0,0 +1,853 @@
...
+                    }
+                }
+            }
+            // An inbound connection can indicate that we are public; adjust the delay to the next probe.
+            ConnectedPoint::Listener { .. } => {
+                if self.confidence == self.config.confidence_max {
+                    if self.nat_status.is_public() {
+                        self.schedule_next_probe(self.config.refresh_interval * 2);
+                    } else {
+                        self.schedule_next_probe(self.config.refresh_interval / 5);
```
mxinden: Good idea!

protocols/autonat/src/behaviour.rs
```diff
@@ -0,0 +1,853 @@
...
+
+    fn inject_new_external_addr(&mut self, addr: &Multiaddr) {
+        self.inner.inject_new_external_addr(addr);
+
+        // New address could be publicly reachable, trigger retry.
+        if !self.nat_status.is_public() {
+            if self.confidence > 0 {
+                self.confidence -= 1;
+            }
+            self.schedule_next_probe(self.config.retry_interval);
```
mxinden: In general, `NetworkBehaviour::poll` is guaranteed to be called after `NetworkBehaviour::inject_xxx` is called unless directly triggered through synchronous methods on `Swarm`, e.g. `Swarm::add_external_address`. As you noted this is a problem here as `NetworkBehaviour::inject_new_external_addr` potentially resets the timer.

I am not sure how to tackle this issue. As you mentioned in the main thread, keeping the previous `Waker` around within `Behaviour` seems rather unclean.

Given that a call to `NetworkBehaviour::poll` is triggered through all kinds of external mechanisms (e.g. a network event, a `Delay` firing in another `NetworkBehaviour`, ...) and thus the `Waker` is properly registered with the `Delay` of `self.schedule_next_probe`, I wonder how often this happens in the wild. In other words I wonder how often a user would call a method on `Swarm` triggering `self.schedule_next_probe` to reset followed by a period with zero external inputs (e.g. network events, ...).

The above said, depending on external mechanisms (e.g. network events) is less than ideal and likely error prone.

Are you able to trigger this error in unit tests?

mxinden (commented):
Looks like a very sturdy implementation. Thanks for the recent improvements!

Couple of comments. Only major thing outstanding from my side is the `Delay` and `Waker` issue.

I will deploy this on one of my public servers so we can continue testing.

protocols/autonat/src/behaviour.rs
```diff
@@ -0,0 +1,853 @@
...
+    }
+
+    // Validate the inbound request and collect the addresses to be dialed.
+    fn resolve_inbound_request(
+        &mut self,
+        sender: PeerId,
+        request: DialRequest,
+    ) -> Result<Vec<Multiaddr>, DialResponse> {
+        // Update list of throttled clients.
+        let i = self.throttled_servers.partition_point(|(_, time)| {
```
mxinden:
```suggestion
        let i = self.throttled_clients.partition_point(|(_, time)| {
```

Copy paste error?

protocols/autonat/src/behaviour.rs
elenaf9: The newlines are on purpose. The `//== Client Config` and `//== Server Config` comments are meant as mental "subheader" for the configs that are listed below them. I changed them now to just `// Client Config` / `// Server Config`, but I am also fine with completely removing them.

protocols/autonat/src/behaviour.rs
elenaf9: True the label is a bit misleading. Changed it to `throttled_for_client`.

protocols/autonat/src/behaviour.rs
elenaf9: Polling the waker after polling the inner behaviour is a good idea. However, depending on the request/ responses that are polled from the inner behaviour, one or multiple `Event`s may have been added to `Behaviour::pending_out_events`, therefore this queue also has to be checked before returning `Poll::Pending`.
I refactored the poll function a bit in dbfc59e: 
1. rearranged it according to you proposal so that the waker is polled after the inner behaviour
2. make it more consistent that all `pending_out_events.push_back(..)` happen within the poll method. Before it was a bit confusing when/ where new events are added to the queue since it was spread out onto different methods.

However, I am not really happy with that solution yet; will try to clean it a bit more in the next days.

protocols/autonat/src/behaviour.rs
elenaf9: I think I did it this way because a previous iteration of the loop could have also pushed an event to the queue, and I want to keep the correct order.

protocols/autonat/src/behaviour.rs
elenaf9: I tried to test this directly on the `autonat::Behaviour` and on a Swarm, but in both cases the test surprisingly passes: b3670b7.
I honestly don't understand why the waker is wakened again in those tests, but not if I reset the delay within `Behaviour::poll`. Any ideas?

elenaf9:
> I want to add a use-case for a more detailed NetworkBehaviour::OutEvent: As an operator of a libp2p node running libp2p-autonat I would like to collect Prometheus metrics, especially when public and thus acting as an AutoNAT server. Exposing additional events via OutEvent would allow us to record these events via libp2p-metrics.
>
> How about exposing outgoing and incoming probe requests and responses as OutEvents in addition to the current NatStatus, though making clear as a doc comment that the latter is the one to be relied on?

Okay yes that sounds reasonable. I added `Request`, `Response` and `Error` events for inbound and outbound probes in d194a931edf793c02a65f20ec8d22e8fa36f63c5. 

I am a bit undecided on whether a `DialError` should be reported as `Response` or `Error`. If we want to be consistent with the  protocol (and also re-use `protocol::ResponseError`) it would be an error, but it differs fundamentally from the other errors because it signals us that we are private, whereas the other errors don't say anything about our NAT status.
Wdyt?

protocols/autonat/src/behaviour.rs
mxinden: >  I honestly don't understand why the waker is wakened again in those tests, but not if I reset the delay within Behaviour::poll. Any ideas?

The following should be the case, no?

1. `inject_new_external_addr` is called.
2. `behaviour::poll` is actively called
    1. `self.schedule_probe.poll_unpin` within `NetworkBehaviour::poll` is called, thus registering the waker.
3. The registered waker is woken up once the timer set through (1) fires.
4. `behaviour::poll` is called, returning an `Event::OutboundProbe`.

mxinden (commented):
Following up on the discussion above:

> However, depending on the request/ responses that are polled from the inner behaviour, one or multiple Events may have been added to Behaviour::pending_out_events, therefore this queue also has to be checked before returning Poll::Pending.

Ah, sorry, I did a mistake in https://github.com/libp2p/rust-libp2p/commit/cb797abd96e2032b98a9f5307215bb8be18d3cd7. The following block should be right at the beginning (within) the loop, not in front of the loop.

``` rust
        if let Some(event) = self.pending_out_events.pop_front() {
            return Poll::Ready(NetworkBehaviourAction::GenerateEvent(event));
        }
```

When that is the case, `self.inner` and `self.schedule_probe` are  only polled when there are no elements in `self.pending_out_events`. If both `self.inner` and `self.schedule_probe` return `Poll::Pending`, then neither of them added an event to `self.pending_out_events` and thus it is safe to return `Poll::Pending` from `NetworkBehaviour::poll`.

---

The "problem" (it is minor) with your current approach is, that you poll `self.inner` and/or `self.schedule_probe` even though we might still have events in `self.pending_out_event`, in other words `NetworkBehaviour::poll` is looking for more events, even though it still has pending ones.

mxinden:
> I am a bit undecided on whether a DialError should be reported as Response or Error. If we want to be consistent with the protocol (and also re-use protocol::ResponseError) it would be an error, but it differs fundamentally from the other errors because it signals us that we are private, whereas the other errors don't say anything about our NAT status.
Wdyt?

Interesting. I see the conflict, though I am not sure I have a clear cut solution either. I am tending towards including `DialError` in `OutboundProbeEvent::Response`. That said, I don't think the current state is too confusing, especially since you document `OutboundProbeEvent::Response` with the word `successfully`.

protocols/autonat/src/behaviour.rs
elenaf9: So if I understand you correctly, the fact that we **actively** call again `Behaviour::poll` / `Swarm::poll` after `inject_new_external_addr` solves it that the waker is registered again?
If that is the case, doesn't this also solves the above issue of the user manually triggering Swarm methods like `inject_new_external_addr`, that reset the timer? Because in order to call such a method, a user would need to stop polling the swarm, call the method, and then actively start polling it again, hence also registering the new waker.

elenaf9:
> When that is the case, self.inner and self.schedule_probe are only polled when there are no elements in self.pending_out_events. If both self.inner and self.schedule_probe return Poll::Pending, then neither of them added an event to self.pending_out_events and thus it is safe to return Poll::Pending from NetworkBehaviour::poll.
>
> The "problem" (it is minor) with your current approach is, that you poll self.inner and/or self.schedule_probe even though we might still have events in self.pending_out_event, in other words NetworkBehaviour::poll is looking for more events, even though it still has pending ones.

Okay yes good point, thanks for the explanation. I followed your proposal in 27762f0.

protocols/autonat/src/behaviour.rs
mxinden: > So if I understand you correctly, the fact that we **actively** call again `Behaviour::poll` / `Swarm::poll` after `inject_new_external_addr` solves it that the waker is registered again?

Correct.

> If that is the case, doesn't this also solves the above issue of the user manually triggering Swarm methods like `inject_new_external_addr`, that reset the timer? Because in order to call such a method, a user would need to stop polling the swarm, call the method, and then actively start polling it again, hence also registering the new waker.

In the ideal case yes, though the user could as well do something like:

``` rust
loop {
  let is_pending = my_swarm.poll_next(cx).is_pending();
  my_swarm.add_external_address(some_address);
  if is_pending {
    return Poll::Pending;
  }
}

File: protocols/autonat/src/protocol.rs [hunk 40]
```diff
@@ -0,0 +1,325 @@
...
+            _ => {
+                log::debug!("Received response with invalid status code.");
+                Err(io::Error::new(
+                    io::ErrorKind::InvalidData,
+                    "invalid response error type",
+                ))
+            }
+        }
+    }
+}
```
mxinden: The generated `ResponseStatus` should implement `TryFrom<i32>` right? If so, how about leveraging that implementation, thus not having two sources of truth on the number values, and manually implement `ResponseStatus` -> `ResponseError`?

```suggestion
impl TryFrom<ResponseStatus> for ResponseError {
    type Error = io::Error;

    fn try_from(value: ResponseStatus) -> Result<Self, Self::Error> {
        match value {
            ResponseStatus::DialError => Ok(ResponseError::DialError),
            // ...
        }
    }
}
```

mxinden (commented):
One small suggestion. Otherwise code looks good to me. Thank you for the continuous work here.

My public server identifies itself as private which is odd (`StatusChanged { old: Unknown, new: Private }`), though that might be on my end. Still need to investigate further.

mxinden:
> My public server identifies itself as private which is odd (`StatusChanged { old: Unknown, new: Private }`), though that might be on my end. Still need to investigate further.

Never mind, this has been a mistake on my end. Public server successfully reports being public (`StatusChanged { old: Unknown, new: Public("/ip4/XXX/tcp/4001") }`). Private node reports being private (`AutoNat(StatusChanged { old: Unknown, new: Private })`).

:rocket:

protocols/autonat/src/protocol.rs
elenaf9: > The generated ResponseStatus should implement TryFrom<i32> right?

Is that done by `prost_build`, or why do you assume it is implemented? If so, could you link me any docs where that's described?
I tried to follow your proposal but I get the error that `TryFrom<i32>` is not implemented for `ResponseStatus`.

elenaf9:
One more note regarding the changed specs for autonat in https://github.com/libp2p/specs/pull/386:

> This restriction as well implies that implementations MUST NOT accept dial requests via relayed connections as one can
not validate the IP address of the requesting node

The request-response protocol does not bubble up the info on via what specific connection a request was received (if we have multiple active connections), so we can not check if that specific connection is relayed or not.
But with 2a73082 I changed it to track all  addresses from active non-relayed connections. When there is at least one active non-relayed connection, we select the remote address as observed address, otherwise (if there are **only** relayed connections) we reject it. 
From how I understood it, the main problem with relayed connections is that we can not know the actual address of the remote. If we have at least one non-relayed connection, that is not a problem, so I would think this fulfills the requirements from the quoted spec?

mxinden:
> From how I understood it, the main problem with relayed connections is that we can not know the actual address of the remote. If we have at least one non-relayed connection, that is not a problem, so I would think this fulfills the requirements from the quoted spec?

Agreed that this fulfills the requirement. Thanks for double-checking.

protocols/autonat/src/protocol.rs
mxinden: Yes, it is generated by `prost_build` for `ResponseStatus`. Not sure whether it is documented anywhere.

See https://github.com/elenaf9/rust-libp2p/pull/3.

mxinden:
>  protocols/autonat: override dial concurrency factor 

That was fast. Thanks! I was about to comment.

mxinden (approved):
:tada: Ready to merge from my side.

Unless there are any further comments I will merge tomorrow.

protocols/autonat/src/behaviour.rs
elenaf9: Ah yes you are right, I didn't think about that. I am able to trigger this error with https://github.com/elenaf9/rust-libp2p/commit/ae95520449da3e807b9efef9b603a3883db43600, the test there never terminates.
Not sure if in practice anyone would implement something like this, but I have to admit it is not too far-fetched.
I don't know if this is a show-stopper or not. If we want to be 100% sure I can just remove the resetting of the timer from any method apart from `NetworkBehaviour::poll`. But worst case this would mean that we have to wait a whole `refresh_period` until a next probe (which is per default 15mins).

mxinden: I am leaning towards keeping it as is, betting on users not to write logic like https://github.com/elenaf9/rust-libp2p/commit/ae95520449da3e807b9efef9b603a3883db43600. I find it intuitive to have to `poll` something after injecting an event (i.e. calling `add_external_address`).

elenaf9: Agreed.

mxinden:
:rocket: Again, thanks @elenaf9 for the continuous work here. A very important piece of libp2p's hole punching story.

PR #2275, opened 2021-10-06
Fix dependabot rand PR

gilescope:
This should fix the rand dependabot PR.


thomaseizinger:
I've triggered CI for you and and it seems there are quite a few more errors. Let me know if you want to continue working on this!

PR #2277, opened 2021-10-06
misc/metrics: Add support for libp2p-gossipsub

mxinden:
A draft adding minimal support for `libp2p-gossipsub` to `libp2p-metrics`.

I think a first version should at least instrument all variants of `GossipsubEvent`.

https://github.com/libp2p/rust-libp2p/blob/937b59de8946eb3773e292691dac4e5f3e081706/protocols/gossipsub/src/behaviour.rs#L115-L144

In the long run we should extend `GossipsubEvent` to e.g. expose details about the _mesh_.

Example on how to use `libp2p-metrics` can be found here: https://github.com/libp2p/rust-libp2p/blob/master/misc/metrics/examples/metrics.rs

A binary using `libp2p-metrics` is e.g. https://github.com/mxinden/rust-libp2p-server/


mxinden:
Merged via https://github.com/libp2p/rust-libp2p/pull/2316.

PR #2295, opened 2021-10-15
core/tests/connection_limit: Fix flake in max_established_incoming

mxinden:
The test `max_established_incoming` starts two networks with a
configured connection limit, spawns up to `limit + 1` connections and
expects the last connection to be closed due to being over the limit.

The previous test implementation depended on both networks to handle
each connection in sequence, which is not always the case.

E.g. while network 2 might expect the last connection to close, network
1 might finish upgrading the last connection before the second to last
  connection, thus expecting the second to last connection to close.

This commit drives network 1 and 2 in sequence and ensures both networks
are finished upgrading a connection before starting a new connection. In
addition it upgrade the test to use `quickcheck`.

### DIFF
--- a/core/tests/connection_limits.rs
+++ b/core/tests/connection_limits.rs
@@ -27,6 +27,7 @@ use libp2p_core::{
     network::{ConnectionLimits, DialError, NetworkConfig, NetworkEvent},
     PeerId,
 };
+use quickcheck::*;
 use rand::Rng;
 use std::task::Poll;
 use util::{test_network, TestHandler};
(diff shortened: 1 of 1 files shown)


mxinden:
@thomaseizinger can you take a look? It is not ideal (mostly due to the conflicts resulting from `NetworkEvent` having a lifetime), still, especially since the test is flaky on our CI, I think it is a step forward.

thomaseizinger:
> @thomaseizinger can you take a look? It is not ideal (mostly due to the conflicts resulting from `NetworkEvent` having a lifetime), still, especially since the test is flaky on our CI, I think it is a step forward.

Will have a look tomorrow!

thomaseizinger (approved):
Looks sound to me!

PR #2297, opened 2021-10-16
protocols/kad: Refactor KademliaEvent

### LINKED ISSUE #2283: Refactor KademliaEvent
I was thinking if we could refactor `InboundPutRecordRequest` and `InboundAddProviderRequest` into a `FilterRequest` or something similar inside `KademliaEvent`.

So instead of this

```rust
KademliaEvent::InboundPutRecordRequest {
                source,
                connection,
                record,
            } => {
                println!(
                    "inbound put record request {:?} {:?}",
                    std::str::from_utf8(record.key.as_ref()).unwrap(),
                    std::str::from_utf8(&record.value).unwrap(),
                );

                self.kademlia.store_mut().put(record);
            }

```

we could have something like this:

```rust
KademliaEvent::FilterRequest{request} => {
        match request {
            KademliaFilter::InboundPutRecordRequest{record, ..} => {
                println!(
                    "inbound put record request {:?} {:?}",
                    std::str::from_utf8(record.key.as_ref()).unwrap(),
                    std::str::from_utf8(&record.value).unwrap(),
                );

                self.kademlia.store_mut().put(record);
            }
            _ => {}
        }
    }

```

supercmmetry:
- Refactored `InboundPutRecordRequest` to `InboundRequest::PutRecord`
- Refactored `InboundAddProviderRequest` to `InboundRequest::AddProvider`
- Refactored `KademliaEvent::InboundRequestServed` to `KademliaEvent::InboundRequest`

Closes #2283

### DIFF
--- a/protocols/kad/src/behaviour.rs
+++ b/protocols/kad/src/behaviour.rs
@@ -1641,17 +1641,19 @@ where
                         return;
                     }
                 },
-                KademliaStoreInserts::FilterBoth => {
-                    self.queued_events
-                        .push_back(NetworkBehaviourAction::GenerateEvent(
-                            KademliaEvent::InboundPutRecordRequest {
-                                source,
-                                connection,
-                                record: record.clone(),
-                            },
-                        ));
-                }
+                KademliaStoreInserts::FilterBoth => {}
             }
+
+            self.queued_events
+                .push_back(NetworkBehaviourAction::GenerateEvent(
+                    KademliaEvent::InboundRequest {
+                        request: InboundRequest::PutRecord {
+                            source,
+                            connection,
+                            record: record.clone(),
+                        },
+                    },
+                ));
         }
 
         // The remote receives a [`KademliaHandlerIn::PutRecordRes`] even in the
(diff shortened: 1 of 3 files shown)


protocols/kad/src/behaviour.rs
```diff
@@ -1641,17 +1637,19 @@ where
...
+                KademliaStoreInserts::FilterBoth => {}
             }
+
+            self.queued_events
+                .push_back(NetworkBehaviourAction::GenerateEvent(
+                    KademliaEvent::InboundRequest {
+                        request: InboundRequest::PutRecord {
+                            source,
+                            connection,
+                            record: record.clone(),
```
mxinden: When `KademliaStoreInserts::Unfiltered` is used, do we need to clone and bubble up the record? In other words, how about making `InboundRequestPutRecord::record` an `Option` with it being `Some(record)` only when `KademliaStoreInserts::FilterBoth` is used and `None` when `Unfiltered`?

protocols/kad/src/behaviour.rs
```diff
@@ -147,10 +147,6 @@ pub enum KademliaStoreInserts {
     /// Whenever a (provider) record is received,
     /// the record is forwarded immediately to the [`RecordStore`].
     Unfiltered,
-    /// Whenever a (provider) record is received, an event is emitted.
```
mxinden, line 150: This documentation is still necessarily and only needs to be updated now that the record is not provided via `KademliaEvent::InboundAddProiderRequest` but `KademliaEvent::InboundRequest`.

protocols/kad/src/behaviour.rs
```diff
@@ -2371,26 +2358,12 @@ pub struct PeerRecord {
 /// See [`NetworkBehaviour::poll`].
 #[derive(Debug)]
 pub enum KademliaEvent {
-    /// A peer sent a [`KademliaHandlerIn::PutRecord`] request and filtering is enabled.
-    ///
-    /// See [`KademliaStoreInserts`] and [`KademliaConfig::set_record_filtering`].
```
mxinden, line 2376: Can you copy parts of this to `InboundRequest::PutRecord`?

mxinden (commented):
A couple of comments, mostly on the documentation side. Thanks for looking into this.

supercmmetry:
@mxinden can you check if there is anything else you need?

mxinden:
Merged via https://github.com/libp2p/rust-libp2p/pull/2321. Thanks @supercmmetry for the help!

PR #2299, opened 2021-10-16
Update rustls 0.20

quininer:


transports/websocket/src/tls.rs
```diff
@@ -105,35 +116,39 @@ impl Builder {
...
             .map_err(|e| Error::Tls(Box::new(e)))?;
         Ok(self)
     }
 
     /// Finish configuration.
     pub fn finish(self) -> Config {
+        let client = rustls::ClientConfig::builder()
+            .with_safe_defaults()
+            .with_root_certificates(self.client_root_store)
+            .with_no_client_auth();
```
thomaseizinger, line 141: I am not 100% what should be happening here but does this make sense if we are constructing a _client_ `Config` here?

transports/websocket/src/tls.rs
```diff
@@ -69,33 +70,43 @@ impl Config {
 
     /// Create a client-only configuration.
     pub fn client() -> Self {
+        let client = rustls::ClientConfig::builder()
+            .with_safe_defaults()
+            .with_root_certificates(client_root_store())
+            .with_no_client_auth();
```
thomaseizinger, line 76: Same here.

thomaseizinger (changes requested):
One comment, otherwise LGTM!

Can you please address the clippy lint failure?

transports/websocket/src/tls.rs
quininer: I'm not sure, I just translated the old code as is.

thomaseizinger: The old code only sets `NoClientAuth` within the `server` constructor (line 108).

For `client`, we don't set anything (line 89).

quininer: `NoClientAuth` is the default value in rustls 0.19, but in rustls 0.20 you need to set it explicitly.

thomaseizinger: Right, thanks for clarifying!

mxinden:
Replaced with https://github.com/libp2p/rust-libp2p/pull/2336. Thanks @quininer for the help!

PR #2306, opened 2021-10-22
Update libsecp256k1 from 0.6 to 0.7

koushiro:

Close #2238


thomaseizinger (approved):
Nice! Pulling through the upgrade of `rand` is a major PITA but I am happy to go with this solution for now as it at least updates `libsecp256k1` to the latest version which is a much more critical dependency IMO.

What do you think @mxinden?

mxinden (approved):
Thanks @koushiro for the help!

Agreed @thomaseizinger. Given that I can not find any release notes nor changelog entries I am assuming this includes security or bug fixes. Let's include it in v0.40.0-rc.3.

PR #2309, opened 2021-10-24
protocols/kad: populate the key field to fix go-libp2p interop

pradt2:
I had to make this change to get the (predominantly go-libp2p based) IPFS DHT nodes to accept any records issued by rust-libp2p. 

I'm a bit surprised that this change has to be made since the key is also present in the Record itself. However, I double checked against the go-libp2p implementation, and there they do indeed set the key both in the Record, as well as directly in the Message.

All tests in the protocols.rs file are commented out, so I didn't bother trying to fit in a test case for this. I hope that's okay.


mxinden:
Great catch @pradt2! Thank you for both debugging and providing a fix.

For the reference, the corresponding Golang logic expecting both a key in the message in and the record:

https://github.com/libp2p/go-libp2p-kad-dht/blob/7724838e46d1822b28fd20611ab9815d4ce2d94f/handlers.go#L151-L164

I will need to do some more testing, communicate with the go-libp2p folks and update the specification. I am sorry for the delay.

Again, great work @pradt2!

mxinden:
https://github.com/libp2p/specs/pull/373 updates the specification accordingly. 

@pradt2 would you mind adding a changelog entry to `protocols/kad/CHANGELOG.md`? Other than that, this is ready to go.

pradt2:
@mxinden done

mxinden (approved):
Cool, that was very quick! Thank you @pradt2!

PR #2311, opened 2021-10-25
protocols/mdns: Use a random alphanumeric string for peer name

### LINKED ISSUE #2285: protocols/mdns: Use a random string as peer-name
With https://github.com/libp2p/specs/pull/368 the definition of the _peer name_ changed in the mDNS specification.

> peer-name is the case-insensitive unique identifier of the peer, and is less than 64 characters.
>
> As the this field doesn't carry any meaning, it is sufficient to ensure the uniqueness of this identifier. Peers SHOULD generate a random, lower-case alphanumeric string of least 32 characters in length when booting up their node. Peers SHOULD NOT use their Peer ID here because a future Peer ID could exceed the DNS label limit of 63 characters.

https://github.com/libp2p/specs/blob/master/discovery/mdns.md

`libp2p-mdns` should be adjusted accordingly.

Also see https://github.com/libp2p/go-libp2p/pull/1222 for the corresponding change on the Golang side.

jochasinga:
Fixes #2285.

### DIFF
--- a/protocols/mdns/CHANGELOG.md
+++ b/protocols/mdns/CHANGELOG.md
@@ -1,6 +1,11 @@
 # 0.33.1 [unreleased]
 
-- Use a random alphanumeric string for mDNS peer name (see [PR 2311]).
+- Use a random alphanumeric string instead of the local peer ID for mDNS peer
+  name (see [PR 2311]).
+
+  Note that previous versions of `libp2p-mdns` expect the peer name to be a
+  valid peer ID. Thus they will be unable to discover nodes running this new
+  version of `libp2p-mdns`.
 
 [PR 2311]: https://github.com/libp2p/rust-libp2p/pull/2311/
 
--- a/protocols/mdns/CHANGELOG.md
+++ b/protocols/mdns/CHANGELOG.md
@@ -1,4 +1,4 @@
-# 0.33.1 [unreleased]
+# 0.34.0 [unreleased]
 
 - Use a random alphanumeric string instead of the local peer ID for mDNS peer
   name (see [PR 2311]).
--- a/protocols/mdns/CHANGELOG.md
+++ b/protocols/mdns/CHANGELOG.md
@@ -1,3 +1,9 @@
+# 0.33.1 [unreleased]
+
+- Use a random alphanumeric string for mDNS peer name (see [PR 2311]).
+
+[PR 2311]: https://github.com/libp2p/rust-libp2p/pull/2311/
+
 # 0.33.0 [2021-11-16]
 
 - Update dependencies.
--- a/protocols/mdns/Cargo.toml
+++ b/protocols/mdns/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "libp2p-mdns"
 edition = "2018"
-version = "0.33.0"
+version = "0.33.1"
 description = "Implementation of the libp2p mDNS discovery method"
 authors = ["Parity Technologies <admin@parity.io>"]
 license = "MIT"
--- a/protocols/mdns/Cargo.toml
+++ b/protocols/mdns/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "libp2p-mdns"
 edition = "2018"
-version = "0.33.1"
+version = "0.34.0"
 description = "Implementation of the libp2p mDNS discovery method"
 authors = ["Parity Technologies <admin@parity.io>"]
 license = "MIT"
--- a/protocols/mdns/src/dns.rs
+++ b/protocols/mdns/src/dns.rs
@@ -271,7 +271,7 @@ fn random_string(length: usize) -> String {
 /// Generates a peer ID bytes for a DNS query.
 fn generate_peer_id() -> Vec<u8> {
     // Use a variable-length random string for mDNS peer name.
-    // See discussion: https://github.com/libp2p/go-libp2p/pull/1222
+    // See https://github.com/libp2p/rust-libp2p/pull/2311/
     let peer_name = random_string(32 + thread_rng().gen_range(0..32));
 
     // allocate with a little extra padding for QNAME encoding
--- a/protocols/mdns/src/dns.rs
+++ b/protocols/mdns/src/dns.rs
@@ -23,6 +23,8 @@
 use crate::{META_QUERY_SERVICE, SERVICE_NAME};
 use libp2p_core::{Multiaddr, PeerId};
 use std::{borrow::Cow, cmp, error, fmt, str, time::Duration};
+use rand::{thread_rng, Rng};
+use rand::distributions::Alphanumeric;
 
 /// Maximum size of a DNS label as per RFC1035.
 const MAX_LABEL_LENGTH: usize = 63;
@@ -280,6 +282,14 @@ fn segment_peer_id(peer_id: String) -> String {
     out
 }
 
+fn random_string(length: usize) -> String {
+    thread_rng()
+        .sample_iter(&Alphanumeric)
+        .take(length)
+        .map(char::from)
+        .collect()
+}
+
 /// Combines and encodes a `PeerId` and service name for a DNS query.
 fn encode_peer_id(peer_id: &PeerId) -> Vec<u8> {
     // DNS-safe encoding for the Peer ID
@@ -287,7 +297,7 @@ fn encode_peer_id(peer_id: &PeerId) -> Vec<u8> {
     // ensure we don't have any labels over 63 bytes long
     let encoded_peer_id = segment_peer_id(raw_peer_id);
     let service_name = str::from_utf8(SERVICE_NAME).expect("SERVICE_NAME is always ASCII");
-    let peer_name = [&encoded_peer_id, service_name].join(".");
+    let peer_name = random_string(32 + thread_rng().gen_range(0..32));
 
     // allocate with a little extra padding for QNAME encoding
     let mut peer_id_bytes = Vec::with_capacity(peer_name.len() + 32);
@@ -454,6 +464,14 @@ mod tests {
             [&str_63, str_63.as_str()].join(".")
         );
         assert_eq!(segment_peer_id(str_127), [&str_63, &str_63, "x"].join("."));
+    }
+
+    #[test]
+    fn test_random_string() {
+        let varsize = thread_rng().gen_range(0..32);
+        let size = 32 + varsize;
+        let name = random_string(size);
+        assert_eq!(name.len(), size);
     }
 
     // TODO: test limits and errors
--- a/protocols/mdns/src/dns.rs
+++ b/protocols/mdns/src/dns.rs
@@ -115,8 +115,8 @@ pub fn build_query_response(
     // Add a limit to 2^16-1 addresses, as the protocol limits to this number.
     let addresses = addresses.take(65535);
 
-    let peer_id_bytes = generate_peer_id();
-    debug_assert!(peer_id_bytes.len() <= 0xffff);
+    let peer_name_bytes = generate_peer_name();
+    debug_assert!(peer_name_bytes.len() <= 0xffff);
 
     // The accumulated response packets.
     let mut packets = Vec::new();
@@ -129,7 +129,7 @@ pub fn build_query_response(
     for addr in addresses {
         let txt_to_send = format!("dnsaddr={}/p2p/{}", addr.to_string(), peer_id.to_base58());
         let mut txt_record = Vec::with_capacity(txt_to_send.len());
-        match append_txt_record(&mut txt_record, &peer_id_bytes, ttl, &txt_to_send) {
+        match append_txt_record(&mut txt_record, &peer_name_bytes, ttl, &txt_to_send) {
             Ok(()) => {
                 records.push(txt_record);
             }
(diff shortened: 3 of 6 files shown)


protocols/mdns/src/dns.rs
```diff
@@ -280,14 +282,22 @@ fn segment_peer_id(peer_id: String) -> String {
...
+        .map(char::from)
+        .collect()
+}
+
 /// Combines and encodes a `PeerId` and service name for a DNS query.
 fn encode_peer_id(peer_id: &PeerId) -> Vec<u8> {
     // DNS-safe encoding for the Peer ID
     let raw_peer_id = data_encoding::BASE32_DNSCURVE.encode(&peer_id.to_bytes());
     // ensure we don't have any labels over 63 bytes long
     let encoded_peer_id = segment_peer_id(raw_peer_id);
```
mxinden: With the encoding of peer IDs no longer needed, this and all of the underlying encoding logic can be removed, right?

mxinden (commented):
Very cool. Thanks a bunch for picking this up @jochasinga!

Would you mind adding a changelog entry as well?

//CC @peat as you have been working on this logic in the past.

protocols/mdns/src/dns.rs
jochasinga: Yes you're right. Looks like every call to `encode_peer_id()` can also get rid of the `peer_id: &PeerId` argument as well?

jochasinga: Inherently the change to `encode_peer_id()` will affect this public function 
```rust
pub fn build_query_response(
    id: u16,
    peer_id: PeerId, /* unused */
    addresses: impl ExactSizeIterator<Item = Multiaddr>,
    ttl: Duration,
) -> Vec<MdnsPacket>
```
which will no longer need `peer_id` in the argument. Is this change fine?

https://github.com/libp2p/rust-libp2p/blob/6e76435b80209bed1eefd38597b3ba25aef00fa7/protocols/mdns/src/dns.rs#L109

mxinden: :+1: removing `peer_id` from `build_query_response` sounds good to me.

jochasinga: My mistake. Look like `build_query_response` still uses `peer_id` so there's no need to touch it.

jochasinga:
> Very cool. Thanks a bunch for picking this up @jochasinga!
> 
> Would you mind adding a changelog entry as well?
> 
> //CC @peat as you have been working on this logic in the past.

In the changelog it is currently 0.32.0rc-1. I didn't find a doc explaining the process. Do you mind advise how this change should be reflected in the versioning?

mxinden:
Sorry for the confusion @jochasinga. Could you apply the diff below?

```diff
diff --git a/protocols/mdns/CHANGELOG.md b/protocols/mdns/CHANGELOG.md
index f3287379..a2376242 100644
--- a/protocols/mdns/CHANGELOG.md
+++ b/protocols/mdns/CHANGELOG.md
@@ -1,3 +1,9 @@
+# 0.32.1 [unreleased]
+
+- Use a random alphanumeric string for mDNS peer name (see [PR 2311]).
+
+[PR 2311]: https://github.com/libp2p/rust-libp2p/pull/2311/
+
 # 0.32.0 [2021-11-01]
 
 - Make default features of `libp2p-core` optional.
diff --git a/protocols/mdns/Cargo.toml b/protocols/mdns/Cargo.toml
index 43370a9b..0bf1605e 100644
--- a/protocols/mdns/Cargo.toml
+++ b/protocols/mdns/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "libp2p-mdns"
 edition = "2018"
-version = "0.32.0"
+version = "0.32.1"
 description = "Implementation of the libp2p mDNS discovery method"
 authors = ["Parity Technologies <admin@parity.io>"]
 license = "MIT"
```

protocols/mdns/src/dns.rs
```diff
@@ -116,7 +115,7 @@ pub fn build_query_response(
     // Add a limit to 2^16-1 addresses, as the protocol limits to this number.
     let addresses = addresses.take(65535);
 
-    let peer_id_bytes = encode_peer_id(&peer_id);
+    let peer_id_bytes = generate_peer_id();
```
mxinden:
```suggestion
    let peer_name_bytes = generate_peer_name();
```

protocols/mdns/src/dns.rs
```diff
@@ -260,34 +259,20 @@ fn append_u16(out: &mut Vec<u8>, value: u16) {
...
 
-/// Combines and encodes a `PeerId` and service name for a DNS query.
-fn encode_peer_id(peer_id: &PeerId) -> Vec<u8> {
-    // DNS-safe encoding for the Peer ID
-    let raw_peer_id = data_encoding::BASE32_DNSCURVE.encode(&peer_id.to_bytes());
-    // ensure we don't have any labels over 63 bytes long
-    let encoded_peer_id = segment_peer_id(raw_peer_id);
-    let service_name = str::from_utf8(SERVICE_NAME).expect("SERVICE_NAME is always ASCII");
-    let peer_name = [&encoded_peer_id, service_name].join(".");
+/// Generates a peer ID bytes for a DNS query.
```
mxinden:
```suggestion
/// Generates a random peer name as bytes for a DNS query.
```

protocols/mdns/src/dns.rs
```diff
@@ -260,34 +259,20 @@ fn append_u16(out: &mut Vec<u8>, value: u16) {
...
+fn generate_peer_id() -> Vec<u8> {
+    // Use a variable-length random string for mDNS peer name.
+    // See https://github.com/libp2p/rust-libp2p/pull/2311/
+    let peer_name = random_string(32 + thread_rng().gen_range(0..32));
 
     // allocate with a little extra padding for QNAME encoding
     let mut peer_id_bytes = Vec::with_capacity(peer_name.len() + 32);
```
mxinden:
```suggestion
    let mut peer_name_bytes = Vec::with_capacity(peer_name.len() + 32);
```

mxinden (commented):
Looks good to me.

One small nit that just now caught my attention: Given that the peer name is no longer based on the peer ID, would you mind renaming the references from peer ID to peer name?

elenaf9:
Just that I understand it correctly: with this change, the `PeerId` that is reported in an `MdnsEvent` (Either in the `DiscoveredAddrsIter` or `ExpiredAddrsIter`) is not the actual id of the peer that is listening on that address?

If yes, shouldn't this then also be changed in the examples? E.g. `/examples/distributed-key-value-store.rs` adds the mdns peer-address pairs to the kademlia DHT, which then wouldn't really make sense.

mxinden:
> Just that I understand it correctly: with this change, the `PeerId` that is reported in an `MdnsEvent` (Either in the `DiscoveredAddrsIter` or `ExpiredAddrsIter`) is not the actual id of the peer that is listening on that address?
> 
> If yes, shouldn't this then also be changed in the examples? E.g. `/examples/distributed-key-value-store.rs` adds the mdns peer-address pairs to the kademlia DHT, which then wouldn't really make sense.

You are raising a very important point here @elenaf9 that I wish I didn't miss before writing https://github.com/libp2p/rust-libp2p/issues/2285. Sorry for the trouble this will cause @jochasinga.

I will write up more details on https://github.com/libp2p/rust-libp2p/issues/2285, just a short summary for now:

When receiving an mDNS response, rust-libp2p actually does parse the peer name as a `PeerId` and discards an address in case it contains a `/p2p/QmXX` `PeerId` which doesn't match the one retrieved from the peer name field.
    
https://github.com/libp2p/rust-libp2p/blob/012287ad1ecbc548c1233e9b4b8387de7af280e6/protocols/mdns/src/query.rs#L162-L176

The above implies that peers not running this patch will discard any responses send by peers that are running this patch.

To prevent a network to split, we will need to do a step in between namely to ignore the peer name of a mDNS response and not discard responses in case there is a mismatch. Once that is rolled out on a network, step two would be to roll out this patch setting the peer name to a random string.

jochasinga:
@mxinden keep me posted on the update. Would love to get this fixed up.

mxinden:
> @mxinden keep me posted on the update. Would love to get this fixed up.

See https://github.com/libp2p/rust-libp2p/issues/2285#issuecomment-964229679.

protocols/mdns/src/query.rs
```diff
@@ -218,9 +208,9 @@ impl MdnsPeer {
     pub fn new(
         packet: &Packet<'_>,
         record_value: String,
-        my_peer_id: PeerId,
         ttl: u32,
-    ) -> MdnsPeer {
+    ) -> Result<MdnsPeer, String> {
```
jochasinga: @mxinden @MarcoPolo should this return a `Result<MdnsPeer, _>` or an `Option<MdnsPeer>`? I feel that the first makes it clearer.

MarcoPolo: I think this is a little tricky, but I think it comes down to two things:
1. Is the returned error helpful or actionable? Can it be propagated to the user, or can the caller do something with that error that is useful?
2. Is the returned error correct (This is actually a corollary to the above)

For 1, I don't think the caller can do anything different. Either it can get a peer from the record+packet, or it can't. The error won't change anything. This is true for `MdnsPacket` too. In some other system you can imagine an error leading to a retry, or a user prompt, or some error specific remediation, but not here.

For 2, I don't think the error returned is always correct. There are many branches that return `None` in the logic below, and they aren't all related to a missing peer id. So this error could actually be confusing if the error was actually that there weren't any `TXT` fields (for example).

So my thoughts are that this should be an `Option<...>` like `MdnsPacket`.

jochasinga: Resolved in 7dc6b1f

jochasinga:
@mxinden Do we need to change the examples per @elenaf9 [comment](https://github.com/libp2p/rust-libp2p/pull/2311#issuecomment-958977601)?

MarcoPolo:
> Just that I understand it correctly: with this change, the PeerId that is reported in an MdnsEvent (Either in the DiscoveredAddrsIter or ExpiredAddrsIter) is not the actual id of the peer that is listening on that address?

(from @elenaf9)

With the changes in here now, I don't think this is true anymore.

Now the peerid is correct in MdnsResponse, which is used to generate the MdnsEvent. So I don't think we need to change the examples. If this is wrong, then the correct fix is to have the correct `PeerId` be reported, not changing the examples.

MarcoPolo:
@jochasinga can you please rebase this branch?

jochasinga:
@mxinden can you please run checks on this?

protocols/mdns/src/query.rs
```diff
@@ -269,10 +253,13 @@ impl MdnsPeer {
             })
             .collect();
 
-        MdnsPeer {
-            addrs,
-            peer_id: my_peer_id,
-            ttl,
+        match my_peer_id {
```
MarcoPolo: nit: you can simplify this a bit by doing:
```
peer_id.map(|peer_id| MdnsPeer {addrs, peer_id, ttl})
```

protocols/mdns/src/query.rs
```diff
@@ -303,3 +290,50 @@ impl fmt::Debug for MdnsPeer {
...
+                    let record_value = match record.data {
+                        RData::PTR(record) => record.0.to_string(),
+                        _ => return None,
+                    };
+                    return Some(record_value);
+                })
+                .next()
+                .unwrap();
+
+            let peer = MdnsPeer::new(&packet, record_value, ttl).unwrap();
```
MarcoPolo: nit, I would make this an `expect` so that if this fails it's a useful message to the test runner.

protocols/mdns/src/query.rs
jochasinga: Great catch. I never knew that.

protocols/mdns/CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+# 0.33.1 [unreleased]
+
+- Use a random alphanumeric string for mDNS peer name (see [PR 2311]).
```
mxinden:
```suggestion
- Use a random alphanumeric string instead of the local peer ID for mDNS peer
  name (see [PR 2311]).

  Note that previous versions of `libp2p-mdns` expect the peer name to be a
  valid peer ID. Thus they will be unable to discover nodes running this new
  version of `libp2p-mdns`.
```

protocols/mdns/Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "libp2p-mdns"
 edition = "2018"
-version = "0.33.0"
+version = "0.33.1"
```
mxinden:
```suggestion
version = "0.34.0"
```

While not a breaking change on the API level, it is on the network level. Using a patch release here would make downstream dependents auto-update. Given that this is a breaking change and might split a network, I would prefer it to be an explicit decision to update.

Could you bump the version of `libp2p-mdns` in the root `Cargo.toml` and mention the update in the root `CHANGELOG.md` file as well?

mxinden (commented):
Code looks good to me. Thank you for the many iterations! 

Just two comments on the release process:

protocols/mdns/CHANGELOG.md
jochasinga: Do I need to correct the `protocols/mdns/CHANGELOG` to 0.34.0 as well here?

mxinden: Yes please :pray:

mxinden (approved):
I will merge later today unless anyone objects.

Thanks @jochasinga for all the work and thanks @MarcoPolo for the additional help!

PR #2316, opened 2021-10-29
protocols/gossipsub: Add mesh metrics

divagant-martian:
Small part of #2235 
I found it increasingly hard to reason about the correctness of the way we update each metrics with so many metrics at once, so this PR focus only on the mesh related metrics and topic_peer metrics. It also includes #2277


protocols/gossipsub/src/metrics.rs
```diff
@@ -0,0 +1,255 @@
...
+    topic_peers_count: Family<TopicHash, Gauge>,
+
+    /* Metrics regarding mesh state */
+    /// Number of peers in our mesh. This metric should be updated with the count of peers for a
+    /// topic in the mesh regardless of inclusion and churn events.
+    mesh_peer_counts: Family<TopicHash, Gauge>,
+    /// Number of times we include peers in a topic mesh for different reasons.
+    mesh_peer_inclusion_events: Family<(TopicHash, Inclusion), Counter>,
+    /// Number of times we remove peers in a topic mesh for different reasons.
+    mesh_peer_churn_events: Family<(TopicHash, Churn), Counter>,
```
mxinden: `(TopicHash, Churn)` would encode as `mesh_peer_churn_events_total{<some-topic-hash>="<some-churn-variant>"} <some-value>`. I don't think that is what you actually want, right? Instead I would create a new struct like:

```rust
#[derive(Encode)]
struct MeshPeerChurnEventsLabels {
    hash: TopicHash,
    churn: Churn,
}
```

And then here do:

```suggestion
    mesh_peer_churn_events: Family<MeshPeerChurnEventsLabels, Counter>,
```

protocols/gossipsub/src/metrics.rs
```diff
@@ -0,0 +1,255 @@
...
+use open_metrics_client::metrics::counter::Counter;
+use open_metrics_client::metrics::family::Family;
+use open_metrics_client::metrics::gauge::Gauge;
+use open_metrics_client::registry::Registry;
+
+use crate::topic::TopicHash;
+
+// Default value that limits how many topics for which there has been a previous or current
+// subscription do we store metrics.
+const DEFAULT_MAX_EXPLICITLY_SUBSCRIBED_TOPICS: usize = 300;
```
mxinden: While a bit of a hack in my eyes, this seems like a practical solution to the problem of metric churn.

What is your experience with this constant limit? Do you ever hit it on your production nodes?

protocols/gossipsub/src/behaviour.rs
```diff
@@ -302,6 +304,9 @@ pub struct Gossipsub<
     /// calculating the message-id and sending to the application. This is designed to allow the
     /// user to implement arbitrary topic-based compression algorithms.
     data_transform: D,
+
+    /// Keep track of a set of internal metrics relating to gossipsub.
+    metrics: Option<Metrics>,
```
mxinden, line 309: :+1:

mxinden (commented):
Sorry for the delay here @divagant-martian. I like the new approach of passing a registry at construction and keeping an `Option<Metrics>` at runtime.

protocols/gossipsub/src/metrics.rs
divagant-martian: Thanks Max, I've fixed it as suggested

protocols/gossipsub/src/metrics.rs
divagant-martian: For us the subscriptions that are really important occur at the beginning and are few, this is a very high upper limit. Still, that's why I left it as an application-configurable value

divagant-martian:
Just solved conflicts with master, in case you can give this a review @mxinden Thx

protocols/gossipsub/src/metrics.rs
AgeManning: We're currently bounded at about 250 topics (we can double our topic subscription around a fork boundary), most operating nodes would have around 125. 
I guess we've set this value selfishly around our needs a little, but hopefully is sufficiently large for others.

protocols/gossipsub/src/behaviour.rs
```diff
@@ -318,6 +323,25 @@ where
...
+    }
+
+    /// Creates a [`Gossipsub`] struct given a set of parameters specified via a
+    /// [`GossipsubConfig`]. This has no subscription filter and uses no compression.
+    /// Metrics can be evaluated by passing a reference to a [`Registry`].
+    pub fn new_with_metrics(
+        privacy: MessageAuthenticity,
+        config: GossipsubConfig,
+        metrics_registry: &mut Registry,
+        metrics_config: MetricsConfig,
```
mxinden, line 339: Just a thought: Instead of passing both a `Registry` and ` MetricsConfig`, one could as well require the user to build `Metrics` themselves outside of `new_with_metrics` and then just passing `Metrics` to `new_with_metrics`.

mxinden (approved):

protocols/gossipsub/src/behaviour.rs
divagant-martian: That would simplify the parameters, but would also expose a type that is otherwise useless (from an user perspective). The `Metrics` type does not expose read functions, or a fancy builder, etc. I'd prefer it this way for these reasons, what are your thoughts?

mxinden: > but would also expose a type that is otherwise useless

Good point. Might as well give users the wrong impression that they can extract the metrics from the `Metrics` struct.

Let's keep it as is.

mxinden (commented):
@divagant-martian could you add a changelog entry in `protocols/gossipsub/CHANGELOG.md` and `misc/metrics/CHANGELOG.md`?

Other than that, this looks good to me and is ready to be merged.

divagant-martian:
I added the Changelog entries and bumped the version of the metrics toml. I also updated the style of this file to follow the other ones. Let me know if the last two steps were not needed / should be reverted. Thanks!

mxinden:
>  Max likes final punctuation. 

Yes indeed :D

PR #2076, opened 2021-05-25
protocols: Implement Direct Connection Upgrade through Relay (dcutr)

mxinden:
This pull request implements the [Direct Connection Upgrade through Relay](https://github.com/libp2p/specs/pull/173) specification. For the greater picture please see https://github.com/libp2p/rust-libp2p/issues/2052 and https://github.com/libp2p/specs/issues/312.

**Status quo**

It builds on top of https://github.com/libp2p/rust-libp2p/pull/2059 and https://github.com/libp2p/rust-libp2p/pull/2066. Please post review comments related to circuit relay v2 and simultaneous open on https://github.com/libp2p/rust-libp2p/pull/2059 and https://github.com/libp2p/rust-libp2p/pull/2066.

PR #2317, opened 2021-10-29
swarm/: Enable advanced dialing requests

### LINKED ISSUE #2249: swarm/: Enable dialing a specific fixed set of addresses for a single peer
**Status Quo**

One can either
- (a) request a dial of a specific peer by their `PeerId` via `NetworkBehaviourAction::DialPeer`, where the concrete addresses to be dialed are collected via `NetworkBehaviour::addresses_of_peer`

- or (b)  request a dial of a single specific address via `NetworkBehaviourAction::DialAddress`

**Problem**

A `NetworkBehaviour` implementation might want to choose the specific fixed set of addresses to be dialed for a given peer. This is the case in https://github.com/libp2p/rust-libp2p/pull/2076 where the remote provides a specific set of addresses to be dialed by the local node.

Using `NetworkBehaviourAction::DialPeer` and providing the set of addresses via `NetworkBehaviour::addresses_of_peer` does not prevent other `NetworkBehaviour` implementations to add bogus addresses to the set via `NetworkBehaviour::addresses_of_peer`.

**Proposal**

- Merge `NetworkBehaviourAction::DialPeer` and `NetworkBehaviourAction::DialAddress` into a single `NetworkBehaviourAction::Dial`.
- In `NetworkBehaviourAction::Dial` allow specifying either:
  - A `PeerId` only, where the `Swarm` retrieves corresponding addresses via `NetworkBehaviour::addresses_of_peer`.
  - A `PeerId` and a set of addresses
    - with the option of the `Swarm` extending the set of addresses via `NetworkBehaviour::addresses_of_peer`.
  - A set of addresses.

**Miscellaneous**

- Next to enabling a `NetworkBehaviour` implementation to specify a specific fixed set of addresses to be dialed, the change above would simplify other implementations. E.g. in the case where a `NetworkBehaviour` implementation wants to dial a peer by its `PeerId`, also providing some addresses, it can provide these addresses right away within `NetworkBehaviourAction::Dial`, instead of first emitting a `NetworkBehaviourAction::DialPeer` and then providing the set of addresses via `NetworkBehaviour::addresses_of_peer`.

- To reduce conflicts I would suggest tackling this after https://github.com/libp2p/rust-libp2p/pull/2248 is merged.

mxinden:
Enable advanced dialing requests both on `Swarm` and via
`NetworkBehaviourAction`. Users can now trigger a dial with a specific
set of addresses, optionally extended via
`NetworkBehaviour::addresses_of_peer`. In addition the whole process is
now modelled in a type safe way via the builder pattern.

Example of a `NetworkBehaviour` requesting a dial to a specific peer
with a set of addresses additionally extended through
`NetworkBehaviour::addresses_of_peer`:

```rust
NetworkBehaviourAction::Dial {
    opts: DialOpts::peer_id(peer_id)
              .condition(PeerCondition::Always)
              .addresses(addresses)
              .extend_addresses_through_behaviour()
              .build(),
    handler,
}
```

Example of a user requesting a dial to an unknown peer with a single
address via `Swarm`:

```rust
swarm1.dial(
    DialOpts::unknown_peer_id()
        .address(addr2.clone())
        .build()
)
```

- Fixes https://github.com/libp2p/rust-libp2p/issues/2249.
- Enables https://github.com/libp2p/rust-libp2p/pull/2076 and https://github.com/libp2p/rust-libp2p/pull/2262/ to provide a fixed exclusive set of addresses to dial (//CC @elenaf9).

---

Creating this as a draft for now. I have only updated `libp2p-swarm` for now.

What do folks think of the interface? (//CC @thomaseizinger as this follows your second suggestion provided in https://github.com/libp2p/rust-libp2p/issues/2249#issuecomment-927453562.)


swarm/src/lib.rs
```diff
@@ -1731,3 +1777,134 @@ mod tests {
...
+
+    #[derive(Debug)]
+    pub enum DialOpts {
+        // TODO:  How about folding these structs into the enum variants?
+        WithPeerId(WithPeerId),
+        WithPeerIdWithAddresses(WithPeerIdWithAddresses),
+        WithoutPeerIdWithAddress(WithoutPeerIdWithAddress),
+    }
+
+    impl DialOpts {
```
elenaf9:
```suggestion
    enum Opts {
        // TODO:  How about folding these structs into the enum variants?
        WithPeerId(WithPeerId),
        WithPeerIdWithAddresses(WithPeerIdWithAddresses),
        WithoutPeerIdWithAddress(WithoutPeerIdWithAddress),
    }
    
    pub struct DialOpts {
        opts: Opts
    }

    impl DialOpts {
```
Consider wrapping the enum in a struct to emphasize that the user should not try to build the variants directly, but instead use the methods.

swarm/src/lib.rs
```diff
@@ -1731,3 +1777,134 @@ mod tests {
...
+
+    #[derive(Debug)]
+    pub struct WithPeerIdWithAddresses {
+        pub(crate) peer_id: PeerId,
+        pub(crate) condition: PeerCondition,
+        pub(crate) addresses: Vec<Multiaddr>,
+        pub(crate) extend_addresses_through_behaviour: bool,
+    }
+
+    impl WithPeerIdWithAddresses {
```
elenaf9:
```suggestion
    impl WithPeerIdWithAddresses {
        pub fn condition(mut self, condition: PeerCondition) -> Self {
            self.condition = condition;
            self
        }
        
```
Maybe add this method here as well.

swarm/src/lib.rs
```diff
@@ -1731,3 +1777,134 @@ mod tests {
...
+        pub fn peer_id(peer_id: PeerId) -> WithPeerId {
+            WithPeerId {
+                peer_id,
+                condition: Default::default(),
+            }
+        }
+
+        pub fn unknown_peer_id() -> WithoutPeerId {
+            WithoutPeerId {}
+        }
```
elenaf9:
```suggestion
        pub fn address(address: Multiaddr) -> WithoutPeerIdWithAddress {
            WithoutPeerIdWithAddress { address }
        }
```
I do like the consistent syntax of `DialOpts::peer_id(..).addresses()` and `DialOpts::unknown_peer_id(..).address(..)`.
But this similar syntax suggests that `WithPeerId` and `WithoutPeerId` are alike, which is not the case because `WithPeerId` can be a "final" state from which you can call `build()`, whereas  `WithoutPeerId` is only intermediate. Not sure about this, but I feel like it could end up confusing the user, since even though it follows the same naming convention, `WithoutPeerId` is not a valid `DialOpt`.

Optionally if we follow the above suggestion, we could keep it consist by also adding
```rust
impl WithoutPeerIdWithAddress {
        pub fn peer_id(self, peer_id: PeerId) -> WithPeerIdWithAddresses {
            WithPeerIdWithAddresses {
                peer_id,
                condition: Default::default(),,
                addresses: vec![self.address],
                extend_addresses_through_behaviour: false,
            }
        }
    }
```
The only problem with this is that  without a peer id you only have one address, whereas with an peer id there can be multiple.

elenaf9 (commented):
I really like the new interface, would be very useful to have this feature!

swarm/src/lib.rs
mxinden: Given the structs `WithPeerId`, `WithPeerIdWithAddresses` and `WithoutPeerIdWithAddress` can not be constructed outside of `dial_opts`, the user can not build `DialOpts` themselves. Still I understand that this is not intuitive and that it would be nice to make it clear that the user should go through the methods on `DialOpts`.

Is wrapping an enum into a struct an idiomatic way in Rust to signal that something should not be constructed directly? I don't recall seeing this in the past. Any prominent examples? I would have just relied on a doc comment on `DialOpts`.

swarm/src/lib.rs
mxinden: I am a bit reluctant to offer two ways to achieve the same thing. In what use-case would one want to call `condition` on `WithPeerIdWithAddresses` and couldn't do so on `WithPeerId`?

swarm/src/lib.rs
mxinden: Say we allowed `DialOpts::address(..)` instead of `DialOpts::unknown_peer_id(..).address(..)`, would that not confuse a user into thinking that they can then later on add a peer id, i.e. that they can do `DialOpts::address(..).peer_id(..)`?

> The only problem with this is that without a peer id you only have one address, whereas with an peer id there can be multiple.

Yes, this is something that I would like to encode at the type level, thus the differentiation between the two. I am not sure how to model your suggestion without loosing the type safety.

mxinden:
Thanks @elenaf9 for giving this a review and helping to design this interface. I replied to all of your comments. I am curious what you think.

swarm/src/lib.rs
elenaf9: > Is wrapping an enum into a struct an idiomatic way in Rust to signal that something should not be constructed directly? I don't recall seeing this in the past.

I would argue it the other way around: I feel like it is very rare and not rust-idiomatic that an enum is **not** constructed directly. Of course a user would notice that it is not possible once they try to build the inner structs, but it still may cause some confusion. With a struct, the API would make it clearer that it should be constructed via methods, especially when looking at it on docs.rs.

> I don't recall seeing this in the past. Any prominent examples?

I feel like this is just following the general api design pattern of wrapping internal/ inner types in a new struct to hide implementation details. Imo this always makes sense when the inner type is not relevant for the user and should never be constructed directly by them, which is the case here.

---
Alternatively, if you decide to fold the structs into the enum variants as you wrote in your TODO in l.1786:
> `// TODO:  How about folding these structs into the enum variants`

the user would be able to build the enum types directly and the builder pattern would not be needed at all.

swarm/src/lib.rs
elenaf9: I am also a bit undecided in this aspect. 
But the `DialOpts` are trying to follow the syntax of a builder pattern, and with a builder it normally doesn't matter in which order you specify the paramters, so you would expect that `DialOpts::peer_id(...).condition(..).addresses(..)` and `DialOpts::peer_id(...).addresses(..).condition(..)` both work. 
That being said, the current implementation doesn't strictly follow the builder pattern anyway (at least not the way it is described in the [Rust API Guidelines](https://rust-lang.github.io/api-guidelines/type-safety.html#builders-enable-construction-of-complex-values-c-builder)), therefore I also see the point of avoiding duplicated code.

swarm/src/lib.rs
elenaf9: > Say we allowed DialOpts::address(..) instead of DialOpts::unknown_peer_id(..).address(..), would that not confuse a user into thinking that they can then later on add a peer id, i.e. that they can do DialOpts::address(..).peer_id(..)?

I agree. But with reference to the above discussion: the pattern currently already expects a fixed order in which the parameters are specified, e.g. as written above: in case of a known peer id you have to specify the `condition` before specifying the `addresses`. Isn't that kind of the same issue?

> Yes, this is something that I would like to encode at the type level, thus the differentiation between the two. I am not sure how to model your suggestion without loosing the type safety.

You are right, the type safety should be kept; `DialOpts::address(..).peer_id(..)` should not be possible.

swarm/src/lib.rs
mxinden: Addressed in https://github.com/libp2p/rust-libp2p/pull/2317/commits/ad083e933a567f1f498e590f22a4022f0d09cfee.

swarm/src/lib.rs
mxinden: Addressed in https://github.com/libp2p/rust-libp2p/pull/2317/commits/26dedbec870842a5b3f9299894db14790b759777.

mxinden:
@elenaf9 do you want to take another look?

@thomaseizinger would very much appreciate your input here.

swarm/src/lib.rs
```diff
@@ -321,72 +323,158 @@ where
...
+    /// );
+    ///
+    /// // Dial an unknown peer.
+    /// swarm.dial(
+    ///   DialOpts::unknown_peer_id()
+    ///      .address("/ip6/::1/tcp/12345".parse().unwrap())
+    ///      .build()
+    /// );
+    /// ```
+    pub fn dial(&mut self, opts: DialOpts) -> Result<(), DialError> {
```
thomaseizinger: Given how common the following is:

```rust
DialOpts::unknown_peer_id().address(listen_addr).build()
```

We could consider accepting `impl Into<DialOpts>`:

```suggestion
    pub fn dial(&mut self, opts: impl Into<DialOpts>) -> Result<(), DialError> {
```

and implement:

```rust
impl From<MultiAddress> for DialOpts { }
```

where we do exactly the above.

swarm/src/dial_opts.rs
```diff
@@ -0,0 +1,205 @@
...
+pub struct WithPeerIdWithAddresses {
+    pub(crate) peer_id: PeerId,
+    pub(crate) condition: PeerCondition,
+    pub(crate) addresses: Vec<Multiaddr>,
+    pub(crate) extend_addresses_through_behaviour: bool,
+}
+
+impl WithPeerIdWithAddresses {
+    /// Specify a [`PeerCondition`] for the dial.
+    pub fn condition(mut self, condition: PeerCondition) -> Self {
```
thomaseizinger, line 147: This API requires users to import one more type: `PeerCondition`. Might be worth providing 3 functions instead that configure the corresponding condition.

Please excuse if that has been discussed and dismissed already!

thomaseizinger (approved):
Nice work!

Two suggestions but those could also be shipped later!

elenaf9 (approved):
Looks good to me!

swarm/src/lib.rs
kpp: Totally agree we can get rid of a boilerplate code like this. Another option is to accept an `impl DialOptsBuilder` and call: `.dial(DialOpts::.address(listen_addr))`.

swarm/src/dial_opts.rs
mxinden: Modelling 3 exclusive states with 3 methods instead of an `enum` seems not intuitive to me. Is the additional import of `PeerCondition` worth the added complexity?

swarm/src/lib.rs
mxinden: > Given how common the following is:
> 
> ```rust
> DialOpts::unknown_peer_id().address(listen_addr).build()
> ```
> 
> We could consider accepting `impl Into<DialOpts>`:
> 
> and implement:
> 
> ```rust
> impl From<MultiAddress> for DialOpts { }
> ```
> 
> where we do exactly the above.

Neat! Thanks for bringing it up @thomaseizinger. I will give this a shot.

mxinden: https://github.com/libp2p/rust-libp2p/pull/2317/commits/5e3edb3bf4c9dcbd5fb5fac8ecfd0b14b956d82b implements `From<Multiaddr>` and changes calls to `unknown_peer_id().address(xxx).build()` to the address only.



> Totally agree we can get rid of a boilerplate code like this. Another option is to accept an `impl DialOptsBuilder` and call: `.dial(DialOpts::.address(listen_addr))`.

I think `From<Multiaddr>` is the more elegant way. Given that the two solve the same issue, I would prefer the earlier over the later suggestion.

swarm/src/dial_opts.rs
thomaseizinger: I think there is a misunderstanding. I would still model the states internally with an enum. However, nothing stops anyone from doing this:

```rust
DialOpts::address(addr).condition(...).condition(...)
```

It might give people a hint that the condition is overridden but in theory, it could also be implemented in an additive way.
I don't feel strongly about it so happy to keep the enum :)

protocols/identify/src/identify.rs
```diff
@@ -743,8 +745,21 @@ mod tests {
...
+                    swarm2.select_next_some().await
+                {
+                    break peer_id;
+                }
+            }
+        });
+
+        // We should still be able to dial now!
+        swarm2
+            .dial(DialOpts::peer_id(swarm1_peer_id).build())
```
kpp: `impl From<PeerId> for DialOpts` too?

mxinden: :+1: good idea. Done via https://github.com/libp2p/rust-libp2p/pull/2317/commits/09293c4d9697d1d883dd69d935c481c2c8a2337c.

kpp (approved):

swarm/src/dial_opts.rs
mxinden: > I think there is a misunderstanding. I would still model the states internally with an enum.

No misunderstanding :)

>  I don't feel strongly about it so happy to keep the enum :)

Will keep the enum for now. Not 100% happy but can't think of a better alternative. We can still further iterate on it. Merging for now.

PR #2318, opened 2021-10-29
swarm/src/behaviour: Fix outdated comment on derive's out_event

mxinden:
The default behaviour changed with
https://github.com/libp2p/rust-libp2p/pull/2214. This updates the doc
comment.

Initially raised in
https://github.com/libp2p/rust-libp2p/pull/2290#issuecomment-944415562.

---

While this fixes the actual mismatch, I think it is missing an example in order for newcomers to actually be able to understand the concepts.

//CC @kpp


mxinden:
> Sorry that I missed this as part of the PR. If none beats me to it, I'll send an update in the next couple of days!

No worries! Just to make sure there is no duplicate work, this PR does the fix @thomaseizinger.

kpp (approved):

Issue #2259, opened 2021-10-01
core/src/peer_id: Support and output peer IDs as CIDs (RFC 0001)


mxinden:
rust-libp2p should implement [RFC 0001](https://github.com/libp2p/specs/blob/master/RFC/0001-text-peerid-cid.md).

Tracking issue: https://github.com/libp2p/specs/issues/216

Also tackle:

https://github.com/libp2p/rust-libp2p/blob/9f331e517f524dd112bf9e12b26d9dfae9162022/protocols/kad/src/protocol.rs#L102-L104

JerryHue:
@mxinden, I have spent a few hours reading over the several documents related to this. 

Would it be okay for me to work on this issue with some guidance?

mxinden:
Hi @JerryHue,

> Would it be okay for me to work on this issue

Yes for sure. Your help would be very much appreciated.

> with some guidance?

:+1: I don't know how familiar you are with the stack. Feel free to post questions here or via mail. Note that there is [a CID implementation in Rust](https://github.com/multiformats/rust-cid).

JerryHue:
Hello again @mxinden!

Despite my absence, I haven't forgotten about this issue! 

I believe I have implemented the changes as per the RFC stated, but I would like to make sure that I have not forgotten anything before I make my PR.

I would like to ask certain questions that have passed through my mind while working on this.

## Previous public methods

### Encoding to the string representation

As per the RFC requirements, the `PeerId` should still support the `base58btc` representation when printing it. This means that the `Debug` and `Display` traits use the `base58` representation.

To encode the `PeerId` to the `base32` representation, one may use `to_base32`.

### Decoding from the string representation

This one was tougher to separate, since we can only implement the `FromStr` once. At the end, I just had them in the same function. So, that means we can use the same function to parse the following strings:

* `bafzbeie5745rpv2m6tjyuugywy4d5ewrqgqqhfnf445he3omzpjbx5xqxe`,
* `QmYyQSo1c1Ym7orWxLYvCrM2EmxFTANf8wXmmE7DWjhx5N`,
* `12D3KooWD3eckifWpRn9wQpMG9R9hX3sD158z7EqHWmweQAJU5SA`.

I believe this fits the criteria of the RFC the best, and causes the least friction when upgrading.

### `to_bytes` and `from_bytes`

[...]

PR #2322, opened 2021-11-01
Change representation of PeerId to CIDv1

JerryHue:
Addresses #2259.

To encapsulate both errors from Multihash and CID, we change the interface of `from_bytes` to return a `PeerIdError` in the case of `Err`.

### DIFF
--- a/core/src/peer_id.rs
+++ b/core/src/peer_id.rs
@@ -394,4 +394,26 @@ mod tests {
 
         assert_eq!(cid_bytes, peer_id.to_cidv1_bytes());
     }
+
+    #[test]
+    fn peer_ids_from_base58_and_from_base32_are_equal() {
+        use std::str::FromStr;
+
+        let peer_id_as_base58_str = "QmYyQSo1c1Ym7orWxLYvCrM2EmxFTANf8wXmmE7DWjhx5N";
+        let peer_id_as_base32_str = "bafzbeie5745rpv2m6tjyuugywy4d5ewrqgqqhfnf445he3omzpjbx5xqxe";
+
+        let peer_id_from_base58_str = PeerId::from_str(peer_id_as_base58_str).unwrap();
+        let peer_id_from_base32_str = PeerId::from_str(peer_id_as_base32_str).unwrap();
+
+        assert_eq!(peer_id_from_base58_str, peer_id_from_base32_str);
+    }
+
+    #[test]
+    fn peer_id_from_base58_str_then_back() {
+        use std::str::FromStr;
+
+        let cidv1_peer_id = identity::Keypair::generate_ed25519().public().to_peer_id();
+
+        assert_eq!(cidv1_peer_id, PeerId::from_str(&cidv1_peer_id.to_base58()).unwrap());
+    }
 }
--- a/core/src/peer_id.rs
+++ b/core/src/peer_id.rs
@@ -168,7 +168,7 @@ impl PeerId {
         bs58::encode(self.to_bytes()).into_string()
     }
 
-    /// Returns a `base32` encoded string of this `PeerId`.
+    /// Returns a `base32` multibase encoded string of this `PeerId`.
     pub fn to_base32(&self) -> String {
         self.cid.to_string_of_base(Base::Base32Lower).unwrap()
     }
--- a/core/src/peer_id.rs
+++ b/core/src/peer_id.rs
@@ -114,18 +114,22 @@ impl PeerId {
 
     /// Tries to turn a `Cid` into a `PeerId`.
     ///
-    /// If the CID's multihash does not fit the specifications:
-    ///
-    /// * use a valid hashing algorithm for peer IDs, or
-    /// * the value does not satisfy the constraints for a hashed peerID,
-    ///
-    /// or, if the CID does not fit the specifications:
-    ///
-    /// * is a CIDv0 with a multihash that fits the specifications
-    /// for multihash, or
-    /// * is a CIDv1 with the proper codec (LIBP2P_KEY_CODEC),
-    ///
-    /// then the CID will be returned as an `Err`.
+    /// If the `Cid` is valid, then a `Some(PeerId)` will be
+    /// returned, otherwise, `Err(Cid)` is returned.
+    /// 
+    /// A `Cid` is considered valid if and only if:
+    /// * the `Cid` is a [`Cidv0`] with a valid `multihash`, or
+    /// * the `Cid` is a [`Cidv1`] with:
+    ///  1. a valid `multihash`, and
+    ///  2. the codec for [`libp2p-key`].
+    /// 
+    /// A `multihash` is considered valid if and only if:
+    /// 1. it uses a valid hashing algorithm for peer IDs, and
+    /// 2. it satisfies the constraints for a hashed peerID.
+    /// 
+    /// [`Cidv0`]: https://github.com/multiformats/cid#cidv0
+    /// [`Cidv1`]: https://github.com/multiformats/cid#cidv1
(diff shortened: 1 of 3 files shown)


core/src/peer_id.rs
```diff
@@ -77,44 +102,86 @@ impl PeerId {
...
+            }),
             _ => Err(multihash),
         }
     }
 
+    /// Tries to turn a `Cid` into a `PeerId`.
+    ///
+    /// If the CID's multihash does not fit the specifications:
+    ///
+    /// * use a valid hashing algorithm for peer IDs, or
```
```suggestion
    /// * use an  invalid hashing algorithm for peer IDs, or
```

Correct?

core/src/peer_id.rs
```diff
@@ -77,44 +102,86 @@ impl PeerId {
...
+    /// * the value does not satisfy the constraints for a hashed peerID,
+    ///
+    /// or, if the CID does not fit the specifications:
+    ///
+    /// * is a CIDv0 with a multihash that fits the specifications
+    /// for multihash, or
+    /// * is a CIDv1 with the proper codec (LIBP2P_KEY_CODEC),
```
mxinden: I am confused by this. Shouldn't the opposite be written, along the lines of: "If XXX does not fit the specification, ..."?

core/src/peer_id.rs
```diff
@@ -77,44 +102,86 @@ impl PeerId {
...
     pub fn to_bytes(&self) -> Vec<u8> {
-        self.multihash.to_bytes()
+        self.cid.hash().to_bytes()
     }
 
-    /// Returns a base-58 encoded string of this `PeerId`.
+    /// Returns a raw bytes representation of the `PeerId`.
+    pub fn to_cidv1_bytes(&self) -> Vec<u8> {
+        self.cid.to_bytes()
+    }
```
mxinden: Is this needed anywhere? If not, I would prefer removing it to ease future breaking changes. If I am not mistaken test logic should be able to call `self.cid.to_bytes` directly.

core/src/peer_id.rs
```diff
@@ -77,44 +102,86 @@ impl PeerId {
...
+
+    /// Returns a base-58 raw encoded string of this `PeerId`.
     pub fn to_base58(&self) -> String {
         bs58::encode(self.to_bytes()).into_string()
     }
 
+    /// Returns a `base32` encoded string of this `PeerId`.
```
```suggestion
    /// Returns a `base32` multibase encoded string of this `PeerId`.
```

core/src/peer_id.rs
```diff
@@ -77,44 +102,86 @@ impl PeerId {
...
+        self.cid.hash().to_bytes()
     }
 
-    /// Returns a base-58 encoded string of this `PeerId`.
+    /// Returns a raw bytes representation of the `PeerId`.
+    pub fn to_cidv1_bytes(&self) -> Vec<u8> {
+        self.cid.to_bytes()
+    }
+
+    /// Returns a base-58 raw encoded string of this `PeerId`.
```
mxinden, line 170:
```suggestion
    /// Returns a base-58 raw (non-multibase) encoded string of this `PeerId`.
```

mxinden, line 170:
```suggestion
    /// Returns a base-58 raw encoded string of this `PeerId`.
    ///
    /// This corresponds to the old peer ID [representation].
    ///
    /// [representation]: https://github.com/libp2p/specs/blob/master/RFC/0001-text-peerid-cid.md
```

core/src/peer_id.rs
```suggestion
    /// Returns a `base32` encoded string of this `PeerId`.
    ///
    /// This corresponds to the new peer ID [representation].
    ///
    /// [representation]: https://github.com/libp2p/specs/blob/master/RFC/0001-text-peerid-cid.md
```

core/src/peer_id.rs
```diff
@@ -77,44 +102,86 @@ impl PeerId {
...
     pub fn to_base58(&self) -> String {
         bs58::encode(self.to_bytes()).into_string()
```
mxinden, line 176: The following would fail given that `<PeerId as FromStr>` expects the string to start with `Qm` or `1`, right?

```rust
assert!(cidv1_peer_id == PeerId::from_str(cidv1_peer_id.to_base58()));
```

I am not quite sure how to proceed in this case. Maybe convert the CIDv1 peer ID to the old format first?

core/src/peer_id.rs
```diff
@@ -164,21 +231,68 @@ impl From<PeerId> for Vec<u8> {
...
 
+#[derive(Debug)]
+pub enum PeerIdError {
+    // Variant to group errors produced by `multihash`
+    MultihashError(MhError),
+    // Variant to group errors produced by `cid`
+    CidError(CidError),
+}
+
+impl fmt::Display for PeerIdError {
```
mxinden, line 250: `libp2p-core` already imports `thiserror`. What do you think of using it instead of this manual implementation?

mxinden (commented):
First off, very solid work! Thanks @JerryHue.

I have a couple of comments below:

core/src/peer_id.rs
```diff
@@ -28,12 +30,15 @@ use thiserror::Error;
...
 const MAX_INLINE_KEY_LENGTH: usize = 42;
 
+/// CIDv1 MUST use the libp2p key codec to identify the content of the CID.
+const LIBP2P_KEY_CODEC: u64 = 0x72;
+
 /// Identifier of a peer of the network.
 ///
-/// The data is a multihash of the public key of the peer.
+/// The data is a cid of the public key of the peer.
 #[derive(Clone, Copy, Eq, Hash, Ord, PartialEq, PartialOrd)]
```
mxinden, line 39: If I am not mistaken the implementation of `Eq` is simply derived on both `PeerId` and `Cid`. Given that we store all `PeerID`s as CIDv1s now, an old style `PeerId` and a new style `PeerId` both parsed from a `String` would still equal, correct? Could you add a test to make sure this is the case going forward?

core/src/peer_id.rs
mxinden: > Maybe convert the CIDv1 peer ID to the old format first?

What do you think @JerryHue?

mxinden:
Friendly ping @JerryHue. Are you still interested in taking this fix to the finish line? :)

JerryHue:
Hello again, @mxinden! I would like to try to bring this to the finish line throughout this week, sorry for the excruciatingly long wait. I wonder if I should improve this PR by taking into account @tomaka's comments on the original issue. What do you think?

thomaseizinger:
@mxinden I believe this is waiting for your input.

tomaka:
This PR would break compatibility in Substrate/Polkadot and is a no-no for us.

tomaka:
This PR would break compatibility with Substrate/Polkadot and is a no-no for us.

thomaseizinger:
> This PR would break compatibility with Substrate/Polkadot and is a no-no for us.

Thanks for mentioning this. This definitely needs an upgrade path where we first support reading in a CIDv1 `PeerId` as described in the RFC.

tomaka:
For what it's worth, I've taken the decision for Polkadot a while ago to always use CIDv0. The spec precisely mentions how to encode it: https://spec.polkadot.network/#defn-peer-id. Being able to easily encode/decode PeerIds, and ensuring uniqueness of PeerIds, are more important than conforming to some libp2p spec outside of our control and that brings zero benefits for us.

It's not a deal breaker for adding support for CIDv1 in rust-libp2p, as long as there's always an option to force CIDv0.

mxinden:
> @mxinden I believe this is waiting for your input.

As far as I can tell blocked on addressing comments above.

> It's not a deal breaker for adding support for CIDv1 in rust-libp2p, as long as there's always an option to force CIDv0.

I suggest we support reading CIDv1 as a first iteration, though always writing and handling CIDv0s.

PR #2325, opened 2021-11-03
update topic_peers when peer_topics is updated

divagant-martian:
Small piece of code where we update peer_topics but not topic_peers

### DIFF
--- a/protocols/gossipsub/CHANGELOG.md
+++ b/protocols/gossipsub/CHANGELOG.md
@@ -1,6 +1,6 @@
 # 0.33.1 [unreleased]
 
-- Fix bug in internal peer's topics tracking (see [PR 2325])
+- Fix bug in internal peer's topics tracking (see [PR 2325]).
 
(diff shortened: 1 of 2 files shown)


AgeManning (approved):
Thanks, looks like this was missed in an earlier revision

mxinden (commented):
Could you bump the patch version and include a changelog entry?

divagant-martian:
> Could you bump the patch version and include a changelog entry?

Is the patch version bump required? There is an unreleased entry already in the Changelog

protocols/gossipsub/CHANGELOG.md
```diff
@@ -1,8 +1,11 @@
 # 0.33.1 [unreleased]
 
+- Fix bug in internal peer's topics tracking (see [PR 2325])
```
```suggestion
- Fix bug in internal peer's topics tracking (see [PR 2325]).
```

mxinden (approved):
> Is the patch version bump required? There is an unreleased entry already in the Changelog

Ah, yes, sorry, not required.

PR #2327, opened 2021-11-05
protocols/gossipsub: Improve bandwidth

AgeManning:
This PR adds some bandwidth improvements to gossipsub. 

After a bit of inspection on live networks a number of improvements have been made that can help reduce unnecessary bandwidth on gossipsub networks. This PR introduces the following:
- A 1:1 tracking of all in-flight IWANT requests. This not only ensures that all IWANT requests are answered and peers penalized accordingly, but gossipsub will no no longer create multiple IWANT requests for multiple peers. Previously, gossipsub sampled the in-flight IWANT requests in order to penalize peers for not responding with a high probability that we would detect non-responsive nodes. Futher, it was possible to re-request IWANT messages that are already being requested causing added duplication in messages and wasted unnecessary IWANT control messages. This PR shifts this logic to only request message ids that we are not currently requesting from peers. 
- Triangle routing naturally gives rise to unnecessary duplicates. Consider a mesh of 4 peers that are interconnected. Peer 1 sends a new message to 2,3,4. 2 propagates to 3,4 and 3 propagates to 2,4 and 4 propagates to 2,3. In this case 3 has received the message 3 times. If we keep track of peers that send us messages, when publishing or forwarding we no longer send to peers that have sent us a duplicate, we can eliminate one of the sends in the scenario above. This only occurs when message validation is async however. This PR adds this logic to remove some elements of triangle-routing duplicates.


protocols/gossipsub/src/behaviour/tests.rs
```diff
@@ -4688,79 +4690,76 @@ mod tests {
...
         }
 
-        //add 100 more peers
+        // add 100 more peers
         let other_peers: Vec<_> = (0..100)
             .map(|_| add_peer(&mut gs, &topics, false, false))
             .collect();
 
-        //each peer sends us two ihave containing each two message ids
+        // each peer sends us two ihave containing each two message ids
```
divagant-martian:
```suggestion
        // each peer sends us an ihave containing each two message ids
```

protocols/gossipsub/src/behaviour/tests.rs
```diff
@@ -4688,79 +4690,76 @@ mod tests {
...
         sleep(Duration::from_secs(4));
 
-        //now we do a heartbeat to apply penalization
+        // now we do a heartbeat to apply penalization
         gs.heartbeat();
 
-        //now we get all the second messages
-        for message in second_messages {
-            gs.handle_received_message(message.clone(), &PeerId::random());
+        // now we get all the second messages
```
divagant-martian:
```suggestion
        // now we get the second messages from the last 80 peers.
```

protocols/gossipsub/src/behaviour.rs
```diff
@@ -1608,13 +1639,26 @@ where
...
+
         if let Some(fast_message_id) = fast_message_id.as_ref() {
-            if let Some(msg_id) = self.fast_messsage_id_cache.get(fast_message_id) {
+            if let Some(msg_id) = self.fast_message_id_cache.get(fast_message_id) {
                 let msg_id = msg_id.clone();
-                self.message_is_valid(&msg_id, &mut raw_message, propagation_source);
-                if let Some((peer_score, ..)) = &mut self.peer_score {
-                    peer_score.duplicated_message(propagation_source, &msg_id, &raw_message.topic);
+                // Report the duplicate
+                if self.message_is_valid(&msg_id, &mut raw_message, propagation_source) {
```
divagant-martian, line 1734: Why just report if the message is valid now?

divagant-martian (changes requested):

protocols/gossipsub/src/behaviour.rs
AgeManning: I think this was a bug in the previous code. In principle we shouldn't have invalid messages in the duplicates, and if we do, we penalize them but we don't register it as a duplicate (for scoring). The duplicate would benefit the peer, but we dont want to improve scores for peers that are sending us invalid messages.

protocols/gossipsub/CHANGELOG.md
```diff
@@ -1,11 +1,15 @@
 # 0.33.1 [unreleased]
 
+- Improve bandwidth performance by tracking IWANTs and reducing duplicate sends
+(see PR 2327)
```
```suggestion
- Improve bandwidth performance by tracking IWANTs and reducing duplicate sends
  (see PR 2327).
```

protocols/gossipsub/src/mcache.rs
```diff
@@ -68,30 +71,44 @@ impl MessageCache {
...
+        {
             // Don't add duplicate entries to the cache.
             self.history[0].push(cache_entry);
+            return true;
+        } else {
+            return false;
+        }
+    }
+
+    /// Keeps track of peers we know have received the message to prevent forwarding peers
```
mxinden:
```suggestion
    /// Keeps track of peers we know have received the message to prevent forwarding to said peers.
```

mxinden (approved):
Looks good to me. Anything else left from your side?

divagant-martian (approved):

AgeManning:
Yep. Just want to run a bit more and run with the new metrics to quantify the gains here. Will report back once tests are complete

AgeManning:
@mxinden - Have confirmed this has a net positive effect and works as expected. 
Could you review and maybe we can get this merged down.

protocols/gossipsub/CHANGELOG.md
```diff
@@ -4,7 +4,11 @@
 
 - Migrate to Rust edition 2021 (see [PR 2339]).
 
+- Improve bandwidth performance by tracking IWANTs and reducing duplicate sends
+  (see PR 2327).
```
```suggestion
  (see [PR 2327]).
```

protocols/gossipsub/src/behaviour.rs
```diff
@@ -1211,25 +1234,33 @@ where
...
 
-            let message_ids = iwant_ids_vec.into_iter().cloned().collect::<Vec<_>>();
+            let message_ids = iwant_ids_vec
+                .into_iter()
+                .map(|id| {
+                    // Add all messages to the pending list
+                    self.pending_iwant_msgs.insert(id.clone());
+                    id.clone()
+                })
+                .collect::<Vec<_>>();
```
mxinden: Nit pick: A side effect (`self.pending_iwant_msgs`) in a `map` is a bit surprising to me. Would an additional imperative `for msg in message_ids` be ok as well? I would guess this does not have any impact on the performance.

mxinden (commented):
Just two small comments. Otherwise looks good to me.

protocols/gossipsub/src/behaviour.rs
AgeManning: I've re-written it with a for loop. Just makes the `message_ids` vec mutable. Just wanted to do the clone and the copy in the one iteration :p

AgeManning:
Hopefully this guy is ready now :)

mxinden (approved):
Thanks for the follow-ups!

PR #2332, opened 2021-11-09
Allow setting custom protocol in mplex

### LINKED ISSUE #2344: [Feature Request] Allow setting custom protocol in mplex
in [go-libp2p-muxer](https://github.com/libp2p/go-libp2p/blob/master/options.go#L101), it allows using custom protocol name. Adding a similar pub fn set_protocol to MplexConfig to connect to go nodes with custom protocol.  

```go
// Muxer configures libp2p to use the given stream multiplexer (or stream
// multiplexer constructor).
//
// Name is the protocol name.
//
// The transport can be a constructed mux.Transport or a function taking any
// subset of this libp2p node's:
// * Peer ID
// * Host
// * Network
// * Peerstore
func Muxer(name string, tpt interface{}) Option {
	mtpt, err := config.MuxerConstructor(tpt)
	err = traceError(err, 1)
	return func(cfg *Config) error {
		if err != nil {
			return err
		}
		cfg.Muxers = append(cfg.Muxers, config.MsMuxC{MuxC: mtpt, ID: name})
		return nil
	}
}
```

hanabi1224:
in [go-libp2p-muxer](https://github.com/libp2p/go-libp2p/blob/master/options.go#L101), it allows using custom protocol name.  Adding a similar `pub fn set_protocol` to `MplexConfig`  to connect to go nodes with custom protocol

```go
// Muxer configures libp2p to use the given stream multiplexer (or stream
// multiplexer constructor).
//
// Name is the protocol name.
//
// The transport can be a constructed mux.Transport or a function taking any
// subset of this libp2p node's:
// * Peer ID
// * Host
// * Network
// * Peerstore
func Muxer(name string, tpt interface{}) Option {
	mtpt, err := config.MuxerConstructor(tpt)
	err = traceError(err, 1)
	return func(cfg *Config) error {
		if err != nil {
			return err
		}
		cfg.Muxers = append(cfg.Muxers, config.MsMuxC{MuxC: mtpt, ID: name})
		return nil
	}
}
```

Closes https://github.com/libp2p/rust-libp2p/issues/2344

### DIFF
--- a/muxers/mplex/CHANGELOG.md
+++ b/muxers/mplex/CHANGELOG.md
@@ -1,3 +1,7 @@
+# unreleased
+
+- Add `set_protocol_name` to MplexConfig
+
 # 0.30.0 [2021-11-01]
 
 - Make default features of `libp2p-core` optional.
--- a/muxers/mplex/CHANGELOG.md
+++ b/muxers/mplex/CHANGELOG.md
@@ -1,7 +1,6 @@
-# unreleased
+# 0.30.1
 
 - Add `fn set_protocol_name(&mut self, protocol_name: &'static [u8])` to MplexConfig
-- Add `fn protocol_name(&self)` to MplexConfig
 
 # 0.30.0 [2021-11-01]
 
--- a/muxers/mplex/CHANGELOG.md
+++ b/muxers/mplex/CHANGELOG.md
@@ -1,4 +1,4 @@
-# 0.30.1
+# 0.30.1 [unreleased]
 
 - Add `fn set_protocol_name(&mut self, protocol_name: &'static [u8])` to MplexConfig
 
--- a/muxers/mplex/CHANGELOG.md
+++ b/muxers/mplex/CHANGELOG.md
@@ -1,6 +1,7 @@
 # unreleased
 
-- Add `set_protocol_name` to MplexConfig
+- Add `fn set_protocol_name(&mut self, protocol_name: &'static [u8])` to MplexConfig
+- Add `fn protocol_name(&self)` to MplexConfig
 
 # 0.30.0 [2021-11-01]
 
--- a/muxers/mplex/Cargo.toml
+++ b/muxers/mplex/Cargo.toml
@@ -2,7 +2,7 @@
 name = "libp2p-mplex"
 edition = "2018"
 description = "Mplex multiplexing protocol for libp2p"
-version = "0.30.0"
+version = "0.31.0"
 authors = ["Parity Technologies <admin@parity.io>"]
 license = "MIT"
 repository = "https://github.com/libp2p/rust-libp2p"
--- a/muxers/mplex/Cargo.toml
+++ b/muxers/mplex/Cargo.toml
@@ -2,7 +2,7 @@
 name = "libp2p-mplex"
 edition = "2018"
 description = "Mplex multiplexing protocol for libp2p"
-version = "0.31.0"
+version = "0.30.1"
 authors = ["Parity Technologies <admin@parity.io>"]
 license = "MIT"
 repository = "https://github.com/libp2p/rust-libp2p"
--- a/muxers/mplex/src/config.rs
+++ b/muxers/mplex/src/config.rs
@@ -21,6 +21,8 @@
 use crate::codec::MAX_FRAME_SIZE;
 use std::cmp;
 
+pub(crate) const DEFAULT_MPLEX_PROTOCOL: &'static [u8] = b"/mplex/6.7.0";
+
 /// Configuration for the multiplexer.
 #[derive(Debug, Clone)]
 pub struct MplexConfig {
@@ -33,6 +35,8 @@ pub struct MplexConfig {
     /// When sending data, split it into frames whose maximum size is this value
     /// (max 1MByte, as per the Mplex spec).
     pub(crate) split_send_size: usize,
+    /// Protocol defaults to b"/mplex/6.7.0"
+    pub(crate) protocol: &'static [u8],
 }
 
 impl MplexConfig {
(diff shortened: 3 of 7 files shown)


muxers/mplex/src/config.rs
```diff
@@ -84,6 +88,17 @@ impl MplexConfig {
         self.split_send_size = size;
         self
     }
+
+    /// Gets protocol
+    pub fn protocol(&self) -> &[u8] {
+        self.protocol
+    }
```
mxinden: None of the other fields have a _getter_ method. Why is this one needed?

muxers/mplex/src/config.rs
```diff
@@ -21,6 +21,8 @@
 use crate::codec::MAX_FRAME_SIZE;
 use std::cmp;
 
+pub(crate) const DEFAULT_MPLEX_PROTOCOL: &'static [u8] = b"/mplex/6.7.0";
```
mxinden:
```suggestion
pub(crate) const DEFAULT_MPLEX_PROTOCOL_NAME: &'static [u8] = b"/mplex/6.7.0";
```

How about referring to this as the protocol **name**?

muxers/mplex/src/config.rs
```diff
@@ -84,6 +88,17 @@ impl MplexConfig {
...
         self
     }
+
+    /// Gets protocol
+    pub fn protocol(&self) -> &[u8] {
+        self.protocol
+    }
+
+    /// Sets protocol, muxer.set_protocol(b"/mplex/6.7.0")
+    pub fn set_protocol(&mut self, protocol: &'static [u8]) -> &mut Self {
```
mxinden:
```suggestion
    pub fn set_protocol_name(&mut self, protocol: &'static [u8]) -> &mut Self {
```

muxers/mplex/src/config.rs
```diff
@@ -84,6 +88,17 @@ impl MplexConfig {
         self.split_send_size = size;
         self
     }
+
+    /// Gets protocol
+    pub fn protocol(&self) -> &[u8] {
+        self.protocol
+    }
+
+    /// Sets protocol, muxer.set_protocol(b"/mplex/6.7.0")
```
mxinden:
```suggestion
    /// Set protocol name.
    ///
    /// ```rust
    /// muxer_config.set_protocol(b"/mplex/6.7.0");
    /// ```
```

muxers/mplex/src/config.rs
```diff
@@ -120,6 +135,7 @@ impl Default for MplexConfig {
             max_buffer_len: 32,
             max_buffer_behaviour: MaxBufferBehaviour::Block,
             split_send_size: 8 * 1024,
+            protocol: DEFAULT_MPLEX_PROTOCOL,
```
mxinden:
```suggestion
            protocol_name: DEFAULT_MPLEX_PROTOCOL,
```

mxinden (commented):
Could you include a changelog entry and bump the version?

https://github.com/libp2p/rust-libp2p/blob/master/docs/release.md

muxers/mplex/src/config.rs
hanabi1224: I believe it's useful for debugging/logging purposes to be able to retrieve the protocol name when it can be customized

hanabi1224:
@mxinden Thanks for your review, I've updated the PR, please take another look.

muxers/mplex/src/config.rs
```diff
@@ -84,6 +88,23 @@ impl MplexConfig {
         self.split_send_size = size;
         self
     }
+
+    /// Gets the protocol anme
+    pub fn protocol_name(&self) -> &[u8] {
+        self.protocol_name
+    }
```
mxinden: Sorry for the bringing this up once more, still, for the sake of consistency, I don't think we should introduce a getter for a single field only.

In case you would like to introduce getters for all fields, I think that should be a separate pull request.

muxers/mplex/Cargo.toml
```diff
@@ -2,7 +2,7 @@
 name = "libp2p-mplex"
 edition = "2018"
 description = "Mplex multiplexing protocol for libp2p"
-version = "0.30.0"
+version = "0.31.0"
```
mxinden:
```suggestion
version = "0.30.1"
```

Is this a breaking change? If not, why not release it as a patch release?

muxers/mplex/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+# unreleased
```
```suggestion
# 0.30.1 [unreleased]
```

hanabi1224:
@mxinden  Comments resolved, plz take another look thx!

muxers/mplex/src/config.rs
hanabi1224: Removed as suggested. Although I don't agree with the point of maintaining consistency. I think it makes sense to have immutable getters only for a selective set of the fields, with the same rationale behind having different access modifiers for different fields in many languages including rust, but I do agree this might belong to a separate PR, hope this addresses ur concern in this PR, thx!

hanabi1224: Removed as suggested. 

Although I don't really agree with the point of maintaining consistency. I think it makes sense to have immutable getters only for a subset of the fields, sharing the same rationale behind having different access modifiers for different fields in many languages including rust, but I do agree this might belong to a separate PR.

Hope this addresses ur concern in this PR, thx!

muxers/mplex/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# 0.30.1
```
```suggestion
# 0.30.1 [unreleased]
```

mxinden (approved):
Thanks @hanabi1224.

hanabi1224:
Thanks!

PR #2338, opened 2021-11-12
protocols/identify Check multiaddr has valid peer id component prior to caching

### LINKED ISSUE #2302: protocols/identify: Validate addresses before caching
With https://github.com/libp2p/rust-libp2p/pull/2232 `libp2p-identify` caches discovered listen addresses to provide additional addresses when dialing a peer via `NetworkBehaviour::addresses_of_peer`.

Before caching received listen addresses, `libp2p-identify` should validate these addresses. Validations:

- If the address contains a `/p2p/QmXXX` make sure `QmXX` matches the peer's `PeerId`.
- ...

//CC @AgeManning: This might be the reason for the dial failures you are seeing.

MarcoPolo:
## Context
Resolves https://github.com/libp2p/rust-libp2p/issues/2302. 

https://github.com/libp2p/rust-libp2p/pull/2232 introduced a change to the [Identify protocol](https://github.com/libp2p/specs/blob/master/identify/README.md) implementation that caches the discovered `listenAddrs` from the [Identify message](https://github.com/libp2p/specs/blob/master/identify/README.md#the-identify-message). The cache is then used in the [addresses_of_peer](https://docs.rs/libp2p/0.40.0/libp2p/swarm/trait.NetworkBehaviour.html#method.addresses_of_peer) method that is called by swarm code to find possible addresses to dial for a given peer.

This lets the Identify assist in peer discovery of a given [network behavior](https://docs.rs/libp2p/0.40.0/libp2p/swarm/trait.NetworkBehaviour.html) automatically. See https://github.com/libp2p/rust-libp2p/issues/2216 for more information.

This cache was disabled in https://github.com/libp2p/rust-libp2p/pull/2312 because we weren't doing any validation. This PR adds the peer id validation.

## Proposed Solution
This adds a simple validation before we cache the multiaddr. We don't cache a multiaddr with a wrong final peer-id component. e.g. something that ends with `.../p2p/Qmsome-other-peer`. While this should fail to dial because we [check this before dialing](https://github.com/libp2p/rust-libp2p/blob/master/core/src/connection/pool/concurrent_dial.rs#L144), it's probably better to not cache this in the first place.

If the multiaddr doesn't have a final `.../p2p/Qmfoo` component, we consider it valid. 

Testing this locally, it seems like all `listenAddrs` in Identify messages generally don't include the `.../p2p/` component (except for [dnsaddr](https://github.com/multiformats/multiaddr/blob/master/protocols/DNSADDR.md), but do these even show up in identify messages?)

## Rollback strategy
This is safe to revert. Without this, multiaddrs with a wrong peerid will [fail to dial](https://github.com/libp2p/rust-libp2p/blob/master/core/src/connection/pool/concurrent_dial.rs#L152).

## Next steps

This doesn't enable the caching, I can do that in this PR or another one if we want to keep things smaller.

### DIFF
--- a/protocols/identify/src/identify.rs
+++ b/protocols/identify/src/identify.rs
@@ -23,6 +23,7 @@ use crate::protocol::{IdentifyInfo, ReplySubstream};
 use futures::prelude::*;
 use libp2p_core::{
     connection::{ConnectionId, ListenerId},
+    multiaddr::Protocol,
     upgrade::UpgradeError,
     ConnectedPoint, Multiaddr, PeerId, PublicKey,
 };
@@ -306,8 +307,14 @@ impl NetworkBehaviour for Identify {
         match event {
             IdentifyHandlerEvent::Identified(info) => {
                 // Replace existing addresses to prevent other peer from filling up our memory.
+                let valid_listen_addrs = info
+                    .listen_addrs
+                    .clone()
+                    .into_iter()
+                    .filter(|addr| multiaddr_matches_peer_id(addr, &peer_id));
+
                 self.discovered_peers
-                    .put(peer_id, HashSet::from_iter(info.listen_addrs.clone()));
+                    .put(peer_id, HashSet::from_iter(valid_listen_addrs));
 
                 let observed = info.observed_addr.clone();
                 self.events.push_back(NetworkBehaviourAction::GenerateEvent(
@@ -497,6 +504,16 @@ fn listen_addrs(params: &impl PollParameters) -> Vec<Multiaddr> {
     let mut listen_addrs: Vec<_> = params.external_addresses().map(|r| r.addr).collect();
     listen_addrs.extend(params.listened_addresses());
     listen_addrs
+}
+
+/// If there is a given peer_id in the multiaddr, make sure it is the same as
+/// the given peer_id. If there is no peer_id for the peer in the mutiaddr, this returns true.
+fn multiaddr_matches_peer_id(addr: &Multiaddr, peer_id: &PeerId) -> bool {
+    let last_component = addr.iter().last();
+    if let Some(Protocol::P2p(multi_addr_peer_id)) = last_component {
+        return multi_addr_peer_id == *peer_id.as_ref();
+    }
+    return true;
 }
 
 #[cfg(test)]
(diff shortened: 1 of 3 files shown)


MarcoPolo:
tagging @mxinden here on GH as well

protocols/identify/src/identify.rs
```diff
@@ -758,4 +775,29 @@ mod tests {
...
+            .parse()
+            .expect("failed to parse multiaddr");
+
+        let addr_without_peer_id: Multiaddr = addr.clone();
+        let mut addr_with_other_peer_id = addr.clone();
+
+        addr.push(Protocol::P2p(peer_id.clone().into()));
+        addr_with_other_peer_id.push(Protocol::P2p(other_peer_id.into()));
+
+        assert_eq!(true, multiaddr_matches_peer_id(&addr, &peer_id));
```
mxinden:
```suggestion
        assert!(multiaddr_matches_peer_id(&addr, &peer_id));
```

Small nit pick. Same for the below.

protocols/identify/src/identify.rs
```diff
@@ -758,4 +775,29 @@ mod tests {
 
         assert_eq!(connected_peer, swarm1_peer_id);
     }
+
+    #[test]
+    fn check_multiaddr_matches_peer_id() {
+        let peer_id = PeerId::random();
+        let other_peer_id = PeerId::random();
+        let mut addr: Multiaddr = ("/ip4/147.75.69.143/tcp/4001".to_string())
```
mxinden:
```suggestion
        let mut addr: Multiaddr = "/ip4/147.75.69.143/tcp/4001".to_string()
```

Are these needed?

mxinden:
```suggestion
        let mut addr: Multiaddr = ("/ip4/147.75.69.143/tcp/4001")
```

`.parse` should work on `&str` directly, no?

protocols/identify/src/identify.rs
```diff
@@ -304,8 +305,14 @@ impl NetworkBehaviour for Identify {
         match event {
             IdentifyHandlerEvent::Identified(info) => {
                 // Replace existing addresses to prevent other peer from filling up our memory.
+                let valid_listen_addrs = info
+                    .listen_addrs
+                    .clone()
```
mxinden: I don't see a reason why a `Multiaddr` that does not fulfill `multiaddr_matches_peer_id` would ever be valid in this context.

With that in mind, how about removing the invalid `Multiaddr`s from `info` directly, thus both (a) only putting valid addresses into `self.discovered_peers` and (b) only emitting valid addresses via `IdentifyEvent::Received`?

mxinden (commented):
:rocket: couple of minor comments. Other than that this looks good to me. Thanks for the help.

Good job getting started so quickly!

mxinden:
> except for dnsaddr, but do these even show up in identify messages?

Yes, a node could advertise a dnsaddr as their listen address.

protocols/identify/src/identify.rs
MarcoPolo: yup! I was doing something else, and didn't rollback far enough 🤦

mxinden:
Thanks for the follow-up. Just missing a changelog entry (see [release docs](https://github.com/libp2p/rust-libp2p/blob/master/docs/release.md)).

MarcoPolo:
Done, should I also enable this by default or do that in another PR? (essentially undoing this PR: https://github.com/libp2p/rust-libp2p/pull/2312)

mxinden (approved):
Thanks @MarcoPolo.

> Done, should I also enable this by default or do that in another PR? (essentially undoing this PR: #2312)

I think it is fine for now. I would suggest enabling it by default once people had the chance to test it out on a larger deployment.

PR #2339, opened 2021-11-14
*: Migrate to Rust 2021 edition

nuke-web3:
Confirmed to build locally on Ubuntu 20.04.3 LTS:
```
rustc 1.56.1 (59eed8a2a 2021-11-01)
rustc 1.58.0-nightly (e90c5fbbc 2021-11-12)
```


mxinden:
Thanks @NukeManDan for the upgrade work!

> The most important rule for editions is that crates in one edition can interoperate seamlessly with crates compiled in other editions. This ensures that the decision to migrate to a newer edition is a "private one" that the crate can make without affecting others.

https://blog.rust-lang.org/2021/05/11/edition-2021.html

If I am not mistaken there is no downside for rust-libp2p to upgrade to edition 2021. Anyone objecting?

Not sure what the best practice is, but I would still consider an upgrade to 2021 to be a breaking change. Thus this pull request would need to add a changelog entry to each crate's `CHANGELOG.md` and bump the version if not already done (look for `[unreleased]` in `CHANGELOG.md`). See [release docs](https://github.com/libp2p/rust-libp2p/blob/master/docs/release.md) for details.

thomaseizinger:
> If I am not mistaken there is no downside for rust-libp2p to upgrade to edition 2021. Anyone objecting?

AFAIK it will bump the MSRV to 1.56. Are we okay with that?

nuke-web3:
It is also strongly suggested that we included a _minimal_ rust version to use, specifically 1.56.1

`Cargo.toml` should include:
```toml
rust-version = "1.56.1"
```

https://doc.rust-lang.org/cargo/reference/manifest.html#the-rust-version-field
> it cannot include semver operators or pre-release identifiers. 
> The rust-version must be equal to or newer than the version that first introduced the configured edition.
> The rust-version may be ignored using the --ignore-rust-version option.

More discussion: 
- https://github.com/paritytech/parity-wasm/pull/322#issuecomment-968709073
- https://github.com/Tamschi/rust-template/issues/28

mxinden:
> AFAIK it will bump the MSRV to 1.56. Are we okay with that?

I am fine with it. Anyone out there objecting?

> It is also strongly suggested that we included a minimal rust version to use, specifically 1.56.1

What is the benefit of explicitly setting the version? Isn't it already implicitly set by moving to Rust `2021` @NukeManDan?

nuke-web3:
I have not played with behavior for different versions with and without this included, but I believe it _enforces_ builds with _no less_ than this version, and in this case the 1.56.1 (with the security patch) will be required to build, without specifying a flag to ignore it.
This will _theoretically_ avoid issues that using older versions may introduce. But to be honest, it's not clear (I have not done a lot of research) if this is worth enforcing. My intuition is to add it.

nuke-web3:
`rust-version = "1.56.1"` introduced

For reference, I tried to get this to `cargo build` with 1.56.0 and it failed with
```
error: package `...` cannot be built because it requires rustc 1.56.1 or newer, while the currently active rustc version is 1.56.0
```
With `--ignore-rust-version` it _does_ build with 1.56.0, as expected.

Should be good to go with a squash merge to make the commit log clean on `master` IMHO :+1:

mxinden:
> `rust-version = "1.56.1"` introduced
> 
> For reference, I tried to get this to `cargo build` with 1.56.0 and it failed with
> 
> ```
> error: package `...` cannot be built because it requires rustc 1.56.1 or newer, while the currently active rustc version is 1.56.0
> ```
> 
> With `--ignore-rust-version` it _does_ build with 1.56.0, as expected.
> 
> Should be good to go with a squash merge to make the commit log clean on `master` IMHO +1

Thanks for testing. I understand that `rust-version` enforces usage of `> VERSION` in downstream dependencies, though I wonder whether we need to set it explicitly, or whether the implicit version restriction through `edition = "2021"` isn't enough?

thomaseizinger:
> Thanks for testing. I understand that `rust-version` enforces usage of `> VERSION` in downstream dependencies, though I wonder whether we need to set it explicitly, or whether the implicit version restriction through `edition = "2021"` isn't enough?

I think it should give a much better error message? At least for version 1.54 on-wards (where `rust-version` stabilized). Not a huge benefit I guess?

mxinden (approved):
Thanks @NukeManDan for the help!

PR #2346, opened 2021-11-16
add performance metrics to gossipsub

divagant-martian:
Add two metrics that help analyze if certain configurations/network are performing as expected.


protocols/gossipsub/src/lib.rs
```diff
@@ -122,6 +122,10 @@
 //! println!("Listening on {:?}", addr);
 //! ```
 
+// Re-export open_metrics_client so external applications can inject the correct Registry when
+// using metrics
+pub use open_metrics_client;
```
mxinden: I would guess users that use `libp2p-gossipsub` metrics also use `libp2p-metrics`? In addition I would assume that users consume both crates via the root level `libp2p` which combines versions of `libp2p-gossipsub` and `libp2p-metrics` using the same `open-metrics-client` version. With that in mind should this re-export not happen within `libp2p-metrics`?

AgeManning: Because we had to inject the `Regsitry` directly into gossipsub, the `libp2p-metrics` addition is rather superfluous (if you're using the metrics inside gossipsub). As it stands, `libp2p-metrics` only logs events seen by the behaviour (i.e message count) and these are relatively redundant compared to the detailed metrics we can get inside gossipsub now. 
I was going to remove the gossipsub addition to `libp2p-metrics` but figured it could still be useful for others who just want a high-level overview of their gossipsub implementation. 

Anyway, we are currently consuming this directly by the `libp2p-gossipsub` crate which is why I re-exported it there. Perhaps there is a better way to handle this? I can also add the re-export to `libp2p-metrics` if you think appropriate.

mxinden: I don't have a strong opinion on re-exports. If I am not mistaken, unless libp2p specific, I think we don't do them at all today.

Off the top of my head, I would prefer not re-exporting `open-metrics-client` at all. Users that try to use a different version than the one used in `libp2p-gossipsub` will get a compile time error like with any other version mismatch. Do we need to special case this here?

AgeManning: Yep sure. Removed the re-export. 

I guess I was being lazy, where I tie the dependency solely to libp2p and upgrading the libp2p dep upgrades the required open metric dep for me. I can manage the dep version in my project too.

AgeManning:
I think this is ready for review and another handy one to get down before some more gossipsub updates.

Will give a basis for future metrics should we need them.

mxinden (approved):
:rocket: 

As a side note, I have created a Grafana dashboard for metrics exposed for `libp2p-identify`, `libp2p-ping` and `libp2p-kad`. Maybe we can join efforts and include the Gossipsub metrics as well.

mxinden:
Mind resolving the merge conflicts now that https://github.com/libp2p/rust-libp2p/pull/2327 is merged? Other than that, this is good to go from my side.

AgeManning:
Yep. I've got a bunch related to the new metrics (some are lighthouse-specific) but we should be able to rip out the gossipsub ones and combine them. 

Do you have a repo somewhere for your dashboards. My current one is here: https://github.com/sigp/lighthouse-metrics/blob/network/dashboards/LighthouseNetwork.json

PR #2350, opened 2021-11-18
swarm/: Patch reporting on banned peer connections

### LINKED ISSUE #2337: swarm/: Never call Behaviour::inject_event without a previous Behaviour::inject_connection_established
Imagine the following scenario:

1. The user calls `Swarm::ban_peer_id` for peer X.
1. The `Swarm` receives a `NetworkEvent::IncomingConnection` for a not-yet authenticated peer.
2. The `Swarm` calls `this.network.accept` to accept the connection and thus starts the upgrade process.
3. The `Swarm` receives a `NetworkEvent::ConnectionEstablished` from the network for the previously accepted connection. The connection appears to be to peer X which is marked as banned and thus the `Swarm` starts closing the connection. **The `Swarm` does not call `NetworkBehaviour::inject_connection_established`.**
2. The `Swarm` receives a `NetworkEvent::ConnectionEvent` from the network from the now closing connection to peer X. **The `Swarm` calls `NetworkBehaviour::inject_event`.**

The above leads to `NetworkBehaviour::inject_event` being called without a previous `NetworkBehaviour::inject_connection_established`. This violates the API contract between `Swarm` and `NetworkBehaviour`. E.g. `libp2p-identify` expects `NetworkBehaviour::inject_event` to never be called without a previous `NetworkBehaviour::inject_connection_established`.

https://github.com/libp2p/rust-libp2p/blob/8b68026bb04a469654020a8dcfe68308404e2510/protocols/identify/src/identify.rs#L330-L333

I see multiple ways moving forward, ordered by preference:

1. Track the connection state (`Established` and `Closing`) in `EstablishedConnectionInfo` (`pool.rs`) and ignore any `task::EstablishedConnectionEvent::Notify` from the connection in case it is in `Closing` state.
1. Check `Swarm::banned_peers` before forwarding an event from the network to the behaviour in `Swarm` (`NetworkEvent::ConnectionEvent`).
3. Drop the promise of `Swarm` to never call `NetworkBehaviour::inject_event` without a previous `NetworkBehaviour::inject_connection_established` and adjust `libp2p-identify` accordingly.

Credit goes to @divagant-martian for associating the `libp2p-identify` panic with the `Swarm` banning system.

//CC @AgeManning 

What do folks think?

divagant-martian:
Closes #2337 
- Track banned connections in the swarm.
- Prevent notifications to the behaviour for banned connections.
- Include contract assertions to the `CallTraceBehaviour`.
- Rework ban test to check for verious conditions.
- `inject_connected` and `inject_closed` are called for the first and last allowed connections, respectively.


divagant-martian:
I added an extremely ugly test to check that `inject_connection_closed` is not called for a banned connection. The test fails on master, succeeds here. I'm not sure if the test adds value, so please let me know if you'd like to maintain it to try to make it more up to your (and mine) standards

mxinden (commented):
Thanks @divagant-martian for the patch!

I would like this change to be contained within `libp2p-swarm`, i.e. the `libp2p-swarm` concept of banning a peer should not leak into `libp2p-core` (here via `is_allowed`).

Instead of the current `is_allowed`, could one keep track of the banned connections in `Swarm` instead? E.g. something along the lines of:

```diff
diff --git a/swarm/src/lib.rs b/swarm/src/lib.rs
index 0059a21a..0d422561 100644
--- a/swarm/src/lib.rs
+++ b/swarm/src/lib.rs
@@ -275,6 +275,10 @@ where
     /// List of nodes for which we deny any incoming connection.
     banned_peers: HashSet<PeerId>,
 
+    /// List of banned connections and whether they have been reported as open to the
+    /// [`NetworkBehaviour`].
+    banned_connections: HashMap<ConnectionId, bool>,
+
     /// Pending event to be delivered to connection handlers
     /// (or dropped if the peer disconnected) before the `behaviour`
     /// can be polled again.
```

Combining the above with @MarcoPolo laws:


- If a peer is banned, all active connections are still in the "unbanned state", but the connections are made to disconnect.
  
  Active connections to now banned peer are added to `Swarm::banned_connections` with `true`.

- If a peer is banned, all new connections will be in the "banned state" and thus not emit any events to the network behavior.

  New connections to banned peer are added to `Swarm::banned_connections` with `false`.

- If a peer is unbanned (and previously banned), all active connections are still in the "banned state".

  Active connections stay in `Swarm::banned_connections` and are only removed from `Swarm::banned_connections` on closing. On closing check the `bool` in `Swarm::banned_connections` to tell whether the closing should be reported to the `NetworkBehaviour`.
- If a peer is unbanned, all new connections will be in the "unbanned state".

  New connections for unbanned peer are not added to `Swarm::banned_connections`.

divagant-martian:
Hey,

> I would like this change to be contained within `libp2p-swarm`, i.e. the `libp2p-swarm` concept of banning a peer should not leak into `libp2p-core` (here via `is_allowed`).

Alright, makes sense. As a library user I still see it all as a single unit but if the ban is contained in the swarms, sounds like the best option.

> Instead of the current `is_allowed`, could one keep track of the banned connections in `Swarm` instead? E.g. something along the lines of:
> 
> ```diff
> diff --git a/swarm/src/lib.rs b/swarm/src/lib.rs
> index 0059a21a..0d422561 100644
> --- a/swarm/src/lib.rs
> +++ b/swarm/src/lib.rs
> @@ -275,6 +275,10 @@ where
>      /// List of nodes for which we deny any incoming connection.
>      banned_peers: HashSet<PeerId>,
>  
> +    /// List of banned connections and whether they have been reported as open to the
> +    /// [`NetworkBehaviour`].
> +    banned_connections: HashMap<ConnectionId, bool>,
> +
>      /// Pending event to be delivered to connection handlers
>      /// (or dropped if the peer disconnected) before the `behaviour`
>      /// can be polled again.
> ```

I think a `HashSet` would be enough, where `!conn.is_allowed()` is equivalent to `banned_connections.contains(&conn.id())`.

divagant-martian:
@mxinden Moved the logic to the swarm. I also removed the ugly test and reworked the existing ban test (see updated PR description) The test however, does not finish every now and then. I'd appreciate if you or @MarcoPolo can give it a look

swarm/src/lib.rs
```diff
@@ -676,14 +691,16 @@ where
...
-                    if num_established == 0 {
-                        this.behaviour.inject_disconnected(&peer_id);
+                    if !this.banned_connections.remove(&id) {
+                        this.behaviour.inject_connection_closed(
+                            &peer_id,
+                            &id,
+                            &endpoint,
+                            handler.into_protocols_handler(),
+                        );
+                        if num_established == 0 {
```
mxinden: What if there is another connection to the same peer, but that connection is in `banned_connections`? `num_established` would be `!= 0` and thus `inject_disconnected` would never be called. Correct?

If that is the case, we could carry the concrete `ConnectionId`s in `ConnectionClosed::num_established`. In case `num_established != 0`, we also check whether the remaining `ConnectionId`s are in `banned_connections` and if so, call `inject_disconnected` as well.

mxinden (commented):
Just one comment. Still need to give this an in-depth review but looking good overall. Thanks for the work @divagant-martian!

swarm/src/lib.rs
divagant-martian: That's a very good catch. I'll think how to solve it tomorrow and hopefully have this ready by your monday

MarcoPolo: I’m sure you’re already planning on doing this, so please forgive the note if so.

We should double check all other usages of `num_established` – particularly the [`inject_connected`](https://github.com/libp2p/rust-libp2p/blob/eefc25d5dffdb464afa5d30e94d7d0eaad42b73e/swarm/src/lib.rs#L669) call. Maybe we can even add a helper method to make this less tricky? something like `num_established_unbanned_conns` (naming?).

There’s some ambiguity whether network events should use the raw `num_established` or they should use the `num_established_unbanned_conns` form instead. I’m not sure. My guess is it should use the `num_established` because that’s the real number of established connections. And otherwise you may get weird cases where the event says 0 established, but some `is_connected` method returns true (maybe I’m wrong here).

divagant-martian: My politic here would be: do the changes that make us fulfill the contract since the behaviour's perspective. From the swarm's API, to ignore banned peers we would need to rework way more than what I think is in scope of the current PR. For completeness:
- `network_info()` with all it's counters and etc will now and then not match the behaviour's view.
- `num_established` will also differ when banned connections are still open.
- `is_connected` might return `true` for a banned peer.
However unexpected, none of these break a contract. The info about the network is untainted with the banning info. From my perspective that's fine.

divagant-martian:
Not sure why I can't re-request a review from @MarcoPolo. But please both give this a review. I think it's ready

core/src/connection/pool.rs
```diff
@@ -156,7 +156,7 @@ where
     /// A new connection has been established.
     ConnectionEstablished {
         connection: EstablishedConnection<'a, THandlerInEvent<THandler>>,
-        num_established: NonZeroU32,
+        established_ids: Vec<ConnectionId>,
```
mxinden:
```suggestion
        /// List of other connections to the same peer.
        /// 
        /// Note: Does not include the connection reported through this event.
        other_established_connection_ids: Vec<ConnectionId>,
```

core/src/network/event.rs
```diff
@@ -100,9 +100,9 @@ where
     ConnectionEstablished {
         /// The newly established connection.
         connection: EstablishedConnection<'a, TInEvent>,
-        /// The total number of established connections to the same peer,
-        /// including the one that has just been opened.
-        num_established: NonZeroU32,
+        /// The ids of the open connections for the peer, not including the one that has just been
+        /// opened.
+        established_ids: Vec<ConnectionId>,
```
mxinden:
```suggestion
        /// List of other connections to the same peer.
        /// 
        /// Note: Does not include the connection reported through this event.
        other_established_connection_ids: Vec<ConnectionId>,
```

I would like to be explicit about the fact that this does not include the currently reported connection, thus the `other_`. What do you think?

core/src/network/event.rs
```diff
@@ -128,8 +128,8 @@ where
         connected: Connected,
         /// The error that occurred.
         error: Option<ConnectionError<<THandler::Handler as ConnectionHandler>::Error>>,
-        /// The remaining number of established connections to the same peer.
-        num_established: u32,
+        /// The remaining established connections to the same peer.
+        established_ids: Vec<ConnectionId>,
```
mxinden:
```suggestion
        /// List of remaining established connections to the same peer.
        remaining_established_connection_ids: Vec<ConnectionId>,
```

swarm/src/lib.rs
```diff
@@ -603,38 +607,56 @@ where
...
-                    let connection = connection.id();
-                    this.behaviour.inject_event(peer, connection, event);
+                    let conn_id = connection.id();
+                    if !this.banned_peer_connections.contains(&conn_id)
+                        || this.banned_peers.contains(&peer)
+                    {
+                        this.behaviour.inject_event(peer, conn_id, event);
+                    } else {
+                        log::debug!(
+                            "Ignoring event from disallowed connection or peer: {:?} {:?}.",
```
```suggestion
                            "Ignoring event from banned peer: {:?} {:?}.",
```

Given that a user _banns_ a peer, wouldn't that be more intuitive here?

swarm/src/lib.rs
```diff
@@ -603,38 +607,56 @@ where
...
+                        this.banned_peer_connections.insert(connection.id());
                         this.network
                             .peer(peer_id)
                             .into_connected()
                             .expect("the Network just notified us that we were connected; QED")
                             .disconnect();
                         return Poll::Ready(SwarmEvent::BannedPeer { peer_id, endpoint });
                     } else {
+                        let num_established =
+                            NonZeroU32::new(u32::try_from(established_ids.len() + 1).unwrap())
```
mxinden: If we rename `established_ids` to `other_established_connection_ids` (see suggestion above), I would find the fact that we `+1` here more intuitive.

swarm/src/lib.rs
```diff
@@ -603,38 +607,56 @@ where
...
                 Poll::Ready(NetworkEvent::ConnectionEstablished {
                     connection,
-                    num_established,
+                    established_ids,
                     concurrent_dial_errors,
                 }) => {
                     let peer_id = connection.peer_id();
                     let endpoint = connection.endpoint().clone();
                     if this.banned_peers.contains(&peer_id) {
+                        // Mark the connection for the banned peer as disallowed.
```
```suggestion
                        // Mark the connection for the banned peer as banned, thus withholding any future events from the connection to the behaviour.
```

This might seem like a nit pick, but still I wonder whether we need to differentiate in _banned_ peers and _disallowed_ connections. Why not call them _banned_ connections instead of _disallowed_ connections? As a newcomer I would find this more intuitive.

swarm/src/lib.rs
```diff
@@ -275,6 +276,9 @@ where
     /// List of nodes for which we deny any incoming connection.
     banned_peers: HashSet<PeerId>,
 
+    /// Connections for which we withhold any reporting. These belong to banned peers.
+    banned_peer_connections: HashSet<ConnectionId>,
```
```suggestion
    /// Connections for which we withhold any reporting. These belong to banned peers.
    ///
    /// Note: Connections to a peer that are established at the time of banning that peer
    /// are not added here. Instead they are simply closed.
    banned_peer_connections: HashSet<ConnectionId>,
```

swarm/src/lib.rs
```diff
@@ -651,7 +673,13 @@ where
...
                             failed_addresses.as_ref(),
                         );
-                        if num_established.get() == 1 {
+                        // The peer is not banned, but there could be previous banned connections
+                        // if the peer was just unbanned. Check if this is the first allowed
+                        // connection.
+                        let first_allowed = established_ids
+                            .into_iter()
+                            .all(|conn_id| this.banned_peer_connections.contains(&conn_id));
+                        if first_allowed {
```
```suggestion
                        // The peer is not banned, but there could be previous banned connections
                        // if the peer was just unbanned. Check if this is the first non-banned
                        // connection.
                        let first_non_banned = established_ids
                            .into_iter()
                            .all(|conn_id| this.banned_peer_connections.contains(&conn_id));
                        if first_non_banned {
```

Again, suggesting to call them `non-banned` and not `allowed` connections. What do you think?

swarm/src/lib.rs
```diff
@@ -676,14 +704,25 @@ where
...
-                        &peer_id,
-                        &id,
-                        &endpoint,
-                        handler.into_protocols_handler(),
-                    );
-                    if num_established == 0 {
-                        this.behaviour.inject_disconnected(&peer_id);
+                    let num_established = u32::try_from(established_ids.len()).unwrap();
+                    let conn_was_informed = !this.banned_peer_connections.remove(&id);
+                    if conn_was_informed {
```
```suggestion
                    let conn_was_reported = !this.banned_peer_connections.remove(&id);
                    if conn_was_reported {
```

Again, nit pick, sorry, trying to make it easy for future me to understand all of this.

We use the word `reported` above. I would prefer to stay consistent.

swarm/src/lib.rs
```diff
@@ -676,14 +704,25 @@ where
...
+                            handler.into_protocols_handler(),
+                        );
+
+                        // This connection was informed. Check if this is the last allowed
+                        // connection for the peer.
+                        let last_allowed = established_ids
+                            .into_iter()
+                            .all(|conn_id| this.banned_peer_connections.contains(&conn_id));
+
+                        if last_allowed {
```
```suggestion
                        // This connection was reported as open to the behaviour. Check if this is the last non-banned
                        // connection for the peer.
                        let last_non_banned = established_ids
                            .into_iter()
                            .all(|conn_id| this.banned_peer_connections.contains(&conn_id));

                        if last non_banned {
```

swarm/src/lib.rs
```diff
@@ -1505,60 +1537,104 @@ mod tests {
...
+        executor::block_on(future::poll_fn(move |cx| loop {
+            let poll1 = Swarm::poll_next_event(Pin::new(&mut swarm1), cx);
+            let poll2 = Swarm::poll_next_event(Pin::new(&mut swarm2), cx);
+            match stage {
+                Stage::Connecting => {
+                    if swarm1.behaviour.assert_connected(s1_expected_conns, 1)
+                        && swarm2.behaviour.assert_connected(s2_expected_conns, 1)
+                    {
+                        // Setup to test that already established connections are correctly closed
+                        // and reported as such after the peer in banned.
```
```suggestion
                        // and reported as such after the peer is banned.
```

swarm/src/lib.rs
```diff
@@ -1505,60 +1537,104 @@ mod tests {
...
+                        // Setup to test that new connections of banned peers are not reported.
+                        swarm1.dial(addr2.clone()).unwrap();
+                        s1_expected_conns += 1;
+                        stage = Stage::BannedDial;
+                    }
+                }
+                Stage::BannedDial => {
+                    if swarm2.network_info().num_peers() == 1 {
+                        // The banned connection was established. Check that the banning swarm was
+                        // not reported about the connection.
```
```suggestion
                        // The banned connection was established. Check that it was
                        // not reported to the behaviour of the banning swarm.
```

swarm/src/test.rs
```diff
@@ -223,6 +275,20 @@ where
...
+            "`inject_event` is called for reported connections."
+        );
+        assert!(
+            !self
+                .inject_connection_closed
+                .iter()
+                .any(|(peer_id, conn_id, _)| *peer_id == p && c == *conn_id),
+            "`inject_event` is never called for closed connections."
+        );
+
```
mxinden, line 291: :pray:

mxinden (commented):
Thanks @divagant-martian. This is great work.

I have a bunch of wording suggestions, but overall I think this is the way to go.

divagant-martian:
@mxinden all suggestions taken and applied. I think it's a good advice to keep the language consistent across the code :+1:

mxinden:
Unfortunately this pull request does introduce a breaking change to `libp2p-core`, thus we will need to release a new minor version of `libp2p-core` and all of its dependents.

I prepared the corresponding cargo and changelog entries (small hacky Python script). Could you cherry-pick https://github.com/libp2p/rust-libp2p/commit/10f7c188813da20147e6c84f114352f7170b8e19 into this pull request?

Also could you add a changelog entry to `libp2p-swarm` describing the bug fix?

Final ask: Can you run this on one of your lighthouse nodes for a bit to make sure all panics are fixed?

core/src/connection/pool.rs
```diff
@@ -183,8 +186,8 @@ where
         error: Option<ConnectionError<THandlerError<THandler>>>,
         /// A reference to the pool that used to manage the connection.
         pool: &'a mut Pool<THandler, TTrans>,
-        /// The remaining number of established connections to the same peer.
-        num_established: u32,
+        /// The remaining established connections to the same peer.
+        established_ids: Vec<ConnectionId>,
```
mxinden:
```suggestion
        remaining_established_connection_ids: Vec<ConnectionId>,
```

Sorry, missed this one. For the sake of consistency with `NetworkEvent`.

mxinden (commented):
Mind including this diff as well?

```diff
diff --git a/swarm/src/lib.rs b/swarm/src/lib.rs
index 0059a21a..2b2262eb 100644
--- a/swarm/src/lib.rs
+++ b/swarm/src/lib.rs
@@ -540,6 +540,10 @@ where
     pub fn ban_peer_id(&mut self, peer_id: PeerId) {
         if self.banned_peers.insert(peer_id) {
             if let Some(peer) = self.network.peer(peer_id).into_connected() {
+                // Note that established connections to the now banned peer are closed but not added
+                // to [`Swarm::banned_peer_connections`]. They have been previously reported as open
+                // to the behaviour and need be reported as closed once closing the connection
+                // finished.
                 peer.disconnect();
             }
         }
```

divagant-martian:
all done. We'll report in a few days with our findings

swarm/src/lib.rs
```diff
@@ -603,38 +614,54 @@ where
                 Poll::Pending => network_not_ready = true,
                 Poll::Ready(NetworkEvent::ConnectionEvent { connection, event }) => {
                     let peer = connection.peer_id();
-                    let connection = connection.id();
-                    this.behaviour.inject_event(peer, connection, event);
+                    let conn_id = connection.id();
+                    if this.banned_peer_connections.contains(&conn_id)
+                        || this.banned_peers.contains(&peer)
```
MarcoPolo: Why do we do the `|| this.banned_peers.contains(&peer)` check? 

My guess: So that if a peer banned we don't `inject_event` even on connections from before the ban. This is done because the disconnect of existing connections takes a bit and we don't want to get events from banned peers.

divagant-martian: That's the reason. That, and every possible condition of peers being banned, unbanned, etc and connections not closing on time

divagant-martian: At the end, it's also the safer bet

divagant-martian:
Based on runs, I think this looks fine @mxinden 
The last CI run failed but it looks like a random infra failure

EDIT: waiting one more day due to changes addressing [this comment](https://github.com/libp2p/rust-libp2p/pull/2350#discussion_r757017654)

swarm/src/lib.rs
mxinden: Sorry in advance for reviving this discussion once more, especially since you already tested this version on your infra @divagant-martian.

Imagine the scenario below:

- Local peer is connected to peer A via a single connection C1.
- Peer A sends two events to the local peer via C1. The second event depends on the first event. The events are about to be delivered to the local `Swarm` from the local `Network`.
- User banns peer A.
- User polls the `Swarm`.
  - The `Swarm` receives the first event from `Network` sent via C1.
  - `Swarm` drops it due to `this.banned_peers.contains(&peer)`.
- User unbanns A.
- User polls the `Swarm`.
  - The `Swarm` receives the second event from `Network` sent via C1.
  - The `Swarm` does not drop the second event as `this.banned_peers.contains(&peer)` is no longer true.
  - `Swarm` delivers the second event to the `NetworkBehaviour`.

Result: `NetworkBehaviour` receives the second but not the first event from peer A.

I don't think the behavior above is intuitive for an implementor of `NetworkBehaviour`. In other words, as an implementor of `NetworkBehaviour` I would expect to either receive all events from a single remote peer in their correct order or not at all.

With the above in mind, would it not be better to remove `|| this.banned_peers.contains(&peer)` @divagant-martian and @MarcoPolo? A `NetworkBehaviour` would potentially receive events from banned peers until already existing connections are closed, but a `NetworkBehaviour` would never receive events _out of order_.

divagant-martian: addressed in b43a169

divagant-martian:
all looking bright on our side

mxinden (approved):
:pray: thanks!

PR #2352, opened 2021-11-21
core: Add support for ECDSA identities

### LINKED ISSUE #556: Add ECDSA public key format
https://github.com/libp2p/specs/pull/100

davxy:
Closes https://github.com/libp2p/rust-libp2p/issues/556

Implementation compliant with https://github.com/libp2p/specs/blob/master/peer-ids/peer-ids.md#ecdsa

### DIFF
--- a/core/src/identity/ecdsa.rs
+++ b/core/src/identity/ecdsa.rs
@@ -1,0 +1,186 @@
+// Copyright 2019 Parity Technologies (UK) Ltd.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a
+// copy of this software and associated documentation files (the "Software"),
+// to deal in the Software without restriction, including without limitation
+// the rights to use, copy, modify, merge, publish, distribute, sublicense,
+// and/or sell copies of the Software, and to permit persons to whom the
+// Software is furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS
+// OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+// FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+// DEALINGS IN THE SOFTWARE.
+
+//! Ecdsa keys.
+
+use super::error::{DecodingError, SigningError};
+use core::fmt;
+use ring::{
+    rand::SystemRandom,
+    signature::{
+        self, EcdsaKeyPair as EcdsaKeyPairImpl, EcdsaSigningAlgorithm, EcdsaVerificationAlgorithm,
+        KeyPair as RingKeyPair, UnparsedPublicKey as PublicKeyImpl,
+    },
+};
+use std::sync::Arc;
+
+#[derive(Debug, Clone, PartialEq, Eq, Copy)]
+pub enum CurveId {
+    Secp256r1,
+}
+
+/// An ECDSA keypair.
+#[derive(Clone)]
+pub struct Keypair {
+    curve_id: CurveId,
+    pair: Arc<EcdsaKeyPairImpl>,
+}
+
+impl Keypair {
+    /// Generate a new random ECDSA p256 keypair.
+    pub fn generate(curve_id: CurveId) -> Keypair {
+        let rng = SystemRandom::new();
+        let alg = get_sign_alg(curve_id);
+        let pkcs8 = EcdsaKeyPairImpl::generate_pkcs8(alg, &rng).unwrap();
+        let imp = EcdsaKeyPairImpl::from_pkcs8(alg, pkcs8.as_ref()).unwrap();
+        Keypair {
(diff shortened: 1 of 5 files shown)


mxinden (approved):
Thank you @davxy for the pull request.

This looks good to me, though I don't think I should be the only reviewer on a crypto focused pull request.

@dignifiedquire or @burdges do you have opinions on `p256` vs `ring`?

dignifiedquire:
@tarcieri should be able to give a better comparison than me.

dignifiedquire (commented):
implementation from a crypto standpoint looks good to me

core/src/identity/ecdsa.rs
```diff
@@ -0,0 +1,161 @@
...
+    }
+}
+
+/// An ECDSA public key.
+#[derive(Clone, PartialEq, Eq)]
+pub struct PublicKey(VerifyingKey);
+
+impl PublicKey {
+    /// Verify an ECDSA signature on a message using the public key.
+    pub fn verify(&self, msg: &[u8], sig: &[u8]) -> bool {
```
thomaseizinger, line 125: Might be good to annotate this with `#[must_use]` to avoid accidentially discarding the result.

core/src/identity/ecdsa.rs
```diff
@@ -0,0 +1,161 @@
...
+/// Demote an ECDSA keypair to a secret key.
+impl From<Keypair> for SecretKey {
+    fn from(kp: Keypair) -> SecretKey {
+        kp.secret
+    }
+}
+
+/// An ECDSA secret key.
+#[derive(Clone)]
+pub struct SecretKey(SigningKey);
```
thomaseizinger, line 87: There is currently no way of constructing one of these from bytes. How are nodes meant to have to persistent identity if all we can do is generate random ones?

core/src/identity/ecdsa.rs
```diff
@@ -0,0 +1,161 @@
...
+
+    /// Sign a message with this secret key, producing a DER-encoded ECDSA signature.
+    pub fn sign(&self, msg: &[u8]) -> Vec<u8> {
+        self.0.sign(msg).to_der().as_bytes().to_owned()
+    }
+}
+
+impl fmt::Debug for SecretKey {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "SecretKey")
```
thomaseizinger, line 115: Over at rust-bitcoin, we settled on including a fingerprint (first couple of bytes of tagged hash) in the debug output to make it possible to differentiate different secret keys. Thoughts on doing the same thing here?

core/src/identity/ecdsa.rs
```diff
@@ -0,0 +1,161 @@
...
+
+    /// Encode a public key into a DER encoded byte buffer as defined by SEC1 standard.
+    pub fn encode_der(&self) -> Vec<u8> {
+        self.0.to_encoded_point(false).as_bytes().to_owned()
+    }
+
+    // Decode a public key from a DER encoded byte buffer as defined for SEC1 standard.
+    pub fn decode_der(k: &[u8]) -> Result<PublicKey, DecodingError> {
+        VerifyingKey::from_sec1_bytes(k)
+            .map_err(|_| DecodingError::new("failed to parse ecdsa p256 public key"))
```
thomaseizinger: Could `?` after this to avoid the last `.map`.

thomaseizinger: We are dropping the source error here. Is that what we do for other keypairs as well? I think we should include the source error message.

thomaseizinger (commented):
Thanks for the contribution!

Some comments / discussion points.

core/src/identity/ecdsa.rs
davxy: I've followed the design of the other key types. I can add the attribute, but then I suggest to do the same for the others.

core/src/identity/ecdsa.rs
davxy: Legitimate. I add the functions

core/src/identity/ecdsa.rs
davxy: I've followed the design of the other key types. I can add the attribute, but then I suggest to do the same for the others.

Can you elaborate on what should be used as private key fingerprint?

core/src/identity/ecdsa.rs
thomaseizinger: I didn't know it is missing from the others!

I think it would be good to add but we can probably leave this to a follow-up PR then.

core/src/identity/ecdsa.rs
thomaseizinger: > I've followed the design of the other key types. I can add the attribute, but then I suggest to do the same for the others.

Leave it then, would probably need more discussion overall :)

> Can you elaborate on what should be used as private key fingerprint?

I'd suggest using the first couple of bytes of a tagged hash (ie `sha256(b"ECDSA-secret-key" + bytes)[0..8]`) but if none of the other secret keys do this, we don't need to add it here.

core/src/identity/ecdsa.rs
davxy: The second `map` has been inserted to map from `Result<VerifyingKey>` into a `Result<PublicKey>`.

With PublicKey defined as : `struct PublicKey(VerifyingKey)`

davxy: This should have been resolved.

Where possible the errors are propagated using `source()`. 

There is an error that cannot be stored within the source because it doesn't implement `Error`. In this case I've added some details to the error message.

```
        let enc_pt = EncodedPoint::from_bytes(k).map_err(|_| {   // <- this doesn't implement error on `no_std` builds
            DecodingError::new("failed to parse ecdsa p256 public key, bad point encoding")
        })?;
```

thomaseizinger (approved):
Looks good to me, thank you!

mxinden (approved):
Thanks @davxy for the contribution.

Also tested with https://github.com/libp2p/rust-libp2p/blob/master/examples/ipfs-kad.rs using ECDSA identity.

mxinden:
Also thanks to @thomaseizinger and @dignifiedquire for the input!

PR #2354, opened 2021-11-21
Request-Response: Add method connected_peers

elenaf9:
The request-response protocol manages a list of connected peers. This list is currently not exposed to the user, only the method `RequestResponse::is_connected` allows to the check the connection for a specific peers.
However, in case that the user needs to know the whole list of connected peers, it forces them to also track connected peers themself, which is redundant as the information is already there, e.g. as it currently has to be done in the AutoNAT draft #2262.
56f1ede190834e3876649d40a44f4e83b807fa8c adds a method `RequestResponse::connected_peers` to share the list of connected peers.

Apart from that, I noticed that at the moment `inject_address_change` is not handled, which means that the associated `Connection` still has the address of the old `ConnectedPoint.` bd6622a1dd3e59bad212f10fc0619198c32084b1 changes this, unless there was a specific reason to do so?  
Note that with this change, if the ConnectedPoint changed from `Dialer` to `Listener`, the address will change from `Some(old_addr)` to `None`. Are there cases in which we still want to keep the old address? If I am not mistaken, the address is only used to return it in `addresses_of_peer`, therefore I would say it's valid to remove an outdated address?

### DIFF
--- a/protocols/request-response/src/lib.rs
+++ b/protocols/request-response/src/lib.rs
@@ -450,6 +450,11 @@ where
         }
     }
 
+    /// List of currently connected peers.
+    pub fn connected_peers(&self) -> Vec<&PeerId> {
+        self.connected.keys().collect()
+    }
+
     /// Checks whether an outbound request to the peer with the provided
     /// [`PeerId`] initiated by [`RequestResponse::send_request`] is still
     /// pending, i.e. waiting for a response.
@@ -686,7 +691,7 @@ where
                     self.pending_events
                         .push_back(NetworkBehaviourAction::GenerateEvent(
                             RequestResponseEvent::OutboundFailure {
-                                peer: peer,
+                                peer,
                                 request_id: request.request_id,
                                 error: OutboundFailure::DialFailure,
                             },
--- a/protocols/request-response/src/lib.rs
+++ b/protocols/request-response/src/lib.rs
@@ -598,6 +598,29 @@ where
         addresses
     }
 
+    fn inject_address_change(
+        &mut self,
+        peer: &PeerId,
+        conn: &ConnectionId,
+        _old: &ConnectedPoint,
+        new: &ConnectedPoint,
+    ) {
+        let new_address = match new {
+            ConnectedPoint::Dialer { address } => Some(address.clone()),
+            ConnectedPoint::Listener { .. } => None,
+        };
+        let connections = self
+            .connected
+            .get_mut(peer)
+            .expect("Address change can only happen on an established connection.");
+
+        let connection = connections
+            .iter_mut()
+            .find(|c| &c.id == conn)
+            .expect("Address change can only happen on an established connection.");
+        connection.address = new_address;
+    }
+
     fn inject_connected(&mut self, peer: &PeerId) {
         if let Some(pending) = self.pending_outbound_requests.remove(peer) {
             for request in pending {
--- a/protocols/request-response/tests/ping.rs
+++ b/protocols/request-response/tests/ping.rs
@@ -122,6 +122,11 @@ fn ping_protocol() {
                     assert_eq!(&peer, &peer2_id);
                 }
                 SwarmEvent::Behaviour(e) => panic!("Peer1: Unexpected event: {:?}", e),
+                SwarmEvent::ConnectionEstablished { peer_id, .. } => {
+                    assert_eq!(peer_id, peer2_id);
+                    assert!(swarm1.behaviour().is_connected(&peer_id));
+                    assert_eq!(swarm1.behaviour().connected_peers(), vec![&peer_id]);
+                }
                 _ => {}
             }
         }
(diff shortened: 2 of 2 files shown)


mxinden:
> The request-response protocol manages a list of connected peers. This list is currently not exposed to the user, only the method `RequestResponse::is_connected` allows to the check the connection for a specific peers. However, in case that the user needs to know the whole list of connected peers, it forces them to also track connected peers themself, which is redundant as the information is already there, e.g. as it currently has to be done in the AutoNAT draft #2262. [56f1ede](https://github.com/libp2p/rust-libp2p/commit/56f1ede190834e3876649d40a44f4e83b807fa8c) adds a method `RequestResponse::connected_peers` to share the list of connected peers.

I am undecided on this one. On the one hand, yes, it is just a small getter method. On the other hand:

- It includes an internal concept in the public API. Say e.g. that at some point we don't want to track all, but only a subset of connected peers. Even though nothing within `libp2p-request-response` would break, it would still require a breaking version release.
- I would argue that tracking the connected peers in each behaviour is easy (via `inject_connected` and `inject_disconnected`) and I doubt it is performance relevant.
- I oftentimes find myself having to track additional data with each peer, specific to the `NetworkBehaviour` implementation. E.g. for `libp2p-autonat` we might want to track in the future, whether the connected peer is within our subnetwork, or connected via the public internet. If I am not mistaken `libp2p-autonat` would then have to do its own tracking anyways.


> Apart from that, I noticed that at the moment `inject_address_change` is not handled, which means that the associated `Connection` still has the address of the old `ConnectedPoint.` [bd6622a](https://github.com/libp2p/rust-libp2p/commit/bd6622a1dd3e59bad212f10fc0619198c32084b1) changes this, unless there was a specific reason to do so? Note that with this change, if the ConnectedPoint changed from `Dialer` to `Listener`, the address will change from `Some(old_addr)` to `None`. Are there cases in which we still want to keep the old address? If I am not mistaken, the address is only used to return it in `addresses_of_peer`, therefore I would say it's valid to remove an outdated address?

Good catch. I am not aware of any cases, where this would not be desired behaviour. Thanks for adding it.

elenaf9:
> It includes an internal concept in the public API. Say e.g. that at some point we don't want to track all, but only a subset of connected peers. Even though nothing within libp2p-request-response would break, it would still require a breaking version release.

Isn't this already the case right now because we have `is_connected`? I can not think of a change that would break the implementation of `connected_peers`, but not of `is_connected`, since they both depend on the same property. 

> I would argue that tracking the connected peers in each behaviour is easy (via inject_connected and inject_disconnected) and I doubt it is performance relevant.

> I oftentimes find myself having to track additional data with each peer, specific to the NetworkBehaviour implementation. E.g. for libp2p-autonat we might want to track in the future, whether the connected peer is within our subnetwork, or connected via the public internet. If I am not mistaken libp2p-autonat would then have to do its own tracking anyways.

In regard to AutoNAT you are probably right, yes. But in general this implies that the user implements the behaviour manually. In case of `#[derive(NetworkBehaviour)]` this would not be possible, so if then a user needs the list of connected peers they would have to implement the whole logic on a higher level when polling the Swarm.

I don't really have a strong opinion on this. For AutoNAT I agree with you that eventually we have to track the connected peers separately anyway. And e.g. `Swarm` also only provides a `is_connected` method, but not a `connected_peers`. 
So by explicitly not providing these methods it enforces the usage of  `NetworkBehaviour` trait implementation and / or `SwarmEvent`s to track the networks state, which is also a valid design.
I am fine with either,  so if you'd like to close this PR I could just open a new one that only cherry picks the `inject_address_change` patch bd6622a1dd3e59bad212f10fc0619198c32084b1.

mxinden:
> > It includes an internal concept in the public API. Say e.g. that at some point we don't want to track all, but only a subset of connected peers. Even though nothing within libp2p-request-response would break, it would still require a breaking version release.
> 
> Isn't this already the case right now because we have `is_connected`? I can not think of a change that would break the implementation of `connected_peers`, but not of `is_connected`, since they both depend on the same property.

Fair point!

> In regard to AutoNAT you are probably right, yes. But in general this implies that the user implements the behaviour manually. In case of `#[derive(NetworkBehaviour)]` this would not be possible, so if then a user needs the list of connected peers they would have to implement the whole logic on a higher level when polling the Swarm.

Do I understand correctly that this mechanism would not only be useful for a `NetworkBehaviour` wrapping `RequestResponse`, but also for a user of `Swarm` who could then do `swarm.behaviour().connected_peers`? For the latter I would much rather prefer exposing `Network::connected_peers` via `Swarm::connected_peers`, thus returning the data right from the source (`Network`), always available no matter if the user uses `RequestResponse` or not.

> I am fine with either, so if you'd like to close this PR I could just open a new one that only cherry picks the `inject_address_change` patch [bd6622a](https://github.com/libp2p/rust-libp2p/commit/bd6622a1dd3e59bad212f10fc0619198c32084b1).

My preference is still to include bd6622a only. Others might have an opinion as well though.

elenaf9:
> Do I understand correctly that this mechanism would not only be useful for a NetworkBehaviour wrapping RequestResponse, but also for a user of Swarm who could then do swarm.behaviour().connected_peers? For the latter I would much rather prefer exposing Network::connected_peers via Swarm::connected_peers, thus returning the data right from the source (Network), always available no matter if the user uses RequestResponse or not.

Good point, yes I agree.
Since nobody else voiced any opinion on this (and it's only a minor thing anyway), I am closing this PR now. Anybody who needs `Swarm::connected_peers` should feel free to add a PR for it themself, or cherry-pick 56f1ede190834e3876649d40a44f4e83b807fa8c if they need it in the Request-Response protocol after all.
PR #2362 will add the address-change handling.

mxinden:
For the record `connected_peers` is added via https://github.com/libp2p/rust-libp2p/pull/2378/.

Issue #2345, opened 2021-11-16
examples/*: Rewrite event loops using async/await


mxinden:
Today, event loops in `examples/` use poll-style futures. Using async/await simplifies the examples, thus helping newcomers to get started easier.

Example of a complicated poll-style loop:

https://github.com/libp2p/rust-libp2p/blob/4bd44c812a577d0889f9b57f5dcda98113a35573/examples/chat.rs#L142-L195

Example of a simple async/await loop:

https://github.com/libp2p/rust-libp2p/blob/4bd44c812a577d0889f9b57f5dcda98113a35573/protocols/rendezvous/examples/discover.rs#L63-L121

jochasinga:
There is a [chat-tokio.rs](./examples/chat-tokio.rs) that is using an async-await on tokio runtime. Maybe we should also update other outdated examples to use tokio instead of async_std?

thomaseizinger:
The others are not outdated as such! Rust-libp2p supports multiple runtimes, hence there are examples on how to use either.

Tokio requires a bit more configuration, hence there is a dedicated example.

mxinden:
Closing here with https://github.com/libp2p/rust-libp2p/pull/2356 merged.

PR #2356, opened 2021-11-22
examples/*: Migrate to async fn main

JerryHue:
Addresses issue #2345.

* Adapt async style loop for chat.rs
* Adapt async style loop for distributed-key-value-store.rs
* Adapt async style loop for gossibsub-chat.rs
* Adapt async style loop for ipfs-private.rs
* Adapt ping to use async
* Update tutorial crate to reflect new changes

### DIFF
--- a/examples/chat.rs
+++ b/examples/chat.rs
@@ -141,8 +141,7 @@ async fn main() -> Result<(), Box<dyn Error>> {
     // Kick it off
     loop {
         select! {
-            line = stdin.next() => match line {
-                Some(line) => swarm
+            line = stdin.next() => swarm.behaviour_mut().floodsub.publish(floodsub_topic.clone(), line.expect("Stdin not to close")?.as_bytes())
                     .behaviour_mut()
                     .floodsub
                     .publish(floodsub_topic.clone(), line?.as_bytes()),
--- a/examples/chat.rs
+++ b/examples/chat.rs
@@ -141,12 +141,10 @@ async fn main() -> Result<(), Box<dyn Error>> {
     // Kick it off
     loop {
         select! {
-            line = stdin.next() => swarm.behaviour_mut().floodsub.publish(floodsub_topic.clone(), line.expect("Stdin not to close")?.as_bytes())
-                    .behaviour_mut()
-                    .floodsub
-                    .publish(floodsub_topic.clone(), line?.as_bytes()),
-                None => panic!("Stdin closed")
-            },
+            line = stdin.select_next_some() => swarm
+                .behaviour_mut()
+                .floodsub
+                .publish(floodsub_topic.clone(), line.expect("Stdin not to close").as_bytes()),
             event = swarm.select_next_some() => match event {
                 SwarmEvent::NewListenAddr { address, .. } => {
                     println!("Listening on {:?}", address);
--- a/examples/chat.rs
+++ b/examples/chat.rs
@@ -50,7 +50,10 @@
 //! The two nodes then connect.
 
 use async_std::{io, task};
-use futures::{future, prelude::*};
+use futures::{
+    prelude::{stream::StreamExt, *},
+    select,
+};
 use libp2p::{
     floodsub::{self, Floodsub, FloodsubEvent},
     identity,
@@ -58,10 +61,7 @@ use libp2p::{
     swarm::SwarmEvent,
     Multiaddr, NetworkBehaviour, PeerId, Swarm,
 };
-use std::{
-    error::Error,
-    task::{Context, Poll},
-};
+use std::error::Error;
 
 #[async_std::main]
 async fn main() -> Result<(), Box<dyn Error>> {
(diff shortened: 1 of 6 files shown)


examples/chat.rs
```diff
@@ -133,50 +133,47 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
 
     // Kick it off
-    task::block_on(future::poll_fn(move |cx: &mut Context<'_>| {
-        loop {
-            match stdin.try_poll_next_unpin(cx)? {
-                Poll::Ready(Some(line)) => swarm
+    loop {
+        select! {
+            line = stdin.next() => match line {
+                Some(line) => swarm
```
mxinden:
```suggestion
            line = stdin.next() => swarm.behaviour_mut().floodsub.publish(floodsub_topic.clone(), line.expect("Stdin not to close")?.as_bytes())
```

What do you think? Should be a bit simpler?

examples/distributed-key-value-store.rs
```diff
@@ -157,35 +154,28 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
-                Poll::Ready(None) => return Poll::Ready(Ok(())),
-                Poll::Pending => break,
+    loop {
+        select! {
+            line = stdin.next() => match line {
+                Some(line) => {
+                    handle_input_line(&mut swarm.behaviour_mut().kademlia, line?)
+                },
+                None => panic!("Stdin closed")
+            },
```
mxinden:
```suggestion
            line = stdin.next() => handle_input_line(&mut swarm.behaviour_mut().kademlia, line.expect("Stdin not to close")?),
```

A bit simpler, no?

mxinden (commented):
Much simpler. Thank you @JerryHue!

I would prefer simply `expect`ing stdin not to be closed. Other than that, this looks good to me.

examples/distributed-key-value-store.rs
JerryHue: Yeah, that looks better!

examples/chat.rs
JerryHue: Yeah, I agree. Getting rid of the other branch does simplify the code. After all, these are examples, there is no need for thorough error handling.

JerryHue: I will apply these changes and test them briefly.

JerryHue:
It seems that the suggested changes might have not applied the proper lines. I will fix them now.

examples/distributed-key-value-store.rs
```diff
@@ -157,35 +154,23 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
-                Poll::Ready(Some(event)) => {
-                    if let SwarmEvent::NewListenAddr { address, .. } = event {
-                        println!("Listening on {:?}", address);
-                    }
-                }
-                Poll::Ready(None) => return Poll::Ready(Ok(())),
-                Poll::Pending => break,
+    loop {
+        select! {
+            line = stdin.next() => handle_input_line(&mut swarm.behaviour_mut().kademlia, line.expect("Stdin not to close")?),
```
thomaseizinger: For the behaviour, we use `select_next_some` to unwrap the `Option`. I think that should work here as well?

It won't panic in that case because `select` is specifically designed to work well with `FusedFuture`: https://docs.rs/futures/0.3.18/futures/stream/trait.StreamExt.html#method.select_next_some

examples/ping.rs
```diff
@@ -75,17 +74,15 @@ fn main() -> Result<(), Box<dyn Error>> {
         println!("Dialed {}", addr)
     }
 
-    block_on(future::poll_fn(move |cx| loop {
-        match swarm.poll_next_unpin(cx) {
-            Poll::Ready(Some(event)) => match event {
+    block_on(async move {
```
thomaseizinger: Can we also use an `async fn main` in this example?

src/tutorial.rs
```diff
@@ -302,17 +301,15 @@
 //!         println!("Dialed {}", addr)
 //!     }
 //!
-//!     block_on(future::poll_fn(move |cx| loop {
-//!         match swarm.poll_next_unpin(cx) {
-//!             Poll::Ready(Some(event)) => match event {
+//!     block_on(async move {
```
thomaseizinger: Same here, can we use `async fn main` here?

thomaseizinger (commented):
Thanks for tackling this!

Commented with some ideas.

examples/ping.rs
JerryHue: That's what I was thinking as well, but I think we need to mark the `async fn main` with an attribute, which means we either have to use `tokio` or `async_std`, is that right? Do we want to mention the usage of `tokio` or `async_std` in the basic example?

examples/distributed-key-value-store.rs
JerryHue: I have used `select_next_some` for all places where we use `stdin`, except in `chat-tokio.rs`, because `tokio` does not have `select_next_some`, like in `async_std`.

examples/ping.rs
mxinden: Yes, I think it is fine. I suggest using `async-std` in all cases apart from the tokio specific ones. Not that I prefer `async-std` over `tokio`, but simply for consistency, given that `async-std` is the default across the whole repository.

thomaseizinger (approved):
Thank you!

Do you mind fixing the rustfmt error? Otherwise looks good to me!

JerryHue:
> Thank you!
> 
> Do you mind fixing the rustfmt error? Otherwise looks good to me!

All done! Should I rebase and squash all the changes?

thomaseizinger:
> > Thank you!
> > 
> > Do you mind fixing the rustfmt error? Otherwise looks good to me!
> 
> All done! Should I rebase and squash all the changes?

No need as we use squash-merge strategy in rust-libp2p!

mxinden (approved):
:rocket: thank you for the work @JerryHue!

PR #2361, opened 2021-11-27
swarm/: Improve `NetworkBehaviour` documentation

mxinden:
- Contrast `NetworkBehaviour` and `Transport`.
- Describe hierarchy of `NetworkBehaviour` implementations.
- Describe `NetworkBehaviour` combinators.
- Describe `NetworkBehaviour` derive macro.
  - Document `out_event` mechanism`.
  - Document `ignore` mechanism.

While I think this is a small step in the right direction, I still feel like the concept of a `NetworkBehaviour` is still hard to understand for newcomers. I have some additional steps in mind, though these need more thought.

@dignifiedquire @MarcoPolo given that both of you have recently felt the pain of having to understand the concept of `NetworkBehaviour`, your input would be valuable here. Obviously this is not to say that input by others is not valuable.


swarm/src/behaviour.rs
```diff
@@ -31,26 +31,116 @@ use std::{task::Context, task::Poll};
...
+/// [`NetworkBehaviour`].
+///
+/// ``` rust
+/// # use libp2p_swarm::DummyBehaviour;
+/// # use libp2p_swarm::toggle::Toggle;
+/// let my_behaviour = DummyBehaviour::default();
+/// let my_toggled_behaviour = Toggle::from(Some(my_behaviour));
+/// ```
+///
+/// ## Derive Macro
```
MarcoPolo, line 66: What made this clearer for me was looking at the expanded code.

Maybe we can add a sentence that explains what the derive macro outputs, just so the user can have a working mental model without necessarily having to dive into the [derive code themselves](https://github.com/libp2p/rust-libp2p/blob/master/swarm-derive/src/lib.rs).

Something like (feel free to reword):
> The derive macro will simply delegate trait methods to each struct member in the order the struct is defined. For `poll` it will first poll the first struct member until it returns Pending before moving on to later members. For `addresses_of_peers` it will delegate to each struct member and return a concatenated array of all results.

swarm/src/behaviour.rs
```diff
@@ -31,26 +31,116 @@ use std::{task::Context, task::Poll};
...
+///
+/// To compose multiple [`NetworkBehaviour`] implementations into a single [`NetworkBehaviour`]
+/// implementation, potentially building a multi-level hierarchy of [`NetworkBehaviour`]s, one can
+/// use one of the [`NetworkBehaviour`] combinators, and/or use the [`NetworkBehaviour`] derive
+/// macro.
+//G/
+/// ## Combinators
+///
+/// [`NetworkBehaviour`] combinators wrap one or more [`NetworkBehaviour`] implementations and
+/// implement [`NetworkBehaviour`] themselves. Example is the [`Toggle`](crate::toggle::Toggle)
```
MarcoPolo, line 56: This makes sense, thanks for this!

swarm/src/behaviour.rs
```diff
@@ -31,26 +31,116 @@ use std::{task::Context, task::Poll};
...
+/// hierarchy of [`NetworkBehaviour`]s where parent implementations delegate to child
+/// implementations. Finally the root of the [`NetworkBehaviour`] hierarchy is passed to
+/// [`Swarm`](crate::Swarm) where it can then control the behaviour of the local node on a libp2p
+/// network.
 ///
-/// Crate users can implement this trait by using the the `#[derive(NetworkBehaviour)]`
+/// # Hierarchy of [`NetworkBehaviour`]
+///
+/// To compose multiple [`NetworkBehaviour`] implementations into a single [`NetworkBehaviour`]
+/// implementation, potentially building a multi-level hierarchy of [`NetworkBehaviour`]s, one can
```
MarcoPolo, line 49: Is there a real-world example of a multi-level hierarchy? Not just the one level, but something that has 2 or more levels.

MarcoPolo (approved):
This makes it a bit clearer what the hierarchy is. I added a suggestion to demystify the derive macro a bit.

swarm/src/behaviour.rs
```diff
@@ -31,26 +31,116 @@ use std::{task::Context, task::Poll};
...
+/// use one of the [`NetworkBehaviour`] combinators, and/or use the [`NetworkBehaviour`] derive
+/// macro.
+//G/
```
dignifiedquire: stray `G`?

swarm/src/behaviour.rs
thomaseizinger: Yes. You could imagine using something like `libp2p-request-response` to make a custom request-response protocol but at the same time compose that with `libp2p-ping` into your root behaviour.

swarm/src/behaviour.rs
```diff
@@ -31,26 +31,116 @@ use std::{task::Context, task::Poll};
...
+///
+/// ## Derive Macro
+///
+/// One can derive [`NetworkBehaviour`] for a custom `struct` via the `#[derive(NetworkBehaviour)]`
 /// proc macro re-exported by the `libp2p` crate. The macro generates a delegating `trait`
 /// implementation for the `struct`, which delegates method calls to all struct members.
 ///
-/// Struct members that don't implement [`NetworkBehaviour`] must be annotated with `#[behaviour(ignore)]`.
-///
 /// By default the derive sets the [`NetworkBehaviour::OutEvent`] as `()` but this can be overridden
```
thomaseizinger, line 77: With the change to `event_process = false`, this might actually be a bad default now.

thomaseizinger: In theory, we could have the derive auto-generate an enum but having that one be "invisible" could be very strange. Thoughts?

swarm/src/behaviour.rs
mxinden: To add an example to @thomaseizinger comment, say you would use both `libp2p-ping` and [`libp2p-autonat`](https://github.com/libp2p/rust-libp2p/pull/2262/), your `NetworkBehaviour` hierarchy would be:

- `MyRootBehaviour`
  - `AutoNat`
    - `RequestResponse`
  - `Ping`

swarm/src/behaviour.rs
mxinden: > With the change to event_process = false, this might actually be a bad default now.

Agreed.

> In theory, we could have the derive auto-generate an enum but having that one be "invisible" could be very strange. 

I am not in favor of auto-generating the events enum. Instead, I would suggest making `event_process` a mandatory field to be set by the user.

swarm/src/behaviour.rs
mxinden: Good idea. Thanks. Added via https://github.com/libp2p/rust-libp2p/pull/2361/commits/4fdd7a966c2feaedf69fc92a054fbe20c5ed21f6.

swarm/src/behaviour.rs
mxinden: Let's continue the discussion in https://github.com/libp2p/rust-libp2p/issues/2373.

mxinden:
Thanks everyone for the input!

PR #2362, opened 2021-11-28
protocols/request-response: handle address change on connection

elenaf9:
Handle address change on a tracked connection in the `Request-Response` protocol.
See #2354.


mxinden (approved):
Thanks @elenaf9 :pray:.

PR #2363, opened 2021-11-28
core/src/transport: Add `Transport::dial_as_listener`

### LINKED ISSUE #2250: {swarm,core}/: Allow specifying the role of a connection in NetworkBehaviourAction::Dial
The _Direct Connection Upgrade through Relay_ protocol (https://github.com/libp2p/rust-libp2p/pull/2076) tries to establish a connection between two peers by synchronizing the two such that they can simultaneously attempt to establish a connection. Steps required by the two peers differ by transport protocol (e.g. TCP and QUIC) and role (client and server). See excerpt from the specification below:

> 5. Simultaneous Connect. The two nodes follow the steps below in parallel for
>    every address obtained from the `Connect` message:
>    - For a TCP address:
>       - Upon receiving the `Sync`, `A` immediately dials the address to `B`.
>       - Upon expiry of the timer, `B` dials the address to `A`.
>       - This will result in a TCP Simultaneous Connect. For the purpose of all
>         protocols run on top of this TCP connection, `A` is assumed to be the
>         client and `B` the server.
>    - For a QUIC address:
>       - Upon receiving the `Sync`, `A` immediately dials the address to `B`.
>       - Upon expiry of the timer, `B` starts to send UDP packets filled with
>         random bytes to `A`'s address. Packets should be sent repeatedly in
>         random intervals between 10 and 200 ms.
>       - This will result in a QUIC connection where `A` is the client and `B` is
>         the server.

https://github.com/libp2p/specs/blob/master/relay/DCUtR.md#the-protocol

In the case of TCP `B` needs to act as a listener once a connection is successfully established. Note that the TCP stack will treat both `A` and `B` as dialers as the connection will be a simultaneous TCP connection.

In the case of `QUIC` `B` needs to send random bytes instead of a connection attempt and then once a connection is established needs to act as a listener just like in the case of TCP.

Code in and based on `libp2p-swarm` does not need to be concerned with these implementation details. Instead, a single flag on `NetworkBehaviourAction::Dial`, indicating to the `Swarm` and `libp2p-core` that the conection to be established should be treated from the role of a listener, is sufficient.

To reduce conflicts, I would suggest tackling this once https://github.com/libp2p/rust-libp2p/issues/2249 is fixed.

mxinden:
Allows `NetworkBehaviour` implementations to dial a peer, but instruct
the dialed connection to be upgraded as if it were the listening
endpoint.

This is needed when establishing direct connections through NATs and/or
Firewalls (hole punching). When hole punching via TCP (QUIC is different
but similar) both ends dial the other at the same time resulting in a
simultaneously opened TCP connection. To disambiguate who is the dialer
and who the listener there are two options:

1. Use the Simultaneous Open Extension of Multistream Select. See
   [sim-open] specification and [sim-open-rust] Rust implementation.

2. Disambiguate the role (dialer or listener) based on the role within
   the DCUtR [dcutr] protocol. More specifically the node initiating the
   DCUtR process will act as a listener and the other as a dialer.

This commit enables (2), i.e. enables the DCUtR protocol to specify the
role used once the connection is established.

While on the positive side (2) requires one round trip less than (1), on
the negative side (2) only works for coordinated simultaneous dials.
I.e. when a simultaneous dial happens by chance, and not coordinated via
DCUtR, the connection attempt fails when only (2) is in place.

[sim-open]: https://github.com/libp2p/specs/blob/master/connections/simopen.md
[sim-open-rust]: https://github.com/libp2p/rust-libp2p/pull/2066
[dcutr]: https://github.com/libp2p/specs/blob/master/relay/DCUtR.md

Fixes https://github.com/libp2p/rust-libp2p/issues/2250.


core/src/connection.rs
```diff
@@ -193,12 +198,47 @@ impl ConnectedPoint {
...
+/// resulting in a _simultaneous open_ TCP connection. On this new
+/// connection both endpoints assume to be the dialer of the connection.
+/// This is problematic during the connection upgrade process where an
+/// upgrade assumes one side to be the listener. With the help of this
+/// option, both peers can negotiate the roles (dialer and listener) for
+/// the new connection ahead of time, through some external channel, and
+/// thus have one peer dial the other as a dialer and one peer dial the
+/// other _as a listener_.
+#[derive(Debug, Default, Copy, Clone, Hash, Eq, PartialEq)]
+pub struct DialAsListener(bool);
```
mxinden: I introduced this newtype instead of passing a `bool` through the various layers.

I like the fact that it (a) increases type safety and (b) carries its documentation. Unfortunately it is not particularly intuitive, nor idiomatic (never seen a newtype around a `bool`) nor convenient (e.g. see `if as_listener == false`).

@thomaseizinger you usually have good ideas when it comes to idiomatic Rust. Any alternatives that come to your mind?

thomaseizinger: What do you think of:

```rust
    Dialer {
        /// An optional override if the desired role for the upgrade process has already been negotiated out of band.
		////
		/// If this is set, the endpoint resulting from a successful dial should assume the given role instead of inferring it from the underlying transport.
        role_override: Option<Endpoint>,
    },
```

We could be even more descriptive and do:

```rust
enum UpgradeRole {
	InferredFromTransport,
	NegotiatedOutOfBand {
		role: Endpoint
	}
}
// -------------
Dialer {
    /// The strategy, how the role within the upgrade process should be determined.
    upgrade_role: UpgradeRole,
},
```

mxinden:
```rust
    Dialer {
        /// An optional override if the desired role for the upgrade process has already been negotiated out of band.
		////
		/// If this is set, the endpoint resulting from a successful dial should assume the given role instead of inferring it from the underlying transport.
        role_override: Option<Endpoint>,
    },
```

This is a good idea, especially as it does not introduce a new type. Though it does allow for duplication of information and thus potentially confusion? `PendingPoint::Dialer { role_override: Some(Endpoint::Dialer) }` is the same as `PendingPoint::Dialer { role_override: None }`.

mxinden:
```rust
enum UpgradeRole {
	InferredFromTransport,
	NegotiatedOutOfBand {
		role: Endpoint
	}
}
```

Note that this whole out-of-band role negotiation does not only concern upgrades, but some transport protocols as well. The TCP transport implementation does not require this information as it will open a simultaneous open TCP connection either way. But QUIC does. In the case of QUIC the listener sends a random-garbage packet instead of a connection-initiation packet.

   > - For a QUIC address:
   >    - Upon receiving the `Sync`, `A` immediately dials the address to `B`.
   >    - Upon expiry of the timer, `B` starts to send UDP packets filled with
   >      random bytes to `A`'s address. Packets should be sent repeatedly in
   >      random intervals between 10 and 200 ms.
   >    - This will result in a QUIC connection where `A` is the client and `B` is
   >      the server.


https://github.com/libp2p/specs/blob/master/relay/DCUtR.md#the-protocol

mxinden: Back to your suggestion of `role_override`, how would the signatures of `Network::dial` and `Peer::dial` look like?

```
    pub fn dial(
        &mut self,
        address: &Multiaddr,
        handler: THandler,
        role_override: Option<Endpoint>,
    ) -> Result<ConnectionId, DialError<THandler>>
```

I find this descriptive but again confusing due to the duplication of information where `Some(Endpoint::Dialer)` is the same as `None`.

mxinden:
@thomaseizinger I gave this another shot with your previous input. I did the following changes:

- Use `Endpoint` instead of `DialAsListener` where `Endpoint::Dialer` is
  the default and `Endpoint::Listener` represents the override.
- Use `override_role` instead of `as_listener`.
- Extend `Transport` through new method `dial_with_role_override`.

Could you take another look?

core/src/transport.rs
```diff
@@ -130,6 +130,19 @@ pub trait Transport {
     where
         Self: Sized;
 
+    /// As [`Transport::dial`] but overrides the role of the local node on the connection . I.e. has
+    /// the local node act as a listener on the outgoing connection.
+    ///
+    /// This option is needed for NAT and firewall hole punching.
+    ///
+    /// See [`ConnectedPoint::Dialer`](crate::connection::ConnectedPoint::Dialer) for related option.
+    fn dial_with_role_override(
```
thomaseizinger: I think on this level, it might be more descriptive to use a terminology like `dial_as_listener`? There is no overriding parameter here so not having been involved this discussion, I would be thinking "overridden with what?"

core/src/transport/and_then.rs
```diff
@@ -76,7 +76,35 @@ where
             .map_err(|err| err.map(EitherError::A))?;
         let future = AndThenFuture {
             inner: Either::Left(Box::pin(dialed_fut)),
-            args: Some((self.fun, ConnectedPoint::Dialer { address: addr })),
+            args: Some((
+                self.fun,
+                ConnectedPoint::Dialer {
+                    address: addr,
+                    role_override: Endpoint::Dialer,
+                },
```
thomaseizinger, line 84: This is indeed a bit awkward. I am not sure I completely understand this:

> Note that this whole out-of-band role negotiation does not only concern upgrades, but some transport protocols as well. The TCP transport implementation does not require this information as it will open a simultaneous open TCP connection either way.

From your PR description, what we are trying to achieve is to explicitly instruct the upgrade phase of a connection, which role to use instead of inferring it from somewhere. If that is true, perhaps we should rename `role_override` to `upgrade_role`. That naming would make the default case a bit less awkward as we are simply always stating the role explicitly.

What do you think?

thomaseizinger (commented):
Sorry for the late reply!

core/src/transport/and_then.rs
mxinden: In the hope to resolve some confusion: The `role_override` passed through `Network::dial` is relevant in two places:

1. in `AndThen` `Transport` and later `upgrade::apply` instructing to either use the `OutboundUpgrade::upgrade_outbound` (default case, no override) or `InboundUpgrade::upgrade_inbound` (override).
2. In certain `Transport` implementations. E.g. `libp2p-quic`, when `role_override` is set and thus `Transport::dial_with_role_override` (or `Transport::dial_as_listener`) is called the local node sends a garbage UDP packet instead of a handshake packet, to punch a hole through its NAT / firewall only.

---

So would you keep `role_override` on `Network::dial`, but rename to `upgrade_role` in `ConnectedPoint::Dialer` and `dial_as_listener` in `Transport`? With 3 different names for related features, would that not be even more confusing?

thomaseizinger: Okay, thanks for giving me more context. I think I would indeed choose a name for each component that makes sense within that scope. For the upgrade phase, an explicit role parameter is pretty clear IMO.

For the transports, wouldn't it be nice if we could use some kind of "back-channel" to feed in this information? We could add a function like this to the transport interface (name subject to bike-shedding):

```rust
trait Transport {
	fn register_hole_punching_attempt(&mut self, address: Multiaddr) { }
}
```

A transport that cares about this (QUIC), can override the function and upon `dial` check whether or not this address need to be handled in a special way. A transport that doesn't care, like TCP, can just not override this function.

Perhaps the entire feature can implement in such a back-channel way? Sorry for going into so many different directions here, I am just brain storming for better solutions because now that I think I understand the overall problem, I don't really like my original suggestion that much :D

mxinden: > For the transports, wouldn't it be nice if we could use some kind of "back-channel" to feed in this information? We could add a function like this to the transport interface (name subject to bike-shedding):

Do I understand your suggestion correctly that protocols like `libp2p-dcutr` would emit a `NetworkBehaviourAction::Dial` with some hole-punching-marker set, this would propagate to `ConcurrentDial` where one would then first call `Transport::register_hole_punching_attempt` and then call `Transport::dial`.

Why associate the hole-punching-marker with a `Multiaddr` instead of a dial attempt (call to `Swarm::dial`, `Network::dial`, `Transport::dial`). In other words, why force a `Transport` to keep track of a `Multiaddr`-to-hole-punching-marker mapping when we can as well just provide it with each call to `Transport::dial` as a parameter, or `Transport::dial_with_role_override`, or `Transport::dial_as_listener`?

> Sorry for going into so many different directions here, I am just brain storming for better solutions because now that I think I understand the overall problem, I don't really like my original suggestion that much :D

Very much appreciated. Exactly why I asked for your review.

mxinden: Also note that `Combinator` `Transport` implementations like `AndThen` would need to keep track of the `Multiaddr` to hole-punching-marker as well, given that they need to forward the right `ConnectedPoint` to the upgrade provided with `Transport::and_then`, e.g. either a `ConnectedPoint::Dialer { role_override: Endpoint::Dialer }` or a `ConnectedPoint::Dialer { role_override: Endpoint::Listener }`.

See below for how this is solved currently in this pull request. Note the different `role_override` values.

https://github.com/libp2p/rust-libp2p/blob/5d4d27166bc64dcf5870f4a6cf9e581cbb24366f/core/src/transport/and_then.rs#L72-L111

thomaseizinger: > Why associate the hole-punching-marker with a `Multiaddr` instead of a dial attempt (call to `Swarm::dial`, `Network::dial`, `Transport::dial`). In other words, why force a `Transport` to keep track of a `Multiaddr`-to-hole-punching-marker mapping when we can as well just provide it with each call to `Transport::dial` as a parameter, or `Transport::dial_with_role_override`, or `Transport::dial_as_listener`?

The idea was to leave the interface that serves 90% of the usecases untouched and model the special case using a dedicated interface. That way, we are

a) not committing to an interface now that might turn out to not be ideal
b) not trying to force a special case into the standard flow

I think some of the awkwardness in handing this data through all the layers is coming from the fact that feature of hole-punching is somewhat unnatural and requires a fair bit of coordination between several components. Thus, it is hard to find abstractions that more generally make sense (like role-override) and yet still support the feature.

Hence the idea of: If we can't find abstractions that we are happy with, lets just not make any new abstractions and create new, dedicated APIs instead.

> Do I understand your suggestion correctly that protocols like `libp2p-dcutr` would emit a `NetworkBehaviourAction::Dial` with some hole-punching-marker set, this would propagate to `ConcurrentDial` where one would then first call `Transport::register_hole_punching_attempt` and then call `Transport::dial`.

I haven't thought it fully through and it is somewhat annoying that we still have to touch the `NetworkBehaviourAction::Dial` API. But yes, I think we will have to adapt that API to somehow carry this information down.

> Also note that `Combinator` `Transport` implementations like `AndThen` would need to keep track of the `Multiaddr` to hole-punching-marker as well, given that they need to forward the right `ConnectedPoint` to the upgrade provided with `Transport::and_then`, e.g. either a `ConnectedPoint::Dialer { role_override: Endpoint::Dialer }` or a `ConnectedPoint::Dialer { role_override: Endpoint::Listener }`.

For those combinators, the function would just call through, the same way `address_translation` does.

mxinden: > > Also note that `Combinator` `Transport` implementations like `AndThen` would need to keep track of the `Multiaddr` to hole-punching-marker as well, given that they need to forward the right `ConnectedPoint` to the upgrade provided with `Transport::and_then`, e.g. either a `ConnectedPoint::Dialer { role_override: Endpoint::Dialer }` or a `ConnectedPoint::Dialer { role_override: Endpoint::Listener }`.
> 
> For those combinators, the function would just call through, the same way `address_translation` does.

But `AndThen` calls both `self.transport.dial` and `self.fun`. For the former it can "just call through", for the latter it needs to pass the hole-punching flag to `self.fun`, thus has to cache calls to `Transport::register_hole_punching_attempt`.

mxinden: > The idea was to leave the interface that serves 90% of the usecases untouched and model the special case using a dedicated interface. That way, we are

What do you think of the current approach, implementing a middle ground? Instead of altering `Transport::dial` which "serves 90% of the usecases", it adds `Transport::dial_with_role_override`. `Transport`s that don't care can simply delegate to `Transport::dial`.

thomaseizinger: > 
> > The idea was to leave the interface that serves 90% of the usecases untouched and model the special case using a dedicated interface. That way, we are
> 
> What do you think of the current approach, implementing a middle ground? Instead of altering `Transport::dial` which "serves 90% of the usecases", it adds `Transport::dial_with_role_override`. `Transport`s that don't care can simply delegate to `Transport::dial`.

Yeah that is fair!

I am not yet convinced that we need to follow the same naming on all levels. I think I'd prefer a name that is more descriptive for the individual functionality and composed together, they allow for hole punching then.

thomaseizinger: For the transport, what do you think of a name like: `Transport::dial_for_nat_traversal` or `Transport::dial_nat_traversing`.

The reason I'd advocate for such a naming is because the transport doesn't know what happens to the connection once it is established.

mxinden: For NAT traversal both endpoints dial each other, but only one needs to call `Transport::dial_for_nat_traversal`, in other words, only one needs to dial _as a listener_, in other words only one needs to dial _with a role override_. The name `dial_for_nat_traversal` would thus be confusing, as one would expect both nodes to use `dial_for_nat_traversal` for a NAT traversal, no?

mxinden:
Updated to use `DialOpts` introduced through #2404.

core/src/transport/and_then.rs
thomaseizinger: Yeah fair point!

I think I've come full circle on this and `dial_as_listener` is my preferred option now :sweat_smile: 
Sorry about the long back and forth that ended up in the original idea!

What do you think:

1. Use `dial_as_listener` in `Transport`
2. Using the "override" notation within the declarative `PendingPoint` struct

This would result in a piece of code that is something like:

```
if role_override == Listener {
	dial_as_listener
} else {
	dial
}
```

Which is quite readable IMO.

mxinden: > I think I've come full circle on this and dial_as_listener is my preferred option now sweat_smile
Sorry about the long back and forth that ended up in the original idea!

:rocket: maybe a [_strange loop_](https://en.wikipedia.org/wiki/Strange_loop) :innocent: ?

af10530 does the rename. Note that `libp2p_swarm::DialOpts` and `libp2p_core::DialOpts` use `override_role`. Let me know in case you would like that to be changed to `as_listener` as well.

See:

- https://github.com/mxinden/rust-libp2p/blob/618874c74fdf81679ae74f0fd5c3e89f20e6a140/swarm/src/dial_opts.rs#L143-L152
- https://github.com/mxinden/rust-libp2p/blob/618874c74fdf81679ae74f0fd5c3e89f20e6a140/core/src/network.rs#L661-L670

thomaseizinger (approved):
LGTM!

mxinden:
Thanks @thomaseizinger for the help!

thomaseizinger:
> Thanks @thomaseizinger for the help!

Thank you for your patience! :)

PR #2365, opened 2021-12-03
*: Fix clippy errors from upgrade to Rust 1.57

thomaseizinger:
The newly released version of clippy is flagging several things.

This PR is an attempt to tackle these quickly to unblock CI for other PRs.


mxinden (approved):
Thanks @thomaseizinger!

PR #2368, opened 2021-12-03
.github/workflow: Use ubuntu-18.04 fixing missing protoc binary 

mxinden:
Debug CI failure on missing protoc binary.

Caused by:
  process didn't exit successfully:
`/home/runner/work/rust-libp2p/rust-libp2p/target/debug/build/libp2p-core-31f9552cf2db9053/build-script-build`
(exit status: 101)
  --- stderr
  thread 'main' panicked at 'called `Result::unwrap()` on an `Err`
value: Custom { kind: NotFound, error: "failed to invoke protoc (hint:
https://docs.rs/prost-build/#sourcing-protoc): No such file or directory
(os error 2)" }', core/build.rs:30:6

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,6 +26,7 @@ jobs:
 
     - name: Check protoc
       run: |
+        ls -la /lib64/ld-linux-x86-64.so.2
         echo $PROTOC
         echo $PROTOC_INCLUDE
         whereis protoc
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,6 +24,11 @@ jobs:
       with:
         access_token: ${{ github.token }}
 
+    - name: Check protoc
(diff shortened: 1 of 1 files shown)


mxinden:
Still not quite sure why this is failing. Might be the `Virtual Environment` `Version` change (`Version: 20211122.1` -> `Version: 20211129.1`).

Interpreter `/lib64/ld-linux-x86-64.so.2` is present.



https://github.com/tokio-rs/prost/blob/d6faf5a0690826bbe07fe2c7e29bbe29d1eb0534/prost-build/build.rs#L38-L54

mxinden:
I will merge here, since this issue is blocking all other pull requests. I am happy to revisit this fix, especially in case anyone has alternative suggestions. Also I will create a reminder to check back in some months whether the bundled `protoc` binaries have been updated in `prost-build`.

Issue #2329, opened 2021-11-08
core/: Report pending connection being aborted


mxinden:
Today when aborting a pending connection, we remove the pending connection from `Pool::pending` directly:

https://github.com/libp2p/rust-libp2p/blob/8b68026bb04a469654020a8dcfe68308404e2510/core/src/connection/pool.rs#L974-L978

Later on, when the task reports that it has been aborted, the `Pool` checks if it has an entry in `Pool::pending`, which it does not, thus dropping the error instead of reporting it.

https://github.com/libp2p/rust-libp2p/blob/8b68026bb04a469654020a8dcfe68308404e2510/core/src/connection/pool.rs#L894-L899

We should instead report the pending connection being aborted by emitting an event.

JerryHue:
@mxinden, hello! Could I try to look into this?

mxinden:
@JerryHue sure thing. Help here is very much appreciated.

Unit test should be fairly simple, checking for a `PendingConnectionError::Aborted` after aborting a pending connection.

MarcoPolo:
@JerryHue have you started working on this? If not, do you mind if I pick this up? Thanks!

JerryHue:
Hello, @MarcoPolo! I will start to work on this today. If by the end of the day I couldn't advance much, I will delegate it to you. Is that alright?

MarcoPolo:
Of course!

JerryHue:
@MarcoPolo, hello! I read through the code, but I feel like this might be a little bit too much for me, especially since I am not familiar with this part of the core (I only worked with `PeerId`s, which arguably, were an easier concept for me).

Sorry for the inconvenience, and thank you for the prompt responses, @mxinden and @MarcoPolo.

MarcoPolo:
@mxinden could you assign this to me please?

jmmaloney4:
@MarcoPolo are you working on this? Do you mind if I take a stab?

MarcoPolo:
yep! please take!

jmmaloney4:
Hi @mxinden, I have a few questions about the code.

Will both of these be needed?

https://github.com/libp2p/rust-libp2p/blob/245b0563ab41a37b52d1d45ddb59f8706e130a53/core/src/connection/pool.rs#L978-L981

https://github.com/libp2p/rust-libp2p/blob/245b0563ab41a37b52d1d45ddb59f8706e130a53/core/src/connection/pool.rs#L902-L904

[...]

PR #2369, opened 2021-12-07
Report aborted connection from Pool::poll

jmmaloney4:
First attempt at fixing #2329.


core/src/connection/pool.rs
```diff
@@ -50,7 +50,8 @@ use std::{
     task::Context,
     task::Poll,
 };
-use void::Void;
+
+use self::task::PendingConnectionCommand;
```
MarcoPolo: nit, I would not `use ...` this directly and instead follow the same pattern as EstablishedConnectionCommand referencing it by the `task` namespace. E.g. `task::PendingConnectionCommand`

MarcoPolo (commented):
Looks good!

Bonus points if you add a test case to this :) 
Maybe something like:
1. create new pool
2. call pool.add_outgoing(...) with a transport (see examples)
3. pool.disconnect
4. pool.pool and see if you see a `PoolEvent::PendingOutboundConnectionError`
(I haven't tried this, so it may not work)

jmmaloney4:
I tried adding a test case, but I'm not sure how to call the `Pool::poll` method without an async context. What would be the right way to do that in a test case?

MarcoPolo:
Nice work @jmmaloney4 I filled in some of the missing pieces here: https://github.com/libp2p/rust-libp2p/commit/0c61938dae44710b003f303824a115f4e86c7b27.

I had to move the test to be within the module since the `pool` module is `pub(crate)` (so not usable from an external crate like those test).

Feel free to cherry pick that commit :)

jmmaloney4:
Thanks! I took your code and moved it into my branch, and I also tried to make the TestHandler implementations shared, but I'm not sure if I did that in an appropriate way--I had to make the util module `pub` here:

https://github.com/libp2p/rust-libp2p/blob/a3ff8e2ef32d6292e18b305530136df249ca0376/core/src/connection.rs#L25

core/src/connection/pool/task.rs
```diff
@@ -141,22 +148,24 @@ pub async fn new_for_pending_outgoing_connection<TTrans>(
...
+    abort_receiver: oneshot::Receiver<PendingConnectionCommand>,
     mut events: mpsc::Sender<PendingConnectionEvent<TTrans>>,
 ) where
     TTrans: Transport,
     TFut: Future<Output = Result<TTrans::Output, TTrans::Error>> + Send + 'static,
 {
-    match futures::future::select(drop_receiver, Box::pin(future)).await {
+    match futures::future::select(abort_receiver, Box::pin(future)).await {
         Either::Left((Err(oneshot::Canceled), _)) => {
+            unreachable!("Pool never drops channel to task.");
```
thomaseizinger: This is not quite true. It is dropped immediately after `abort` finishes because we are consuming `self` in there.

I've read through #2329 but I don't quite understand why we need to send a dedicated value now and the drop notifier isn't enough?

1. We always initialize `abort_receiver` with `Some` so there is never a case where we call `abort` and wouldn't be able to send `PendingConnectionCommand::Abort`.
1. We are still dropping the entire struct at the end of `abort` so all the data is gone anyway.
1. We are not sending any information along with `PendingConnectionCommand::Abort`.

Unless I am missing something, reacting to `Canceled` should be equivalent to what we currently have here? The `abort` method can just consume `self` and do nothing with it which will immediately drop it and trigger the drop notifier. That will still be the same public API albeit it may be worth considering removing `abort` and using `std::mem::drop` instead if we want to be explicit about the dropping.

core/src/connection/util.rs
```diff
@@ -0,0 +1,38 @@
...
+
+use multiaddr::Multiaddr;
+
+use crate::muxing::StreamMuxerBox;
+
+use super::{ConnectionHandler, Substream, SubstreamEndpoint, ConnectionHandlerEvent};
+
+
+#[derive(Debug)]
+pub struct TestHandler();
```
thomaseizinger: Instead of making this handler public, can we move the test you added to `tests/` and hence reuse the handler from there? I don't think we should be adding this one to the public about of `libp2p-core`.

core/src/connection/pool/task.rs
```diff
@@ -39,11 +39,16 @@ use futures::{
...
+/// Commands that can be sent to a task driving a pending connection.
 #[derive(Debug)]
-pub enum Command<T> {
+pub enum PendingConnectionCommand {
+    Abort,
+}
+
+/// Commands that can be sent to a task driving an established connection.
+#[derive(Debug)]
+pub enum EstablishedConnectionCommand<T> {
```
thomaseizinger: If my assumption above is correct, I'd vote for undoing this rename to reduce the noise of the diff.

thomaseizinger (changes requested):
Thanks!

Left some comments inline.

core/src/connection/util.rs
thomaseizinger: To resolve the issue about `Pool` being `pub(crate)`, I'd suggest to write the unit test against the `Network` component (which is the public API of `libp2p-core`).

The structure should roughly be the same:

1. Make a new `Network`
2. `dial` a peer
3. Retrieve it with `peer`
4. Convert it into a `DialingPeer`
5. `disconnect` the `DialingPeer`
6. Observe the aborted event

core/src/connection/pool/task.rs
MarcoPolo: > This is not quite true. It is dropped immediately after abort finishes because we are consuming self in there.

Maybe there's a misunderstanding. The thing that gets dropped is `PendingConnection`, the thing that has the abort_notifier is `PendingConnectionInfo`. The `PendingConnectionInfo` stays around in the hash map of the pool.

Whether you actually need to send abort vs just taking the notfier and dropping it. I think you can do the same thing by dropping the notifier. The reason you may want the abort value here is to make the distinction between this was dropped and we aborted this. Although with the `unreachable` we're saying we would never drop this (I'm not fully sure that's true, I need to look more closely).

core/src/connection/util.rs
MarcoPolo: Oh! good idea, thanks!

core/src/connection/pool/task.rs
thomaseizinger: > > This is not quite true. It is dropped immediately after abort finishes because we are consuming self in there.
> 
> Maybe there's a misunderstanding. The thing that gets dropped is `PendingConnection`, the thing that has the abort_notifier is `PendingConnectionInfo`. The `PendingConnectionInfo` stays around in the hash map of the pool.

Ah what a subtle difference, I missed that, thank you!

> The reason you may want the abort value here is to make the distinction between this was dropped and we aborted this. Although with the `unreachable` we're saying we would never drop this (I'm not fully sure that's true, I need to look more closely).

That is what my closer look was triggered by! We are adding a way of distinguishing the two and then asserting that one of them never happens. So why add that in the first place?

It seems to be sufficient to change the `oneshot::Sender<Void>` to an `Option<oneshot::Sender<Void>>` so we can `take` and `mem::drop` it which will trigger the cancellation. That should avoid the `unreachable` within the `select`.

MarcoPolo:
@jmmaloney4 added the suggestions from @thomaseizinger in my branch here: https://github.com/libp2p/rust-libp2p/compare/master...MarcoPolo:marco/fix-2329-v2?expand=1 (I can't push to this branch). Feel free to cherry pick those commits here.

mxinden:
Friendly ping @jmmaloney4.

mxinden:
There have been some larger changes (#2492). To make things easier on your end I resolved the conflicts. As far as I can there is nothing else missing, thus I am closing here with the goal of merging via https://github.com/libp2p/rust-libp2p/pull/2517. 

Please leave any comments on https://github.com/libp2p/rust-libp2p/pull/2517 in case I am missing something.

Thanks @jmmaloney4 @MarcoPolo and @thomaseizinger for the work!

PR #2383, opened 2021-12-14
protocols/mdns: Support multiple network interfaces

dvc94ch:
so had some issues with mdns between my laptop and my phone when android was acting as a hotspot. this is due to binding to 0.0.0.0 which selects the iface with internet connectivity leading to failure for the devices to detect each other. this fixes that by properly handling multiple interfaces in mdns. the socket logic was moved into an instance while the mdns behaviour watches for interface changes and creates new instances with a dedicated send/recv socket.


protocols/mdns/src/instance.rs
```diff
@@ -0,0 +1,228 @@
...
+use std::{
+    collections::VecDeque,
+    io, iter,
+    net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr, UdpSocket},
+    pin::Pin,
+    task::Context,
+    time::{Duration, Instant},
+};
+
+/// A `NetworkBehaviour` for mDNS. Automatically discovers peers on the local network and adds
```
rklaehn: So this is basically the code from before, per interface. Not sure if instance is a good name, it is a bit generic. Maybe InterfaceState or something? But in any case the doc comments need updating.

protocols/mdns/src/instance.rs
```diff
@@ -0,0 +1,228 @@
...
+    recv_socket: Async<UdpSocket>,
+    /// Send socket.
+    send_socket: Async<UdpSocket>,
+    /// Buffer used for receiving data from the main socket.
+    /// RFC6762 discourages packets larger than the interface MTU, but allows sizes of up to 9000
+    /// bytes, if it can be ensured that all participating devices can handle such large packets.
+    /// For computers with several interfaces and IP addresses responses can easily reach sizes in
+    /// the range of 3000 bytes, so 4096 seems sensible for now. For more information see
+    /// [rfc6762](https://tools.ietf.org/html/rfc6762#page-46).
+    recv_buffer: [u8; 4096],
```
rklaehn: This is not new code, but just copied from the previous global state.

protocols/mdns/tests/smoke.rs
```diff
@@ -146,39 +133,12 @@ async fn test_expired_async_std_ipv4() -> Result<(), Box<dyn Error>> {
         .map_err(|e| Box::new(e) as Box<dyn Error>)
 }
 
-#[async_std::test]
```
rklaehn, line 149: Why can these tests be removed without replacement?

protocols/mdns/tests/smoke.rs
```diff
@@ -146,39 +133,12 @@ async fn test_expired_async_std_ipv4() -> Result<(), Box<dyn Error>> {
...
 #[tokio::test]
-async fn test_expired_tokio_ipv6() -> Result<(), Box<dyn Error>> {
+async fn test_expired_tokio() -> Result<(), Box<dyn Error>> {
+    env_logger::try_init().ok();
     let config = MdnsConfig {
-        ttl: Duration::from_millis(500),
-        query_interval: Duration::from_secs(1),
-        multicast_addr: *IPV6_MDNS_MULTICAST_ADDRESS,
+        ttl: Duration::from_secs(1),
+        query_interval: Duration::from_secs(10),
```
rklaehn, line 160: Why is the timeout increased?

protocols/mdns/src/instance.rs
```diff
@@ -0,0 +1,228 @@
...
+                }
+            }
+            MdnsPacket::ServiceDiscovery(disc) => {
+                let resp = build_service_discovery_response(disc.query_id(), self.ttl);
+                self.send_buffer.push_back(resp);
+            }
+        }
+    }
+
+    pub fn poll(
```
rklaehn: This is mostly unchanged from before, except that it is now per interface.

protocols/mdns/src/instance.rs
```diff
@@ -0,0 +1,228 @@
...
+                .now_or_never()
+            {
+                Some(Ok((len, from))) => {
+                    if let Some(packet) = MdnsPacket::new_from_bytes(&self.recv_buffer[..len], from)
+                    {
+                        self.inject_mdns_packet(packet, params);
+                    }
+                }
+                Some(Err(err)) => log::error!("Failed reading datagram: {}", err),
+                None => unreachable!(),
```
rklaehn: So this is impossible because we call is_ready() before?

protocols/mdns/src/instance.rs
dvc94ch: well since it's an implementation detail of mdns calling it MdnsInstance is superfluous. I don't think that it's too confusing.

protocols/mdns/src/instance.rs
dvc94ch: yes

protocols/mdns/tests/smoke.rs
dvc94ch: this makes sure that the ttl is much shorter than the query interval, as otherwise the expired event will never be emitted.

protocols/mdns/src/instance.rs
dvc94ch: yep

protocols/mdns/src/instance.rs
dvc94ch: exactly. although can revert as I was just trying to rule out some code paths during debugging

protocols/mdns/tests/smoke.rs
dvc94ch: fixed ipv6 support

dvc94ch:
ping @mxinden do you have some time to review? Think it's pretty reliable now.

protocols/mdns/tests/smoke.rs
rklaehn: So how long will the test run on average? Tests with long running timeouts are not great.

rklaehn: So how long will the test run on average? Tests with long running timeouts are not great.

dvc94ch: immediate

protocols/mdns/src/instance.rs
rklaehn: I think it would be nice to indicate that this is the state *per interface*. The fact that it is a part of MDns does not need to be in the name, you are right about that.

rklaehn (approved):

transports/tcp/Cargo.toml
```diff
@@ -14,7 +14,7 @@ categories = ["network-programming", "asynchronous"]
 async-io-crate = { package = "async-io", version = "1.2.0", optional = true }
 futures = "0.3.8"
 futures-timer = "3.0"
-if-watch = { version = "0.2.0", optional = true }
+if-watch = { version = "0.3.0", optional = true }
```
mxinden: Done via https://github.com/libp2p/rust-libp2p/pull/2387.

protocols/mdns/src/iface.rs
```diff
@@ -0,0 +1,238 @@
...
+    collections::VecDeque,
+    io, iter,
+    net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr, UdpSocket},
+    pin::Pin,
+    task::Context,
+    time::{Duration, Instant},
+};
+
+/// An mDNS instance for a networking interface. To discover all peers when having multiple
+/// interfaces an instance is required for each interface.
```
mxinden:
```suggestion
/// interfaces an [`InterfaceState`] is required for each interface.
```

protocols/mdns/src/iface.rs
```diff
@@ -0,0 +1,238 @@
...
+    /// Multicast address.
+    multicast_addr: IpAddr,
+    /// Discovered addresses.
+    discovered: VecDeque<(PeerId, Multiaddr, Instant)>,
+    /// TTL
+    ttl: Duration,
+}
+
+impl InterfaceState {
+    /// Builds a new `Instance`.
```
mxinden:
```suggestion
    /// Builds a new [`InterfaceState`].
```

mxinden (commented):
Listening and sending via multiple interfaces sounds good to me. Split into `InterfaceState` makes sense to me.

I will need a bit more time to give this an in-depth review.

protocols/mdns/src/behaviour.rs
```diff
@@ -18,79 +18,34 @@
...
-    /// bytes, if it can be ensured that all participating devices can handle such large packets.
-    /// For computers with several interfaces and IP addresses responses can easily reach sizes in
-    /// the range of 3000 bytes, so 4096 seems sensible for now. For more information see
-    /// [rfc6762](https://tools.ietf.org/html/rfc6762#page-46).
-    recv_buffer: [u8; 4096],
-
-    /// Buffers pending to send on the main socket.
-    send_buffer: VecDeque<Vec<u8>>,
+    /// Mdns interface states.
+    iface_states: FnvHashMap<IpAddr, InterfaceState>,
```
mxinden: I understand that using a `FnvHashMap` increases performance here. Though is it worth the additional dependency? How is this a bottleneck in the first place? A lookup should be on the order of 100 ns either way, no?

protocols/mdns/src/iface.rs
```diff
@@ -0,0 +1,238 @@
+// Copyright 2018 Parity Technologies (UK) Ltd.
```
mxinden: Should this file not rather live in `src/behaviour/iface.rs`? It is only used within `src/behaviour.rs`, right?

protocols/mdns/src/behaviour.rs
```diff
@@ -295,131 +120,98 @@ impl NetworkBehaviour for Mdns {
...
-                                log::trace!("leaving multicast on iface {}", addr);
-                                if let Err(err) = socket.leave_multicast_v6(&multicast, 0) {
-                                    log::error!("leave multicast failed: {}", err);
-                                }
+                    if let Entry::Vacant(e) = self.iface_states.entry(addr) {
+                        match InterfaceState::new(addr, self.config.clone()) {
+                            Ok(iface_state) => {
+                                e.insert(iface_state);
                             }
+                            Err(err) => log::error!("failed to create instance: {}", err),
```
mxinden:
```suggestion
                            Err(err) => log::error!("failed to create `InterfaceState`: {}", err),
```

protocols/mdns/src/behaviour.rs
```diff
@@ -295,131 +120,98 @@ impl NetworkBehaviour for Mdns {
...
+        });
+        if !expired.is_empty() {
+            let event = MdnsEvent::Expired(ExpiredAddrsIter {
+                inner: expired.into_iter(),
+            });
+            return Poll::Ready(NetworkBehaviourAction::GenerateEvent(event));
+        }
+        if let Some(closest_expiration) = closest_expiration {
+            let mut timer = Timer::at(closest_expiration);
+            Pin::new(&mut timer).poll(cx).is_pending();
```
mxinden: The return type is unused.

mxinden (commented):
Couple of comments and missing a changelog entry. Looks good to me otherwise. Thank you for the contribution.

transports/tcp/Cargo.toml
dvc94ch: guess it will disappear during the merge then

protocols/mdns/src/iface.rs
dvc94ch: not sure how adding more directories for a small crate helps. is there some "policy" for the rust-libp2p repo somewhere?

protocols/mdns/src/behaviour.rs
dvc94ch: removed it, but it's not really an additional dependency as libp2p-core uses it

protocols/mdns/src/iface.rs
mxinden: I am following the convention: If module B is used in module A only, move it into `a/b.rs`. I think we are doing a decent job following this convention.

protocols/mdns/src/behaviour.rs
mxinden: Thanks. I know. I think it is still worth removing.

protocols/mdns/src/iface.rs
dvc94ch: well, I don't like large nested repos. some directories make sense sometimes when organizing files, but if a crate has 4 files, I don't think it's helpful. Made the change (still open to reverting if you change your mind)

mxinden:
Mind running a final `rustfmt` @dvc94ch?

mxinden (approved):

Issue #2391, opened 2021-12-17
gossipsub peer sharing


Frederik-Baetens:
I'm trying to build a system where nodes connect to a set of known nodes to join the gossipsub network.

However, it seems like these peers don't reach out to eachother after establishing a connection to the known nodes.

Is there any way to get some sort of peer sharing & automatic connection handling?

I saw [this pr](https://github.com/libp2p/rust-libp2p/pull/1720) and it mentions peer eXchange being held up by [an addition to the spec](https://github.com/libp2p/specs/pull/217) which has been merged for a while now, but i'm not sure if that spec has been implemented yet.

An example of what i'm trying to do (adapted from the libp2p examples): https://github.com/Frederik-Baetens/libp2p-tokiochat

It allows starting the first node with `cargo run`. The following nodes are then connected with `cargo run  -- *multiaddr of first node*`

I'd like the additional nodes to find eachother, and for the chat application to keep working after the first node has been shut down. Currently I get an insufficientpeers error after the original node gets shut down.

Am I supposed to implement this functionality myself based on things like https://docs.rs/libp2p/latest/libp2p/core/struct.Network.html#method.connected_peers ? Or could this be made to work of the box?

Older issue about this: https://github.com/libp2p/rust-libp2p/issues/1559

mxinden:
Thanks for the detailed issue @Frederik-Baetens. Are you calling `GossipsubConfig::do_px`?

Frederik-Baetens:
I'm constructing my config as follows:

```
        // Set a custom gossipsub
        let gossipsub_config = gossipsub::GossipsubConfigBuilder::default()
            .heartbeat_interval(Duration::from_secs(10)) // This is set to aid debugging by not cluttering the log space
            .validation_mode(ValidationMode::Strict) // This sets the kind of message validation. The default is Strict (enforce message signing)
            .message_id_fn(message_id_fn) // content-address messages. No two messages of the
            // same content will be propagated.
            .build()
            .expect("Valid config");
```

So i think it should be enabled, since it says that it's enabled by default in the docs. I tried adding .do_px() explicitly to the configbuilder, but that doesn't seem to make a difference. Should I do something for peer exchange when calling the eventloop/handling gossipsub events?

Frederik-Baetens:
I've pushed an update to my example repository with the explicit do_px(). It still doesn't seem to work.

I made a small recording of what happens:

[...]

PR #2397, opened 2021-12-19
protocols/gossipsub/src/config: Document peer exchange default is false

mxinden:
```rust
impl Default for GossipsubConfigBuilder {
    fn default() -> Self {
        GossipsubConfigBuilder {
            config: GossipsubConfig {
                // ...
                do_px: false,
                // ...
            },
        }
    }
}
```

https://github.com/libp2p/rust-libp2p/blob/b6c82a499c185eae6af716a5ecae2a8c7a4b6476/protocols/gossipsub/src/config.rs#L418

Spotted through https://github.com/libp2p/rust-libp2p/issues/2391.

### DIFF
--- a/protocols/gossipsub/src/config.rs
+++ b/protocols/gossipsub/src/config.rs
@@ -248,7 +248,7 @@ impl GossipsubConfig {
     }
 
     /// Whether Peer eXchange is enabled; this should be enabled in bootstrappers and other well
-    /// connected/trusted nodes. The default is true.
+    /// connected/trusted nodes. The default is false.
     pub fn do_px(&self) -> bool {
         self.do_px
     }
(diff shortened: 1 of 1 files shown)


AgeManning:
Yep, this could be better documented. We currently don't support peer exchange because we don't (or didn't at the time) have a legitimate way to send peer-id and multiaddr. The specs say we are supposed to use a signed peer record, which didn't exist in rust-libp2p at the time. Maybe we have them now?

The relevant parts of the libp2p code where this is documented is here: https://github.com/libp2p/rust-libp2p/blob/master/protocols/gossipsub/src/behaviour.rs#L1554
and here:
https://github.com/libp2p/rust-libp2p/blob/master/protocols/gossipsub/src/behaviour.rs#L1573

So currently we are sending PeerId's as a Vec but we can't dial them when they are recieved. This is not to spec, which is why it is disabled by default. Even if enabled, rust-libp2p wont connect to new nodes.

mxinden:
Thanks for the update @AgeManning. Added https://github.com/libp2p/rust-libp2p/issues/2398 and corresponding references in the code. Does this look good to you?

AgeManning (approved):
LGTM

PR #2403, opened 2021-12-27
Unsub backoff spec changes

divagant-martian:
Implements the changes specified by https://github.com/libp2p/specs/pull/383 (to my best understanding)


mxinden:
//CC @Menduist

mxinden:
@AgeManning do you have some time to give this pull request a review?

protocols/gossipsub/src/behaviour.rs
```diff
@@ -1092,10 +1093,15 @@ where
         self.backoffs
```
AgeManning: Shouldn't we be updating our local store also, as well as telling the peer what the backoff time should be?

divagant-martian: I added a test that should check if the used and sent backoff are the unsubscribe backoff

divagant-martian:
@mxinden could you give CI a check? Not sure if related to the latest rust release, but _it works on my machine_. I tried clearing the cache too, still works

AgeManning (approved):
lgtm

PR #2404, opened 2021-12-28
{core,swarm}/: Allow configuring dial concurrency factor per dial

### LINKED ISSUE #2385: {core,swarm}: Allow configuring dial concurrency factor per dial
Today the dial concurrency factor can be set global and for once.

https://github.com/libp2p/rust-libp2p/blob/b6c82a499c185eae6af716a5ecae2a8c7a4b6476/swarm/src/lib.rs#L1241-L1245

I argue a `NetworkBehaviour` should be able to configure the factor for each dial individually.

While it is necessary to dial concurrently when hole punching to increase success rate, I don't think this should be the case for `libp2p-autonat` when probing someones addresses. In other words, to reduce the amount of work a remote node can make the local node do at a time, I think the local node should probe the remotes addresses in sequence and not in parallel. In order to be able to do so, `libp2p-autonat` needs to be able to set the dial concurrency factor to `0` for probe dials.

@elenaf9 would you agree with the above?

https://github.com/libp2p/rust-libp2p/pull/2363 already adds a new parameter to  `Network::dial`, namely `role_override`. Instead of adding yet another one, it might be time to have `Network::dial` take a `DialOpts` like e.g. `Swarm::dial` does.

mxinden:
Enable a `NetworkBehaviour` or a user via `Swarm::dial` to override the
dial concurrency factor per dial. This is especially relevant in the
case of libp2p-autonat where one wants to probe addresses in sequence to
reduce the amount of work a remote peer can force onto the local node.

To enable the above, this commit also:

- Introduces `libp2p_core::DialOpts` mirroring `libp2p_swarm::DialOpts`.
  Passed as an argument to `Network::dial`.
- Removes `Peer::dial` in favor of `Network::dial`.
- Simplifies `Swarm::dial_with_handler`.

The introduction of `libp2p_core::DialOpts` will be useful beyond this
feature, e.g. for https://github.com/libp2p/rust-libp2p/pull/2363.

In the long run I would like to move and merge `libp2p_core::Network`
and `libp2p_core::Pool` into `libp2p_swarm::Swarm` thus deduplicating
`libp2p_core::DialOpts` and `libp2p_swarm::DialOpts`.

Fixes #2385.

//CC @elenaf9


mxinden:
@thomaseizinger @elenaf9 @MarcoPolo does one of you have some time to give this a review?

MarcoPolo:
Yep! I'll tal tomorrow morning

core/src/network.rs
```diff
@@ -560,31 +557,122 @@ impl NetworkConfig {
...
+
+    /// Dial an unknown peer.
+    pub fn unknown_peer_id() -> WithoutPeerId {
+        WithoutPeerId {}
+    }
+}
+
+impl From<Multiaddr> for DialOpts {
+    fn from(address: Multiaddr) -> Self {
+        DialOpts::unknown_peer_id().address(address).build()
```
MarcoPolo, line 602: Should we check if the multiaddr ends with `/p2p/<peerid>` and use the withpeerid variant?

core/src/network/peer.rs
```diff
@@ -432,7 +403,7 @@ where
     THandler: IntoConnectionHandler,
 {
     peer_id: PeerId,
-    network: &'a mut Network<TTrans, THandler>,
```
MarcoPolo, line 435: Why this change? Are you planning on removing this field?

core/src/network.rs
MarcoPolo: ah, nvm. Handled above.

MarcoPolo (approved):

core/src/network/peer.rs
mxinden: Mostly for the sake of consistency. Each of the `XXXPeer`, e.g. `ConnectedPeer` and `DisconnectedPeer`, contain a `&mut Network`. Holding this mutable reference is relevant e.g. for `ConnectedPeer::some_connection`:

``` rust
    /// Obtains some established connection to the peer.
    pub fn some_connection(&mut self) -> EstablishedConnection<THandlerInEvent<THandler>> {
        self.connections()
            .into_first()
            .expect("By `Peer::new` and the definition of `ConnectedPeer`.")
    }
```

Given that `ConnectedPeer` is only created when the peer is actually connected and given that `ConnectedPeer` holds a `&mut Network` and thus a user could not disconnect the peer in the meantine, `ConnectedPeer::some_connection` can do the `expect` call.

Does that make sense @MarcoPolo?

> In the long run I would like to move and merge `libp2p_core::Network` and `libp2p_core::Pool` into  libp2p_swarm::Swarm` thus deduplicating `libp2p_core::DialOpts` and `libp2p_swarm::DialOpts`.

Following up on the above from the PR description, I would likely suggest getting rid of all of `peer.rs` in the same go. While I think the `Peer` construct offers a cool user experience, I don't think it makes sense for an rust-libp2p internal API.

MarcoPolo: Yeah, I think that makes sense. Even though we never use `_network` anywhere, we still want to keep it so that we know that nothing else has a `&mut` on `Network`. Is that correct?

mxinden: Correct.

Issue #2341, opened 2021-11-15
Types should eagerly implement common traits


elenaf9:
At the moment many config-, event- and error-types that are exposed to the user do not implement `Clone`, some not even `Debug`. This can be very inconvenient for the user and force them to create their own copy of the type just to derive the required traits. 
Following chapter [C-COMMON-TRAITS  of the Rust API Guidlines](https://rust-lang.github.io/api-guidelines/interoperability.html#c-common-traits):

> Rust's trait system does not allow orphans: roughly, every impl must live either in the crate that defines the trait or the implementing type. Consequently, crates that define new types should eagerly implement all applicable, common traits.

For rust-libp2p, I would suggest to at least derive `Clone`, `Debug` and `PartialEq` for every type. The latter mostly for convenience, quoting _Rust for Rustaceans_, Chapter 3 _Common Traits for Types_:

> The `PartialEq` trait is particularly desirable, because users will at some point inevitably have two instances of your type that they wish to compare with `==` or `assert_eq!`. Even if your type would compare equal for only the same instance of the type, it’s worth implementing `PartialEq` to enable your users to use `assert_eq!`.

Apart from these very basic traits, I would like to put up for discussion if it would make sense to also derive `serde::{Deserialize, Serialize}` for all types, quoting again _Rust for Rustaceans_:

> Finally, for most types, it makes sense to implement the serde crate’s Serialize and Deserialize traits. These can be easily derived, and the serde_derive crate even comes with mechanisms for overwriting the serialization for just one field or enum variant. Since serde is a third-party crate, you may not wish to add a required dependency on it. Most libraries therefore choose to provide a serde feature that adds support for serde only when the user opts into it.

At least in our project, this is quite relevant and it would make things easier to have them implemented.

Now I don't know how easy it is to derive `Serialize` and `Deserialize` for libp2p, especially for types that libp2p inherits from its dependencies, but if this is generally something that you'd approve, I would have a look at it.
And of course also `Clone`, `Debug` and `PartialEq`, which should imo added either way and not be much work.

thomaseizinger:
I agree with Clone, Debug and PartialEq, esp. for config and event types.

For serde I think we would have to decide on a case by case basis by generally happy to provide it, although I think it should be feature-flagged and off by default.

Which types would you like to serialize/deserialize?

[...]

PR #2405, opened 2021-12-28
core: Add Serialize and Deserialize to PeerId and MessageId

laptou:
- added a new feature `serde` to the `libp2p` crate
- when this feature is enabled, `Serialize` and `Deserialize` are implemented for `PeerId` and `gossipsub::MessageId`
- fixes an issue with `GossipsubConfigBuilder::build()` where it requires `&self` to be borrowed for `'static` lifetime due to lifetime elision

### DIFF
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -23,9 +23,16 @@ lazy_static = "1.2"
 libsecp256k1 = { version = "0.7.0", optional = true }
 log = "0.4"
 multiaddr = { version = "0.13.0" }
-multihash = { version = "0.14", default-features = false, features = ["std", "multihash-impl", "identity", "sha2"] }
+multihash = { version = "0.14", default-features = false, features = [
+  "std",
+  "multihash-impl",
+  "identity",
+  "sha2",
+] }
 multistream-select = { version = "0.11", path = "../misc/multistream-select" }
-p256 = { version = "0.10.0", default-features = false, features = ["ecdsa"], optional = true }
+p256 = { version = "0.10.0", default-features = false, features = [
+  "ecdsa",
+], optional = true }
 parking_lot = "0.11.0"
 pin-project = "1.0.0"
 prost = "0.9"
@@ -37,9 +44,13 @@ thiserror = "1.0"
 unsigned-varint = "0.7"
 void = "1"
 zeroize = "1"
+serde = { version = "1", optional = true }
 
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
-ring = { version = "0.16.9", features = ["alloc", "std"], default-features = false }
+ring = { version = "0.16.9", features = [
+  "alloc",
+  "std",
+], default-features = false }
 
 [dev-dependencies]
 async-std = { version = "1.6.2", features = ["attributes"] }
@@ -56,9 +67,9 @@ rand07 = { package = "rand", version = "0.7" }
 prost-build = "0.9"
 
 [features]
-default = [ "secp256k1", "ecdsa" ]
-secp256k1 = [ "libsecp256k1" ]
-ecdsa = [ "p256" ]
+default = ["secp256k1", "ecdsa"]
+secp256k1 = ["libsecp256k1"]
+ecdsa = ["p256"]
 
 [[bench]]
 name = "peer_id"
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -70,6 +70,7 @@ prost-build = "0.9"
 default = ["secp256k1", "ecdsa"]
 secp256k1 = ["libsecp256k1"]
 ecdsa = ["p256"]
+use-serde = ["multihash/serde-codec", "serde"]
 
 [[bench]]
 name = "peer_id"
--- a/core/src/peer_id.rs
+++ b/core/src/peer_id.rs
@@ -24,6 +24,9 @@ use rand::Rng;
 use std::{convert::TryFrom, fmt, str::FromStr};
 use thiserror::Error;
 
+#[cfg(feature = "serde")]
+use serde::{Serialize, Deserialize};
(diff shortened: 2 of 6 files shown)


core/Cargo.toml
```diff
@@ -56,9 +67,10 @@ rand07 = { package = "rand", version = "0.7" }
 prost-build = "0.9"
 
 [features]
-default = [ "secp256k1", "ecdsa" ]
-secp256k1 = [ "libsecp256k1" ]
-ecdsa = [ "p256" ]
+default = ["secp256k1", "ecdsa"]
+secp256k1 = ["libsecp256k1"]
+ecdsa = ["p256"]
+use-serde = ["multihash/serde-codec", "serde"]
```
thomaseizinger, line 73: We can avoid the awkward `use-serde` name by renaming the `serde` dependency like this:

```toml
serde_crate = { package = "serde", version = "1" }
```

and then do:

```rust
extern crate serde_crate as serde;
```

Inside `lib.rs`.

thomaseizinger (changes requested):
Thanks!

One comment on the feature name in `libp2p-core`!

core/Cargo.toml
```diff
@@ -23,9 +23,16 @@ lazy_static = "1.2"
...
+  "std",
+  "multihash-impl",
+  "identity",
+  "sha2",
+] }
 multistream-select = { version = "0.11", path = "../misc/multistream-select" }
-p256 = { version = "0.10.0", default-features = false, features = ["ecdsa"], optional = true }
+p256 = { version = "0.10.0", default-features = false, features = [
+  "ecdsa",
+], optional = true }
```
thomaseizinger, line 35: Can we undo this reformatting please?

core/Cargo.toml
```diff
@@ -37,9 +44,13 @@ thiserror = "1.0"
...
 void = "1"
 zeroize = "1"
+serde = { version = "1", optional = true }
 
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
-ring = { version = "0.16.9", features = ["alloc", "std"], default-features = false }
+ring = { version = "0.16.9", features = [
+  "alloc",
+  "std",
+], default-features = false }
```
thomaseizinger, line 53: Same here.

core/Cargo.toml
thomaseizinger: Also here, please undo the unnecessary re-formatting.

core/src/peer_id.rs
```diff
@@ -24,13 +24,17 @@ use rand::Rng;
...
+use serde::{Serialize, Deserialize};
+
 /// Public keys with byte-lengths smaller than `MAX_INLINE_KEY_LENGTH` will be
 /// automatically used as the peer id using an identity multihash.
 const MAX_INLINE_KEY_LENGTH: usize = 42;
 
 /// Identifier of a peer of the network.
 ///
 /// The data is a multihash of the public key of the peer.
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
```
thomaseizinger, line 37:
```suggestion
/// Public keys with byte-lengths smaller than `MAX_INLINE_KEY_LENGTH` will be
/// automatically used as the peer id using an identity multihash.
const MAX_INLINE_KEY_LENGTH: usize = 42;

/// Identifier of a peer of the network.
///
/// The data is a multihash of the public key of the peer.
#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
```

This would contain it to a single line.

protocols/gossipsub/Cargo.toml
```diff
@@ -32,6 +32,7 @@ pin-project = "1.0.8"
 instant = "0.1.11"
 # Metrics dependencies
 open-metrics-client = "0.13.0"
+serde = { version = "1", optional = true }
```
thomaseizinger, line 35: I think we should activate the `derive` feature here because we depend on it and at the moment, it seems to be carried along by something else but that might not always be true and can lead to compile errors.

thomaseizinger (commented):
Some more comments :)

core/Cargo.toml
mxinden: Referencing https://github.com/rust-lang/api-guidelines/discussions/180 here as a general Rust discussion. If I understand correctly, ideally we could use `serde` for both feature and crate, though that seems [not to be stable yet](https://github.com/rust-lang/cargo/issues/5565#issuecomment-412296421).

I don't have an opinion. I would go with whatever is most widely used within the Rust ecosystem.

mxinden:
//CC @elenaf9 as you proposed this in https://github.com/libp2p/rust-libp2p/issues/2341.

mxinden:
> fixes an issue with GossipsubConfigBuilder::build() where it requires &self to be borrowed for 'static lifetime due to lifetime elision

@laptou would you mind proposing this change in a separate pull request?

core/Cargo.toml
laptou: I did this b/c I've seen a `use-serde` feature on other crates. I tried @thomaseizinger's solution, but it seems that `serde` doesn't support being aliased, because when I tried aliasing `serde` as `serde_crate` or `_serde`, then I started getting this error on the derive macros:

```plain
error[E0463]: can't find crate for `serde`
  --> core/src/peer_id.rs:37:38
   |
37 | #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
   |                                      ^^^^^^^^^ can't find crate
   |
   = note: this error originates in the derive macro `Serialize` (in Nightly builds, run with -Z macro-backtrace for more info)
```

laptou:
I've created a new branch on my fork that has just the changes relevant to adding Serde, but I can't change the branch associated with this PR, so I will close it and open another one.

PR #2406, opened 2021-12-28
use SystemTime instead of Instant

laptou:
This branch is based on the `master` branch in my fork, so it also contains the changes from #2405.
In this PR, I changed all instances of `Instant` to `SystemTime` in the Kademlia behaviour, because I am trying to make a `RecordStore` that serializes the records to an SQLite database, but `Instant`s are extremely resistant to serialization. For now, I have only modified the structures used by Kademlia and their transitive dependencies, so there are still many uses of `Instant` in the ping, mDNS, and gossipsub behaviours.


protocols/kad/src/behaviour.rs
```diff
@@ -50,6 +49,7 @@ use libp2p_swarm::{
 use log::{debug, info, warn};
 use smallvec::SmallVec;
 use std::collections::{BTreeMap, HashSet, VecDeque};
+use std::time::SystemTime;
```
thomaseizinger, line 52: IF we end up merging this, we need to use `instant::SystemTime` here.

protocols/kad/src/behaviour.rs
```diff
@@ -697,7 +697,7 @@ where
         let inner = QueryInner::new(info);
         let id = self.queries.add_iter_closest(target.clone(), peers, inner); // (*)
 
-        // Instantly finish the query if we already have enough records.
+        // SystemTimely finish the query if we already have enough records.
```
thomaseizinger, line 700: This seems to be an erroneous change :D

protocols/kad/src/behaviour.rs
```diff
@@ -721,7 +721,7 @@ where
     /// After the initial publication of the record, it is subject to (re-)replication
     /// and (re-)publication as per the configured intervals. Periodic (re-)publication
     /// does not update the record's expiration in local storage, thus a given record
-    /// with an explicit expiration will always expire at that instant and until then
+    /// with an explicit expiration will always expire at that SystemTime and until then
```
thomaseizinger, line 724: This too.

thomaseizinger (commented):
I think we only use `Instant` across the codebase because `SystemTime` is a very recent [addition](https://github.com/sebcrozet/instant/pull/42).

I can't think of any downside on using `SystemTime` over `Instant` so feel free to go ahead with these changes :)

mxinden:
> RecordStore that serializes the records to an SQLite database

Great to see this happening @laptou!

> I think we only use Instant across the codebase because SystemTime is a very recent addition.

`SystemTime` is not guaranteed to be monotonic:

> Distinct from the Instant type, this time measurement is not monotonic. This means that you can save a file to the file system, then save another file to the file system, and the second file has a SystemTime measurement earlier than the first. In other words, an operation that happens after another operation in real time may have an earlier SystemTime!

https://doc.rust-lang.org/nightly/std/time/struct.SystemTime.html

I am not sure whether we rely on this property anywhere within the codebase.

> I am trying to make a RecordStore that serializes the records to an SQLite database

Would it not be enough to use `SystemTime` in the serialization logic only, i.e. transform an `Instant` to a `SystemTime` via `Duration` when serializing and deserializing a record?

tomaka:
Replacing `Instant` with `SystemTime` is most certainly wrong. At any point, the value returned by `SystemTime::now()` can change. Maybe one moment `SystemTime::now()` returns December 29th 2021, and the next moment it returns June 2nd 1995, you can't know.

In order to serialize an `Instant`, one proper way is to do something like:

```rust
let now = Instant::now();
let now_system_time = SystemTime::now();
if instant > now {
    now_system_time + (instant - now)
} else {
    now_system_time - (now - instant)
}
```

thomaseizinger:
Learned something today, thanks for the comments!

PR #2408, opened 2021-12-29
*: Add Serialize and Deserialize to PeerId gossipsub MessageId and kad Key

laptou:
Continuation of #2405 on a new branch.


core/src/peer_id.rs
```diff
@@ -24,13 +24,17 @@ use rand::Rng;
...
+use serde::{Deserialize, Serialize};
+
 /// Public keys with byte-lengths smaller than `MAX_INLINE_KEY_LENGTH` will be
 /// automatically used as the peer id using an identity multihash.
 const MAX_INLINE_KEY_LENGTH: usize = 42;
 
 /// Identifier of a peer of the network.
 ///
 /// The data is a multihash of the public key of the peer.
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
```
thomaseizinger: You should be able to use the `crate` attribute of serde for the case where it has been renamed.

https://serde.rs/container-attrs.html#crate

thomaseizinger (commented):
One more suggestion on how we can use `serde` as the feature name. That would be a cleaner API IMO :)

core/src/peer_id.rs
laptou: It worked, it's done

core/src/peer_id.rs
```diff
@@ -24,13 +24,18 @@ use rand::Rng;
...
```
thomaseizinger: Sorry for only noticing this now but what does `derive(Serialize)` actually result in? I would expect that we serialize to the string representation if the serializer returns true for `is_human_readable` and otherwise returns the efficient byte representation of `Multihash`.

Looking at https://github.com/multiformats/rust-multihash/blob/ebd0d91be8f76ce84000d103243162373d8d08fa/src/multihash.rs#L84-L95, the derive seems to serialize to the internal representation of `Multihash`? This could be argued to be an upstream bug.

Can we add a test to confirm what this serializes to? (And fix the serialization to be _not_ the internal multihash if that is what it turns out to be?)

@mxinden thoughts?

thomaseizinger (changes requested):
Nice, thank you!

Sorry for the turn-around but I have another concern.

core/src/peer_id.rs
mxinden: Unfortunately I have little to no experience with `serde`. Things that come to my mind:

- We already have a text representation of `PeerId`s. (Actually we have two, see https://github.com/libp2p/rust-libp2p/issues/2259.) Would one implement a custom `Serialize` and `Deserialize` or depend on the internal Rust structures instead?
- In case we do serialize to the internal multihash structure, that would make `multihash` an externally facing dependency of `libp2p`. In other words, every breaking change in `multihash`s serialization code is a breaking change in `libp2p`. Is that something we want?

laptou: I added a custom `Serialize`/`Deserialize` impl and some unit tests.

laptou:
I also derived `Serialize`, `Deserialize` for `kad::store::record::Key`. Let me know if you want this in a separate PR, or if I can squeeze it into this one 😁

core/tests/serde.rs
```diff
@@ -0,0 +1,52 @@
...
+    assert_eq!(
+        &buf[..],
+        &[
+            0xc4, 38, // msgpack buffer header
+            0x00, 0x24, 0x08, 0x01, 0x12, 0x20, 0xe7, 0x37, 0x0c, 0x66, 0xef, 0xec, 0x80, 0x00,
+            0xd5, 0x87, 0xfc, 0x41, 0x65, 0x92, 0x8e, 0xe0, 0x75, 0x5f, 0x94, 0x86, 0xcb, 0x5c,
+            0xf0, 0xf7, 0x80, 0xd8, 0xe0, 0x6c, 0x98, 0xce, 0x7d, 0xa9
+        ]
+    );
+}
```
thomaseizinger, line 32: Have you seen the `serde_test` package? It allows for [easy assertion](https://docs.rs/serde_test/1.0.133/serde_test/trait.Configure.html) of human-readable / non-human-readable encoding.

thomaseizinger (approved):
Thanks for this effort!

One non-blocking comment so good to go in from my side. Any more considerations @mxinden ?

core/tests/serde.rs
laptou: It looks useful for writing future tests, but I don't really want to rewrite these existing tests. Thanks for bringing it to my attention :)

thomaseizinger: Yep no worries!

core/src/peer_id.rs
```diff
@@ -164,6 +167,60 @@ impl From<PeerId> for Vec<u8> {
...
+            serializer.serialize_bytes(&self.to_bytes()[..])
+        }
+    }
+}
+
+#[cfg(feature = "serde")]
+impl<'de> Deserialize<'de> for PeerId {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: _serde::Deserializer<'de>,
```
```suggestion
        D: serde::Deserializer<'de>,
```

Nit pick, would this work as well?

protocols/kad/src/record.rs
```diff
@@ -25,10 +25,14 @@ pub mod store;
...
 use instant::Instant;
 use libp2p_core::{multihash::Multihash, Multiaddr, PeerId};
+#[cfg(feature = "serde")]
+use serde::{Deserialize, Serialize};
 use std::borrow::Borrow;
 use std::hash::{Hash, Hasher};
 
 /// The (opaque) key of a record.
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "serde", serde(crate = "_serde"))]
```
mxinden, line 35: I am missing something here. Why is this line needed in `libp2p-kad` but not in `libp2p-gossipsub`?

laptou: `libp2p-kad`'s `serde` feature is declared explicitly in Cargo.toml, and it does other stuff besides just enabling the serde crate (it enables serde features of other dependencies), which means that I had to rename the serde crate in the dependencies of `libp2p-kad` to `_serde` (because a feature and an optional dependency cannot have the same name in stable Rust). On the other hand, it is simpler in `libp2p-gossipsub` because all that needs to happen is for the optional `serde` dependency to be enabled, so we don't need to rename `serde` to `_serde`.

mxinden: :pray: Thank you @laptou. That makes sense.

mxinden (approved):
Thanks @laptou!

PR #2409, opened 2021-12-29
protocols/gossipsub: Fix elided lifetime on gossipsub config builder

laptou:
- The Gossipsub config builder is currently hard to use b/c its error type is `&str`, and due to lifetime elision, `&self` needs to live as long as this `&str`. This PR fixes that.


mxinden (approved):
Looks good to me. @divagant-martian @AgeManning any comments?

divagant-martian:
lgtm!

PR #2414, opened 2021-12-30
protocols/kad: Implement `Error` and `Clone` for error types

laptou:
- implement `Error` for `kad::record::store::Error`, `kad::AddProviderError`, `kad::GetProvidersError`, `kad::GetClosestPeersError`, `kad::BootstrapError` using `thiserror`
- implement `Clone` for `kad::record::store::Error`, `kad::AddProviderError`, `kad::GetProvidersError`, `kad::GetClosestPeersError`, `kad::BootstrapError`

### DIFF
--- a/protocols/kad/src/record/store.rs
+++ b/protocols/kad/src/record/store.rs
@@ -30,7 +30,7 @@ use std::borrow::Cow;
 pub type Result<T> = std::result::Result<T, Error>;
 
 /// The possible errors of a `RecordStore` operation.
-#[derive(Debug)]
+#[derive(Debug, Clone)]
 pub enum Error {
     /// The store is at capacity w.r.t. the total number of stored records.
     MaxRecords,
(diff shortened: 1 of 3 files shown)


protocols/kad/src/record/store.rs
```diff
@@ -30,14 +31,15 @@ use std::borrow::Cow;
 pub type Result<T> = std::result::Result<T, Error>;
 
 /// The possible errors of a `RecordStore` operation.
-#[derive(Debug)]
+#[derive(Error, Debug, Clone)]
 pub enum Error {
-    /// The store is at capacity w.r.t. the total number of stored records.
+    #[error("the store cannot contain any more records")]
```
elenaf9, line 37: I would be in favor of also keeping the code docs. `#[error(..)]` only sets the `Display` message for this type, but isn't visible to a user that is looking at this type through the API docs on docs.rs.

laptou: done

thomaseizinger (approved):
LGTM. Needs conflicts resolved please :)

Neilblaze (approved):

PR #2424, opened 2022-01-06
.github/workflows/: Fix test-wasm and pin third-party action versions

galargh:
# Description

## `test-wasm`

I remove the `container` configuration from `test-wasm` job so that the job starts using `rust` installed here https://github.com/libp2p/rust-libp2p/blob/eefed59877609148562286a7a1bd2a38e974f443/.github/workflows/ci.yml#L61

Previously, we would run last step(https://github.com/libp2p/rust-libp2p/blob/eefed59877609148562286a7a1bd2a38e974f443/.github/workflows/ci.yml#L82) inside the container with `rust` from the container.

## third-party action versions

I pinned the third-party action versions to the exact commit SHAs. We follow the same strategy in protocol go repositories to ensure the code we use doesn't change.

# Testing

- [x] CI workflow in forked repo
- [x] CI workflow in this repo

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -44,10 +44,11 @@ jobs:
         include:
           - toolchain: wasm32-unknown-unknown
             args: "--features wasm-bindgen"
-    container:
-      image: rust
-      env:
-        CC: clang-11
+    env:
+      CC: clang-11
+    defaults:
+      run:
+        shell: bash
     steps:
 
     - name: Cancel Previous Runs
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -67,11 +67,11 @@ jobs:
 
     - name: Install a recent version of clang
       run: |
-        apt-get update
-        apt-get install -y clang-11
+        sudo apt-get update
+        sudo apt-get install -y clang-11
 
     - name: Install CMake
-      run: apt-get install -y cmake
+      run: sudo apt-get install -y cmake
 
     - uses: Swatinem/rust-cache@v1.3.0
       with:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -20,13 +20,13 @@ jobs:
     steps:
 
     - name: Cancel Previous Runs
-      uses: styfle/cancel-workflow-action@0.9.1
+      uses: styfle/cancel-workflow-action@a40b8845c0683271d9f53dfcb887a7e181d3918b # 0.9.1
```
mxinden, line 23: Would @dependabot still be able to detect newer versions and create pull requests like https://github.com/libp2p/rust-libp2p/pull/2300?

mxinden (commented):
Thanks @galargh for your help!

.github/workflows/ci.yml
galargh: https://github.com/dependabot/dependabot-core/issues/2835

It seems it won't have a problem with it. It won't update the version number in the comment though :( We could use https://github.com/timmeinerzhagen/dependabot-sha-comment-action for that if we wanted to.

mxinden: :rocket: I don't mind updating the version comment manually.

mxinden (approved):

.github/workflows/ci.yml
galargh: Alright :) Let me know if it becomes annoying. Hopefully, we'll be able to move the workflow to unified CI repo before that happens and then we'll be handling the action upgrades there.

PR #2427, opened 2022-01-10
protocols/: Update to lru v0.7.1

mxinden:
See https://rustsec.org/advisories/RUSTSEC-2021-0130


mxinden:
[Published](https://crates.io/crates/libp2p/0.41.1) and [tagged](https://github.com/libp2p/rust-libp2p/releases/tag/v0.41.1).

Note that this does not include the latest changes from `master`.

Issue #2425, opened 2022-01-09
inject_dial_failure called with wrong peer ID


rkuhn:
When dialling PeerA at a listen address of PeerB, `inject_dial_failure` is called with the peer ID of PeerB (and an error “Dial error: Invalid peer ID”). Proper error handling on the behaviour side would be easier if the peer ID of PeerA was used here (i.e. the expected ID, not the negotiated one).

mxinden:
Agreed that the above would be more intuitive.

Actually, the comment on `PendingOutboundConnectionError` is wrong as of today

https://github.com/libp2p/rust-libp2p/blob/71a54d8227a0043c26f83c61bceb68b174a343b3/core/src/connection/pool.rs#L202-L203

given that we return the actual peer ID instead of the expected one as you noted above

https://github.com/libp2p/rust-libp2p/blob/71a54d8227a0043c26f83c61bceb68b174a343b3/core/src/connection/pool.rs#L832-L837

How about we:

1. Call `inject_dial_failure` with the expected peer ID.
2. Include the actual peer ID in the `DialError`.

What do you think @rkuhn? Do you want to create a patch?

rkuhn:
sure, will do today or tomorrow — good idea of adding to DialError (as an `Option<PeerId>`, I think)

mxinden:
Closing here given that https://github.com/libp2p/rust-libp2p/pull/2428 is merged.

PR #2428, opened 2022-01-11
fix peer_id in case of DialError::InvalidPeerId (#2425)

rkuhn:
Previously, the negotiated PeerId was included in the swarm event and
inject_dial_failure’s arguments while the expected one was absent. This
patch adds the negotiated PeerId to the DialError and includes the expected
one in the notifications.

### DIFF
--- a/core/src/connection/error.rs
+++ b/core/src/connection/error.rs
@@ -18,9 +18,9 @@
 // FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
 // DEALINGS IN THE SOFTWARE.
 
-use crate::connection::ConnectionLimit;
 use crate::transport::TransportError;
 use crate::Multiaddr;
+use crate::{connection::ConnectionLimit, PeerId};
 use std::{fmt, io};
 
 /// Errors that can occur in the context of an established `Connection`.
@@ -84,8 +84,8 @@ pub enum PendingConnectionError<TTransErr> {
     Aborted,
 
     /// The peer identity obtained on the connection did not
-    /// match the one that was expected or is otherwise invalid.
-    InvalidPeerId,
+    /// match the one that was expected or is the local one.
+    InvalidPeerId(PeerId),
 
     /// An I/O error occurred on the connection.
     // TODO: Eventually this should also be a custom error?
@@ -110,8 +110,8 @@ where
             PendingConnectionError::ConnectionLimit(l) => {
                 write!(f, "Connection error: Connection limit: {}.", l)
             }
-            PendingConnectionError::InvalidPeerId => {
-                write!(f, "Pending connection: Invalid peer ID.")
+            PendingConnectionError::InvalidPeerId(id) => {
+                write!(f, "Pending connection: Unexpected peer ID {}.", id)
             }
         }
     }
@@ -125,7 +125,7 @@ where
         match self {
             PendingConnectionError::IO(err) => Some(err),
             PendingConnectionError::Transport(_) => None,
-            PendingConnectionError::InvalidPeerId => None,
+            PendingConnectionError::InvalidPeerId(_) => None,
             PendingConnectionError::Aborted => None,
             PendingConnectionError::ConnectionLimit(..) => None,
         }
--- a/core/src/connection/error.rs
+++ b/core/src/connection/error.rs
@@ -85,11 +85,30 @@ pub enum PendingConnectionError<TTransErr> {
 
     /// The peer identity obtained on the connection did not
     /// match the one that was expected or is the local one.
-    InvalidPeerId(PeerId),
+    WrongPeerId {
+        obtained: PeerId,
+        address: Multiaddr,
+    },
 
     /// An I/O error occurred on the connection.
     // TODO: Eventually this should also be a custom error?
     IO(io::Error),
+}
+
+impl<T> PendingConnectionError<T> {
+    pub fn map<U>(self, f: impl FnOnce(T) -> U) -> PendingConnectionError<U> {
+        match self {
+            PendingConnectionError::Transport(t) => PendingConnectionError::Transport(f(t)),
+            PendingConnectionError::ConnectionLimit(l) => {
+                PendingConnectionError::ConnectionLimit(l)
+            }
+            PendingConnectionError::Aborted => PendingConnectionError::Aborted,
+            PendingConnectionError::WrongPeerId { obtained, address } => {
+                PendingConnectionError::WrongPeerId { obtained, address }
+            }
+            PendingConnectionError::IO(e) => PendingConnectionError::IO(e),
+        }
+    }
 }
 
 impl<TTransErr> fmt::Display for PendingConnectionError<TTransErr>
@@ -110,8 +129,12 @@ where
             PendingConnectionError::ConnectionLimit(l) => {
                 write!(f, "Connection error: Connection limit: {}.", l)
             }
-            PendingConnectionError::InvalidPeerId(id) => {
-                write!(f, "Pending connection: Unexpected peer ID {}.", id)
+            PendingConnectionError::WrongPeerId { obtained, address } => {
+                write!(
+                    f,
+                    "Pending connection: Unexpected peer ID {} at {}.",
+                    obtained, address
+                )
             }
         }
     }
@@ -125,7 +148,7 @@ where
         match self {
             PendingConnectionError::IO(err) => Some(err),
             PendingConnectionError::Transport(_) => None,
-            PendingConnectionError::InvalidPeerId(_) => None,
+            PendingConnectionError::WrongPeerId { .. } => None,
             PendingConnectionError::Aborted => None,
             PendingConnectionError::ConnectionLimit(..) => None,
         }
--- a/core/src/connection/pool.rs
+++ b/core/src/connection/pool.rs
@@ -771,7 +771,7 @@ where
 
                     enum Error {
                         ConnectionLimit(ConnectionLimit),
-                        InvalidPeerId,
+                        InvalidPeerId(PeerId),
                     }
 
                     impl<TransportError> From<Error> for PendingConnectionError<TransportError> {
@@ -780,7 +780,7 @@ where
                                 Error::ConnectionLimit(limit) => {
                                     PendingConnectionError::ConnectionLimit(limit)
                                 }
-                                Error::InvalidPeerId => PendingConnectionError::InvalidPeerId,
+                                Error::InvalidPeerId(x) => PendingConnectionError::InvalidPeerId(x),
                             }
                         }
                     }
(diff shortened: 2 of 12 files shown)


core/tests/network_dial_error.rs
```diff
@@ -83,6 +84,70 @@ fn deny_incoming_connec() {
...
 }
 
+#[test]
+fn invalid_peer_id() {
+    // Checks whether dialling with the wrong peer id raises the correct error
+
+    let mut swarm1 = test_network(NetworkConfig::default());
+    let mut swarm2 = test_network(NetworkConfig::default());
+
+    swarm1.listen_on("/memory/0".parse().unwrap()).unwrap();
```
rkuhn, line 99: does this need a random number instead?

core/src/connection/error.rs
```diff
@@ -84,8 +84,8 @@ pub enum PendingConnectionError<TTransErr> {
     Aborted,
 
     /// The peer identity obtained on the connection did not
-    /// match the one that was expected or is otherwise invalid.
-    InvalidPeerId,
+    /// match the one that was expected or is the local one.
+    InvalidPeerId(PeerId),
```
mxinden:
```suggestion
    InvalidPeerId {
      obtained: PeerId,
    }
```

What do you think? That would make it clear that the provided one in `InvalidPeerId` is the one obtained on the connection.

core/tests/network_dial_error.rs
```diff
@@ -83,6 +84,70 @@ fn deny_incoming_connec() {
     .unwrap();
 }
 
+#[test]
+fn invalid_peer_id() {
+    // Checks whether dialling with the wrong peer id raises the correct error
```
mxinden:
```suggestion
    // Checks whether dialing an address containing the wrong peer id raises an error
    // for the expected peer id instead of the obtained peer id.
```

core/tests/network_dial_error.rs
mxinden: On `0` the port is randomized, i.e. no need for a random number here.

https://github.com/libp2p/rust-libp2p/blob/5617481e56e3949dabc7ead26f15a75c0f50136f/core/src/transport/memory.rs#L61-L73

core/tests/network_dial_error.rs
```diff
@@ -83,6 +84,70 @@ fn deny_incoming_connec() {
...
+
+    let mut swarm1 = test_network(NetworkConfig::default());
+    let mut swarm2 = test_network(NetworkConfig::default());
+
+    swarm1.listen_on("/memory/0".parse().unwrap()).unwrap();
+
+    let address = async_std::task::block_on(future::poll_fn(|cx| match swarm1.poll(cx) {
```
mxinden:
```suggestion
    let address = futures::executor::block_on(future::poll_fn(|cx| match swarm1.poll(cx) {
```

Just a nit pick. We will always depend on `futures` but might now always depend on `async_std`.

core/tests/network_dial_error.rs
```diff
@@ -83,6 +84,70 @@ fn deny_incoming_connec() {
...
+        Poll::Ready(NetworkEvent::NewListenerAddress { listen_addr, .. }) => {
+            Poll::Ready(listen_addr)
+        }
+        Poll::Pending => Poll::Pending,
+        _ => panic!("Was expecting the listen address to be reported"),
+    }));
+
+    let other_keys = identity::Keypair::generate_ed25519();
+    let other_id = other_keys.public().to_peer_id();
```
mxinden:
```suggestion
    let other_id = PeerId::random();
```

core/tests/network_dial_error.rs
```diff
@@ -83,6 +84,70 @@ fn deny_incoming_connec() {
...
+    let other_addr = address.with(Protocol::P2p(other_id.into()));
+
+    swarm2.dial(&other_addr, TestHandler()).unwrap();
+
+    let (peer_id, error) = async_std::task::block_on(future::poll_fn(
```
mxinden:
```suggestion
    let (peer_id, error) = futures::executor::block_on(future::poll_fn(
```

core/tests/network_dial_error.rs
mxinden: Side comment: It is unfortunate that `Network` does not implement `Future`. Otherwise this could be simplified a lot.

core/tests/network_dial_error.rs
```diff
@@ -83,6 +84,70 @@ fn deny_incoming_connec() {
...
+                        }
+                        PendingConnectionError::ConnectionLimit(l) => {
+                            PendingConnectionError::ConnectionLimit(l)
+                        }
+                        PendingConnectionError::Aborted => PendingConnectionError::Aborted,
+                        PendingConnectionError::InvalidPeerId(id) => {
+                            PendingConnectionError::InvalidPeerId(id)
+                        }
+                        PendingConnectionError::IO(e) => PendingConnectionError::IO(e),
+                    };
```
mxinden: Why is this conversion needed? Is there a type conflict with the type parameter of `PendingConnectionError::Transport`?

core/tests/network_dial_error.rs
```diff
@@ -83,6 +84,70 @@ fn deny_incoming_connec() {
...
+                Poll::Ready(x) => panic!("unexpected {:?}", x),
+                Poll::Pending => Poll::Pending,
+            }
+        },
+    ));
+    assert_eq!(peer_id, other_id);
+    match error {
+        PendingConnectionError::InvalidPeerId(id) => assert_eq!(id, *swarm1.local_peer_id()),
+        x => panic!("wrong error {:?}", x),
+    }
```
mxinden: Why not do both of these within the `match` above?

swarm/src/lib.rs
```diff
@@ -1335,8 +1335,8 @@ pub enum DialError {
     /// Pending connection attempt has been aborted.
     Aborted,
     /// The peer identity obtained on the connection did not
-    /// match the one that was expected or is otherwise invalid.
-    InvalidPeerId,
+    /// match the one that was expected or is the local one.
+    InvalidPeerId(PeerId),
```
```suggestion
    InvalidPeerId {
      obtained: PeerId,
    },
```

mxinden (commented):
Thank you for taking care of this @rkuhn!

core/src/connection/error.rs
rkuhn: Yes, and I’ll want to add the failed address if possible (looking into it now).

core/tests/network_dial_error.rs
rkuhn: okay, will fix all occurrences in that file

core/tests/network_dial_error.rs
rkuhn: great, thanks for the pointer!

core/tests/network_dial_error.rs
rkuhn: No, it was a left-over from some earlier version.

core/tests/network_dial_error.rs
rkuhn: I prefer to assert on the test function level when possible, for ease of stack trace reading (and reliability of reporting).

rkuhn:
@mxinden Thanks for the review! While trying this out and working a bit more on Actyx & ipfs-embed I found that InvalidPeerId should contain the address at which the mismatch occurred — since the user will very likely want to scratch that Multiaddr from any lists they may be maintaining (in conjunction with the expected PeerId).

The WASM failures seem unrelated to my changes (at least I hope so).

core/src/connection/pool.rs
```diff
@@ -763,49 +763,39 @@ where
...
                                 }
                             }
                             Ok(())
                         })
                         // Check peer is not local peer.
                         .and_then(|()| {
                             if self.local_id == peer_id {
-                                Err(Error::InvalidPeerId)
+                                Err(PendingConnectionError::InvalidPeerId {
+                                    obtained: peer_id,
```
mxinden: How about renaming `peer_id` to `obtained_peer_id`? That would be consistent with `expected_peer_id` and would be less confusing.

mxinden (commented):
One last comment. Otherwise looks good to me.

core/src/connection/pool.rs
rkuhn: yes, will do that once I get a minute — crunch time right now

rkuhn:
Seeing that you added an actual case of “invalid peer id” in #2404, I think this PR needs to cover a bit more ground: we want to signal two distinct cases now. How about the following:

- `InvalidPeerId(multihash)`: the `Protocol::P2p(multihash)` didn’t correspond to a PeerId (why does P2p not hold a PeerId?)
- `WrongPeerId { obtained, address }`: the ID obtained from the peer didn’t match our expectations

mxinden:
> Seeing that you added an actual case of “invalid peer id” in #2404, I think this PR needs to cover a bit more ground.

:+1: thanks for being rigorous here.

>   * `InvalidPeerId(multihash)`: the `Protocol::P2p(multihash)` didn’t correspond to a PeerId 

:+1: 

> (why does P2p not hold a PeerId?)

I haven't been involved back then. I guess due to `PeerId` being defined inside the `libp2p-core` crate and `multiaddr` crate not wanting to depend on `libp2p-core`.

That said, I agree that this is not ideal. I would very much like this to be enforced at the type system level.

>   * `WrongPeerId { obtained, address }`: the ID obtained from the peer didn’t match our expectations

:+1:

swarm/src/lib.rs
```diff
@@ -1326,10 +1327,13 @@ pub enum DialError {
     DialPeerConditionFalse(dial_opts::PeerCondition),
     /// Pending connection attempt has been aborted.
     Aborted,
-    /// The provided peer identity is invalid or the peer identity obtained on
-    /// the connection did not match the one that was expected or is otherwise
-    /// invalid.
-    InvalidPeerId,
+    /// The provided peer identity is invalid
```
mxinden, line 1330:
```suggestion
    /// The provided peer identity is invalid.
```

core/src/connection/error.rs
```diff
@@ -84,14 +84,33 @@ pub enum PendingConnectionError<TTransErr> {
     Aborted,
 
     /// The peer identity obtained on the connection did not
-    /// match the one that was expected or is otherwise invalid.
-    InvalidPeerId,
+    /// match the one that was expected or is the local one.
+    WrongPeerId {
+        obtained: PeerId,
+        address: Multiaddr,
```
mxinden, line 90:
```suggestion
        address: ConnectedPoint,
```
How about a `ConnectedPoint` instead of a plain `Multiaddr` here? Mostly for the sake of additional context.

mxinden (commented):
Thanks for splitting the two error cases @rkuhn.

mxinden:
Seems like I am not allowed to push to this branch. I will merge this via https://github.com/libp2p/rust-libp2p/pull/2441.

PR #2430, opened 2022-01-12
.github/workflows: Add cargo audit

mxinden:
Run cargo audit on each push and on a daily schedule.


mxinden:
Blocked on https://github.com/libp2p/rust-libp2p/issues/2432.

PR #2433, opened 2022-01-13
protocols/: Fix clippy and fmt warnings

elenaf9:
Github actions are failing in #2262 due to new rustfmt and clippy versions. This PR fixes clippy and rustfmt. 
Note: in 4fbaa2d9673eb8970d1d95f57261b03a4a5d6dd9 I had to remove the tuple from the function signature and instead add it in the function body, because it caused some internal error on rustfmt.


mxinden (approved):
Again ahead of me. :)

Thanks for the fix!

PR #2437, opened 2022-01-14
misc/metrics: Set default-features = false on libp2p-core

folex:
Right now, `libp2p-metrics` uses all default features from `libp2p-core`. As a result, it is very hard (may be impossible) to disable e.g. `ecdsa` in `libp2p-core` in a project that depends on `libp2p`.

After this change, the problem is completely fixed.

### DIFF
--- a/misc/metrics/Cargo.toml
+++ b/misc/metrics/Cargo.toml
@@ -17,7 +17,7 @@ kad = ["libp2p-kad"]
 ping = ["libp2p-ping"]
 
 [dependencies]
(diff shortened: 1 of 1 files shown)


mxinden (approved):
My bad. Thanks for looking into it!

mxinden:
@folex seems like I am not allowed to push recent `master` to this branch. Can you either grant me access or merge latest `master` yourself?

folex:
@mxinden merged

Issue #2052, opened 2021-04-20
NAT traversal


mxinden:
This GitHub issue tracks the status of NAT traversal capabilities for rust-libp2p. For an overview across libp2p implementations see libp2p/specs#312.

Please comment below to suggest additions or corrections.

Contributions of any form are welcome. Before you start working on something, please comment below to make sure no work is duplicated.

---

### Transport protocols
- [x] TCP + port-reuse
- [x] QUIC
	- https://github.com/libp2p/rust-libp2p/pull/2289
    - libp2p/rust-libp2p#2801
- [x] Websockets (for signaling with browsers)
- [x] WebRTC
    - libp2p/rust-libp2p#1066
    - https://github.com/libp2p/rust-libp2p/pull/2622

### TURN-like protocols

- [x] circuit relay v1
- [x] circuit relay v2
    - libp2p/specs#314
    - libp2p/rust-libp2p#2059

### Signaling protocols

- [x] Direct connection upgrade through relay
    - libp2p/specs#173
    - libp2p/rust-libp2p#2076
    - Blocked on:
        - https://github.com/libp2p/rust-libp2p/issues/1896
        - https://github.com/libp2p/rust-libp2p/issues/2249
        - https://github.com/libp2p/rust-libp2p/issues/2250

### STUN-like protocols

- [ ] STUN
- [x] AutoNAT
    - libp2p/specs#180
    - https://github.com/libp2p/rust-libp2p/pull/2262/

### Other protocols

- [x] multistream-select v1 simultaneous open
     - libp2p/specs#196
     - libp2p/rust-libp2p#2066
- [x] UPnP
     - https://github.com/libp2p/rust-libp2p/issues/3903

### Misc

- [x] Parallel connection attempts to increase hole punch success rate
    - https://github.com/libp2p/rust-libp2p/issues/1896
    - https://github.com/libp2p/rust-libp2p/pull/2248
- [x] Enable dialing a specific fixed set of addresses for a single peer
    - https://github.com/libp2p/rust-libp2p/issues/2249
- [x] Allow specifying the role of a connection in NetworkBehaviourAction::Dial
    - https://github.com/libp2p/rust-libp2p/issues/2250
    - https://github.com/libp2p/rust-libp2p/pull/2363

mxinden:
**Changelog**
- Referenced https://github.com/libp2p/rust-libp2p/pull/2059 as the rust-libp2p circuit relay v2 implementation.

[...]

PR #2438, opened 2022-01-15
protocols/: Implement Direct Connection Upgrade through Relay (DCUtR)

mxinden:
Enables two peers to coordinate a hole punch (direct connection upgrade)
via a relayed connection.

See https://github.com/libp2p/specs/blob/master/relay/DCUtR.md for
specification.

Depends on either:
- https://github.com/libp2p/rust-libp2p/pull/2363
- https://github.com/libp2p/rust-libp2p/pull/2066

Replaces https://github.com/libp2p/rust-libp2p/pull/2076.


protocols/dcutr/src/handler.rs
```diff
@@ -0,0 +1,81 @@
...
+                }
+            }
+            Self::DirectConnection {
+                relayed_connection_id,
+                ..
+            } => {
+                assert!(
+                    !endpoint.is_relayed(),
+                    "`Prototype::DirectConnection` is never created for relayed connection."
+                );
```
mxinden, line 65: TODO: Ensure remote can not trick us into dialing a relayed address as a hole punch.

mxinden: Tackled via https://github.com/libp2p/rust-libp2p/pull/2438/commits/7bc75a5216a32bf211d095a81a04cfba463df1ea.

mxinden:
This is the last step needed for basic hole punching in rust-libp2p (see also https://github.com/libp2p/rust-libp2p/issues/2052). 

Anyone interested in reviewing? @thomaseizinger maybe? :innocent:

protocols/dcutr/tests/lib.rs
```diff
@@ -0,0 +1,285 @@
...
+use futures::stream::StreamExt;
+use futures::task::Spawn;
+use libp2p::core::multiaddr::{Multiaddr, Protocol};
+use libp2p::core::muxing::StreamMuxerBox;
+use libp2p::core::transport::upgrade::Version;
+use libp2p::core::transport::{Boxed, MemoryTransport, OrTransport, Transport};
+use libp2p::core::PublicKey;
+use libp2p::core::{identity, PeerId};
+use libp2p::dcutr;
+use libp2p::ping::{Ping, PingConfig, PingEvent};
```
MarcoPolo: Do you need to use the Ping behavior for this test?

protocols/dcutr/src/behaviour.rs
```diff
@@ -0,0 +1,401 @@
...
+        handler: Self::ProtocolsHandler,
+        _error: &DialError,
+    ) {
+        match handler {
+            handler::Prototype::DirectConnection {
+                relayed_connection_id,
+                role: handler::Role::Initiator { attempt },
+            } => {
+                let peer_id =
+                    peer_id.expect("Prototype::DirectConnection is always for known peer.");
```
MarcoPolo: typo? or rephrase? maybe "Peer is always known for a `Prototype::DirectConnection`"

protocols/dcutr/src/behaviour.rs
```diff
@@ -0,0 +1,401 @@
...
+            Action::AcceptInboundConnect {
+                inbound_connect,
+                handler,
+                peer_id,
+            } => NetworkBehaviourAction::NotifyHandler {
+                handler,
+                peer_id,
+                event: Either::Left(handler::relayed::Command::AcceptInboundConnect {
+                    inbound_connect,
+                    obs_addrs: poll_parameters
```
MarcoPolo: nit (optional), could we move this duplicate logic to a fn in the parent scope?

protocols/dcutr/examples/client.rs
```diff
@@ -0,0 +1,275 @@
...
+
+    // Wait to listen on localhost.
+    block_on(async {
+        let mut delay = futures_timer::Delay::new(std::time::Duration::from_secs(1)).fuse();
+        loop {
+            futures::select! {
+                            event = swarm.next() => {
+                                match event.unwrap() {
+                                    SwarmEvent::NewListenAddr { address, .. } => {
+            info!("Listening on {:?}", address);
```
MarcoPolo: nit, fix indent here.

protocols/dcutr/src/handler/relayed.rs
```diff
@@ -0,0 +1,380 @@
...
+        attempt: u8,
+    },
+    AcceptInboundConnect {
+        obs_addrs: Vec<Multiaddr>,
+        inbound_connect: protocol::inbound::PendingConnect,
+    },
+    /// Upgrading the relayed connection to a direct connection either failed for good or succeeded.
+    /// There is no need to keep the relayed connection alive for the sake of upgrading to a direct
+    /// connection.
+    UpgradeFinishedDontKeepAlive,
```
MarcoPolo, line 53: This makes sense. But it makes me wonder what happens if the relayed connection stays up, but the direct connection fails? Do we have a mechanism that will restart dcutr?

protocols/dcutr/src/handler/relayed.rs
```diff
@@ -0,0 +1,380 @@
...
+    }
+
+    fn inject_fully_negotiated_inbound(
+        &mut self,
+        output: <Self::InboundProtocol as upgrade::InboundUpgrade<NegotiatedSubstream>>::Output,
+        _: Self::InboundOpenInfo,
+    ) {
+        match output {
+            EitherOutput::First(inbound_connect) => {
+                let remote_addr = match & self.endpoint {
```
MarcoPolo:
```suggestion
                let remote_addr = match &self.endpoint {
```

protocols/dcutr/src/handler/relayed.rs
```diff
@@ -0,0 +1,380 @@
...
+    type OutboundProtocol = protocol::outbound::Upgrade;
+    type OutboundOpenInfo = u8; // Number of upgrade attempts.
+    type InboundOpenInfo = ();
+
+    fn listen_protocol(&self) -> SubstreamProtocol<Self::InboundProtocol, Self::InboundOpenInfo> {
+        match self.endpoint {
+            ConnectedPoint::Dialer { .. } => {
+                SubstreamProtocol::new(upgrade::EitherUpgrade::A(protocol::inbound::Upgrade {}), ())
+            }
+            ConnectedPoint::Listener { .. } => {
```
MarcoPolo, line 182: I think this needs a bit of documentation explaining why this is backwards. Took me a bit to figure out but its because the listener on the relayed connection is the one that starts DCUtR dance, not the dialer.

Or maybe this is obvious. Not sure.

protocols/dcutr/src/handler/relayed.rs
```diff
@@ -0,0 +1,380 @@
...
+            Self::OutboundProtocol,
+            Self::OutboundOpenInfo,
+            Self::OutEvent,
+            Self::Error,
+        >,
+    > {
+        // Check for a pending (fatal) error.
+        if let Some(err) = self.pending_error.take() {
+            // The handler will not be polled again by the `Swarm`.
+            return Poll::Ready(ProtocolsHandlerEvent::Close(err));
```
MarcoPolo, line 359: This closes the actual socket connection, not a substream, correct?

The code isn't wrong, I just want to make sure I understand.

protocols/dcutr/src/handler/relayed.rs
```diff
@@ -0,0 +1,380 @@
...
+                self.inbound_connects
+                    .push(inbound_connect.accept(obs_addrs).boxed());
+            }
+            Command::UpgradeFinishedDontKeepAlive => {
+                self.keep_alive = KeepAlive::No;
+            }
+        }
+    }
+
+    fn inject_listen_upgrade_error(
```
MarcoPolo, line 259: I know this isn't related at all, but shouldn't these be renamed to `inject_error_inbound` to match the above `inject_fully_negotiated_inbound`, or is there some subtlety that I'm missing?

protocols/dcutr/src/handler/relayed.rs
```diff
@@ -0,0 +1,380 @@
...
+        &mut self,
+        _: Self::InboundOpenInfo,
+        error: ProtocolsHandlerUpgrErr<<Self::InboundProtocol as InboundUpgradeSend>::Error>,
+    ) {
+        self.keep_alive = KeepAlive::No;
```
MarcoPolo: Are we sure we want to kill this handler in this timeout case? What if it was a one-off issue? 

As is does the user have to track timeouts and restart the connection?

protocols/dcutr/src/handler/relayed.rs
```diff
@@ -0,0 +1,380 @@
...
+            }
+        }
+    }
+
+    fn inject_dial_upgrade_error(
+        &mut self,
+        _open_info: Self::OutboundOpenInfo,
+        error: ProtocolsHandlerUpgrErr<<Self::OutboundProtocol as OutboundUpgradeSend>::Error>,
+    ) {
+        self.keep_alive = KeepAlive::No;
```
MarcoPolo, line 310: Should we be checking attempts here before setting keep alive = no?

MarcoPolo (commented):
First pass. Looking good! I can't believe this is so simple.

Overall I guess I'm wondering how much of the failure cases have to be handled by the end user vs the library. Things such as:
1. The direct connection fails but the relay connection still works (can happen if the nat state gets reset for example).
2. timeouts

and some questions in general from me.

Thanks!

thomaseizinger:
> This is the last step needed for basic hole punching in rust-libp2p (see also #2052).
> 
> Anyone interested in reviewing? @thomaseizinger maybe? innocent

Sorry, GitHub for some reason filtered this out of my "review requested" list :(

protocols/dcutr/examples/client.rs
```diff
@@ -0,0 +1,275 @@
...
+use std::error::Error;
+use std::net::Ipv4Addr;
+use std::str::FromStr;
+use std::task::{Context, Poll};
+use structopt::StructOpt;
+
+#[derive(Debug, StructOpt)]
+#[structopt(name = "libp2p DCUtR client")]
+struct Opts {
+    /// The mode (relay, client-listen, client-dial)
```
thomaseizinger: Seems stale:

```suggestion
    /// The mode (client-listen, client-dial)
```

protocols/dcutr/examples/client.rs
```diff
@@ -0,0 +1,275 @@
...
+        .unwrap();
+
+    // Wait to listen on localhost.
+    block_on(async {
+        let mut delay = futures_timer::Delay::new(std::time::Duration::from_secs(1)).fuse();
+        loop {
+            futures::select! {
+                            event = swarm.next() => {
+                                match event.unwrap() {
+                                    SwarmEvent::NewListenAddr { address, .. } => {
```
thomaseizinger: Why not return from this if that is what we are waiting for (based on the comment?).

protocols/dcutr/examples/client.rs
```diff
@@ -0,0 +1,275 @@
...
+                })) => {
+                    info!("Observed address: {:?}", observed_addr);
+                    break;
+                }
+                event => panic!("{:?}", event),
+            }
+        }
+    });
+
+    if matches!(opts.mode, Mode::Dial) {
```
thomaseizinger: Nit: Could be a `==` if you would derive `PartialEq`.

protocols/dcutr/examples/client.rs
```diff
@@ -0,0 +1,275 @@
...
+            .dial(
+                opts.relay_address
+                    .clone()
+                    .with(Protocol::P2pCircuit)
+                    .with(Protocol::P2p(opts.remote_peer_id.unwrap().into())),
+            )
+            .unwrap();
+    }
+
+    block_on(futures::future::poll_fn(move |cx: &mut Context<'_>| {
```
thomaseizinger: Why `poll_fn` here and and `async` loop above?

protocols/dcutr/src/behaviour.rs
```diff
@@ -0,0 +1,401 @@
...
+    PollParameters, ProtocolsHandler, ProtocolsHandlerUpgrErr,
+};
+use std::collections::{HashMap, HashSet, VecDeque};
+use std::task::{Context, Poll};
+use thiserror::Error;
+
+/// The events produced by the [`Behaviour`].
+#[derive(Debug)]
+pub enum Event {
+    InitiateDirectConnectionUpgrade {
```
thomaseizinger: The naming of this variant doesn't indicate an event but a command. Is that on purpose?

protocols/dcutr/src/protocol/outbound.rs
```diff
@@ -0,0 +1,164 @@
...
+            let msg = HolePunch {
+                r#type: hole_punch::Type::Sync.into(),
+                obs_addrs: vec![],
+            };
+
+            let mut encoded_msg = BytesMut::new();
+            msg.encode(&mut encoded_msg)
+                .expect("BytesMut to have sufficient capacity.");
+
+            substream.send(encoded_msg.freeze()).await?;
```
thomaseizinger: Might be worthwhile to extract some helpers to hide the details of constructing one of these messages from the overall flow.

protocols/dcutr/src/behaviour.rs
thomaseizinger: Unless I misunderstand the logic, this could be named Initiate**d**DirectConnectionUpgrade?

thomaseizinger: `DirectConnectedUpgradeInitiated` might be even better but then it is not in line with `RemoteInitiatedDirectConnectionUpgrade`.

protocols/dcutr/src/behaviour.rs
```diff
@@ -0,0 +1,401 @@
...
+        }
+
+        Poll::Pending
+    }
+}
+
+/// A [`NetworkBehaviourAction`], either complete, or still requiring data from [`PollParameters`]
+/// before being returned in [`Behaviour::poll`].
+#[allow(clippy::large_enum_variant)]
+enum Action {
```
thomaseizinger:
```suggestion
enum ActionBuilder {
```

protocols/dcutr/src/handler/relayed.rs
thomaseizinger: Yep, this will close the socket AFAIK.

protocols/dcutr/src/behaviour.rs
```diff
@@ -0,0 +1,401 @@
...
+
+    fn new_handler(&mut self) -> Self::ProtocolsHandler {
+        handler::Prototype::UnknownConnection
+    }
+
+    fn addresses_of_peer(&mut self, _peer_id: &PeerId) -> Vec<Multiaddr> {
+        vec![]
+    }
+
+    fn inject_connected(&mut self, _peer_id: &PeerId) {}
```
thomaseizinger: Nit: I think these have a default now.

protocols/dcutr/src/behaviour.rs
```diff
@@ -0,0 +1,401 @@
...
+                    ConnectedPoint::Listener { local_addr, .. } => local_addr,
+                    ConnectedPoint::Dialer { .. } => unreachable!("Due to outer if."),
+                };
+                self.queued_actions.push_back(
+                    NetworkBehaviourAction::GenerateEvent(Event::InitiateDirectConnectionUpgrade {
+                        remote_peer_id: *peer_id,
+                        local_relayed_addr: local_addr.clone(),
+                    })
+                    .into(),
+                );
```
thomaseizinger: This could be written as a single `.extend` of an array to make it clear that the actions belong together.

thomaseizinger (approved):
Some comments, nothing blocking :)

protocols/dcutr/tests/lib.rs
mxinden: Indeed I don't. Thanks for the catch. Removed with 4a5461c0ba29410dce05fe957290abf3b3dcb0d2.

protocols/dcutr/src/behaviour.rs
mxinden: Thanks. Fixed with e7fa5e9c5ee2354a27218ef830ec9e7e92cd9a15.

protocols/dcutr/src/behaviour.rs
mxinden: You are right. Removed with 831deb5625a58d3b2b65e7873d6bd0d58561aa4c.

protocols/dcutr/src/behaviour.rs
mxinden: Good idea. Renamed to `InitiatedDirectConnectionUpgrade ` for now with 8384b4e48b27e807a2b066cf21f9aea26b8ac1f7.

I am fine with `DirectConnectedUpgradeInitiated` as well.

protocols/dcutr/examples/client.rs
mxinden: Good point. Replaced with `async` in 128d8015d6e6dc68c824292d661f1ee62029fb49.

protocols/dcutr/examples/client.rs
mxinden: :+1: fixed with 10046f21c644118a46dac74c25a3b1c927d532ae. I think I previously carried something within the `Dial` variant, though no longer.

protocols/dcutr/examples/client.rs
mxinden: Yes, this is misleading. We are giving `Swarm` some time to set up listeners. Though it might be more than one, depending on the number of network interfaces, thus returning after a single `NewListenAddr` might wait for one network interface, but not all.

Waiting for a second instead is a hack, but a hack that works. Happy for alternative suggestions.

At least better documented with 21ae03bbe9a30bc4ce567511f81343d0af54043c.

protocols/dcutr/examples/client.rs
mxinden: Oh surprising. Even though it is a macro, I would expect Rust fmt to at least do the indentation outside of the macro correctly. Either way. Thanks for the pointer. Fixed with 21ae03bbe9a30bc4ce567511f81343d0af54043c.

protocols/dcutr/examples/client.rs
mxinden: Oh, thanks for the catch! Fixed with da5e48f10d7eee80a97d7d2f8c349b30696eec11.

mxinden (commented):
Sorry for the delay here, still not yet done going through all the comments. Very much appreciate that you two took the time for a review @MarcoPolo and @thomaseizinger.

protocols/dcutr/src/behaviour.rs
mxinden: :+1: Moved to a closure in 3b75add8.

protocols/dcutr/src/handler/relayed.rs
mxinden: > Do we have a mechanism that will restart dcutr?

DCUtR has an internal mechanism that retries twice, resulting in 3 upgrade attempts. If all 3 fail it does not retry again.

> But it makes me wonder what happens if the relayed connection stays up, but the direct connection fails?

DCUtR itself will not keep the relayed connection alive. Though other protocols might still use it up until the time or bandwidth limit of the relayed connection is hit and thus the connection breaks.

I have been thinking about whether upper-level protocols need this to be more explicit, in other words whether they need to know that they are sending data over a relayed connection which might break at any point in time. Upper-level protocols will need to be able to cope with connection failures anyways. In the current state of DCUtR we expect upper-level protocols to handle upgrade failures and relay connection terminations with that same mechanism.

I am happy for alternative suggestions. Would be cool to be able to provide more information to upper-level applications. Though unless done in a generic way, I am not sure it is worth the additional complexity.

@MarcoPolo what do you think?

protocols/dcutr/src/handler/relayed.rs
mxinden: Not obvious. Definitely worth a comment. Added in 9d998502.

protocols/dcutr/src/handler/relayed.rs
mxinden: :+1: reference to the `enum`:

https://github.com/libp2p/rust-libp2p/blob/ea180aa3a88641f74f9fa7f89d6efdf48509892d/swarm/src/protocols_handler.rs#L311-L321

Would a comment have helped here? Grateful for suggestions.

protocols/dcutr/src/handler/relayed.rs
mxinden: Good catch. Changed with 0194655f661f02c5afcef6234cbdd2af28c7045c.

protocols/dcutr/src/handler/relayed.rs
mxinden: Today the retry mechanism in DCUtR retries on hole-punch failure only. In other words when the direct connection establishment itself fails.

`inject_dial_upgrade_error` called on `relayed::Handler` implies that the communication via the relayed connection failed, not the hole-punch.

I would argue that there is not much use in retrying a _direct connection upgrade_ over a erroneous relayed connection. Would you agree with that reasoning @MarcoPolo?

protocols/dcutr/src/protocol/outbound.rs
mxinden: Good idea. We can even go one step further and implement `asynchronous_codec::Encoder` and `asynchronous_codec::Decoder` for a custom `Codec` that takes care of all the `prost::Message` encoding, wrapping a `UviBytes`.

What do you think of 354900fd1eafa108d9f70f54ec5c15ae3527656b @thomaseizinger?

We could take this one step further, making `Codec` abstract over `prost::Message` and thus use the same `Codec` in all `protocols/XXX`. What do you think?

protocols/dcutr/src/handler/relayed.rs
MarcoPolo: is the connection terminally failed? As in it may not recover in the future? If so, then I think it's okay to have a KeepAlive::No;. Otherwise Wouldn't it make sense to only set it in the terminal error case? 

For example if this is just a timeout, maybe we want to try again.

protocols/dcutr/src/handler/relayed.rs
MarcoPolo: yep I saw that. I just wasn't sure if it meant (in practice) the socket connection or the substream. I know this is tricky since it may not even be a socket connection. Not sure what else we could add, but maybe something like:

```
/// Close the connection for the given reason. Note this will affect all handlers that use this connection.
```

protocols/dcutr/src/behaviour.rs
mxinden: Sounds good to me. Done across the whole file with b50708b7fe237e59ae549499a6b992a664386b60.

protocols/dcutr/src/handler/relayed.rs
mxinden: Just to avoid any confusion. setting `KeepAlive::No` signals that this handler has no more use for the connection. In case another handler returns `KeepAlive::Yes` the connection is still kept alive.

> is the connection terminally failed?

No. One could still reuse the connection.

> For example if this is just a timeout, maybe we want to try again.

We might want to do that in the future, but today there is no mechanism that will trigger a retry. In other words, DCUtR will never use the connection again. Given that it never uses the connection again, it should set `KeepAlive::No`.

Does the above make sense @MarcoPolo? Am I missing something?

protocols/dcutr/src/handler/relayed.rs
MarcoPolo: > DCUtR itself will not keep the relayed connection alive. Though other protocols might still use it up until the time or bandwidth limit of the relayed connection is hit and thus the connection breaks.

I think this makes sense. Here's a use case I'm thinking of as a user of this behavior:

1. I use this behavior as part of my root behavior.
2. I successfully upgrade my connection to a direct one.
3. Some time passes.
4. The direct connection fails (this is fine and expected).
5. How do I try to get another direct connection?

Does the other peer have to close all existing relayed connections with me and start another one so my `inject_connection_established` callback can happen?
Or maybe (in the future, not this PR) there should be method in the behavior that allows one to try to re-setup a dcutr dance.

protocols/dcutr/src/handler/relayed.rs
mxinden: What do you think of https://github.com/libp2p/rust-libp2p/pull/2470 @MarcoPolo?

protocols/dcutr/src/behaviour.rs
mxinden: :+1: Done with f18a8638e310c1ba87ddd3ce1e08baf2046fb59a.

protocols/dcutr/src/handler/relayed.rs
MarcoPolo: oh!! I realized my confusion this is in the relayed handler that is trying to orchestrate the dance before we start the simultaneous tcp connections. For some reason I thought this was part of the rety logic during the simultaneous tcp connections.

Thanks, and sorry for the confusion.

mxinden: No worries. Actually I had the same thought as well when I first read your initial comment.

protocols/dcutr/src/handler/relayed.rs
mxinden: > How do I try to get another direct connection?

For now you would need to initiate a new relayed connection, which would lead to the destination of the relayed connection to initiate a _direct connection upgrade_.

>  Or maybe (in the future, not this PR) there should be method in the behavior that allows one to try to re-setup a dcutr dance.

Over an existing relayed connection, potentially the relayed connection previously used? We would first have to update the [DCUtR specification](https://github.com/libp2p/specs/blob/master/relay/DCUtR.md) as it only details a _direct connection upgrade_ when a new relayed connection is established.

We could e.g. agree that the listening side of the relayed connection is allowed to initiate a new _direct connection upgrade_ in case the first direct connection dies.

---

From a more general perspective, something that might be worth noting here:

Say a `ProtocolsHandler` returned `KeepAlive::No` on `ProtocolsHandler::connection_keep_alive`. It basically signals: "I don't need the connection right now. I am fine with others using it.".

Returning `KeepAlive::No` does not stop the `ProtocolsHandler` from continuing to operate and it will continue to be `poll`ed by the task of the connection. E.g. `libp2p-ping` always returns `KeepAlive::No` as it is just an auxiliary protocol with no utility in itself.

In case the connection is kept alive by some other protocol, in other words in case the connection has not been closed, the `ProtocolsHandler` is free to change its mind, i.e. return `KeepAlive::Yes` again.

MarcoPolo: > For now you would need to initiate a new relayed connection, which would lead to the destination of the relayed connection to initiate a direct connection upgrade.

ah this makes sense. I think this is okay to do as a user.

---

> Returning KeepAlive::No does not stop the ProtocolsHandler from continuing to operate and it will continue to be polled by the task of the connection. E.g. libp2p-ping always returns KeepAlive::No as it is just an auxiliary protocol with no utility in itself.

Ah! thanks! that makes sense. The `KeepAlive` is a signal for the connection to close, and the handler itself is **only closed** when the [connection is closed](https://github.com/libp2p/rust-libp2p/blob/master/core/src/connection/pool/task.rs#L204). A connection will only be closed (barring failure modes) when all handlers using that connection return `KeepAlive::No` in the case of a `ProtocolsHandlerSelect` [code](https://github.com/libp2p/rust-libp2p/blob/master/swarm/src/protocols_handler/select.rs#L319).

If this were the only handler associate with the connection (only one behavior), returning `KeepAlive::No` would close the connection and drop the handler. But since usually there are many behaviours in a hierarchy, the top level handler in the hierarchy returns the `cmp::max` of all handlers in the hierarchy.

This is pretty nice since this means the swarm code doesn't know the difference between the handlers. In its perspective there's only a single handler for the connection.

mxinden: Marking as resolved. Please re-open in case I am missing something.

protocols/dcutr/src/handler/relayed.rs
mxinden: With https://github.com/libp2p/rust-libp2p/pull/2470 merged, I am resolving here.

mxinden:
I addressed all the comments. Thanks for the reviews!

@thomaseizinger and @MarcoPolo would you mind taking another look?

MarcoPolo (approved):
TIL about the custom codec. Neat.

mxinden:
Thanks @MarcoPolo and @thomaseizinger for the help here!

PR #2441, opened 2022-01-18
*: Rework reporting of invalid and wrong PeerIds

mxinden:
Replaces https://github.com/libp2p/rust-libp2p/pull/2428.

### DIFF
--- a/core/src/connection/pool.rs
+++ b/core/src/connection/pool.rs
@@ -22,8 +22,8 @@
 use crate::{
     connection::{
         handler::{THandlerError, THandlerInEvent, THandlerOutEvent},
-        Connected, ConnectionError, ConnectionHandler, ConnectionId, ConnectionLimit, IncomingInfo,
-        IntoConnectionHandler, PendingConnectionError, PendingInboundConnectionError,
+        Connected, ConnectionError, ConnectionHandler, ConnectionId, ConnectionLimit, Endpoint,
+        IncomingInfo, IntoConnectionHandler, PendingConnectionError, PendingInboundConnectionError,
         PendingOutboundConnectionError, PendingPoint, Substream,
     },
     muxing::StreamMuxer,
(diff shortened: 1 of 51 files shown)


rkuhn:
Thanks a lot for pushing this over the line, @mxinden!

core/src/connection/pool.rs
```diff
@@ -777,49 +777,42 @@ where
...
                         })
                         // Check peer is not local peer.
                         .and_then(|()| {
-                            if self.local_id == peer_id {
-                                Err(Error::InvalidPeerId)
+                            if self.local_id == obtained_peer_id {
+                                Err(PendingConnectionError::WrongPeerId {
+                                    obtained: obtained_peer_id,
+                                    endpoint: endpoint.clone(),
+                                })
```
bkchr, line 815: Why is that a `WrongPeerId`

bkchr (commented):
In general I think this pr broke something. We have this check that prints a warning when we connect to a bootnode and that provides a different than expected peer id. Any idea what could be broken. See the following issue: https://github.com/paritytech/substrate/issues/11709

core/src/connection/pool.rs
bkchr: CC @mxinden

tomaka: You're connecting to yourself, which isn't allowed. It's true that the error is unclear, but this was an `InvalidPeerId` error before this PR and in general has been an error for a very long time now, so I don't see how this specific change could have broken anything.

bkchr: Yeah, I also looked over this changeset and didn't find anything that could cause the issue we are seeing. However, I think that after the corresponding libp2p update in Substrate this error started to appear. There is also some internal parity devops issue about the Kusama bootnodes reporting a wrong peerid, but devops assured to me multiple times that all keys are correct. So yeah, I'm assuming now that maybe something changed it libp2p :shrug:

PR #2444, opened 2022-01-19
protocols/autonat: Add changelog and start version at v0.1.0

mxinden:
The `libp2p-autonat` crate has never been released, thus it should start at
`v0.1.0`.

@elenaf9 @dvc94ch am I missing something? Not sure why https://github.com/libp2p/rust-libp2p/pull/1672 started with `v0.20.0`.


elenaf9 (approved):
I would guess that it was a copy-paste error; I never noticed it, thank you for the fix!

Edit: the version also needs to be updated in the top-level Cargo.toml.

Issue #2434, opened 2022-01-13
`inject_connected` and `inject_disconnected` might introduce small inefficiencies in `NetworkBehaviour` implementers


divagant-martian:
Checking the implemented protocols (gossipsub, kad, relay, req/resp and floodsub, mdns and identify) all but floodsub and mdns implement at least a pair of the coupled methods: `inject_connection_(closed|establised)` and the corresponding `inject_(disconnected|connected)` from the `NetworkBehaviour` trait. 

This seems to be because the protocols require pieces of information that are not given together: i.e connection direction and whether this connection is the last/first. In the general case the means the protocol implementing the trait retrieves the same `Peer_id` twice for every first connection and final disconnection. This could be avoided removing `inject_disconnected` and `inject_connected` and simply passing the number of connections as a parameter. If a protocol cares to check for the conditions for disconnections/connections it would mean to move the check that the swarm already does about the number of connections to the protocol.

mxinden:
I don't think the fact that we have both `inject_connection_(closed|establised)` and `inject_(disconnected|connected)` is an inefficiency in a performance sense. I would argue that the cost of two function calls is negligible compared to the cost of establishing the connections themselves. Especially as `inject_(disconnected|connected)` involves a `PeerId` only, which should fit a single CPU cache line. Could you expand on which logic you would expect to be a hot-path?

I do agree that having both `inject_connection_(closed|establised)` and `inject_(disconnected|connected)` is non-intuitive from a user-experience perspective.

divagant-martian:
implementation wise, protocols typically store information in a `HashMap`. Two calls = two searches

divagant-martian:
fixed title to make the point more clear. 

On one hand it may be a bit confusing the existence of both methods; but beyond this it also splits the logic required to handle new and closed connections/peers in a two step process in each protocol. Doing what protocols require to "unify" back this logic is what typically will produce a small inefficiency. I also see as an advantage to removing the `inject_disconnected/connected`  methods we no longer needing to state the promise/contract about when is each method called with respect to the other. Some protocols actually have checks in place for this. 

It also aligns nicely with `SwarmEvent::ConnectionEstablished`

thomaseizinger:
> On one hand it may be a bit confusing the existence of both methods

This has historical reasons because originally, `rust-libp2p` did not support multiple connections to the same peer and hence only `inject_connected` and `inject_disconnected` were needed.

mxinden:
[...]

PR #2445, opened 2022-01-20
Merge inject_* paired methods

divagant-martian:
Addresses #2434 

- Include the number of other reported connections in `inject_connection_established`
- remove `inject_connected`
- Include the number of remaining reported open connections in `inject_connection_closed`
- remove `inject_disconnected`

I'll leave the changelogs for later to avoid cluttering the diff for review.
In an attempt to not interfere with the logic of the protocols I left (for now) a TODO note in the autonat behaviour regarding a mapping that seems to never be cleared


protocols/gossipsub/src/behaviour/tests.rs
```diff
@@ -228,6 +228,32 @@ mod tests {
...
+            for conn_id in peer_connections.connections.clone() {
+                let handler = gs.new_handler();
+                gs.inject_connection_closed(
+                    peer_id,
+                    &conn_id,
+                    &fake_endpoint,
+                    handler,
+                    active_connections,
+                );
+                active_connections = active_connections.saturating_sub(1);
```
mxinden:
```suggestion
                active_connections = active_connections.checked_sub(1).unwrap();
```

In which case would an overflow be valid? Would the above help catch future bugs?

protocols/relay/src/v2/relay.rs
```diff
@@ -213,8 +213,10 @@ impl NetworkBehaviour for Relay {
         connection: &ConnectionId,
         _: &ConnectedPoint,
         _handler: Either<handler::Handler, DummyProtocolsHandler>,
+        _remaining_established: usize,
     ) {
         if let Some(connections) = self.reservations.get_mut(peer) {
+            // TODO: it looks like `peer` is never removed from `self.reservations`?
```
mxinden: Good catch! Mind taking a look at https://github.com/libp2p/rust-libp2p/pull/2464 @divagant-martian?

mxinden (commented):
Thanks for looking into this! :pray:

elenaf9:
> In an attempt to not interfere with the logic of the protocols I left (for now) a TODO note in the autonat behaviour regarding a mapping that seems to never be cleared.

@divagant-martian Could you please point out to me what line/ mapping in the autonat behaviour you are referring to?

mxinden:
> > In an attempt to not interfere with the logic of the protocols I left (for now) a TODO note in the autonat behaviour regarding a mapping that seems to never be cleared.
> 
> @divagant-martian Could you please point out to me what line/ mapping in the autonat behaviour you are referring to?

I think @divagant-martian mixed up `libp2p-autonat` and `libp2p-relay`. I think the comment refers to https://github.com/libp2p/rust-libp2p/pull/2445#discussion_r795583829.

protocols/relay/src/v2/relay.rs
divagant-martian: done. I completely forgot about this finding after writing this, ty for addressing it

protocols/gossipsub/src/behaviour/tests.rs
divagant-martian: we are counting the remaining connections after removal of one, so we start in `active - 1` (because we want to inform of how many are remaining after this one). This line updates the count for _the next removal_. So this means that when the last connection is removed we are already in 0 remaining connections. So basically the suggested line would always panic for the last connection

divagant-martian: This can probably be rewritten in a better way. I'll look into it

divagant-martian: https://github.com/libp2p/rust-libp2p/pull/2445/commits/2d47bd344fe5127d666752d5e37eca8adb5a3612

protocols/relay/src/v2/relay.rs
mxinden: This was perfect. Big fan of `TODO` comments up until code is merged to `master`. Thanks.

With https://github.com/libp2p/rust-libp2p/pull/2464 merged, I think this can now be removed.

mxinden (commented):
> I'll leave the changelogs for later to avoid cluttering the diff for review.

Looks good to me overall apart from changelog entries and version bumps. Given the recent release (v0.42.0 and v0.42.1) let me know if you need help with the various semver version updates.

protocols/relay/src/v2/relay.rs
divagant-martian: nice. removed

divagant-martian:
Hey @mxinden can you share your upgrade version tools?

mxinden:
> Hey @mxinden can you share your upgrade version tools?

Version upgrades have been done via https://github.com/libp2p/rust-libp2p/pull/2469, thus only the normal changelog entries are needed here.

(The version tool is a VERY hacky Python script written by a total Python novice (me). I am not sure it is yet ready to be seen by other human beings :D )

divagant-martian:
Grand, I'll add the Changelog entries!

mxinden:
Thanks. Can you add an entry for `swarm/CHANGELOG.md` as well?

divagant-martian:
done! Ty @mxinden

mxinden:
Merge needs to incorporate https://github.com/libp2p/rust-libp2p/pull/2076 which is still using the old methods. Could you incorporate that @divagant-martian? Sorry for the trouble.

divagant-martian:
@mxinden `dcutr` does not have a CHANGELOG, how would it look like for a first time?

mxinden:
Oh, my bad. I am not a good rust-libp2p citizen. No need to add an entry for `libp2p-dcutr` here, given that it has never been released. I will add a changelog in a separate pull request. Thanks for raising it.

mxinden (approved):
Thanks for the clean up!

PR #2440, opened 2022-01-17
*: Prepare v0.42.0

mxinden:
As promised, I am proposing to cut a new release, namely `v0.42.0`.

Pull requests I would like to wait for:

- [x] https://github.com/libp2p/rust-libp2p/pull/2428
- [x] https://github.com/libp2p/rust-libp2p/pull/2443
- [x] https://github.com/libp2p/rust-libp2p/pull/2444

Any other blockers or things you would like to see included?

I would very much appreciate help testing this branch on production setups before the release.

- @AgeManning @divagant-martian would you mind running this branch on a lighthouse node?
- @kpp @tomaka  do you have some time to do a burn-in on Substrate / Polkadot?
- @thomaseizinger can you test this branch on the Comit network stack?
- @elenaf9 could you test whether things run smoothly on the iotledger stack.
- @ec2 got some time to run this against the Filecoin network via Forest?
- @rkuhn want to give this a spin via ipfs-embed?
- Any other folks running rust-libp2p in larger environments?

PR #2124, opened 2021-07-07
protocols/relay: Fix flaky innactive_connection_timeout test

mxinden:
Happening e.g. in https://github.com/libp2p/rust-libp2p/pull/2123.

**Status**

Running in a loop for now to find root cause.

PR #2450, opened 2022-01-24
protocols/autonat/tests: ignore irrelevant events

elenaf9:
CI failed in #2440  due to a flaky test in libp2p_autonat with "Unexpected Swarm Event". Wild guess is that a listener event happened. This PR changes it so that the autonat tests only panic on swarm-events that indicate an error in the autonat behaviour protocol, and ignore irrelevant events.

### DIFF
--- a/protocols/autonat/tests/test_server.rs
+++ b/protocols/autonat/tests/test_server.rs
@@ -165,8 +165,10 @@ async fn test_dial_back() {
                     };
                     break observed_client_ip;
                 }
-                SwarmEvent::IncomingConnection { .. } | SwarmEvent::NewListenAddr { .. } => {}
-                _ => panic!("Unexpected Swarm Event"),
+                SwarmEvent::IncomingConnection { .. }
+                | SwarmEvent::NewListenAddr { .. }
+                | SwarmEvent::ExpiredListenAddr { .. } => {}
+                other => panic!("Unexpected swarm event: {:?}.", other),
             }
         };
         let expect_addr = Multiaddr::empty()
@@ -184,7 +186,7 @@ async fn test_dial_back() {
                 assert_eq!(addresses[0], expect_addr);
                 probe_id
             }
-            other => panic!("Unexpected Event: {:?}", other),
+            other => panic!("Unexpected behaviour event: {:?}.", other),
         };
 
         loop {
@@ -206,8 +208,8 @@ async fn test_dial_back() {
                     break;
                 }
                 SwarmEvent::Dialing(peer) => assert_eq!(peer, client_id),
-                SwarmEvent::NewListenAddr { .. } => {}
-                _ => panic!("Unexpected Swarm Event"),
+                SwarmEvent::NewListenAddr { .. } | SwarmEvent::ExpiredListenAddr { .. } => {}
+                other => panic!("Unexpected swarm event: {:?}.", other),
             }
         }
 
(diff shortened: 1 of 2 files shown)


protocols/autonat/tests/test_server.rs
```diff
@@ -252,8 +258,11 @@ async fn test_dial_error() {
                     break;
                 }
                 SwarmEvent::Dialing(peer) => assert_eq!(peer, client_id),
-                SwarmEvent::NewListenAddr { .. } => {}
-                _ => panic!("Unexpected Swarm Event"),
```
mxinden, line 256:
``` diff
-                 _ => panic!("Unexpected Swarm Event"),
+                 e => panic!("Unexpected Swarm Event: {:?}", e),
```

I would as well be fine to only log the error for now, thus enabling us to debug the actual issue.

mxinden (approved):
Alternative suggestion. I am fine either way. Your choice :)

Thanks for preparing a fix!

protocols/autonat/tests/test_server.rs
elenaf9: I am bit undecided on this. But I think panicking by "blanked-implementation" could cause headaches in the future. If for example a new, unrelated swarm event is added that can happen here, the test could suddenly fail. For someone not familiar with this protocol it could take a while to figure out if the error is an actual error in the implementation, or just in the test itself because it is not up-to-date anymore. Even worse if it's a flaky error, like the one that caused #2440 to fail.

On the other hand I generally think it's good to test as much as possible, so that we also detect edge cases and unexpected interference. Considering that most swarm events don't just "randomly" occur, I am mostly concerned only about listener events. E.g. with a TCP transport a listener event is reported each time a network interface becomes available/ unavailable, which can happen at any time. 
So we could find a middle way by ignoring all listener events, and panicking on all other events like you proposed. Wdyt?

thomaseizinger: > a TCP transport a listener event is reported each time a network interface becomes available/ unavailable, which can happen at any time. So we could find a middle way by ignoring all listener events, and panicking on all other events like you proposed. Wdyt?

The test setup should be deterministic enough for this to not happen though, shouldn't it?

elenaf9: 👍 I'll change it to use the `MemoryTransport` and follow the above suggestion from @mxinden to panic in all cases and print the unexpected event.

elenaf9: Correction: The client has to listen on a TCP address. For DOS attack prevention the server only dials addresses that are bases on the observed sender IP, an refuses the dial if there is none. See https://github.com/libp2p/specs/pull/369. 
So we have to stick with the (nondeterministic) TCP transport in the tests.

mxinden (approved):
:pray: Thanks!

elenaf9:
Turns out that printing the swarm event was a good idea, since the issue was is not the listener event:
https://github.com/libp2p/rust-libp2p/runs/4990919849?check_suite_focus=true. Sorry about that.
I have been running the test locally in a loop for a while now, but this error never happens there. I have the wild guess that the boot delay and/ or retry_interval is too short so that probes overlap or start while the peer is still waiting for its listening addresses.
@mxinden  I was thinking about looping the test on github actions to test it, as it was also done in https://github.com/libp2p/rust-libp2p/pull/2124, but just want double check beforehand if that's okay?

mxinden:
> @mxinden I was thinking about looping the test on github actions to test it, as it was also done in #2124, but just want double check beforehand if that's okay?

Yes, for sure. Thanks for investigating!

---

> I have been running the test locally in a loop for a while now

Might as well be worth running single-threaded to mimic the CI environment.

https://docs.rs/async-std/latest/async_std/#runtime-configuration

PR #482, opened 2018-09-14
Fix old task not interrupted after replacing

tomaka:
If we receive a new connection to a peer that replaces an existing task, we have to put the former task in the `Interrupted` state so that events don't propagate.

PR #2453, opened 2022-01-27
Implement cli tool to handle key material

demfabris:
So for decoding, if I understood correctly we are reading from a IPFS config file passed as argument.
Do we check for this file under `IPFS_PATH` as well?

For encoding are we reusing this implementation: [peer-id-generator](https://github.com/libp2p/rust-libp2p/blob/master/misc/peer-id-generator/src/main.rs) ?

@mxinden 
Refs: [#482](https://github.com/rs-ipfs/rust-ipfs/issues/482)


mxinden:
Thanks for following up here @fabricio7p!

>  Do we check for this file under IPFS_PATH as well?

I would require passing a config file path for now. I would argue that libp2p shouldn't depend on IPFS concepts much. While we would depend on their file-format, at least we wouldn't depend on their filesystem conventions or environment variables. Does that make sense?

> For encoding are we reusing this implementation: peer-id-generator ?

Something along those lines. In my eyes, by default it should just generate a random identity. The feature of choosing a custom prefix should be a nice-to-have.

mxinden:
I think the [`ipfs key XXX` subcommand](https://docs.ipfs.io/reference/cli/#ipfs-key) is a good guiding example (thanks @lidel for the pointer). No need to thrive for feature parity though. The rust-libp2p client should be a lot more minimalist and without the IPFS related concepts (IPNS).

demfabris:
> I think the [`ipfs key XXX` subcommand](https://docs.ipfs.io/reference/cli/#ipfs-key) is a good guiding example (thanks @lidel for the pointer). No need to thrive for feature parity though. The rust-libp2p client should be a lot more minimalist and without the IPFS related concepts (IPNS).

Sure thing, I'll post updates soon.
Thanks for the investigation

misc/keygen/src/main.rs
```diff
@@ -0,0 +1,94 @@
...
+                // Find peer IDs in a multithreaded fashion.
+                for _ in 0..num_cpus::get() {
+                    let prefix = prefix.clone();
+                    let tx = tx.clone();
+
+                    thread::spawn(move || loop {
+                        let keypair = ed25519::Keypair::generate();
+                        let peer_id = identity::PublicKey::Ed25519(keypair.public()).to_peer_id();
+                        let base58 = peer_id.to_base58();
+                        if base58[8..].starts_with(&prefix) {
```
demfabris, line 99: `if base58[2..]` was hanging, not 100% sure why but the peerIds generated in my machine always starts with "12D3KooW" so I had to increase this prefix check

misc/keygen/src/main.rs
```diff
@@ -0,0 +1,94 @@
...
+                            tx.send((peer_id, identity::Keypair::Ed25519(keypair)))
+                                .expect("to send");
+                        }
+                    });
+                }
+
+                rx.recv().expect("to recv")
```
demfabris, line 106: exits on the first key found

demfabris:
I choose to use `argh` to be way way lighter than `clap`/`structopt`.
Did a couple of changes to your `peer-id-generator` logic because it was hanging on my PC, check review

**cli api**
```shell
# Gen random key material, with prefix at your option
keygen rand [--prefix <prefix>]

# Gen key material from IPFS config
keygen from --config <path>
```

Cargo.toml
```diff
@@ -122,6 +122,7 @@ members = [
     "misc/metrics",
     "misc/multistream-select",
     "misc/peer-id-generator",
```
mxinden: Would there still be any use for `peer-id-generator`? In other words, how about removing it within this pull request?

misc/keygen/src/cli.rs
```diff
@@ -0,0 +1,45 @@
...
+/// generate random key material
+#[argh(subcommand, name = "rand")]
+pub struct RandCmd {
+    #[argh(option, default = "String::new()")]
+    /// prefix
+    pub prefix: String,
+
+    #[argh(option, default = "String::from(\"ed25519\")")]
+    /// encryption type
+    pub r#type: String,
```
mxinden:
```suggestion
    pub r#type: String,
    
    /// Path to output config file.
    pub output_path: Option<PathBuf>,
```

When generating a new key, what do you think of supporting it to be written to a config file?

mxinden (commented):
Great. Thanks for the work here!

> I choose to use argh to be way way lighter than clap/structopt

We use `structopt` in all other examples across the project. I don't really have a strong opinion on which one to use, but as a maintainer I would prefer being consistent.

Unless you feel strongly about it, I would prefer if this one would use `structopt` as well.

demfabris:
Added `--json` flag to output json formatted. Exactly the way it's on the IPFS config file
Using `structopt`

misc/keygen/Cargo.toml
```diff
@@ -0,0 +1,18 @@
...
+repository = "https://github.com/libp2p/rust-libp2p"
+keywords = ["peer-to-peer", "libp2p", "networking"]
+categories = ["network-programming", "asynchronous"]
+
+[dependencies]
+structopt = "0.3.26"
+zeroize = "1.5.2"
+serde = { version = "1.0.136", features = ["derive"] }
+serde_json = "1.0.79"
+libp2p = { version = "0.42.2", default-features = false }
```
mxinden:
```suggestion
 libp2p-core = { path = "../../core", default-features = false, version = "0.32.0"}
```

I would suggest only depending on `libp2p-core` and to depend on the version in the relative path.

misc/keygen/Cargo.toml
```diff
@@ -0,0 +1,18 @@
...
+edition = "2021"
+authors = ["demfabris <demfabris@gmail.com>"]
+license = "MIT"
+repository = "https://github.com/libp2p/rust-libp2p"
+keywords = ["peer-to-peer", "libp2p", "networking"]
+categories = ["network-programming", "asynchronous"]
+
+[dependencies]
+structopt = "0.3.26"
+zeroize = "1.5.2"
```
mxinden: A specific reason to not allow any versions below?
```suggestion
zeroize = "1"
```

misc/keygen/src/main.rs
```diff
@@ -0,0 +1,126 @@
...
+    #[structopt(long, global = true)]
+    json: bool,
+
+    #[structopt(subcommand)]
+    cmd: Command,
+}
+
+#[derive(Debug, StructOpt)]
+enum Command {
+    /// Generate from...
```
mxinden:
```suggestion
    /// Read from config file
```


That would make for a nicer `--help` output. Currently:

```
SUBCOMMANDS:
    from    Generate from...
    help    Prints this message or the help of the given subcommand(s)
    rand    Generate random
```

mxinden (commented):
3 more comments. Otherwise this is good to go. Thanks for the new tool!

misc/keygen/Cargo.toml
```diff
@@ -0,0 +1,18 @@
...
+serde = { version = "1.0.136", features = ["derive"] }
+serde_json = "1.0.79"
+libp2p = { version = "0.42.2", default-features = false }
+base64 = "0.13.0"
+num_cpus = "1.13.1"
```
mxinden: What do you think of using https://doc.rust-lang.org/std/thread/fn.available_parallelism.html ?

demfabris: Nice @mxinden! I just learned about this new addition earlier today xD.

I'm on a busy week but will soon tackle the pending stuff

demfabris:
All points addressed 👍🏻 

ready to go?

mxinden (approved):
Looks good to me. Thanks @demfabris. This tool will be helpful going forward.

mxinden:
Unrelated test failure:

```
---- test_dial_error stdout ----
thread 'test_dial_error' panicked at 'Unexpected swarm event: Behaviour(InboundProbe(Error { probe_id: ProbeId(0), peer: PeerId("12D3KooWL97LtsmyDtSbDxJ8675kG5Ne49hZX41y5Na7h64RXZNz"), error: Response(DialError) })).', protocols/autonat/tests/test_server.rs:257:26
```

https://github.com/libp2p/rust-libp2p/runs/5510014225?check_suite_focus=true

//CC @elenaf9

PR #2464, opened 2022-01-31
protocols/relay/src/v2: Remove empty peer entries in self.reservations

mxinden:
When a peer disconnects, reservations associated with that peer are removed from
the set of reservations of the peer. In case the set of reservations for the
peer is now empty, remove the entire peer.

Same when a reservation times out.


divagant-martian:
lgtm!

divagant-martian (approved):

thomaseizinger (approved):

PR #2468, opened 2022-02-02
core/src/transport/memory: Test parse_memory_addr with PeerId

cobward:
Add some test cases for memory transport with p2p. Mostly adding this so that it is obvious from the tests that these are valid addresses for MemoryTransport.

### DIFF
--- a/core/src/transport/memory.rs
+++ b/core/src/transport/memory.rs
@@ -388,6 +388,22 @@ mod tests {
             parse_memory_addr(&"/memory/1234567890".parse().unwrap()),
             Ok(1_234_567_890)
         );
+        assert_eq!(
+            parse_memory_addr(
+                &"/memory/5/p2p/12D3KooWETLZBFBfkzvH3BQEtA1TJZPmjb4a18ss5TpwNU7DHDX6"
+                    .parse()
+                    .unwrap()
+            ),
+            Ok(5)
+        );
+        assert_eq!(
+            parse_memory_addr(
+                &"/memory/5/p2p/12D3KooWETLZBFBfkzvH3BQEtA1TJZPmjb4a18ss5TpwNU7DHDX6/p2p-circuit/p2p/12D3KooWLiQ7i8sY6LkPvHmEymncicEgzrdpXegbxEr3xgN8oxMU"
+                    .parse()
+                    .unwrap()
+            ),
+            Ok(5)
+        );
     }
 
     #[test]


mxinden (approved):
Looks good to me. Thanks.

mxinden:
Unrelated, might be of some help for @elenaf9:

```
failures:

error: test failed, to rerun pass '-p libp2p-autonat --test test_server'
---- test_dial_error stdout ----
thread 'test_dial_error' panicked at 'Unexpected swarm event: Behaviour(InboundProbe(Error { probe_id: ProbeId(1), peer: PeerId("12D3KooWFP21kH18bMBCoywZN2UpKH2J93TTZogiztMQgk91RuKn"), error: Response(DialRefused) })).', protocols/autonat/tests/test_server.rs:258:26
```

https://github.com/libp2p/rust-libp2p/runs/5042997171?check_suite_focus=true

core/src/transport/memory.rs
```diff
@@ -388,6 +388,22 @@ mod tests {
...
+            parse_memory_addr(
+                &"/memory/5/p2p/12D3KooWETLZBFBfkzvH3BQEtA1TJZPmjb4a18ss5TpwNU7DHDX6"
+                    .parse()
+                    .unwrap()
+            ),
+            Ok(5)
+        );
+        assert_eq!(
+            parse_memory_addr(
+                &"/memory/5/p2p/12D3KooWETLZBFBfkzvH3BQEtA1TJZPmjb4a18ss5TpwNU7DHDX6/tcp/150"
```
mxinden:
```suggestion
                &"/memory/5/p2p/12D3KooWETLZBFBfkzvH3BQEtA1TJZPmjb4a18ss5TpwNU7DHDX6"
```

Ah, missed this one. Why add a `/tcp` at the end? While well-formed, this is semantically not correct.

cobward: ~Thanks for pointing this out. I'm pretty new to multiaddr, so I'm not sure why this isn't semantically correct. I wanted to add a test here to show that protocols after the `p2p` protocol were also preserved. Would something like `/memory/5/p2p/<peer-id>/p2p-circuit/p2p/<peer-id>` make sense semantically? If not could you suggest an address that does?~

I've read up on it and now I understand.

cobward: Also I would ask: if the structure of semantically correct addresses is well defined, should the `parse_memory_addr` function fail to parse well-formed but invalid addresses, or do we prefer to let that error happen when attempting a connection?

mxinden: `/memory/5/p2p/<peer-id>/p2p-circuit/p2p/<peer-id>` yes, that would be correct.

> Also I would ask: if the structure of semantically correct addresses is well defined, should the parse_memory_addr function fail to parse well-formed but invalid addresses, or do we prefer to let that error happen when attempting a connection?

In the ideal case semantically incorrect addresses would be prevented at the type system level. I have a couple of thoughts on how one could do that within `rust-multiaddr`, though I haven't had time to explore it yet. For now most libp2p Multiaddr related functions take the laissez faire approach, only checking validity of the parts they are interested in.

In the concrete case of `parse_memory_addr`, we could make it more strict, but in my eyes that would (a) add a lot of complexity and (b) only solve the problem in the memory transport and not across libp2p.

rkuhn: Is there a specification for what is semantically correct? Especially with the example of p2p-circuit in mind, is it not possible that currently unsupported combinations will become valid in the future? If there were a poll, I’d vote for keeping Multiaddr as a purely syntactical construct — semantics are defined by the middleware that interprets them (which includes rejection).

cobward: Yeah my above comments and questions came more from lack of understanding. I absolutely agree that only syntax should be checked here, however I think it's reasonable to not include semantically "wrong" test cases.

rkuhn: Hehe, given this discussion I’d argue that it is valuable to assert “syntax only” by including some currently known-nonsensical examples in the tests :-)

mxinden: > 

Hehe, given this discussion I’d argue that it is valuable to assert “syntax only” by including some currently known-nonsensical examples in the tests :-)


:D If we go this far, we should as well generate those addresses either via quickcheck or use fuzzing.

PR #2469, opened 2022-02-02
*: Update multihash and multiaddr

vmx:
multiaddr also depends on multihash, hence updating it as well.


mxinden (approved):
Thank you for the help Volker!

PR #2470, opened 2022-02-02
swarm/src/protocols_handler: Document alternative to Event::Close

mxinden:
There are 3 ways to close a connection:

1. `ProtocolsHandlerEvent::Close`
2. `KeepAlive::No` via `ProtocolsHandler::connection_keep_alive`
3. `NetworkBehaviourAction::CloseConnection`

This commit references (2) as an alternative to (1) in the documentation
of (1).

Do you think this is helpful @MarcoPolo?

### DIFF
--- a/swarm/src/protocols_handler.rs
+++ b/swarm/src/protocols_handler.rs
@@ -321,8 +321,8 @@ pub enum ProtocolsHandlerEvent<TConnectionUpgrade, TOutboundOpenInfo, TCustom, T
     ///
     /// Note this will affect all [`ProtocolsHandlerEvent`] handling this
     /// connection, in other words it will close the connection for all
-    /// [`ProtocolHandler`]s. To signal that one has no more need for the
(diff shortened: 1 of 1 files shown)


MarcoPolo (approved):
Yes, I think this is clearer. Thanks for adding the reference to keepAlive, since I can imagine myself reaching for close instead of setting keepAlive.

swarm/src/protocols_handler.rs
```diff
@@ -318,6 +318,13 @@ pub enum ProtocolsHandlerEvent<TConnectionUpgrade, TOutboundOpenInfo, TCustom, T
     },
 
     /// Close the connection for the given reason.
+    ///
+    /// Note this will affect all [`ProtocolsHandlerEvent`] handling this
```
mxinden:
```suggestion
    /// Note this will affect all [`ProtocolsHandler`]s handling this
```

PR #2477, opened 2022-02-04
Tiny refactoring to reduce cloning

nazar-pc:
Not very meaningful, but it bothered me that `&Key` is only there to clone it later and in my code I have to throw away owned value that I could have provided directly.

It is more idiomatic to require owned value instead of reference that will be unconditionally cloned later.

### DIFF
--- a/protocols/kad/CHANGELOG.md
+++ b/protocols/kad/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 - Update to `libp2p-swarm` `v0.34.0`.
 
-- Require owned key in `get_record()` method (see [PR 2477])
+- Require owned key in `get_record()` method (see [PR 2477]).
 
 [PR 2477]: https://github.com/libp2p/rust-libp2p/pull/2477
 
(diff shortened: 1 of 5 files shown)


thomaseizinger (approved):
Reasoning makes sense to me!

Can you add a changelog entry please?

protocols/kad/CHANGELOG.md
```diff
@@ -4,6 +4,10 @@
 
 - Update to `libp2p-swarm` `v0.34.0`.
 
+- Require owned key in `get_record()` method (see [PR 2477])
```
```suggestion
- Require owned key in `get_record()` method (see [PR 2477]).
```

mxinden (approved):
Looks good to me. Thanks @nazar-pc!

mxinden:
Could you address the test failures @nazar-pc?

nazar-pc:
Interestingly it didn't run all tests with `cargo test` when running at the root (probably because of the root package), should be fixed now.

mxinden:
Could you also run Rust fmt :innocent:  ?

PR #2481, opened 2022-02-05
protocols/gossipsub: Emit gossip of all non empty topics

divagant-martian:
tiny censoring bug in gossipsub. When a topic had no messages to gossip about, `emit_gossip` was returning early


AgeManning (approved):
Nice. Good catch!

mxinden:
:rocket: thanks!

Let me know in case you need a release with this included sometime soon. I also plan to release `v0.43.0` in the next couple of weeks.

PR #2482, opened 2022-02-06
document no blocking code in networkbehaviour impl

### LINKED ISSUE #2474: Blocking code is not allowed in sync #[behaviour(event_process = true)] trait implementation
As the traits generated by event_process=true are standard sync traits, there's no warning to the user that blocking code is not allowed.

I realize that async fn in traits isn't possible as of yet, but in the meantime it seems wise to document this, or alternatively devise a mechanism whereby these sync method calls are spawned on a separate thread, or with block_on.

Something like this already exists today:

```
        libp2p::swarm::SwarmBuilder::new(transport, mybehaviour, peer_id)
            .executor(Box::new(|fut| {
                tokio::spawn(fut);
            }))
            .build()
```

Perhaps we could allow the user to pass an additional closure, for blocking calls, like so:

```
        libp2p::swarm::SwarmBuilder::new(transport, mybehaviour, peer_id)
            .executor(Box::new(|fut| {
                tokio::spawn(fut);
            }))
            .blocking_spawn(Box::new(|function| {
                tokio::task::spawn_blocking(function);
            }))
            .build()
```
(Idk if my closure is correct)

Alternatively we could add feature flags for various executors, which would remove the need to pass these kinds of closures (I think, correct me if that's wrong).

Would it be possible to use [the async trait crate](https://crates.io/crates/async-trait) for this?

Attached you can find a demonstration of problems that can occur when using blocking code in the trait implementation:

https://user-images.githubusercontent.com/18269095/152414077-06be9795-fdbc-4e3e-8983-6f7b12808b8c.mp4

Frederik-Baetens:
Closes  #2474

Opinions welcome on whether the link to the blog post about blocking code is appropriate. It's a nice explanation, but there's little guarantee that the link will still work in ~5 years.

### DIFF
--- a/swarm/src/behaviour.rs
+++ b/swarm/src/behaviour.rs
@@ -155,10 +155,9 @@ type THandlerInEvent<THandler> =
 /// Because these implementations are granted exclusive access to the [`NetworkBehaviour`],
 /// [blocking code](https://ryhl.io/blog/async-what-is-blocking/) in these implementations will
 /// block the entire [`Swarm`](crate::Swarm) from processing new events, since the swarm cannot progress
-/// without also having exclusive access to the [`NetworkBehaviour`]. A better alternative is to spawn
-/// blocking or asynchronous tasks on a (separate) executor, which may be provided to the behaviour's
-/// constructor.
-///
+/// without also having exclusive access to the [`NetworkBehaviour`]. A better alternative is to execute
+/// blocking or asynchronous logic on a separate task, perhaps with the help of a bounded channel to
+/// maintain backpressure. The sender for the channel could be included in the NetworkBehaviours constructor.
 /// Optionally one can provide a custom `poll` function through the `#[behaviour(poll_method =
 /// "poll")]` attribute. This function must have the same signature as the [`NetworkBehaviour#poll`]
 /// function and will be called last within the generated [`NetworkBehaviour`] implementation.
--- a/swarm/src/behaviour.rs
+++ b/swarm/src/behaviour.rs
@@ -155,7 +155,7 @@ type THandlerInEvent<THandler> =
 /// Because these implementations are granted exclusive access to the [`NetworkBehaviour`],
 /// [blocking code](https://ryhl.io/blog/async-what-is-blocking/) in these implementations will
 /// stop the entire swarm from processing new events, since the swarm cannot progress without also
-/// having access exclusive access to the NetworkBehaviour. A better alternative is to spawn
+/// having exclusive access to the [`NetworkBehaviour`]. A better alternative is to spawn
 /// blocking or asynchronous tasks on a (separate) executor, which may be provided to the behaviour's
 /// constructor.
 ///
--- a/swarm/src/behaviour.rs
+++ b/swarm/src/behaviour.rs
@@ -152,6 +152,13 @@ type THandlerInEvent<THandler> =
 /// struct members are delegated to [`NetworkBehaviourEventProcess`] implementations. Those must be
 /// provided by the user on the type that [`NetworkBehaviour`] is derived on.
 ///
+/// Because these implementations are granted exclusive access to the [`NetworkBehaviour`],
+/// [blocking code](https://ryhl.io/blog/async-what-is-blocking/) in these implementations will
+/// stop the entire swarm from processing new events, since the swarm cannot progress without also
+/// having access exclusive access to the NetworkBehaviour. A better alternative is to spawn
+/// blocking or asynchronous tasks on a (separate) executor, which may be provided to the behaviour's
+/// constructor.
+///
 /// Optionally one can provide a custom `poll` function through the `#[behaviour(poll_method =
 /// "poll")]` attribute. This function must have the same signature as the [`NetworkBehaviour#poll`]
 /// function and will be called last within the generated [`NetworkBehaviour`] implementation.
--- a/swarm/src/behaviour.rs
+++ b/swarm/src/behaviour.rs
@@ -154,8 +154,8 @@ type THandlerInEvent<THandler> =
 ///
 /// Because these implementations are granted exclusive access to the [`NetworkBehaviour`],
 /// [blocking code](https://ryhl.io/blog/async-what-is-blocking/) in these implementations will
-/// block the entire swarm from processing new events, since the swarm cannot progress without also
-/// having exclusive access to the [`NetworkBehaviour`]. A better alternative is to spawn
+/// block the entire [`Swarm`](crate::Swarm) from processing new events, since the swarm cannot progress
+/// without also having exclusive access to the [`NetworkBehaviour`]. A better alternative is to spawn
 /// blocking or asynchronous tasks on a (separate) executor, which may be provided to the behaviour's
 /// constructor.
 ///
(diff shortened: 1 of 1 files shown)


swarm/src/behaviour.rs
```diff
@@ -152,6 +152,11 @@ type THandlerInEvent<THandler> =
 /// struct members are delegated to [`NetworkBehaviourEventProcess`] implementations. Those must be
 /// provided by the user on the type that [`NetworkBehaviour`] is derived on.
 ///
+/// Because these implementations are granted exclusive access to the [`NetworkBehaviour`],
+/// [blocking code](https://ryhl.io/blog/async-what-is-blocking/) in these implementations will
+/// stop the entire swarm from processing new events, since the swarm cannot progress without also
+/// having access exclusive access to the NetworkBehaviour.
```
rkuhn: How about adding something like

> A better alternative is to spawn blocking or asynchronous tasks on a (separate) executor, which may be provided to the behaviour’s constructor.

It would be nice if the network pool’s `executor` were available to (sub)behaviours for spawning tasks. @mxinden what do you think?

rkuhn:
If the blog link stops working at some point, we can fix the doc comment to point to a new resource — I think the wording “cannot make progress” should be clear enough to let the reader figure it out without the external text as well.

swarm/src/behaviour.rs
Frederik-Baetens: Added it, thanks for the good suggestion!

Frederik-Baetens: And I think the suggestion of providing the async executor in the networkbehaviour implementation would be great, I went with the dirty solution of rewriting the function I needed to call with blocking code. In the future since it's a relatively fast fsync call, but I'll probably try to provide access to the executor by including it in the NetworkBehaviour struct as you wrote in your suggested addition.

swarm/src/behaviour.rs
```diff
@@ -152,6 +152,13 @@ type THandlerInEvent<THandler> =
 /// struct members are delegated to [`NetworkBehaviourEventProcess`] implementations. Those must be
 /// provided by the user on the type that [`NetworkBehaviour`] is derived on.
 ///
+/// Because these implementations are granted exclusive access to the [`NetworkBehaviour`],
+/// [blocking code](https://ryhl.io/blog/async-what-is-blocking/) in these implementations will
+/// stop the entire swarm from processing new events, since the swarm cannot progress without also
+/// having access exclusive access to the NetworkBehaviour. A better alternative is to spawn
```
```suggestion
/// having exclusive access to the [`NetworkBehaviour`]. A better alternative is to spawn
```

swarm/src/behaviour.rs
```diff
@@ -152,6 +152,13 @@ type THandlerInEvent<THandler> =
 /// struct members are delegated to [`NetworkBehaviourEventProcess`] implementations. Those must be
 /// provided by the user on the type that [`NetworkBehaviour`] is derived on.
 ///
+/// Because these implementations are granted exclusive access to the [`NetworkBehaviour`],
+/// [blocking code](https://ryhl.io/blog/async-what-is-blocking/) in these implementations will
+/// stop the entire swarm from processing new events, since the swarm cannot progress without also
```
```suggestion
/// block the entire swarm from processing new events, since the swarm cannot progress without also
```

```suggestion
/// stop the entire [`Swarm`](crate::Swarm) from processing new events, since the swarm cannot progress without also
```

swarm/src/behaviour.rs
mxinden: > It would be nice if the network pool’s executor were available to (sub)behaviours for spawning tasks. @mxinden what do you think?

Indeed, that would be an option and would make things a bit more consistent instead of requiring users to pass yet another executor to the behaivour.

What would be your use-case @rkuhn?

In my eyes, rust-libp2p greatly benefits of being designed around a single synchronous hierarchical state machine (`Swarm` + `NetworkBehaviour`) + one hierarchical state machine per connection (managed via `Pool`).

rkuhn: My use-case is a converter between Rust structs (from a channel) and substream frames (going into AsyncWrite). Polling the whole swarm (easily a hierarchy of a dozen or more state machines, most of them uninterested), for passing a single message through is wasteful. Or sending a notification from a behaviour to client code via a bounded channel — unbounded channels creep me out, they’re an OOM waiting to happen. In the second case the computational overhead may in some cases be tolerable, but the coding overhead is burdensome — FuturesUnordered galore.

Frederik-Baetens:
Added your suggestions, thanks for the review!

swarm/src/behaviour.rs
mxinden: > unbounded channels creep me out, they’re an OOM waiting to happen.

:+1: 

> My use-case is a converter between Rust structs (from a channel) and substream frames (going into AsyncWrite). Polling the whole swarm (easily a hierarchy of a dozen or more state machines, most of them uninterested), for passing a single message through is wasteful.

Does this has to happen in the main task (`Swarm` and `NetworkBehaviour`) or can this happen in the connection task (`ProtocolsHandler`)?

swarm/src/behaviour.rs
```diff
@@ -152,6 +152,13 @@ type THandlerInEvent<THandler> =
 /// struct members are delegated to [`NetworkBehaviourEventProcess`] implementations. Those must be
 /// provided by the user on the type that [`NetworkBehaviour`] is derived on.
 ///
+/// Because these implementations are granted exclusive access to the [`NetworkBehaviour`],
+/// [blocking code](https://ryhl.io/blog/async-what-is-blocking/) in these implementations will
+/// block the entire [`Swarm`](crate::Swarm) from processing new events, since the swarm cannot progress
+/// without also having exclusive access to the [`NetworkBehaviour`]. A better alternative is to spawn
+/// blocking or asynchronous tasks on a (separate) executor, which may be provided to the behaviour's
+/// constructor.
```
```suggestion
/// without also having exclusive access to the [`NetworkBehaviour`]. A better alternative is to execute
/// blocking or asynchronous logic on a separate task.
```

I don't think spawning a new task on events is the right approach for most applications. You would need some external back-pressure mechanism, in other words you need some mechanism to stop you from spawning a large amount of tasks. (Yes they are cheap, still they break back-pressure.)

How about the wording above? If you want to execute your blocking logic somewhere else, I would suggest having a dedicated task or a limited set of dedicated tasks.

Frederik-Baetens: Thanks, that seems like a better idea. I added the concrete suggestion to use a bounded channel, so that it's more obvious for new users how they could accomplish such a thing.

swarm/src/behaviour.rs
```diff
@@ -152,6 +152,12 @@ type THandlerInEvent<THandler> =
 /// struct members are delegated to [`NetworkBehaviourEventProcess`] implementations. Those must be
 /// provided by the user on the type that [`NetworkBehaviour`] is derived on.
 ///
+/// Because these implementations are granted exclusive access to the [`NetworkBehaviour`],
+/// [blocking code](https://ryhl.io/blog/async-what-is-blocking/) in these implementations will
+/// block the entire [`Swarm`](crate::Swarm) from processing new events, since the swarm cannot progress
+/// without also having exclusive access to the [`NetworkBehaviour`]. A better alternative is to execute
+/// blocking or asynchronous logic on a separate task, perhaps with the help of a bounded channel to
+/// maintain backpressure. The sender for the channel could be included in the NetworkBehaviours constructor.
```
mxinden, line 160:
```suggestion
/// maintain backpressure. The sender for the channel could be included in the NetworkBehaviours constructor.
///
```

mxinden (approved):
Thanks for bearing with me @Frederik-Baetens :)

Issue #2449, opened 2022-01-23
libp2p-websocket: replace /wss with /tls/ws


tomaka:
According to the multiaddr spec, the `/wss` specifier is deprecated and `/tls/ws` should be preferred.
`libp2p-websocket` only supports `/wss`.

(I've taken the liberty to mark this as "easy", as it's just about interpreting the multiaddr)

jochasinga:
Shouldn't this change be made on the multiaddr crate too?

mxinden:
> Shouldn't this change be made on the multiaddr crate too?

Should be done via https://github.com/multiformats/rust-multiaddr/pull/48/.

jochasinga:
I can tackle this @mxinden @tomaka.

jochasinga:
Looks to me like even `/wss` scheme isn't supported currently? Keep getting `MultiaddrNotSupported("/ip6/::1/tcp/0/wss")` running test.

Since [multiformats/rust-multiaddr#48 ](https://github.com/multiformats/rust-multiaddr/pull/48) added tls support, will this be just upgrading libp2p-websocket to use it?

jochasinga:
Does this test case for rust-multiaddr make sense @tomaka?

```rust

// rust-multiaddr/src/from_url.rs

#[test]
fn normal_usage_wss() {
    let addr = from_url("wss://127.0.0.1:8000").unwrap();
    assert_eq!(addr, "/ip4/127.0.0.1/tcp/8000/tls/ws".parse().unwrap());
    assert_eq!(addr, "/ip4/127.0.0.1/tcp/8000/wss".parse().unwrap());
}
```

Basically, when parsing a multiaddress, when a `tls` is found, we peek to see if the next one is "ws" and write that off as `Protocol::Wss(..)`.

tomaka:
I don't think it's a good idea to make `/tls/ws` and `/wss` equal. This is too magic and could lead to violated assumptions.
In my opinion the only change needed is making `libp2p-websocket` support `/tls/ws`.

navasvarela:
Hello everyone, are there any updates on TLS support for websockets?

thomaseizinger:
> Hello everyone, are there any updates on TLS support for websockets? 

Support is there and there is an example in the docs! This ticket is only about the deprecated multiaddr component.

navasvarela:
Thank you!

navasvarela:
> > Hello everyone, are there any updates on TLS support for websockets?
> 
[...]

PR #2487, opened 2022-02-07
Add get_address_proto to handle trailing "tls/ws" in multiaddr

jochasinga:
Before returning `Ws` protocol when hitting the last trailing `/ws` in a multiaddr,
check to see if the next path segment is `/tls`. If it is, then return `Wss` instead.
This addresses #2449 the deprecated `/wss` specifier for the new `/tls/ws` in the spec.

### DIFF
--- a/transports/websocket/src/framed.rs
+++ b/transports/websocket/src/framed.rs
@@ -197,7 +197,6 @@ where
                 ListenerEvent::NewAddress(mut a) => {
                     a = a.with(proto.clone());
                     debug!("Listening on {}", a);
-                    println!("Listening on {}", a);
                     ListenerEvent::NewAddress(a)
                 }
                 ListenerEvent::AddressExpired(mut a) => {
--- a/transports/websocket/src/framed.rs
+++ b/transports/websocket/src/framed.rs
@@ -110,34 +110,32 @@ impl<T> WsConfig<T> {
         self.use_deflate = flag;
         self
     }
-
-    /// Look backward to if addr has a trailing "/ws" and returns a Wss protocol if "/tls" was found.
-    pub(crate) fn get_address_proto(
-        addr: &mut Multiaddr,
-    ) -> Result<(bool, Protocol<'static>), String> {
-        match addr.pop() {
-            Some(p @ Protocol::Wss(_)) => {
-                let mut addr = addr.clone();
-                if let Some(Protocol::Tls) = addr.pop() {
-                    return Err(format!("{} is not a websocket multiaddr", addr));
-                } else {
-                    Ok((true, p))
-                }
-            }
-            Some(ref p @ Protocol::Ws(ref s)) => {
-                let mut addr = addr.clone();
-                if let Some(Protocol::Tls) = addr.pop() {
-                    Ok((true, Protocol::Wss(s.clone())))
-                } else {
-                    Ok((false, p.clone()))
-                }
-            }
-            _ => Err(format!("{} is not a websocket multiaddr", addr)),
-        }
-    }
 }
 
 type TlsOrPlain<T> = EitherOutput<EitherOutput<client::TlsStream<T>, server::TlsStream<T>>, T>;
+
+/// Look backward to if addr has a trailing "/ws" and returns a Wss protocol if "/tls" was found.
+pub(crate) fn get_address_proto(
+    mut addr: Multiaddr,
+) -> Result<(bool, Protocol<'static>, Multiaddr), String> {
+    match addr.pop() {
+        Some(p @ Protocol::Wss(_)) => {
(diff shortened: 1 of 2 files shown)


transports/websocket/src/framed.rs
```diff
@@ -98,6 +98,31 @@ impl<T> WsConfig<T> {
         self.use_deflate = flag;
         self
     }
+
+    /// Look backward to if addr has a trailing "/ws" and returns a Wss protocol if "/tls" was found.
+    pub(crate) fn get_address_proto(
```
thomaseizinger: Does this have to be on `Self`? Could be a free function :)

thomaseizinger: Also, we could refactor this to not mutate its argument and instead consume the argument and return a new address.

thomaseizinger (commented):
Thanks!

Some ideas for how this could be improved :)

transports/websocket/src/framed.rs
jochasinga: Sounds good! Thanks. You mean a public function? Nothing private about it, just feels like it's a helper rather than an API.

jochasinga: And by not mutating its argument and returning a new address, do you mean the function also returns another `Multiaddr` back to the caller? i.e. something like

```rust
fn get_address_proto(addr: Multiaddr) -> Result<(bool, Protocol<'static>, Multiaddr), String>
```

Consuming the `Multiaddr` will likely cause one additional clone as opposed to borrowing mutably since we need to pop the `Multiaddr` one more time to peek at the second to last path segment while keeping the original addr for the caller.

tomaka:
I guess this is a first step :+1: , but for #2449 to be finished we need a couple more changes, such as producing `/tls/ws` when receiving an incoming connection. Ultimately `Protocol::Wss` shouldn't be used except specifically for backwards compatibility.

jochasinga:
> I guess this is a first step 👍 , but for #2449 to be finished we need a couple more changes, such as producing `/tls/ws` when receiving an incoming connection. Ultimately `Protocol::Wss` shouldn't be used except specifically for backwards compatibility.

Could you elaborate on those two points @tomaka ?

transports/websocket/src/framed.rs
thomaseizinger: > Consuming the Multiaddr will likely cause one additional clone

Looking at the current usage, it is already cloned before being passed in or am I missing something?

The reasoning was: If the caller is okay with mutation (which it currently seems to be), then they can just pass over ownership and use the new return value so it shouldn't make a difference but a non-mutable API would be easier to understand IMO.

thomaseizinger: > You mean a public function? Nothing private about it, just feels like it's a helper rather than an API.

Uhm, still pub(crate) or whatever it needs to be but just not on `Self` and instead free-standing (outside an `impl` block).

jochasinga: How's this? [1e2117a](https://github.com/libp2p/rust-libp2p/pull/2487/commits/1e2117ad5a7046b6123f34e627851d075169d708)

thomaseizinger: Yeah that is much easier to understand IMO, thank you :)

jochasinga:
@tomaka or @thomaseizinger some help on finishing this up?

transports/websocket/src/framed.rs
```diff
@@ -142,14 +150,15 @@ where
...
-            .listen_on(inner_addr)
+            .listen_on(next_addr)
             .map_err(|e| e.map(Error::Transport))?;
         let listen = transport
             .map_err(Error::Transport)
             .map_ok(move |event| match event {
                 ListenerEvent::NewAddress(mut a) => {
                     a = a.with(proto.clone());
                     debug!("Listening on {}", a);
+                    println!("Listening on {}", a);
```
thomaseizinger: I don't think we should include `println!`s.

thomaseizinger (commented):
I can't comment on what @tomaka said unfortunately. I am not familiar enough with the code here.

transports/websocket/src/framed.rs
jochasinga: Of course that slipped through the crack.

mxinden:
I am not particularly familiar with this code. @tomaka your input would be more helpful here.

In regards to above comment on this still missing some parts, if I am not mistaken the code below would append a `ws` or `wss`, but not a `tls/ws`:

https://github.com/libp2p/rust-libp2p/blob/90140a6eafd8a500e654b14040941f9f5074474d/transports/websocket/src/framed.rs#L165-L166

Is that of some help @jochasinga?

PR #2493, opened 2022-02-08
protocols/gossipsub: Do not overwrite msg's peers if put again into mcache

divagant-martian:
Putting a message in the mcache with a duplicate key would reset its peers. This makes it so that duplicates do not alter the cache


AgeManning (approved):
Yep nice catch

PR #2495, opened 2022-02-09
Derive `Debug` and `Clone`/`Copy` on a few data structures

nazar-pc:
This makes it possible to derive `Debug` for structs containing configs for different components downstream as well as `Clone` and even `Copy` in some cases (not all configs are clonable though).

Clone/Copy is more controversial, let me know if it is undesirable.


protocols/gossipsub/src/metrics.rs
```diff
@@ -40,6 +40,7 @@ const DEFAULT_MAX_TOPICS: usize = 300;
 // store metrics.
 const DEFAULT_MAX_NEVER_SUBSCRIBED_TOPICS: usize = 50;
 
+#[derive(Debug, Clone, Copy)]
```
rkuhn: Copy is a rather strict promise, I’d only add it where there is a clear benefit that would justify some pain to keep it (like for PeerId). For `*Config` structures it can easily happen that some String or other heap allocated structure is added in the future, which would then be a larger breaking change (field addition can for users be done in a mostly compatible fashion using `default` and `with_x()` functions).

rkuhn (commented):
Nice one!

protocols/gossipsub/src/metrics.rs
nazar-pc: Indeed. On a similar note I saw that some configs have `#[non_exhaustive]` on them and others do not. Maybe it is related to the fact that some are considered to be "done" and some are still evolving, but I found it interesting as well.

thomaseizinger: I think it is okay to make that promise now and perhaps remove it again later. Yes it is a breaking change but it is

a) easily fixed
b) may not break users in all circumstances (only if they use `Copy`)
c) we make minor releases (which are breaking < `0.x`) all the time

Hence I think it is okay to give users you may benefit from this the benefits now and accept it if we need to break it again later.

@mxinden Thoughts?

thomaseizinger (approved):

rkuhn:
Switching between `x.clone()` and `*x` can be a lot of work — and clippy users will be strongly pushed towards removing the clones as soon as we add the Copy impl. OTOH runtime cost of cloning on these structures is zero (AFAIK — happy to be proven wrong), so not having Copy is future proof and has only a syntactical downside.

nazar-pc:
I only added `Copy` where I thought it is unlikely that `Clone` will be useful, but I'm fine changing it either way.

thomaseizinger:
> Switching between `x.clone()` and `*x` can be a lot of work

I agree, if it is a widely used struct. But we are talking about the `Config` structs here which, unless someone is doing something really funky are usually only used a handful of times.

I am okay with not adding `Copy` if people feel strongly about it.

core/src/identity.rs
```diff
@@ -63,7 +63,7 @@ use std::convert::{TryFrom, TryInto};
 /// let keypair = Keypair::rsa_from_pkcs8(&mut bytes);
 /// ```
 ///
-#[derive(Clone)]
+#[derive(Debug, Clone)]
```
mxinden, line 66: I would like to be extra careful with private key material. E.g. I would like to prevent users ever accidentally printing private key material to stdout, potentially being persisted in some log store, ...

How about a custom `Debug` implementation that only prints the key type. In case specific `xxx::Keypair` `Debug` implementations ever accidentally leak private key material, it would still not be exposed.

Next to the key type we could include the `PeerId`.

core/src/identity/rsa.rs
```diff
@@ -30,7 +30,7 @@ use std::{fmt, sync::Arc};
 use zeroize::Zeroize;
 
 /// An RSA keypair.
-#[derive(Clone)]
+#[derive(Debug, Clone)]
```
mxinden: With my suggestion above, we would not need this `Debug`.

protocols/gossipsub/src/metrics.rs
mxinden: Is there much value in offering `Copy` for this particular config? In other words, I would suspect downstream code to only need to clone it once or twice, thus I don't see the benefit of offering `Copy` here.

> Indeed. On a similar note I saw that some configs have #[non_exhaustive] on them and others do not. Maybe it is related to the fact that some are considered to be "done" and some are still evolving, but I found it interesting as well.

Correct. This is a pattern used in the past. I personally don't see much value in it. As a downstream user I would prefer a breaking change over accidentally missing some new config option or event variant.

mxinden (commented):
A couple of comments. What do folks think?

core/src/identity.rs
thomaseizinger: The folks over a `rust-bitcoin` opted for printing a cryptographic _hash_ of the private key. That way you can still distinguish different keys from each other in the logs but no sensitive information is leaked.

Could apply the same thing here.

mxinden: Would one not get the same benefit from the `PeerId`? If so, I would prefer the `PeerId` over a hash of the private key.

core/src/identity/rsa.rs
nazar-pc: A good practice is to have `Debug` on all public data structures (pedantic preset in clippy suggests this IIRC).

Turns out other keypairs (ed25519 and secp256k1) already implemented `Debug` by printing public key, doing the same here for symmetry.

protocols/gossipsub/src/metrics.rs
nazar-pc: Removed `Copy` from all places where I have added as it seems controversial/not useful.

mxinden (approved):

PR #2501, opened 2022-02-09
swarm/src/behaviour: Clarify custom networkbehaviour deriviation

Frederik-Baetens:
I'm creating this pull request, because the following line made it seem like one could use `event_process=false` without specifying an out_event:

> By default the derive sets the [`NetworkBehaviour::OutEvent`] as `()` but this can be overridden with `#[behaviour(out_event = "AnotherType")]`.

I further clarified that the user needs to choose between two mutually exclusive options, one being the out_event, the other being event_process=true.


thomaseizinger:
> I'm creating this pull request, because the following line made it seem like one could use `event_process=false` without specifying an out_event:

Technically speaking, you can do that IF your inner `NetworkBehaviour` also has an `OutEvent` of `()`.

swarm/src/behaviour.rs
```diff
@@ -77,11 +77,20 @@ type THandlerInEvent<THandler> =
 /// it will delegate to each `struct` member and return a concatenated array of all addresses
 /// returned by the struct members.
 ///
-/// By default the derive sets the [`NetworkBehaviour::OutEvent`] as `()` but this can be overridden
-/// with `#[behaviour(out_event = "AnotherType")]`.
+/// When creating a custom [`NetworkBehaviour`], you must choose one of two methods to respond to
```
thomaseizinger, line 83: There is also the possibility of:

- Setting `out_event` to a type
- Setting `event_process = true`

Which allows you to use `NetworkBehaviourEventProcess` but still emit an event as part of `poll`. You will likely need `poll_method` for that as well. This is useful if the `NetworkBehaviour` you are deriving is meant to be composed again and is not the root one!

thomaseizinger (approved):
Thank you!

If we are enumerating all the options, I think we should also explicitly talk about the third one that is possible.

swarm/src/behaviour.rs
Frederik-Baetens: Is there an example of this? I'd suggest something like the following to document this, but an example would likely help as I don't really understand what a custom_poll method should be used for:

> It is also possible to combine a custom `out_event` with `event_process=true`. This lets you handle emitted events in the `NetworkEventBehaviourProcess` trait implementations, while still emitting events as part of `poll`. This is useful if the `NetworkBehaviour` you are deriving is not the root behaviour and is meant to be composed again, .

thomaseizinger: I don't have an example at hand.

If you set `OutEvent` to something else other than `()`, you need a custom `poll` method so you can do `return NetworkBehaviourAction::GenerateEvent`.The event you would likely `pop` off a `Vec` or some other data structure in your `NetworkBehaviour` and you would `push` into the data structure in the `NetworkBehaviourEventProcess` implementations.

Frederik-Baetens: Could you suggest some documentation to add for this third possibility? I don't have a firm grasp on how that would work, so I think it would be more convenient if someone more familiar suggested something.

Frederik-Baetens: > If you set OutEvent to something else other than (), you need a custom poll method

Is this only the case when event_process=true? I don't remember having to add a custom poll method when having my own OutEvent.

thomaseizinger: Essentially, the three parameters `out_event`, `event_process` and `poll_method` are orthogonal. The derived implementation looks something like this:

```rust
#[derive(NetworkBehaviour)]
#[behaviour(out_event = "MyEvent", poll_method = "my_poll", event_process = true)]
struct FooBehaviour {
	inner: BarBehaviou
}

impl FooBehaviour {
	fn my_poll(&mut self, params: impl PollParameters) -> Poll<...> {
		// do some work on every poll
	}
}

impl NetworkBehaviourEventProcess<BarEvent> for FooBehaviour {
	fn inject_event(&mut self, event: BarEvent) {

	}
}

// this is generated
impl NetworkBehaviour for FooBehaviour {
	type OutEvent = MyEvent;


	fn poll(&mut self, params: impl PollParameters) -> Poll<...> {
		
		self.my_poll(params)?;

		match self.inner.poll(params) {
			NetworkBehaviourAction::GenerateEvent(bar_event) => {
				self.inject_event(bar_event)
			}
			// other cases
		}
	}
}
```

Now what can you do with that?

1. If you want to process events as part of the `NetworkBehaviour`, use `event_process` true and you will be given each event in the event handler.
2. If don't want that (`event_process = false`), then the `poll` method cannot delegate to anything and will attempt to `return` your event from the `poll` function. For this to work, we need `impl From<BarEvent> for MyEvent`:

```rust
match self.inner.poll(params) {
	NetworkBehaviourAction::GenerateEvent(bar_event) => {
		return Poll::Ready(NetworkBehaviourAction::GenerateEvent(bar_event.into())) // `From` impl called here
	}
	// other cases
}
```
3. If you choose to use `poll_method = "my_poll`, you can return a `NetworkBehaviourAction` from there, which includes the `GenerateEvent` variant meaning it can make sense to have `OutEvent` set as well as `poll_method`. For example, prior to introducing `event_process = false`, you had to store events in `NetworkBehaviourEventProcess` in a `Vec` and then pop them off again in `my_poll` if you wanted to "bubble" them up.
4. If you really wanted to, you could do: `OutEvent = ()` (which is the default) and `event_process = false`. For this to work, you would need to implement `From<BarEvent> for ()` which is typically non-sensical but technically possible.

As you can see, these are all orthogonal and it just happens to be that the typical combination is:

- `OutEvent = MyEvent` + `event_process = false`: Allows you to process all events outside the behaviour
- `OutEvent = ()` + `event_process = true`: Allows you to process all events inside the behaviour

Should we document all these pathological cases? I don't know. I think the documentation that you have written here is really great and I think we should merge it.

@mxinden What do you think?

mxinden (approved):

mxinden:
@Frederik-Baetens mind resolving the merge conflict?

Frederik-Baetens:
Conflict resolved :)

Issue #2497, opened 2022-02-09
protocols/gossipsub: Custom interval grows and is not performant


AgeManning:
@mxinden - Your input here might also be helpful. 

After a bit of digging, we've found that the gossipsub heartbeat can be set to the order of 1s but grow to the order of many minutes. This heartbeat is crucial for gossipsub to maintain its peers and meshes as well as bound its memory cache and gossip messages on time. 

A many-minute heartbeat can cause terrible conditions on the network and significantly grow memory use of gossipsub users.

We've tracked the issue down to the custom implementation of `Interval` in https://github.com/libp2p/rust-libp2p/blob/master/protocols/gossipsub/src/interval.rs. 

As an example, we have three nodes running, in one (green one) I've replaced the custom `Interval` with a `tokio::time::Interval`. The heartbeat is supposed to be 700ms.

![allg](https://user-images.githubusercontent.com/7454587/153130760-b41a0426-7af1-401f-9506-58c75e67bdcf.png)

Halfway through these graphs I've switched the green node to `tokio::time::Interval`. The heartbeat interval (time between heartbeats) then stays steady on 700ms whereas the other nodes have grown and continue to do so. 
It's also useful to note that the heartbeat duration (time taken to do a heartbeat) itself is about 3-4x faster with `tokio::time::Interval` than with the custom implementation we are currently using.

After a very quick skim, I think the issue of the growing interval lies in this calcualtion: https://github.com/libp2p/rust-libp2p/blob/master/protocols/gossipsub/src/interval.rs#L86

Where if we miss a interval the `fires_at` and `delay` can potentially become out of sync. 

It is useful to note that the heartbeat interval grows in multiples of the desired interval. 

It would be nice to get the current implementation to work correctly and be as performant as the tokio version. Failing this, potentially we add a feature flag or something so that tokio users can benefit from the performance gains.

cc @divagant-martian @pawanjay176

divagant-martian:
Checked our metrics again now that some hours have passed and gains are consistent
![image](https://user-images.githubusercontent.com/26765164/153191578-d524ceb1-01e6-45f8-bae9-923117640728.png)
The step increases are what @AgeManning mentions as increases by multiples of the heartbeat duration

divagant-martian:
[...]

PR #2506, opened 2022-02-10
protocols/gossipsub: Revert back to wasm_timer for interval

AgeManning:
As described in https://github.com/libp2p/rust-libp2p/issues/2497 the current implementation for Interval in gossipsub can cause some serious issues with how gossipsub functions. 

The heartbeat interval can grow unbounded, which unbounds the memory cache, exploding RAM and causes messages to be gossiped much later than they should be which can lead to very old messages being re-propagated on networks. 

This PR reverts the Interval implementation back to `wasm_timer`. This should be sufficient to get gossipsub to work as expected however it was noted that this is not as performant as a straight `tokio::time::Interval`. A future PR should be made that ports the interval to be `tokio` agnostic, yet equally performant and compatible with WASM on NodeJS.


mxinden (commented):
Looks good to me. Thanks for reverting. Could you include a changelog entry as well?

Also I think this needs to use `wasm_timer::Instant` instead of `instant::Instant`.


//CC @wngr

divagant-martian:
> Also I think this needs to use `wasm_timer::Instant` instead of `instant::Instant`.

https://github.com/libp2p/rust-libp2p/pull/2245 made the `ProtocolsHandler` trait dependent on `instant`

divagant-martian:
I've removed support for gossipsub in the wasm targets. It's sad since before https://github.com/libp2p/rust-libp2p/pull/2245 there was support for the other two targets, but the `instant::Instant` vs `wasm_timer::Instant` incompatibility makes this the only reasonable solution that I can see for now.

divagant-martian:
Following @mxinden suggestion, Gossipsub should now be removed only in the `wasm32-unknown-unknown` target. I'm using `cargo tree` to verify this:
```bash
> cargo tree -e features --target=wasm32-unknown-unknown | rg gossipsub
# empty
> cargo tree -e features --target=wasm32-wasi | rg gossipsub
#├── libp2p-gossipsub feature "default"
#│   └── libp2p-gossipsub v0.36.0 (/home/me/rust-libp2p/protocols/gossipsub)
#│       ├── libp2p-gossipsub feature "default" (*) # libp2p-metrics dep
> cargo tree -e features --target=wasm32-unknown-emscripten | rg gossipsub
#├── libp2p-gossipsub feature "default"
#│   └── libp2p-gossipsub v0.36.0 (/home/me/rust-libp2p/protocols/gossipsub)
#│       ├── libp2p-gossipsub feature "default" (*) # libp2p-metrics dep
```

protocols/gossipsub/CHANGELOG.md
```diff
@@ -10,9 +10,12 @@
 
 - Merge NetworkBehaviour's inject_\* paired methods (see PR 2445).
 
+- Revert to wasm-timer (see PR 2506).
```
```suggestion
- Revert to wasm-timer (see [PR 2506]).
```

mxinden (approved):

mxinden:
Thanks @divagant-martian and @AgeManning!

Issue #2504, opened 2022-02-10
Pull `rw-stream-sink` into this `rust-libp2p`?


smoelius:
Should [`rw-stream-sink`](https://github.com/paritytech/rw-stream-sink) be brought into `rust-libp2p`'s workspace?

Doing so would help ensure that `rw-stream-sink` is regularly linted and tested.

The only potential downside I can see is: it could complicate how you publish `libp2p` and/or `rw-stream-sink`.

If this change is considered reasonable, I could submit a PR.

mxinden:
> Doing so would help ensure that rw-stream-sink is regularly linted and tested.

Has there been a recent incident that would warrant the move of `rw-stream-sink` specifically? Are you interested in moving `rw-stream-sink` only, or all smaller crates? I am having difficulties understanding the concrete motivation here.

smoelius:
> Has there been a recent incident ...

No, and I'm sorry for being cryptic.

`rw-stream-sink` uses an old version of [`pin-project`](https://github.com/taiki-e/pin-project) (^0.4.6). @hamamo opened an [issue](https://github.com/paritytech/rw-stream-sink/issues/2) about this on October 1, but the issue seems to have gone unnoticed. In short, it seems that `rw-stream-sink` is not getting the attention it deserves. Having `rw-stream-sink` use the same Cargo.lock file as the other crates in this repository should help with the dependency aspect of this problem. And there would likely be additional benefits, as I alluded to above.

Re other small crates, I think it depends on several factors, such as "who are the crate's consumers?" and "does the crate warrant independent development?" For `rw-streak-sink`, its primary consumer appears to be `libp2p`, and it doesn't seemed to have received any independent development. (It has just one commit.)

Again, sorry for appearing out of left field. I hope the question makes more sense now.

mxinden:
I am happy to maintain `rw-stream-sink`. @tomaka in case you think this is a good idea, i.e. moving `rw-stream-sink` into `rust-libp2p/misc/`, can you make me an owner on https://crates.io/crates/rw-stream-sink/?

mathiversen:
I'm not able to compile the repository locally and it seems to be because of this. 
![image](https://user-images.githubusercontent.com/22917480/161249832-86aacb35-63cf-4c21-bb3a-de8fcd8ec247.png)
![image](https://user-images.githubusercontent.com/22917480/161249598-beed03f7-f189-4f74-9951-3a8c9d93dff3.png)

[...]

PR #2507, opened 2022-02-11
Incorporate `rw-stream-sink`

smoelius:
If it helps, here's what #2504 might look like.


koushiro:
@mxinden I think this PR could be closed because you already did the same in PR #2641

PR #2516, opened 2022-02-14
Don't take ownership of key in SignedEnvelope::new (#2510)

### LINKED ISSUE #2510: core/src/signed_envelope: Don't take ownership of `key` in `SignedEnvelope::new`
This boils down to the patch below:

``` diff
diff --git a/core/src/signed_envelope.rs b/core/src/signed_envelope.rs
index 73efcc99..cdf8e9f3 100644
--- a/core/src/signed_envelope.rs
+++ b/core/src/signed_envelope.rs
@@ -19,7 +19,7 @@ pub struct SignedEnvelope {
 impl SignedEnvelope {
     /// Constructs a new [`SignedEnvelope`].
     pub fn new(
-        key: Keypair,
+        key: &Keypair,
         domain_separation: String,
         payload_type: Vec<u8>,
         payload: Vec<u8>,
```

laurentsenta:
Fixes issues #2510 

- [x] don't take ownership of key in SignedEnvelope::new
- [x] clarify if we want / should use a ref through all the code instead of making a clone
- [x] don't take ownership of key in PeerContrustor::new

### DIFF
--- a/core/CHANGELOG.md
+++ b/core/CHANGELOG.md
@@ -12,9 +12,12 @@
 
 - Validate PeerRecord signature matching peer ID. See [RUSTSEC-2022-0009].
 
+- Don't take ownership of key in `PeerRecord::new` and `SignedEnvelope::new`. See [PR 2516].
+
 [PR 2456]: https://github.com/libp2p/rust-libp2p/pull/2456
 [RUSTSEC-2022-0009]: https://rustsec.org/advisories/RUSTSEC-2022-0009.html
 [PR 2492]: https://github.com/libp2p/rust-libp2p/pull/2492
+[PR 2516]: https://github.com/libp2p/rust-libp2p/pull/2516
 
 # 0.31.0 [2022-01-27]
 
--- a/core/src/peer_record.rs
+++ b/core/src/peer_record.rs
@@ -59,7 +59,7 @@ impl PeerRecord {
     /// Construct a new [`PeerRecord`] by authenticating the provided addresses with the given key.
     ///
     /// This is the same key that is used for authenticating every libp2p connection of your application, i.e. what you use when setting up your [`crate::transport::Transport`].
-    pub fn new(key: Keypair, addresses: Vec<Multiaddr>) -> Result<Self, SigningError> {
+    pub fn new(key: &Keypair, addresses: Vec<Multiaddr>) -> Result<Self, SigningError> {
         use prost::Message;
 
         let seq = SystemTime::now()
@@ -88,7 +88,7 @@ impl PeerRecord {
         };
 
         let envelope = SignedEnvelope::new(
-            key,
+            &key,
             String::from(DOMAIN_SEP),
             PAYLOAD_TYPE.as_bytes().to_vec(),
             payload,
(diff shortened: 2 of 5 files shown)


mxinden:
>  clarify if we want / should use a ref through all the code instead of making a clone

I am not quite sure I follow. Could you expand on this @laurentsenta?

mxinden:
To reproduce CI testfailures locally you can run `cd core/ && cargo test`.

laurentsenta:
@mxinden
regarding:
> clarify if we want / should use a ref through all the code instead of making a clone

According to this [doc](https://github.com/libp2p/rust-libp2p/blob/master/core/src/peer_record.rs#L61), the keypair is the one used to authenticate every connection. But it looks like we [clone this key](https://github.com/libp2p/rust-libp2p/blob/master/protocols/rendezvous/src/client.rs#L237) when we want to use it.

Would that make sense to make the relationship explicit by using a reference in the `PeerRecord` too, instead of a clone?
So similar to borrowing the key in `SignedEnvelop`, should we also borrow the key when we instantiate a `PeerRecord`?

mxinden:
> > clarify if we want / should use a ref through all the code instead of making a clone
> 
> According to this [doc](https://github.com/libp2p/rust-libp2p/blob/master/core/src/peer_record.rs#L61), the keypair is the one used to authenticate every connection. But it looks like we [clone this key](https://github.com/libp2p/rust-libp2p/blob/master/protocols/rendezvous/src/client.rs#L237) when we want to use it.
> 
> Would that make sense to make the relationship explicit by using a reference in the `PeerRecord` too, instead of a clone? So similar to borrowing the key in `SignedEnvelop`, should we also borrow the key when we instantiate a `PeerRecord`?

Good callout @laurentsenta. Off the top of my head there is no reason why `PeerRecord` and `SignedRecord` should not take a reference only. Maybe @thomaseizinger has thoughts here?

thomaseizinger:
> > > clarify if we want / should use a ref through all the code instead of making a clone
> > 
> > According to this [doc](https://github.com/libp2p/rust-libp2p/blob/master/core/src/peer_record.rs#L61), the keypair is the one used to authenticate every connection. But it looks like we [clone this key](https://github.com/libp2p/rust-libp2p/blob/master/protocols/rendezvous/src/client.rs#L237) when we want to use it.
> > 
> > Would that make sense to make the relationship explicit by using a reference in the `PeerRecord` too, instead of a clone? So similar to borrowing the key in `SignedEnvelop`, should we also borrow the key when we instantiate a `PeerRecord`?
> 
> Good callout @laurentsenta. Off the top of my head there is no reason why `PeerRecord` and `SignedRecord` should not take a reference only. Maybe @thomaseizinger has thoughts here?

Yes, skimming the code it makes sense :)

laurentsenta:
Thanks for the feedbacks! I pushed another commit with the tweak,
The PR is code complete, cargo check & tests are passing on my machine

thomaseizinger (approved):

laurentsenta:
Thanks for the review, @thomaseizinger! could you merge, or should I ping Max? I don't have write access on the repo yet.

mxinden (commented):
Missing a changelog entry in `core/CHANGELOG.md`. Otherwise looks good to me.

laurentsenta:
@mxinden Fixed, thanks for the check!

mxinden (approved):
Thanks for the help @laurentsenta.

mxinden:
First patch merged @laurentsenta  :tada:

PR #2517, opened 2022-02-14
swarm/: Report aborted connections

### LINKED ISSUE #2329: core/: Report pending connection being aborted

mxinden:
- Fixes https://github.com/libp2p/rust-libp2p/issues/2329
- Replaces https://github.com/libp2p/rust-libp2p/pull/2369.
- Resolves merge conflicts with https://github.com/libp2p/rust-libp2p/pull/2492
- Fixes `Swarm::disconnect` to disconnect pending connections https://github.com/libp2p/rust-libp2p/commit/ff4c071451c2e2119eaf7a9ea05c7ae73290b032


thomaseizinger (approved):
Thanks @mxinden for resolving the conflicts, this is a nice patch that improves correctness :)

PR #2519, opened 2022-02-15
swarm/src: Remove ConnectionHandler

mxinden:
The `ConnectionHandler` trait is not exposed to users. The only
implementor of `ConnectionHandler` is `NodeHandlerWrapper`. Thus
`ConnectionHandler` is a superfluous abstraction. This commit removes
`ConnectionHandler`.

Next to this large change, this commit removes the `Tmuxer` trait
parameter. `Swarm` enforces dynamic dispatching via `StreamMuxerBox`
anyways, thus the trait parameter is useless.

As a follow up to this commit one could rename `ProtocolsHandler` to
`ConnectionHandler` and `NodeHandlerWrapper` to
`ConnectionHandlerWrapper` or just `Wrapper`.

Follow up to #2492.


swarm/src/connection.rs
```diff
@@ -54,28 +56,26 @@ pub struct Connected {
...
-    TMuxer: StreamMuxer,
-    THandler: ConnectionHandler<Substream = Substream<TMuxer>>,
+    THandler: ProtocolsHandler,
 {
     /// Node that handles the muxing.
-    muxing: substream::Muxing<TMuxer, THandler::OutboundOpenInfo>,
+    muxing: substream::Muxing<StreamMuxerBox, NodeHandlerWrapperOutboundOpenInfo<THandler>>,
     /// Handler that processes substreams.
-    handler: THandler,
+    handler: NodeHandlerWrapper<THandler>,
```
thomaseizinger, line 73: Could we inline this struct into `Connection` now? Then we could just get rid of `NodeHandlerWrapper` completely!

I think the only reason we are passing this around at the moment is because we are constructing it "too early". We could just pass `THandler` around everywhere have all other fields of `NodeHandlerWrapper` be in `Connection`.

thomaseizinger (approved):
Nice work! I love how the layers are disappearing :)

swarm/src/connection.rs
mxinden: > I think the only reason we are passing this around at the moment is because we are constructing it "too early".

I think you are confusing the concept of `IntoProtocolsHandler` -> `ProtocolsHandler`, `NodeHandlerWrapperBuilder` -> `NodeHandlerWrapper` and now obsolete `IntoConnectionHandler` -> `ConnectionHandler` with the relation between `ProtocolsHandler` and `NodeHandlerWrapper`, right?

(Puuuh, so many `XXXHandler`. I hope we end up with only `ProtocolsHandler` (renamed to `ConnectionHandler`) once all the refactorings are in.)

> Could we inline this struct into `Connection` now?

I think so. At least move the two closer together. I suggest we rename `ProtocolsHandler` to `ConnectionHandler` first ( including some smaller clean-ups) and then consider folding `NodeHandlerWrapper` into `Connection`.

mxinden:
> I love how the layers are disappearing :)

Same here :)

swarm/src/connection.rs
thomaseizinger: I may confuse some things yes 😁

Either order is fine with me, it was a more general remark, not suggesting it should be done next or at all! :)

PR #2522, opened 2022-02-16
Return signing key in SignedEnvelope.payload

### LINKED ISSUE #2511: core/src/signed_envelope: Enforce verification of public key on `SignedEnvelope::payload`
`PeerRecord::from_signed_envelope` would previously (before https://github.com/libp2p/rust-libp2p/pull/2491) not validate that the public key of the `SignedEnvelope` matches the `PeerId` of the `PeerRecord`.

Today `PeerRecord` directly accesses `SignedEnvelope::key` as a quick fix. Long term we should enforce verification through the type system. Ideas thus far:

- @MarcoPolo: Return both the payload and the public key on `SignedEnvelope::payload`.
- @thomaseizinger: Require passing a close to `SignedEnvelope::payload` validating the public key.

Any additional suggestions? Thoughts? Volunteers?

MarcoPolo:
## Context

Fixes https://github.com/libp2p/rust-libp2p/issues/2511.

Another solution would be to pass in a closure as a key validation function to `payload`. I originally tried this, but quickly noticed that it made the `PeerRecord` use case awkward. This is because the peer id we are validating against is inside the payload. So the closure would need access to the payload. If you passed in the payload to the closure, then you would need to parse the payload twice (or awkwardly mutate an outer variable).

So instead, I return the signing key from the `payload` method.

If I'm misunderstanding something @thomaseizinger please let me know.


core/src/signed_envelope.rs
```diff
@@ -44,15 +44,19 @@ impl SignedEnvelope {
...
 
-    /// Extract the payload of this [`SignedEnvelope`].
+    /// Extract the payload and signing key of this [`SignedEnvelope`].
     ///
     /// You must provide the correct domain-separation string and expected payload type in order to get the payload.
     /// This guards against accidental mis-use of the payload where the signature was created for a different purpose or payload type.
+    ///
+    /// It is the caller's responsibility to check that the signing key is what
+    /// is expected. For example, checking that the signing key is from a
+    /// certain peer.
```
mxinden, line 54: How about enforcing this constraint via `#[must_use]`? Something along the lines of the playground below:

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=1a32dedc2d59c3298dc73a8b48d9d76c

mxinden (commented):
Thanks for picking this up! :pray:

core/src/signed_envelope.rs
MarcoPolo: Oh excellent! I was wondering if there was a way I could leverage `#[must_use]`. Thanks!

mxinden: As far as I can tell, unfortunately one can not do it on an anonymous tuple. E.g. instead of the additional `struct` I would prefer to do:

``` rust
fn foo() -> Result<(a, #[must_use] b), c> {}
```

MarcoPolo: Yeah that was my first thought, but I hadn't considering using a struct instead.

mxinden: Sorry, the above suggestion actually does not work. The compiler reports the field not being used in general, rather than not being used in this particular case.

Also nesting `#[must_use]` structs doesn't work see https://github.com/rust-lang/rust/issues/39524.

Also @MarcoPolo pointed out that users would likely do something along the lines of `let (payload, _) = payload().handle_error();` which would thus circumvent the `#[must_use]`.

Sorry for the noise here.

core/src/signed_envelope.rs
```diff
@@ -44,15 +44,19 @@ impl SignedEnvelope {
...
     pub fn payload(
```
mxinden:
```suggestion
    pub fn payload_and_signing_key(
```

What do you think of renaming the method as well? If I recall correctly you suggested this in your initial vulnerability report, right?

MarcoPolo: Changed in https://github.com/libp2p/rust-libp2p/pull/2522/commits/ef3afcdeab1b752a52063e3aa3c2be4b356c1474

core/src/signed_envelope.rs
thomaseizinger: We could do something like this: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=92e6fc6cf971873eb590d9ada72f85fa

But I am not sure if it is useful. Assigning `_` will silence the `must_use` warning ...

thomaseizinger (approved):
Thanks for tackling this!

mxinden:
I suggest we move forward and merge here. @MarcoPolo given that this is a breaking change, would you mind adding a changelog entry to `core/CHANGELOG.md`?

core/src/signed_envelope.rs
MarcoPolo: Unfortunately assigning the return value to anything and not using the key value will also satisfy the must_use lint. e.g. `let data = foo().unwrap();`

mxinden (approved):
:pray:

PR #2526, opened 2022-02-17
Optionally use only global IPs for AutoNAT

### LINKED ISSUE #2514: AutoNAT: how to handle non-global IP addresses?
Currently, AutoNAT seems to consider private IP addresses as possible external addresses when two peers run in the same private network. In my case, this is one computer where I run multiple instances of my code, but it is conceivable that the same happens when separate machines run in the same private network (i.e. behind the same router or within some corporate network).

I don't really know which behavior would be most appropriate. My initial guess is that non-public IP-Adresses (i.e. those for which `Ipv4Addr::is_global()` or `Ipv6Addr::is_global()` returns `false`) should not be considered as candidates for probes, but I don't know yet whether there are use cases where private IP addresses are valid external addresses (is multi-layered NAT a thing?).

If people see this as a valid concern, it should probably be communicated to the libp2p AutoNAT spec team for inclusion in the spec.

hamamo:
Fixes #2514.

Beware, this is my first attempt at a pull request. And my first attempt to contribute Rust code to a public project...

This change introduces a config option for AutoNAT that forces it to only consider globally routable IP addresses for NAT probing. The default behavior is unchanged (all addresses are considered), and the restriction to globally routable addresses can be switched on using `use_only_global_ips` option.

This branch is based on v0.42 as that is the libp2p version currently in crates.io. It requires a crate level attribute `#![feature(ip)]` to use the `is_global()` methods of Ipv4Addr and Ipv6Addr.

### DIFF
--- a/protocols/autonat/src/behaviour/as_client.rs
+++ b/protocols/autonat/src/behaviour/as_client.rs
@@ -21,8 +21,8 @@
 use crate::ResponseError;
 
 use super::{
-    Action, AutoNatCodec, Config, DialRequest, DialResponse, Event, HandleInnerEvent, NatStatus,
-    ProbeId,
+    Action, AutoNatCodec, Config, DialRequest, DialResponse, Event, GlobalAddress,
+    HandleInnerEvent, NatStatus, ProbeId,
 };
 use futures::FutureExt;
 use futures_timer::Delay;
@@ -201,6 +201,12 @@ impl<'a> AsClient<'a> {
 
                 let mut addresses: Vec<_> = params.external_addresses().map(|r| r.addr).collect();
                 addresses.extend(params.listened_addresses());
+                if self.config.use_only_global_ips {
+                    addresses = addresses
+                        .into_iter()
+                        .filter(|a| a.contains_global_address())
+                        .collect::<Vec<_>>();
+                };
 
                 let probe_id = self.probe_id.next();
                 let event = match self.do_probe(probe_id, addresses) {
--- a/protocols/autonat/src/behaviour/as_server.rs
+++ b/protocols/autonat/src/behaviour/as_server.rs
@@ -19,8 +19,8 @@
 // DEALINGS IN THE SOFTWARE.
 
 use super::{
-    Action, AutoNatCodec, Config, DialRequest, DialResponse, Event, HandleInnerEvent, ProbeId,
-    ResponseError,
+    Action, AutoNatCodec, Config, DialRequest, DialResponse, Event, GlobalAddress,
+    HandleInnerEvent, ProbeId, ResponseError,
 };
 use instant::Instant;
 use libp2p_core::{connection::ConnectionId, multiaddr::Protocol, Multiaddr, PeerId};
@@ -114,12 +114,21 @@ impl<'a> HandleInnerEvent for AsServer<'a> {
             } => {
                 let probe_id = self.probe_id.next();
                 match self.resolve_inbound_request(peer, request) {
-                    Ok(addrs) => {
+                    Ok(mut addrs) => {
                         log::debug!(
                             "Inbound dial request from Peer {} with dial-back addresses {:?}.",
                             peer,
                             addrs
                         );
+
+                        if self.config.use_only_global_ips {
+                            addrs = addrs
+                                .into_iter()
+                                .filter(|a| {
+                                    !self.config.use_only_global_ips || a.contains_global_address()
+                                })
+                                .collect();
+                        }
 
                         self.ongoing_inbound
                             .insert(peer, (probe_id, request_id, addrs.clone(), channel));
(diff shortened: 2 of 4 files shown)


hamamo:
Oops, just now I see I left some redundant checks in the code. (as_server.rs line 128)

hamamo:
While using this in my mixed environment, AutoNAT seems to switch from the correct Public() state to Private, perhaps when the random choice of peers selects the wrong one. Will need to investigate further, so this isn't production-ready yet.

elenaf9:
> While using this in my mixed environment, AutoNAT seems to switch from the correct Public() state to Private, perhaps when the random choice of peers selects the wrong one. Will need to investigate further, so this isn't production-ready yet.

I'll do a proper review later, but I think what is missing right now is that also from the client side only servers are used that are not in the same private network. Because otherwise the client sends a dial-back request to the server in the same network, and the server filters all the addresses from the client because they are not public and then reports back a dial failure, which results in the Private NAT Status.
The information about the observed address from connected peers is already there, so you just would have to use that information in the `as_client` module when selecting the server.

mxinden:
> This branch is based on v0.42 as that is the libp2p version currently in crates.io. It requires a crate level attribute `#![feature(ip)]` to use the `is_global()` methods of Ipv4Addr and Ipv6Addr.

rust-libp2p does not depend on any night features today. If I am not mistaken using nightly in rust-libp2p would require all applications build on top of rust-libp2p to also be build with nightly. I don't think we should force that upon our users.

Instead I suggest simply copying the `is_global` method from the standard library into the rust-libp2p code-base, giving credit to the rust std authors. Once it stabilized, we can use the standard library method directly.

Does that make sense @hamamo?

hamamo:
> rust-libp2p does not depend on any night features today. If I am not mistaken using nightly in rust-libp2p would require all applications build on top of rust-libp2p to also be build with nightly. I don't think we should force that upon our users.
> 
> Instead I suggest simply copying the `is_global` method from the standard library into the rust-libp2p code-base, giving credit to the rust std authors. Once it stabilized, we can use the standard library method directly.
> 
> Does that make sense @hamamo?

Yes, absolutely, I'm just unsure where to put it then, probably side by side to the function that uses it.

mxinden:
> I'm just unsure where to put it then, probably side by side to the function that uses it.

I don't have an opinion here. Suggestion sounds good.

protocols/autonat/src/behaviour/as_client.rs
```diff
@@ -201,6 +201,12 @@ impl<'a> AsClient<'a> {
 
                 let mut addresses: Vec<_> = params.external_addresses().map(|r| r.addr).collect();
                 addresses.extend(params.listened_addresses());
+                if self.config.use_only_global_ips {
+                    addresses = addresses
+                        .into_iter()
+                        .filter(|a| a.contains_global_address())
+                        .collect::<Vec<_>>();
+                };
```
elenaf9, line 209: When a client sends a dial-back request to the server, the server replaces the IP in the sent addresses with the IP at which it observes the client. So effectively, the server never tests an IP the client sends, but instead only the ports of each address. Therefore it is not necessary to filter the addresses here. On the contrary: if we filter them here we remove all of the `params.listened_addresses()`, since those are always private.

protocols/autonat/src/behaviour/as_server.rs
```diff
@@ -114,13 +114,22 @@ impl<'a> HandleInnerEvent for AsServer<'a> {
...
 
+                        if self.config.use_only_global_ips {
+                            addrs = addrs
+                                .into_iter()
+                                .filter(|a| {
+                                    !self.config.use_only_global_ips || a.contains_global_address()
+                                })
+                                .collect();
+                        }
+
```
elenaf9, line 132: I think this logic should be moved into the `AsServer::resolve_inbound_request` method. If the client is in the same private network (= the observed IP is private), we want to directly return an `Error` and not do a dial-attempt.

elenaf9 (commented):
Thank you very much for the PR @hamamo , it is very much appreciated!

One comment to help the general understanding:
For DOS attack prevention (https://github.com/libp2p/specs/pull/369) the server does not dial the exact addresses that the client sends in the dial-request. 
Instead, the `AutoNAT` behaviour remembers for each connected peer the address at which it observes the peer (= address from the endpoint on `Behaviour::inject_connection_established`).
On an inbound dial-back request, it then iterates through all the addresses of the request, and for each address it replaces the address's IP with the client's observed on.  https://github.com/libp2p/rust-libp2p/blob/2ff9acee227bc8787fb41d657de5f174250e2588/protocols/autonat/src/behaviour/as_server.rs#L323-L338
Therefore, the "global-ips-only" restriction does not apply individually for each address in the dial-request and they don't need to be filtered. Instead, the observed address has to be checked and based on that the whole dial-request is accepted, or not. 
Does that explanation make sense @hamamo?

elenaf9:
Friendly ping @hamamo. Are you still planning to do this fix? I am happy to help in case that you need more input.

elenaf9:
Fixed via #2618, thus I am closing here.

PR #2527, opened 2022-02-18
swarm/: Rename ProtocolsHandler to ConnectionHandler

mxinden:
A `ProtocolsHandler`, now `ConnectionHandler`, handels a connection, not
a protocol. Thus the name `CONNECTIONHandler` is more appropriate.

Next to the rename of `ProtocolsHandler` this commit renames the `mod
protocols_handler` to `mod handler`. Finally all combinators (e.g.
`ProtocolsHandlerSelect`) are renamed appropriately.

---

I find the name `ConnectionHandler` more intuitive, both for me familiar with the library, as well as for newcomers while explaining the concepts of the library. What do folks think?


thomaseizinger (approved):
I think it is a good change but it will take some getting used to.

I wonder if there is a way of making this change in a more graceful way. I guess users can just start with a search-replace and fix up remaining errors. Should be  easy enough!

mxinden:
> I guess users can just start with a search-replace and fix up remaining errors. Should be easy enough!

:+1: I think an additional call-out in the changelog should be good enough.

PR #2528, opened 2022-02-19
Misc metrics tide to hyper

esotericbyte:
Metrics example server based on hyper replaces a component based on Tide framework. This removes dependency on Tide which triggered audit warnings.


esotericbyte:
I could probably clean up unnecessary commits.

mxinden:
Thanks @esotericbyte. I will take a look.

> I could probably clean up unnecessary commits.

No need. We will _squash merge_ anyways.

misc/metrics/Cargo.toml
```diff
@@ -32,7 +32,9 @@ prometheus-client = "0.15.0"
 libp2p-gossipsub =  { version = "0.36.0", path = "../../protocols/gossipsub", optional = true }
 
 [dev-dependencies]
+log = "0.4.0"
 env_logger = "0.8.1"
 futures = "0.3.1"
 libp2p = { path = "../../", features = ["metrics"] }
-tide = "0.16"
+hyper = { version="0.14", features = ["full"] }
+tokio = { version = "1", features = ["full"] }
```
mxinden: For the sake of reducing CI compile times, do we need `full` for each of these here?

misc/metrics/examples/metrics_http_service.rs
```diff
@@ -0,0 +1,127 @@
...
+use std::future::Future;
+use std::pin::Pin;
+use std::task::{Context, Poll};
+use log::{error,info};
+use hyper::service::Service;
+use hyper::{Body, Request, Response, Server};
+use hyper::http::{StatusCode};
+
+pub async fn metrics_server(registry: Registry) -> Result<(), std::io::Error> {
+    //serve on localhost
```
mxinden:
```suggestion
    // Serve on localhost.
```

misc/metrics/examples/metrics_http_service.rs
```diff
@@ -0,0 +1,127 @@
...
+    let addr = ([127, 0, 0, 1], 0).into();
+    
+    // Use the tokio runtime to run the hyper server
```
mxinden:
```suggestion
    // Use the tokio runtime to run the hyper server.
```

misc/metrics/examples/metrics_http_service.rs
```diff
@@ -0,0 +1,127 @@
...
+
+    fn poll_ready(&mut self, _cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+        Poll::Ready(Ok(()))
+    }
+
+    fn call(&mut self, req: Request<Body>) -> Self::Future {
+        let resp = 
+        if (req.method() == & hyper::Method::GET) &&
+            (req.uri().path() == "/metrics") {
+            //encode and serve meterics from registry
```
mxinden:
```suggestion
            // Encode and serve metrics from registry.
```

misc/metrics/examples/metrics_http_service.rs
```diff
@@ -0,0 +1,127 @@
+// Copyright 2022 Protocol Labs.
```
mxinden: `cargo` interprets this as a full example, how about doing the file structure below to have a single example only?

```
misc/metrics/examples/metrics/main.rs
misc/metrics/examples/metrics/http_service.rs
```

mxinden (commented):
Couple of small comments. Looks good to me overall. Thanks for the help @esotericbyte!

misc/metrics/Cargo.toml
esotericbyte: The problem was related to structs a couple of levels deep with missing traits implementations. I will do some research and hacking to see if the features can be narrowed.

esotericbyte: Cargo.toml is fixed and I just rebased.

mxinden:
@esotericbyte can you run `cargo fmt` on the code?

esotericbyte:
I ran rustfmt. I had to run it with --config  edition="2021" even though it's set in Cargo.toml.
The first time around I had set max_width=80 which is why it did not pass! Sorry! 
I fixed a one line monster that was unsatisfactory by adding some intermediate variables that I would expect to be optimized out anyway and are useful if anyone adds another route to the server.  I re-based and tested before I pushed.

mxinden (approved):
Thanks @esotericbyte!

PR #2532, opened 2022-02-22
swarm/src/lib: Improve connection counting for test_behaviour_disconnect_all

mxinden:
`test_behaivour_disconnect_all` showed to be flaky on CI in the past. This pull request includes a couple of smaller fixes.


.github/workflows/ci.yml
```diff
@@ -14,8 +14,8 @@ jobs:
       matrix:
         args: [
           "--no-default-features",
-          "--all-features",
-          "--benches --all-features",
+          # "--all-features",
```
```suggestion
          "--all-features",
```

swarm/src/lib.rs
```diff
@@ -1583,27 +1583,36 @@ mod tests {
         TBehaviour: NetworkBehaviour,
         <<TBehaviour::ConnectionHandler as IntoConnectionHandler>::Handler as ConnectionHandler>::OutEvent: Clone,
     {
-        [swarm1, swarm2]
-            .iter()
-            .all(|s| s.behaviour.inject_connection_established.len() == num_connections)
```
mxinden, line 1588: Counting the `inject_connection_established` calls only might mislead in case there are more than 2 swarms interconnecting. Also some of these connections might already be closed.

swarm/src/lib.rs
```diff
@@ -1850,18 +1859,17 @@ mod tests {
                             },
                         );
                         state = State::Disconnecting;
+                        continue;
```
mxinden, line 1862: Just above we provided a new event to be emitted by the behaviour. `continue` to ensure the behaviour is polled again.

thomaseizinger (approved):
Thanks! LGTM :)

Issue #2509, opened 2022-02-11
*: Replace implementations of `Display` and `Error` on errors with `thiserror`


mxinden:
I would argue that a generated implementation of `Display` and `Error` via `thiserror` is faster to write and easier to maintain. Thus I advocate for using `thiserror` for most of not all libp2p error types.

demfabris:
I got this

dignifiedquire:
This means adding another error related dependency, that has no `no_std` support. :/ 

Ref https://github.com/dtolnay/thiserror/pull/64

demfabris:
@mxinden thoughts?

mxinden:
> This means adding another error related dependency, that has no `no_std` support. :/

We already depend on `thiserror` in most `libp2p-xxx` crates, e.g. `libp2p-swarm` and `libp2p-core`. In addition, in case we ever decide to be `no_std` compatible, I don't think it is particularly hard to revert this change, in the sense that it would not be particularly interwoven with other functionality.

I don't have much experience with `no_std`. With the discussions in https://github.com/libp2p/rust-libp2p/issues/627 in mind, I wonder whether this is a goal we can achieve in the first place. @dignifiedquire do you think `no_std` is a goal worth striving for?

dignifiedquire:
if we already depend on it, then nvm in regards to this specific instance

mxinden:
@demfabris still want to tackle this one? :innocent:

demfabris:
Yep @mxinden 👍🏻

thomaseizinger:
I'll close this as stale. I don't think it worth having an issue around to track this.

PR #2533, opened 2022-02-24
Don't print inner error when returning as source

thomaseizinger:
According to https://github.com/rust-lang/project-error-handling/issues/44#issuecomment-858151860, we should not be printing the inner error AND returning it as source. Libraries like `anyhow` will traverse the chain of `source` errors and build up a composed error message. Printing it and returning the same error from `source` results in duplicate error messages in that case.


mxinden:
Thanks for the pointer. I wasn't aware of this convention.

How about taking this one step further and deriving `Error` and `Display` for `TransportError` via `thiserror`?

thomaseizinger:
> Thanks for the pointer. I wasn't aware of this convention.
> 
> How about taking this one step further and deriving `Error` and `Display` for `TransportError` via `thiserror`?

I support the idea (there is a ticket for it I think?). Can we still merge this fix and we do the thiserror migration later? :)

mxinden (approved):

mxinden:
> (there is a ticket for it I think?)

https://github.com/libp2p/rust-libp2p/issues/2509

PR #2535, opened 2022-02-24
swarm/src/behaviour: Remove Send bound from NetworkBehaviour

dignifiedquire:
Needs confirmation this is okay design wise, but asfaict this is not actuallly required for the core of libp2p to work. 

cc @mxinden


mxinden:
I am considering this a non-breaking change, thus the patch version bump only. Please comment in case you think otherwise.

thomaseizinger:
Technically it is breaking.

People could rely on the implicit Send bound if they write code that it is abstract over NetworkBehaviour and sent across thread boundaries.

I think that can fairly easily happen if someone builds a generic piece of code for spawning a Swarm into a task pool as we can see from the tests.

I think publishing it as breaking change would be better. I don't know if any _library_ code exists that does depend on this but if it does, having things break in a dependency is a pretty shitty experience.

dignifiedquire:
agree with @thomaseizinger , it will likely require some folks to change their trait bounds if they are generic over it.

CHANGELOG.md
```diff
@@ -1,3 +1,4 @@
+
```
```suggestion
```

mxinden (approved):
Thanks @dignifiedquire for the find.

PR #2540, opened 2022-02-25
*: Apply clippy suggestions

mxinden:
Current `master` fails on clippy stage, see https://github.com/libp2p/rust-libp2p/runs/5338775279?check_suite_focus=true

Apart from the change in `libp2p-swarm` I don't think any of these contain a breaking change. Changelog entries for the breaking change in `libp2p-swarm` are added through https://github.com/libp2p/rust-libp2p/pull/2535.


thomaseizinger (approved):
Nice!

To catch these things earlier, we could contemplate running beta clippy in CI but not require it passing for merging!

mxinden:
> To catch these things earlier, we could contemplate running beta clippy in CI but not require it passing for merging!

Happy to accept it, though I don't think these "breaking" changes have a large impact on our development flow.

PR #2551, opened 2022-03-07
Add `PeerId::try_from_multiaddr`

thomaseizinger:
Little convenience method I needed in one of my projects. Thoughts?


rkuhn:
Yes, this looks very useful — in fact, I have such a helper in nearly every crate using libp2p. Companions are:

- `normalize_addr<'a>(addr: &<'a>Multiaddr, peer_id: &PeerId) -> Cow<'a, Multiaddr>` (name can be different, like “addr_with_peer_id” etc.)
- `addr_without_peer(addr: &Multiaddr) -> Cow<'_, Multiaddr>`

I’ve wished more than once that the presence or absence of a trailing `Protocol::P2p(_)` be captured in the type …

mxinden:
> I’ve wished more than once that the presence or absence of a trailing Protocol::P2p(_) be captured in the type …

:+1:

core/src/peer_id.rs
```diff
@@ -88,6 +89,17 @@ impl PeerId {
         }
     }
 
+    /// Tries to extract a [`PeerId`] from the given [`Multiaddr`].
+    ///
+    /// In case the given [`Multiaddr`] ends with `/p2p/<peer-id>`, this function
+    /// will return the encapsulated [`PeerId`], otherwise it will return `None`.
+    pub fn from_multiaddr(address: &Multiaddr) -> Option<PeerId> {
```
```suggestion
    pub fn try_from_multiaddr(address: &Multiaddr) -> Option<PeerId> {
```

I wonder whether this would be more in line with the Rust naming conventions. What do you think?

mxinden (commented):
Why not implement `TryFrom<&Multiaddr> for PeerId`?

rkuhn:
Implementing `TryFrom` would only feel right to me if there is exactly one way to possibly extract a PeerId from a Multiaddr. Syntactically the `P2p` protocol may appear in other places than at the end, right?

Which reminds me: perhaps the name should hint at the applied algorithm, like `PeerId::from_multiaddr_suffix`.

thomaseizinger:
> Implementing `TryFrom` would only feel right to me if there is exactly one way to possibly extract a PeerId from a Multiaddr. Syntactically the `P2p` protocol may appear in other places than at the end, right?

This is a good point. Personally, I also tend to not implement traits like `TryFrom` unless I also have a need to abstract over it for this particular case. I don't feel strongly about that argument but I do think @rkuhn's point of there not being a canonical way of extracting a PeerId from a Multiaddr is a show-stopper for providing a `TryFrom` impl.

core/src/peer_id.rs
thomaseizinger: Implemented!

core/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# Unreleased
```
```suggestion
# 0.32.1 [unreleased]
```

Mind bumping the version in `core/Cargo.toml` and the `libp2p-core` version in `./Cargo.toml`?

mxinden (commented):
Looks good to me otherwise.

core/CHANGELOG.md
thomaseizinger: > and the libp2p-core version in ./Cargo.toml?

Shouldn't a patch release get picked up automatically? I can still bump it though :)

mxinden: Yes. Though, given that we have so many subcrates, it is hard to keep track of which one has a pending release. Bumping the version in `libp2p` makes `cargo publish` on `libp2p` fail unless all pending subcrate releases are published, thus in this case reminding me to release `libp2p-core`.

Not ideal. Happy for alternative suggestions.

mxinden (approved):
:pray:

PR #2554, opened 2022-03-11
Revert ".github/workflow: Use ubuntu-18.04 fixing missing protoc binary "

mxinden:
Reverts libp2p/rust-libp2p#2368


Jerrody:
I think this failure can be fixed via installing LLVM:
`bash -c "$(wget -O - https://apt.llvm.org/llvm.sh)"`.

PR #2563, opened 2022-03-15
Correcting commands to use ip4 instead of dns4

MitaliBo:
For listening and dialing clients , tutorial uses /dns4 with ip_address instead of /ip4 for multiaddr


MitaliBo:
@mxinden

mxinden (approved):
My bad. Thanks for the fix @MitaliBo! I was using a relay server behind a DNS name to test this.

PR #2574, opened 2022-03-18
*: Import `libp2p` with `default-features = false`

mxinden:
While not a win in most cases, it reduces compile time for tests of
individual crates.


thomaseizinger (approved):
Concept ACK

Eventually, I'd like to find time for flipping this around so we don't have to deactivate default features!

mxinden:
> Eventually, I'd like to find time for flipping this around so we don't have to deactivate default features!

:+1: I would be in favor of that @thomaseizinger!

PR #2578, opened 2022-03-19
Rework `gossipsub-chat` example.

Jerrody:
And I made little refactoring of `Cargo.toml`.

TODO:
- [ ] Change instruction for the tutorial.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -167,7 +167,7 @@ required-features = ["request-response"]
 
 [[example]]
 name = "gossipsub-chat"
-required-features = ["gossipsub", "tcp-tokio", "mdns"]
+required-features = ["gossipsub", "tcp-tokio", "mdns", "dns-tokio"]
 
 [[example]]
 name = "ipfs-private"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -57,12 +57,7 @@ rendezvous = ["libp2p-rendezvous"]
 tcp-async-io = ["libp2p-tcp", "libp2p-tcp/async-io"]
 tcp-tokio = ["libp2p-tcp", "libp2p-tcp/tokio"]
 uds = ["libp2p-uds"]
-wasm-bindgen = [
-    "futures-timer/wasm-bindgen",
-    "instant/wasm-bindgen",
-    "getrandom/js",
-    "rand/wasm-bindgen",
-]
+wasm-bindgen = ["futures-timer/wasm-bindgen", "instant/wasm-bindgen", "getrandom/js", "rand/wasm-bindgen"]
 wasm-ext = ["libp2p-wasm-ext"]
 wasm-ext-websocket = ["wasm-ext", "libp2p-wasm-ext/websocket"]
 websocket = ["libp2p-websocket"]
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -7,7 +7,7 @@ version = "0.44.0"
 authors = ["Parity Technologies <admin@parity.io>"]
 license = "MIT"
 repository = "https://github.com/libp2p/rust-libp2p"
-keywords = ["peer-to-peer", "libp2p", "networking"]
+keywords = ["peer-to-peer", "libp2p", "networking", "p2p"]
```
mxinden: For the sake of a clean git history, would you mind doing these unrelated changes in separate pull requests?

examples/gossipsub-chat.rs
```diff
@@ -68,12 +139,25 @@ async fn main() -> Result<(), Box<dyn Error>> {
     let local_peer_id = PeerId::from(local_key.public());
     println!("Local peer id: {:?}", local_peer_id);
 
+    let noise_keys = libp2p::noise::Keypair::<libp2p::noise::X25519Spec>::new()
+        .into_authentic(&local_key)
+        .expect("Signing libp2p-noise static DH keypair failed.");
+
     // Set up an encrypted TCP Transport over the Mplex and Yamux protocols
-    let transport = libp2p::development_transport(local_key.clone()).await?;
+    let transport = libp2p::tcp::TokioTcpConfig::new()
```
mxinden: How about using the `tokio_development_transport` function instead?

examples/gossipsub-chat.rs
```diff
@@ -45,21 +45,92 @@
 //! ```
 //!
 //! The two nodes should then connect.
-
-use async_std::io;
```
mxinden, line 49: Why the change to tokio?

mxinden (commented):
I don't quite understand the goal of this refactoring. What are you trying to showcase?

Cargo.toml
Jerrody: Alright then!

examples/gossipsub-chat.rs
Jerrody: Hm, maybe it is better than my variant. I just don't understand what's differences between `development_transport` (`tokio_development_transport`) and `TcpConfig` (or `TokioTcpConfig`).

examples/gossipsub-chat.rs
Jerrody: Because this is a standard in Rust's world. Everybody uses `tokio` and this is the fact. `async-std` lost competition with `tokio` and `tokio` turned out to be more viable.
I accept another point of view, so, I want to see your position, pretty good to have many opinions.

Jerrody:
So, I made changes with `Cargo.toml`.
If my PR will be accepted I need to change instructions for the usage in the example's code.
And maybe make more comments on what things do.

examples/gossipsub-chat.rs
mxinden: The former works with async-io / async-std, the latter with Tokio.

Jerrody: > The former works with async-io / async-std, the latter with Tokio.

Alright, let's change to `tokio_development_transport`!

Jerrody:
> I don't quite understand the goal of this refactoring. What are you trying to showcase?

Sorry, I didn't see this message. The main reason is that for the current example we need to pass explicitly in args the address for connecting to the node. In my "refactored" example don't need to specify in the command line the address of the node, the connection is automatic.

Jerrody:
@mxinden So, I wait for your word mate. What should I change else?

thomaseizinger:
This is obsolete with #2996.

Issue #2546, opened 2022-03-03
File sharing example cannot support the sample plot situation


YoshieraHuang:
Hi, I am new to libp2p but very interested for this. While surfing through examples, I find the provided `file-sharing` example are not so ideal. I run these three scripts at the base directory of the repository in different console to test the sample plot.

In first console:
```
 cargo run --example file-sharing -- --listen-address /ip4/127.0.0.1/tcp/40837 --secret-key-seed 1 provide --path ./Cargo.lock --name lock
```
for serving `Cargo.lock` file with name `lock`. It works. The listening multiaddr is `/ip4/127.0.0.1/tcp/40837/p2p/12D3KooWPjceQrSwdWXPyLLeABRXmuqt69Rg3sBYbU1Nft9HyQ6X`

In second console:
```
cargo run --example file-sharing -- --listen-address /ip4/127.0.0.1/tcp/40800 --peer /ip4/127.0.0.1/tcp/40837/p2p/12D3KooWPjceQrSwdWXPyLLeABRXmuqt69Rg3sBYbU1Nft9HyQ6X --secret-key-seed 2 provide --path ./Cargo.toml --name toml
```
for serving `Cargo.toml` file with name `toml` and connect to the first console. It also works. The listening multiaddr is `/ip4/127.0.0.1/tcp/40800/p2p/12D3KooWH3uVF6wv47WnArKHk5p6cvgCJEb74UTmxztmQDc298L3`.

In thrid console, I connect to the second console and request file `lock` in first console:
```
cargo run --example file-sharing -- --peer /ip4/127.0.0.1/tcp/40800/p2p/12D3KooWH3uVF6wv47WnArKHk5p6cvgCJEb74UTmxztmQDc298L3 get --name toml
```
Now the third console panics with this informations:
```
thread 'async-std/runtime' panicked at 'Dialing(PeerId("12D3KooWPjceQrSwdWXPyLLeABRXmuqt69Rg3sBYbU1Nft9HyQ6X"))', examples/file-sharing.rs:502:22
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
thread 'main' panicked at 'Sender not to be dropped.: Canceled', examples/file-sharing.rs:332:28
```
Meanwhile, the first console also panics with this informations:
```
[...]

PR #2580, opened 2022-03-21
Exclude two swarm event from panic in file-sharing example

YoshieraHuang:
Related to #2546

### DIFF
--- a/examples/file-sharing.rs
+++ b/examples/file-sharing.rs
@@ -499,7 +499,7 @@ mod network {
                         }
                     }
                 }
-                SwarmEvent::IncomingConnectionError { .. } => {},
+                SwarmEvent::IncomingConnectionError { .. } => {}
                 SwarmEvent::Dialing(peer_id) => println!("Dialing {}", peer_id),
                 e => panic!("{:?}", e),
             }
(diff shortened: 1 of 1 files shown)


examples/file-sharing.rs
```diff
@@ -499,6 +499,9 @@ mod network {
                         }
                     }
                 }
+                // Exclude these two events from panic to prevent unexpected panic. (Issue #2546)
+                SwarmEvent::IncomingConnectionError { error, .. } => println!("{}", error),
+                SwarmEvent::Dialing(peer_id) => println!("{}", peer_id),
```
mxinden:
```suggestion
                SwarmEvent::IncomingConnectionError { .. } => {},
                SwarmEvent::Dialing(peer_id) => println!("Dialing {}", peer_id),
```

mxinden: I don't think we need to print the `IncomingConnectionError`. I would guess these are only due to concurrent dialing where the remote favors one established connection over another.

mxinden (approved):
Thanks @YoshieraHuang!

examples/file-sharing.rs
```diff
@@ -499,6 +499,8 @@ mod network {
                         }
                     }
                 }
+                SwarmEvent::IncomingConnectionError { .. } => {},
```
mxinden:
```suggestion
                SwarmEvent::IncomingConnectionError { .. } => {}
```

PR #2586, opened 2022-03-23
Expose Error Types of Relay and DCUtR

canewsin:
This Exposure is required for passing swarm event to handler methods, instead dealing them at one place.
```
event = swarm1.select_next_some() => {
      match event {
                 event =>  DecentNetworkBehaviour::handle_swarm_event(&mut swarm1, event, &client_config1)
     }
}
```

### DIFF
--- a/protocols/relay/src/v2.rs
+++ b/protocols/relay/src/v2.rs
@@ -30,6 +30,13 @@ mod copy_future;
 mod protocol;
 pub mod relay;
 
+pub use protocol::{
+    inbound_hop::FatalUpgradeError as InHopUpgradeError,
+    inbound_stop::FatalUpgradeError as InStopUpgradeError,
+    outbound_hop::FatalUpgradeError as OutHopUpgradeError,
+    outbound_stop::FatalUpgradeError as OutStopUpgradeError,
+};
+
 /// The ID of an outgoing / incoming, relay / destination request.
 #[derive(Debug, Copy, Clone, PartialEq, Eq, Hash)]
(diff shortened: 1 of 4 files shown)


protocols/relay/src/v2.rs
```diff
@@ -30,6 +30,13 @@ mod copy_future;
 mod protocol;
 pub mod relay;
 
+pub use protocol::{
+    inbound_hop::FatalUpgradeError as InHopUpgradeError,
```
mxinden:
```suggestion
    inbound_hop::FatalUpgradeError as InboundHopFatalUpgradeError,
```

While lengthy, I would prefer writing these out for the sake of being easier to understand.

mxinden (commented):
Thanks for the patch. One comment. Also missing changelog entries in `protocols/dcutr/changelog.md` and `protocols/relay/changelog.md`.

canewsin:
> Thanks for the patch. One comment. Also missing changelog entries in `protocols/dcutr/changelog.md` and `protocols/relay/changelog.md`.

@mxinden 
Should I add change log under unreleased versions for each protocol?

mxinden:
> Should I add change log under unreleased versions for each protocol?

Yes please. See https://github.com/libp2p/rust-libp2p/blob/master/docs/release.md

canewsin:
@mxinden 👍

mxinden:
Merged via https://github.com/libp2p/rust-libp2p/pull/2605. Thanks @canewsin!

PR #2595, opened 2022-03-28
protocols/gossipsub: Allow score buckets to be set by the user and/or adjusted from the scoring thresholds

divagant-martian:
Moves the score buckets to the metrics configuration, setting the same defaults as what we had but also allowing the user to pass a reference to the score thresholds to create them from that. Having them in the config also allows users to set them directly

### DIFF
--- a/protocols/gossipsub/CHANGELOG.md
+++ b/protocols/gossipsub/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 - Fix gossipsub metric (see [PR 2558]).
 
+- Allow the user to set the buckets for the score histogram, and to adjust them from the score thresholds.
+
 [PR 2558]: https://github.com/libp2p/rust-libp2p/pull/2558
 
 # 0.36.0 [2022-02-22]
--- a/protocols/gossipsub/CHANGELOG.md
+++ b/protocols/gossipsub/CHANGELOG.md
@@ -4,9 +4,10 @@
 
 - Fix gossipsub metric (see [PR 2558]).
 
-- Allow the user to set the buckets for the score histogram, and to adjust them from the score thresholds.
+- Allow the user to set the buckets for the score histogram, and to adjust them from the score thresholds. See [PR 2595].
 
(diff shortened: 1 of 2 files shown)


protocols/gossipsub/src/metrics.rs
```diff
@@ -49,13 +49,48 @@ pub struct Config {
...
     pub max_never_subscribed_topics: usize,
+    /// Buckets used for the score histograms.
+    pub score_buckets: Vec<f64>,
+}
+
+impl Config {
+    /// Create buckets for the score histograms based on score thresholds.
+    pub fn buckets_using_scoring_thresholds(&mut self, params: &crate::PeerScoreThresholds) {
+        self.score_buckets = vec![
+            params.graylist_threshold,
```
mxinden, line 60: I am not sure how valuable values calculated based on the bucket values are if the buckets themselves don't follow some consistent distribution. Would it not make sense to take the smallest of all the values and do an exponential distribution to the max value (`100.0`)?

mxinden: You probably know best from your current measurements. How well does the histogram mirror real world values?

divagant-martian: The value here is to understand how often nodes are treated in a special way because they have passed the thresholds, more than to understand the distribution of scores itself. In a way is more a measure of network health/score param behaviour than a measure of the scores

protocols/gossipsub/CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
 
 - Fix gossipsub metric (see [PR 2558]).
 
+- Allow the user to set the buckets for the score histogram, and to adjust them from the score thresholds.
+
 [PR 2558]: https://github.com/libp2p/rust-libp2p/pull/2558
```
```suggestion
- Allow the user to set the buckets for the score histogram, and to adjust them from the score thresholds. See [PR 2595].

[PR 2558]: https://github.com/libp2p/rust-libp2p/pull/2558
[PR 2595]: https://github.com/libp2p/rust-libp2p/pull/2595
```

mxinden (approved):

PR #2600, opened 2022-03-30
Maintenance: update to clap 3.1 from structopt

TotalKrill:
structopt is no longer actively maintained, and all features are in clap v3 that is also released. This basically just shortens the dependency tree a bit, since structopt anyway was depending on clap.


mxinden (approved):
Very kind. Thanks for the help @TotalKrill!

Issue #2461, opened 2022-01-30
Websocket yamux regression in v0.42


appaquet:
I tried upgrading my [project](https://github.com/appaquet/exocore) to 0.42 in which I use WebSocket transport on web nodes. Inbetween TCP nodes everything is fine, but it seems like there is a regression when using WebSocket transport in my wasm compiled web client. 

I've bisected the issue to the bump to yamux 0.10, and more precisely this [commit](https://github.com/libp2p/rust-yamux/commit/eb6c73201822d2835320414b133ad6c36a5bbafb) in yamux.

It seems like communication is either not working, sporadic or very slow. One example is the ping protocol that takes 10s to ping the server from the web client. One interesting fact is that it seems like more messages get through if I generate more traffic (ex: generate more messages to the server). My hunch is that the WebSocket transport has different behaviour in terms of connection states that is not handled correctly by yamux, and generating traffic gets it out of some kind of pending state.

See [this gist](https://gist.github.com/appaquet/8b6b95fc3150c1abec8f5e73f1412339) for libp2p & yamux logs on the server and on the web client.

Unfortunately, my project may be a bit complex to get running to show a minimal working example. Do you have any WebSocket / wasm example code that I could use to create a minimal reproducible code? 

Thanks !

mxinden:
Thanks for reporting. Definitely worth investigating.

> Unfortunately, my project may be a bit complex to get running to show a minimal working example. Do you have any WebSocket / wasm example code that I could use to create a minimal reproducible code?

I don't have one handy. Maybe @wngr?

appaquet:
It seems like I was wrong on the origin of the problem in yamux's commit history. I've narrowed the problem to https://github.com/libp2p/rust-yamux/pull/112

It seems like the issue comes from not fully awaiting the flush of the underlying socket in the connection. 

The problems get solved by correctly awaiting the flush like this:
```
diff --git a/src/connection.rs b/src/connection.rs
index 60bed2c..4287dca 100644
--- a/src/connection.rs
+++ b/src/connection.rs
@@ -883,8 +883,7 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Connection<T> {
     /// Try to flush the underlying I/O stream, without waiting for it.
     async fn flush_nowait(&mut self) -> Result<()> {
         future::poll_fn(|cx| {
-            let _ = self.socket.get_mut().poll_flush(cx)?;
-            Poll::Ready(Ok(()))
[...]

PR #2606, opened 2022-04-06
yamux 0.10.1, set_split_send_size

folex:
- Update yamux to 0.10.1
- Add `set_split_send_size` to `YamuxConfig`

As I understand, it affects #2461.

P.S. I'm kinda surprised you don't commit Cargo.lock, is there a reason for that? From my PoV, it makes CI builds unpredictable and generates hard-to-investigate dependency bugs.

### DIFF
--- a/muxers/yamux/src/lib.rs
+++ b/muxers/yamux/src/lib.rs
@@ -290,6 +290,13 @@ impl YamuxConfig {
         self
     }
 
+    /// Set the max. payload size used when sending data frames. Payloads larger
+    /// than the configured max. will be split.
+    pub fn set_split_send_size(&mut self, n: usize) -> &mut Self {
+        self.inner.set_split_send_size(n);
+        self
+    }
+
     /// Converts the config into a [`YamuxLocalConfig`] for use with upgrades
     /// of I/O streams that are ![`Send`].
     pub fn into_local(self) -> YamuxLocalConfig {
--- a/muxers/yamux/CHANGELOG.md
+++ b/muxers/yamux/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 - Update to `libp2p-core` `v0.33.0`.
 
+- Update `yamux` to `v0.10.1`.
+
+- Add `set_split_send_size` to `YamuxConfig`
+
 # 0.36.0 [2022-02-22]
 
 - Update to `libp2p-core` `v0.32.0`.
--- a/muxers/yamux/Cargo.toml
+++ b/muxers/yamux/Cargo.toml
@@ -15,4 +15,4 @@ futures = "0.3.1"
 libp2p-core = { version = "0.33.0", path = "../../core", default-features = false }
 parking_lot = "0.12"
 thiserror = "1.0"
-yamux = "0.10.0"
+yamux = "0.10.1"


Cargo.toml
```diff
@@ -117,7 +117,7 @@ libp2p-gossipsub = { version = "0.37.0", path = "protocols/gossipsub", optional
 async-std = { version = "1.6.2", features = ["attributes"] }
 async-trait = "0.1"
 env_logger = "0.9.0"
-clap = {version = "3.1.6", features = ["derive"]}
+structopt = "0.3.21"
```
mxinden: Accidental merge, or is it on purpose to revert https://github.com/libp2p/rust-libp2p/pull/2600?

muxers/yamux/src/lib.rs
```diff
@@ -290,6 +290,13 @@ impl YamuxConfig {
         self
     }
 
+    /// Set the max. payload size used when sending data frames. Payloads larger
+    /// than the configured max. will be split.
+    pub fn set_split_send_size(&mut self, n: usize) -> &mut Self {
```
mxinden, line 297: Out of curiosity, why do you need this flag?

mxinden (commented):
> P.S. I'm kinda surprised you don't commit Cargo.lock, is there a reason for that? From my PoV, it makes CI builds unpredictable and generates hard-to-investigate dependency bugs.

This is a best practice as a library. Which dependency bug did you run into?

muxers/yamux/src/lib.rs
ValeryAntopol: We had a hard time debugging issue when we encountered UnexpectedEOF errors. Payload data was `N * split_send_size`, so we thought that it can affect the problem, and wondered why it is impossible to change this value. The reason was that we forgot to close stream after write and it got dropped immediately after write. 

Regarding this issue, i thought that it can be useful to add panic, assert or at least warning when garbage collecting stream in "Open" state [here](https://github.com/libp2p/rust-yamux/blob/058010157e5c58976ce0fe78a84bf88ec9dd8656/src/connection.rs#L1003). I have seen that you claim that `poll_close` must be called before dropping the stream [here](https://github.com/libp2p/rust-yamux/issues/117#issuecomment-824027292), so it may help users catch the same issue before it leads to serious consequences.

ValeryAntopol: Also you can add a comment in the examples mentioning the importance of the closing of the socked, like [here](https://github.com/libp2p/rust-libp2p/blob/master/examples/file-sharing.rs#L711). I can make a PR or issue if you agree

Cargo.toml
folex: accidental, will remove

folex:
> This is a best practice as a library. Which dependency bug did you run into?

We're running build on nightly, so sometimes `cargo update` brings compilation errors. Without `Cargo.lock` commited, every build on CI is basically preceded by `cargo update`. 

It's the usual practice to commit `.lock` files in JS and [Haskell](https://docs.haskellstack.org/en/stable/lock_files/). But in Rust, Cargo book suggests against it for library authors. And it makes little sense to me, cuz `Cargo.lock` only affects CI/local builds, it doesn't affect publishing to crates.io.

So I'm curious on your reasons to do that :)

folex:
Anything else I can do to merge this PR?

folex:
@mxinden anything else I can do to merge this PR?

muxers/yamux/src/lib.rs
mxinden: > Also you can add a comment in the examples mentioning the importance of the closing of the socked, like [here](https://github.com/libp2p/rust-libp2p/blob/master/examples/file-sharing.rs#L711). I can make a PR or issue if you agree

Sure. That sounds good to me @ValeryAntopol.

mxinden:
> It's the usual practice to commit `.lock` files in JS and [Haskell](https://docs.haskellstack.org/en/stable/lock_files/). But in Rust, Cargo book suggests against it for library authors. And it makes little sense to me, cuz `Cargo.lock` only affects CI/local builds, it doesn't affect publishing to crates.io.
> 
> So I'm curious on your reasons to do that :)

No particular reasoning on my end other than following the best practice. I guess it does reduce the maintenance work a bit as one does not need to keep it up to date, e.g. via Dependabot.

mxinden:
> @mxinden anything else I can do to merge this PR?

I still don't understand why you need to configure this flag. Can you expand on that once more? I am reluctant to add a configuration flag just for the sake of debugging.

folex:
> > @mxinden anything else I can do to merge this PR?
> 
> I still don't understand why you need to configure this flag. Can you expand on that once more? I am reluctant to add a configuration flag just for the sake of debugging.

I needed it to overcome the issue as I described. I've no other use-case. In general, I feel that it should be possible to configure underlying mechanisms without going through rather hard process of forking the whole libp2p.

thomaseizinger:
> > This is a best practice as a library. Which dependency bug did you run into?
> 
> We're running build on nightly, so sometimes `cargo update` brings compilation errors. Without `Cargo.lock` commited, every build on CI is basically preceded by `cargo update`.
> 
> It's the usual practice to commit `.lock` files in JS and [Haskell](https://docs.haskellstack.org/en/stable/lock_files/). But in Rust, Cargo book suggests against it for library authors. And it makes little sense to me, cuz `Cargo.lock` only affects CI/local builds, it doesn't affect publishing to crates.io.
> 
> So I'm curious on your reasons to do that :)

As library authors, we should strive for having our library work with as many version combinations as possible. Locking down dependencies to specific patch versions is not helpful in that regard.

[By default](https://doc.rust-lang.org/cargo/reference/resolver.html#semver-compatibility), cargo uses the caret modifier for dependency versions, meaning it will only build against that exact version or newer patch versions.

Within semver, there should never be breaking changes within patch versions. I think it is useful to recommend against tracking Cargo.lock in Git for library authors because it has the ecosystem-wide effect that breaking API changes in patch versions are extremely rare and when they happen, they are detected very early. Libraries tend to have less dependencies so it is also easier for library authors to work out, which dependency is the problem instead of putting this burden onto the application projects.

PR #2608, opened 2022-04-06
protocols/ping: Log remote PeerId instead of payload

### LINKED ISSUE #2583: protocols/ping: Reduce noise on logs
Instead of printing the random ping payload, the log should print who we are sending a ping to and from whom we received a ping.

https://github.com/libp2p/rust-libp2p/blob/0c24821a144b10ba805e3648e05c5d44bbd59e42/protocols/ping/src/protocol.rs#L88

leviathanbeak:
In this PR we move Ping logging out of the protocol to the behaviour.

And we also reduce noise by only logging to whom (PeerId) we sent the Ping payload - and from whom we received it,
instead of the payload itself.

it closes #2583


mxinden (approved):
Thanks @leviathanbeak !

PR #2610, opened 2022-04-07
swarm/behaviour: make `either` mod private

elenaf9:
Change the `swarm::behaviour::either` module to be private, since it has no public items: https://docs.rs/libp2p/latest/libp2p/swarm/behaviour/either/index.html.


mxinden (approved):
Thanks @elenaf9!

thomaseizinger (approved):

PR #2615, opened 2022-04-11
*: Fix clippy warnings

LesnyRumcajs:
Hello! After some time tidying up (tried to do something different but I always start with a linter :D ) I noticed you use a custom clippy command. Figured I'll propose some of the changes I made anyway.

Is there a specific reason why many of the lints are disabled? Or do you think it'd be useful to fix the offending pedantic and style issues and disable the suppressions? (complex type may be overkill though)

### DIFF
--- a/core/src/identity/rsa.rs
+++ b/core/src/identity/rsa.rs
@@ -263,7 +263,7 @@ impl DerDecodable<'_> for Asn1SubjectPublicKey {
             )));
         }
 
-        let pk_der: Vec<u8> = object.value().into_iter().skip(1).cloned().collect();
+        let pk_der: Vec<u8> = object.value().iter().skip(1).cloned().collect();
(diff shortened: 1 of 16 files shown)


mxinden (commented):
These changes look good to me. Feel free to enable the corresponding lints in case all of their ocurrences are now fixed with this pull request.

Regarding enabling all of them, I am not familiar with how intrusive that would be.

LesnyRumcajs:
@mxinden Thanks for looking into it. What do you think about merging this and iterating over it to avoid one massive PR on a long-living branch? There are quite a lot of other issues reported by the pedantic clippy (most of which make sense).

mxinden (approved):

mxinden:
> What do you think about merging this and iterating over it

Sounds good to me.

PR #2616, opened 2022-04-12
src/tutorials/hole-punching: Clarify use of libp2p-lookup

winksaville:
Also, add some additional instructions for setting RELAY_SERVER_IP.

### DIFF
--- a/src/tutorials/hole_punching.rs
+++ b/src/tutorials/hole_punching.rs
@@ -75,45 +75,45 @@
 //!
 //! 1. Test that you can connect on Layer 3 (IP).
 //!
-//!     ``` bash
-//!     ping $RELAY_SERVER_IP
-//!     ```
+//!    ``` bash
+//!    ping $RELAY_SERVER_IP
+//!    ```
 //!
 //! 2. Test that you can connect on Layer 4 (TCP).
 //!
-//!     ``` bash
-//!     telnet $RELAY_SERVER_IP 4001
-//!     ```
+//!    ``` bash
+//!    telnet $RELAY_SERVER_IP 4001
+//!    ```
 //!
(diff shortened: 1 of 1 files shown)


mxinden (approved):
Appreciated. Thanks @winksaville!

winksaville:
The branch is out-of-date, should I change the "Update branch" to "Update with rebase" or leave it at "Update with merge commit"?

mxinden:
> The branch is out-of-date, should I change the "Update branch" to "Update with rebase" or leave it at "Update with merge commit"?

No need to update the branch. I can do this through the UI as I just did. In general, I prefer merge commits over rebases, as that does not change history and thus does not requires a reviewer to start from scratch, but only requires a reviewer to review the additional changes.

mxinden:
Not sure why CI is failing though:

```
error: unknown start of token: `
 --> src/tutorials/hole_punching.rs:100:1
  |
4 | ``` bash
  | ^
  |
help: Unicode character '`' (Grave Accent) looks like ''' (Single Quote), but it is not
```

winksaville:
That's weird, I'll take a look later today.

winksaville:
Turns out the error was caused by the indentation being 4 rather than 3 spaces as I explained in the second commit! I've decided to change the other 3 sections too three spaces so there would be consistency.

I have kept the two commits so you can see the change, but maybe these two changes should be squashed,
which I can do if you want, or you can leave it as two or squash it when merging or ...

I've tested with `cargo test --workspace --no-default-features` and it failed before the change and now succeeds!

Anyway, let me know what more, if anything, you'd like me to do.

mxinden:
Thanks for the help @winksaville!

PR #2618, opened 2022-04-17
protocols/autonat: optionally use only global IPs

### LINKED ISSUE #2514: AutoNAT: how to handle non-global IP addresses?

elenaf9:
# Description

Optionally only perform dial-backs on peers that are observed at a global ip-address. 
This is relevant when multiple peers are in the same local network, in which case a peer could incorrectly assume themself to be public because a peer in the same local network was able to dial them. Thus servers should reject dial-back requests from clients with a non-global IP address, and at the same time clients should only pick connected peers as servers if they are global.
Behind a config flag (enabled by default) to also allow use-cases where AutoNAT is needed within a private network.
Continuation/ alternative to #2526 (cc @hamamo). Opened as a draft for now because of the open questions below.

Fixes #2514.

## Open Questions

The `is_global` check for Ipv4 and Ipv6 addresses currently implements the same logic as `std::net::Ipv4Addr::is_global`, `std::net::Ipv6Addr::is_global`, which are at the time of writing unstable behind the `ip` flag. 
(See https://github.com/rust-lang/rust/issues/27709 and https://github.com/rust-lang/rust/issues/85604.)

For Ipv4 most of the internally called methods are stable, apart from the checks for `is_shared`, `is_reserved` and `is_benchmarking`. Could we just remove those checks? In case of `is_shared` the peer is still eligible for a dial-back and the other two cases should never occur in practice, no? In general, would it be enough to just check for `is_private`, `is_loopback` and `is_link_local`? I am not sure, but since we only use this check for observed addresses, can any of the other cases even happen?

For Ipv6 there is still some ongoing discussion around the `is_global` check, because is currently checks for the (global) scope of the address rather than global reachability (See https://github.com/rust-lang/rust/pull/86634 and https://github.com/rust-lang/rust/issues/85604). There are ongoing efforts to fix this (and eventually stabilize all `ip` methods), but they seem to be inactive since last year. Is this something that we could / should help push to completion?

## Change checklist


protocols/autonat/src/behaviour.rs
```diff
@@ -512,3 +521,97 @@ trait HandleInnerEvent {
...
+impl GlobalIp for std::net::Ipv4Addr {
+    // NOTE: The below logic is copied from `std::net::Ipv4Addr::is_global`, which is at the time of
+    // writing behind the unstable `ip` feature.
+    // See https://github.com/rust-lang/rust/issues/27709 for more info.
+    //
+    // TODO: Consider removing check check with the unstable methods `is_shared`, `is_reserved` and `is_benchmarking`? In case
+    // of the former the peer is still eligible for a dial-back, the latter cases should never occur in practice.
+    //
+    // TODO: Consider to only check for loopback, private ip and link-local. Can any other case ever happen in practice if we solely
+    // use this check for observed addresses?
```
mxinden: I don't think it is worth deviating from the `IPv4Addr::is_global` function for the sake of not copying the code below. I don't expect it to change any time soon. Am I missing something?

protocols/autonat/src/behaviour.rs
```diff
@@ -512,3 +521,97 @@ trait HandleInnerEvent {
...
+}
+
+impl GlobalIp for std::net::Ipv6Addr {
+    // NOTE: The below logic is copied from `std::net::Ipv6Addr::is_global`, which is at the time of
+    // writing behind the unstable `ip` feature.
+    // See https://github.com/rust-lang/rust/issues/27709 for more info.
+    //
+    // Note that contrary to `Ipv4Addr::is_global_ip` this currently checks for global scope
+    // rather than global reachability.
+    // TODO: There is a PR to change this, but seems inactive. Should we help push this forward?
```
mxinden: Is the distinction between reachability and scope relevant in our case?

Either way, it is always worth it for the sake of helping the Rust community.

mxinden (commented):
This looks good to me overall. Just two comments:

protocols/autonat/src/behaviour.rs
```diff
@@ -512,3 +521,97 @@ trait HandleInnerEvent {
...
+        // globally routable addresses in the 192.0.0.0/24 range.
+        if u32::from_be_bytes(self.octets()) == 0xc0000009
+            || u32::from_be_bytes(self.octets()) == 0xc000000a
+        {
+            return true;
+        }
+
+        // Copied from the unstable method `std::net::Ipv4Addr::is_shared`.
+        let is_shared =
+            || self.octets()[0] == 100 && (self.octets()[1] & 0b1100_0000 == 0b0100_0000);
```
thomaseizinger: It might be worth to do a literal 1-to-1 copy of the function and simply replace the `self` parameter with `ip: Ipv4Addr`. The only difference would be the calling convention but it may be easier to audit that they are exactly the same.

protocols/autonat/src/behaviour.rs
elenaf9: It would be for the sake of avoiding unstable methods. `is_shared`, `is_reserved` and `is_benchmarking` are the only non-stable methods calls within the `is_global` implementation. In particular for `is_reserved` there is a warning:
> As IANA assigns new addresses, this logic will be updated. This may result in non-reserved addresses being treated as reserved in code that relies on an outdated version of this method.

If those are not relevant for our case anyway, we would avoid the risk of being outdated if the std implementation changes.

protocols/autonat/src/behaviour.rs
elenaf9: 👍 Done with 099b258 and 9c2d901.

protocols/autonat/src/behaviour.rs
mxinden: Off the top of my head I would favor consistency with `std` over avoiding the risk of being outdated. I.e. this code being an exact copy of `std` makes it easier to maintain.

protocols/autonat/src/behaviour/as_server.rs
```diff
@@ -295,8 +295,8 @@ impl<'a> AsServer<'a> {
             .values()
             .find_map(|a| a.as_ref())
             .ok_or_else(|| {
-                let status_text = "no dial-request over relayed connections".to_string();
-                (status_text, ResponseError::DialError)
+                let status_text = "refusing to dial peer with blocked observed address".to_string();
+                (status_text, ResponseError::DialRefused)
```
elenaf9, line 299: Note: the go implementation returns a `ResponseError::DialError` here (but with the same status text). I decided to change it here because imo a `DialRefused` makes more sense. A client should not flip their status to private or reduce the confidence in their public status just because they accidentally picked a server that rejects them because of  the observed address (though if the client also enabled `only_global_ips` this should not happen in practice).
Edit: should it be part of the spec to list the cases  in which a server returns `DialError`?

protocols/autonat/CHANGELOG.md
```diff
@@ -6,6 +6,11 @@
 
 - Update to `libp2p-request-response` `v0.18.0`.
 
+- Add `Config::only_global_ips` to skip peers that are observed at a private IP-address
+  (see PR 2618).
```
mxinden:
```suggestion
  (see [PR 2618]).
```

protocols/autonat/src/behaviour/as_server.rs
mxinden: >  Edit: should it be part of the spec to list the cases in which a server returns DialError?

Yes. I think this is worth consolidating on. Mind driving it on libp2p/specs?

mxinden (approved):
This is good to go from my end. Just a small nit above.

protocols/autonat/src/behaviour/as_server.rs
elenaf9: Opened https://github.com/libp2p/specs/issues/411.

elenaf9: Also added 7035422, which does the same change for the case that `filter_valid_addrs` filters out all addresses send from the client.

protocols/autonat/src/behaviour.rs
```diff
@@ -512,3 +524,119 @@ trait HandleInnerEvent {
...
+            is_unicast(addr)
+                && !addr.is_loopback()
+                && !is_unicast_link_local(addr)
+                && !is_unique_local(addr)
+                && !addr.is_unspecified()
+                && !is_documentation(addr)
+        }
+
+        // Variation of unstable method [`std::net::Ipv6Addr::multicast_scope`] that instead of the
+        // `Ipv6MulticastScope` just return if the scope is global or not.
```
elenaf9:
```suggestion
        // `Ipv6MulticastScope` just returns if the scope is global or not.
```

mxinden (approved):
Latest changes look good to me. Feel free to merge once ready.

protocols/autonat/CHANGELOG.md
```diff
@@ -6,6 +6,11 @@
 
 - Update to `libp2p-request-response` `v0.18.0`.
 
+- Add `Config::only_global_ips` to skip peers that are observed at a private IP-address
+  (see [PR 2618]).
+
+[PR 2445]: https://github.com/libp2p/rust-libp2p/pull/2618
```
elenaf9:
```suggestion
[PR 2618]: https://github.com/libp2p/rust-libp2p/pull/2618
```

PR #2620, opened 2022-04-25
*: Activate `clippy::style` lint group

LesnyRumcajs:
# Description

Enabled clippy style warnings and fixed the corresponding issues. I had to suppress some of them to not break the existing API (we could consider resolving those for a major release).

https://github.com/libp2p/rust-libp2p/pull/2615

## Change checklist

### DIFF
--- a/protocols/gossipsub/src/behaviour.rs
+++ b/protocols/gossipsub/src/behaviour.rs
@@ -1226,19 +1226,19 @@ where
             }
 
             for id in ids {
-                if !self.duplicate_cache.contains(&id) && !self.pending_iwant_msgs.contains(&id) {
-                    if self
+                if !self.duplicate_cache.contains(&id)
+                    && !self.pending_iwant_msgs.contains(&id)
+                    && self
                         .peer_score
                         .as_ref()
                         .map(|(_, _, _, promises)| !promises.contains(&id))
                         .unwrap_or(true)
-                    {
-                        // have not seen this message and are not currently requesting it
-                        if iwant_ids.insert(id) {
-                            // Register the IWANT metric
-                            if let Some(metrics) = self.metrics.as_mut() {
-                                metrics.register_iwant(&topic);
-                            }
+                {
+                    // have not seen this message and are not currently requesting it
+                    if iwant_ids.insert(id) {
+                        // Register the IWANT metric
+                        if let Some(metrics) = self.metrics.as_mut() {
+                            metrics.register_iwant(&topic);
                         }
                     }
                 }
@@ -1353,7 +1353,7 @@ where
             } else if let Some(m) = self.metrics.as_mut() {
                 // Sending of messages succeeded, register them on the internal metrics.
                 for topic in topics.iter() {
-                    m.msg_sent(&topic, msg_bytes);
+                    m.msg_sent(topic, msg_bytes);
                 }
             }
         }
@@ -2136,7 +2136,7 @@ where
             for peer_id in self.connected_peers.keys() {
                 scores
                     .entry(peer_id)
-                    .or_insert_with(|| peer_score.metric_score(&peer_id, self.metrics.as_mut()));
+                    .or_insert_with(|| peer_score.metric_score(peer_id, self.metrics.as_mut()));
             }
         }
 
--- a/protocols/gossipsub/src/behaviour.rs
+++ b/protocols/gossipsub/src/behaviour.rs
@@ -1215,6 +1215,21 @@ where
 
         let mut iwant_ids = HashSet::new();
 
+        let want_message = |id: &MessageId| {
+            if self.duplicate_cache.contains(id) {
+                return false;
+            }
+
+            if self.pending_iwant_msgs.contains(id) {
+                return false;
+            }
+
+            self.peer_score
+                .as_ref()
+                .map(|(_, _, _, promises)| !promises.contains(id))
+                .unwrap_or(true)
+        };
+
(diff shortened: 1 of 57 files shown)


protocols/rendezvous/src/codec.rs
```diff
@@ -254,6 +254,7 @@ impl Decoder for RendezvousCodec {
     }
 }
 
+#[allow(clippy::enum_variant_names)]
```
thomaseizinger: I wouldn't mind breaking this API and fix the lint instead. We make new minor releases of `libp2p-swarm` all the time.

Just add a changelog entry please if you decide to do it :)

swarm/src/connection.rs
```diff
@@ -176,6 +176,7 @@ pub struct IncomingInfo<'a> {
 
 impl<'a> IncomingInfo<'a> {
     /// Builds the [`ConnectedPoint`] corresponding to the incoming connection.
+    #[allow(clippy::wrong_self_convention)]
```
thomaseizinger: Same here.

I'd be in favor of fixing this, even though it is a small API break.

protocols/gossipsub/src/behaviour.rs
```diff
@@ -1226,19 +1226,19 @@ where
...
             for id in ids {
-                if !self.duplicate_cache.contains(&id) && !self.pending_iwant_msgs.contains(&id) {
-                    if self
+                if !self.duplicate_cache.contains(&id)
+                    && !self.pending_iwant_msgs.contains(&id)
+                    && self
                         .peer_score
                         .as_ref()
                         .map(|(_, _, _, promises)| !promises.contains(&id))
                         .unwrap_or(true)
```
thomaseizinger: This is quite the monster boolean condition now. Perhaps we should extract 3 boolean variables that we can then chain together to make this more readable?

thomaseizinger (commented):
I am a fan of clippy myself, happy to move forward with this.
What do you think @mxinden?

protocols/gossipsub/src/behaviour.rs
LesnyRumcajs: For readability, it would make sense, but it would kill the possibility of short-circuiting, which may result in a performance drop.

dignifiedquire:
big fan of clippy myself, very much in favor of this in general

protocols/rendezvous/src/codec.rs
LesnyRumcajs: Great, will do! :)

swarm/src/connection.rs
LesnyRumcajs: Done!

protocols/gossipsub/src/behaviour.rs
thomaseizinger: I guess functions would allow for readability and short-circuiting?

thomaseizinger:
Side-note @LesnyRumcajs : We use squash-merge so please don't force-push :)
Makes it easier to see the diff since the last review!

protocols/gossipsub/src/behaviour.rs
LesnyRumcajs: Indeed, a lambda would do the trick! For the first two conditions I don't think it would help much, i.e. `!self.duplicate_cache.contains(&id)` would most likely become `duplicate_cache.contains(&id)`. Is it fine for you?

The last one is definitely a monster though and I'd be happy to turn it into a short lambda. Any suggestions on a meaningful name for it?

LesnyRumcajs:
> Side-note @LesnyRumcajs : We use squash-merge so please don't force-push :) Makes it easier to see the diff since the last review!

Sure thing. I was uncertain how it would behave coming from my fork which was a bit outdated so I defaulted to my typical routine. :)

protocols/gossipsub/src/behaviour.rs
thomaseizinger: Probably even better would be to do:

```rust
impl Gossipsub {
	fn want_message(&self, id: ???) -> bool {
		if self.duplicate_cache.contains(&id) {
			return false;
		}

		if self.pending_iwant_msgs.contains(&id) {
			return false;
		}

		self.peer_score
     		  .as_ref()
            .map(|(_, _, _, promises)| !promises.contains(&id))
            .unwrap_or(true)
	}
}
```

Although I think clippy might complain about this one being possible to write as a `&&` chain.

LesnyRumcajs: Good idea! I went ahead with similar change. I was a bit confused with a `contains` method being mutable for no visible reason so I made them immutable. Maybe pedantic clippy would have caught that before.

protocols/kad/CHANGELOG.md
```diff
@@ -6,7 +6,10 @@
 
 - Derive `Error` for `GetRecordError` (see [PR 2614]).
 
-[pr 2614]: https://github.com/libp2p/rust-libp2p/pull/2614
+- Renamed `PeriodicJob::is_ready` to `PeriodicJob::check_ready`. See [PR 2620].
```
mxinden: This method is not exposed and thus would not need to be mentioned in the changelog, right?

mxinden (approved):
Just one nit pick. Otherwise looks good to me. Thanks for the help!

protocols/kad/CHANGELOG.md
LesnyRumcajs: Indeed! Fixed.

PR #2625, opened 2022-05-02
swarm/src/connection/pool: Remove 'a lifetime in PoolEvent

mxinden:
# Description

Simplifies `PoolEvent`, no longer carrying a reference to an
`EstablishedConnection` or the `Pool`, but instead the `PeerId`,
`ConnectionId` and `ConnectedPoint` directly.

## Change checklist

    - No change in functionality.
    - Not needed. `PoolEvent` is not exposed.

### DIFF
--- a/swarm/src/connection/pool.rs
+++ b/swarm/src/connection/pool.rs
@@ -517,10 +517,7 @@ where
         Ok(connection_id)
     }
     /// Polls the connection pool for events.
-    ///
-    /// > **Note**: We use a regular `poll` method instead of implementing `Stream`,
-    /// > because we want the `Pool` to stay borrowed if necessary.
-    pub fn poll<'a>(&'a mut self, cx: &mut Context<'_>) -> Poll<PoolEvent<THandler, TTrans>>
+    pub fn poll(&mut self, cx: &mut Context<'_>) -> Poll<PoolEvent<THandler, TTrans>>
     where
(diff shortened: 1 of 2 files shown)


swarm/src/connection/pool.rs
```diff
@@ -586,7 +520,7 @@ where
     ///
     /// > **Note**: We use a regular `poll` method instead of implementing `Stream`,
     /// > because we want the `Pool` to stay borrowed if necessary.
-    pub fn poll<'a>(&'a mut self, cx: &mut Context<'_>) -> Poll<PoolEvent<'a, THandler, TTrans>>
+    pub fn poll<'a>(&'a mut self, cx: &mut Context<'_>) -> Poll<PoolEvent<THandler, TTrans>>
```
elenaf9:
```suggestion
    pub fn poll(&mut self, cx: &mut Context<'_>) -> Poll<PoolEvent<THandler, TTrans>>
```
I think we can remove the lifetime here as well?

elenaf9 (approved):
Nice!

swarm/src/connection/pool.rs
```diff
@@ -586,7 +520,7 @@ where
     ///
     /// > **Note**: We use a regular `poll` method instead of implementing `Stream`,
     /// > because we want the `Pool` to stay borrowed if necessary.
```
mxinden:
```suggestion
```

thomaseizinger (approved):
Nice work!

PR #2626, opened 2022-05-02
swarm/src/connection: Prioritize handler over connection

mxinden:
# Description

Prioritize work in handler over work on connection, thus prioritizing local work
over work coming from a remote.

I would like to make _prioritizing local work over work coming from a remote_ a pattern throughout the codebase. More to come.

I will run this change on a larger machine to make sure it does not introduce any subtle errors.

## Change checklist

- [ ] A changelog entry has been made in the appropriate crates.
    - I don't think this change needs a call-out in the changelog.


thomaseizinger (approved):
Concept ACK

Was there a specific motivation for this change in prioritization?

mxinden:
> Was there a specific motivation for this change in prioritization?

General theme of prioritizing local work over work from remote. I hope for this to improve both performance and security, though applying the pattern across the stack is likely required for that to show.

Thanks for the review @thomaseizinger!

PR #2627, opened 2022-05-02
swarm/src/lib: Prioritize Behaviour over Pool and Pool over Listeners 

mxinden:
# Description

Builds on top of https://github.com/libp2p/rust-libp2p/pull/2625. 

Have the main event loop (`Swarm::poll_next_event`) prioritize:

1. Work on `NetworkBehaviour` over work on `Pool`, thus prioritizing
   local work over work coming from a remote.

2. Work on `Pool` over work on `ListenersStream`, thus prioritizing work
   on existing connections over upgrading new incoming connections.

Refactors `NetworkBehaviourAction`, `PoolEvent` and `ListenersEvent` handling along the way through 9f0ca957d170a4364f18991a1c10f4d8fbc19436 87b7cb3cb59ec25731e130bc4e966b75c9a0d4d8 00346de814a950f317018e4aa79ed8c79a85f41d to simplify the actual change, namely 426023ecdfb35f78c6a03a8f1d5397700adf97d7. Thus best reviewed one commit at a time.

I will run this change on a larger machine to make sure it does not introduce any subtle errors.

Related to https://github.com/libp2p/rust-libp2p/pull/2626.

## Change checklist

    - I don't think this needs to be called out in the changelog.


swarm/src/lib.rs
```diff
@@ -514,7 +514,7 @@ where
                     };
                     match dial {
                         Ok(fut) => fut
-                            .map(|r| (address, r.map_err(TransportError::Other)))
+                            .map(|r| (address, r.map_err(|e| TransportError::Other(e))))
```
thomaseizinger: Why not stick to passing the variant as a function reference?

swarm/src/lib.rs
```diff
@@ -1018,88 +1018,92 @@ where
         // across a `Deref`.
         let this = &mut *self;
 
+        // This loop polls the components below in a prioritized order.
+        //
+        // 1. [`NetworkBehaviour`]
```
thomaseizinger, line 1023: Nit: Unless you are in rustdoc land (`///`), I don't think this reference will do anything.

swarm/src/lib.rs
```diff
@@ -1018,88 +1018,92 @@ where
...
+        // 2. Connection [`Pool`]
+        // 3. [`ListenersStream`]
+        //
+        // (1) is polled before (2) to prioritize local work over work coming from a remote.
+        //
+        // (2) is polled before (3) to prioritize existing connections over upgrading new incoming connections.
         loop {
-            let mut listeners_not_ready = false;
-            let mut connections_not_ready = false;
```
thomaseizinger: :+1: for getting rid of these flags!

swarm/src/lib.rs
```diff
@@ -1018,88 +1018,92 @@ where
...
-                };
-                this.behaviour.poll(cx, &mut parameters)
             };
 
-            match behaviour_poll {
-                Poll::Pending if listeners_not_ready && connections_not_ready => {
-                    return Poll::Pending
-                }
+            // Poll the listener(s) for new connections.
+            match ListenersStream::poll(Pin::new(&mut this.listeners), cx) {
```
thomaseizinger, line 1095: If `ListenersStream` would actually implement `Stream` then we could utilise `poll_unpin` here.

thomaseizinger (approved):
Nice work! :)

swarm/src/lib.rs
mxinden: Yeah, I am hoping for this to eventually become a feature and if not, at least it is consistent when find-replacing. I would keep it as is unless you feel strongly about it @thomaseizinger.

mxinden:
As an example @winksaville https://github.com/libp2p/rust-libp2p/pull/2627/commits/426023ecdfb35f78c6a03a8f1d5397700adf97d7 is one of those changes on a complex `poll` `loop` method. Error prone indeed, though I don't see a better way for doing it.

swarm/src/lib.rs
mxinden: True, though we would have to handle the `None` case which can never happen. https://github.com/rust-lang/futures-rs/issues/1857 would be very useful here.

swarm/src/lib.rs
mxinden:
```suggestion
                            .map(|r| (address, r.map_err(TransportError::Other)))
```

swarm/src/lib.rs
thomaseizinger: No feeling strongly at all, just figured it might have been a mistake :)

elenaf9:
>  2. Work on Pool over work on ListenersStream, thus prioritizing work
on existing connections over upgrading new incoming connections.

👍 Makes sense.

> 1. Work on NetworkBehaviour over work on Pool, thus prioritizing
local work over work coming from a remote.

Not sure if I agree with this. Many NetworkBehaviours internally have a state that is influenced by events in the pool (e.g. list of connected peers) and the behaviour might rely on the fact that this state is really the most recent one. Also when the behaviour now initiates a dial-attempt, with `PeerCondition::Disconnected`, it could happen that we initiate a dial attempt even though there is a connection to that peer pending in the pool, that has just not been polled to completion yet.
I think prioritization of local work <-> remote work has to happen within the NetworkBehaviour (e.g. do local work first before handling events reported by the ConnectionHandler). 
Could you give an example where you think it makes more sense to poll the NetworkBehaviour before polling the Pool?

mxinden:
> Many NetworkBehaviours internally have a state that is influenced by events in the pool (e.g. list of connected peers) and the behaviour might rely on the fact that this state is really the most recent one

As an aside a `NetworkBehaviour` must not depend on their view of the connection pool to be up-to-date, as they are updated in an eventual consistent way only.

> Also when the behaviour now initiates a dial-attempt, with `PeerCondition::Disconnected`, it could happen that we initiate a dial attempt even though there is a connection to that peer pending in the pool, that has just not been polled to completion yet.

True, though I would argue that this can happen anyways, thus each `NetworkBehaviour` needs to be able to cope with the situation.

> I think prioritization of local work <-> remote work has to happen within the NetworkBehaviour (e.g. do local work first before handling events reported by the ConnectionHandler).

But how would a `NetworkBehaviour` push back on events coming from the `ConnectionHandler`? Today they are  injected via `NetworkBehaviour::inject_event` with no way for the `NetworkBehaivour` to refuse an event other than dropping it.

> Could you give an example where you think it makes more sense to poll the NetworkBehaviour before polling the Pool?

Good question. I should have included an example in the pull request description.

Say we have a request-response style protocol. The remote is sending requests to the local node back-to-back. They are read from the socket in the corresponding `ConnectionHandler`, send to the `Pool` via the connection->pool channel. The `Pool` bubbles the events up the `Swarm`, which in turn forwards the request to the `NetworkBehaviour`.

Say that the `NetworkBehaviour` is busy handling a set of requests. The `Swarm` would thus poll the `NetworkBehaviour` and not the `Pool`. Given that the `Pool` is not polled, new requests are not taken out of the connection->pool channel. Thus the channel eventually fills up and the `ConnectionHandler` stops reading new requests from the socket. Not reading new requests from the socket results in TCPs back-pressure mechanism to kick-in, thus stopping the remote from sending new requests into the socket. Thus the remote is being forced to slow down in order to not overwhelm the local node.

Does the above example make sense @elenaf9?

elenaf9:
Thanks for the explanation @mxinden! Yes that makes sense, I agree with you then that `NetworkBehaviour` should be polled before `Pool`.

mxinden:
> > I think prioritization of local work <-> remote work has to happen within the NetworkBehaviour (e.g. do local work first before handling events reported by the ConnectionHandler).
> 
> But how would a `NetworkBehaviour` push back on events coming from the `ConnectionHandler`? Today they are injected via `NetworkBehaviour::inject_event` with no way for the `NetworkBehaivour` to refuse an event other than dropping it.

By the way, long term it would be cool for a `NetworkBehaviour` to be able to _push back_ on events it can currently not handle. Though I have not been able to think of a way to do so.

elenaf9 (approved):
👍
Also very much in favor of splitting larger `poll` methods into smaller chunks.

rkuhn:
> By the way, long term it would be cool for a NetworkBehaviour to be able to push back on events it can currently not handle. Though I have not been able to think of a way to do so.

So you want Erlang-style selective receive from a Mailbox, or its close cousin [join calculus](https://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.91.946&rep=rep1&type=pdf).

MarcoPolo (approved):
Looks great!

mxinden:
> > By the way, long term it would be cool for a NetworkBehaviour to be able to push back on events it can currently not handle. Though I have not been able to think of a way to do so.
> 
> So you want Erlang-style selective receive from a Mailbox

I guess so. Still have to give it more thought.

> or its close cousin [join calculus](https://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.91.946&rep=rep1&type=pdf).

Will take a look. Thanks @rkuhn.

elenaf9:
One more thing that just came to my mind: 
Because the behaviour is now polled before the pool, I think the promise stated in https://github.com/libp2p/rust-libp2p/pull/2262#discussion_r770868799 does not hold true anymore: 
> In general, `NetworkBehaviour::poll` is guaranteed to be called after `NetworkBehaviour::inject_xxx` is called

This is relevant for the very specific case that a behaviour internally holds a timer that is reset within those method calls. For almost all method calls this won't be a problem because a `SwarmEvent` is returned and afterwards everything is polled again. 
Only for `inject_event` and `inject_address_change` this does not happen.
Looking through our existing behaviour this won't cause issues for any of them. But maybe this is reason enough to amend a changlog stating this change, in case that a user's behaviour relies on this.

mxinden:
> Because the behaviour is now polled before the pool, I think the promise stated in [#2262 (comment)](https://github.com/libp2p/rust-libp2p/pull/2262#discussion_r770868799) does not hold true anymore:
> 
> > In general, `NetworkBehaviour::poll` is guaranteed to be called after `NetworkBehaviour::inject_xxx` is called

I think it does:

Say that `this.pool.poll` resolves to a `PoolEvent::ConnectionEvent`.

https://github.com/libp2p/rust-libp2p/blob/f04f6bb4fcf83a056b399dada4d57f844b29cd9a/swarm/src/lib.rs#L1082-L1092

Thus `this.handle_pool_event(pool_event)` would be called with the `PoolEvent::ConnectionEvent`.

The corresponding `match` arm would call `inject_event` on the `NetworkBehaviour`, potentially making the `NetworkBehaviour` reset a timer.

https://github.com/libp2p/rust-libp2p/blob/f04f6bb4fcf83a056b399dada4d57f844b29cd9a/swarm/src/lib.rs#L805-L811

Given that the match arm does not return, `None` is returned at the end of the method.

https://github.com/libp2p/rust-libp2p/blob/f04f6bb4fcf83a056b399dada4d57f844b29cd9a/swarm/src/lib.rs#L829

Back in `Swarm::poll_next_event` the conditional in the `if` is false (i.e. not `None`) and thus the `continue` statement is hit.

https://github.com/libp2p/rust-libp2p/blob/f04f6bb4fcf83a056b399dada4d57f844b29cd9a/swarm/src/lib.rs#L1086-L1090

Thus execution jumps to the top of the loop where the `NetworkBehaviour` will be called again.

https://github.com/libp2p/rust-libp2p/blob/f04f6bb4fcf83a056b399dada4d57f844b29cd9a/swarm/src/lib.rs#L1031-L1080

@elenaf9 am I missing something?

elenaf9:
> @elenaf9 am I missing something?

Nope, I was missing that there was the `continue` statement after `this.handle_pool_event`. Sorry for the noise.

mxinden:
> > @elenaf9 am I missing something?
> 
> Nope, I was missing that there was the `continue` statement after `this.handle_pool_event`. Sorry for the noise.

By now I feel comfortable with the _either `return` or `continue` once progress (`Poll::Ready`) was made_ pattern. That said, I don't think this is intuitive when first seeing it. Thus happy for alternative suggestions @elenaf9.

Issue #2500, opened 2022-02-09
Migrate all protocols to use `quick-protobuf-codec`


mxinden:
Instead of implementing Protobuf en-/decoding in each protocol and for each message exchange, one can instead implement a custom encoder abstracting the en-/decoding of Protobuf messages.
```[tasklist]
### Tasks
- [ ] https://github.com/libp2p/rust-libp2p/issues/4488
- [ ] https://github.com/libp2p/rust-libp2p/issues/4487
- [ ] https://github.com/libp2p/rust-libp2p/issues/4489
```

nloadholtes:
Hi! I have read through this and would like to take it on. I'm still pretty new to Rust, so apologies in advance if I ask some basic questions. :smiley: 

I see that the `Encoder` and `Decoder` are now in prost-codec, and it looks like `protocols/dcutr`, `protocols/identity`, and `protocols/relay` are using it. I'll work on refactoring the other protocols packages to use prost-codec.

For this ticket, would you prefer 1 PR per package, or everything in 1 big PR? (My plan is to do 1 package in a draft PR and then have you/someone look at it to make sure I'm on the right path, and then go from there)

thomaseizinger:
> Hi! I have read through this and would like to take it on. I'm still pretty new to Rust, so apologies in advance if I ask some basic questions. :smiley: 
> 
> I see that the `Encoder` and `Decoder` are now in prost-codec, and it looks like `protocols/dcutr`, `protocols/identity`, and `protocols/relay` are using it. I'll work on refactoring the other protocols packages to use prost-codec.
> 
> For this ticket, would you prefer 1 PR per package, or everything in 1 big PR? (My plan is to do 1 package in a draft PR and then have you/someone look at it to make sure I'm on the right path, and then go from there)

1 PR per crate would be preferable!

mxinden:
Thanks @nloadholtes for volunteering.

nloadholtes:
Hi @mxinden and @thomaseizinger  , I've made a WIP PR that I'd like some feedback on: #3070

Basically I'm looking at `protocols/dcutr` and trying to make sure that I can make `protocols/gossipsub` "look" like that. I ran the changes locally and the tests passed so I'm hopeful that is a good sign. :smile: 

Thank you in advance for any feedback, I'm still learning the ways of rust, but I'm really liking what I'm seeing so far.

nloadholtes:
Putting this here to track progress (since we're doing 1 PR per crate)

[...]

PR #2630, opened 2022-05-03
 misc/prost-codec: Introduce codec for varint prefixed Protobuf messages 

mxinden:
# Description

Extracts the Protobuf en-/decoding pattern into its separate crate and applies it to `libp2p-identify`.

In case this is accepted, we can update the other crates.

1a8b24747dd4c6147a11400759339554b3b7440d also makes `libp2p-identify` expose an explicit `UpgradeError` instead of an `io::Error`.

https://github.com/libp2p/rust-libp2p/issues/2500

## Change checklist


misc/prost-codec/Cargo.toml
```diff
@@ -0,0 +1,21 @@
+[package]
+name = "prost-codec"
```
thomaseizinger, line 2: Something to include in `asynchronous-codec` perhaps? Feature-flagged of course.

misc/prost-codec/src/lib.rs
```diff
@@ -0,0 +1,81 @@
...
+}
+
+#[derive(Debug, Error)]
+pub enum Error {
+    #[error("Failed to decode response: {0}.")]
+    Decode(
+        #[from]
+        #[source]
+        prost::DecodeError,
+    ),
```
thomaseizinger, line 74: If you are using `#[source]`, you shouldn't include inner error message yourself, otherwise `anyhow` and friends will duplicate the message!

misc/prost-codec/Cargo.toml
thomaseizinger: I guess that would only work if we move the `UviBytes` into `asynchronous-codec` as well.

thomaseizinger (approved):

misc/prost-codec/Cargo.toml
mxinden: > I guess that would only work if we move the `UviBytes` into `asynchronous-codec` as well.

Yes. That is the reason why I didn't move it there. Not sure whats best. I feel like using Uvi is a libp2p thing and thus should be inside the `rust-libp2p` repo. Given that it is a separate crate folks outside of libp2p can still use it though.

thomaseizinger: It would probably be cleaner to not have any specific codecs in `asynchronous-codec` and make it an interface-only crate. In that case it would be clear that all codecs live outside the crate :)

But, it is probably not worthwhile going through the effort of refactoring that now ...

misc/prost-codec/src/lib.rs
mxinden: Good point. You mentioned this before, though I seem to have forgotten it :( . Thanks for the pointer!

protocols/identify/CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
 
 - Update to `libp2p-swarm` `v0.36.0`.
 
+- Expose explicits errors via `UpgradeError` instead of generic `io::Error`.
+
```
mxinden:
```suggestion
- Expose explicits errors via `UpgradeError` instead of generic `io::Error`. See [PR 2630].

[PR 2630]: https://github.com/libp2p/rust-libp2p/pull/2630
```

misc/prost-codec/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+
+[dependencies]
+asynchronous-codec = { version = "0.6" }
+bytes = { version = "1" }
+prost = "0.10"
+thiserror = "1.0"
+unsigned-varint = { version = "0.7", features = ["asynchronous_codec"] }
+
+[dev-dependencies]
+prost-build = "0.10"
```
elenaf9:
```suggestion
prost-build = "0.10"

```
Nit: newline at end of file.

protocols/dcutr/src/protocol/inbound.rs
```diff
@@ -107,7 +110,7 @@ impl PendingConnect {
             self.substream
                 .next()
                 .await
-                .ok_or(super::codec::Error::Io(std::io::Error::new(
+                .ok_or(prost_codec::Error::Io(std::io::Error::new(
                     std::io::ErrorKind::UnexpectedEof,
                     "",
                 )))??;
```
elenaf9: Maybe also add for dcutr a `{inbound, outbound}::UpgradeError::StreamClosed` like it was done for identify?

protocols/identify/src/protocol.rs
```diff
@@ -184,77 +188,99 @@ where
...
 
     Ok(())
 }
 
-async fn recv<T>(mut socket: T) -> io::Result<IdentifyInfo>
+async fn recv<T>(mut socket: T) -> Result<IdentifyInfo, UpgradeError>
 where
     T: AsyncRead + AsyncWrite + Unpin,
 {
     socket.close().await?;
```
elenaf9, line 206: Just to help my understanding: Why do we close the socket here before creating the `FramedRead`, but don't do this in dcutr on `upgrade_inbound`?

mxinden: The standard identify exchange (not identify push) opens a stream to a remote, never writes on it, and expects the remote to send the response on said stream. Given that the local node never writes on the stream it opened it is safe to close it early. Not sure it is strictly necessary though.

(Also given that the local node never writes on the channel, it can use a `FramedRead` instead of a `Framed`.)

In DCUtR `upgrade_inbound` the remote node sends a request and the local node replies with a response. Closing the stream before creating the `Framed` would prevent the local node from being able to respond.

Does the above explanation make sense @elenaf9?

mxinden (commented):
Thanks for the reviews @thomaseizinger and @elenaf9. All comments should be addressed.

protocols/dcutr/src/protocol/inbound.rs
mxinden: Good idea. Done with the recent commit.

protocols/identify/src/protocol.rs
elenaf9: Yes makes sense, thanks!

elenaf9 (approved):

PR #2638, opened 2022-05-10
swarm/src/connection: Prioritize handler over negotiating streams

mxinden:
# Description

The `HandlerWrapper` polls three components:

1. `ConnectionHandler`
2. Outbound negotiating streams
3. Inbound negotiating streams

The `ConnectionHandler` itself might itself poll already negotiated streams.

By polling the three components above in the listed order one:

- Prioritizes local work and work coming from negotiated streams over
  negotiating streams.
- Prioritizes outbound negotiating streams over inbound negotiating
  streams, i.e. outbound requests over inbound requests.

Follows the general theme of https://github.com/libp2p/rust-libp2p/pull/2627 and https://github.com/libp2p/rust-libp2p/pull/2626.

## Change checklist

    - No change in functionality.
    - No change in functionality.


swarm/src/connection/handler_wrapper.rs
```diff
@@ -307,27 +307,57 @@ where
...
+            if let Poll::Ready(Some((user_data, res))) = self.negotiating_out.poll_next_unpin(cx) {
+                match res {
+                    Ok(upgrade) => self
+                        .handler
+                        .inject_fully_negotiated_outbound(upgrade, user_data),
+                    Err(err) => self.handler.inject_dial_upgrade_error(user_data, err),
+                }
+
+                // After the `inject_*` calls, the [`ConnectionHandler`] might be able to make progress.
+                continue;
```
thomaseizinger, line 353: I think we could also recurse here and thus avoid the loop? Not sure if that is necessarily easier to understand.

swarm/src/connection/handler_wrapper.rs
```diff
@@ -307,27 +307,57 @@ where
...
+                        let id = self.unique_dial_upgrade_id;
+                        let timeout = *protocol.timeout();
+                        self.unique_dial_upgrade_id += 1;
+                        let (upgrade, info) = protocol.into_upgrade();
+                        self.queued_dial_upgrades.push((id, SendWrapper(upgrade)));
+                        return Poll::Ready(Ok(Event::OutboundSubstreamRequest((
+                            id, info, timeout,
+                        ))));
+                    }
+                    ConnectionHandlerEvent::Close(err) => return Poll::Ready(Err(err.into())),
```
thomaseizinger: It is not immediately clear that all these branches will `return` because of the noise of constructing the return value. We could extract this into a separate function (`self.poll_handler(cx)`) and only have 1 `return` statement to make this more obvious. What do you think?

swarm/src/connection/handler_wrapper.rs
mxinden: I don't follow. What do you mean with "recurse"? As in a recursive call to the same method?

swarm/src/connection/handler_wrapper.rs
mxinden: What do you think of https://github.com/libp2p/rust-libp2p/pull/2638/commits/1de8039e853f45c248a2e860b898dcac50121499 @thomaseizinger?

swarm/src/connection/handler_wrapper.rs
thomaseizinger: Yeah. You should be able to just call `self.poll` here I believe?

swarm/src/connection/handler_wrapper.rs
thomaseizinger: Yeah that is better :)

swarm/src/connection/handler_wrapper.rs
mxinden: I have not seen recursion used on any serious Rust project. Not saying that it does not exist. Thus far I have avoided recursion in Rust due to:

- No Tail Call Ellimination
- Call stack limits
- Sensing that this is not idiomatic, thus not intuitive

Am I missing something? Do the above arguments not apply here?

elenaf9 (approved):
LGTM.
One question (not related to this PR but something I was wondering about while reviewing):
When asking the handler whether to keep the connection alive or not, in case of `KeepAlive::Until` it checks if the "until" timestamp is in the future. 
If it is, we set the `Shutdown::Later` accordingly. 
But in case that it is in the past (`checked_duration_since` returns `None`), we do not do anything. Shouldn't in that case the shutdown be set to `Shutdown::Asap`?
https://github.com/libp2p/rust-libp2p/blob/8f0f9a803ce26160892ce0f71db64431a33ef068/swarm/src/connection/handler_wrapper.rs#L385-L389
Am I missing something?

swarm/src/connection/handler_wrapper.rs
thomaseizinger: I think in this particular case, the recursion is very likely to stop after one call, is it not?

But I understand that this might be too much of a detail to notice and thus make the code harder to understand. Also, it is not guaranteed and thus risky to use with the usual problems of recursion.

Happy for the loop to stay, was just an idea :)

mxinden:
> One question (not related to this PR but something I was wondering about while reviewing): When asking the handler whether to keep the connection alive or not, in case of `KeepAlive::Until` it checks if the "until" timestamp is in the future. If it is, we set the `Shutdown::Later` accordingly. But in case that it is in the past (`checked_duration_since` returns `None`), we do not do anything. Shouldn't in that case the shutdown be set to `Shutdown::Asap`?
> 
> https://github.com/libp2p/rust-libp2p/blob/8f0f9a803ce26160892ce0f71db64431a33ef068/swarm/src/connection/handler_wrapper.rs#L385-L389

Not sure what the reasoning for this was. I could imagine that a `KeepAlive::Until` with a time in the past was considered a mistake. That said, I agree that `Shutdown::Asap` makes more sense. Want to create a patch @elenaf9?

PR #2639, opened 2022-05-10
identify/handler: Improve property name

elenaf9:
# Description

Nit: Rename `IdentifyHandler.next_id` to `IdentifyHandler.trigger_next_identify` to better reflect what it is actually doing.

Throughout rust-libp2p, we use `next_*id` for identifiers (e.g. `next_connection_id: ConnectionId`). 
`IdentifyHandler.next_id`is not an id, but instead a `Delay` that triggers the next outbound substream request, hence the old name is misleading.

## Change checklist

     - don't think a changelog entry is needed for this since it is only internally?

### DIFF
--- a/protocols/identify/src/handler.rs
+++ b/protocols/identify/src/handler.rs
@@ -50,12 +50,12 @@ pub struct IdentifyHandler {
     >,
 
     /// Future that fires when we need to identify the node again.
-    next_id: Delay,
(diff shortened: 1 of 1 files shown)


thomaseizinger (approved):
Thanks!

That is indeed a much better name :)

mxinden:
:+1: Thanks!

PR #2642, opened 2022-05-11
protocols/dcutr/example: Wait for relay to accept reservation request

mxinden:
# Description

When in listening mode, wait for the relay to accept our reservation
request. Only then can a client in dialing mode establish a relayed
connection to us via the relay.

Might help with https://github.com/libp2p/rust-libp2p/issues/2621#issuecomment-1123549348.

## Change checklist


mxinden:
//CC @foemass

mxinden:
@elenaf9 @thomaseizinger can one of you give this a review?

protocols/dcutr/examples/client.rs
```diff
@@ -197,25 +197,41 @@ fn main() -> Result<(), Box<dyn Error>> {
...
                 SwarmEvent::Dialing { .. } => {}
                 SwarmEvent::ConnectionEstablished { .. } => {}
                 SwarmEvent::Behaviour(Event::Ping(_)) => {}
+                SwarmEvent::Behaviour(Event::Relay(client::Event::ReservationReqAccepted {
+                    ..
+                })) => {
+                    info!("Relay accepted our reservation request.");
+                    relay_accepted_reservation = true
+                }
                 SwarmEvent::Behaviour(Event::Relay(_)) => {}
```
elenaf9: Maybe also catch the case of `client::Event::ReservationReqFailed` and return an error in that case. 
Shouldn't happen in practice but I'd rather have an explicit error than a non-terminating program. 
Alternatively, I think there is no valid case in which any other relay event should occur anyway/ So we could remove the match block here and let it fall to the blank panic below.

elenaf9 (approved):

protocols/dcutr/examples/client.rs
```diff
@@ -197,25 +197,41 @@ fn main() -> Result<(), Box<dyn Error>> {
...
-                    info!("Observed address: {:?}", observed_addr);
-                    break;
+                    info!("Relay observes us under the address: {:?}", observed_addr);
+                    learned_observed_addr = true;
                 }
                 event => panic!("{:?}", event),
             }
+
+            if learned_observed_addr
+                && (matches!(opts.mode, Mode::Dial) || relay_accepted_reservation)
```
thomaseizinger: Would this be easier to understand if we matched explicitly on `Mode` and returned `true` for dial and `relay_accepted_reservation` for listen?

thomaseizinger (approved):
LGTM

protocols/dcutr/examples/client.rs
mxinden: What do you think of https://github.com/libp2p/rust-libp2p/pull/2642/commits/75697c253d774294abdf2a00a40718f8a6d32493?

thomaseizinger: Much better, thank you!

PR #2643, opened 2022-05-11
src/bandwidth: Drop `atomic` dependency

hrxi:
Replace `atomic::Atomic<u64>` by `std::sync::atomic:AtomicU64`.

The original motivation of using `atomic::Atomic<u64>` instead of
`std`'s version in https://github.com/libp2p/rust-libp2p/pull/1670 was the following:

> I used the atomic crate and an Atomic<u64> because the AtomicU64 type
> isn't available on all architectures. The atomic crate automatically
> falls back to using a Mutex on platforms that don't support AtomicU64.

This argumentation is moot because the crate directly depends on
`libp2p-core` which also uses `AtomicU64`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -69,7 +69,6 @@ serde = ["libp2p-core/serde", "libp2p-kad/serde", "libp2p-gossipsub/serde"]
 all-features = true
 
 [dependencies]
-atomic = "0.5.0"
 bytes = "1"
 futures = "0.3.1"
 futures-timer = "3.0.2" # Explicit dependency to be used in `wasm-bindgen` feature
--- a/src/bandwidth.rs
+++ b/src/bandwidth.rs
@@ -26,7 +26,6 @@ use crate::{
     Multiaddr,
 };
 
-use atomic::Atomic;
 use futures::{
     io::{IoSlice, IoSliceMut},
     prelude::*,
@@ -36,7 +35,10 @@ use std::{
     convert::TryFrom as _,
     io,
     pin::Pin,
-    sync::{atomic::Ordering, Arc},
+    sync::{
+        atomic::{AtomicU64, Ordering},
+        Arc,
+    },
     task::{Context, Poll},
 };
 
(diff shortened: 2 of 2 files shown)


thomaseizinger (changes requested):
Thanks for the contribution!

1. Can you write a few lines about this change? i.e. is the `atomic` crate fully superseded by the std lib?
2. We don't have a lockfile checked into the repository because we are only developing a library here. Can you post the output of `cargo tree -i -p atomic` to show that we are actually dropping the dependency?
3. Rustfmt doesn't seem to be happy with the changes, please fix that :)

hrxi:
> 1. Can you write a few lines about this change? i.e. is the `atomic` crate fully superseded by the std lib?

Done.

> 2. We don't have a lockfile checked into the repository because we are only developing a library here. Can you post the output of `cargo tree -i -p atomic` to show that we are actually dropping the dependency?

```
$ git checkout pr_drop_atomic
Previous HEAD position was f04f6bb4 identify/handler: Improve property name (#2639)
Switched to branch 'pr_drop_atomic'
$ cargo tree -i -p atomic
error: package ID specification `atomic` did not match any packages

	Did you mean `tokio`?
$ git checkout HEAD~
HEAD is now at f04f6bb4 identify/handler: Improve property name (#2639)
$ cargo tree -i -p atomic | head
atomic v0.5.1
├── libp2p v0.45.0 (/home/u/nimiq/rust-libp2p)
│   [dev-dependencies]
│   ├── libp2p-autonat v0.4.0 (/home/u/nimiq/rust-libp2p/protocols/autonat)
│   │   └── libp2p v0.45.0 (/home/u/nimiq/rust-libp2p) (*)
│   ├── libp2p-dcutr v0.3.0 (/home/u/nimiq/rust-libp2p/protocols/dcutr)
│   │   ├── libp2p v0.45.0 (/home/u/nimiq/rust-libp2p) (*)
│   │   └── libp2p-metrics v0.6.0 (/home/u/nimiq/rust-libp2p/misc/metrics)
│   │       └── libp2p v0.45.0 (/home/u/nimiq/rust-libp2p) (*)
│   ├── libp2p-mdns v0.37.0 (/home/u/nimiq/rust-libp2p/protocols/mdns)
```

> 3. Rustfmt doesn't seem to be happy with the changes, please fix that :)

Done.

thomaseizinger (approved):

mxinden:
Thanks @hrxi!

PR #2644, opened 2022-05-11
Update `arrayvec` dependency to latest version 0.7.2

hrxi:


mxinden (approved):
Thanks for the upgrade! Help is is much appreciated.

I am fine merging here. I wonder whether the optimization (stack allocated bucket keys) is actually worth the complexity here. In case you would like to explore simplifications of this code, suggestions would be welcome!

Issue #2286, opened 2021-10-13
Metrics dependencies are always pulled via protocol features


thomaseizinger:
I am not sure if there is much we can do about it but I noticed that `libp2p-metrics` is automatically pulled in if any of the protocols that activate metrics features are activated.

When trying to upgrade to the most recent rust-libp2p HEAD, I saw `libp2p-metrics` come in as a dependency even though we are not specifying it as a feature. I suspect it is being pulled in by cargo due to `ping` activating `libp2p-metrics/ping`.

https://github.com/libp2p/rust-libp2p/blob/6d3ab8a3debe8d69dcd004173999732f12d0da96/Cargo.toml#L47

Is there a way of how we can design this differently? Perhaps having a dedicated `metrics` feature within each protocol would solve this?

mxinden:
Good catch!

Corresponding upstream tracking issue (_weak dependency feature_): https://github.com/rust-lang/cargo/issues/8832

As far as I can tell there is no solution to this today.

I don't it hurts much, given that `libp2p-metrics` is a pretty light-weight dependency. Still it would be great to not pull it in each time e.g. when the `ping` feature is set.

thomaseizinger:
Thanks for pointing me to that, I had no idea that this is being worked on! Very exciting :)

mxinden:
If I am not mistaken, this can be fixed with Rust `1.60`. Anyone interested in sending a patch?

https://blog.rust-lang.org/2022/04/07/Rust-1.60.0.html

maschad:
Hey @mxinden I've added a patch in this [PR](https://github.com/libp2p/rust-libp2p/pull/2646). Lmk what you think.

mxinden:
Closing here, since fixed by @maschad :tada:

PR #2646, opened 2022-05-12
Set Metrics as weak dependency

maschad:
# Description
Upgraded to [1.60.0](https://blog.rust-lang.org/2022/04/07/Rust-1.60.0.html) in order to enable metrics as a _weak dependency_  https://github.com/rust-lang/cargo/issues/8832

https://github.com/libp2p/rust-libp2p/issues/2286

## Change checklist


thomaseizinger (changes requested):
Concept ACK

I am okay with the semver bump but I think it should have a changelog entry plus we should use this syntax for all features!

mxinden:
:+1: Also in favor of this. Thanks @maschad!

Do I understand correctly that this forces downstream users to use Rust `>= v1.60.0`? I am fine with that, though to reduce conflicts, I suggest we merge here after the next release. (Which I am planning to prepare next week.)

maschad:
> 👍 Also in favor of this. Thanks @maschad!
> 
> Do I understand correctly that this forces downstream users to use Rust `>= v1.60.0`? I am fine with that, though to reduce conflicts, I suggest we merge here after the next release. (Which I am planning to prepare next week.)

I believe so seeing as it's the root manifest. @thomaseizinger that's a good idea, will do.

maschad:
I've made some changes. Look forward to your feedback post release of `0.45.0`

Cargo.toml
```diff
@@ -36,26 +36,26 @@ default = [
...
-dcutr = ["libp2p-dcutr", "libp2p-metrics/dcutr"]
+dcutr = ["libp2p-dcutr", "libp2p-metrics?/dcutr"]
 deflate = ["libp2p-deflate"]
 dns-async-std = ["libp2p-dns", "libp2p-dns/async-std"]
 dns-tokio = ["libp2p-dns", "libp2p-dns/tokio"]
 floodsub = ["libp2p-floodsub"]
-identify = ["libp2p-identify", "libp2p-metrics/identify"]
-kad = ["libp2p-kad", "libp2p-metrics/kad"]
-gossipsub = ["libp2p-gossipsub", "libp2p-metrics/gossipsub"]
+identify = ["dep:libp2p-identify", "libp2p-metrics?/identify"]
```
thomaseizinger, line 44: For consistency, I'd be in favor of using the `dep:` syntax for all features.

It has the benefit that all the additional feature flags - added by the mere fact that these are optional dependencies - are removed.

Cargo.toml
```diff
@@ -36,26 +36,26 @@ default = [
...
 plaintext = ["libp2p-plaintext"]
 pnet = ["libp2p-pnet"]
-relay = ["libp2p-relay", "libp2p-metrics/relay"]
+relay = ["dep:libp2p-relay", "libp2p-metrics?/relay"]
 request-response = ["libp2p-request-response"]
 rendezvous = ["libp2p-rendezvous"]
-tcp-async-io = ["libp2p-tcp", "libp2p-tcp/async-io"]
-tcp-tokio = ["libp2p-tcp", "libp2p-tcp/tokio"]
+tcp-async-io = ["dep:libp2p-tcp", "libp2p-tcp?/async-io"]
+tcp-tokio = ["dep:libp2p-tcp", "libp2p-tcp?/tokio"]
```
thomaseizinger, line 58: I think here, the `?` is unnecessary? We are activating a feature on the same dependency that we are activating.

thomaseizinger (changes requested):

Cargo.toml
maschad: If I am not mistaken that `dep:` syntax is implicitly defined whenever a dependency is marked as optional as explained [here](https://doc.rust-lang.org/cargo/reference/features.html#optional-dependencies). Perhaps I am misinterpreting what you mean.

Cargo.toml
maschad: Good catch, I overlooked this.

Cargo.toml
thomaseizinger: I am referring to this part of the docs:

> If you specify the optional dependency with the dep: prefix anywhere in the [features] table, that disables the implicit feature.

But reading it again now, I think I misunderstood how it works! I thought one had to use `dep:` everywhere to disable the implicit features!

maschad:
Thanks for your review @thomaseizinger I have made the requested changes.

CHANGELOG.md
```diff
@@ -43,6 +43,12 @@
 
 # `libp2p` facade crate
 
+# 0.45.1 [unreleased]
+- Semver bump Rust from `1.56.1` to `1.60.0` . See [PR 2646].
```
thomaseizinger: I am not sure if we have a written policy here but I think a minor version bump would be appropriate for a MSRV increase.

Cargo.toml
thomaseizinger: Can we verify that all the implicit features created by optional dependencies are gone?

You should be able to do that by making a new crate and try depending on this version of libp2p (probably best done with a `path` dependency) and see if features like `libp2p-ping` work.

CHANGELOG.md
maschad: I'm not able to find a written policy either but I would agree with you based on [Rust's  API Evolution's RFC](https://github.com/rust-lang/rfcs/blob/master/text/1105-api-evolution.md#principles-of-the-policy)

_In principle, it should be possible to produce a version of dependency code that will not break when upgrading other dependencies, or Rust itself, to a new minor revision._

Cargo.toml
maschad: Good idea, I ran this test and yes I can confirm that the optional dependencies are no longer pulled in.

thomaseizinger: Thank you :)

thomaseizinger (approved):
I think this is good to go!

`0.45.0` is not released yet so we will have to wait with merging this until that is in. Thank you for your effort and contribution :)

Cargo.toml
mxinden: Just to make sure I understand this correctly, given that we use `dep:` in the line above, implicit features of **all** optional dependencies are removed?

I am still a bit confused why we do it for `identify`, `ping` and `relay`, but not for `dcutr`, `kad` and `gossipsub`.

maschad:
I updated the Changelog again given the release of `0.45.0`

Cargo.toml
mxinden: Mentioning @maschad and @thomaseizinger to make sure this doesn't get lost. Accidentally posted on a resolved conversation.

thomaseizinger: > Just to make sure I understand this correctly, given that we use `dep:` in the line above, implicit features of **all** optional dependencies are removed?

Yes, that is my understanding from the documentation.

> I am still a bit confused why we do it for `identify`, `ping` and `relay`, but not for `dcutr`, `kad` and `gossipsub`.

Good find! I think we need/should use it there as well!

maschad: Good catch on `dcutr` :) but for `kad`  and `gossipsub` which are only enabled by the `serde` feature if we use `dep:` syntax then there would be no implicit features with the names `libp2p-kad` and ``libp2p-gossipsub` and so it becomes redundant (since they are already optional dependencies)

The docs are a bit confusing though, especially for people like myself who are newer to Rust.

thomaseizinger: > Good catch on `dcutr` :) but for `kad` and `gossipsub` which are only enabled by the `serde` feature if we use `dep:` syntax then there would be no implicit features with the names `libp2p-kad` and ``libp2p-gossipsub` and so it becomes redundant (since they are already optional dependencies)
> 
> The docs are a bit confusing though, especially for people like myself who are newer to Rust.

I don't quite understand. Shouldn't we also set `libp2p-gossipsub/serde` as a weak dependency to avoid `libp2p-gossipsub` being pulled if only the `serde` feature is activated? Same for `libp2p-kad`.

Perhaps it would be better to better to go back to the original proposal of using the `dep:` syntax everywhere for simplicity reasons.

maschad: I agree, that may make more sense in terms of a mental model of the dependency tree. I have updated the PR.

Cargo.toml
```diff
@@ -35,35 +35,35 @@ default = [
...
-relay = ["libp2p-relay", "libp2p-metrics/relay"]
-request-response = ["libp2p-request-response"]
-rendezvous = ["libp2p-rendezvous"]
-tcp-async-io = ["libp2p-tcp", "libp2p-tcp/async-io"]
-tcp-tokio = ["libp2p-tcp", "libp2p-tcp/tokio"]
-uds = ["libp2p-uds"]
+autonat = ["dep:libp2p-autonat"]
+dcutr = ["dep:libp2p-dcutr", "libp2p-metrics?/dcutr"]
+deflate = ["dep:libp2p-deflate"]
+dns-async-std = ["dep:libp2p-dns", "libp2p-dns?/async-std"]
```
thomaseizinger, line 41: Technically, setting `async-std` as a weak-dependency feature would not be necessary because we activate `libp2p-dns` anyway. But I also don't think it is harmful :)

thomaseizinger (approved):

mxinden:
Thanks for the continued work here! I will do a couple of patch releases off of `master` this week. Will merge this afterwards.

mxinden (approved):
Thanks @maschad and @thomaseizinger!

Issue #2635, opened 2022-05-09
*: Consider using semverver


mxinden:
## Description

Consider using https://github.com/rust-lang/rust-semverver to enforce semantic versioning. Instructions on how to integrate semverver in CI can be found on https://github.com/rust-lang/rust-semverver#ci-setup.

## Motivation

Complying with Semver makes it easier for users to consume rust-libp2p. Having an automated compliance check makes maintaining rust-libp2p easier.

## Are you planning to do it yourself in a pull request?

Help wanted!

maschad:
> ## Description
> Consider using [rust-lang/rust-semverver](https://github.com/rust-lang/rust-semverver) to enforce semantic versioning. Instructions on how to integrate semverver in CI can be found on [rust-lang/rust-semverver#ci-setup](https://github.com/rust-lang/rust-semverver#ci-setup).
> 
> ## Motivation
> Complying with Semver makes it easier for users to consume rust-libp2p. Having an automated compliance check makes maintaining rust-libp2p easier.
> 
> ## Are you planning to do it yourself in a pull request?
> Help wanted!

Hey @mxinden  I am giving this a shot. Please see this [PR](https://github.com/libp2p/rust-libp2p/pull/2647)

maschad:
Hey @mxinden I am giving this a shot. Please see this [PR](https://github.com/libp2p/rust-libp2p/pull/2647)

maschad:
Our current version bump results in a build failure see https://github.com/libp2p/rust-libp2p/pull/2647#issuecomment-1135208285

mxinden:
Instead of `semverver`, we could consider using https://crates.io/crates/cargo-semver-checks.

maschad:
Thanks for reviving this @mxinden - I had lost track of it because of work,  I agree, particularly because semverver is tightly coupled to specific nightly compiler versions and [takes work to stay in sync](https://github.com/rust-lang/rust-semverver/search?q=Rustup+to&type=commits). I will give it a try on the PR.

mxinden:
With https://github.com/libp2p/rust-libp2p/pull/2647 I am closing here.

PR #2647, opened 2022-05-12
.github/workflows: Enforce semver compliance with `cargo semver-checks`

maschad:
# Description

Complying with Semver makes it easier for users to consume rust-libp2p. 

https://github.com/libp2p/rust-libp2p/issues/2635

## Change checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -210,7 +210,52 @@ jobs:
         ALL_FEATURES=$(cargo metadata --format-version=1 --no-deps | jq -r '.packages[] | select(.name == "libp2p") | .features | keys | map(select(. != "full")) | sort | join(" ")')
         FULL_FEATURE=$(cargo metadata --format-version=1 --no-deps | jq -r '.packages[] | select(.name == "libp2p") | .features["full"] | sort | join(" ")')
         
+        test "$ALL_FEATURES = $FULL_FEATURE"
+
         echo "$ALL_FEATURES";
         echo "$FULL_FEATURE";
         
         test "$ALL_FEATURES" = "$FULL_FEATURE"
+
+  gather_crates_for_semver_checks:
+    runs-on: ubuntu-latest
+    outputs:
+      members: ${{ steps.cargo-metadata.outputs.members }}
+    steps:
+      - uses: actions/checkout@v3
+
+      - id: cargo-metadata
+        run: |
+          WORKSPACE_MEMBERS=$(cargo metadata --format-version=1 --no-deps | jq -c '.packages | .[] | select(.publish == null) | .name' | jq -s '.' | jq -c '.')
+          echo "::set-output name=members::${WORKSPACE_MEMBERS}"
+
+  semver-check:
+    runs-on: ubuntu-latest
+    needs: gather_crates_for_semver_checks
+    strategy:
+      fail-fast: false
+      matrix:
+        crate: ${{ fromJSON(needs.gather_crates_for_semver_checks.outputs.members) }}
+    steps:
+    - name: Cancel Previous Runs
+      uses: styfle/cancel-workflow-action@bb6001c4ea612bf59c3abfc4756fbceee4f870c7 # 0.10.0
+      with:
+        access_token: ${{ github.token }}
+
+    - uses: actions/checkout@v3
+
+    - uses: actions-rs/toolchain@16499b5e05bf2e26879000db0c1d13f7e13fa3af # v1.0.7
+      with:
+        profile: minimal
+        toolchain: stable
+        override: true
+
+    - name: Install Protoc
+      uses: arduino/setup-protoc@v1
+
+    - uses: Swatinem/rust-cache@6720f05bc48b77f96918929a9019fb2203ff71f8 # v2.0.0
+
+    - run: cargo install cargo-semver-checks
+    
+    - name: Semver Check
+      run: cargo semver-checks check-release -p ${{ matrix.crate }}
--- a/misc/keygen/Cargo.toml
+++ b/misc/keygen/Cargo.toml
@@ -7,6 +7,7 @@ license = "MIT"
 repository = "https://github.com/libp2p/rust-libp2p"
 keywords = ["peer-to-peer", "libp2p", "networking"]
 categories = ["network-programming", "asynchronous"]
+publish = false
 
 [dependencies]
 clap = {version = "3.1.6", features = ["derive"]}


.github/workflows/ci.yml
```diff
@@ -172,3 +172,26 @@ jobs:
...
+
+    - uses: actions-rs/toolchain@v1
+      with:
+        profile: minimal
+        toolchain: nightly-2022-04-20
+        override: true
+        components: rustc-dev,llvm-tools-preview
+
+    - name: Install semver
+      run: cargo install semver
```
thomaseizinger: This should be semverver!

.github/workflows/ci.yml
```diff
@@ -172,3 +172,26 @@ jobs:
...
+
+    - name: Fetch the version in the manifest
+      run: eval "current_version=$(grep -e '^version = .*$' Cargo.toml | cut -d ' ' -f 3)"
```
thomaseizinger: I think all this bash code will have to be one `run` step.

thomaseizinger (changes requested):
Thanks :)

I suggest you enable GH actions in your fork and iterate there, it is much quicker because you can force-push to master and don't have to wait for the CI approval.

maschad:
Good idea @thomaseizinger , thanks for the recommendation.

.github/workflows/ci.yml
maschad: Good point

maschad:
Currently this step fails with `version bump: 0.44.0 -> (breaking) -> 0.44.1` we should look into fixing that first before merging this.

thomaseizinger:
That is a lot of breaking changes that I am not recognising 😅

Also, I think @mxinden is planning on releasing the next version as 0.45 (see #2662). Where is semverver getting the 0.44.1 from?

In any case, strong concept ACK for adding this. Enforcing correct version bumps is super important IMO.

maschad:
I suspect the `0.44.1` was derived from the fact that the [most recent package published](https://crates.io/crates/libp2p/0.44.0) was `0.44.0` I tried changing the version to `0.44.1` in root manifest and also `1.0.0` and it still reports these changes as breaking. I will continue to debug.

thomaseizinger:
> I suspect the `0.44.1` was derived from the fact that the [most recent package published](https://crates.io/crates/libp2p/0.44.0) was `0.44.0` I tried changing the version to `0.44.1` in root manifest and also `1.0.0` and it still reports these changes as breaking. I will continue to debug.

Well, the fact that they are breaking will not change with the chosen version number :)

What should change is whether the tool reports it as a problem. With version `0.45.0`, it should hopefully say that the version bump is fine! Not sure how semverver does this but would have assumed that the exitcode changes?

maschad:
Good point, well then I am unsure why it says the changes are breaking. In the event there are no breaking changes,  outputs to a file which contains the current version i.e. `0.45.0`

mxinden:
I am not sure why the new CI step is failing, any ideas?

```
 version bump: 0.45.0 -> (breaking) -> 0.45.1
error: breaking changes in `development_transport`
   --> /home/runner/work/rust-libp2p/rust-libp2p/src/lib.rs:200:1
    |
200 | / pub async fn development_transport(
201 | |     keypair: identity::Keypair,
202 | | ) -> std::io::Result<core::transport::Boxed<(PeerId, core::muxing::StreamMuxerBox)>> {
    | |____________________________________________________________________________________^
    |
    = warning: type error: expected enum `old::identity::Keypair`, found enum `new::identity::Keypair` (breaking)
```

maschad:
> I am not sure why the new CI step is failing, any ideas?
> 
> ```
>  version bump: 0.45.0 -> (breaking) -> 0.45.1
> error: breaking changes in `development_transport`
>    --> /home/runner/work/rust-libp2p/rust-libp2p/src/lib.rs:200:1
>     |
> 200 | / pub async fn development_transport(
> 201 | |     keypair: identity::Keypair,
> 202 | | ) -> std::io::Result<core::transport::Boxed<(PeerId, core::muxing::StreamMuxerBox)>> {
>     | |____________________________________________________________________________________^
>     |
>     = warning: type error: expected enum `old::identity::Keypair`, found enum `new::identity::Keypair` (breaking)
> ```

Sorry I took so long to reply to this, was on vacation. My only hypothesis right now is that https://github.com/libp2p/rust-libp2p/commit/245b0563ab41a37b52d1d45ddb59f8706e130a53#diff-4ef9da87778def7bb67a144d06f9cc429d901994dd8b054812c14c1dac2b0df6R76-R78 introduced a new field and based on the current RFC [adding new fields to a variant in an `enum` is a **Major change**](https://github.com/rust-lang/rfcs/blob/master/text/1105-api-evolution.md#major-change-adding-new-fields-to-a-variant) _although_ a [postponed RFC](https://github.com/rust-lang/rfcs/pull/757) discusses a language feature that allows an `enum` to be marked as "extensible", which modifies the way that exhaustiveness checking is done and would make it possible to extend the `enum` without breakage.


Suffice to say, it may require quite a bit of backwork for this PR to be merged.I think despite the strong concept ACK, we may want to explore the opportunity cost of enforcing semver's current perspectives on breaking changes a bit more before beginning that work / moving ahead with adding these steps.

mxinden:
Thanks for digging into this once more @maschad.

> My only hypothesis right now is that [245b056#diff-4ef9da87778def7bb67a144d06f9cc429d901994dd8b054812c14c1dac2b0df6R76-R78](https://github.com/libp2p/rust-libp2p/commit/245b0563ab41a37b52d1d45ddb59f8706e130a53#diff-4ef9da87778def7bb67a144d06f9cc429d901994dd8b054812c14c1dac2b0df6R76-R78) introduced a new field and based on the current RFC [adding new fields to a variant in an `enum` is a **Major change**](https://github.com/rust-lang/rfcs/blob/master/text/1105-api-evolution.md#major-change-adding-new-fields-to-a-variant) _although_ a [postponed RFC](https://github.com/rust-lang/rfcs/pull/757) discusses a language feature that allows an `enum` to be marked as "extensible", which modifies the way that exhaustiveness checking is done and would make it possible to extend the `enum` without breakage.

Correct, https://github.com/libp2p/rust-libp2p/commit/245b0563ab41a37b52d1d45ddb59f8706e130a53#diff-4ef9da87778def7bb67a144d06f9cc429d901994dd8b054812c14c1dac2b0df6R76-R78 has been a breaking change, but it was released pre `v0.45.0`. `rustsemver` on the other hand complains about a breaking change between `v0.45.0` and `v0.45.1`:

>  version bump: 0.45.0 -> (breaking) -> 0.45.1

Am I missing something?

.github/workflows/ci.yml
```diff
@@ -172,3 +172,34 @@ jobs:
...
+    steps:
+
+     - name: Cancel Previous Runs
+       uses: styfle/cancel-workflow-action@a40b8845c0683271d9f53dfcb887a7e181d3918b # 0.9.1
+       with:
+        access_token: ${{ github.token }}
+
+     - uses: actions/checkout@v3
+
+     - name: Install Rust nightly-2022-05-16
```
```suggestion
     - name: Install Rust nightly-2022-06-20
```

.github/workflows/ci.yml
```diff
@@ -172,3 +172,34 @@ jobs:
...
+       uses: actions-rs/toolchain@v1
+       with:
+        profile: minimal
+        toolchain: nightly-2022-05-16
```
```suggestion
        toolchain: nightly-2022-06-20
```

.github/workflows/ci.yml
```diff
@@ -172,3 +172,34 @@ jobs:
...
+        override: true
+        components: rustc-dev,llvm-tools-preview
+
+     - name: Install CMake
+       run: sudo apt-get install -y cmake
+
+     - name: Install Semverver
+       run: cargo +nightly-2022-05-16 install --git https://github.com/rust-lang/rust-semverver
```
```suggestion
       run: cargo +nightly-2022-06-20 install --git https://github.com/rust-lang/rust-semverver
```

mxinden (commented):
rust-semver now recommends `nightly-2022-06-20`. That might solve the compilation errors on rust-semver itself.

thomaseizinger:
Random guess: Could these issues arise from our workspace and re-exporting crates?

mxinden:
Or potentially due to https://github.com/libp2p/rust-libp2p/pull/2720, as that imports another `libp2p-core` version which is incompatible.

thomaseizinger:
I think your theory is better :D

mxinden:
> Or potentially due to #2720, as that imports another `libp2p-core` version which is incompatible.

Unfortunately, this is not the case. Merged #2720 and updated here with latest `master`. Still failing.

mxinden:
I am out of ideas :/ Maybe worth involving the developers of rust-semverver here?

maschad:
That sounds worthwhile, I opened an [issue](https://github.com/rust-lang/rust-semverver/issues/325)

thomaseizinger:
There is a lot of reformatting going on. Can we undo that so the actual diff is easier to see?

Plus, as per your [comment](https://github.com/libp2p/rust-libp2p/issues/2635#issuecomment-1235715811), you are now experimenting with cargo-semver-checks, correct? I am in favor of that!

maschad:
Sorry about that, I did intend to investigate why my formatter was making all these changes, this was actually just a test @thomaseizinger , I had meant to push to my master to run my git actions but I hadn't worked on this in a while and forgotten that workflow.  I will convert this back to draft and only push again when this is ready for review.

.github/workflows/ci.yml
```diff
@@ -187,3 +187,28 @@ jobs:
...
+          profile: minimal
+
+      - name: Install Protoc
+        uses: arduino/setup-protoc@v1
+
+      - name: Install Semver Checks
+        run:  cargo install cargo-semver-checks
+
+      - name: Check Semver
+        run: cargo semver-checks check-release
```
thomaseizinger: Please add the newline at the end so this annoying icon goes away :D

.github/workflows/ci.yml
```diff
@@ -187,3 +187,28 @@ jobs:
...
+      - uses: actions/checkout@v3
+
+      - name: Install rust nightly
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: nightly
+          profile: minimal
+
+      - name: Install Protoc
+        uses: arduino/setup-protoc@v1
```
thomaseizinger: Do we need `protoc` for building the docs?

.github/workflows/ci.yml
```diff
@@ -187,3 +187,28 @@ jobs:
...
+
+      - name: Install Semver Checks
+        run:  cargo install cargo-semver-checks
```
thomaseizinger: Installing this on every CI run seems pretty wasteful. Can you add a cache please?

https://github.com/Swatinem/rust-cache

thomaseizinger (commented):
Very exciting! Thank you for working on this!

Why does the workflow say that we skipped 18 checks? https://github.com/libp2p/rust-libp2p/actions/runs/3074952477/jobs/4968044609#step:7:9

.github/workflows/ci.yml
maschad: Haha will do!

.github/workflows/ci.yml
maschad: Yup, otherwise we would get 

```
Error: Failed when running cargo-doc on....

failed to run custom build command for `libp2p-core v0.36.0 (/home/runner/work/rust-libp2p/rust-libp2p/core)`

Caused by:
  process didn't exit successfully: `/home/runner/work/rust-libp2p/rust-libp2p/target/semver-checks/target/debug/build/libp2p-core-ed4cd93bb4d7b68e/build-script-build` (exit status: 101)
  --- stderr
  thread 'main' panicked at '
  Could not find `protoc` installation and this build crate cannot proceed without
  this knowledge.
```

thomaseizinger:
The comments are resolved but no new code has been pushed. Did you forget that maybe? :)

maschad:
I had resolved them locally but I am still investigating why the `check-release` is skipping 18 checks.

maschad:
> Very exciting! Thank you for working on this!
> 
> Why does the workflow say that we skipped 18 checks? [libp2p/rust-libp2p/actions/runs/3074952477/jobs/4968044609#step:7:9](https://github.com/libp2p/rust-libp2p/actions/runs/3074952477/jobs/4968044609#step:7:9)

After digging through and switching over to [the author's own github action](https://github.com/obi1kenobi/cargo-semver-checks-action) I am inclined to think that the tests are being skipped since [this is a major version update](https://github.com/libp2p/rust-libp2p/actions/runs/3116777668/jobs/5054906175#step:7:876).

thomaseizinger:
> > Very exciting! Thank you for working on this!
> > Why does the workflow say that we skipped 18 checks? [libp2p/rust-libp2p/actions/runs/3074952477/jobs/4968044609#step:7:9](https://github.com/libp2p/rust-libp2p/actions/runs/3074952477/jobs/4968044609#step:7:9)
> 
> After digging through and switching over to [the author's own github action](https://github.com/obi1kenobi/cargo-semver-checks-action) I am inclined to think that the tests are being skipped since [this is a major version update](https://github.com/libp2p/rust-libp2p/actions/runs/3116777668/jobs/5054906175#step:7:876).

Okay, so because we are making a major update, it doesn't even bother to run the checks because anything is allowed anyway? Because of our workspace, I am inclined to say that we should actually run this check on all workspace members separately. Otherwise we won't necessarily learn a bad semver bump in one of the `libp2p-XYZ` crates.

We could do something like the following:

```yml
  gather_workspace_members:
    runs-on: ubuntu-latest
    outputs:
      members: ${{ steps.cargo-metadata.outputs.members }}
    steps:
      - uses: actions/checkout@v3
      - id: cargo-metadata
        run: |
          WORKSPACE_MEMBERS=$(cargo metadata --format-version=1 --no-deps | jq -c '.packages |.[] | .name' | jq -s | jq -c '.')
          echo "::set-output name=members::${WORKSPACE_MEMBERS}"
  semver_check:
    needs: gather_workspace_members
    strategy:
      matrix:
        crate: ${{ fromJson(needs.gather_workspace_members.outputs.members) }}
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v3
      - uses: obi1kenobi/cargo-semver-checks-action@v1
        with:
          crate-name: ${{ matrix.crate }}
          version-tag-prefix: v
```

maschad:
Thanks for that recommendation @thomaseizinger , so it turns out we have [some issues](https://github.com/maschad/rust-libp2p/actions/runs/3162236722/jobs/5148652564#step:7:14) to fix. This was on `dcutr`, looks like we removed a `pub enum variant` potentially in [this PR](https://github.com/libp2p/rust-libp2p/pull/2698/files#diff-2616480fabfd04daae9c6663797415112373156836109a3f9bbfac74e598fdb3L172-L173) perhaps we need to add the [non_exhaustive attribute](https://doc.rust-lang.org/reference/attributes/type_system.html#the-non_exhaustive-attribute) ?

thomaseizinger:
> Thanks for that recommendation @thomaseizinger , so it turns out we have [some issues](https://github.com/maschad/rust-libp2p/actions/runs/3162236722/jobs/5148652564#step:7:14) to fix. This was on `dcutr`, looks like we removed a `pub enum variant` potentially in [this PR](https://github.com/libp2p/rust-libp2p/pull/2698/files#diff-2616480fabfd04daae9c6663797415112373156836109a3f9bbfac74e598fdb3L172-L173) perhaps we need to add the [non_exhaustive attribute](https://doc.rust-lang.org/reference/attributes/type_system.html#the-non_exhaustive-attribute) ?

Glad to see it working. It might be worth to mark this job as `fail-fast: false` so we can actually see the results of all crates!

I've merged latest master in to resolve the merge conflicts and this includes a version bump for several crates so it seems to problem is fixed :)
I don't think there is much value in chasing that down, as long as we are making the correct bump now!

.github/workflows/ci.yml
```diff
@@ -188,6 +188,44 @@ jobs:
...
+      matrix:
+        crate: ${{ fromJSON(needs.gather_workspace_members.outputs.members) }}
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Install rust nightly
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: nightly
+          profile: minimal
```
thomaseizinger: Do we need to run on nightly? This will poison our cache effectively daily ...

.github/workflows/ci.yml
```diff
@@ -188,6 +188,44 @@ jobs:
...
+
+      - name: Install Protoc
+        uses: arduino/setup-protoc@v1
+
+      - run: cargo install cargo-semver-checks
+      
+      - uses: Swatinem/rust-cache@v2
+
+      - name: Semver Check
+        if: ${{ matrix.crate != 'libp2p' && contains(matrix.crate, 'libp2p') }}
```
thomaseizinger: What exactly is this doing? If we don't want to run the tool on certain crates, it is probably better to exclude them in the `jq` expression above!

.github/workflows/ci.yml
```diff
@@ -188,6 +188,44 @@ jobs:
...
+      - name: Install rust nightly
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: nightly
+          profile: minimal
+
+      - name: Install Protoc
+        uses: arduino/setup-protoc@v1
+
+      - run: cargo install cargo-semver-checks
```
thomaseizinger: We should do this after the cache action so we can cache this binary too!

.github/workflows/ci.yml
maschad: you're absolutely right, that makes much more sense.

.github/workflows/ci.yml
maschad: Good catch

.github/workflows/ci.yml
```diff
@@ -211,3 +211,42 @@ jobs:
...
+  gather_workspace_members:
+    runs-on: ubuntu-latest
+    outputs:
+      members: ${{ steps.cargo-metadata.outputs.members }}
+    steps:
+      - uses: actions/checkout@v3
+
+      - id: cargo-metadata
+        run: |
+          WORKSPACE_MEMBERS=$(cargo metadata --format-version=1 --no-deps | jq -c '.packages |.[] | .name'| grep 'libp2p-' | jq -s | jq -c '.')
```
thomaseizinger: Not sure how familiar you are with `jq` but it does have a filter mechanism (called `select`) as well, so we don't need to break up the expression only pipe it to `jq` again directly after :)

```suggestion
          WORKSPACE_MEMBERS=$(cargo metadata --format-version=1 --no-deps | jq -c '.packages |.[] | select(.name | startsWith("libp2p-")')
```

thomaseizinger: Really, the even better way would be to select them based on the `publish` value under `package` in `Cargo.toml`.

We want to check all semver compliance for all packages that we publish to crates.io. That f.e. includes `multistream-select` which isn't currently included in this.

I think the above should give you the necessary tools to get that done! Happy jq hacking 😁

.github/workflows/ci.yml
```diff
@@ -211,3 +211,42 @@ jobs:
         FULL_FEATURE=$(cargo metadata --format-version=1 --no-deps | jq -r '.packages[] | select(.name == "libp2p") | .features["full"] | sort | join(" ")')
         
         test "$ALL_FEATURES = $FULL_FEATURE"
+  gather_workspace_members:
```
thomaseizinger: This needs a better name to describe, which crates we select here (something with publish?). Otherwise someone might accidentially reuse this for another job.

.github/workflows/ci.yml
```diff
@@ -211,3 +211,42 @@ jobs:
...
+    runs-on: ubuntu-latest
+    needs: gather_workspace_members
+    strategy:
+      fail-fast: false
+      matrix:
+        crate: ${{ fromJSON(needs.gather_workspace_members.outputs.members) }}
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Install rust nightly
```
thomaseizinger: 1. Typo here.
2. Please use the same step here as in other jobs.

.github/workflows/ci.yml
```diff
@@ -211,3 +211,42 @@ jobs:
...
+      members: ${{ steps.cargo-metadata.outputs.members }}
+    steps:
+      - uses: actions/checkout@v3
+
+      - id: cargo-metadata
+        run: |
+          WORKSPACE_MEMBERS=$(cargo metadata --format-version=1 --no-deps | jq -c '.packages |.[] | .name'| grep 'libp2p-' | jq -s | jq -c '.')
+          echo "::set-output name=members::${WORKSPACE_MEMBERS}"
+
+  semver-check:
```
thomaseizinger, line 232: Please update this job to match the others: cancel previous etc.

.github/workflows/ci.yml
```diff
@@ -211,3 +211,42 @@ jobs:
...
+        with:
+          toolchain: stable
+          profile: minimal
+
+      - name: Install Protoc
+        uses: arduino/setup-protoc@v1
+
+      - uses: Swatinem/rust-cache@v2
+
+      - run: cargo install cargo-semver-checks
```
thomaseizinger: I remember there being issued with `cargo install` and trying to cache the binary. Please confirm using your own fork that the caching actually works here.

.github/workflows/ci.yml
```diff
@@ -211,3 +211,42 @@ jobs:
...
+      
+      - name: Semver Check
+        if: ${{ matrix.crate }}
```
thomaseizinger: That seems redundant? :)

thomaseizinger (changes requested):
Thanks! This is looking really good :)

A few minor corrections and then I think we can ship this!

.github/workflows/ci.yml
maschad: Thank you for being so detailed in your suggestions :) I'm not so familiar with `jq` so this is a very enlightening experience! Very cool. 

I don't think it would be wise to use the publish value considering they are all `null`-

```bash
cargo metadata --format-version=1 --no-deps | jq -c '.packages | .[] | .publish'
``` 
shows all `null` values -  which makes sense they were never given a value in the manifests, and the [publish field is actually used to prevent it  from being published to a package registry](https://doc.rust-lang.org/cargo/reference/manifest.html#the-publish-field).  

We could actually run `semver-checks` on all the packages but it will throw an error for new packages since there would be no available baseline versions f.e. with `keygen` or `quickcheck-ext`. Perhaps we should run it on all packages greater than `0.1.0` ?

.github/workflows/ci.yml
maschad: Perhaps gather crates for semver checks?

.github/workflows/ci.yml
maschad: Good catch! Thanks

.github/workflows/ci.yml
thomaseizinger: The reason why I proposed using `publish` is because we only care about semver in packages we publish :)

If you set it to `false` in keygen, does it show up then? It is a shame cargo doesn't autofill defaults for the  manifests. The default for `publish` is `true`. I'd be fine with having that logic in here and selecting all crates that have either `null` or `true` for their `publish` value.

maschad: I agree it is a shame, I opened [this issue](https://github.com/oli-obk/cargo_metadata/issues/207) to address it. In the interim we can select on a `null` value.

maschad: > We could actually run semver-checks on all the packages but it will throw an error for new packages since there would be no available baseline versions f.e. with keygen or quickcheck-ext. Perhaps we should run it on all packages greater than 0.1.0 ?

Also how do you think this is a reasonable suggestion?

thomaseizinger: > > We could actually run semver-checks on all the packages but it will throw an error for new packages since there would be no available baseline versions f.e. with keygen or quickcheck-ext. Perhaps we should run it on all packages greater than 0.1.0 ?
> 
> Also how do you think this is a reasonable suggestion?

I don't think it is a big problem that it throws an error on new packages. That is a temporary state when adding a new crate so it shouldn't happen all too often. @mxinden Do you think it is acceptable to have a failing job between merging the PR for a new crate and releasing it?

Selecting based on the version is a bit of a hack because it is not directly tied to the issue. Someone could add a package with version `0.0.1` if they want so if we were to go by version, we'd have to select based on `> 0.0.0` really (and have all sample / dev crates use `0.0.0`). Using `publish` is a more direct link to what we want to enforce: Not violating semver for packages we publish to `crates.io`.

maschad: Let me clarify, I was suggesting that if there is no `baseline version` for a crate to us to compare to, then we would avoid doing a `semver check` - _since there would be no version for us to check against._ 

This is not to suggest that we should select by `version` as an alternative to `publish` , but rather that we do this check per package that we publish.

Otherwise this would continuously fail on packages like `keygen` for example (which is over 3 years old) as it has only one version.

thomaseizinger: Ah right, I misunderstood you!

So we would build the matrix from all publishable crates but then check if we actually have a baseline version? How would we achieve that in particular? If the crate is published as `0.1.0` at the moment, we still want to check.

To me, this sounds like a missing feature in `cargo semver-checks`: We need a flag to have it not fail with missing baseline version to make it more CI friendly.

maschad: Good point, upon reconsideration I don't think there is a version detection that could indicate that there wasn't a semver violation in a package

I raised an issue on the repo https://github.com/obi1kenobi/cargo-semver-check/issues/146 

In the interim we could ignore the failures that are related to `No available baseline versions` or ignore those packages manually. What do you think @thomaseizinger ?

thomaseizinger: Given that they are all separate jobs, I think we can just have those fail for the time period between merging a new crate and releasing it :)

.github/workflows/ci.yml
maschad: It seems to be working! https://github.com/maschad/rust-libp2p/actions/runs/3215742329/jobs/5257056422#step:7:11

thomaseizinger: Nice :)

thomaseizinger (approved):
Nice!

You will need to set `publish` to false for `keygen` :)

maschad:
> Nice!
> 
> You will need to set `publish` to false for `keygen` :)

Thank you! Done :)

thomaseizinger:
I kicked CI again, was a flaky DNS resolve.

thomaseizinger:
@mxinden This is ready to merge from my end.

mxinden (approved):
Thanks @maschad for the continued work. Thanks @thomaseizinger for all the help. And of course thanks to the semver-checks maintainers (@obi1kenobi and @epage) for the tool!

Very excited for this!

thomaseizinger:
We currently have a failing semver check for 3 crates: https://github.com/libp2p/rust-libp2p/actions/runs/3254423615/jobs/5342659716

This is despite none of this code being touched since the last release. For `libp2p-core`, can this be related to us not building the docs with all features? https://docs.rs/libp2p-core/latest/libp2p_core/identity/enum.PublicKey.html doesn't mention the other variants. I guess this would improve with https://github.com/libp2p/rust-libp2p/pull/2983?

obi1kenobi:
If none of the code changed, then it's also possible this is a bug in cargo-semver-checks. I'll open a triage issue in the cargo-semver-checks repo and take a look as well when I have a bit of time.

thomaseizinger:
> If none of the code changed, then it's also possible this is a bug in cargo-semver-checks. I'll open a triage issue in the cargo-semver-checks repo and take a look as well when I have a bit of time.

Thank you! That is much appreciated!

obi1kenobi:
Confirmed false-positive bug in cargo-semver-checks, my apologies. obi1kenobi/cargo-semver-check#147 has the details. TL;DR, baseline data generated by looking up a crate version on crates.io incorrectly only includes default features, so the variants from `--all-features` look like additions.

The other baseline options (git rev / path / pre-generated rustdoc file) are not affected: for example, `cargo semver-checks check-release -p libp2p-core --baseline-rev libp2p-core-v0.37.0` finds no errors. Definitely not ideal, but hopefully it can be a workaround while we fix the bug.

Unrelated, `libp2p-core-v0.37.0` is where this repo's git tag pattern first started to include a `v` before the version: compare `libp2p-core-v0.37.0` vs `libp2p-core-0.36.0`. Just a thing I noticed and thought you might want to know.

thomaseizinger:
> Confirmed false-positive bug in cargo-semver-checks, my apologies. [obi1kenobi/cargo-semver-check#147](https://github.com/obi1kenobi/cargo-semver-check/issues/147) has the details. TL;DR, baseline data generated by looking up a crate version on crates.io incorrectly only includes default features, so the variants from `--all-features` look like additions.
> 
> The other baseline options (git rev / path / pre-generated rustdoc file) are not affected: for example, `cargo semver-checks check-release -p libp2p-core --baseline-rev libp2p-core-v0.37.0` finds no errors. Definitely not ideal, but hopefully it can be a workaround while we fix the bug.

Excellent finding, thank you for that!

> Unrelated, `libp2p-core-v0.37.0` is where this repo's git tag pattern first started to include a `v` before the version: compare `libp2p-core-v0.37.0` vs `libp2p-core-0.36.0`. Just a thing I noticed and thought you might want to know.

I think this is because @mxinden used `cargo release` for this release. We are planning on using that moving forward so we'll either need to adjust to that or fix how we use of `cargo release`.

mxinden:
First off, thanks @obi1kenobi for stepping in!

> Unrelated, `libp2p-core-v0.37.0` is where this repo's git tag pattern first started to include a `v` before the version: compare `libp2p-core-v0.37.0` vs `libp2p-core-0.36.0`. Just a thing I noticed and thought you might want to know.

That is my bad. Sorry for not making this change explicit.

> I think this is because @mxinden used `cargo release` for this release. We are planning on using that moving forward so we'll either need to adjust to that or fix how we use of `cargo release`.

In case this only requires a change on rust-libp2p, I would prefer the `cargo-release` way, i.e. `libp2p-core-v0.0.0` with the `v`.

obi1kenobi:
No worries on the tag format change. It's all the same to both me and to `cargo-semver-checks` — I just wanted to flag it to make sure you know about it, in case it was an unintentional change.

~~Btw, I'm testing a new `cargo-semver-checks` update with a fix for this bug, and I seem to have run into a problem: it appears that dependency resolution fails on `libp2p-core v0.37.0` with all features enabled.~~ **EDIT:** Resolved after deleting `Cargo.lock`, see next post.

To reproduce, make a new library project and add the following to its `Cargo.toml`:
```toml
[dependencies]
libp2p-core = { version = '=0.37.0', features = ["secp256k1", "ecdsa", "rsa", "serde"] }
```

The above is how `cargo-semver-checks` generates rustdoc for projects when looking up a version number from the index (crates.io). This is what it sees when it attempts to generate a valid lockfile using which to generate the rustdoc JSON:
```
$ cargo semver-checks check-release -p libp2p-core
    Updating index
     Parsing libp2p-core v0.37.0 (current)
     Parsing libp2p-core v0.37.0 (baseline)
Error: Failed when running cargo-doc on /.../rust-libp2p/target/semver-checks/registry-libp2p_core-0_37_0/Cargo.toml: error: failed to select a version for `zeroize`.
    ... required by package `elliptic-curve v0.12.1`
    ... which satisfies dependency `elliptic-curve = "^0.12.1"` of package `p256 v0.11.1`
    ... which satisfies dependency `p256 = "^0.11.1"` of package `libp2p-core v0.37.0`
    ... which satisfies dependency `libp2p-core = "=0.37.0"` (locked to 0.37.0) of package `rustdoc v0.0.0 (/.../rust-libp2p/target/semver-checks/registry-libp2p_core-0_37_0)`
versions that meet the requirements `^1.5` are: 1.5.7, 1.5.6, 1.5.5, 1.5.4, 1.5.3

all possible versions conflict with previously selected packages.

  previously selected package `zeroize v1.3.0`
    ... which satisfies dependency `zeroize = ">=1, <1.4"` (locked to 1.3.0) of package `curve25519-dalek v3.2.1`
    ... which satisfies dependency `curve25519-dalek = "^3"` (locked to 3.2.1) of package `ed25519-dalek v1.0.1`
    ... which satisfies dependency `ed25519-dalek = "^1.0.1"` (locked to 1.0.1) of package `libp2p-core v0.37.0`
    ... which satisfies dependency `libp2p-core = "=0.37.0"` (locked to 0.37.0) of package `rustdoc v0.0.0 (/.../rust-libp2p/target/semver-checks/registry-libp2p_core-0_37_0)`

failed to select a version for `zeroize` which could resolve this conflict
```

Not sure why this is happening, or why the test suite didn't catch it already — I noticed that an `--all-features` test is included in the suite.

Perhaps there's something wrong with my machine? Would love some help if you have a moment, since this is blocking me from fixing obi1kenobi/cargo-semver-check#147. Happy to move the discussion to that issue, if you'd prefer.

obi1kenobi:
Huh, perhaps a glitch in (or some mismatch in my mental model for) cargo?

I deleted the `Cargo.lock` file and it regenerated just fine after that.

epage:
And feel free to ping me about cargo-release questions.  I am in the middle of a new version where the goal is to make the workspace experience even better.

mxinden:
I am not sure I am following. What input would be helpful from our end @obi1kenobi ? Does the "comment has been hidden" that you are no longer facing the issue?

@thomaseizinger what shall we do in the meantime? I suggest we ignore the CI failure on changes that are clearly not related. Objections?

mxinden:
> And feel free to ping me about cargo-release questions. I am in the middle of a new version where the goal is to make the workspace experience even better.

Thanks a lot @epage. Based on the little experience I have with cargo-release, everything works as expected and in an intuitive way.

obi1kenobi:
> I am not sure I am following. What input would be helpful from our end @obi1kenobi ? Does the "comment has been hidden" that you are no longer facing the issue?

Yeah, all sorted out. It was something due to cargo (either a bug or a wrong mental model on my part), and deleting the lockfile solved it.

I'm onboarding a couple of new contributors on the `cargo-semver-checks` and this fix is one of their onboarding tasks. I think we should have a fix for you at some point soon (~2-3 weeks max, hopefully much much sooner).

> @thomaseizinger what shall we do in the meantime? I suggest we ignore the CI failure on changes that are clearly not related. Objections?

The `check-release` command is designed to run right before `cargo publish`, so if running it on every PR, it's probably a good idea to allow failures and just review its findings. This is because if `check-release` sees that the current version and the baseline (last published) version are an exact match, it'll conservatively assume the new code is just planned to be a patch release — which will[0] fail if new APIs are added (minor version) or if breaking changes are made (major version).

[0]: It won't actually fail today because we don't have those lints yet, but we will have them eventually.

In the future, I'd like to add a "report findings" mode in addition to `check-release` that would be even better suited to running on every PR. But we aren't there yet.

thomaseizinger:
> @thomaseizinger what shall we do in the meantime? I suggest we ignore the CI failure on changes that are clearly not related. Objections?

Works for me!

obi1kenobi:
> Confirmed false-positive bug in cargo-semver-checks, my apologies. [obi1kenobi/cargo-semver-checks#147](https://github.com/obi1kenobi/cargo-semver-checks/issues/147) has the details. TL;DR, baseline data generated by looking up a crate version on crates.io incorrectly only includes default features, so the variants from `--all-features` look like additions.

The false-positive bug is fixed in v0.13.0 of `cargo-semver-checks` which is now on `crates.io`, and should automatically get picked up by your CI setup here: https://github.com/libp2p/rust-libp2p/blob/master/.github/workflows/ci.yml#L220

In addition, we've added a regression test in `cargo-semver-checks` that uses the problematic commit in your repo in order to ensure this is the last we see of this bug: https://github.com/obi1kenobi/cargo-semver-checks/blob/main/.github/workflows/ci.yml#L118-L126

Thank you for your patience, it's been a pleasure to have you as some of `cargo-semver-checks`' earliest adopters 😊

thomaseizinger:
> > Confirmed false-positive bug in cargo-semver-checks, my apologies. [obi1kenobi/cargo-semver-checks#147](https://github.com/obi1kenobi/cargo-semver-checks/issues/147) has the details. TL;DR, baseline data generated by looking up a crate version on crates.io incorrectly only includes default features, so the variants from `--all-features` look like additions.
> 
> The false-positive bug is fixed in v0.13.0 of `cargo-semver-checks` which is now on `crates.io`, and should automatically get picked up by your CI setup here: [https://github.com/libp2p/rust-libp2p/blob/master/.github/workflows/ci.yml#L220](https://github.com/libp2p/rust-libp2p/blob/master/.github/workflows/ci.yml?rgh-link-date=2022-11-16T15%3A28%3A31Z#L220)

Awesome, thank you!

> Thank you for your patience, it's been a pleasure to have you as some of `cargo-semver-checks`' earliest adopters blush

Thank you for making and improving it :)

mxinden:
Our `cargo semver-checks` CI steps currently fail for `libp2p-dcutr` and `libp2p-relay`, even though neither of them merged any changes since the last [`libp2p` `v0.50.0` release](https://github.com/libp2p/rust-libp2p/releases/tag/v0.50.0).

- `libp2p-dcutr` https://github.com/libp2p/rust-libp2p/actions/runs/3576177811/jobs/6013685638
- `libp2p-relay` https://github.com/libp2p/rust-libp2p/actions/runs/3576177811/jobs/6013686091

@thomaseizinger @obi1kenobi any ideas why this is failing?

obi1kenobi:
Confirmed bug, thanks for reporting. Will open an issue in `cargo-semver-checks` and triage further.

As usual, we'll add those packages to our regression suite (here's the entry for [the previous bug you reported](https://github.com/obi1kenobi/cargo-semver-checks/blob/main/.github/workflows/ci.yml#L92-L133)) so this problem doesn't happen again after it's fixed.

obi1kenobi:
> Our `cargo semver-checks` CI steps currently fail for `libp2p-dcutr` and `libp2p-relay`, even though neither of them merged any changes since the last [`libp2p` `v0.50.0` release](https://github.com/libp2p/rust-libp2p/releases/tag/v0.50.0).

Unfortunately, I found that this is because of a combination of renaming re-exports causing confusion between multiple enums with the same name. For example, a `pub enum UpgradeError` exists in all of:
```
protocols/dcutr/src/protocol/outbound.rs
protocols/dcutr/src/protocol/inbound.rs
protocols/dcutr/src/behavior.rs
```
The false-positives happen because the lint tries to match up all the variants across different enums, which obviously isn't going to do the right thing.

`cargo-semver-checks` doesn't currently handle re-export renames (and re-exports are already tricky because of the possibility of having infinitely many re-exported names for each item: https://github.com/obi1kenobi/cargo-semver-checks/issues/152 ), so this isn't a trivial fix.

That said, it's still important to fix and this should be sufficient motivation to come up with a permanent solution to the re-exports problem :)

Thanks for bearing with me on this.

thomaseizinger:
@mxinden How should we proceed in the meantime? We can either:

- Bump the version to make `cargo semver-checks` happy.
- Exclude it from running for those two crates.
- Any other ideas?

I'd like to use our mergify automation again for merging PRs but that one doesn't work as long as our CI is red.

obi1kenobi:
If you're open to tweaking the code a bit, I think it should be possible to work around the `cargo-semver-checks` false-positive without making a breaking change.

The issue tripping up `cargo-semver-checks` is the renaming re-exports:
https://github.com/libp2p/rust-libp2p/blob/master/protocols/dcutr/src/lib.rs#L31

Renaming `inbound::UpgradeError` to `InboundUpgradeError`, and `outbound::UpgradeError` to `OutboundUpgradeError` (i.e. the same names they are being renamed to in the re-export) should avoid the false-positive problem. I believe this isn't even a breaking change: we'd just be removing the renaming re-export and replacing it with a re-export without rename.

If the above isn't something you'd like to do, then the next best thing would be excluding those two crates from the `cargo-semver-checks` runs for now. I don't believe that bumping the version will prevent the false-positive from coming back.

I'm also trying to work out a proper fix for `cargo-semver-checks`, but obviously that's a project with a longer timescale than needed here.

thomaseizinger:
Thanks! That is helpful!

I am definitely open to tweaking the code. Do you think type aliases would work?

obi1kenobi:
Unfortunately I don't think they would, since I don't believe they are reflected in the rustdoc JSON data.

Happy to review PRs or write one together with you if that'd be helpful.

thomaseizinger:
I've opened a PR here with your suggestion, let's see if it works: https://github.com/libp2p/rust-libp2p/pull/3213

thomaseizinger:
Your proposed fix works but my planned deprecation of other symbols that I wanted to do in one go doesn't :(

```
❯ cargo semver-checks check-release --package libp2p-dcutr
    Updating index
     Parsing libp2p-dcutr v0.8.1 (current)
     Parsing libp2p-dcutr v0.8.0 (baseline)
    Checking libp2p-dcutr v0.8.0 -> v0.8.1 (minor change)
   Completed [   0.076s] 22 checks; 20 passed, 2 failed, 0 unnecessary

--- failure enum_missing: pub enum removed or renamed ---

Description:
A publicly-visible enum cannot be imported by its prior path. A `pub use` may have been removed, or the enum itself may have been renamed or removed entirely.
        ref: https://doc.rust-lang.org/cargo/reference/semver.html#item-remove
       impl: https://github.com/obi1kenobi/cargo-semver-check/tree/v0.13.1/src/queries/enum_missing.ron

Failed in:
  enum libp2p_dcutr::UpgradeError, previously in file /home/thomas/.cargo/registry/src/github.com-1ecc6299db9ec823/libp2p-dcutr-0.8.0/src/protocol/inbound.rs:120
  enum libp2p_dcutr::behaviour::UpgradeError, previously in file /home/thomas/.cargo/registry/src/github.com-1ecc6299db9ec823/libp2p-dcutr-0.8.0/src/behaviour.rs:62
  enum libp2p_dcutr::behaviour::Event, previously in file /home/thomas/.cargo/registry/src/github.com-1ecc6299db9ec823/libp2p-dcutr-0.8.0/src/behaviour.rs:43
  enum libp2p_dcutr::UpgradeError, previously in file /home/thomas/.cargo/registry/src/github.com-1ecc6299db9ec823/libp2p-dcutr-0.8.0/src/protocol/outbound.rs:118

--- failure struct_missing: pub struct removed or renamed ---

Description:
A publicly-visible struct cannot be imported by its prior path. A `pub use` may have been removed, or the struct itself may have been renamed or removed entirely.
        ref: https://doc.rust-lang.org/cargo/reference/semver.html#item-remove
       impl: https://github.com/obi1kenobi/cargo-semver-check/tree/v0.13.1/src/queries/struct_missing.ron

Failed in:
  struct libp2p_dcutr::behaviour::Behaviour, previously in file /home/thomas/.cargo/registry/src/github.com-1ecc6299db9ec823/libp2p-dcutr-0.8.0/src/behaviour.rs:70
       Final [   0.081s] semver requires new major version: 2 major and 0 minor checks failed
```

thomaseizinger:
I've split the renaming into a separate PR, so we can merge the fix in isolation: https://github.com/libp2p/rust-libp2p/pull/3214

I am not sure if it is the same underlying issue but supporting type aliases is more important. They allow us to deprecate APIs which we use quite extensively. I am not a fan of renaming imports myself so if you have to pick which one to implement first, I'd vote for type aliases :)

obi1kenobi:
Thanks for the suggestion, definitely very valuable feedback! I opened https://github.com/obi1kenobi/cargo-semver-checks/issues/215 for that. I'm hopeful that reworking the import-handling code might help resolve both issues at once 🤞

obi1kenobi:
I'm putting the finishing touches on https://github.com/obi1kenobi/cargo-semver-checks/pull/330 as a fix for the re-export and typedef false-positives, and if all goes well it should be part of cargo-semver-checks 0.17 🤞

Thanks for your patience!

thomaseizinger:
> I'm putting the finishing touches on [obi1kenobi/cargo-semver-checks#330](https://github.com/obi1kenobi/cargo-semver-checks/pull/330) as a fix for the re-export and typedef false-positives, and if all goes well it should be part of cargo-semver-checks 0.17 crossed_fingers
> 
> Thanks for your patience!

Awesome work! Looking forward to seeing this in a release!

obi1kenobi:
https://github.com/obi1kenobi/cargo-semver-checks/releases/tag/v0.17.0 is out now 🚀

If you notice any re-export bugs after upgrading (or any other bugs, really), please ping me and I'll fix them ASAP.

PR #2622, opened 2022-04-26
feat: Add WebRTC transport

melekes:
## Description

Hey 👋 This is a WebRTC transport implemented in accordance w/ the [spec](https://github.com/libp2p/specs/pull/412). It's based on the [webrtc-rs](https://github.com/webrtc-rs/webrtc) library.

Resolves: #1066.

## Links to any relevant issues

- https://github.com/libp2p/rust-libp2p/issues/1066
- https://github.com/libp2p/specs/issues/220
- https://github.com/paritytech/smoldot/issues/1712

## Change checklist

- [x] I have performed a self-review of my own code
- [x] I have made corresponding changes to the documentation
- [x] I have added tests that prove my fix is effective or that my feature works
- [ ] A changelog entry has been made in the appropriate crates

Issue #1389, opened 2020-01-10
Finish the work related to Pin/Unpin


tomaka:
After #1384, there are still a lot of things to do:

- [x] #1329
- [ ] Maybe part of #1329, but pass `Pin<&mut Self>`  to `NetworkBehaviour::poll`/`ProtocolsHandler::poll`/`NodeHandler::poll`.
- [x] Change all the methods of `StreamMuxer` to require `Pin<&mut Self>`. In practice, the muxer itself is already pinned as it is within an `Arc`.
- [ ] Maybe remove all the `Unpin` trait requirements that remain everywhere? There is one major problematic area: when we pass a substream from multistream-select to the actual handler, we move it. I'm not really sure what to do here.

tomaka:
About the `StreamMuxer`:

Ideally, the `Substream`s and `OutboundSubstream`s that the muxer produces would be pinned as well. However we have the `destroy_substream` and `destroy_outbound` methods that require being given ownership of the substream/outboundsubstream. This would require unpinning them.

I can see two options:

- Don't pin `Substream`/`OutboundSubstream`, which is a shame because they mimic the `AsyncRead`/`AsyncWrite` and `Future` traits respectively, which accept a pinned version of them.

- Modify `destroy_substream` and `destroy_outbound` to take a `&mut Substream` and `&mut OutboundSubstream` respectively, rather than being given ownership.

PR #2648, opened 2022-05-15
core/muxing: Redesign `StreamMuxer` trait

thomaseizinger:
# Description

This PR re-designs the `StreamMuxer` trait to:

1. Use `&mut self` for all methods
2. Follow the style of "mutate state and return result via `poll`", same as we do for many other parts of `rust-libp2p`

## Splitting out separate PRs

- [x] https://github.com/libp2p/rust-libp2p/pull/2664
- [x] https://github.com/libp2p/rust-libp2p/pull/2685
- [x] https://github.com/libp2p/rust-libp2p/pull/2669
- [x] https://github.com/libp2p/rust-libp2p/pull/2665
- [x] https://github.com/libp2p/rust-libp2p/pull/2666
- [x] https://github.com/libp2p/rust-libp2p/pull/2667
- [x] https://github.com/libp2p/rust-libp2p/pull/2668
- [x] https://github.com/libp2p/rust-libp2p/pull/2677
- [x] https://github.com/libp2p/rust-yamux/pull/135
- [x] https://github.com/libp2p/rust-libp2p/pull/2688
- [x] https://github.com/libp2p/rust-libp2p/pull/2691
- [x] https://github.com/libp2p/rust-libp2p/pull/2703
- [x] https://github.com/libp2p/rust-libp2p/pull/2706
- [x] https://github.com/libp2p/rust-libp2p/pull/2707
- [x] https://github.com/libp2p/rust-libp2p/pull/2710

## Additional tasks

- [x] Retain `Substream` associated type on libp2p-core level
- [x] Fix mplex performance regression (use channels of `Bytes`/`[u8; X]`?)
- [ ] Make `StreamMuxer` use `Pin<&mut self>`
- [ ] ~Include "first close, then flush" logic within `SubstreamBox`~
- [ ] ~Release `yamux` once https://github.com/libp2p/rust-yamux/pull/135 is merged~
- [ ] ~Update PR with latest yamux release~
- [x] Drop b6c91d138b2d4078c2ba08b436256b4d035d59ab from PR now that cec117667b3e0a23b8acb3acbc342e087ef8df6b is landed and the bug is fixed.
- [ ] Write a more detailed changelog entry for `libp2p-yamux`.

## Open Questions

- ~Is there a better name for `AsyncReadWriteBox`?~

## Change checklist


muxers/yamux/src/lib.rs
```diff
@@ -84,116 +85,69 @@ where
...
-    ) -> Poll<YamuxResult<usize>> {
-        Pin::new(s)
-            .poll_write(c, b)
-            .map_err(|e| YamuxError(e.into()))
-    }
+        let result = ready!(self.incoming.poll_next_unpin(c))
+            .transpose()
+            .map_err(to_io_error)?
+            .map(|stream| StreamMuxerEvent::InboundSubstream(AsyncReadWriteBox::new(stream)))
+            .ok_or_else(|| to_io_error(yamux::ConnectionError::Closed));
```
thomaseizinger: Can you `?` here.

muxers/yamux/src/lib.rs
```diff
@@ -84,116 +85,69 @@ where
...
-        match ready!(inner.incoming.poll_next_unpin(c)) {
-            Some(Ok(s)) => Poll::Ready(Ok(StreamMuxerEvent::InboundSubstream(s))),
-            Some(Err(e)) => Poll::Ready(Err(e)),
-            None => Poll::Ready(Err(yamux::ConnectionError::Closed.into())),
+    fn open_outbound(&mut self) -> SubstreamId {
+        self.substreams.open_outbound()
+    }
+
+    fn poll(&mut self, c: &mut Context<'_>) -> Poll<io::Result<StreamMuxerEvent>> {
+        match self
```
thomaseizinger: This could be an `if let Poll::Ready`.

swarm/src/connection.rs
```diff
@@ -79,7 +85,7 @@ where
 {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         f.debug_struct("Connection")
-            .field("muxing", &self.muxing)
+            // .field("muxing", &self.muxing)
```
thomaseizinger: Put this back in.

thomaseizinger (commented):
Things found in self-review.

mxinden:
Great. I meant to touch this code part since some time. I will take a look. Thanks @thomaseizinger for proposing the patch set.

thomaseizinger:
> Great. I meant to touch this code part since some time. I will take a look. Thanks @thomaseizinger for proposing the patch set.

That is what happens when you get bored on a plane 😂

I have a few more patches locally that should fix the build and do a bit more tidy-up! Will push shortly!

thomaseizinger:
@mxinden @elenaf9 In this PR, I inlined the `Muxing` type into the `Connection` struct. Both because it felt like an unnecessary abstraction and because some of the trait bounds were in the way of what I wanted to do :D 

I am tempted to do the same thing for `HandlerWrapper`.

1. The name is not great which indicates that it should probably not exist.
2. It is only used once: within `Connection`. I don't see it being useful anywhere else and we can employ various other techniques like extracting functions etc to retain the readability/maintainability of `Connection`.
3. I think some of the logic would be easier to understand if we directly control the `StreamMuxer` and the `ConnectionHandler` within one type.

What do you think? I am happy to do it as a follow-up as well, doesn't have to be in this PR.

thomaseizinger:
Okay, this is actually ready for review now, no more commits I promise :D

thomaseizinger:
As part of fixing the benchmarks, I ran them against `master` and this branch and unfortunately, there seems to be some severe performance regressions in mplex:

```
Gnuplot not found, using plotters backend
tcp/256                 time:   [5.3159 ms 5.3836 ms 5.4592 ms]
                        thrpt:  [183.18 MiB/s 185.75 MiB/s 188.11 MiB/s]
                 change:
                        time:   [-0.2956% +2.5125% +5.8465%] (p = 0.10 > 0.05)
                        thrpt:  [-5.5236% -2.4509% +0.2965%]
                        No change in performance detected.
Found 3 outliers among 100 measurements (3.00%)
  1 (1.00%) high mild
  2 (2.00%) high severe
tcp/512                 time:   [3.4658 ms 3.5983 ms 3.7190 ms]
                        thrpt:  [268.89 MiB/s 277.91 MiB/s 288.53 MiB/s]
                 change:
                        time:   [-13.413% -5.8873% +2.6998%] (p = 0.18 > 0.05)
                        thrpt:  [-2.6289% +6.2556% +15.491%]
                        No change in performance detected.
Found 15 outliers among 100 measurements (15.00%)
  13 (13.00%) low severe
  2 (2.00%) high mild
tcp/1024                time:   [3.2309 ms 3.3600 ms 3.4761 ms]
                        thrpt:  [287.68 MiB/s 297.62 MiB/s 309.51 MiB/s]
                 change:
                        time:   [+6.6190% +15.824% +26.712%] (p = 0.00 < 0.05)
                        thrpt:  [-21.081% -13.662% -6.2081%]
                        Performance has regressed.
Found 7 outliers among 100 measurements (7.00%)
  6 (6.00%) low severe
  1 (1.00%) high mild
tcp/8192                time:   [2.8161 ms 2.8594 ms 2.9051 ms]
                        thrpt:  [344.22 MiB/s 349.73 MiB/s 355.09 MiB/s]
                 change:
                        time:   [+44.152% +57.258% +71.712%] (p = 0.00 < 0.05)
                        thrpt:  [-41.763% -36.410% -30.629%]
                        Performance has regressed.
Found 5 outliers among 100 measurements (5.00%)
  4 (4.00%) high mild
  1 (1.00%) high severe
tcp/16384               time:   [2.7120 ms 2.7387 ms 2.7652 ms]
                        thrpt:  [361.64 MiB/s 365.13 MiB/s 368.73 MiB/s]
                 change:
                        time:   [+30.950% +42.147% +54.770%] (p = 0.00 < 0.05)
                        thrpt:  [-35.388% -29.650% -23.635%]
                        Performance has regressed.
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) low mild
tcp/65536               time:   [2.2610 ms 2.4366 ms 2.6072 ms]
                        thrpt:  [383.56 MiB/s 410.41 MiB/s 442.29 MiB/s]
                 change:
                        time:   [+140.64% +162.39% +182.64%] (p = 0.00 < 0.05)
                        thrpt:  [-64.619% -61.888% -58.445%]
                        Performance has regressed.
tcp/262144              time:   [3.0009 ms 3.0434 ms 3.0865 ms]
                        thrpt:  [324.00 MiB/s 328.58 MiB/s 333.24 MiB/s]
                 change:
                        time:   [+99.851% +115.74% +132.65%] (p = 0.00 < 0.05)
                        thrpt:  [-57.017% -53.647% -49.963%]
                        Performance has regressed.
tcp/1048576             time:   [3.8665 ms 3.8915 ms 3.9158 ms]
                        thrpt:  [255.37 MiB/s 256.97 MiB/s 258.63 MiB/s]
                 change:
                        time:   [+42.686% +53.377% +65.578%] (p = 0.00 < 0.05)
                        thrpt:  [-39.606% -34.801% -29.916%]
                        Performance has regressed.
Found 3 outliers among 100 measurements (3.00%)
  2 (2.00%) low mild
  1 (1.00%) high mild

memory/256              time:   [3.9661 ms 4.0088 ms 4.0507 ms]
                        thrpt:  [246.87 MiB/s 249.45 MiB/s 252.14 MiB/s]
                 change:
                        time:   [+54.112% +61.755% +68.979%] (p = 0.00 < 0.05)
                        thrpt:  [-40.821% -38.178% -35.112%]
                        Performance has regressed.
memory/512              time:   [2.4825 ms 2.5069 ms 2.5319 ms]
                        thrpt:  [394.96 MiB/s 398.89 MiB/s 402.82 MiB/s]
                 change:
                        time:   [+51.312% +58.870% +66.709%] (p = 0.00 < 0.05)
                        thrpt:  [-40.015% -37.056% -33.911%]
                        Performance has regressed.
Found 13 outliers among 100 measurements (13.00%)
  4 (4.00%) low mild
  9 (9.00%) high mild
Benchmarking memory/1024: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 9.7s, enable flat sampling, or reduc
memory/1024             time:   [1.6728 ms 1.7484 ms 1.8086 ms]
                        thrpt:  [552.90 MiB/s 571.95 MiB/s 597.79 MiB/s]
                 change:
                        time:   [+23.997% +34.583% +46.069%] (p = 0.00 < 0.05)
                        thrpt:  [-31.539% -25.697% -19.353%]
                        Performance has regressed.
Benchmarking memory/8192: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 6.2s, enable flat sampling, or reduce sample count to 60.
memory/8192             time:   [1.3502 ms 1.3676 ms 1.3871 ms]
                        thrpt:  [720.93 MiB/s 731.22 MiB/s 740.66 MiB/s]
                 change:
                        time:   [+34.785% +44.974% +56.196%] (p = 0.00 < 0.05)
                        thrpt:  [-35.978% -31.022% -25.808%]
                        Performance has regressed.
memory/16384            time:   [791.93 us 908.92 us 1.0127 ms]
                        thrpt:  [987.44 MiB/s 1.0744 GiB/s 1.2331 GiB/s]
                 change:
                        time:   [+11.338% +22.345% +33.895%] (p = 0.00 < 0.05)
                        thrpt:  [-25.315% -18.264% -10.183%]
                        Performance has regressed.
Benchmarking memory/65536: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 5.7s, enable flat sampling, or reduce sample count to 60.
memory/65536            time:   [1.1447 ms 1.1558 ms 1.1671 ms]
                        thrpt:  [856.80 MiB/s 865.22 MiB/s 873.62 MiB/s]
                 change:
                        time:   [+68.560% +82.823% +98.392%] (p = 0.00 < 0.05)
                        thrpt:  [-49.595% -45.302% -40.674%]
                        Performance has regressed.
Found 8 outliers among 100 measurements (8.00%)
  1 (1.00%) low mild
  7 (7.00%) high mild
Benchmarking memory/262144: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 5.7s, enable flat sampling, or reduce sample count to 60.
memory/262144           time:   [1.0463 ms 1.1525 ms 1.2377 ms]
                        thrpt:  [807.98 MiB/s 867.68 MiB/s 955.77 MiB/s]
                 change:
                        time:   [+13.879% +30.134% +47.883%] (p = 0.00 < 0.05)
                        thrpt:  [-32.379% -23.156% -12.188%]
                        Performance has regressed.
memory/1048576          time:   [2.1756 ms 2.1942 ms 2.2129 ms]
                        thrpt:  [451.90 MiB/s 455.75 MiB/s 459.65 MiB/s]
                 change:
                        time:   [+72.594% +79.643% +86.679%] (p = 0.00 < 0.05)
                        thrpt:  [-46.432% -44.334% -42.061%]
                        Performance has regressed.
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) high mild
```

thomaseizinger:
Additionally, there seems to be some dead-lock bug somewhere because CI is running forever, will investigate.

core/src/muxing.rs
```diff
@@ -79,9 +79,6 @@ mod singleton;
 ///   `destroy_outbound` methods allow controlling these entries.
 ///
 pub trait StreamMuxer {
-    /// Type of the object that represents the raw substream where data can be read and written.
-    type Substream;
```
mxinden: What is the reasoning for using dynamic dispatching (`AsyncReadWriteBox`) instead of using an associated type that implements `AsyncRead` and `AsyncWrite`?

core/src/muxing.rs
```diff
@@ -99,31 +104,13 @@ pub trait StreamMuxer {
...
-    /// Polls the outbound substream.
-    ///
-    /// If `Pending` is returned, then the current task will be notified once the substream
-    /// is ready to be polled, similar to the API of `Future::poll()`.
-    /// However, for each individual outbound substream, only the latest task that was used to
-    /// call this method may be notified.
-    ///
-    /// May panic or produce an undefined result if an earlier polling of the same substream
-    /// returned `Ready` or `Err`.
-    fn poll_outbound(
```
mxinden, line 119: What if, as a user, I want to poll for outbound substreams, while at the same time signaling to the remote that I am not willing to accept any more inbound substreams? In other words, I do I exercise back-pressure on the number of inbound substreams?

I know this is not supported in mplex or yamux today. It is supported by QUIC and will be supported by https://github.com/libp2p/specs/pull/394.

swarm/src/connection/substream.rs
```diff
@@ -18,196 +18,9 @@
...
 }
-
-/// Implementation of `Stream` that handles substream multiplexing.
-///
-/// The stream will receive substreams and can be used to open new outgoing substreams. Destroying
-/// the `Muxing` will **not** close the existing substreams.
-///
-/// The stream will close once both the inbound and outbound channels are closed, and no more
-/// outbound substream attempt is pending.
-pub struct Muxing<TMuxer, TUserData>
```
mxinden, line 42: :rocket:

muxers/yamux/src/lib.rs
```diff
@@ -39,15 +39,7 @@ use std::{
 use thiserror::Error;
 
 /// A Yamux connection.
-pub struct Yamux<S>(Mutex<Inner<S>>);
```
mxinden, line 41: Neat!

muxers/mplex/src/lib.rs
```diff
@@ -236,3 +205,9 @@ impl AsyncWrite for Substream {
         this.muxer.lock().poll_close_stream(cx, this.id)
     }
 }
+
+impl Drop for Substream {
```
mxinden: Is this change necessary for the general goal of _redesigning the `StreamMuxer` trait_? I would prefer unrelated changes to be split into separate pull requests, thus not having one very large one.

mxinden (commented):
Some very cool simplifications in here.

Any chance this pull request can be split into multiple autonomous ones to ease the review process?

core/src/muxing.rs
thomaseizinger: Part of the motivation for this change was the idea that protocols (like `libp2p-ping`) should depend on a type that is independent from the specific muxer in question and instead are implemented on top of an opaque stream type. This would allow standalone usage of the protocol (state machines) which hopefully brings us closer to using some of the protocol logic independently from `libp2p-swarm`.

Additionally, I am not even sure an associated type would work unless we parametrize all protocols over `M: StreamMuxer` to be able to reference the concrete substream type in methods like `inject_fully_negotiated_inbound`?

core/src/muxing.rs
thomaseizinger: > while at the same time signaling to the remote that I am not willing to accept any more inbound substreams?

If this is a requirement, I'd suggest we make a function on `StreamMuxer` to trigger this behaviour.

In general though, I am a bit hesitant to do this. The original design of the `StreamMuxer` trait had numerous features that were completely unused (at least within the libp2p ecosystem that is known to me). Unless we use this functionality as part of `libp2p-swarm`, I'd refrain from adding it.

thomaseizinger: > I know this is not supported in mplex or yamux today. It is supported by QUIC and will be supported by [libp2p/specs#394](https://github.com/libp2p/specs/pull/394).

This sounds like it would be better of as a method on the _concrete_ muxer type. Part of the benefits of the changes in this PR is that the `StreamMuxer` implementations are much easier to use standalone, meaning any usage outside of `libp2p-swarm` does not necessarily need to go through the `StreamMuxerBox` abstraction and hence can access functionality outside of the trait interface.

muxers/yamux/src/lib.rs
thomaseizinger: Well, looking at the current build failures this may have been a mistake :D

muxers/mplex/src/lib.rs
thomaseizinger: I think it is coupled to it, yes. Where-else would you clean up the resources associated with a substream within the mplex implementation?

thomaseizinger:
I had to drop some `Sync` bounds, see https://github.com/libp2p/rust-libp2p/pull/2648/commits/f9379a0962515d59b2810b2b9f4d4e6d4af807be. I think that is okay but please have a good think about it @mxinden :)

core/src/muxing.rs
thomaseizinger: My current thinking is that it would be nice to have `libp2p-swarm` be an optional feature of protocol crates.

thomaseizinger:
> Any chance this pull request can be split into multiple autonomous ones to ease the review process?

I could split out some of the removal of functions from the `StreamMuxer` crate into a separate PR perhaps? To be functionally atomic, the changes of passing out an opaque stream type, changing `poll` etc are coupled I believe :/

I can invest some time into tidying up the commits if you'd like?

thomaseizinger:
> I can invest some time into tidying up the commits if you'd like?

I am actually unhappy with how messy it is so once I have CI finally passing, I'll pull the PR apart and make distinct commits so that the entire thing is reviewable patch-by-patch.

thomaseizinger:
@mxinden @elenaf9 This is (finally) ready for a full review. I completely redid the commit history to separate a few changes that could be done in isolation. Each commit should compile and pass the tests.

I couldn't make the final commit much smaller unfortunately.

core/src/muxing.rs
mxinden: > My current thinking is that it would be nice to have `libp2p-swarm` be an optional feature of protocol crates.

Protocol crates like e.g. `libp2p-gossipsub`? Would they still be using `NetworkBehaviour`?

> Additionally, I am not even sure an associated type would work unless we parametrize all protocols over `M: StreamMuxer` to be able to reference the concrete substream type in methods like `inject_fully_negotiated_inbound`?

We might decide to do dynamic dispatching at the `libp2p-swarm` level, though I am not sure we need it at the `libp2p-core` level.

core/src/muxing.rs
mxinden: > > while at the same time signaling to the remote that I am not willing to accept any more inbound substreams?
> 
> If this is a requirement, I'd suggest we make a function on `StreamMuxer` to trigger this behaviour.

This will be a requirement. It isn't today, due to rust-libp2p not yet implementing a sophisticated stream multiplexer. Will follow up with more details.

mxinden:
Thanks for cleaning up the commit history!

As far as I can tell, the following commits can stand by themselves, correct? Mind proposing them as separate pull requests? While small, it would enforce a clean git history, proper changelogs, easier pull request reviews and in the worst case easier `git bisect`.

- Fix StreamMuxer::Error to io::Error
- Remove the StreamMuxer::flush_all function
- Remove deprecated function
- Rename close to poll_close
- Remove various Sync bounds
- Remove unnecessary mapping to StreamMuxerBox
- Refactor Yamux::poll_close

Happy to do the work (cherry-picking) if you want.

muxers/mplex/tests/async_write.rs
```diff
@@ -64,8 +64,8 @@ fn async_write() {
             .await
             .unwrap();
 
-        let mut buf = Vec::new();
-        outbound.read_to_end(&mut buf).await.unwrap();
+        let mut buf = vec![0u8; 11];
+        outbound.read_exact(&mut buf).await.unwrap();
```
mxinden: Would that not defeat the purpose of the unit test?

Above:

>     // Tests that `AsyncWrite::close` implies flush.

Not using `read_to_end` thus no longer tests the closing and flushing, no?

core/src/muxing.rs
thomaseizinger: Fair point! Can we agree to design this once we have/add a sophisticated stream multiplexer? I am not confident that I can add this functionality now in a way that will be useful once we need it!

core/src/muxing.rs
thomaseizinger: > > My current thinking is that it would be nice to have `libp2p-swarm` be an optional feature of protocol crates.
> 
> Protocol crates like e.g. `libp2p-gossipsub`? Would they still be using `NetworkBehaviour`?

They would, optionally, by dispatching to standalone state machines that only depend on `libp2p-core` and could thus be used without the `libp2p-swarm` abstractions. Without derailing the conversation too much: I see one of two avenues for making `rust-libp2p` easier for "arbitrary stream usage".

1. Re-designing the `NetworkBehaviour` trait to provide more flexibility
2. Making it easier to not use `libp2p-swarm` but still be able to re-use the existing work and protocols.

> > Additionally, I am not even sure an associated type would work unless we parametrize all protocols over `M: StreamMuxer` to be able to reference the concrete substream type in methods like `inject_fully_negotiated_inbound`?
> 
> We might decide to do dynamic dispatching at the `libp2p-swarm` level, though I am not sure we need it at the `libp2p-core` level.

That is an interesting idea, I will go and explore that.

thomaseizinger:
> As far as I can tell, the following commits can stand by themselves, correct? Mind proposing them as separate pull requests? While small, it would enforce a clean git history, proper changelogs, easier pull request reviews and in the worst case easier `git bisect`.
> 
>  * Fix StreamMuxer::Error to io::Error
>  * Remove the StreamMuxer::flush_all function
>  * Remove deprecated function
>  * Rename close to poll_close
>  * Remove various Sync bounds
>  * Remove unnecessary mapping to StreamMuxerBox
>  * Refactor Yamux::poll_close

Yes they can stand by themselves. Happy to post them as individual PRs :)

muxers/mplex/tests/async_write.rs
thomaseizinger: Yeah you might be right. I wasn't too happy about these changes but extracted them as a separate commit so we can discuss it easier.

I suspect there might be a bug somewhere but I couldn't spot it so far.

core/src/muxing.rs
thomaseizinger: What do yo think of https://github.com/libp2p/rust-libp2p/pull/2648/commits/61f07e0271e143992a00463f8d0ffe50b92bd32f?

thomaseizinger:
I opted to force push this PR because I expect to extract more commits out of it, hence the final one should stay at the end. No changes have been made other than resolving a few conflicts.

thomaseizinger:
> * Fix mplex performance regression (use channels of `Bytes`/`[u8; X]`?)

I tried to implement the communication between mplex's substreams and the stream muxer via channels but I don't think it actually makes any sense. Not unless we also rewrite the internals of `io::Multiplexed`. The issue is that `io::Multiplexed` is optimised to store individual wakers per substreams but with a channel-based implementation, we would always be waking the same single task for every substream.

With the current implementation of a mutex, only the specific task of the substream is woken because the `Context` is passed all the way through to the `io::Multiplexed`.

Running the benchmark on a single thread (using `ASYNC_STD_THREAD_COUNT=1`) unsurprisingly mitigates the performance problems because the Mutex is never congested. Within `libp2p-swarm`, we run every connection handler on its own task, which I think should yield about the same result.

Hence, I am tempted to conclude that the benchmark (and also its regressions with this PR) are not representative of actual usage and I'd propose to keep the `Mutex`-based implementation.

Thoughts @mxinden @elenaf9?

core/src/muxing.rs
```diff
@@ -51,178 +50,61 @@
...
     /// or polls must generate an error or be ignored.
     ///
-    /// Calling this method implies `flush_all`.
-    ///
     /// > **Note**: You are encouraged to call this method and wait for it to return `Ready`, so
     /// >           that the remote is properly informed of the shutdown. However, apart from
     /// >           properly informing the remote, there is no difference between this and
     /// >           immediately dropping the muxer.
-    fn poll_close(&self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>>;
+    fn poll_close(&mut self, cx: &mut Context<'_>) -> Poll<io::Result<()>>;
```
elenaf9, line 98:
```suggestion
    fn poll_close(mut self, cx: &mut Context<'_>) -> Poll<io::Result<()>>;
```
Why not move ownership if the muxer should not be used afterwards anyway?

core/src/muxing/async_read_write_box.rs
```diff
@@ -0,0 +1,105 @@
+// Copyright 2022 Thomas Eizinger.
```
elenaf9: I think it would make sense to also move`StreamMuxerBox` and `Wrap` into this module, since they are only used in combination with each other. Maybe also rename the module to `boxed` analogous to `transport::boxed`, which does something similar for transports. 

> - Is there a better name for AsyncReadWriteBox?

`SubstreamBox`? Even though it is theoretically abstract and could be used independently of the Muxer logic, in practice we only use it to box the `Mutex::Substream`, so maybe just reflect that in the name?

muxers/yamux/CHANGELOG.md
```diff
@@ -1,6 +1,9 @@
 # 0.37.0 [unreleased]
 
 - Update to `libp2p-core` `v0.33.0`.
+- Remove `YamuxResult` and `YamuxError` from the public API. See [PR 2648].
+
+[PR 2648]: https://github.com/libp2p/rust-libp2p/pull/2648
```
elenaf9, line 6: I think the changelog for yamux should also mention the muxer trait & substream handling changes. The current entry imo does not sufficiently inform users that this PR did internal changes that might affect performance or could have introduced bugs.

elenaf9 (approved):
The muxer logic in libp2p is still new to me so I can't give a profound review right now, but from a high-level perspective I like the changes of this PR very much! The API is much more intuitive, and consistent with the rest of rust-libp2p.

core/src/muxing.rs
thomaseizinger: It is still a poll function that may return Pending and needs to be polled again!

core/src/muxing/async_read_write_box.rs
thomaseizinger: Excellent ideas! Will implement!

muxers/yamux/CHANGELOG.md
thomaseizinger: This changelog entry will already be included within https://github.com/libp2p/rust-libp2p/pull/2664.

I'll write another a more detailed one, once https://github.com/libp2p/rust-libp2p/pull/2664 is merged and this is rebased on top of it!

core/src/muxing.rs
mxinden: > I am not confident that I can add this functionality now in a way that will be useful once we need it!

Sounds good to me, especially since we don't prioritize outbound streams over inbound ones today.

https://github.com/libp2p/rust-libp2p/blob/3611d69e0911c5551270eaf8a6aa43d9b49f8782/swarm/src/connection/substream.rs#L154-L175

core/src/muxing.rs
```diff
@@ -231,511 +114,92 @@ pub enum StreamMuxerEvent<T> {
...
 
-    #[inline]
-    fn shutdown_substream(
-        &self,
-        cx: &mut Context<'_>,
-        s: &mut Self::Substream,
-    ) -> Poll<Result<(), Self::Error>> {
-        self.inner.shutdown_substream(cx, s)
-    }
+    fn fetch_and_increment(&mut self) -> Self {
```
mxinden:
```suggestion
    fn get_then_increment(&mut self) -> Self {
```

Would be more intuitive to me. Not sure whether the current version is more idiomatic?

core/src/muxing.rs
```diff
@@ -231,511 +114,92 @@ pub enum StreamMuxerEvent<T> {
...
+            return Poll::Pending;
+        }
 
-    #[inline]
-    fn destroy_substream(&self, substream: Self::Substream) {
-        let mut list = self.substreams.lock();
-        self.inner
-            .destroy_substream(list.remove(&substream).unwrap())
-    }
+        let stream = futures::ready!(poll_open())?;
```
mxinden, line 198: This indirection seems counterintuitive to me. While more code, I would expect the muxer calling the `OutboundStreams` instead of the `OutboundStreams` calling the muxer to be easier to follow.

muxers/yamux/src/lib.rs
```diff
@@ -386,23 +333,16 @@ where
...
-#[error("yamux error: {0}")]
-pub struct YamuxError(#[from] yamux::ConnectionError);
-
-impl From<YamuxError> for io::Error {
-    fn from(err: YamuxError) -> Self {
-        match err.0 {
-            yamux::ConnectionError::Io(e) => e,
-            e => io::Error::new(io::ErrorKind::Other, e),
-        }
+fn to_io_error(e: yamux::ConnectionError) -> io::Error {
```
mxinden, line 339: What do you think of upstreaming this to the `yamux` crate?

https://github.com/libp2p/rust-yamux

core/src/muxing.rs
```diff
@@ -51,178 +50,62 @@
...
-/// - A list of substreams that are open. The `poll_outbound`, `read_substream`, `write_substream`,
-///   `flush_substream`, `shutdown_substream` and `destroy_substream` methods allow controlling
-///   these entries.
-/// - A list of outbound substreams being opened. The `open_outbound`, `poll_outbound` and
-///   `destroy_outbound` methods allow controlling these entries.
+/// Provides multiplexing for a connection by allowing users to open substreams.
 ///
+/// A [`StreamMuxer`] is a stateful object and needs to be continuously polled to make progress.
 pub trait StreamMuxer {
-    /// Type of the object that represents the raw substream where data can be read and written.
```
mxinden, line 77: Comment is roughly accurate, no?

mxinden (commented):
>  +583 −1,401 

Very cool! 

A couple of comments. Sorry, large change, thus I am still not all the way through.

core/src/muxing.rs
thomaseizinger: Naming was inspired by `Atomic::fetch_add`: https://doc.rust-lang.org/std/sync/atomic/struct.AtomicU8.html#method.fetch_add

Could probably just call it `fetch_add` directly?

muxers/yamux/src/lib.rs
thomaseizinger: We can upstream that yes!

core/src/muxing.rs
thomaseizinger: Correct. Removing this is a left-over from first removing the associated type and reintroducing it based on the review feedback :D

I'll restore the doc string as well!

core/src/muxing.rs
mxinden: :+1: for reintroducing `type Substream`.

core/src/muxing.rs
mxinden: Sounds good to me.

core/src/muxing.rs
thomaseizinger: The _concrete_ muxers are calling the `OutboundSubstreams` component.

mplex: https://github.com/thomaseizinger/rust-libp2p/blob/db7cb6b47ace2f844a7a6da6ece345d8d2f6657a/muxers/mplex/src/lib.rs#L118-L120
yamux: https://github.com/thomaseizinger/rust-libp2p/blob/db7cb6b47ace2f844a7a6da6ece345d8d2f6657a/muxers/yamux/src/lib.rs#L110-L113

thomaseizinger: I am not sure how to otherwise design this abstraction so it can be reused by each muxer implementation.

swarm/src/connection/substream.rs
```diff
@@ -18,240 +18,9 @@
...
 // DEALINGS IN THE SOFTWARE.
 
-use futures::prelude::*;
-use libp2p_core::multiaddr::Multiaddr;
-use libp2p_core::muxing::{substream_from_ref, StreamMuxer, StreamMuxerEvent, SubstreamRef};
-use smallvec::SmallVec;
-use std::sync::Arc;
-use std::{fmt, io::Error as IoError, pin::Pin, task::Context, task::Poll};
-
 /// Endpoint for a received substream.
```
mxinden: How about moving this into `connection.rs?

muxers/yamux/src/lib.rs
```diff
@@ -84,110 +83,58 @@ where
...
+        }
 
-    fn flush_substream(
-        &self,
-        c: &mut Context<'_>,
-        s: &mut Self::Substream,
-    ) -> Poll<YamuxResult<()>> {
-        Pin::new(s).poll_flush(c).map_err(|e| YamuxError(e.into()))
-    }
+        let event = ready!(self.incoming.poll_next_unpin(c))
```
mxinden: Nit pick, instead of the `ready!`, how about following the style above using `if let Poll::Ready`. Say that we introduce a statement below this chain, we might miss the early return.

core/src/muxing.rs
mxinden: I am more referring to the closer (`poll_open`) passed by the muxer to the `OutboundSubstreams`. Is that a strict requirement?

muxers/yamux/src/lib.rs
thomaseizinger: See https://github.com/libp2p/rust-yamux/pull/135.

muxers/yamux/src/lib.rs
thomaseizinger: Happy to do that!

thomaseizinger: I did the same thing in mplex!

swarm/src/connection/substream.rs
thomaseizinger: Good idea!

core/src/muxing.rs
thomaseizinger: I am not sure I completely follow but here is the thinking:

1. We should only poll to open a new open substream if the user requested one.
2. Requesting to open a new substream issues an `OutboundSubstreamId`.
3. Thus, opening a new substream is based on the state of having issued `OutboundSubstreamId`s
4. The correct IDs need to be mapped to the `OutboundSubstreamId`s that were handed out.

I could inline this abstraction but it felt a bit error prone to duplicate the behaviour of only polling for new substreams if we have issued IDs but not serviced them yet.

core/src/muxing.rs
elenaf9: Ahh yes of course you're right. Somehow that completely slipped my mind.

core/src/muxing.rs
```diff
@@ -51,178 +50,63 @@
...
-    ) -> Poll<Result<StreamMuxerEvent<Self::Substream>, Self::Error>>;
-
-    /// Opens a new outgoing substream, and produces the equivalent to a future that will be
-    /// resolved when it becomes available.
-    ///
-    /// The API of `OutboundSubstream` is totally opaque, and the object can only be interfaced
-    /// through the methods on the `StreamMuxer` trait.
-    fn open_outbound(&self) -> Self::OutboundSubstream;
+    fn poll(&mut self, cx: &mut Context<'_>)
+        -> Poll<io::Result<StreamMuxerEvent<Self::Substream>>>;
```
elenaf9, line 79: In what cases does it makes sense in poll functions to have `&mut self`, and when should it be `Pin<&mut Self>`? I wasn't sure which of the two to take in `Transport::poll`, and settled for `Pin<&mut Self>` to mirror the `Stream` trait. But it would be good to have more profound reasoning for when to use what in our API

thomaseizinger: It is my understanding that `Pin` was introduced to allow self-referential structs in futures that are generated from async blocks. Within our hand-written state machines, we don't have self-referential structs and thus, they don't need to be pinned during execution within an executor. In other words, they will automatically implement `Unpin`.

Ideally, `Pin` wouldn't exist and all poll-functions would take `&mut self` but this wouldn't allow for safe self-referential structs within futures. Thus I think the way to go is: Use `&mut self` until the compiler complains :)

muxers/mplex/src/lib.rs
```diff
@@ -46,30 +48,38 @@ impl UpgradeInfo for MplexConfig {
...
-    type Output = Multiplex<C>;
+    type Output = Multiplex;
     type Error = io::Error;
     type Future = future::Ready<Result<Self::Output, io::Error>>;
 
     fn upgrade_outbound(self, socket: C, _: Self::Info) -> Self::Future {
         future::ready(Ok(Multiplex {
-            io: Mutex::new(io::Multiplexed::new(socket, self)),
+            io: Arc::new(Mutex::new(io::Multiplexed::new(
+                SubstreamBox::new(socket),
```
thomaseizinger: @elenaf9 I remember why I didn't name it `SubstreamBox` initially! Because it is also used here to erase the type of the socket which would otherwise also show up in the type signature of `libp2p_mplex::Substream`.

However, I think that is actually not a problem because _now_ we box the substream later on anyway. In an earlier design we didn't!

muxers/mplex/tests/async_write.rs
thomaseizinger: I fixed it! The important commit is https://github.com/libp2p/rust-libp2p/commit/cec117667b3e0a23b8acb3acbc342e087ef8df6b.

This means I can drop this commit from the PR history but that will require a force-push. I've noted it in the PR description as a task for the final clean-up once this PR is closer to being merged :)

core/src/muxing.rs
elenaf9: Thank you for the explanation @thomaseizinger!

> Thus I think the way to go is: Use &mut self until the compiler complains :)

But for traits we won't really know it the struct that implements it will have self-references, no? So the compiler will not complain when defining the trait, but might complain once a user implements it on their type.

thomaseizinger: > > Thus I think the way to go is: Use &mut self until the compiler complains :)
> 
> But for traits we won't really know it the struct that implements it will have self-references, no? So the compiler will not complain when defining the trait, but might complain once a user implements it on their type. 

That is true!

The safer way would be to use `Pin` everywhere but it is also additional complexity. I could go either way.

There are probably no implementations of `StreamMuxer` outside of this repository so leaving it at `&mut self` shouldn't cause any problems but at the same time, the additional complexity would also only really affect us core maintainers 🤷‍♂️

thomaseizinger:
Force-push only dropped commit b6c91d138b2d4078c2ba08b436256b4d035d59ab and rebased onto latest master. Diff here: https://github.com/libp2p/rust-libp2p/compare/cec117667b3e0a23b8acb3acbc342e087ef8df6b..c2a11bd25d7987e3f727fb8ffea29ff2bf690442

core/src/muxing.rs
mxinden: In my eyes, for correctness reasons, we should mirror the `Future` standard of using `Pin<&mut Self>`. Also see https://github.com/libp2p/rust-libp2p/issues/1389.

muxers/mplex/src/lib.rs
```diff
@@ -79,117 +83,142 @@ where
...
+            .outbound_substreams
+            .poll(|| Pin::new(&mut self.io.lock()).poll_open_stream(cx))?
+        {
+            let substream = Substream::new(mplex_id, self.io.clone());
+
+            return Poll::Ready(Ok(StreamMuxerEvent::OutboundSubstream(
+                substream,
+                outbound_id,
+            )));
+        }
```
mxinden: We should favor local work over remote work, i.e. `self.io.lock()).poll_open_stream(cx)` should be polled before `self.io.lock().poll_next_stream(cx)`. Does that make sense in this scenario?

core/src/muxing.rs
thomaseizinger: Interesting! Happy to move towards `Pin<&mut self>`. I guess without knowing, I was working towards #1389 :D

mxinden:
Do the mplex benchmarks still show a large regression?

I am surprised about it. Also I am surprised that this is due to the `Mutex`. Didn't the previous implementation use mutual exclusion as well? I would expect most CPU cycles to be spend on the many cases where we do heap based copying, not so on the cache coherence overhead introduced through the `Mutex`. Just an intuition.

muxers/mplex/src/lib.rs
thomaseizinger: Makes sense!

thomaseizinger:
> Do the mplex benchmarks still show a large regression?

They do unless run on a single thread, see https://github.com/libp2p/rust-libp2p/pull/2648#issuecomment-1141257327.

> Didn't the previous implementation use mutual exclusion as well? I would expect most CPU cycles to be spend on the many cases where we do heap based copying, not so on the cache coherence overhead introduced through the `Mutex`. Just an intuition.

The previous `StreamMuxerBox` [used 2 different Mutex](https://github.com/libp2p/rust-libp2p/blob/fcc987e0f6dd82d0f702acd8b819b4f99a9fb26a/core/src/muxing.rs#L612-L615) for various internal data structures so perhaps there was less congestion there?

muxers/mplex/src/lib.rs
thomaseizinger: Done in https://github.com/libp2p/rust-libp2p/pull/2648/commits/e9506aae7048184e90ceabd4dcbad3f07126308d.

muxers/mplex/benches/split_send_size.rs
```diff
@@ -132,24 +130,16 @@ fn run(transport: &mut BenchTransport, payload: &Vec<u8>, listen_addr: &Multiadd
...
-                .await
-                .unwrap();
-            off += n;
-            if off == payload.len() {
-                poll_fn(|cx| conn.flush_substream(cx, &mut stream))
-                    .await
-                    .unwrap();
-                return;
-            }
-        }
```
thomaseizinger, line 152: @mxinden This is also differently implemented in the benchmark now but I can't see how this would impact performance much.

muxers/mplex/src/lib.rs
```diff
@@ -79,117 +83,142 @@ where
 /// This implementation isn't capable of detecting when the underlying socket changes its address,
 /// and no [`StreamMuxerEvent::AddressChange`] event is ever emitted.
 pub struct Multiplex<C> {
-    io: Mutex<io::Multiplexed<C>>,
+    io: Arc<Mutex<io::Multiplexed<C>>>,
```
thomaseizinger, line 86: This is a `parkling_lot::Mutex`. I'll try if a native async one makes a difference.

mxinden:
I would guess that https://github.com/flamegraph-rs/flamegraph proves useful here. One can compile the benchmarks to a binary and then run them via `cargo flamegraph`.

muxers/mplex/benches/split_send_size.rs
mxinden: > but I can't see how this would impact performance much.

Me neither :/

thomaseizinger:
> I would guess that https://github.com/flamegraph-rs/flamegraph proves useful here. One can compile the benchmarks to a binary and then run them via `cargo flamegraph`.

Great idea. I am familiar with flamegraph-rs, just didn't think of using it.

thomaseizinger:
**EDIT**: I managed to extract the crucial bit into a dedicated PR: https://github.com/libp2p/rust-libp2p/pull/2706 Best to benchmark that in isolation.

OLD COMMENT BELOW:

---

Interestingly enough, I can no longer reproduce the performance problems. Would you mind trying @elenaf9 / @mxinden? What I did was:

- Install `cargo-criterion`
- Run `cargo criterion` within the `muxers/mplex` directory for `master` branch
- Run `cargo criterion` within the `muxers/mplex` directory for `refactoring-multiplexing` branch (Easy to checkout via `gh pr checkout 2648`)

Funnily, I can now actually see a performance _improvement_ of 5% for some of the TCP benchmarks and even 10% for the memory benchmarks.

`tcp/16384` and `memory/16384` are are an outlier with ~5% regression.

mxinden:
@elenaf9 just double checking, none of these changes make QUIC in rust-libp2p harder to implement, right? I would expect the opposite, namely for all of this to simplify https://github.com/libp2p/rust-libp2p/pull/2289.

mxinden:
> Interestingly enough, I can no longer reproduce the performance problems. Would you mind trying @elenaf9 / @mxinden? What I did was:

Sorry, late reply here. Happy to do it. Would I be running against this branch or against `master` given that #2706 is already merged?

thomaseizinger:
> > Interestingly enough, I can no longer reproduce the performance problems. Would you mind trying @elenaf9 / @mxinden? What I did was:
> 
> Sorry, late reply here. Happy to do it. Would I be running against this branch or against `master` given that #2706 is already merged?

If you want, you can benchmark https://github.com/libp2p/rust-libp2p/pull/2706 against master before it was merged :)

mxinden:
@melekes given that https://github.com/libp2p/rust-libp2p/pull/2622 now also implements the `StreamMuxer` trait, I want to make sure you are aware of Thomas refactorings around the trait.

Let us know in case there is something we should be considering (with regards to WebRTC).

I would expect the overall change set to simplify the WebRTC muxing implementation.

thomaseizinger:
I am closing this PR in favor of a ticket where we can track all of this: https://github.com/libp2p/rust-libp2p/issues/2722

mxinden:
> > > Interestingly enough, I can no longer reproduce the performance problems. Would you mind trying @elenaf9 / @mxinden? What I did was:
> > 
> > 
> > Sorry, late reply here. Happy to do it. Would I be running against this branch or against `master` given that #2706 is already merged?
> 
> If you want, you can benchmark #2706 against master before it was merged :)

@thomaseizinger: Ran on a `Intel(R) Core(TM) i7-8550U CPU @ 1.80GHz`:

```
critcmp 3c120ef9 ea487aeb
group             3c120ef9                                ea487aeb
-----             --------                                --------
memory/1024       1.00   518.4±42.66µs  1929.1 MB/sec     1.01   525.2±48.46µs  1904.1 MB/sec
memory/1048576    1.00   492.1±21.08µs  2032.2 MB/sec     1.03   508.9±44.19µs  1965.0 MB/sec
memory/16384      1.00   288.6±39.33µs     3.4 GB/sec     1.20   346.2±58.03µs     2.8 GB/sec
memory/256        1.00  1319.6±74.77µs   757.8 MB/sec     1.01  1328.2±60.90µs   752.9 MB/sec
memory/262144     1.00   284.4±39.42µs     3.4 GB/sec     1.00   284.5±38.24µs     3.4 GB/sec
memory/512        1.00   804.1±47.89µs  1243.7 MB/sec     1.00   803.6±47.80µs  1244.5 MB/sec
memory/65536      1.00   264.0±37.24µs     3.7 GB/sec     1.01   266.7±37.35µs     3.7 GB/sec
memory/8192       1.00   317.7±47.46µs     3.1 GB/sec     1.01   321.2±51.34µs     3.0 GB/sec
tcp/1024          1.02   810.9±57.88µs  1233.2 MB/sec     1.00   798.6±85.29µs  1252.2 MB/sec
tcp/1048576       1.00   885.9±56.03µs  1128.7 MB/sec     1.00   885.9±91.76µs  1128.8 MB/sec
tcp/16384         1.01   668.6±93.15µs  1495.7 MB/sec     1.00   659.9±98.35µs  1515.5 MB/sec
tcp/256           1.32  1927.9±258.41µs   518.7 MB/sec    1.00  1456.4±176.39µs   686.6 MB/sec
tcp/262144        1.03   679.6±96.01µs  1471.4 MB/sec     1.00   657.3±78.15µs  1521.4 MB/sec
tcp/512           1.12  1152.5±164.17µs   867.7 MB/sec    1.00  1027.3±134.77µs   973.4 MB/sec
tcp/65536         1.01   626.5±83.42µs  1596.1 MB/sec     1.00   619.0±78.21µs  1615.5 MB/sec
tcp/8192          1.04  686.9±117.84µs  1455.8 MB/sec     1.00  663.1±107.80µs  1508.1 MB/sec
```

where 3c120ef9 is the base and ea487aeb includes the change. I don't see any of this as a sign of a regression.

elenaf9:
> @elenaf9 just double checking, none of these changes make QUIC in rust-libp2p harder to implement, right? I would expect the opposite, namely for all of this to simplify #2289.

I did not get the chance to review the `QuicMuxer` in depth yet. I have been focused on the transport thus far. 
But the QuicMuxer will definitely need some refactoring after #2707, see https://github.com/libp2p/rust-libp2p/pull/2289#discussion_r885718446.  Because we now need to hand out substreams instead of just Ids, there is some additional logic needed. I can not say right now if it overall simplifies things or not.

But either way I think the changes of this PR are the right direction.

thomaseizinger:
> > @elenaf9 just double checking, none of these changes make QUIC in rust-libp2p harder to implement, right? I would expect the opposite, namely for all of this to simplify #2289.
> 
> I did not get the chance to review the `QuicMuxer` in depth yet. I have been focused on the transport thus far. 
> But the QuicMuxer will definitely need some refactoring after #2707, see https://github.com/libp2p/rust-libp2p/pull/2289#discussion_r885718446.  Because we now need to hand out substreams instead of just Ids, there is some additional logic needed. I can not say right now if it overall simplifies things or not.
> 
> But either way I think the changes of this PR are the right direction. 

I think worst case, you can follow an approach similar to the mplex muxer!

PR #2652, opened 2022-05-16
Transport: Poll Transport directly, remove ListenersStream 

elenaf9:
# Description

<s>_**Opened as a draf**t: I'd like to open the discussion on whether this direction is desired, and to discuss how this could be implement / how existing transport could be adapted to the change. Not ready for review yet._</s>

Remove the concept of `Listener`s from the `Transport` trait. Instead require the Transport itself to implement a stream-like API with `Transport::poll`. Transport implementations can then internally implement the logic for managing one or multiple listeners.

The main motivation for this change is that the concept of multiple listeners is rather TCP specific and introduces unnecessary complexity for the QUIC transport (https://github.com/libp2p/rust-libp2p/pull/2289#issuecomment-1119872606). 
Apart from that, this generally follows the same direction as #2648 :
> 2. Follow the style of "mutate state and return result via poll", same as we do for many other parts of rust-libp2p

The change includes:
- Remove `Transport::Listener` and change `Transport::listen_on` to just return a `Result<ListenerId, _>`
- Add `Transport::poll`, which returns `TransportEvent`s  (replacing the former `swarm/connection/ListenersEvent). Listeners should be driven here. 
- Remove `swarm/connection/ListenersStream`, instead poll the boxed transport directly in the swarm
- Adapt transports/tcp: Wrap the `GenTcpConfig` into a new struct `GenTcpTransport` that implements `Transport`:
   - Manage multiple listener streams (this is essentially the logic that used to be in `swarm/connection/ListenersStream`
- Adapt some of the existing Transports to this (not all yet, but that's only because I'd like to discuss the general direction first, not because I have concerns with specific transports)
- Adapt tests of transports/tcp, core and swarm to this change -> tests pass

Relevant discussions:
- https://github.com/libp2p/rust-libp2p/pull/2529#issuecomment-1087384988
- https://github.com/libp2p/rust-libp2p/discussions/2113#discussioncomment-944923
- https://github.com/libp2p/rust-libp2p/pull/2289#issuecomment-1119872606
- similar direction as #2648 

## Open Questions

**Transport::dial**

I did not change how `Transport::dial` works. I originally planned to also remove `Transport::Dial` and push that logic into `Transport::poll` as well. The problem with this would be that it does not work well with the concurrent-dial logic. 
 `ConcurrentDial` currently drives multiple dial futures (from potentially multiple transports) at the same time. Once the first succeeds, it drops the other ongoing dials. If the Transport would internally manage the `Dial` future, this logic would not work anymore. What we could do is something like the following:
  - Each dial is associated with a `DialId`, `ConcurrentDial` tracks the dial-ids of concurrent dials
  - We add `Transport::abort_dial` that can be called from the `Pool` once the first dial of a concurrent dialing batch succeeds

I am not sure if this is worth the extra complexity. Happy to hear alternative suggestions.

## Open TODOs
- [x] Add `Transport::remove_listener`
- [x] Track listening addresses
- [x] Adapt all remaining transports to the change 
- [x] Clean code, Fix docs
- [x] (not part of this PR, but should be done before merging this) draft how the QUIC implementation of #2289 could be adapted to this change (CC @kpp)

## Change checklist


core/src/transport.rs
```diff
@@ -115,12 +108,20 @@ pub trait Transport {
...
-    /// and addresses this transport is listening on (cf. [`ListenerEvent`]).
+    /// and addresses this transport is listening on (cf. [`TransportEvent`]).
     ///
     /// Returning an error from the stream is considered fatal. The listener can also report
-    /// non-fatal errors by producing a [`ListenerEvent::Error`].
-    fn listen_on(&mut self, addr: Multiaddr) -> Result<Self::Listener, TransportError<Self::Error>>
+    /// non-fatal errors by producing a [`TransportEvent::Error`].
+    fn listen_on(
+        &mut self,
+        id: ListenerId,
```
elenaf9: `Transport`s now have to be aware of `ListenerId`s since they need to include in `TransportEvent`s the information which listener produced the event. This is not ideal for users that use `core/transport` directly outside of a swarm, but I don't think there's any other way to solve this.

core/src/transport.rs
```diff
@@ -115,12 +108,20 @@ pub trait Transport {
     /// obtained from [dialing](Transport::dial).
     type Dial: Future<Output = Result<Self::Output, Self::Error>>;
 
+    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<TransportEvent<Self>>
```
elenaf9: Decided to add `Transport::poll` instead of requiring the Transport to implement `Stream`. Reason for this is that a transport's stream should never terminate (never return `None`). This is also aligned with how most other structs in `rust-libp2p` do it. 
Q: Should we auto-implement `Stream` for `T: Transport`? Would ease polling if users use a Transport directly.

core/src/transport.rs
```diff
@@ -152,9 +153,8 @@ pub trait Transport {
     /// Boxes the transport, including custom transport errors.
     fn boxed(self) -> boxed::Boxed<Self::Output>
     where
-        Self: Transport + Sized + Send + Sync + 'static,
+        Self: Transport + Sized + Send + Sync + Unpin + 'static,
```
elenaf9: I did not manage to implement the polling of the inner transport without it being `Unpin`.  Happy for any suggestions on how to solve this without the trait bound.

swarm/src/lib.rs
```diff
@@ -545,7 +548,9 @@ where
 
     /// Returns an iterator that produces the list of addresses we're listening on.
     pub fn listeners(&self) -> impl Iterator<Item = &Multiaddr> {
-        self.listeners.listen_addrs()
+        // TODO
+        vec![].into_iter()
+        // self.transport.listen_addrs()
     }
```
elenaf9: Should the Transport track the listening addresses or should the swarm do this according to new / expired addresses reported in `TransportEvent`s?

core/tests/transport_upgrade.rs
```diff
@@ -118,12 +119,17 @@ fn upgrade_pipeline() {
...
+        .listen_on(ListenerId::new(1), listen_addr1)
+        .unwrap();
 
     let server = async move {
         loop {
-            let (upgrade, _remote_addr) =
-                match listener.next().await.unwrap().unwrap().into_upgrade() {
+            let (upgrade, _send_back_addr) =
+                match poll_fn(|cx| Pin::new(&mut listener_transport).poll(cx))
+                    .await
```
elenaf9: Polling transports this way is rather inconvenient. Should we auto-implement `Stream` for transports?

transports/tcp/src/lib.rs
```diff
@@ -33,14 +33,14 @@ pub use provider::async_io;
...
-pub type TcpConfig = GenTcpConfig<async_io::Tcp>;
+pub type TcpConfig = GenTcpTransport<async_io::Tcp>;
 
 #[cfg(feature = "tokio")]
 pub use provider::tokio;
 
 /// The type of a [`GenTcpConfig`] using the `tokio` implementation.
 #[cfg(feature = "tokio")]
-pub type TokioTcpConfig = GenTcpConfig<tokio::Tcp>;
+pub type TokioTcpConfig = GenTcpTransport<tokio::Tcp>;
```
elenaf9: Should probably rename these to `TcpTransport` and `TokioTcpTransport`.

thomaseizinger:
> **Opened as a draf**t: I'd like to open the discussion on whether this direction is desired, and to discuss how this could be implement / how existing transport could be adapted to the change. Not ready for review yet.

Converted it to a draft for you :)

core/src/either.rs
```diff
@@ -493,26 +452,37 @@ where
...
+            },
+        }
+    }
+
     fn listen_on(
         &mut self,
+        id: ListenerId,
         addr: Multiaddr,
-    ) -> Result<Self::Listener, TransportError<Self::Error>> {
+    ) -> Result<(), TransportError<Self::Error>> {
```
thomaseizinger: I think `listen_on` should return a `ListenerId` rather than expect one.

Now, I think this is a bit difficult because of how `Transport`s compose. The IDs should obviously be unique but they might not be if we compose `Transport`s together and sometimes return an ID from an inner transport and sometimes our own.

What if we "namespace" `ListenerId`s? We could extend `ListenerId` with a `TypeId` of the `Transport` they are tracking. E.g. each `Transport` would be encouraged to construct a their "current" `ListenerId` with `ListenerId::new::<Self>()`.

That would assume that each kind of `Transport` is only composed once. I think that is generally true although not impossible to violate. On the whole, I think it is a fair assumption to make though:

To users of `Transport` we can guarantee that the `ListenerId` returned within `TransportEvent`s unambiguously maps to their call to `listen_on` _as long as_ they compose `Transport`s correctly.

core/src/transport.rs
thomaseizinger: > Reason for this is that a transport's stream should never terminate (never return `None`).

I think exposing a bare poll as the baseline is good. We shouldn't force users into a trait if not necessary.

> Q: Should we auto-implement `Stream` for `T: Transport`? Would ease polling if users use a Transport directly.

Yeah I would suggest we do. I might do the same for `StreamMuxer` over in https://github.com/libp2p/rust-libp2p/pull/2648 actually.

core/src/transport.rs
thomaseizinger: > This is not ideal for users that use `core/transport` directly outside of a swarm, but I don't think there's any other way to solve this.

I think it is perfectly fine. IMO any serious user of `Transport` should care about unambiguously mapping attempts of listening on an address to the outcome.

core/src/transport.rs
thomaseizinger: Whilst I understanding pinning in principle, I find it tricky to wrap my head around it every time. I am wondering though: We are _moving_ self here and constructing a trait object. I think because of the type erasure, it is not safe to move this type around unless we can guarantee that it is `Unpin`. Thus requiring this trait bound makes sense I think.

core/src/transport.rs
```diff
@@ -170,28 +170,28 @@ pub trait Transport {
...
+    // /// The returned transport will act like `self`, except that if `listen_on` or `dial`
+    // /// return an error then `other` will be tried.
+    // fn or_transport<U>(self, other: U) -> OrTransport<Self, U>
+    // where
+    //     Self: Sized,
+    //     U: Transport,
+    //     <U as Transport>::Error: 'static,
+    // {
+    //     OrTransport::new(self, other)
+    // }
```
thomaseizinger: Are these functions possible and you just didn't do them because it is a PoC or did you hit some limitation here?

core/src/transport.rs
```diff
@@ -221,149 +221,252 @@ pub trait Transport {
...
+    }
+}
+
+impl std::ops::Add<u64> for ListenerId {
+    type Output = Self;
+
+    fn add(self, other: u64) -> Self {
+        Self(self.0 + other)
+    }
+}
```
thomaseizinger: If we end up doing the namespace thing, I believe we should make this type completely opaque and remove any reference to `u64` from the public API.

core/src/transport.rs
```diff
@@ -221,149 +221,252 @@ pub trait Transport {
...
+        listener_id: ListenerId,
+        /// The new address that is being listened on.
+        listen_addr: Multiaddr,
+    },
+    /// A connection is incoming on one of the listeners.
+    Incoming {
+        /// The listener that produced the upgrade.
+        listener_id: ListenerId,
+        /// The produced upgrade.
+        upgrade: T::ListenerUpgrade,
```
thomaseizinger: It would be nice if this were already the final connection.

But that changes a few assumptions. For example, it means that parallelizing connection upgrades would be a concern of the `Transport` implementation and wouldn't be possible as part of `libp2p-swarm`.

On the other hand, it would allow us to simplify the `Transport` trait even more by allowing us to remove another associated type.

core/tests/transport_upgrade.rs
thomaseizinger: I'd be in favor yes.

swarm/src/lib.rs
thomaseizinger: I think that is a `Swarm` responsibility.

thomaseizinger:
> **Transport::dial**
> 
> I did not change how `Transport::dial` works. I originally planned to also remove `Transport::Dial` and push that logic into `Transport::poll` as well. The problem with this would be that it does not work well with the concurrent-dial logic. `ConcurrentDial` currently drives multiple dial futures (from potentially multiple transports) at the same time. Once the first succeeds, it drops the other ongoing dials. If the Transport would internally manage the `Dial` future, this logic would not work anymore. What we could do is something like the following:
> 
>  * Each dial is associated with a `DialId`, `ConcurrentDial` tracks the dial-ids of concurrent dials
>  * We add `Transport::abort_dial` that can be called from the `Pool` once the first dial of a concurrent dialing batch succeeds
> 
> I am not sure if this is worth the extra complexity. Happy to hear alternative suggestions.

I think from an API PoV, this would be better because it unifies handling of new connections to the caller of the `poll` function. At the moment, `listen_on` eventually triggers an `Incoming` event. In a similar manner, `dial` could trigger an `Outgoing` event that contains the actual `DialFuture`.

This way, concurrent dialling (and aborting the dial at any point) can still be implemented on top and is not a concern of the `Transport` implementation.

thomaseizinger (commented):
Thank you!

Strong concept ACK for going in this direction. Left some comments :)

core/src/transport.rs
elenaf9: I did not do them because it's a PoC.
I don't think there should be any issues when adapting `MapErr` or `OrTransport` (or any other transport for that matter). If the general design is approved the next TODO (that I forgot to add to the above list of TODOs) is adapting the remaining transports. This change should not be the reason to remove any existing transport.

core/src/transport.rs
mxinden: Agreed with everything above. 

At some point it might make sense to drive the `InfiniteStream` effort upstream, or introduce our own `InfiniteStream` trait  in rust-libp2p.

core/src/transport.rs
mxinden: > But that changes a few assumptions. For example, it means that parallelizing connection upgrades would be a concern of the `Transport` implementation and wouldn't be possible as part of `libp2p-swarm`.

Off the top of my head I prefer decisions on parallelism and concurrency to happen in a single place and not make it a per-transport decision. I think rust-libp2p greatly benefits from the fact that (1) we only spawn one task per connection and (2) everything else is driven by a single task.

core/src/transport.rs
```diff
@@ -221,149 +221,252 @@ pub trait Transport {
...
-    /// An upgrade, consisting of the upgrade future, the listener address and the remote address.
-    Upgrade {
-        /// The upgrade.
-        upgrade: TUpgr,
-        /// The local address which produced this upgrade.
+pub enum TransportEvent<T: Transport> {
+    /// A new address is being listened on.
+    NewAddress {
+        /// The listener that is listening on the new address.
+        listener_id: ListenerId,
```
mxinden, line 250: _Still need to give this more thought._

What is the use-case where someone needs to differentiate events by listener? In other words, what is the use-case where a user needs to know about the concept of listeners at all?

Don't users just want to be able to:

- Listen on an address X.
- Stop listening on an address X.

Is it wildcard (e.g. `/ip4/0.0.0.0`) addresses that make this difficult?

Tracking this thought as a comment here for now.

mxinden: //CC @rkuhn as you might need it for ipfs-embed.

core/tests/transport_upgrade.rs
mxinden: `impl<T: Transport> Stream for T` sounds good to me.

core/src/transport.rs
thomaseizinger: In general, I think it is a good idea to hand out unique identifiers in asynchronous systems so clients can unambiguously associate events with their actions. The actions in our case are `listen_on` and `dial` and the events are returned on `poll`.

> In other words, what is the use-case where a user needs to know about the concept of listeners at all?

I don't think they need to know about the concept of _listeners_ as such. That is an implementation detail to me. Perhaps we should be calling them `ListenRequestId`s?

I think it is risky to assume that we can go through all usecases and decide that for `Transport`s in particular, we can optimise this away but for the sake of an example, one usecase I can think of is:

Re-triggering a `listen_on` request upon failure. If I have previously issued 4 `listen_on` calls and one of them fails, I want to know, which one failed to accurately track that state. If `listen_on` doesn't hand me a unique identifier for each invocation, this is not possible.

core/src/transport.rs
thomaseizinger: Yes, I think that is fair. I've actually changed my mind on this. Returning the `Future` here makes more sense and also allows us to move the dialing into this. See https://github.com/libp2p/rust-libp2p/pull/2652#issuecomment-1128536102.

core/src/transport.rs
thomaseizinger: I guess what you are suggesting is that the address passed in by the user might already be unique? Wildcard addresses are one example that show that they might not be suitable identifiers. I think there might be others too.

Overall, I think uniformity is more important here than optimising individual situations. I really like how we are recently aligning different abstractions in the codebase.

rkuhn: ipfs-embed itself doesn’t care about ListenerIds at this point, but it does expose them to its users. The conclusions people should draw from this information are not very clear, though — most people will probably only be interested when an address starts and stops being listened on. What is the consequence of an error in a listener for the addresses this listener is servicing?

Currently ipfs-embed allows user code to start listening on an address, but there is no API to stop listening. Implementing this missing features will require some sort of ListenerId, will it not?

core/src/either.rs
elenaf9: Not sure if it's worth the extra complexity. I agree that it would be a cleaner API. But apart from that, what are the benefits of having the `ListenerId` generated by the transport?

core/src/transport.rs
elenaf9: I am undecided on this. 

> In general, I think it is a good idea to hand out unique identifiers in asynchronous systems so clients can unambiguously associate events with their actions. The actions in our case are listen_on and dial and the events are returned on poll.

Generally I agree with this. But then we should be consistent and also add an identifier for initiated dials (and include that in all related events).

On the other hand, I also like reducing the number of publicly exposed types where possible, for the sake of simplicity. And in the end users really only care about what addresses are currently active.

> I think it is risky to assume that we can go through all usecases and decide that for Transports in particular

Before https://github.com/libp2p/rust-libp2p/pull/2123 we also did not have a `ListenerId` included listener events. I actually did that PR back then to include it, but iirc that was mostly so we can expose it to our (Stronghold's) users.

> Re-triggering a listen_on request upon failure. If I have previously issued 4 listen_on calls and one of them fails, I want to know, which one failed to accurately track that state. If listen_on doesn't hand me a unique identifier for each invocation, this is not possible.

Upon failure we can include the multiaddr that the user gave as parameter in the event (even if it's a wildcard address). If the user calls listen_on e.g. 2 times with wildcard addresses and one of them fails, would they really care if it is the first or the second one that failed?  
 
But if we decided to remove the `ListenerId`, I think we should also remove the event variants `ListenerClosed` and `ListenerError`. For the former we emit `AddressExpired` for all affected addresses, and the latter is just informative (I would actually be interested if anyone really handles that case/ takes action?), so if anything we could just add a general `TransportError` variant.

---
 As already mentioned, I am really undecided here. Both directions are valid, so I am fine with either as long as we are consistent.

core/src/either.rs
thomaseizinger: I think it is easier for clients to use if we don't require additional state tracking outside of a `Transport`. I am coming from the perspective of using these APIs _without_ `libp2p-swarm` and would rather have the `Transport`s itself take care of this state tracking.

The advantage is that if the `Transport` takes care of the state tracking, it is optional to use for the client (could just ignore the return value) whereas if it is a parameter, I HAVE to pass one to make the API work!

Additionally, a client may get the parameters wrong, hence the API is overall more complex to use.

core/src/transport.rs
thomaseizinger: > But then we should be consistent and also add an identifier for initiated dials (and include that in all related events).

Yes. I'd be in favor of this because it also allows us to return the `DialFuture` as part of the events. See https://github.com/libp2p/rust-libp2p/pull/2652#issuecomment-1128536102.

thomaseizinger: I would really like to see events being the only source of new connections because it simplifies client code. Not only outside but also within `libp2p-swarm` :)

Whether or not we include IDs is a question of correctness IMO. I'd vote for including them because it is a a - in my opinion - small price to pay for added correctness and allows to serve more use cases.

Here is a sketch of how I envision the API (modulo errors and bike shedding on names etc):

```rust
pub trait Transport {
    type IncomingConnection;
    type OutgoingConnection;
    
    fn dial(&mut self, address: MultiAddress) -> Result<DialId>;
    fn listen_on(&mut self, address: MultiAddress) -> Result<ListenId>;
    
    fn poll(&mut self, cx: &mut Context<'_>) -> Poll<TransportEvent>;
}

pub enum TransportEvent<T: Transport> {
    NewIncomingConnection {
        listen_id: ListenId,
        connection: T::IncomingConnection
    },
    NewOutgoingConnection {
        dial_id: DialId,
        connection: T::OutgoingConnection
    }
}
```

core/src/either.rs
elenaf9: Good point. I only considered the `Transport` implementer's PoV and did not think about users that directly use transports without a swarm.

core/src/transport.rs
elenaf9: Definitely agree that this would be a very clean and unified API. 
However, I think it would add complexity to the swarm rather than simplify it. The way how concurrent dial currently works is that it calls `dial` for all addresses, pushes the obtained dial futures into a list and hands it to a newly spawned task that drives them. This all happens within the `Swarm::dial_with_handler` call.
If we change the transport according to your proposal, this would be more complicated because we don't obtain the futures until we poll the transport, which only happens in `Swarm::poll`. 
We could for example solve this by using a mpsc channel to "send over" the futures on every `TransportEvent::NewOutgoingConnection` from the swarm to the related spawned task. But I don't think that's very clean.
Also on the transport side I think a transport would on `Transport::dial` still initiate the future, but then have to hold it somewhere until `poll` is called.

That being said, I am generally not opposed to your proposal. I very much appreciate that you are pushing for a simple, unified API. I just think that we have to be careful that we are actually making the implementation and integration easier with this. But maybe we can come up with the clean solution that works well with the concurrent dialing? Interested to hear your ideas @thomaseizinger!

core/src/transport.rs
elenaf9: Added remaining transports from core in d6f0e75.

core/tests/transport_upgrade.rs
elenaf9: > impl<T: Transport> Stream for T sounds good to me

Rust unfortunately does not allow this (I forget that every single time). 
I considered adding a `StreamTransport` that just wraps a transport and implemented `Stream`, but since we already have the `Boxed` type I just implemented it for `Boxed` and adjusted the tests (bfd5fb0). <s> Also implemented Stream for `GenTcpTransport` (90721b9) because you can't box a TCP transport directly.</s> Edit: with d7f5019 we can now also box the tcp transport and don't need an extra Stream impl anymore.

elenaf9:
While trying to adapt `dns::GenDnsConfig` and `websockets::framed::WsConfig` I noticed that they wrap the inner transport in a Mutex. @mxinden if I understand it correctly this was done as part of https://github.com/libp2p/rust-libp2p/pull/2529 so that they can still implement `Clone`? I wonder if it is really necessary for them to be clone? Don't think it makes sense after this PR's change. When implementing `poll` on it it becomes a race condition when a waker is triggered which of the clones acquires the lock first and thus gets the `TransportEvent`.

mxinden:
> While trying to adapt `dns::GenDnsConfig` and `websockets::framed::WsConfig` I noticed that they wrap the inner transport in a Mutex. @mxinden if I understand it correctly this was done as part of #2529 so that they can still implement `Clone`? I wonder if it is really necessary for them to be clone?

Expanding on this for the case of DNS. Websockets should be analogous.

On `<GenDnsConfig as Transport>::dial` the method returns a dialing `Future`. That `Future`, when polled, first resolves the DNS addresses in the provided `Multiaddr`, if any. It then calls `Transport::dial` on its `inner` `Transport`, e.g. `libp2p-tcp`. For this later step it needs a clone of the `inner` transport within the dialing `Future`.

Does the above make sense? Happy to expand further.

We could return a `DialId` in `Transport::dial` only, driving the DNS resolution in `Transport::poll` and only once resolved return the dialing `Future` which now only has to establish the TCP connection. That said, I don't think we should do this work in the main event loop (i.e. within `Transport::poll`), but instead in the per-connection-task.

I can not think of a way to get rid of the `Arc<Mutex<_>>` while still driving the DNS resolution on the per-connection-task instead of the main event loop.

core/src/transport.rs
mxinden: > We could for example solve this by using a mpsc channel to "send over" the futures on every `TransportEvent::NewOutgoingConnection` from the swarm to the related spawned task. But I don't think that's very clean.

Agreed.

mxinden:
If I am not mistaken the efforts of (1) refactoring `Transport::listen` and (2) refactoring `Transport::dial` can be done in two pull requests, i.e. in two in itself atomic changes, correct?

In addition, I am sensing that we have rough consensus on (1), namely of getting rid of listeners. Also, please correct me if I am wrong, we agree on (1) being a step forward even if we don't do (2).

With all of the above in mind, I suggest we do (1) only for now. Delaying (2) to another pull request.

What do you think?

mxinden:
> Don't think it makes sense after this PR's change. When implementing `poll` on it it becomes a race condition when a waker is triggered which of the clones acquires the lock first and thus gets the `TransportEvent`.

Haven't thought of this. All the clones only ever call `Transport::dial`, never `Transport::poll`, thus in practice this is not an issue, correct? Ideally this would be enforced at the type level though :/

elenaf9:
> On <GenDnsConfig as Transport>::dial the method returns a dialing Future. That Future, when polled, first resolves the DNS addresses in the provided Multiaddr, if any. It then calls Transport::dial on its inner Transport, e.g. libp2p-tcp. For this later step it needs a clone of the inner transport within the dialing Future.
>
> Does the above make sense? Happy to expand further.

Yes makes sense, thanks! I agree then that we need the `Mutex`.

> > Don't think it makes sense after this PR's change. When implementing poll on it it becomes a race condition when a waker is triggered which of the clones acquires the lock first and thus gets the TransportEvent.
>
> Haven't thought of this. All the clones only ever call Transport::dial, never Transport::poll, thus in practice this is not an issue, correct? Ideally this would be enforced at the type level though :/

We could at least prevent it for users of `GenDnsConfig` (but not within its implementation). We can remove the `Clone` derivation for `GenDnsConfig`, but keep the Mutex around the inner transport. That way within one `GenDnsConfig` we can still clone the inner transport, but not `GenDnsConfig` itself, which prevents the described race condition.

core/src/transport.rs
thomaseizinger: > Also on the transport side I think a transport would on `Transport::dial` still initiate the future, but then have to hold it somewhere until `poll` is called.

That is a problem many implementations of `NetworkBehaviour` have to deal with and is typically solved with a `VecDeque` in here. I agree that it is not pretty but I think an unavoidable consequence of such an asynchronous design :/

> However, I think it would add complexity to the swarm rather than simplify it. The way how concurrent dial currently works is that it calls `dial` for all addresses, pushes the obtained dial futures into a list and hands it to a newly spawned task that drives them. This all happens within the `Swarm::dial_with_handler` call.
> If we change the transport according to your proposal, this would be more complicated because we don't obtain the futures until we poll the transport, which only happens in `Swarm::poll`.

Thanks for digging into the details, I haven't looked at those parts yet!

Nothing particularly novel comes to my mind at the moment but some areas (these are different ideas) worth exploring could be:

- Delay the call to `self.pool.add_outgoing` until we've collected all futures returned from `Transport::poll`
- Implement concurrent dialing through the pool directly by adding the futures from `TransportEvent::NewOutgoingConnection` individually to the pool
- Move the establishing of connections from `Pool` to a wrapper around `Transport` to also encapsulate (and thus simplify) some of the state required for concurrent dialing. The `Pool` when then only deal with established connections and drive those in individual tasks.

To move forward with this, I think we need more reasons other than this being the "cleaner API". I wonder whether the discussion below about having to `Arc` certain inner transports could be resolved by a design along these lines. Although if we want concurrency, sharing a reference to the inner transport may be unavoidable.

core/src/either.rs
thomaseizinger: More generally, having clients decide on identifiers is really useful if idempotency is required. If not, I'd default to having the "invoked system" (for lack of a better word) generate the identifier.

HTTP POST does this for example or databases and tables with primary keys.
I am guessing that one of the design choices for TCP's sequence number being client-generated is because it needs to replay packets in case they are not ACK'd: Idempotency.

transports/tcp/src/lib.rs
```diff
@@ -425,14 +481,111 @@ where
...
+        mut self: Pin<&mut Self>,
+        cx: &mut Context<'_>,
+    ) -> Poll<TransportEvent<Self::ListenerUpgrade, Self::Error>> {
+        // Return pending events from closed listeners.
+        if let Some(event) = self.pending_events.pop_front() {
+            return Poll::Ready(event);
+        }
+        // We remove each element from `listeners` one by one and add them back.
+        let mut remaining = self.listeners.len();
+        while let Some(mut listener) = self.listeners.pop_back() {
```
mxinden, line 520: Could we make `self.listeners` a `FuturesUnordered`?



While we could not remove listeners, we could instead access them in a mutable fashion via [`iter_mut`](https://docs.rs/futures/latest/futures/stream/struct.FuturesUnordered.html#method.iter_mut) and have them resolve instantly instead.

Just a thought. I think this is out of scope for this pull request.

core/src/transport.rs
mxinden: Unresolving here to continue the discussion whether the `ListenerId` as an argument is needed here.

If I understand correctly, `ListenerId` is passed as an argument to prevent to `Transport` implementations to have clashing IDs, correct?

Do we depend on the monotonicity of the ID anywhere? Asked differently, could the `ListenerId` as well be a random e.g. `u64`? With this approach, there is no need to pass the `ListenerId` as an argument, but instead can be generated by each `Transport` implementations without the need for synchronization.

thomaseizinger: > could the `ListenerId` as well be a random e.g. `u64`

I would be in favor of making it an opaque type. We should probably remove the `PartialOrd` and `Ord` derives if we generate them randomly.

transports/dns/src/lib.rs
```diff
@@ -197,14 +197,11 @@ where
...
-        &mut self,
-        id: ListenerId,
-        addr: Multiaddr,
-    ) -> Result<(), TransportError<Self::Error>> {
+    fn listen_on(&mut self, addr: Multiaddr) -> Result<ListenerId, TransportError<Self::Error>> {
         self.inner
             .lock()
-            .listen_on(id, addr)
+            .listen_on(addr)
+            .map(ListenerId::map_type::<Self>)
```
thomaseizinger: Interesting. I did not anticipate the need for this. Do you mind explaining this in more detail?

core/src/transport.rs
elenaf9: With 2226092 I followed @thomaseizinger idea (https://github.com/libp2p/rust-libp2p/pull/2652#discussion_r874461833) to namespace `ListenerId`s to avoid clashing. The advantage of this implementation is that we can keep the monotonicity that I think all other Ids in libp2p have. Drawback is that it's a bit more complicated when we wrap inner transports. I have implemented it now that transports in `/transports` that wrap an inner transport with additional logic (like `WsConfig` and `DnsConfig`) change the TypeId of ListenerIds from inner transports, but wrapper transport in `core/transport` just pass it on. For the former case it is needed so user can compose a transport multiple times with different wrappers (like our development_transport does it with dns-tcp and ws-dns-tcp), for the latter I don't think the TypeId should be changed since the "actual" transport is still the same.
Also, wrapper transports that map the ListenerId need to also map it for all TransportEvent, but unfortunately this is not enforced on a type level which make is rather error-prone.

Alternative of making it a random id would also work and safe us a bit of complexity, but I am also content with the current solution.

Don't have a strong opinion here, I am fine with either. Wdyt @thomaseizinger @mxinden.

swarm/src/lib.rs
```diff
@@ -1067,7 +1067,7 @@ where
                         let mut parameters = SwarmPollParameters {
                             local_peer_id: &this.local_peer_id,
                             supported_protocols: &this.supported_protocols,
-                            listened_addrs: &this.listened_addrs,
+                            listened_addrs: this.listened_addrs.values().flatten().collect(),
```
thomaseizinger, line 1067: I think you may be able to get away with returning a `Flatten<Values<'a>>` iterator here?

elenaf9: `SwarmPollParamters` implements `PollParameters`, which requires for `PollParameters::ListenedAddressesIter` to implement `ExactSizeIterator`, which unfortunately is not implemented for `Flatten`.

core/src/transport.rs
thomaseizinger: > For the former case it is needed so user can compose a transport multiple times with different wrappers (like our development_transport does it with dns-tcp and ws-dns-tcp)

We never clone "leaf" transports like Tcp right? And only these transports generate `ListenerId`s. As long as we don't clone such a leaf transport, shouldn't the generated `ListenerId`s still be unique without mapping the type?

An alternative solution would be to offer a way of combining `ListenerId`s and even wrappers generate their own `ListenerId`s but combine them with the ones they get returned from inner transports. All of this seems very complicated though so perhaps choosing them randomly is better :)

swarm/src/lib.rs
thomaseizinger: Damn, thanks for the explanation.

transports/dns/src/lib.rs
thomaseizinger: Resolved in a different thread.

core/src/transport.rs
elenaf9: > We never clone "leaf" transports like Tcp right? And only these transports generate ListenerIds. As long as we don't clone such a leaf transport, shouldn't the generated ListenerIds still be unique without mapping the type?

Yes we clone "leaf" transports like tcp. For example in `development_transport`, we create two tcp transports wrapped in a dns transport, and then wrap one in a  WsConfig and combine them in a `OrTransport`. If we would only bubble up the ids generated by the tcp transport, both "or_transport"-ed transports would generate the same Ids:
https://github.com/libp2p/rust-libp2p/blob/4aa84bf7cd78c8553294f4f352fce9e20b6fbd4e/src/lib.rs#L203-L208

> An alternative solution would be to offer a way of combining ListenerIds and even wrappers generate their own ListenerIds but combine them with the ones they get returned from inner transports

Why not just the current solution that replaces the TypeId? I think two wrappers of the same type, but with different inner transports (e.g. `DnsConfig<A>` and `DnsConfig<B>`) will have different TypeIds so that should be enough, no?

core/src/either.rs
elenaf9: Resolving here since this discussion moved to another comment https://github.com/libp2p/rust-libp2p/pull/2652#discussion_r884311198. We agree that the inner Transport should generate the `ListenerId`, now it's more a questions of how the `ListenerId` API should look like.

transports/tcp/src/lib.rs
elenaf9: Not sure if I am missing something, but how would this work with the listeners being Streams and not Futures? 
But I just saw that there is [futures::stream::SelectAll](https://docs.rs/futures/latest/futures/stream/select_all/struct.SelectAll.html#); maybe we could use that?

> While we could not remove listeners, we could instead access them in a mutable fashion via [`iter_mut`](https://docs.rs/futures/latest/futures/stream/struct.FuturesUnordered.html#method.iter_mut) and have them resolve instantly instead.

How would we know which future belongs to which ListenerId?

core/src/transport.rs
thomaseizinger: > Yes we clone "leaf" transports like tcp. For example in `development_transport`, we create two tcp transports wrapped in a dns transport, and then wrap one in a WsConfig and combine them in a `OrTransport`. If we would only bubble up the ids generated by the tcp transport, both "or_transport"-ed transports would generate the same Ids:

Looking at the implementation of `DnsConfig`, the inner transport is `Arc<Mutex>`'d, meaning there is only one inner tcp transport, regardless how many times the wrapped DNS transport is cloned. Consequently, there is only one source of `ListenerId`s and there should never be a collision.

> > An alternative solution would be to offer a way of combining ListenerIds and even wrappers generate their own ListenerIds but combine them with the ones they get returned from inner transports
> 
> Why not just the current solution that replaces the TypeId? I think two wrappers of the same type, but with different inner transports (e.g. `DnsConfig<A>` and `DnsConfig<B>`) will have different TypeIds so that should be enough, no?

Yes, it will be enough but I think it would be better if we can find a solution that doesn't require knowledge of the implementation details of `ListenerId` at various places, i.e. when wrapping a `Transport`. Like you said elsewhere, it is not enforced through the type system so chances of it being forgotten are definitely non-zero.

Taking a step back: With transports getting a `poll` function, cloning them should certainly be the exception and if you want to clone one, you will need to `Arc<Mutex>` it. From this perspective, every transport that actually creates listeners (what I called leaf transports) should only ever exist once.

src/lib.rs
```diff
@@ -201,10 +201,13 @@ pub async fn development_transport(
...
-        let dns_tcp = dns::DnsConfig::system(tcp).await?;
-        let ws_dns_tcp = websocket::WsConfig::new(dns_tcp.clone());
-        dns_tcp.or_transport(ws_dns_tcp)
+        let dns_tcp = || {
+            let tcp_config = tcp::GenTcpConfig::default().nodelay(true);
+            let tcp = tcp::TcpTransport::new(tcp_config);
+            dns::DnsConfig::system(tcp)
+        };
+        let ws_dns_tcp = websocket::WsConfig::new(dns_tcp().await?);
+        dns_tcp().await?.or_transport(ws_dns_tcp)
```
elenaf9: Opening a discussion thread here related to what @thomaseizinger wrote in https://github.com/libp2p/rust-libp2p/pull/2652#discussion_r884856191
Since the original discussion was about the ListenerId I am moving this discussion here.

> > Yes we clone "leaf" transports like tcp. For example in development_transport, we create two tcp transports wrapped in a dns transport, and then wrap one in a WsConfig and combine them in a OrTransport. If we would only bubble up the ids generated by the tcp transport, both "or_transport"-ed transports would generate the same Ids:
>
> Looking at the implementation of DnsConfig, the inner transport is Arc<Mutex>'d, meaning there is only one inner tcp transport, regardless how many times the wrapped DNS transport is cloned. Consequently, there is only one source of ListenerIds and there should never be a collision.

Within this PR I am removing the Clone derivation on `DnsConfig` and `WsConfig`, so with the code here I would actually not clone a single `Mutex`'ed TCP transport but instead create two separate TCP Transports. See the above implementation for construction two dns_tcp transports.
I think we have to do it this way because otherwise the race condition of polling the inner transport will cause errors.
For example in WsConfig:
We wrap an tcp transport, and for all addresses that have a `/ws` or `/wss` suffix we apply some logic, more specifically we apply logic on the `TransportEvents` associated with `ws` or `wss` listeners. In particular, we map the inbound upgrade on `TransportEvent::Incoming`. Now the problem is that because we are "second" in an `OrTransport`, any event that is happening on the inner TCP Transport will never make it to use. Because if the waker is polled, the "first" transport, that is the non-websocket one, will be polled first, poll the inner tcp transport and get the event.

@mxinden @thomaseizinger because of this, I think we can never allow Clone on a transport in general, and thus users will have to build the same transport multiple times. Please correct me in case that I am missing something.

Edit: Related PR #2682.

core/src/transport.rs
elenaf9: I have moved the discussion about Cloning transports to a separate thread https://github.com/libp2p/rust-libp2p/pull/2652#discussion_r884914910. This way we can keep this thread about the ListenerId API and come back to it once we agree on whether we can / have to construct multiple transports of the same type or not.

elenaf9:
Marking the PR as ready for review. There a still some things missing (see below) but if someone (@mxinden, @thomaseizinger?) has time for a more in-depth review already it would appreciated. Especially since basically every transport needs to be reviewed that I did not introduce any errors while adapting it. On my side I'll focus next on how the QUIC implementation can be integrated with this change. 

 **Quick summary of the Status Quo:**

- `core/transport/`: All transports adapted to the change
 - `transports/`: All transports apart from `wasm-ext` adapted to the change
 - All tests / examples adapted to the change (apart from the transports that are not done yet)
- Todo:
  - [x] Adapt `transports/wasm-ext`
  - [x] Adapt `relay::v2::client::ClientTransport`
  - [x] Fix docs
  - [x] Adapt QUIC implementation  (#2289)

**Applied from first round of reviews**:
- `Transport::listen_on` now returns a `ListenerId` instead of expecting it as parameter.
- The swarm is responsible for tracking listening addresses. As a result I also removed the `addresses` parameter from `TransportEvent::ListenerClosed` since this would force the transport to also track the addresses. We can just insert that information in the swarm when creating the `SwarmEvent::ListenerClosed`
- Implement `Stream` for `Boxed` so that we can poll transport more conveniently. This means that we have to box the transport first, which is now done in almost all tests, but I don't see a reason for not doing this.

src/lib.rs
thomaseizinger: Yep, I think you are correct. It also feels more natural to _not_ clone `Transport`s, now that they have `&mut` receivers.

core/src/transport.rs
thomaseizinger: Assuming we can't clone them, then I'd be in favor of exploring the idea of randomly generate `ListenerId`s because it will result in a smaller API.

core/src/transport/memory.rs
```diff
@@ -221,6 +251,56 @@ impl Transport for MemoryTransport {
...
+
+    fn poll(
+        mut self: Pin<&mut Self>,
+        cx: &mut Context<'_>,
+    ) -> Poll<TransportEvent<Self::ListenerUpgrade, Self::Error>>
+    where
+        Self: Sized,
+    {
+        let mut remaining = self.listeners.len();
+        while let Some(mut listener) = self.listeners.pop_back() {
```
mxinden, line 254: As you mentioned before, moving this into a  shared abstraction might be worth exploring :+1:

src/lib.rs
mxinden: For the record https://github.com/libp2p/rust-libp2p/pull/2682 is merged and thus I think this is resolved?

transports/tcp/src/lib.rs
mxinden: Discussed out of band.

> But I just saw that there is [futures::stream::SelectAll](https://docs.rs/futures/latest/futures/stream/select_all/struct.SelectAll.html#); maybe we could use that?

:+1: 



> > While we could not remove listeners, we could instead access them in a mutable fashion via [`iter_mut`](https://docs.rs/futures/latest/futures/stream/struct.FuturesUnordered.html#method.iter_mut) and have them resolve instantly instead.
> 
> How would we know which future belongs to which ListenerId?

We would push a wrapper type into the `SelectAll`, e.g.:

```rust
struct ListenerWrapper {
  id: ListenerId,
  terminate: bool,
  listener: _,
}
```

`ListenerWrapper` would implement `Stream` and if `self.terminate` is `false` delegate to `self.listener` and otherwise return `Poll::Ready(None)`.

Removing a listener is means iterating the list and setting the flag to `true` on the matching one.

mxinden (commented):
Thanks for pushing this!

As discussed out-of-band, following up on https://github.com/libp2p/rust-libp2p/pull/2652#discussion_r875694730  I think it is worth exploring whether we can actually do without the `ListenerId` concept entirely, identifying by `Multiaddr` only.

I still need to give this a more in-depth review.

thomaseizinger:
> As discussed out-of-band, following up on #2652 (comment) I think it is worth exploring whether we can actually do without the ListenerId concept entirely, identifying by Multiaddr only.

FWIW, I am in favor of this as long as it doesn't introduce ambiguity.

elenaf9:
> As discussed out-of-band, following up on https://github.com/libp2p/rust-libp2p/pull/2652#discussion_r875694730 I think it is worth exploring whether we can actually do without the ListenerId concept entirely, identifying by Multiaddr only.

Opened PR https://github.com/elenaf9/rust-libp2p/pull/4 to discuss it separately, would be great to have your review @mxinden @thomaseizinger.

core/src/transport.rs
elenaf9: Done in a357d71.

elenaf9:
Removing the `ListenerId` proved to be problematic for `websocket::WsConfig` since we need to track there what listening address belongs to which inner listener (see https://github.com/elenaf9/rust-libp2p/pull/4#issuecomment-1160571457). We can still add this in a follow-up PR after this one, but for now I decided for the simplest solution, which is having random ListenerIds generated within the transport as proposed in https://github.com/libp2p/rust-libp2p/pull/2652#discussion_r884311198 by @mxinden.

core/src/transport/memory.rs
elenaf9: 👍
To not delay this PR further, I would follow up on this in a separate PR after this one is merged. Wdyt?

core/src/transport.rs
elenaf9: Sorry for the late reply here. If I understand it correctly @thomaseizinger, we agree that:
- Returning the dial future via a `TransportEvent` does not work well with the `ConcurrentDial` implementation and there is no trivial solution for this
- We don't have any reason for this change other that it being the "cleaner API"

Thus I would stick with returning the future directly in `Transport::dial`. Please correct me if I am missing something.

thomaseizinger: Yep, works for me!

core/src/transport/memory.rs
```diff
@@ -418,28 +455,25 @@ mod tests {
...
-            .is_ok());
-        assert!(transport
-            .listen_on("/memory/1639174018481".parse().unwrap())
-            .is_ok());
+
+        let addr_1: Multiaddr = "/memory/1639174018481".parse().unwrap();
+        let addr_2: Multiaddr = "/memory/8459375923478".parse().unwrap();
+
+        let listener_id_1 = transport.listen_on(addr_1.clone()).unwrap();
+        assert!(transport.remove_listener(listener_id_1));
```
thomaseizinger: Perhaps we could add panic messages here to make it a bit easier to understand, what is being asserted.

core/src/transport/optional.rs
```diff
@@ -28,7 +29,8 @@ use multiaddr::Multiaddr;
 /// enabled (read: contains `Some`), then dialing and listening will be handled by the inner
 /// transport.
 #[derive(Debug, Copy, Clone)]
-pub struct OptionalTransport<T>(Option<T>);
+#[pin_project::pin_project]
```
thomaseizinger, line 32: This is necessary because we don't want to require `Unpin` everywhere, correct?

swarm/src/lib.rs
```diff
@@ -831,10 +829,13 @@ where
 
     fn handle_listeners_event(
```
```suggestion
    fn handle_transport_event(
```

swarm/src/lib.rs
```diff
@@ -973,11 +977,11 @@ where
...
                 let translated_addresses = {
-                    let transport = self.listeners.transport();
                     let mut addrs: Vec<_> = self
-                        .listeners
-                        .listen_addrs()
-                        .filter_map(move |server| transport.address_translation(server, &address))
+                        .listened_addrs
+                        .values()
+                        .flatten()
+                        .filter_map(|server| self.transport.address_translation(server, &address))
```
thomaseizinger, line 985: Unrelated to this PR but `translate_address` might be a more natural name for this function.

transports/websocket/src/framed.rs
```diff
@@ -277,14 +163,99 @@ where
...
+    fn poll(
+        mut self: Pin<&mut Self>,
+        cx: &mut Context<'_>,
+    ) -> Poll<libp2p_core::transport::TransportEvent<Self::ListenerUpgrade, Self::Error>> {
+        let mut transport = self.transport.lock();
+        let inner_event = match Transport::poll(Pin::new(transport.deref_mut()), cx) {
+            Poll::Ready(ev) => ev,
+            Poll::Pending => return Poll::Pending,
+        };
+        drop(transport);
```
thomaseizinger: A way to avoid the explicit drop is to create a smaller, inner scope but that is a style-preference :)

thomaseizinger (commented):
It is a big diff but I gave everything a decent read, looking forward to these changes landing!

core/src/transport/memory.rs
mxinden: Works for me. :+1:

core/src/transport/optional.rs
elenaf9: Yes

core/src/transport/memory.rs
elenaf9: Done in b19e11a.

swarm/src/lib.rs
elenaf9: Done in b19e11a.

transports/websocket/src/framed.rs
elenaf9: Good idea; done in b19e11a.

protocols/relay/src/v2/client/transport.rs
```diff
@@ -217,6 +225,35 @@ impl Transport for ClientTransport {
...
+                        let _ = self.to_behaviour.start_send(msg);
+                        continue;
+                    }
+                    Poll::Ready(Err(_)) => unreachable!("Receiver is never dropped."),
+                    Poll::Pending => {}
+                }
+            }
+            match self.listeners.poll_next_unpin(cx) {
+                Poll::Ready(Some(event)) => return Poll::Ready(event),
+                _ => return Poll::Pending,
```
elenaf9, line 253: This also returns Poll::Pending in case of `Poll::Ready(None)`, in which case no waker is informed.
`Poll::Ready(None)` will only happen if we currently don't have a listener. If we add a new listener again with `listen_on`, I would assume we poll the transport afterwards again, thus registering the waker with the new listener.  Thus I think it's okay to do this here (same in the wasm-ext transport).

core/src/either.rs
```diff
@@ -385,11 +343,12 @@ impl<A: ProtocolName, B: ProtocolName> ProtocolName for EitherName<A, B> {
         }
     }
 }
-
+#[pin_project(project = EitherTransportProj)]
 #[derive(Debug, Copy, Clone)]
+#[must_use = "futures do nothing unless polled"]
```
mxinden:
```suggestion
#[must_use = "transports do nothing unless polled"]
```

Not quite a `Future` but rather a `Transport`, right?

core/src/transport.rs
```diff
@@ -144,6 +144,14 @@ pub trait Transport {
     where
         Self: Sized;
 
+    // TODO: Add docs
+    fn poll(
+        self: Pin<&mut Self>,
+        cx: &mut Context<'_>,
+    ) -> Poll<TransportEvent<Self::ListenerUpgrade, Self::Error>>
+    where
+        Self: Sized;
```
mxinden: Probably unrelated to the changes here, but where is this requirement coming from?

core/src/transport.rs
```diff
@@ -221,149 +228,276 @@ pub trait Transport {
...
+impl Default for ListenerId {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+/// Event produced by [`Transport`]s.
 ///
-/// Transports are expected to produce `Upgrade` events only for
+/// Transports are expected to produce `Incoming` events only for
```
mxinden:
```suggestion
/// Transports are expected to produce [`TransportEvent::Incoming`] events only for
```

core/src/transport.rs
```diff
@@ -221,149 +228,276 @@ pub trait Transport {
...
 /// listen addresses which have previously been announced via
 /// a `NewAddress` event and which have not been invalidated by
```
mxinden:
```suggestion
/// a [`TransportEvent::NewAddress`] event and which have not been invalidated by
```

core/src/transport.rs
```diff
@@ -221,149 +228,276 @@ pub trait Transport {
...
 /// an `AddressExpired` event yet.
```
mxinden:
```suggestion
/// an [`TransportEvent::AddressExpired`] event yet.
```

core/src/transport.rs
```diff
@@ -221,149 +228,276 @@ pub trait Transport {
...
-    AddressExpired(Multiaddr),
-    /// A non-fatal error has happened on the listener.
+    /// A listener errored.
     ///
-    /// This event should be generated in order to notify the user that something wrong has
-    /// happened. The listener, however, continues to run.
-    Error(TErr),
+    /// The listener will continue to be polled for new events and the event
+    /// is for informational purposes only.
+    Error {
```
mxinden: This is strictly scoped to listener errors, right? How about:

```suggestion
    ListenerError {
```

core/src/transport/and_then.rs
```diff
@@ -21,15 +21,17 @@
...
     connection::{ConnectedPoint, Endpoint},
     either::EitherError,
-    transport::{ListenerEvent, Transport, TransportError},
+    transport::{ListenerId, Transport, TransportError, TransportEvent},
 };
 use futures::{future::Either, prelude::*};
 use multiaddr::Multiaddr;
 use std::{error, marker::PhantomPinned, pin::Pin, task::Context, task::Poll};
 
 /// See the `Transport::and_then` method.
```
mxinden:
```suggestion
/// See the [`Transport::and_then`] method.
```

Hope you don't mind all the minor unrelated suggestions. Feel free to ignore in case you think it adds to much noise to the pull request.

core/src/transport/memory.rs
```diff
@@ -245,51 +316,17 @@ impl error::Error for MemoryTransportError {}
...
 pub struct Listener {
+    id: ListenerId,
     /// Port we're listening on.
     port: NonZeroU64,
     /// The address we are listening on.
     addr: Multiaddr,
     /// Receives incoming connections.
     receiver: ChannelReceiver,
-    /// Generate `ListenerEvent::NewAddress` to inform about our listen address.
+    /// Generate `TransportEvent::NewAddress` to inform about our listen address.
```
mxinden:
```suggestion
    /// Generate [`TransportEvent::NewAddress`] to inform about our listen address.
```

Throws a docs compile error in case we ever rename this event.

protocols/dcutr/examples/client.rs
```diff
@@ -95,7 +95,10 @@ fn main() -> Result<(), Box<dyn Error>> {
 
     let transport = OrTransport::new(
         relay_transport,
-        block_on(DnsConfig::system(TcpConfig::new().port_reuse(true))).unwrap(),
+        block_on(DnsConfig::system(TcpTransport::new(
```
mxinden, line 98: What do you think of splitting `DnsConfig` into `DnsTransport` and `DnsConfig` like you did for `TcpConfig`?

protocols/relay/src/v2/client/transport.rs
```diff
@@ -147,25 +149,31 @@ impl Transport for ClientTransport {
...
-        })
+            is_closed: false,
+        };
+        self.listeners.push(listener);
+        Ok(listener_id)
+    }
+
+    fn remove_listener(&mut self, id: ListenerId) -> bool {
+        if let Some(listener) = self.listeners.iter_mut().find(|l| l.listener_id == id) {
+            listener.close(Ok(()));
```
mxinden, line 172: :rocket:

protocols/relay/src/v2/client/transport.rs
mxinden: > If we add a new listener again with `listen_on`, I would assume we poll the transport afterwards again,

:+1: that is the case today in `libp2p-swarm`.

We have to be careful in the future. E.g. in case we ever use some intelligent polling mechanism that only polls the underlying component if the passed sub-waker was woken (similar to `FuturesUnordered` and `SelectAll`), this transport would not be polled again once returning `Poll::Pending here.

swarm/src/lib.rs
```diff
@@ -1231,21 +1235,21 @@ where
...
+    type ListenedAddressesIter = std::iter::Cloned<std::vec::IntoIter<&'a Multiaddr>>;
     type ExternalAddressesIter = AddressIntoIter;
 
     fn supported_protocols(&self) -> Self::SupportedProtocolsIter {
         self.supported_protocols.iter().cloned()
     }
 
     fn listened_addresses(&self) -> Self::ListenedAddressesIter {
-        self.listened_addrs.iter().cloned()
+        self.listened_addrs.clone().into_iter().cloned()
```
mxinden, line 1252:
```suggestion
        self.listened_addrs.clone().into_iter()
```

Is this needed? Isn't it already an owned iterator?

mxinden (commented):
Large change. Haven't yet reviewed all. Still need to review the `transports/` folder. Good work!

core/src/transport/and_then.rs
elenaf9: Don't mind them at all. Thanks!

swarm/src/lib.rs
elenaf9: `self.listened_address` is an `Vec<&'a Multiaddr>`, but we need `ExactSizeIterator<Item = Multiaddr>`.
So the first clone is to make at an owned iterator, the second `cloned` is for `Item = &Multiaddr` -> `Item = Multiaddr`.

In the end what we need is to create a `ExactSizeIterator<Item = Multiaddr>` from a `HashMap<ListenerId, SmalLVec<Multiaddr>>`. The current implementation is the only one I was able to come up with, but I'd be grateful for any ideas of how to do this cleaner.

protocols/dcutr/examples/client.rs
elenaf9: Not sure if it makes much sense for `DnsConfig`. It does not really have any config parameters; just the `DnsConfig::custom` constructor for a custom resolver. But not a strong opinion. 
If we decide to do it I'd be in favor of making it a separate follow-up PR and to do it for `WsConfig` as well.

core/src/transport.rs
elenaf9: Actually: it's not needed at all. Same for some of the other methods where this bound existed already. 
When I first added the `poll` method I believe I still got some error, but I can not reproduce it anymore.
Removed it now, everything still compiles and tests pass.

protocols/relay/src/v2/client/transport.rs
elenaf9: The alternative would be to store the waker and call it once we add a new listener. 
Don't have a strong opinion here; happy to change it if you prefer the latter.

swarm/src/lib.rs
mxinden: Thanks for detailing this! You are right.

> The current implementation is the only one I was able to come up with, but I'd be grateful for any ideas of how to do this cleaner.

Played around with this as well. Couldn't come up with something better. Also, performance wise, I don't think it matters. And long term I would like to get rid of `PollParameters`, though I don't see this happening any time soon.

protocols/relay/src/v2/client/transport.rs
mxinden: This is not the only place which we will have to consider in case we ever do the advanced polling. Thus, I think we can keep this as is for now, being consistent with the remaining code base. Thanks @elenaf9!

core/src/transport.rs
mxinden: >  Removed it now, everything still compiles and tests pass.

Great to see this simplification. :pray:

protocols/dcutr/examples/client.rs
mxinden: :+1: Let's defer.

protocols/relay/src/v2/client/transport.rs
```diff
@@ -320,12 +320,21 @@ fn parse_relayed_multiaddr(
 
 pub struct RelayListener {
     listener_id: ListenerId,
+    /// Queue of events to report when polled.
     queued_events: VecDeque<<Self as Stream>::Item>,
+    /// Channel for messages from the behaviour [`Handler`][super::handler::Handler].
```
mxinden, line 325:
```suggestion
    /// Channel for messages from the behaviour [`Client`][super::Client].
```

Why reference the handler here? Wouldn't the `NetworkBehaviour` implementation be more appropriate?

transports/tcp/Cargo.toml
```diff
@@ -20,6 +20,8 @@ ipnet = "2.0.0"
 libc = "0.2.80"
 libp2p-core = { version = "0.34.0", path = "../../core", default-features = false  }
 log = "0.4.11"
+pin-project = "1.0.0"
```
mxinden:
```suggestion
```

Needed? Seems to compile on my machine without.

transports/tcp/Cargo.toml
```diff
@@ -20,6 +20,8 @@ ipnet = "2.0.0"
...
+smallvec = "1.6.1"
```
mxinden:
```suggestion
```

Needed? Seems to compile on my machine without.

transports/tcp/src/lib.rs
```diff
@@ -238,40 +232,40 @@ where
...
     /// > of a peer per local listening socket address.
     ///
-    /// `GenTcpConfig` keeps track of the listen socket addresses as they
-    /// are reported by polling [`TcpListenStream`]s obtained from
-    /// [`GenTcpConfig::listen_on()`]. It is possible to listen on multiple
+    /// [`GenTcpTransport`] keeps track of the listen socket addresses as they
+    /// are reported by polling it. It is possible to listen on multiple
     /// addresses, enabling port reuse for each, knowing exactly which listen
-    /// address is reused when dialing with a specific `GenTcpConfig`, as in the
+    /// address is reused when dialing with a specific `GenTcpTransport`, as in the
```
```suggestion
    /// address is reused when dialing with a specific [`GenTcpTransport`], as in the
```

transports/tcp/src/lib.rs
```diff
@@ -287,23 +281,58 @@ where
     /// case, one is chosen whose IP protocol version and loopback status is the
     /// same as that of the remote address. Consequently, for maximum control of
     /// the local listening addresses and ports that are used for outgoing
-    /// connections, a new `GenTcpConfig` should be created for each listening
+    /// connections, a new `GenTcpTransport` should be created for each listening
```
```suggestion
    /// connections, a new [`GenTcpTransport`] should be created for each listening
```

transports/tcp/src/lib.rs
```diff
@@ -441,9 +505,106 @@ where
...
+                }
+                Poll::Ready(Some(Err(err))) => {
+                    return Poll::Ready(TransportEvent::ListenerClosed {
+                        listener_id: listener.listener_id,
+                        reason: Err(err),
+                    });
+                }
+            }
+        }
+        // We register the current task to be woken up if a new listener is added.
```
mxinden: Where is the waker of the current task stored to be woken up later?

mxinden (commented):
Made it to the end. Couple of more comments, though nothing major.

protocols/relay/src/v2/client/transport.rs
elenaf9: Because the NetworkBehaviour never uses the channel, it just passes it to the Handler. The handler stores it in the `Handler::reservation` and from there sends messages to the transport in `Reservation::forward_messages_to_transport_listener`. 
Took me a while myself to understand this, which is why I pointed it out in the docs.

---

Circling back to an out of band discussion we had:
The fact that the handler sends the messages instead of the behaviour was actually the reason why I decided for now to stick with the concept of Listeners in the relay transport. We currently have a 1:1 relationship between the handler of the [first connection](https://github.com/libp2p/rust-libp2p/blob/9f7d0fe0ac6bc75a9922b7e60d0d146b3e48f94a/protocols/relay/src/v2/client.rs#L245-L249) to the relay, and the transport's listener for the relayed address. They share the above `mpsc::channel::<ToListenerMsg>`, with only a single sender (that is managed by the one handler). If either side is dropped the other one is informed and can close as well, which is quite convenient.
Of course we can do without listeners and just have a mpsc channel between behaviour and transport, but I am not sure whether this would really simplify things.

transports/tcp/src/lib.rs
elenaf9: I copied that logic, including the comment, from the old `swarm::connection::ListenersStream::poll`. 
It seems like a left-over from before we even used stable futures and probably only served to explain how `Poll::Pending` (back then `Ok(Async::NotReady)`) works.
Removed it now. As we already discussed it in https://github.com/libp2p/rust-libp2p/pull/2652#discussion_r906871039. we assume that the transport is polled again after a new listener is added, which will then register the new listener's waker.

elenaf9:
Thank you both for the reviews @mxinden @thomaseizinger! I know it is a large diff, so it's very much appreciated. 
I think I addressed all comments, please point out if in case that I missed something.

protocols/relay/src/v2/client/transport.rs
mxinden: Thanks for elaborating! The above makes sense. No opinion on what is the simpler approach. Fine leaving it as is.

mxinden:
Missing changelog entries. Otherwise this looks good to me :tada:

transports/tcp/CHANGELOG.md
```diff
@@ -6,6 +6,13 @@
   establishment errors early. See also [PR 2458] for the related async-io
   change.
 
+- Split `GenTcpConfig` into `GenTcpConfig` and `GenTcpTransport`. Drive the `TcpListenStream`s
+  within the `GenTcpTransport`. Add `Transport::poll` and `Transport::remove_listener`
+  for `GenTcpTransport`. See [PR 2652].
+
+[PR 2701]: https://github.com/libp2p/rust-libp2p/pull/2701/
```
mxinden: Why reference #2701 here?

mxinden (approved):
Looks good to me. :rocket: 

@thomaseizinger any objections to merging this?

transports/tcp/CHANGELOG.md
elenaf9: Woops good catch. Copy-paste error.

elenaf9:
```suggestion
```

thomaseizinger:
> Looks good to me. :rocket: 
> 
> @thomaseizinger any objections to merging this?

Nope, not at all!

thomaseizinger (approved):

mxinden:
Thanks @elenaf9 for taking this from start to finish. Very large undertaking!

Issue #2651, opened 2022-05-16
Tcp transport port_reuse_dialing test missing some assert


stormshield-pj50:
## Summary
The port_reuse_dialing test in transports/tcp/lib.rs seems to be incomplete, it does not check for port reuse.

## Expected behaviour
In this test the dialer outgoing connection must reuse the listener port. So an assert like this should be added near line 930 in the port_reuse_dialing/dialer function:
```
assert!(tcp.port_reuse.local_dial_addr(&listener.listen_addr.ip()).is_some());
```

## Actual behaviour
The assert is missing and if I add it it fails. It seems to be a consequence of the commit https://github.com/libp2p/rust-libp2p/commit/2ad905f35a61d7588f74aeea4e8010ccfbc70fcf. Previously the listen_addrs field in PortReuse was a Arc<RwLock<HashSet<(IpAddr, Port)>>> so it could be shared between GenTcpConfig and TcpListenStream. One ugly work around to make the assert pass in the test is to clone listener.port_reuse into tcp:
```
tcp.port_reuse = listener.port_reuse.clone(); 
```

## Possible Solution
Don't know if only the test is broken or if libp2p port reuse feature is also broken.

## Would you like to work on fixing this bug?
Yes

mxinden:
Thanks for the bug report @pj50. Great job catching this! This was my mistake.

For others following along here is where the `HashMap` instead of the  `Arc` is being cloned.

https://github.com/libp2p/rust-libp2p/blob/d21cd5fed76d879b02c4f78ea00e02737498b8ca/transports/tcp/src/lib.rs#L327

> Would you like to work on fixing this bug?
> 
> Yes

Great! Let me know in case you need any help.

mxinden:
Friendly ping @pj50. Still interested in providing a patch?

stormshield-pj50:
Sorry being late for a patch, I had to work on other topics last week. Working on it now.

mxinden:
Closing here with https://github.com/libp2p/rust-libp2p/pull/2670 merged.

PR #2659, opened 2022-05-19
protocols/dcutr/examples: Wait to tell relay its public addr

mxinden:
# Description

As a listening client, when requesting a reservation with a relay, the relay
responds with its public addresses. The listening client can then use the public
addresses of the relay to advertise itself as reachable under a relayed
address (`/<public-relay-addr>/p2p-circuit/p2p/<listening-client-peer-id>`).

The above operates under the assumption that the relay knows its public address.
A relay learns its public address from remote peers, via the identify protocol.
In the case where the relay just started up, the listening client might be the
very first node to connect to it.

Such scenario allows for a race condition. The listening client requests a
reservation from the relay, while the relay requests its public address from the
listening client. The former needs to contain the response from the latter.

This commit serializes the two requests, making sure, in the case of a freshly
started relay, that the listening client tells the relay its public address
before requesting a reservation from the relay.

Follow up to https://github.com/libp2p/rust-libp2p/pull/2642. Suggested in https://github.com/libp2p/rust-libp2p/issues/2621#issuecomment-1127963348.

Note that hole punching is still broken due to https://github.com/libp2p/rust-libp2p/issues/2651.

## Change checklist

### DIFF
--- a/protocols/dcutr/examples/client.rs
+++ b/protocols/dcutr/examples/client.rs
@@ -186,22 +186,12 @@ fn main() -> Result<(), Box<dyn Error>> {
         }
     });
 
-    match opts.mode {
-        Mode::Dial => {
-            swarm.dial(opts.relay_address.clone()).unwrap();
-        }
-        Mode::Listen => {
-            swarm
-                .listen_on(opts.relay_address.clone().with(Protocol::P2pCircuit))
-                .unwrap();
-        }
-    }
-
-    // Wait till connected to relay to learn external address. In case we are in listening mode,
-    // wait for the relay to accept our reservation request.
+    // Connect to the relay server. Not for the reservation or relayed connection, but to (a) learn
+    // our local public address and (b) enable a freshly started relay to learn its public address.
+    swarm.dial(opts.relay_address.clone()).unwrap();
     block_on(async {
         let mut learned_observed_addr = false;
-        let mut relay_accepted_reservation = false;
+        let mut told_relay_observed_addr = false;
 
         loop {
             match swarm.next().await.unwrap() {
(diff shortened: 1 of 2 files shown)


mxinden:
//CC @foemass

protocols/dcutr/examples/client.rs
```diff
@@ -186,69 +186,75 @@ fn main() -> Result<(), Box<dyn Error>> {
...
+                        .clone()
+                        .with(Protocol::P2pCircuit)
+                        .with(Protocol::P2p(opts.remote_peer_id.unwrap().into())),
+                )
+                .unwrap();
+        }
+        Mode::Listen => {
+            swarm
+                .listen_on(opts.relay_address.clone().with(Protocol::P2pCircuit))
+                .unwrap();
```
elenaf9:
```suggestion
                    opts.relay_address
                        .with(Protocol::P2pCircuit)
                        .with(Protocol::P2p(opts.remote_peer_id.unwrap().into())),
                )
                .unwrap();
        }
        Mode::Listen => {
            swarm
                .listen_on(opts.relay_address.with(Protocol::P2pCircuit))
                .unwrap();
```
Nit: don't think we need to clone the address here; it's not used afterwards anymore.

protocols/dcutr/examples/client.rs
```diff
@@ -186,69 +186,75 @@ fn main() -> Result<(), Box<dyn Error>> {
...
+                            client::Event::ReservationReqAccepted { .. },
+                        )) => {
+                            info!("Relay accepted our reservation request.");
+                            break;
+                        }
+                        SwarmEvent::Behaviour(Event::Identify(_)) => {}
+                        event => panic!("{:?}", event),
+                    }
+                }
+            });
```
elenaf9: Why do we need to do a separate loop here to wait for the `ReservationReqAccepted`? We don't initiate any following action, but instead directly jump into the next loop below. Couldn't we just remove this block?

mxinden: Good idea!

protocols/dcutr/examples/client.rs
```diff
@@ -186,69 +186,55 @@ fn main() -> Result<(), Box<dyn Error>> {
...
-                    .with(Protocol::P2p(opts.remote_peer_id.unwrap().into())),
-            )
-            .unwrap();
+    match opts.mode {
+        Mode::Dial => {
+            swarm
+                .dial(
+                    opts.relay_address
+                        .clone()
+                        .with(Protocol::P2pCircuit)
```
elenaf9:
```suggestion
                    opts.relay_address
                        .with(Protocol::P2pCircuit)
```
😇

elenaf9 (approved):

thomaseizinger (approved):

protocols/dcutr/examples/client.rs
mxinden: My bad. Thank you! :)

PR #2658, opened 2022-05-19
transports/dns: Remove `Clone` trait bound

dignifiedquire:
# Description

Follow on to https://github.com/libp2p/rust-libp2p/pull/2529, otherwise this error gets generated when trying to use it

```rust
let transport = libp2p::tcp::TokioTcpConfig::new().nodelay(true);
let transport = libp2p::websocket::WsConfig::new(transport.clone()).or_transport(transport);
let transport = libp2p::dns::TokioDnsConfig::system(transport).unwrap();

transport.upgrade(core::upgrade::Version::V1Lazy) // < this line fails

```

```
error[E0599]: the method `upgrade` exists for struct `GenDnsConfig<OrTransport<libp2p::libp2p_websocket::WsConfig<GenTcpConfig<libp2p::libp2p_tcp::tokio::Tcp>>, GenTcpConfig<libp2p::libp2p_tcp::tokio::Tcp>>, trust_dns_resolver::name_server::connection_provider::GenericConnection, trust_dns_resolver::name_server::connection_provider::GenericConnectionProvider<trust_dns_resolver::name_server::connection_provider::tokio_runtime::TokioRuntime>>`, but its trait bounds were not satisfied
   --> iroh-p2p/src/service.rs:349:10
    |
349 |           .upgrade(core::upgrade::Version::V1Lazy)
    |            ^^^^^^^ method cannot be called on `GenDnsConfig<OrTransport<libp2p::libp2p_websocket::WsConfig<GenTcpConfig<libp2p::libp2p_tcp::tokio::Tcp>>, GenTcpConfig<libp2p::libp2p_tcp::tokio::Tcp>>, trust_dns_resolver::name_server::connection_provider::GenericConnection, trust_dns_resolver::name_server::connection_provider::GenericConnectionProvider<trust_dns_resolver::name_server::connection_provider::tokio_runtime::TokioRuntime>>` due to unsatisfied trait bounds
    |
   ::: /Users/dignifiedquire/.cargo/git/checkouts/rust-libp2p-98135dbcf5b63918/999a212/core/src/transport/choice.rs:27:1
    |
27  |   pub struct OrTransport<A, B>(A, B);
    |   ----------------------------------- doesn't satisfy `_: Clone`
    |
   ::: /Users/dignifiedquire/.cargo/git/checkouts/rust-libp2p-98135dbcf5b63918/999a212/transports/dns/src/lib.rs:111:1
    |
111 | / pub struct GenDnsConfig<T, C, P>
112 | | where
[...]

PR #2660, opened 2022-05-19
protocols/autonat: Fix flaky test

elenaf9:
# Description

Follow-up on https://github.com/libp2p/rust-libp2p/pull/2658#issuecomment-1131942213.
In rare cases (that only really happen in a test setup where both peers may run in the same thread) the server may observe a connection (which indicates a successful dial attempt) and already report the response back to the client, before the inbound connection event is reported at the client. Most likely occurring now because with #2627 the behaviour (where we receive the server's response) is polled before the pool (that reports the established connection). 
I assume that in the real world this is not relevant because it would take much longer for the server to send the response back and in the meantime the client would already have reported the inbound connection.

Handle in the autonat `test_client::test_auto_probe` that a `OutboundProbeEvent::Response` and as a result the `Event::StatusChanged` can be reported before the related inbound connection event.

## Change checklist

### DIFF
--- a/protocols/autonat/tests/test_client.rs
+++ b/protocols/autonat/tests/test_client.rs
@@ -189,15 +189,14 @@ async fn test_auto_probe() {
             other => panic!("Unexpected behaviour event: {:?}.", other),
         };
 
-        let mut has_received_response = false;
-        // Expect inbound dial from server.
+        let mut had_connection_event = false;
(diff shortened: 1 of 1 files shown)


mxinden (approved):
This makes sense to me. Thanks for debugging this and thanks for the great explainer.

PR #2669, opened 2022-05-23
core/muxing: Remove the `StreamMuxer::flush_all` function

thomaseizinger:
# Description

`libp2p-core` provides the `StreamMuxer` abstraction so it can provide
functionality that abstracts over this trait.

We never use the `flush_all` function as part of our abstractions.
No one else is going to use it so we can remove it from the abstraction.

- Extracted out of https://github.com/libp2p/rust-libp2p/pull/2648.

## Change checklist

### DIFF
--- a/core/src/muxing/singleton.rs
+++ b/core/src/muxing/singleton.rs
@@ -149,7 +149,7 @@ where
 
     fn destroy_substream(&self, _: Self::Substream) {}
 
-    fn poll_close(&self, cx: &mut Context<'_>) -> Poll<Result<(), io::Error>> {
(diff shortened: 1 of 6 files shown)


mxinden (approved):
:pray:

core/src/muxing/singleton.rs
```diff
@@ -150,11 +150,6 @@ where
     fn destroy_substream(&self, _: Self::Substream) {}
 
     fn poll_close(&self, cx: &mut Context<'_>) -> Poll<Result<(), io::Error>> {
```
mxinden:
```suggestion
    fn poll_close(&self, _cx: &mut Context<'_>) -> Poll<Result<(), io::Error>> {
```

PR #2670, opened 2022-05-23
transports/tcp: Fix port reuse using Arc<RwLock> for listen_addrs

stormshield-pj50:
Sponsored by: Stormshield

# Description
See reported issue https://github.com/libp2p/rust-libp2p/issues/2651 for details.

## Change checklist
- two tests in transports/tcp (port_reuse_dialing and port_reuse_listening) have been enhanced for testing fixed regression.
- protocols/dcutr/examples test has been run successfully with the fix.

### DIFF
--- a/transports/tcp/src/lib.rs
+++ b/transports/tcp/src/lib.rs
@@ -59,6 +59,7 @@ use std::{
     io,
     net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr, TcpListener},
     pin::Pin,
+    sync::{Arc, RwLock},
     task::{Context, Poll},
     time::Duration,
 };
@@ -95,7 +96,7 @@ enum PortReuse {
     Enabled {
         /// The addresses and ports of the listening sockets
         /// registered as eligible for port reuse when dialing.
-        listen_addrs: HashSet<(IpAddr, Port)>,
+        listen_addrs: Arc<RwLock<HashSet<(IpAddr, Port)>>>,
     },
 }
 
@@ -106,7 +107,10 @@ impl PortReuse {
     fn register(&mut self, ip: IpAddr, port: Port) {
         if let PortReuse::Enabled { listen_addrs } = self {
             log::trace!("Registering for port reuse: {}:{}", ip, port);
-            listen_addrs.insert((ip, port));
+            listen_addrs
+                .write()
+                .expect("`register()` and `unregister()` never panic while holding the lock")
+                .insert((ip, port));
         }
     }
 
@@ -116,7 +120,10 @@ impl PortReuse {
     fn unregister(&mut self, ip: IpAddr, port: Port) {
         if let PortReuse::Enabled { listen_addrs } = self {
             log::trace!("Unregistering for port reuse: {}:{}", ip, port);
-            listen_addrs.remove(&(ip, port));
+            listen_addrs
+                .write()
+                .expect("`register()` and `unregister()` never panic while holding the lock")
+                .remove(&(ip, port));
         }
     }
 
@@ -131,7 +138,11 @@ impl PortReuse {
     /// listening socket address is found.
     fn local_dial_addr(&self, remote_ip: &IpAddr) -> Option<SocketAddr> {
         if let PortReuse::Enabled { listen_addrs } = self {
-            for (ip, port) in listen_addrs.iter() {
+            for (ip, port) in listen_addrs
+                .read()
+                .expect("`local_dial_addr` never panic while holding the lock")
+                .iter()
+            {
                 if ip.is_ipv4() == remote_ip.is_ipv4()
                     && ip.is_loopback() == remote_ip.is_loopback()
                 {
@@ -286,7 +297,7 @@ where
     pub fn port_reuse(mut self, port_reuse: bool) -> Self {
         self.port_reuse = if port_reuse {
             PortReuse::Enabled {
-                listen_addrs: HashSet::new(),
+                listen_addrs: Arc::new(RwLock::new(HashSet::new())),
             }
         } else {
             PortReuse::Disabled
@@ -707,7 +718,7 @@ fn ip_to_multiaddr(ip: IpAddr, port: u16) -> Multiaddr {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use futures::channel::mpsc;
+    use futures::channel::{mpsc, oneshot};
 
     #[test]
     fn multiaddr_to_tcp_conversion() {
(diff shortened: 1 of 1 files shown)


transports/tcp/src/lib.rs
```diff
@@ -106,7 +107,10 @@ impl PortReuse {
     fn register(&mut self, ip: IpAddr, port: Port) {
         if let PortReuse::Enabled { listen_addrs } = self {
             log::trace!("Registering for port reuse: {}:{}", ip, port);
-            listen_addrs.insert((ip, port));
+            listen_addrs
+                .write()
+                .expect("`register()` and `unregister()` never panic while holding the lock")
```
mxinden, line 112: I will update these in a follow up pull request.

transports/tcp/src/lib.rs
```diff
@@ -927,7 +938,16 @@ mod tests {
...
+                    // Check that tcp and listener share the same port reuse SocketAddr
+                    let port_reuse_tcp = tcp.port_reuse.local_dial_addr(&listener.listen_addr.ip());
+                    let port_reuse_listener = listener
+                        .port_reuse
+                        .local_dial_addr(&listener.listen_addr.ip());
+                    assert!(port_reuse_tcp.is_some());
+                    assert_eq!(port_reuse_tcp, port_reuse_listener);
+
                     // Obtain a future socket through dialing
+                    // FIXME: Check that the port used by socket is the same as the port returned by local_dial_addr
```
mxinden: How about checking the `ListenerEvent::Upgrade { remote_addr, ..}` in `async fn listener` above instead?

mxinden (commented):
Thank you @pj50 for providing the patch as well as augmenting the test.

With this change, I find the cloning behavior of `GenTcpConfig` non-intuitive, given that the clone would share the `listen_addrs` field when `PortReuse::Enabled`.

What do you think of removing the derived `Clone` on `GenTcpConfig`?


``` diff
diff --git a/transports/tcp/src/lib.rs b/transports/tcp/src/lib.rs
index 6e6eb771..0badcc24 100644
--- a/transports/tcp/src/lib.rs
+++ b/transports/tcp/src/lib.rs
@@ -66,7 +66,7 @@ use std::{
 use provider::{IfEvent, Provider};
 
 /// The configuration for a TCP/IP transport capability for libp2p.
-#[derive(Clone, Debug)]
+#[derive(Debug)]
 pub struct GenTcpConfig<T> {
     /// The type of the I/O provider.
     _impl: std::marker::PhantomData<T>,
@@ -247,7 +247,7 @@ where
     /// let listen_addr2: Multiaddr = "/ip4/127.0.0.1/tcp/9002".parse().unwrap();
     ///
     /// let mut tcp1 = TcpConfig::new().port_reuse(true);
-    /// let mut listener1 = tcp1.clone().listen_on(listen_addr1.clone()).expect("listener");
+    /// let mut listener1 = tcp1.listen_on(listen_addr1.clone()).expect("listener");
     /// match listener1.next().await.expect("event")? {
     ///     ListenerEvent::NewAddress(listen_addr) => {
     ///         println!("Listening on {:?}", listen_addr);
@@ -258,7 +258,7 @@ where
     /// }
     ///
     /// let mut tcp2 = TcpConfig::new().port_reuse(true);
-    /// let mut listener2 = tcp2.clone().listen_on(listen_addr2).expect("listener");
+    /// let mut listener2 = tcp2.listen_on(listen_addr2).expect("listener");
     /// match listener2.next().await.expect("event")? {
     ///     ListenerEvent::NewAddress(listen_addr) => {
     ///         println!("Listening on {:?}", listen_addr);
```

mxinden:
> Sponsored by: Stormshield

You asked whether this can be included in the git history. I am not sure what legal implications this might have. For the sake of simplicity I would not include it. Is that fine by you?

That is not to say, that I don't appreciate the sponsorship! Thanks for the help.

transports/tcp/src/lib.rs
stormshield-pj50: If you remove Clone on GenTcpConfig what about GenDnsConfig<GenTcpConfig, ...> which also implements Clone ?

transports/tcp/src/lib.rs
stormshield-pj50: OK I updated the code using an mpsc channel to pass the reused port between the dialer and the listener, and cheking in the listener upgrade event that the remote port is the same as the one received in the channel.

transports/tcp/src/lib.rs
```diff
@@ -943,8 +985,9 @@ mod tests {
             #[cfg(feature = "async-io")]
             {
                 let (ready_tx, ready_rx) = mpsc::channel(1);
-                let listener = listener::<async_io::Tcp>(addr.clone(), ready_tx);
-                let dialer = dialer::<async_io::Tcp>(addr.clone(), ready_rx);
+                let (port_reuse_tx, port_reuse_rx) = mpsc::channel(1);
```
mxinden: How about using a oneshot here?

https://docs.rs/futures/latest/futures/channel/oneshot/fn.channel.html

transports/tcp/src/lib.rs
```diff
@@ -900,15 +911,28 @@ mod tests {
...
-                    ListenerEvent::Upgrade { upgrade, .. } => {
+                    ListenerEvent::Upgrade {
+                        upgrade,
+                        local_addr: _,
+                        mut remote_addr,
+                    } => {
+                        // Receive the dialer tcp port reuse
+                        let remote_port_reuse = port_reuse_rx.next().await.unwrap();
+                        // And check it is the same as the remote port used for upgrade
+                        assert_eq!(remote_addr.pop().unwrap(), remote_port_reuse);
```
mxinden: :pray:

transports/tcp/src/lib.rs
mxinden: I would advocate for removing `Clone` from `GenDnsConfig` as well, though I would prefer that to happen in a separate pull request.

Can `GenDnsConfig` be `Clone` even if it is wrapping ` GenTcpConfig` which is not `Clone`?

Yes, given that `GenDnsConfig` wraps the `GenTcpConfig` in an `Arc<Mutex<_>>`.

Hope I am not missing something @pj50.

mxinden (commented):
Thanks for the addition to the tests!

Followed up on the `Clone` discussion. What do you think?

transports/tcp/src/lib.rs
stormshield-pj50: OK. Updated the commit with a oneshot.

transports/tcp/src/lib.rs
stormshield-pj50: I tried removing Clone from GenDnsConfig, however
https://github.com/libp2p/rust-libp2p/blob/4aa84bf7cd78c8553294f4f352fce9e20b6fbd4e/src/lib.rs#L206
does not compile any more. I don't know how to fix it properly, so I leave it to you, and you're right I think it would be much better to fix it in another pull request.

transports/tcp/src/lib.rs
```diff
@@ -953,8 +995,9 @@ mod tests {
             #[cfg(feature = "tokio")]
             {
                 let (ready_tx, ready_rx) = mpsc::channel(1);
-                let listener = listener::<tokio::Tcp>(addr.clone(), ready_tx);
-                let dialer = dialer::<tokio::Tcp>(addr.clone(), ready_rx);
+                let (port_reuse_tx, port_reuse_rx) = oneshot: channel();
```
```suggestion
                let (port_reuse_tx, port_reuse_rx) = oneshot::channel();
```

mxinden (approved):
Thanks @pj50 for the help here, going all the way from debugging to proposing the fix :pray:

PR #2679, opened 2022-05-29
protocols/dcutr/examples: Set libp2p-ping to keep conn alive

mxinden:
# Description

The example is for testing out the hole punching capabilities only. Thus best to
remove the complexity of connection keep alives via the ping protocol.

Might make https://github.com/libp2p/rust-libp2p/issues/2609#issuecomment-1135270283 easier to debug. //CC @T2JOESl4m2ZpNC

## Change checklist

### DIFF
--- a/protocols/dcutr/examples/client.rs
+++ b/protocols/dcutr/examples/client.rs
@@ -145,7 +145,7 @@ fn main() -> Result<(), Box<dyn Error>> {
 
     let behaviour = Behaviour {
         relay_client: client,
-        ping: Ping::new(PingConfig::new()),
+        ping: Ping::new(PingConfig::new().with_keep_alive(true)),
         identify: Identify::new(IdentifyConfig::new(
             "/TODO/0.0.1".to_string(),
             local_key.public(),


thomaseizinger (commented):
I am okay with the change but confused by your PR description.

> Thus best to remove the complexity of connection keep alives via the ping protocol.

You are talking about removing but the diff says you are activating `keep_alive`?

thomaseizinger (approved):
Ah, I think it just clicked for me.

I think what is meant is: Keeping a connection alive without ping is complex and by activating ping's `keep-alive` feature, we remove this complexity.

elenaf9 (changes requested):
Not sure if I understand this change. How is not keeping the connection alive currently causing additional complexity? 
As a listener, we keep a long-lived connection to the relay anyway, and as a dialer we only want to try hole-punching to the target. After the hole-punch we exchange Identify events and I think it makes sense that the connection eventually closes again.  
With the keep alive as a dialer we would keep the connection to the relay, and even after both client successfully hole-punched they would still also keep their relayed connection alive. If someone would be using this example as a reference it might cause some unwanted behaviour for them.

mxinden:
> With the keep alive as a dialer we would keep the connection to the relay, and even after both client successfully hole-punched they would still also keep their relayed connection alive. If someone would be using this example as a reference it might cause some unwanted behaviour for them.

True, though users seem to be confused that they get a `DirectConnectionUpgradeSucceeded` followed by a `ConnectionClosed`.

I don't have a strong opinion here. You have a good point above. As a compromise, how about calling out that one would not use `keep_alive(true)` in production code?

PR #2681, opened 2022-05-30
protocols/request-response: close io after reading response

melekes:
Otherwise `read_response` will be always failing.

# Description

This PR fixes an issue (unless this is not a bug, but a feature 😆) with `OutboundUpgrade` for `RequestProtocol` where IO was closed before reading the response. Note this is different from `inbound_upgrade` above where IO is closed after the response is written (i.e. after we're done with IO).

## Change checklist

- [ ] ~~I have made corresponding changes to the documentation~~

### DIFF
--- a/protocols/request-response/src/handler/protocol.rs
+++ b/protocols/request-response/src/handler/protocol.rs
@@ -179,9 +179,9 @@ where
         async move {
             let write = self.codec.write_request(&protocol, &mut io, self.request);
             write.await?;
-            io.close().await?;
             let read = self.codec.read_response(&protocol, &mut io);
             let response = read.await?;
+            io.close().await?;
             Ok(response)
         }
         .boxed()


melekes:
Please let me know if this deserves a changelog entry or you want me to try to write a test.

tomaka:
This is intentional. It's only the writing side that is closed when `close()` is called, not the reading side.

melekes:
> This is intentional. It's only the writing side that is closed when `close()` is called, not the reading side.

Hmm.. what about protocols where closing "writing" side immediately leads to closing of the "reading" side? Even if that was the intention, can we still move the `io.close`? Honestly even looking at this code confuses me because `io` is closed before  it's used in `read_response` (i.e. there's no indication in the code that `io.close` only closes "writing" side)

thomaseizinger:
> (i.e. there's no indication in the code that `io.close` only closes "writing" side)

Unless I am mistaken, this should be apparent from looking at the [`AsyncRead`](https://docs.rs/futures/0.3.21/futures/io/trait.AsyncRead.html) and [`AsyncWrite`](https://docs.rs/futures/0.3.21/futures/io/trait.AsyncWrite.html) traits. You cannot close a `AsyncRead`, only `AsyncWrite` offers the `poll_close` function.

tomaka:
> Hmm.. what about protocols where closing "writing" side immediately leads to closing of the "reading" side?

The contract of the `StreamMuxer` trait is that calling `shutdown_substream` only closes the writing side, and that when that happens then the other side will receive `Ok(0)` (i.e. EOF) when they call `read_substream`.
If WebRTC can't support this, then we have to modify the trait.

> Even if that was the intention, can we still move the io.close?

Changing the way the trait is used just so that you don't hit a bug is not a solution.

PR #2687, opened 2022-06-02
core/either: Use `io::Result` type alias instead of renaming io::Error

thomaseizinger:
# Description

Using the type alias is more idiomatic because it avoids renaming types on import.

## Change checklist


elenaf9 (approved):

mxinden (approved):
:+1:

PR #2688, opened 2022-06-02
muxers/mplex/benches: Use combinators to reduce nesting

thomaseizinger:
# Description

Instead of matching on the result, we can use combinators to directly
extract the substream from the event.

Will make the diff in https://github.com/libp2p/rust-libp2p/pull/2648 smaller.

## Change checklist


elenaf9 (approved):

mxinden (approved):
Cool. Much better! Thanks.

PR #2689, opened 2022-06-02
Adding example for Identify protocol

### LINKED ISSUE #2684: Add more examples
## Description

I'll like to start contributing to rust-libp2p. A way I think I could use to get my feet wet is to add more examples of how to use [rust-libp2p](https://github.com/libp2p/rust-libp2p). 

I am thinking of porting some of the examples in [js-libp2p](https://github.com/libp2p/js-libp2p/tree/master/examples) that are not present here and which I think are interesting enough. 

Before I go ahead, I'll like to ask if this is a welcome contribution. 

Specifically I am thinking of adding these examples:

- [ ] auto-relay
- [ ] echo
- [ ] peer-and-content-routing
    - [ ] Bootstrap list of Peers when booting a node
    - [ ] Pubsub based Peer Discovery
- [ ] discovery mechanisms

And those that is not in the js-libp2p examples but I think might be interesting

- [x] Identify protocol

## Motivation

The examples helps makes onboarding to rust-libp2p easier. 

Yes

dadepo:
# Description

Added example for the Identify protocol.

Part of #2684

## Change checklist


examples/identify.rs
```diff
@@ -0,0 +1,82 @@
+// Copyright 2018 Parity Technologies (UK) Ltd.
```
mxinden: Given that this example is focused on `libp2p-identify`, how about moving it to `protocols/identify/examples/`? To make sure one can still discover it, it can be mentioned in https://github.com/libp2p/rust-libp2p/blob/master/examples/README.md.

examples/identify.rs
```diff
@@ -0,0 +1,82 @@
...
+
+//! identify example
+//!
+//! In the first terminal window, run:
+//!
+//! ```sh
+//! cargo run --example identify
+//! ```
+//! It will print the PeerId and the listening addresses, e.g. `Listening on
+//! "/ip4/0.0.0.0/tcp/24915"`
```
mxinden:
```suggestion
//! "/ip6/2001:db8:: /tcp/24915"`
```

Let's not use a wildcard address here.

examples/identify.rs
```diff
@@ -0,0 +1,82 @@
...
+// DEALINGS IN THE SOFTWARE.
+
+//! identify example
+//!
+//! In the first terminal window, run:
+//!
+//! ```sh
+//! cargo run --example identify
+//! ```
+//! It will print the PeerId and the listening addresses, e.g. `Listening on
```
mxinden:
```suggestion
//! It will print the `PeerId` and the listening addresses, e.g. `Listening on
```

examples/identify.rs
```diff
@@ -0,0 +1,82 @@
...
+//! "/ip4/0.0.0.0/tcp/24915"`
+//!
+//! In the second terminal window, start a new instance of the example with:
+//!
+//! ```sh
+//! cargo run --example identify -- /ip4/127.0.0.1/tcp/24915
+//! ```
+//! The two nodes establish a connection, negotiate the identify protocol
+//! The dialing node prints out the peer id of the node it is sending identify info to
```
mxinden:
```suggestion
//! The dialing node prints out the `PeerId` of the node it is sending identify info to
```

examples/identify.rs
```diff
@@ -0,0 +1,82 @@
...
+//! The other node prints out the received identify info
```
mxinden:
```suggestion
//! The other node prints out the received identify info.
```

examples/identify.rs
```diff
@@ -0,0 +1,82 @@
...
+    swarm.listen_on("/ip4/0.0.0.0/tcp/0".parse()?)?;
+
+    // Dial the peer identified by the multi-address given as the second
+    // command-line argument, if any.
+    if let Some(addr) = std::env::args().nth(1) {
+        let remote: Multiaddr = addr.parse()?;
+        swarm.dial(remote)?;
+        println!("Dialed {}", addr)
+    }
+
```
mxinden:
```suggestion
```

examples/identify.rs
```diff
@@ -0,0 +1,82 @@
...
+
+    loop {
+        match swarm.select_next_some().await {
+            SwarmEvent::NewListenAddr { address, .. } => println!("Listening on {:?}", address),
+            // prints peer id identify info is being sent to
```
mxinden:
```suggestion
            // Prints peer id identify info is being sent to.
```

Let's have each code comment be a full sentence with a period at the end.

mxinden (approved):
Appreciate the help here. Though still a couple of comments.

dadepo:
> Appreciate the help here. Though still a couple of comments.

Gladly done. I will resolve/respond the comments in a bit. 

Quick question, seeing you already approved, do you prefer I create individual PR for each example added?
My initial thought was to include all the added examples in a single PR, hence why I set this PR to draft for now. Plan was to update it to _ready_ PR once all the examples have been added

Let me know which approach the team finds more convenient.

examples/identify.rs
dadepo: I was not even aware that there are some examples placed together with some of the individual protocols. Is there a reason why not to have all the examples listed in `https://github.com/libp2p/rust-libp2p/tree/master/examples/*`? 

Having one canonical place where all the examples are listed, IMHO makes discovery easier.

mxinden:
> Quick question, seeing you already approved

That was a mistake. Sorry.

> do you prefer I create individual PR for each example added?

Yes. I prefer small atomic pull requests. Makes my life as a maintainer a lot easier. Thanks.

examples/identify.rs
mxinden: No strong opinion on my end. I see `/examples` as a folder for examples that combine many protocols and `protocols/xxx/examples` as a folder for examples targeting a single protocol.

> Having one canonical place where all the examples are listed, IMHO makes discovery easier.

I see your point.

Maybe @elenaf9 or @thomaseizinger have an opinion here.

examples/identify.rs
```diff
@@ -0,0 +1,81 @@
...
+// DEALINGS IN THE SOFTWARE.
+
+//! identify example
+//!
+//! In the first terminal window, run:
+//!
+//! ```sh
+//! cargo run --example identify
+//! ```
+//! It will print the `PeerId` and the listening addresses, e.g. `Listening on
```
mxinden:
```suggestion
//! It will print the [`PeerId`] and the listening addresses, e.g. `Listening on
```

examples/identify.rs
elenaf9: I am leaning towards having examples focused on a single protocol in that protocol's folder. Since they are individual crates, imo they should "own" their own examples.

> I was not even aware that there are some examples placed together with some of the individual protocols.

We should definitely mention this in the top-level `examples/README.md`, which is currently not the case.

thomaseizinger: > I am leaning towards having examples focused on a single protocol in that protocol's folder. Since they are individual crates, imo they should "own" their own examples.

I share the same view. These protocols are individual crates and I think we should follow cargo conventions despite this being a large workspace.

> > I was not even aware that there are some examples placed together with some of the individual protocols.
> 
> We should definitely mention this in the top-level `examples/README.md`, which is currently not the case.

Another solution to this would be to do away with the current meta crate being "special" by being a workspace manifest and a crate at the same time. i.e. we could create a directory `meta` which could house the `libp2p` crate. This would also move the top-level `examples/` directory which might make it more obvious to users that they will find examples within each crate.

mxinden (dismissed):
:rocket: Thank you!

mxinden:
CI failures are due to https://github.com/mxinden/if-watch/pull/19.

mxinden:
> CI failures are due to [mxinden/if-watch#19](https://github.com/mxinden/if-watch/pull/19).

This is fixed now.

To make CI green you would need to:

- Run rustfmt
- Add `libp2p` as an explicit `dev-dependency` to `protocols/identify/Cargo.toml` similar to https://github.com/libp2p/rust-libp2p/blob/eb490c08e9ebd1c4bc1c3180a4c66f62e4cc3bd8/protocols/dcutr/Cargo.toml#L33

examples/identify.rs
elenaf9: Given that we called `swarm.listen_on` with a ipv4 wildcard address, this can never result in a ipv6 listen address, right? Thus think `"/ip4/127.0.0.1/tcp/24915"` would make more sense here.

protocols/identify/examples/identify.rs
```diff
@@ -0,0 +1,88 @@
...
+//! "/ip6/2001:db8:: /tcp/24915"`
+//!
+//! In the second terminal window, start a new instance of the example with:
+//!
+//! ```sh
+//! cargo run --example identify -- /ip4/127.0.0.1/tcp/24915
+//! ```
+//! The two nodes establish a connection, negotiate the identify protocol
+//! The dialing node prints out the `PeerId` of the node it is sending identify info to
+//! The other node prints out the received identify info.
```
elenaf9: I think not only the dialer, but instead both peers will sent each other identify info once they establish a connection, no?

elenaf9 (commented):
Thanks @dadepo! Two comments, otherwise looks good to me.

elenaf9 (approved):

PR #2691, opened 2022-06-03
core/muxing: Introduce `StreamMuxerEvent::map_inbound_stream`

thomaseizinger:
# Description

This reduces the noise within core/src/either.rs.

Will reduce diff in https://github.com/libp2p/rust-libp2p/pull/2648.

## Change checklist


mxinden (commented):
Looks good to me. Very much in favor of these small helper methods. Thanks for pushing for these.

Mind adding a changelog entry and bumping the `libp2p-core` patch version?

core/src/muxing.rs
```diff
@@ -241,6 +241,16 @@ impl<T> StreamMuxerEvent<T> {
             None
         }
     }
+
+    /// Map the stream within [`StreamMuxerEvent::InboundSubstream`] to a new type.
+    pub fn map_stream<O>(self, map: impl FnOnce(T) -> O) -> StreamMuxerEvent<O> {
```
mxinden:
```suggestion
    pub fn map_inbound_stream<O>(self, map: impl FnOnce(T) -> O) -> StreamMuxerEvent<O> {
```

Just to avoid the confusion of folks expecting the `StreamMuxerEvent` to somehow contain both inbound and outbound streams. What do you think?

thomaseizinger: It will contain both streams once #2648 is merged. If we do this rename now, I will have to rename it again :)

What do you prefer?

mxinden: Though extra work, I would prefer the rename. That way I can merge to `master` already, keeping `master` clean. The two pull requests might not be released in the same release. Mind doing the change (twice)? :innocent:

core/CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+# 0.33.1
```
```suggestion
# 0.33.1 - unreleased
```

mxinden (approved):

PR #2694, opened 2022-06-06
protocols/identify: Allow at most one inbound identify push stream

mxinden:
# Description

An identify push contains the whole identify information of a remote
peer. Upgrading multiple inbound identify push streams is useless.
Instead older streams are dropped in favor of newer streams.

## Change checklist

### DIFF
--- a/protocols/identify/CHANGELOG.md
+++ b/protocols/identify/CHANGELOG.md
@@ -1,3 +1,7 @@
+# 0.36.1 - unreleased
+
+- Allow at most one inbound identify push stream.
+
 # 0.36.0
 
 - Update to `libp2p-core` `v0.33.0`.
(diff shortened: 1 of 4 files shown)


elenaf9 (approved):

thomaseizinger (approved):
Fine with me.

Is this going towards the "remove the Upgrade traits" idea that you mentioned at some point?
Perhaps we should also build some infrastructure where protocols can define that concurrency factor of their substreams.

mxinden:
> Is this going towards the "remove the Upgrade traits" idea that you mentioned at some point?

The lazy upgrade used in this pull request, yes. Though even if we don't go all the way, I think this is a step forward.

I will write up a proposal at some point.

PR #2697, opened 2022-06-07
swarm/: Limit negotiating inbound substreams per connection

mxinden:
# Description

This limit is shared across all `ConnectionHandler`s on a single connection. It
only enforces a limit on the number of negotiating substreams. Once negotiated a
`ConnectionHandler` manages the lifecycle of the substream and has to enforce
limits themselves.

## Change checklist


swarm/src/connection/handler_wrapper.rs
```diff
@@ -243,6 +247,11 @@ where
     ) {
         match endpoint {
             SubstreamEndpoint::Listener => {
+                if self.negotiating_in.len() == MAX_NUM_NEGOTIATING_IN {
```
thomaseizinger: Instead of doing this check here, we could create a wrapper around `FuturesUnordered` that enforces a limit on the number of concurrent futures. This would foster reuse and actually enforce this limit. At the moment, there is a risk that this check is circumvented in case the `FuturesUnordered` is modified in a different place without this check.

swarm/src/connection/handler_wrapper.rs
```diff
@@ -81,6 +77,15 @@ where
...
     substream_upgrade_protocol_override: Option<upgrade::Version>,
+    /// The maximum number of inbound streams concurrently negotiating. New inbound
+    /// streams exceeding the limit are dropped and thus reset.
+    ///
+    /// Note: This only enforces a limit on the number of concurrently
+    /// negotiating inbound streams. The total number of inbound streams on a
+    /// connection is the sum of negotiating and negotiated streams. A limit on
+    /// the total number of streams can be enforced at the [`StreamMuxerBox`]
+    /// level.
+    max_number_negotiating_inbound_streams: usize,
```
thomaseizinger: Nit: `number` might be redundant here.

swarm/src/connection/handler_wrapper.rs
```diff
@@ -35,6 +35,10 @@ use libp2p_core::{
 };
 use std::{error, fmt, pin::Pin, task::Context, task::Poll, time::Duration};
 
+/// The maximum number of incoming streams concurrently negotiated. Streams are
+/// dropped and thus reset.
+const MAX_NUM_NEGOTIATING_IN: usize = 128;
```
thomaseizinger: How was this number chosen?

thomaseizinger (approved):

swarm/src/connection/handler_wrapper.rs
mxinden: :+1: for the reusable abstraction, though I would much rather like to remove the entire upgrade mechanism for substreams in the future.

elenaf9 (approved):
Lgtm!

swarm/src/connection/handler_wrapper.rs
mxinden: :+1: Renamed with https://github.com/libp2p/rust-libp2p/pull/2697/commits/2b187af6507f29a552e3dfd0eff026aee3fd3317.

swarm/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# 0.36.1 - unreleased
+
+- Limit negotiating inbound substreams per connection.
```
```suggestion
- Limit negotiating inbound substreams per connection. See [PR 2697].

[PR 2697]: https://github.com/libp2p/rust-libp2p/pull/2697
```

Issue #2696, opened 2022-06-07
Multiaddr duplicated p2p protocol in relay reservation


stormshield-pj50:
## Summary
P2p protocol is duplicated in Multiaddr when a relay, using TCP transport with port reuse enabled, accepts a reservation request. This leads to an invalid Multiaddr.

In order to reproduce the bug, just turn on port reuse when creating TcpConfig  in relay_v2 example: https://github.com/libp2p/rust-libp2p/blob/fcc987e0f6dd82d0f702acd8b819b4f99a9fb26a/protocols/relay/examples/relay_v2.rs#L49

Please find some of the logs dumped on the listening client showing the duplicated P2p protocol:
```
[2022-06-07T08:38:16Z INFO  client] Relay accepted our reservation request.
[2022-06-07T08:38:16Z INFO  client] Listening on "/ip4/$RELAY_SERVER_IP/tcp/4001/p2p/12D3KooWDpJ7As7BWAwRMfu1VU2WCqNjvq387JEYKDBj4kx6nXTN/p2p/12D3KooWDpJ7As7BWAwRMfu1VU2WCqNjvq387JEYKDBj4kx6nXTN/p2p-circuit/p2p/12D3KooWPjceQrSwdWXPyLLeABRXmuqt69Rg3sBYbU1Nft9HyQ6X"
```

## Expected behavior
The relayed Multiaddr should not contain a duplicated P2p protocol.

## Possible Solution

Thought about 3 possibilities:
1. In Identify: https://github.com/libp2p/rust-libp2p/blob/fcc987e0f6dd82d0f702acd8b819b4f99a9fb26a/protocols/identify/src/identify.rs#L327 when building the observed response in Identify, do not include P2p protocol in the Multiaddr.
1. In transports/tcp: https://github.com/libp2p/rust-libp2p/blob/fcc987e0f6dd82d0f702acd8b819b4f99a9fb26a/transports/tcp/src/lib.rs#L438 the address_translation function should return the same kind of Multiaddr whether port reuse is enabled or not. So maybe patch the function in the match case when port reuse is on, and pop the P2p protocol if present.
1. In relay_v2: https://github.com/libp2p/rust-libp2p/blob/fcc987e0f6dd82d0f702acd8b819b4f99a9fb26a/protocols/relay/src/v2/relay.rs#L758 don't append P2p protocol if already present.

Also wondering if external and observed addresses need to contain the P2p protocol ?

## Version
libp2p version 0.45

[...]

PR #2701, opened 2022-06-10
Multiaddr duplicated p2p protocol in relay reservation (#2696)

stormshield-pj50:
# Description
In protocols/relay/src/v2/relay.rs do not append P2p protocol if already present.

Now the built multiaddr does not contain a dupped P2p protocol.  Logs are OK:
```
[2022-06-07T08:38:16Z INFO  client] Relay accepted our reservation request.
[2022-06-07T08:38:16Z INFO  client] Listening on "/ip4/$RELAY_SERVER_IP/tcp/4001/p2p/12D3KooWDpJ7As7BWAwRMfu1VU2WCqNjvq387JEYKDBj4kx6nXTN/p2p-circuit/p2p/12D3KooWPjceQrSwdWXPyLLeABRXmuqt69Rg3sBYbU1Nft9HyQ6X"
```

## Change checklist

- [ X] I have performed a self-review of my own code


protocols/relay/src/v2/relay.rs
```diff
@@ -757,8 +757,12 @@ impl Action {
                     addrs: poll_parameters
                         .external_addresses()
                         .map(|a| {
-                            a.addr
-                                .with(Protocol::P2p((*poll_parameters.local_peer_id()).into()))
+                            let p2p_proto =
+                                Protocol::P2p(*poll_parameters.local_peer_id().as_ref());
+                            match a.addr.iter().last() {
+                                Some(p) if p == p2p_proto => a.addr,
```
mxinden: I don't think we need to check whether the `PeerId` is our own. This check should happen earlier in the process, in case it is not already happening.

mxinden (commented):
Thanks for tackling this.

Small suggestions. What do you think of the diff below?

``` diff
diff --git a/protocols/relay/src/v2/relay.rs b/protocols/relay/src/v2/relay.rs
index 8d8bb75c..ed5fe6ca 100644
--- a/protocols/relay/src/v2/relay.rs
+++ b/protocols/relay/src/v2/relay.rs
@@ -756,9 +756,13 @@ impl Action {
                     inbound_reservation_req,
                     addrs: poll_parameters
                         .external_addresses()
-                        .map(|a| {
-                            a.addr
-                                .with(Protocol::P2p((*poll_parameters.local_peer_id()).into()))
+                        .map(|a| a.addr)
+                        // Add local peer ID in case it isn't present yet.
+                        .filter_map(|a| match a.iter().last()? {
+                            Protocol::P2p(_) => Some(a),
+                            _ => Some(
+                                a.with(Protocol::P2p(*poll_parameters.local_peer_id().as_ref())),
+                            ),
                         })
                         .collect(),
                 }),
```

Also, would you mind including a changelog entry and a `libp2p-relay` patch version bump?

mxinden (approved):
Sorry for the delay here. Thanks @pj50. 

I am planning to cut a new release soonish. In case you need this earlier, let me know.

PR #2706, opened 2022-06-15
muxers/mplex: Implement `AsyncRead` and `AsyncWrite` for `Substream`

thomaseizinger:
# Description

This aligns the public API of the `libp2p-mplex` module with the one
from `libp2p-yamux`. This change has two benefits:

1. For standalone users of `libp2p-mplex`, the substreams itself are
now useful, similar to `libp2p-yamux` and don't necessarily need to
be polled via the `StreamMuxer`. The `StreamMuxer` only forwards to
the `Async{Read,Write}` implementations.

2. This will reduce the diff of #2648 because we can chunk the one
giant commit into smaller atomic ones.

Whilst keeping master functional, the state of `StreamMuxer` is now
a bit awkward until we merge #2648. However, given we are already
doing things this way for `libp2p-yamux` it shouldn't be an issue?

Will reduce diff in #2648.

## Change checklist


thomaseizinger:
I've benchmarked this branch separately and cannot reliably find any regressions. The benchmarks do sometimes report changes in the 5% range but those are not stable so I think this change is harmless for performance.

mxinden (approved):

mxinden:
> Whilst keeping master functional, the state of `StreamMuxer` is now
> a bit awkward until we merge #2648. However, given we are already
> doing things this way for `libp2p-yamux` it shouldn't be an issue?

Agreed. I like that this is making it more consistent with `libp2p-yamux`.

Thanks isolating this change!

muxers/mplex/Cargo.toml
```diff
@@ -3,7 +3,7 @@ name = "libp2p-mplex"
 edition = "2021"
 rust-version = "1.56.1"
 description = "Mplex multiplexing protocol for libp2p"
-version = "0.33.0"
+version = "0.34.0"
```
mxinden, line 6:
```suggestion
version = "0.33.1"
```

Is this a breaking change?

muxers/mplex/CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+# 0.34.0
+
+- `Substream` now implements `AsyncRead` and `AsyncWrite`. See [PR 2706].
+
+[PR XXXX]: https://github.com/libp2p/rust-libp2p/pull/2706/
```
```suggestion
[PR 2706]: https://github.com/libp2p/rust-libp2p/pull/2706/
```

muxers/mplex/Cargo.toml
thomaseizinger: Yes because `Substream` is a public type and has type parameter now!

mxinden: Oh, good point. Thanks!

muxers/mplex/CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+# 0.34.0
```
```suggestion
# 0.34.0 [unreleased]
```

PR #2707, opened 2022-06-15
core/muxing: Force `StreamMuxer::Substream` to implement `Async{Read,Write}`

### LINKED ISSUE #2721: Remove flush_all from docs
https://github.com/libp2p/rust-libp2p/blob/cb3784a8b307bf1da9cc0956f2ed60081dad2ac7/core/src/muxing.rs#L68

Because the method was removed in https://github.com/libp2p/rust-libp2p/pull/2669

cc @thomaseizinger

thomaseizinger:
# Description

Instead of delegating reading and writing of the substream to the `StreamMuxer`,
we introduce a `SubstreamBox` which implements `Async{Read,Write}`. This is
used as the `Substream` associated type in `StreamMuxerBox`. This allows us to
delete the `read_substream` etc functions from the `StreamMuxer` trait.

Extracted out of #2648.

## Change checklist


core/src/muxing.rs
```diff
@@ -237,252 +166,3 @@ impl<T> StreamMuxerEvent<T> {
...
-                ShutdownState::Flush => match this.muxer.flush_substream(cx, s) {
-                    Poll::Ready(Ok(())) => this.shutdown_state = ShutdownState::Done,
-                    Poll::Ready(Err(err)) => return Poll::Ready(Err(err.into())),
-                    Poll::Pending => return Poll::Pending,
-                },
-                ShutdownState::Done => {
-                    return Poll::Ready(Ok(()));
-                }
-            }
-        }
```
thomaseizinger, line 476: I think it is important to focus on this bit of deleted code here.

Previously, it would have been very likely that a substream created by _any_ `StreamMuxer` implementation would have gone through this logic because using the `StreamMuxer` interface without the help of `SubstreamRef` would be extremely cumbersome.

With this PR, substreams are forced to implement `AsyncRead` and `AsyncWrite` and thus, users can directly use them. Even if we put this piece of logic into `SubstreamBox`, it may be bypassed. The mplex implementation follows the pattern of "first close, then flush" anyway and is also separately tested.

My take here is: It is fine to delete this code as long as we document that all implementations of `StreamMuxer` must create substreams that flush after closing. If we ever get to do https://github.com/libp2p/rust-libp2p/issues/508, this could be part of that test suite. Perhaps we can extract such a test suite out of the mplex tests.

thomaseizinger:
@elenaf9 @mxinden This is stacked on top of https://github.com/libp2p/rust-libp2p/pull/2706 but can be reviewed independently. It is a draft because I still need to update the docs, otherwise it is ready.

core/src/muxing.rs
mxinden: > "first close, then flush" anyway and is also separately tested.

I am a bit confused here. My intuition would be: A call to `poll_close` first flushes the substream before closing it. Same as documented e.g. in Tokio's `AsyncWrite:

> Invocation of a shutdown implies an invocation of flush. Once this method returns Ready it implies that a flush successfully happened before the shutdown happened. That is, callers don’t need to call flush before calling shutdown. They can rely that by calling shutdown any pending buffered data will be written out.

https://docs.rs/tokio/1.6.0/tokio/io/trait.AsyncWrite.html#tymethod.poll_shutdown

Is this not the convention in libp2p? Based on the above it is not. Am I missing something?

thomaseizinger: > Is this not the convention in libp2p? 

At least for mplex, no: https://github.com/libp2p/rust-libp2p/blob/ea487aebfe6eb672b05d2bec2d9d79bbd92450ba/muxers/mplex/src/io.rs#L527

I am not 100% sure why this code was added on this abstraction level. With your quote from above, I think this PR is actually making things more correct then because the now added `AsyncWrite` impl on `libp2p_mplex::Substream` implies flush on close.

Maybe @tomaka can provide some background why this was solved on a generic level previously?

thomaseizinger:
Hmm, I thought I am clever by stacking PRs but with our squash merge strategy, those first commits don't get dropped :/

thomaseizinger:
Rebased only to solve conflicts: https://github.com/libp2p/rust-libp2p/compare/cdfa81ef4d46ec47cf8f1602ee707b986d2569a5..f014c299064236cbf2af11debe57428771935e7b

thomaseizinger:
Docs are updated, this is ready for a full review @elenaf9 @mxinden ! :)

core/CHANGELOG.md
```diff
@@ -1,8 +1,11 @@
-# 0.33.1 - unreleased
+# 0.34.0 - unreleased
```
mxinden, line 1: Missing version bump in `core/Cargo.toml`. I prepared this in https://github.com/libp2p/rust-libp2p/commit/6d4e09aa8f1db8357983f61d2e3a0c3c3a0fb618 with my **very hacky** script. Hope this is helpful @thomaseizinger.

core/src/muxing/boxed.rs
```diff
@@ -206,50 +155,81 @@ impl StreamMuxer for StreamMuxerBox {
...
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "SubstreamBox({})", self.0.type_name())
+    }
+}
+
+/// Workaround because Rust does not allow `Box<dyn AsyncRead + AsyncWrite>`.
+trait AsyncReadWrite: AsyncRead + AsyncWrite + Unpin {
+    /// Helper function to capture the erased inner type.
+    ///
+    /// Used to make the `Debug` implementation of `SubstreamBox` more useful.
```
mxinden:
```suggestion
    /// Used to make the [`Debug`] implementation of [`SubstreamBox`] more useful.
```

mxinden (commented):
Just two comments. Otherwise this looks great to me. Very happy to see this happening.

Might want to give @tomaka a bit more time, in case he has opinions.

core/CHANGELOG.md
thomaseizinger: Hmm, that seems to be based off a different version of master? I applied the commit but it generates a lot of merge conflicts.

Investingating ...

thomaseizinger: Okay, seems to be due to the recent release off master! I've resolved the conflicts :)

muxers/mplex/CHANGELOG.md
```diff
@@ -1,6 +1,9 @@
 # 0.34.0 [unreleased]
 
+- Update to `libp2p-core` `v0.34.0`.
+
 - `Substream` now implements `AsyncRead` and `AsyncWrite`. See [PR 2706].
+- Update to `libp2p-core` `v0.34.0`
```
thomaseizinger:
```suggestion
- `Substream` now implements `AsyncRead` and `AsyncWrite`. See [PR 2706].
- Update to `libp2p-core` `v0.34.0`
```

mxinden (approved):
Feel free to hit the merge button any time.

PR #2710, opened 2022-06-16
core/muxing: Replace `Into<io::Error>` bound on `StreamMuxer` with `std::error::Error`

thomaseizinger:
# Description

This allows us to preserve the type information of a muxer's concrete
error as long as possible. For `StreamMuxerBox`, we leverage `io::Error`'s
capability of wrapping any error that implements `Into<Box<dyn Error>>`.

~This is a stacked PR on top of https://github.com/libp2p/rust-libp2p/pull/2707 and thus a draft but is otherwise ready to review.~

## Change checklist

### DIFF
--- a/core/src/muxing/boxed.rs
+++ b/core/src/muxing/boxed.rs
@@ -1,7 +1,7 @@
 use crate::muxing::StreamMuxerEvent;
 use crate::StreamMuxer;
 use fnv::FnvHashMap;
-use futures::{AsyncRead, AsyncWrite};
+use futures::{ready, AsyncRead, AsyncWrite};
 use parking_lot::Mutex;
 use std::error::Error;
 use std::fmt;
@@ -50,12 +50,13 @@ where
         &self,
         cx: &mut Context<'_>,
     ) -> Poll<Result<StreamMuxerEvent<Self::Substream>, Self::Error>> {
-        match self.inner.poll_event(cx).map_err(into_io_error)? {
-            Poll::Pending => return Poll::Pending,
-            Poll::Ready(StreamMuxerEvent::AddressChange(a)) => {
+        let event = ready!(self.inner.poll_event(cx).map_err(into_io_error)?);
+
+        match event {
+            StreamMuxerEvent::AddressChange(a) => {
                 Poll::Ready(Ok(StreamMuxerEvent::AddressChange(a)))
             }
-            Poll::Ready(StreamMuxerEvent::InboundSubstream(s)) => {
+            StreamMuxerEvent::InboundSubstream(s) => {
                 Poll::Ready(Ok(StreamMuxerEvent::InboundSubstream(SubstreamBox::new(s))))
             }
         }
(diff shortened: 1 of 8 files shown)


core/src/muxing/boxed.rs
```diff
@@ -38,6 +39,7 @@ impl<T> StreamMuxer for Wrap<T>
 where
     T: StreamMuxer,
     T::Substream: AsyncRead + AsyncWrite + Send + Unpin + 'static,
+    T::Error: Into<Box<dyn Error + Send + Sync>>,
```
elenaf9:
```suggestion
    T::Error: Error + Send + Sync>,
```
Nit: Wouldn't this simpler trait bound also work, given that `Into<Box<dyn ..>>` is auto-implemented for `E: Error + Send + Sync`?

swarm/src/connection/substream.rs
```diff
@@ -134,7 +134,7 @@ where
     pub fn poll(
         &mut self,
         cx: &mut Context<'_>,
-    ) -> Poll<Result<SubstreamEvent<TMuxer, TUserData>, IoError>> {
+    ) -> Poll<Result<SubstreamEvent<TMuxer, TUserData>, TMuxer::Error>> {
```
elenaf9, line 137: Not in scope for this PR, but while reviewing I was wondering: Do we even need the `TMuxer` generic on `Muxing`? I think the actual type is always `StreamMuxerBox`. `Muxing` is private the swarm crate and only used by `Connection` and there we already use the concrete `StreamMuxerBox`.

elenaf9 (approved):

swarm/src/connection/substream.rs
thomaseizinger: I am going to inline the entire `Muxing` type into `Connection` once `poll_outbound` is removed from `StreamMuxer`. This is already done in #2648 if you want to see what it looks like :)

I am also tempted to inline `HandlerWrapper` and centralize all the connection handling logic directly in `Connection`. This would remove quite a bit of indirection because substreams would directly be passed from the `StreamMuxer` to the `ConnectionHandler`.

core/src/muxing/boxed.rs
thomaseizinger: Probably yes. I've copied the bounds from `std::io::Error::new` because that is where I am passing the type to. Omitting the `Into` here is probably okay :)

thomaseizinger: I decided to put the `Error` bound directly on the associated type so I don't have to repeat it everywhere. The `Send + Sync` bounds stay on higher layers for consistency with other bounds.

thomaseizinger:
Rebase only to stay up to date with #2707.

core/src/muxing/boxed.rs
mxinden: > I decided to put the `Error` bound directly on the associated type so I don't have to repeat it everywhere.

It is only in 4 locations, right? https://github.com/libp2p/rust-libp2p/pull/2710/commits/a479eb88ed0a8914bedbad856ff18b181b2246df I would prefer bounds to be as close to where they are used as possible. If I am not mistaken `StreamMuxer` doesn't need `type Error` to implement `Error`, right? I liked how previously this pull request would remove the bound on `StreamMuxer`.

thomaseizinger: It doesn't. It also doesn't necessarily need `Substream` to implement `AsyncRead + AsyncWrite` but it is a bit pointless if it doesn't.

If we remove the `Error` bounds, we should also remove `AsyncRead + AsyncWrite` for consistency but that will get a bit redundant I think.

The current style seems to be that _functional_ bounds are put on associated types and auto-traits are left off.

thomaseizinger:
Rebase only includes diff of master since this has been opened: https://github.com/libp2p/rust-libp2p/compare/c2106db78e6af1fc78f33c1f780498c4c5f59a70..a28a7c51eded33886ccd279e64c423afafb5e907

mxinden (approved):

PR #2712, opened 2022-06-16
protocols/kad: Improve options to efficiently retrieve

dignifiedquire:
> Still a draft as I am opening this to discuss the details and direction of this.

# Description

The end goal is to allow the following 

- when calling `get_providers` give the caller results back in real time, not collect all provider records
- when calling `get_providers` allow the total number of records to be of a fixed limit, such that not necessarily 20 peers need to be contacted (as is the case in the current impl)

# Implementation

There are currently two different commits, the first one implements the basic limit functionality, but this gets replaced with a more general solution in 2868e7bb14c9b267b2c50d811895f10cf3945b46, which streams the results and allows the caller to `.finish()` the query when they have enough.

# Things Left To Do

- [x] decide which queries should be transitioned to the new progress based api
- [x] potentially remove `Quorum` based on the new functionality, as this now could be done by the caller if using the progression based api 
- [x] write changelog


protocols/kad/src/behaviour.rs
```diff
@@ -2701,21 +2751,26 @@ impl GetClosestPeersError {
...
 pub type GetProvidersResult = Result<GetProvidersOk, GetProvidersError>;
 
-/// The successful result of [`Kademlia::get_providers`].
+/// The successful progress result of [`Kademlia::get_providers`].
 #[derive(Debug, Clone)]
-pub struct GetProvidersOk {
+pub struct GetProvidersProgress {
     pub key: record::Key,
-    pub providers: HashSet<PeerId>,
+    pub provider: PeerId,
```
mxinden: Why return them one by one? Why not as a batch? One always discovers them in batches, no?

protocols/kad/src/behaviour.rs
```diff
@@ -2412,16 +2445,26 @@ pub enum KademliaEvent {
...
         id: QueryId,
-        /// The result of the query.
+        /// The optional final result of the query.
         result: QueryResult,
         /// Execution statistics from the query.
         stats: QueryStats,
     },
 
+    /// An outbound query has produced a result.
+    OutboundQueryProgressed {
```
mxinden: _Just a slight preference for now. Should probably put more thoughts into this._

How about the following:

- Remove `OutboundQueryProgressed`.
- Rename `OutboundQueryCompleted` to `OutboundQueryProgressed`.
- Add field `index: Index` to `OutboundQueryProgressed` (formerly `OutboundQueryCompleted`) where `enum Index { Continuing(usize), Final(usize) }` indicating the sequence number and whether it is the final update for the query.

mxinden (commented):
I am guessing that this is already significantly speeding up your query times @dignifiedquire? Thanks for providing a patch with the discussion right away.

>  decide which queries should be transitioned to the new progress based api

Without having looked into every one, I would appreciate the consistency across the query types, thus in favor of this.

>  potentially remove Quorum based on the new functionality, as this now could be done by the caller if using the progression based api

I would be in favor of that for the sake of simplicity. I don't think anyone depends on the `Quorum` concept specifically. Not offering `Quorum` is at the expense of potentially doing extra unneeded work, e.g. when contacting another peer immediately followed by a `finish` call through a user. I doubt this is an issue though.

dignifiedquire:
> I am guessing that this is already significantly speeding up your query times @dignifiedquire?

yes very much

protocols/kad/src/behaviour.rs
dignifiedquire: Yeah, I went back and for on this one. The nice thing about this, is that I can reuse the sequence number as the count of results, and the processing is a little more streamlined. But batching is definitely more efficient given they come in as batches.

protocols/kad/src/behaviour.rs
dignifiedquire: I don't have a strong preference of my solution over this, happy to change it, but lets get decide on a version before I go change it again 😅

mxinden:
@kpp @tomaka do you have thoughts on this? This will require a small change in the _Authority Discovery_ module.

@koivunej what do you think of this proposal? I would guess rust-ipfs could benefit from the performance gains as well.

protocols/kad/src/behaviour.rs
mxinden: Unless anyone of the recently tagged people (or anyone else) raises any objections, I suggest going with the above.

protocols/kad/src/behaviour.rs
thomaseizinger: Do we have any idea about the difference in the performance profile?

From an API PoV, it seems nicer to return them one-by-one given that we are already returning a list of events now. For example, if I want to write a log line for each progress entry, being given a set makes this slightly more cumbersome.

protocols/kad/src/behaviour.rs
thomaseizinger: I am unfamiliar with the internals of kademlia but do we know ahead of time, how many we will get? If yes, then always returning both numbers could be another design.

What does the number within `Final` mean? Would it work to include a boolean `completed` in the event and only have `count` otherwise?

The one thing I am worried about with your design suggestion @mxinden is that it introduces more nesting which can be cumbersome to deal with.

dignifiedquire: > I am unfamiliar with the internals of kademlia but do we know ahead of time, how many we will get?

At least not in the providers case

protocols/kad/src/behaviour.rs
dignifiedquire: > Do we have any idea about the difference in the performance profile?

I don't know, the cost is `1` vs `n` events per discovery. From my understanding that cost should be quite low, but I might be mistaken.

koivunej:
Thanks for the ping @mxinden but I've been out of the loop for a while again so I cannot really comment anything. Will try to keep up to date with this and any follow-up release.

protocols/kad/src/behaviour.rs
mxinden: Intuitively I am also assuming that neither has a performance benefit.

Returning them one by one does loose information. Returning them batched gives the user the information that all records within the same batch came from the same source.

kpp:
>  This will require a small change in the Authority Discovery module.

I am not sure how many changes this will require. I looked through the code and found 0 usages of Providers.

protocols/kad/src/behaviour.rs
mxinden: > What does the number within `Final` mean? Would it work to include a boolean `completed` in the event and only have `count` otherwise?

Yep, that is fine as well. I want to provide two pieces of information. (1) how many of these progress updates for the same query have been emitted thus far and (2) whether this one is the final one, i.e. the query overall finished.

mxinden:
> > This will require a small change in the Authority Discovery module.
> 
> I am not sure how many changes this will require. I looked through the code and found 0 usages of Providers.

@kpp the API changes would as well apply to `get_record`, which Substrate is using last time I worked on it.

mxinden:
@dignifiedquire let me know once this is ready for another review. Excited for this to eventually land.

protocols/kad/src/behaviour.rs
dignifiedquire: @mxinden I have started implementing this, but running into challenges with the `QueryStats`. The current design assumes that these can be only extracted once the query is completed, but with your suggested change we would want to provide stats on every progress event. Thoughts?

mxinden: > but running into challenges with the `QueryStats`. The current design assumes that these can be only extracted once the query is completed

I don't see why `QueryStats` could not be returned on each `OutboundQueryProgressed`. The `QueryStats` of a `Query` are updated on each request, thus they are always up-to-date. Am I missing something @dignifiedquire?

I am also fine returning them on the final event only, e.g. embedding them as: `Index { Continuing(usize), Final(usize, QueryStats) }` (ideally more self describing than a tuple).

protocols/kad/src/behaviour.rs
dignifiedquire: @mxinden updated, please check out

protocols/kad/src/behaviour.rs
dignifiedquire: you were right, I was being stupid, please check out the new structure

protocols/kad/src/behaviour.rs
```diff
@@ -2704,7 +2808,10 @@ pub type GetProvidersResult = Result<GetProvidersOk, GetProvidersError>;
 #[derive(Debug, Clone)]
 pub struct GetProvidersOk {
     pub key: record::Key,
+    /// The new set of providers discovered.
     pub providers: HashSet<PeerId>,
+    /// How many providers have been discovered so var.
+    pub providers_so_far: usize,
```
mxinden: Why is this needed? In case a user is interested in all of the providers found, would they not need to keep track of all of them anyways?

protocols/kad/src/behaviour.rs
```diff
@@ -2472,6 +2557,25 @@ pub enum KademliaEvent {
     PendingRoutablePeer { peer: PeerId, address: Multiaddr },
 }
 
+/// Information about progress events.
+#[derive(Debug, Clone)]
+pub struct Index {
```
mxinden: While I proposed the name `Index`, I don't think it is intuitive.

In case anyone has more intuitive naming suggestions, that would be appreciated. //CC @thomaseizinger who usually has good ideas.

protocols/kad/src/behaviour.rs
```diff
@@ -2411,14 +2494,16 @@ pub enum KademliaEvent {
     // is made of multiple requests across multiple remote peers.
     InboundRequest { request: InboundRequest },
 
-    /// An outbound query has produced a result.
-    OutboundQueryCompleted {
+    /// An outbound query has finished.
+    OutboundQueryProgressed {
```
mxinden: Is the doc comment still up to date? Should it not be "An outbound query progressed"?

protocols/kad/src/behaviour.rs
```diff
@@ -2411,14 +2494,16 @@ pub enum KademliaEvent {
...
         /// The ID of the query that finished.
         id: QueryId,
-        /// The result of the query.
+        /// The optional final result of the query.
```
mxinden: Would "The interim result of the query" not be more descriptive? I find this doc comment confusing.

protocols/kad/src/behaviour.rs
```diff
@@ -2411,14 +2494,16 @@ pub enum KademliaEvent {
...
         result: QueryResult,
         /// Execution statistics from the query.
         stats: QueryStats,
+        /// Indicates which event this is, if therer are multiple responses for a single request.
```
```suggestion
        /// Indicates which event this is, if therer are multiple responses for a single query.
```

I think we refer to it as _query_ everywhere else, right?

protocols/kad/src/behaviour/test.rs
```diff
@@ -1354,28 +1358,128 @@ fn get_providers() {
...
                 }
-                SwarmEvent::Behaviour(e) => panic!("Unexpected event: {:?}", e),
-                _ => {}
             }
         });
     }
     QuickCheck::new().tests(10).quickcheck(prop as fn(_))
 }
+
+fn get_providers_limit<const N: usize>() {
```
mxinden, line 1410: I think our first const generic :tada:

mxinden (commented):
Thank you for making the relevant changes and thanks for propagating them across the code-base.

I left a couple of comments. Overall the direction looks great to me.

I am of the opinion that we should apply this pattern to all query types (e.g. as well to `get_closest_peers`). As far as I understand no one is objecting.

I would like to keep the `master` branch consistent at all times. Thus I would prefer the changes to the other query types to happen in this pull request as well. Would you be willing to do that work as well @dignifiedquire?

protocols/kad/src/behaviour.rs
dignifiedquire: My use case was that I need it, and it is a lot easier to track in here, than to track externally. We might also need it internally if we want to do any early cancellations.

protocols/kad/src/behaviour.rs
dignifiedquire: I also dislike the name, but haven't had any better ideas 😓

protocols/kad/src/behaviour.rs
dignifiedquire: it is not

dignifiedquire:
> I would like to keep the `master` branch consistent at all times. Thus I would prefer the changes to the other query types to happen in this pull request as well. Would you be willing to do that work as well @dignifiedquire?

Yeah, I'll work on it, just wanted to get the structure right first.

protocols/kad/src/behaviour.rs
mxinden: Ok. I am fine with leaving it as is.

protocols/kad/src/behaviour.rs
thomaseizinger: What about `ProgressStep`?

```rust
pub struct ProgressStep {
    index: usize,
    last: bool
}
```

mxinden: Sounds good to me. I find `ProgressStep` a lot more intuitive than `Index`.

dignifiedquire: updated

protocols/kad/src/behaviour.rs
dignifiedquire: fixed

protocols/kad/src/behaviour.rs
dignifiedquire: fixed

protocols/kad/src/behaviour.rs
dignifiedquire: fixed

protocols/kad/src/behaviour.rs
dignifiedquire: @mxinden I started expanding this to `GetRecord` and am running into issues with this construction. There is a need to differentiate the type and information between the progress events and the last event, e.g. for `GetRecord` the last event will not contain an actual record, only the optional cache. 
Thoughts?

dignifiedquire:
@mxinden if we remove the quorum from `get_record`, is your expectation that there is no termination condition anymore, and the caller always has to manually terminate, or should there be a default termination?

mxinden:
> @mxinden if we remove the quorum from `get_record`, is your expectation that there is no termination condition anymore, and the caller always has to manually terminate, or should there be a default termination?

I would expect the `QueryPeerIter` to eventually run out of peers and thus I would expect the query to eventually terminate.

Do you think this is too implicit / not intuitive for users? In other words should we consider the potentially unnecessary work significant and thus the whole API a footgun?

protocols/kad/src/behaviour.rs
mxinden: How about instead returning the `cache_candidates` on each progression? I would expect this `BTreeMap` to be small and thus cloning to be cheap. Would you agree @dignifiedquire? Would that solve the issue?

dignifiedquire: not entirely, as I would still need to make the record be optional, which is pretty confusing as a consumer

mxinden: Or we return an event for each contacted peer directly with an optional record.

mxinden:
Friendly ping @dignifiedquire. Would be unfortunate for this to go stale. Anything you would need from my end?

Do I understand correctly, that this is the only reason why iroh depends on a fork of rust-libp2p?

dignifiedquire:
> Do I understand correctly, that this is the only reason why iroh depends on a fork of rust-libp2p?

yes, that is correct

dignifiedquire:
> Do you think this is too implicit / not intuitive for users? In other words should we consider the potentially unnecessary work significant and thus the whole API a footgun?

I am a little afraid that could happen to be honest.

protocols/kad/src/behaviour.rs
dignifiedquire: I actually realized it doesn't help me much after all, so I think we could savely remove this.

dignifiedquire:
> Friendly ping @dignifiedquire. Would be unfortunate for this to go stale. Anything you would need from my end?

Honestly I am not sure how much sense it really makes to change the others behaviour atm, other than the changes I made so far. I will likely have a better understanding down the line when starting to use the rest of the API similarly intensely as I do provider finding atm. 

So I would love if we can find something close to what is there that we can merge, and I (or others) can work on improving the other pieces of the API as needed.

mxinden:
> Honestly I am not sure how much sense it really makes to change the others behaviour atm, other than the changes I made so far.

Do you see any issues with applying the new mechanism to `GetRecord`?

I understand your reasoning on only doing this change to the `GetRecord` system once we have a deeper understanding of it. That said, I do want to keep the `libp2p-kad` API surface consistent, i.e. not offer two different mechanisms to execute operations on the DHT. I argue that this consistency helps both user experience and maintainability. Thus I suggest to either include the changes across the entire API surface, or delaying this pull request until we have more information, but not merge as is.

dignifiedquire:
@mxinden fair enough, I updated the `GetRecord` api, looks pretty alright to me in tests. Do you think that is enough, in terms of updates or do you want any other api calls to be changed to return multiple results?

protocols/kad/src/behaviour.rs
```diff
@@ -680,37 +685,50 @@ where
...
+                Some(PeerRecord {
                     peer: None,
                     record: record.into_owned(),
-                });
+                })
             }
-        }
+        } else {
+            None
+        };
```
mxinden, line 698:
```suggestion
        }.flatten();
```

Should work as well, right?

protocols/kad/src/behaviour.rs
```diff
@@ -680,37 +685,50 @@ where
...
             key,
-            records,
-            quorum,
+            count: 1,
+            record_to_cache: record.as_ref().map(|r| r.record.clone()),
             cache_candidates: BTreeMap::new(),
         };
         let peers = self.kbuckets.closest_keys(&target);
         let inner = QueryInner::new(info);
         let id = self.queries.add_iter_closest(target.clone(), peers, inner); // (*)
```
```suggestion
        let id = self.queries.add_iter_closest(target.clone(), peers, inner);
```

Is this reference this relevant?

protocols/kad/src/behaviour.rs
```diff
@@ -2538,8 +2654,8 @@ pub type GetRecordResult = Result<GetRecordOk, GetRecordError>;
 /// The successful result of [`Kademlia::get_record`].
 #[derive(Debug, Clone)]
 pub struct GetRecordOk {
-    /// The records found, including the peer that returned them.
-    pub records: Vec<PeerRecord>,
+    /// The record found in this iteration, including the peer that returned them.
```
```suggestion
    /// The record found in this iteration, including the peer that returned it.
```

protocols/kad/src/behaviour.rs
```diff
@@ -2133,40 +2228,39 @@ where
...
+                        ref mut count,
                         cache_candidates,
+                        ref mut record_to_cache,
                     } = &mut query.inner.info
                     {
                         if let Some(record) = record {
-                            records.push(PeerRecord {
+                            if record_to_cache.is_none() {
+                                *record_to_cache = Some(record.clone());
+                            }
```
mxinden: Say we discover multiple records for the same key. With the above only the first record is replicated. `libp2p-kad` can not know which of the records is the one to be cached, i.e. it can not make a qualified judgement on which of the records is the best one.

How about removing this automated publishing mechanism and depend on the user to call `Kademlia::put_record_to` instead?

https://github.com/libp2p/rust-libp2p/blob/6da8497af5c3ce076534be68116c01b7f1748383/protocols/kad/src/behaviour.rs#L778-L798

Either way, the comment on `put_record_to` needs updating.

mxinden (commented):
Thanks for updating this pull request refactoring `GetRecord` to the iterative approach. This looks good to me and in my eyes the consistency makes the interface easier to understand.

If I am not mistaken, from a high level, only `FindNode` is missing. In other words, I do think `GetRecord`, `GetProvider` and `FindNode` should follow the same mechanism. Do folks agree?

I don't think we need the iterative mechanism to `PutProvider` and `PutRecord` as of today. I expect these to be happening mostly async, i.e. not within a users hot path. @dignifiedquire would you agree with that intuition with iroh in mind?

dignifiedquire:
> If I am not mistaken, from a high level, only `FindNode` is missing

I am not sure exactly which public facing api you are referring to with `FindNode`?

protocols/kad/src/behaviour.rs
dignifiedquire: I have removed the caching, but I am confused why the behaviour of `put_record_to` changed if I only changed the behaviour of `get_record`, does that rely on that caching to do this?

protocols/kad/src/behaviour.rs
dignifiedquire: doesn't compile for me

protocols/kad/src/behaviour.rs
dignifiedquire: removed, not sure what this was from

protocols/kad/src/behaviour.rs
mxinden: According to the Kademlia paper:

> For caching purposes, once a lookup succeeds, the requesting node stores the (key, value) pair at the closest node it observed to the key that did not return the value.

Before this pull request, when we did a `Kademlia::get_record` with a quorum of 1, we could replicate the (key, value) pair to the closest node that did not have the pair. But if we did a quorum of >1, the `Kademlia::get_record` query would potentially discover more than one value for the given key. Which value would you replicate to the closest node that did not have the value? `libp2p-kad` can not make that choice. With a quorum of >1 `Kademlia::get_record` would thus not replicate the pair to the closest node that did not have the value. Instead it relied on the user first picking the right value and then calling `Kademlia::put_record_to`.

This pull request removes the quorum mechanism. To keep things simple, I suggest also removing the automated replication to the closest node that does not have the value, but instead entirely rely on the user to call `put_record_to`.

Does that make sense @dignifiedquire? Happy to expand further.

mxinden:
> > If I am not mistaken, from a high level, only `FindNode` is missing
> 
> I am not sure exactly which public facing api you are referring to with `FindNode`?

Sorry, this was confusing. I am referring to `Kademlia::get_closest_peers` which sends Kademlia `FIND_NODE` messages to nodes in the network.

protocols/kad/src/behaviour.rs
```diff
@@ -2471,6 +2547,25 @@ pub enum KademliaEvent {
     PendingRoutablePeer { peer: PeerId, address: Multiaddr },
 }
 
+/// Information about progress events.
+#[derive(Debug, Clone)]
+pub struct ProgressStep {
+    /// The index into the event
+    pub count: usize,
```
```suggestion
    pub count: NonZeroUsize,
```

To avoid confusion on whether this is zero-indexed or not, how about using a `NonZeroUsize` here?

mxinden (commented):
@dignifiedquire is this ready for another review?

protocols/kad/src/behaviour.rs
dignifiedquire: I believe https://github.com/libp2p/rust-libp2p/pull/2712/commits/d8ac894f573955c9eef914f7b832798f16e6eb3d does what you requested

protocols/kad/src/behaviour.rs
dignifiedquire: done

dignifiedquire:
> @dignifiedquire is this ready for another review?

should be now 😅

protocols/kad/src/behaviour.rs
```diff
@@ -2472,6 +2461,41 @@ pub enum KademliaEvent {
     PendingRoutablePeer { peer: PeerId, address: Multiaddr },
 }
 
+/// Information about progress events.
+#[derive(Debug, Clone)]
+pub struct ProgressStep {
```
mxinden, line 2552: Needs to be exported in `kad/src/lib.rs`:

https://github.com/libp2p/rust-libp2p/blob/90df86d9e77172123562f8247f436f3e6da891e0/protocols/kad/src/lib.rs#L57-L64

protocols/kad/src/behaviour.rs
```diff
@@ -2472,6 +2461,41 @@ pub enum KademliaEvent {
...
+    }
+
+    /// Generates an index for the case only a single event is emitted.
+    pub(crate) fn single() -> Self {
+        Self::nth_last(1)
+    }
+
+    pub(crate) fn nth(count: usize) -> Self {
+        Self {
+            count: NonZeroUsize::new(count).unwrap(),
```
mxinden: I would prefer not unwrapping here, but instead track a `ProgressStep` internally. See cc1b2bdfc.

protocols/kad/src/behaviour.rs
```diff
@@ -2539,18 +2563,8 @@ pub type GetRecordResult = Result<GetRecordOk, GetRecordError>;
...
-    /// did not return the record, sorted by distance to the record key
-    /// from closest to farthest. How many of these are tracked is configured
-    /// by [`KademliaConfig::set_caching`]. If the lookup used a quorum of
-    /// 1, these peers will be sent the record as a means of caching.
-    /// If the lookup used a quorum > 1, you may wish to use these
-    /// candidates with [`Kademlia::put_record_to`] after selecting
-    /// one of the returned records.
-    pub cache_candidates: BTreeMap<kbucket::Distance, PeerId>,
+    /// The record found in this iteration, including the peer that returned it.
+    pub record: Option<PeerRecord>,
```
mxinden: I find it confusing that this is an `Option`. At least we need to document what `None` means here. Or we refactor `GetRecordOk`:

```rust
enum GetRecordOk {
  FoundRecord (PeerRecord),
  FinishedWithNoAdditionalRecord,
}
```

protocols/kad/src/behaviour.rs
```diff
@@ -2885,18 +2904,14 @@ pub enum QueryInfo {
...
-        /// The records with the id of the peer that returned them. `None` when
-        /// the record was found in the local store.
-        records: Vec<PeerRecord>,
-        /// The number of records to look for.
-        quorum: NonZeroUsize,
-        /// The peers closest to the `key` that were queried but did not return a record,
-        /// i.e. the peers that are candidates for caching the record.
-        cache_candidates: BTreeMap<kbucket::Distance, PeerId>,
+        /// Current index of events.
+        count: usize,
```
mxinden: In my eyes tracking this as a `usize` is error prone and non-descriptive. Why not use `ProgressStep` here?

Done via cc1b2bdfc in case you want to cherry-pick.

protocols/kad/src/behaviour.rs
mxinden: Would require a side-effect within a `Option::map`, which in my eyes is intuitive. Sorry for the noise.

mxinden:
Hope to get a full review done tomorrow. That way you don't have to follow up on many (small) comments, but instead one large batch.

examples/file-sharing.rs
```diff
@@ -326,12 +326,16 @@ mod network {
 
         /// Find the providers for the given file on the DHT.
         pub async fn get_providers(&mut self, file_name: String) -> HashSet<PeerId> {
-            let (sender, receiver) = oneshot::channel();
+            let (sender, mut receiver) = mpsc::channel(0);
```
mxinden: Given that you only put one element into the channel below, why use a channel? Why not stick with the oneshot?

https://github.com/libp2p/rust-libp2p/pull/2712/files#diff-c0d5fae7178c7ba3a6da42df7d8c536e26fdd58762971784e3b554b7b3fc3642R454-R458

misc/metrics/src/kad.rs
```diff
@@ -182,7 +183,7 @@ impl super::Recorder<libp2p_kad::KademliaEvent> for Metrics {
                     libp2p_kad::QueryResult::GetRecord(result) => match result {
                         Ok(ok) => self
                             .query_result_get_record_ok
-                            .observe(ok.records.len() as f64),
+                            .observe(ok.record.as_ref().map(|_| 1).unwrap_or_default() as f64),
```
mxinden: Does it make sense to still track `query_result_get_record_ok` as a histogram if the only values are `0` and `1`? Wouldn't a `Counter` make more sense?

misc/metrics/src/kad.rs
```diff
@@ -18,6 +18,7 @@
 // FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
 // DEALINGS IN THE SOFTWARE.
 
+use libp2p_kad::GetProvidersOk;
```
mxinden: All other `libp2p_kad` imports are happening in-line. I don't have an opinion on either, but I would prefer this file to be consistent.

protocols/kad/src/behaviour.rs
```diff
@@ -446,10 +413,14 @@ where
...
             local_addrs: HashSet::new(),
-            caching: config.caching,
         }
     }
 
+    /// Returns the currently configured `replication_factor`.
+    pub fn replication_factor(&self) -> NonZeroUsize {
+        self.queries.config().replication_factor
+    }
+
```
mxinden: Is this method necessary? We don't provide getters for any of the other config parameters.

Given that it is a static value and given that the user can set it at construction time, why is this getter needed?

protocols/kad/src/behaviour.rs
```diff
@@ -174,24 +171,6 @@ pub struct KademliaConfig {
...
-    /// that do not return a record are not tracked, i.e.
-    /// [`GetRecordOk::cache_candidates`] is always empty.
-    Disabled,
-    /// Up to `max_peers` peers not returning a record that are closest to the key
-    /// being looked up are tracked and returned in [`GetRecordOk::cache_candidates`].
-    /// Furthermore, if [`Kademlia::get_record`] is used with a quorum of 1, the
-    /// found record is automatically sent to (i.e. cached at) these peers. For lookups with a
-    /// quorum > 1, the write-back operation must be performed explicitly, if
-    /// desired and after choosing a record from the results, via [`Kademlia::put_record_to`].
-    Enabled { max_peers: u16 },
```
mxinden, line 196: With caching removed, `PutRecordContext::Cache` needs to be removed as well.

protocols/kad/src/behaviour.rs
```diff
@@ -922,21 +895,44 @@ where
...
+        let stats = QueryStats::empty();
+
+        self.queued_events
+            .push_back(NetworkBehaviourAction::GenerateEvent(
+                KademliaEvent::OutboundQueryProgressed {
+                    id,
+                    result: QueryResult::GetProviders(Ok(GetProvidersOk {
+                        key,
+                        providers_so_far: providers.len(),
+                        providers,
```
mxinden: Why return an event in case `providers` is empty? Same for `get_record` above.

protocols/kad/src/behaviour.rs
```diff
@@ -1455,37 +1437,43 @@ where
...
 
-                Some(KademliaEvent::OutboundQueryCompleted {
+                Some(KademliaEvent::OutboundQueryProgressed {
                     id: query_id,
                     stats: result.stats,
                     result: QueryResult::Bootstrap(Err(BootstrapError::Timeout {
                         peer,
                         num_remaining,
                     })),
+                    step: ProgressStep::single(),
```
mxinden: There might have been `OutboundQueryProgressed` events emitted for this `bootstrap` call.

protocols/kad/src/behaviour.rs
mxinden: Please remove it then. Thanks.

protocols/kad/src/behaviour.rs
```diff
@@ -2704,7 +2714,10 @@ pub type GetProvidersResult = Result<GetProvidersOk, GetProvidersError>;
 #[derive(Debug, Clone)]
 pub struct GetProvidersOk {
     pub key: record::Key,
+    /// The new set of providers discovered.
     pub providers: HashSet<PeerId>,
+    /// How many providers have been discovered so var.
+    pub providers_so_far: usize,
     pub closest_peers: Vec<PeerId>,
```
mxinden: Is `closest_peers` helpful on any of the `OutboundQueryProgressed` but the very last one? Maybe we can split `GetProvidersOk` via an enum similar to `GetRecordOk`:

``` rust 
enum GetProvidersOk {
  FoundProviders {
    ...
  },
  FinishedWithNoAdditionalRecord {
    closest_peers: Vec<PeerId>,
  }
}
```

If I am not mistaken we currently also emit a `OutboundQueryProgressed` with an empty `providers` in `query_finished`. With the above, this would no longer need to be the case.

protocols/kad/src/behaviour.rs
```diff
@@ -2539,18 +2563,8 @@ pub type GetRecordResult = Result<GetRecordOk, GetRecordError>;
...
-    /// If caching is enabled, these are the peers closest
-    /// _to the record key_ (not the local node) that were queried but
-    /// did not return the record, sorted by distance to the record key
-    /// from closest to farthest. How many of these are tracked is configured
-    /// by [`KademliaConfig::set_caching`]. If the lookup used a quorum of
-    /// 1, these peers will be sent the record as a means of caching.
-    /// If the lookup used a quorum > 1, you may wish to use these
-    /// candidates with [`Kademlia::put_record_to`] after selecting
-    /// one of the returned records.
-    pub cache_candidates: BTreeMap<kbucket::Distance, PeerId>,
```
mxinden, line 2573: I think there was a misunderstanding here.

> This pull request removes the quorum mechanism. To keep things simple, I suggest also removing the automated replication to the closest node that does not have the value, but instead entirely rely on the user to call `put_record_to`.

`https://github.com/libp2p/rust-libp2p/pull/2712/#discussion_r984483930`

You removed the automated caching in `query_finished`. That is good. We now rely on the user to call `put_record_to`. But given that you remove the `cache_candidates` on the `GetRecordOk` `struct`, the user no longer knows with which peers it is supposed to call `put_record_to`. We still need to track and emit `cache_candidates`.

Does the above make sense?

protocols/kad/src/behaviour.rs
```diff
@@ -2854,14 +2866,21 @@ pub enum QueryInfo {
     },
 
     /// A query initiated by [`Kademlia::get_closest_peers`].
```
```suggestion
    /// A (repeated) query initiated by [`Kademlia::get_closest_peers`].
```

To be consistent with the comment updates below.

mxinden (commented):
Went through all of the changes. In my eyes there is still work to do to make the API changes consistent.

protocols/kad/src/behaviour.rs
mxinden: Would be fixed with https://github.com/libp2p/rust-libp2p/commit/cc1b2bdfc09a89a56318e38861204ed2c204cfb8.

protocols/kad/src/behaviour.rs
dignifiedquire: oh sweet, thank you!

protocols/kad/src/behaviour.rs
dignifiedquire: good idea, fixed

protocols/kad/src/behaviour.rs
dignifiedquire: fixed

protocols/kad/src/behaviour.rs
dignifiedquire: fixed

protocols/kad/src/behaviour.rs
dignifiedquire: should be fixed

protocols/kad/src/behaviour.rs
dignifiedquire: you are totally right, fixed this

protocols/kad/src/behaviour.rs
dignifiedquire: done

protocols/kad/src/behaviour.rs
dignifiedquire: only needed internally, adjusted visibility

misc/metrics/src/kad.rs
dignifiedquire: fixed

examples/file-sharing.rs
dignifiedquire: reverted, this was an artificat from a different version

misc/metrics/src/kad.rs
dignifiedquire: yes it does, changed

protocols/kad/src/behaviour.rs
dignifiedquire: yes, very good idea, done

protocols/kad/src/behaviour.rs
dignifiedquire: fixed

protocols/kad/src/behaviour.rs
dignifiedquire: Yes very much, my mistake entirely. I brought back `KademliaCaching` and `cache_candidates` to enable tracking this now.

dignifiedquire:
> Went through all of the changes.

Thank you :) I think addressed all of them now

>  In my eyes there is still work to do to make the API changes consistent.

In addition to the inline comments you left?

thomaseizinger:
Let me know if you want input on this from me as well, otherwise I'll leave it to you @mxinden :)

examples/file-sharing.rs
```diff
@@ -331,7 +331,11 @@ mod network {
                 .send(Command::GetProviders { file_name, sender })
                 .await
                 .expect("Command receiver not to be dropped.");
-            receiver.await.expect("Sender not to be dropped.")
+            let mut out = HashSet::new();
+            if let Ok(h) = receiver.await {
+                out.extend(h);
+            }
+            out
```
mxinden: `h` is already a `HashSet`, why create a new one and then extend it?

examples/file-sharing.rs
```diff
@@ -440,6 +448,19 @@ mod network {
...
+                        result: QueryResult::GetProviders(..),
+                        ..
+                    },
+                )) => {
+                    // Drop channel to signal query is complete.
+                    let _ = self
+                        .pending_get_providers
+                        .remove(&id)
+                        .expect("Completed query to be previously pending.");
+                }
```
mxinden: How about calling `Kademlia::query_mut()::finish()` when discovering the first provider record? We found one record, we are not interested in any additional records.

With this suggestion the second match on `QueryResult::GetProviders` would only need to match on `GetProvidersOk::FinishedWithNoAdditionalRecord`.

protocols/kad/src/behaviour.rs
```diff
@@ -446,10 +445,14 @@ where
             provider_record_ttl: config.provider_record_ttl,
             connection_idle_timeout: config.connection_idle_timeout,
             local_addrs: HashSet::new(),
-            caching: config.caching,
         }
     }
 
+    /// Returns the currently configured `replication_factor`.
+    pub(crate) fn replication_factor(&self) -> NonZeroUsize {
```
```suggestion
    fn replication_factor(&self) -> NonZeroUsize {
```

Why does this need to be `pub(crate)`? Tests succeed with this diff on my machine.

examples/file-sharing.rs
mxinden: See 1398ee6a6.

examples/file-sharing.rs
```diff
@@ -440,6 +448,19 @@ mod network {
...
+                    KademliaEvent::OutboundQueryProgressed {
+                        id,
+                        result: QueryResult::GetProviders(..),
+                        ..
+                    },
+                )) => {
+                    // Drop channel to signal query is complete.
+                    let _ = self
+                        .pending_get_providers
+                        .remove(&id)
```
mxinden: Given the `remove` a couple of lines above, wouldn't this panic?

examples/file-sharing.rs
mxinden: See f8d36f23c.

examples/file-sharing.rs
mxinden: See f8d36f23c.

protocols/kad/src/behaviour.rs
mxinden: See 24f54bb90.

protocols/kad/src/behaviour.rs
```diff
@@ -680,36 +685,49 @@ where
...
+        };
 
-        let done = records.len() >= quorum.get();
         let target = kbucket::Key::new(key.clone());
         let info = QueryInfo::GetRecord {
             key,
-            records,
-            quorum,
+            step: ProgressStep::first(),
+            found_a_record: record.is_some(),
```
mxinden: I find it confusing that `step` is later on altered based on `record` and `found_a_record` is altered based on `record` right away.

15b30a25d would fix that.

protocols/kad/src/behaviour.rs
```diff
@@ -922,21 +938,47 @@ where
...
+                            providers,
+                        })),
+                        step: ProgressStep::first(),
+                        stats,
+                    },
+                ));
+            let query = self.query_mut(&id).expect("just inserted");
+            if let QueryInfo::GetProviders { step, .. } = &mut query.query.inner.info {
+                *step = step.next();
+            }
```
mxinden: Same here. Altering this after creation is confusing to me. I suggest to do the same change as done for `get_record`. See 5e8f3cce8.

protocols/kad/src/behaviour.rs
```diff
@@ -2472,6 +2523,43 @@ pub enum KademliaEvent {
...
+pub struct ProgressStep {
+    /// The index into the event
+    pub count: NonZeroUsize,
+    /// Is this the final event?
+    pub last: bool,
+}
+
+impl ProgressStep {
+    // TODO Why the pub?
+    pub(crate) fn first() -> Self {
```
mxinden: I don't think `pub(crate)` is needed. See 0f0399bae.

protocols/kad/src/behaviour.rs
```diff
@@ -2702,10 +2787,15 @@ pub type GetProvidersResult = Result<GetProvidersOk, GetProvidersError>;
...
-    pub providers: HashSet<PeerId>,
-    pub closest_peers: Vec<PeerId>,
+pub enum GetProvidersOk {
+    FoundProviders {
+        key: record::Key,
+        /// The new set of providers discovered.
+        providers: HashSet<PeerId>,
+    },
+    FinishedWithNoAdditionalRecord {
+        closest_peers: Vec<PeerId>,
```
mxinden, line 2815: As an aside, while we have a `put_record_to` to put a record to an individual node, we don't have a `put_provider_record_to`. 

That seems surprising to me, but I guess not needed by anyone out there thus far.

(Not to be fixed in this pull request.)

protocols/kad/src/behaviour/test.rs
```diff
@@ -801,7 +803,6 @@ fn get_record() {
 
 #[test]
 fn get_record_many() {
-    // TODO: Randomise
```
mxinden: Why should we no longer randomise this value?

protocols/kad/src/behaviour/test.rs
```diff
@@ -1352,30 +1366,137 @@ fn get_providers() {
...
+}
+
+#[test]
+fn get_providers_limit_n_1() {
+    get_providers_limit::<1>();
+}
+
+#[test]
+fn get_providers_limit_n_2() {
+    get_providers_limit::<1>();
```
mxinden:
```suggestion
    get_providers_limit::<2>();
```

protocols/kad/src/query.rs
```diff
@@ -419,6 +419,18 @@ impl<TInner> Query<TInner> {
...
+        match self.peer_iter {
+            QueryPeerIter::Closest(ref iter) => {
+                Either::Left(Either::Left(iter.as_intermediary_result()))
+            }
+            QueryPeerIter::ClosestDisjoint(ref iter) => {
+                Either::Left(Either::Right(iter.as_intermediary_result()))
+            }
+            QueryPeerIter::Fixed(ref iter) => Either::Right(iter.as_intermediary_result()),
+        }
+    }
```
mxinden: Given that we only return the closest nodes on the alst `QueryProgressed`, this is no longer called, right? Mind removing it?

protocols/kad/src/query/peers/closest.rs
```diff
@@ -365,6 +365,19 @@ impl ClosestPeersIter {
...
+            .filter_map(|(_, peer)| {
+                if let PeerState::Succeeded = peer.state {
+                    Some(peer.key.clone().into_preimage())
+                } else {
+                    None
+                }
+            })
+            .take(self.config.num_results.get())
+    }
+
```
mxinden: Same here.

protocols/kad/src/query/peers/closest/disjoint.rs
```diff
@@ -331,6 +331,15 @@ impl ClosestDisjointPeersIter {
...
     }
+
+    pub fn as_intermediary_result(&self) -> impl Iterator<Item = PeerId> + '_ {
+        let result_per_path = self
+            .iters
+            .iter()
+            .map(|iter| iter.clone().into_result().map(Key::from));
+
+        ResultIter::new(self.target.clone(), result_per_path).map(Key::into_preimage)
+    }
```
mxinden: And here.

protocols/kad/src/query/peers/fixed.rs
```diff
@@ -170,6 +170,16 @@ impl FixedPeersIter {
...
+
+    pub fn as_intermediary_result(&self) -> impl Iterator<Item = PeerId> + '_ {
+        self.peers.iter().filter_map(|(p, s)| {
+            if let PeerState::Succeeded = s {
+                Some(*p)
+            } else {
+                None
+            }
+        })
+    }
```
mxinden: And this one.

mxinden (commented):
Thanks for applying my suggestions. This is likely the last large review.

Most (not all) of the new suggestions are fixed on my branch. If you like them all, you can just merge my branch into yours.

https://github.com/mxinden/rust-libp2p/tree/feat-kad-count

protocols/kad/src/behaviour.rs
```diff
@@ -446,10 +445,14 @@ where
             provider_record_ttl: config.provider_record_ttl,
             connection_idle_timeout: config.connection_idle_timeout,
             local_addrs: HashSet::new(),
-            caching: config.caching,
         }
     }
 
+    /// Returns the currently configured `replication_factor`.
```
```suggestion
```

Thinking about this some more, this is only needed in tests. In those tests, we can get the value from the `Default` implementation of `KademliaConfig`. 

I don't think we should introduce a public facing method just for testing.

examples/file-sharing.rs
dignifiedquire: @mxinden I am not seeing that commit anywhere, did you push this?

mxinden: My bad. Updated my branch now:

https://github.com/mxinden/rust-libp2p/commits/feat-kad-count

Unfortunately the hashes above don't link. I added the comments before I pushed the commits.

dignifiedquire: thanks, what a service, getting suggestions already commited <3

protocols/kad/src/behaviour.rs
dignifiedquire: 👍

protocols/kad/src/behaviour.rs
dignifiedquire: fixed

protocols/kad/src/query/peers/fixed.rs
dignifiedquire: removed

protocols/kad/src/query/peers/closest/disjoint.rs
dignifiedquire: removed

protocols/kad/src/query/peers/closest.rs
dignifiedquire: removed

protocols/kad/src/query.rs
dignifiedquire: removed

protocols/kad/src/behaviour/test.rs
dignifiedquire: fixed

protocols/kad/src/behaviour/test.rs
dignifiedquire: readded

protocols/kad/src/behaviour.rs
dignifiedquire: merged, 🙏

protocols/kad/src/behaviour.rs
dignifiedquire: merged, 🙏

protocols/kad/src/behaviour.rs
dignifiedquire: merged, 🙏

examples/file-sharing.rs
dignifiedquire: merged, 🙏

protocols/kad/src/behaviour.rs
dignifiedquire: merged, 🙏

examples/file-sharing.rs
dignifiedquire: merged, 🙏

dignifiedquire:
@mxinden applied CR & your patches and rebased

protocols/kad/src/behaviour.rs
```diff
@@ -2492,6 +2547,42 @@ pub enum KademliaEvent {
...
+    }
+
+    fn next(&self) -> Self {
+        assert!(!self.last);
+
+        Self {
+            count: self
+                .count
+                .checked_add(1)
+                .expect("Not to exhaust usize space."),
```
```suggestion
                // TODO: Use checked_add with Rust 1.64.
                .saturating_add(1),
```

This is only in stable with `1.64`.

```
error[E0658]: use of unstable library feature 'nonzero_ops'
    --> protocols/kad/src/behaviour.rs:2579:18
     |
2579 |                 .checked_add(1)
     |                  ^^^^^^^^^^^
     |
     = note: see issue #84186 <https://github.com/rust-lang/rust/issues/84186> for more information
     ```

Comment (dignifiedquire) [reply protocols/kad/src/behaviour.rs]: `saturating_add` is only available on `1.64` as well: https://doc.rust-lang.org/std/num/struct.NonZeroUsize.html#method.saturating_add 

Fixed it in an ugly way..

File: protocols/kad/src/behaviour.rs [hunk 44]
```diff
@@ -2492,6 +2547,37 @@ pub enum KademliaEvent {
...
+
+    fn first_and_last() -> Self {
+        let mut first = ProgressStep::first();
+        first.last = true;
+        first
+    }
+
+    fn next(&self) -> Self {
+        assert!(!self.last);
+        let count = NonZeroUsize::new(self.count.get() + 1).expect("Not to exhaust usize space.");
```
mxinden: Nit pick: I don't think `+` in Rust overflows but instead panics, thus the `expect` is actually not checking for `exhaust`.

``` rust
#[allow(arithmetic_overflow)]

fn main() {
    assert_eq!(255_u8 + 1, 0);
}
```

```
Compiling playground v0.0.1 (/playground)
    Finished dev [unoptimized + debuginfo] target(s) in 1.61s
     Running `target/debug/playground`
thread 'main' panicked at 'attempt to add with overflow', src/main.rs:4:16
note: [run with `RUST_BACKTRACE=1` environment variable to display a backtrace](https://play.rust-lang.org/#)
```

```suggestion
        let count = NonZeroUsize::new(self.count.get() + 1).expect("Adding 1 not to result in 0.");
```

dignifiedquire: It's a bit more complicated, overflow checks are only inserted in `debug` builds, in `release` builds this will wrap around ( and the above program you posted will pass)

mxinden:
@dignifiedquire please don't force push on rust-libp2p pull requests. In particular this makes collaborating on a pull request hard. We squash-merge, thus a clean commit history, while appreciated, isn't necessary.

dignifiedquire:
Sorry about that, force of habit  😓

protocols/kad/src/behaviour.rs
thomaseizinger: > `saturating_add` is only available on `1.64` as well: https://doc.rust-lang.org/std/num/struct.NonZeroUsize.html#method.saturating_add 
> 
> Fixed it in an ugly way.. 

I don't mind bump MSRV as long as it is documented in the changelog :)

dignifiedquire: seems not really worth imho for this

dignifiedquire:
@mxinden anything I can help with to move this forward?

mxinden (commented):
Great. This looks good to me from a code perspective. Only missing piece would be a changelog entry, allowing folks unfamiliar with this change to upgrade their downstream code easily. Things to cover:

- Rename to `OutboundQueryProgressed`.
- Mention that multiple events per query are emitted.
- Mention that one needs to manually finish a query in case one event is enough.
- ...

dignifiedquire:
@mxinden first pass at changelog done

protocols/kad/CHANGELOG.md
```diff
@@ -16,10 +16,18 @@
   This would eventually lead to warning that says: "New inbound substream to PeerId exceeds inbound substream limit. No older substream waiting to be reused."
   See [PR 3152].
 
+- Refactor APIs to be streaming.
+  - Renamed `KademliaEvent::OutboundQueryCompleted` to `KademliaEvent::OutboundQueryProgressed`
+  - Instead of a single event `OutboundQueryCompleted`, there are now multiple events emitted, allowing the user to process them as they come in (via the new `OutboundQueryProgressed`).
```
```suggestion
  - Instead of a single event `OutboundQueryCompleted`, there are now multiple events emitted, allowing the user to process them as they come in (via the new `OutboundQueryProgressed`). See `ProgressStep` to identify the final `OutboundQueryProgressed` of a single query.
```

protocols/kad/CHANGELOG.md
```diff
@@ -16,10 +16,18 @@
...
+  - `get_providers` and `get_record` do not terminate automatically anymore, the caller needs to call `query.finish()`, on termination conditions.
```
```suggestion
  - To finish a query early, i.e. before the final `OutboundQueryProgressed` of the query, a caller needs to call `query.finish()`.
```

They do terminate automatically. Though not before the final `OutboundQueryProgressed`. The above suggestion would make that easier to understand for me.

protocols/kad/CHANGELOG.md
```diff
@@ -16,10 +16,18 @@
...
+  - There is no more automatic caching of records, the user has to manually call `put_record_to` on the `QueryInfo::GetRecord.cache_candiates`.
```
```suggestion
  - There is no more automatic caching of records. The user has to manually call `put_record_to` on the `QueryInfo::GetRecord.cache_candidates` to cache a record to a close peer that did not return the record on the foregone query.
```

mxinden (commented):
Some small suggestions. Otherwise this ready to go from my end.

dignifiedquire:
> Some small suggestions

thanks a lot, merged them all

mxinden (approved):
Looks good to me. Thanks for bearing with me @dignifiedquire!

rkuhn:
Hmm, this is quite a lot to read: now that I’m facing a somewhat non-obvious upgrade for ipfs-embed (as I’ve never used Kademlia), what is the upgrade path for the quorum removal from `get_record`? @mxinden @dignifiedquire

mxinden:
@rkuhn can you ping me on your upgrade pull request? Happy to guide or provide patches.

I am assuming that you already read the changelog entry. Unfortunately that is not a step-by-step instruction manual.

rkuhn:
Here’s [what I came up with](https://github.com/ipfs-rust/ipfs-embed/commit/11bef10b9d360d902a2f1e92a57ade4367897f6a#diff-79804e689185ec7a5af51be11f39d083f554e434a50058564a49ff8381741a1dR245-R264), hope it isn’t too far off :-)

PR #2713, opened 2022-06-19
Upgrade if-watch to fix windows crate vulnerability

TheQuantumPhysicist:
Hello

This is important and should be merged ASAP as it fixes a vulnerability

https://github.com/microsoft/windows-rs/commit/afe32525c22209aa8f632a0f4ad607863b51796a

Please create a new release after merging this.

Thank you.

### DIFF
--- a/protocols/mdns/Cargo.toml
+++ b/protocols/mdns/Cargo.toml
@@ -15,7 +15,7 @@ async-io = "1.3.1"
 data-encoding = "2.3.2"
 dns-parser = "0.8.0"
 futures = "0.3.13"
-if-watch = "1.0.0"
+if-watch = "1.1.0"
 lazy_static = "1.4.0"
 libp2p-core = { version = "0.33.0", path = "../../core", default-features = false }
 libp2p-swarm = { version = "0.36.0", path = "../../swarm" }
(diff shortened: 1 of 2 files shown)


rkuhn:
Thanks for raising this! The wording is a bit too hasty, though, as there is no actual vulnerability being fixed — since no code changes were necessary in if-watch, all handlers were `Send` already. Still good to keep things updated :-)

thomaseizinger:
Additionally, the new version is semver compatible so any end-user can already update to this version: `cargo update -p if-watch` should do the trick.

I guess we should still merge this because it is not good to allow a version that can cause a security problem.

mxinden (commented):
Looks good to me, though needs a patch version bump on `libp2p-mdns` and `libp2p-tcp` and a changelog entry each. See https://github.com/libp2p/rust-libp2p/blob/master/docs/release.md#development-between-releases.

mxinden:
> Additionally, the new version is semver compatible so any end-user can already update to this version: `cargo update -p if-watch` should do the trick.

Agreed here. Thus I won't cut a new release right away. @TheQuantumPhysicist let us know in case we are missing something.

TheQuantumPhysicist:
You got it right guys. Though to clarify, the motivation for this is that every audit software, including git's dependabot, was screaming that there's a vulnerability, which we don't dismiss usually to avoid mistakes. I think with a precise update of if-watch (which wasn't possible before for the windows crate alone) the problem can be solved and this isn't super-urgent. Cheers!

mxinden:
> Looks good to me, though needs a patch version bump on `libp2p-mdns` and `libp2p-tcp` and a changelog entry each. See https://github.com/libp2p/rust-libp2p/blob/master/docs/release.md#development-between-releases.

Closing here since these are still missing. Feel free to comment or open a new pull request in case you still want this updated.

Issue #2709, opened 2022-06-15
Override protocol id prefix on GossipsubConfigBuilder


bernardoaraujor:
## Description

I need to be able to set the entire `GossipSub`'s `ProtocolId`, not only the prefix.

## Motivation

I'm working on a Rust implementation of [Waku](https://rfc.vac.dev/spec/10/), an evolution of Ethereum Whisper: https://github.com/bernardoaraujor/waku-rs

At this point, I have a minimal working implementation of [11/WAKU2-RELAY](https://rfc.vac.dev/spec/11/), [13/WAKU2-STORE](https://rfc.vac.dev/spec/13/) and [19/WAKU2-LIGHTPUSH](https://rfc.vac.dev/spec/19/).

When trying to connect it to the [Nim reference implementation](https://github.com/status-im/nwaku) without success, I noticed the following messages on the Nim logs:
```
DBG 2022-06-15 18:23:17.558-03:00 no handlers                                topics="libp2p multistream" tid=77736 file=multistream.nim:173 conn=12D*PH8vPd:62aa4dc5649138700bc0af39 protocol=//vac/waku/relay/2.0.0/1.1.0
DBG 2022-06-15 18:23:17.558-03:00 Could not establish send connection        topics="libp2p pubsubpeer" tid=77736 file=pubsubpeer.nim:205 msg="Unable to select sub-protocol @[\"/vac/waku/relay/2.0.0\", \"/vac/waku/relay/2.0.0\"]
```

where `protocol=//vac/waku/relay/2.0.0/1.1.0` is clearly not what I intended with [`RELAY_PROTOCOL_ID`](https://github.com/bernardoaraujor/waku-rs/blob/1176db196b720545ebb3e3150142b38c10e8dad8/waku-protocol/src/waku_relay/network_behaviour.rs#L16).

It turns out that's a result of [`GossipsubConfigBuilder::default().protocol_id_prefix(RELAY_PROTOCOL_ID)`](https://github.com/bernardoaraujor/waku-rs/blob/1176db196b720545ebb3e3150142b38c10e8dad8/waku-protocol/src/waku_relay/network_behaviour.rs#L44), which [uses `RELAY_PROTOCOL_ID` as a mere prefix](https://docs.rs/libp2p-gossipsub/latest/src/libp2p_gossipsub/protocol.rs.html#97) for `ProtocolId`, adding extra `/` and `/1.1.0`.

## Requirements

[...]

PR #2718, opened 2022-06-20
allow custom gossipsub protocol id

bernardoaraujor:
# Description

Currently, the only way to set a GossipSub protocol ID is via `GossipsubConfigBuilder::protocol_id_prefix()`, which uses the input as a mere prefix, and the resulting protocol ID is in the format `/{prefix}/1.x.0 `.

This PR introduces the possibility of custom GossipSub protocol IDs, to be set via `GossipsubConfigBuilder::protocol_id()`.

If `GossipsubConfigBuilder::protocol_id()` is called, then `GossipsubHandler` is created via `GossipsubHandler::new_custom` and whatever was previously set via `GossipsubConfigBuilder::protocol_id_prefix()` is simply ignored.

https://github.com/libp2p/rust-libp2p/issues/2709

## Open Questions

The existing test-cases all pass.
I'm also able to successfully test on the use-case described on #2709 

Is it necessary to add more test cases?
The only tests I could find that operate on a similar level were [`create_config_with_message_id_as_*`](https://github.com/libp2p/rust-libp2p/blob/ea487aebfe6eb672b05d2bec2d9d79bbd92450ba/protocols/gossipsub/src/config.rs#L895), however that was still not exactly the same thing.

I'm also not sure what are the appropriate crates for a changelog entry.

## Change checklist


MarcoPolo:
Haven't looked too closely at this yet, but:

> Is it necessary to add more test cases?

Yes, please do.

> I'm also not sure what are the appropriate crates for a changelog entry.

https://github.com/libp2p/rust-libp2p/blob/0d9bf4f2e785cf09042b3e147d18a5c7462929ee/protocols/gossipsub/CHANGELOG.md See https://github.com/libp2p/rust-libp2p/blob/master/docs/release.md#development-between-releases for more details.

I wonder if the change can be simplified if you did the prefix logic sooner. Create the protocol id at the beginning and use that later on rather than adding the `x_custom` variants to functions.

thomaseizinger (commented):
As far as I know, the current functionality of only configuring the prefix was introduced because that is the _minimal_ amount of configuration needed to allow re-use of the code between floodsub, gossipsub v1 and gossipsub v1.1.

If we now want to support entirely custom protocols, I think a cleaner way of achieving this would be to build up the list of protocol strings in the `GossipsubConfigBuilder` and pass the final list down to the handler.

Is this what you also meant @MarcoPolo?

Also tagging @AgeManning as the original author.

MarcoPolo:
> If we now want to support entirely custom protocols, I think a cleaner way of achieving this would be to build up the list of protocol strings in the GossipsubConfigBuilder and pass the final list down to the handler.
>
> Is this what you also meant @MarcoPolo?

Yup. Thank you

AgeManning (commented):
The reason I went with just a prefix is to allow users to form their own networks and not be stuck connecting to other gossipsub peers on the same network. I kept the version and general format because there were live networks hosting a mix of gossipsub versions. 

There's a few things I think we need to consider for this change:
1. What versions does a custom protocol-id support? - The current design optionally supports floodsub, gossipsub 1.0 and gossipsub 1.1. It will advertise these protocol-ids such that by default this can connect to networks that have old versions of gossipsub. If we use a custom protocol id without versioning, what versions specifically should this support? There are protobuf message changes between 1.0 and 1.1 (such as peer exchange in pruning) and currently every node that connects to us we identify which version they are (by the protocol-id) so we know what messages we can/cannot send. There is currently work being done on a v 1.2 that is backwards compatible. For 1.2, we can just add a new protocol-id to the list and add these nodes to existing networks and have it work fine. If we say that a custom protocol id only supports 1.1 and nothing else, these networks will not be upgradeable or work well when 1.2 gets released. The reason only the prefix was given to users is to allow for this upgradability of networks. If we want to add a custom protocol id that can be versionless and just uses the latest version of gossipsub, then when gossipsub updates it will not work with previous networks and/or the user must manually adjust the custom protocol-id of the network to distinguish versions. I think we should be explicit of how we want to handle this. 
2. We currently use the protocol-id list to identify each node that connects to us. We then preform different logic in the behaviour based on the version of the node. If this PR wants the `PeerKind::Custom` variant to behave as a v1.1 node, some of the behaviour match statements need to be modified to reflect this.
3. As @MarcoPolo points out, i think the code can be simplified. I dont think we need two functions to instantiate a handler. We just need to make a decision of what versions `PeerKind::Custom` is actually going to support. You could manually disable floodsub support and only support v1.1 and use the functions that already exist. Again this may run into complications when the versions get updated. I think that if the decision is that the custom protocol id, just behaves as a 1.1 node, then we don't need the `Custom` variant, and you can just set the protocol id to be a 1.1 node (this would simplify the code). 

The fundamental issue I think is that gossipsub needs to be able to distinguish other versions of itself in order for us to upgrade gossipsub and still be backwards compatible with existing networks. If your network removes the gossipsub versioning, you may run into issues with future gossipsub versions working on your network. If this is really what the network wants to do, perhaps the best way is to just set the custom protocol-id to be the protocol-id for the current version (currently PeerKind::Gossipsub1_1, but this will probably change in the future to Gossipsub1_2, when its complete).

bernardoaraujor:
After the inputs from @AgeManning I realized that introducing a new `PeerKind::Custom` is likely creating more problems than solving, so I got rid of it.

Now we're using `GossipsubConfig.protocol_id` to hold the actual string, to be used as either prefix or custom protocol id. And `GossipsubConfig.protocol_id_prefix` is now a `bool` flag used to signal which logic to follow.

I also followed the advice from @MarcoPolo + @thomaseizinger, and moved the changes up in the logic. Now a `ProtocolConfig` is created inside `Gossipsub::new_handler`, and the flag is used to toggle the prefix logic.

So we're effectively simply introducing the option to override the prefix logic without introducing a new `PeerKind`, which I believe makes things simpler. Regardless of how the `protocol_id` is structured, the `PeerKind` will still be `Gossipsubv1_1` and/or `Gossipsub`.

@AgeManning do you see any issues with this approach? I couldn't find any part of the code that actually matches on the `protocol_id` string, only on `PeerKind`, so I would assume this approach wouldn't introduce any undefined behaviour?

protocols/gossipsub/src/protocol.rs
```diff
@@ -59,20 +59,20 @@ impl ProtocolConfig {
...
         validation_mode: ValidationMode,
         support_floodsub: bool,
+        id_is_prefix: bool,
     ) -> ProtocolConfig {
-        // support version 1.1.0 and 1.0.0 with user-customized prefix
         let mut protocol_ids = vec![
-            ProtocolId::new(id_prefix.clone(), PeerKind::Gossipsubv1_1),
-            ProtocolId::new(id_prefix, PeerKind::Gossipsub),
+            ProtocolId::new(id.clone(), PeerKind::Gossipsubv1_1, id_is_prefix),
+            ProtocolId::new(id, PeerKind::Gossipsub, id_is_prefix),
```
AgeManning: This is kind of what I was saying. We need to be explicit about what version a custom id will have. 

The way you have currently implemented this, the node will advertise two protocol ids (assuming `id_is_prefix` is false). 
- `id` - which matches `Gossipsubv1_1` and
- `id` - which matches `Gossipsub`. 

When we see another node on the network with this custom id, it will just match the first one and therefore we will assume all nodes are 1.1 nodes. If there is a 1.0 node on the network, we may send it messages it doesn't understand and they will treat these as invalids. 

To be explicit, perhaps the config should allow a user to provide a `Vec<ProtocolId>` which can map which node versions match to which id. 
If you just want to assume every single node on the network is 1.1 (which is likely to cause issues when 1.2 comes) then maybe the config parameter can be something like `protocol_id_v1_1` and we disable support for all other versions (this is done by just advertising this protocol_id and leaving everything else unsupported).

bernardoaraujor:
@AgeManning thanks for the feedback.

I introduced `GossipsubVersion` enum to `config.rs`. That way, whenever the user makes the conscious choice of using a custom protocol id (by calling `GossipsubConfigBuilder::protocol_id` instead of `GossipsubConfigBuilder::protocol_id_prefix`), they have to provide either `GossipsubVersion::V1_0` or `GossipsubVersion::V1_1` (or `GossipsubVersion::V1_2` in the future) as argument.

During the construction of `ProtocolConfig`, this enum is matched so that if a custom protocol id is to be set, only a single `PeerKind` will be used.

protocols/gossipsub/src/config.rs
```diff
@@ -943,4 +968,28 @@ mod test {
         let result = builder.message_id(&get_gossipsub_message());
         assert_eq!(result, get_expected_message_id());
     }
+
+    #[test]
+    fn create_config_with_protocol_id_prefix() {
```
thomaseizinger, line 984: Can we extend these tests to go all the way to serialised protocol IDs?

thomaseizinger (commented):
Thanks! I think this is much simpler already :)

One question.

protocols/gossipsub/src/config.rs
bernardoaraujor: thanks for the suggestion @thomaseizinger 
done on https://github.com/libp2p/rust-libp2p/pull/2718/commits/c7ef6fa182c7f3a92f9b87d44cb9662302b4083c

AgeManning (approved):
Yep this looks good to me.  Thanks :)

mxinden (commented):
Needs a changelog entry. Otherwise looks good to me.

Thanks for the collaboration here everyone.

bernardoaraujor:
@mxinden done.

protocols/gossipsub/CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
 
 - Update to `libp2p-swarm` `v0.37.0`.
 
+- Allow for custom protocol ID via `GossipsubConfigBuilder::protocol_id()`. See [PR 2718].
```
mxinden, line 7:
```suggestion
- Allow for custom protocol ID via `GossipsubConfigBuilder::protocol_id()`. See [PR 2718].

[PR 2718]: https://github.com/libp2p/rust-libp2p/pull/2718/
```

mxinden (approved):

mxinden:
@bernardoaraujor could you commit after running Rust fmt on your changes?

bernardoaraujor:
@mxinden done.

Issue #2722, opened 2022-06-22
Re-design the `StreamMuxer` trait


thomaseizinger:
This issue tracks the redesign of the `StreamMuxer` trait which was originally started in #2648.

The primary goals are:

- Emitting substreams which implement `AsyncRead` + `AsyncWrite`
- Use `&mut self` for the `StreamMuxer` (This may extend to `Pin<&mut Self>`)

## PRs related to this issue

- [x] https://github.com/libp2p/rust-libp2p/pull/2664
- [x] https://github.com/libp2p/rust-libp2p/pull/2685
- [x] https://github.com/libp2p/rust-libp2p/pull/2669
- [x] https://github.com/libp2p/rust-libp2p/pull/2665
- [x] https://github.com/libp2p/rust-libp2p/pull/2666
- [x] https://github.com/libp2p/rust-libp2p/pull/2667
- [x] https://github.com/libp2p/rust-libp2p/pull/2668
- [x] https://github.com/libp2p/rust-libp2p/pull/2677
- [x] https://github.com/libp2p/rust-yamux/pull/135
- [x] https://github.com/libp2p/rust-libp2p/pull/2688
- [x] https://github.com/libp2p/rust-libp2p/pull/2691
- [x] https://github.com/libp2p/rust-libp2p/pull/2703
- [x] https://github.com/libp2p/rust-libp2p/pull/2706
- [x] https://github.com/libp2p/rust-libp2p/pull/2707
- [x] https://github.com/libp2p/rust-libp2p/pull/2710
- [x] https://github.com/libp2p/rust-libp2p/pull/2723
- [x] https://github.com/libp2p/rust-libp2p/pull/2724
- [x] https://github.com/libp2p/rust-libp2p/pull/2763
- [x] https://github.com/libp2p/rust-libp2p/pull/2764
- [x] https://github.com/libp2p/rust-libp2p/pull/2776
- [x] https://github.com/libp2p/rust-libp2p/pull/2775
- [x] https://github.com/libp2p/rust-libp2p/pull/2765
- [x] https://github.com/libp2p/rust-libp2p/pull/2797
- [ ] Rename `StreamMuxerEvent` to `Event`

elenaf9:
I'd like to discuss again the interface of the `StreamMuxer` trait, in particular the changes done in #2724.
I am sorry for raising this now and not in the PR itself, but I only now started working on adapting the `QuicMuxer` (#2289) to that interface change.


[...]

PR #2724, opened 2022-06-22
core/muxing: Flatten `StreamMuxer` interface to `poll_{inbound,outbound,address_change,close}`

thomaseizinger:
# Description

This is the next changeset on our journey to #2722!

Instead of having a mix of `poll_event`, `poll_outbound` and `poll_close`, we flatten the entire interface of `StreamMuxer` into 4 individual functions:

- `poll_inbound`
- `poll_outbound`
- `poll_address_change`
- `poll_close`

This design is closer to the design of other async traits like `AsyncRead` and `AsyncWrite`. It also allows us to delete the `StreamMuxerEvent`.

Part of #2722.

## Change checklist


thomaseizinger:
Rebased to stay on top of https://github.com/libp2p/rust-libp2p/pull/2710.

mxinden:
Appreciate the elaborate pull request!

Still have to give this more thought.

As you mentioned in https://github.com/libp2p/rust-libp2p/pull/2648, given that we don't have a muxer with backpressure on stream creation today (e.g. QUIC), we will likely not do a good job designing this abstraction at this point in time.

For now, we could get rid of `poll_outbound` for now, returning an ID on `open_outbound`, returning the actual outbound stream through the `poll_event` method.

Again, have to think about it some more.

thomaseizinger:
> For now, we could get rid of `poll_outbound` for now, returning an ID on `open_outbound`, returning the actual outbound stream through the `poll_event` method.

If somehow possible, I'd like to get rid of this ID because it is actually unnecessarily complicated. Without this ID, we can drop abstractions like `OutboundSubstreams` from #2648.

There is nothing special about a substream that requires us to do this explicit book-keeping. All we need is a mechanism for the caller to say: I want a new one.

I started with a simple `poll_outbound: bool` flag on the poll function but that felt like an odd API.

thomaseizinger:
[Clean](https://github.com/libp2p/rust-libp2p/compare/a2c64aec3aa4f8bcfd420bdcd33a3bed122dfcad..3c3bed3c89837d6a1841dd844584c5b0b6e5ac20) rebase onto master, no changes other than dropping commits of previous parent PR.

thomaseizinger:
I just had what I think is a really good idea!

Instead of having a `StreamMuxerEvent`, we can have one poll function per concern. Not only is this more consistent with the existing `poll_close`, this is also how traits like `AsyncRead` and `AsyncWrite` are designed.

core/CHANGELOG.md
```diff
@@ -1,13 +1,14 @@
 # 0.34.0 - unreleased
 
-- Introduce `StreamMuxerEvent::map_inbound_stream`. See [PR 2691].
```
thomaseizinger, line 3: I deleted this changelog entry because it is not released yet and it no longer makes any sense if the entire type is gone now!

mxinden:
> Instead of having a `StreamMuxerEvent`, we can have one poll function per concern. Not only is this more consistent with the existing `poll_close`, this is also how traits like `AsyncRead` and `AsyncWrite` are designed.

This is a cool idea and might even give downstream users of `StreamMuxer` the ability to backpressure incoming streams. Sorry for the delay here. I will take a more in-depth look.

thomaseizinger:
> > Instead of having a `StreamMuxerEvent`, we can have one poll function per concern. Not only is this more consistent with the existing `poll_close`, this is also how traits like `AsyncRead` and `AsyncWrite` are designed.
> 
> This is a cool idea and might even give downstream users of `StreamMuxer` the ability to backpressure incoming streams.

Yep that is what I thought too! Exposing individual knobs has a lot more flexibility.

I do want to try and build a wrapper around this with different trade-offs, i.e. "simpler API, less knobs".

> Sorry for the delay here. I will take a more in-depth look.

No rush. I am away from my laptop for the next week or so anyway :)

muxers/yamux/src/lib.rs
```diff
@@ -97,44 +98,33 @@ where
...
-        OpenSubstreamToken(())
+    fn poll_inbound(&self, cx: &mut Context<'_>) -> Poll<Result<Self::Substream, Self::Error>> {
+        self.0
+            .lock()
+            .incoming
+            .poll_next_unpin(cx)
+            .map(|maybe_stream| {
+                maybe_stream
+                    .transpose()?
+                    .ok_or(YamuxError(ConnectionError::Closed))
```
thomaseizinger: TODO: Use another `?` here.

swarm/src/connection.rs
```diff
@@ -110,8 +120,9 @@ where
             max_negotiating_inbound_streams,
         );
         Connection {
-            muxing: Muxing::new(muxer),
+            muxing: muxer,
             handler: wrapped_handler,
+            open_info: VecDeque::with_capacity(8),
```
thomaseizinger, line 123: This replaces a `SmallVec` from `Muxing` to ensure that we use the `OpenInfo` in a FIFO manner. A capacity of 8 at least makes only one allocation and thus, the (unbenchmarked) performance profile should be similar.

Open to better ideas.

swarm/src/connection.rs
```diff
@@ -140,35 +151,35 @@ where
...
+            if let Poll::Ready(substream) = self.muxing.poll_inbound(cx)? {
+                self.handler
+                    .inject_substream(substream, SubstreamEndpoint::Listener);
+                continue;
+            }
+
+            if let Poll::Ready(address) = self.muxing.poll_address_change(cx)? {
+                self.handler.inject_address_change(&address);
+                return Poll::Ready(Ok(Event::AddressChange(address)));
             }
```
thomaseizinger: I think it makes sense to flip these.

swarm/src/connection.rs
```diff
@@ -140,35 +151,35 @@ where
...
-                Poll::Ready(SubstreamEvent::AddressChange(address)) => {
-                    self.handler.inject_address_change(&address);
-                    return Poll::Ready(Ok(Event::AddressChange(address)));
-                }
+            }
+
+            if let Poll::Ready(substream) = self.muxing.poll_inbound(cx)? {
+                self.handler
+                    .inject_substream(substream, SubstreamEndpoint::Listener);
+                continue;
```
thomaseizinger: TODO: Add docs how this loop works.

thomaseizinger:
> > Instead of having a `StreamMuxerEvent`, we can have one poll function per concern. Not only is this more consistent with the existing `poll_close`, this is also how traits like `AsyncRead` and `AsyncWrite` are designed.
> 
> This is a cool idea and might even give downstream users of `StreamMuxer` the ability to backpressure incoming streams. Sorry for the delay here. I will take a more in-depth look.

Did you get a chance to look at this @mxinden ?

mxinden:
> > > Instead of having a `StreamMuxerEvent`, we can have one poll function per concern. Not only is this more consistent with the existing `poll_close`, this is also how traits like `AsyncRead` and `AsyncWrite` are designed.
> > 
> > 
> > This is a cool idea and might even give downstream users of `StreamMuxer` the ability to backpressure incoming streams. Sorry for the delay here. I will take a more in-depth look.
> 
> Did you get a chance to look at this @mxinden ?

I think we should do this change :+1:  As far as I can tell, the QUIC implementation would be easy to adapt:

https://github.com/libp2p/rust-libp2p/pull/2289/files#diff-d59efbc1f8dc6ed1521dd36e906216f4ff4ade561dca9b50beea1da354240cf7

Can we prevent users (not implementors) forgetting to call one of the `poll_*` methods. Would `#[must_use]` prevent that? If not, I guess given that `libp2p-swarm` is the main user, I think it is fine.

I suggest we wait for #2289 and #2622 to adapt to the latest `StreamMuxer` changes before merging here. That way we can better predict the impact on these implementations. @thomaseizinger what do you think?

//CC @elenaf9 and @kpp for #2289  and //CC @melekes for #2622

thomaseizinger:
> Can we prevent users (not implementors) forgetting to call one of the `poll_*` methods. Would `#[must_use]` prevent that? If not, I guess given that `libp2p-swarm` is the main user, I think it is fine.

I don't think it is possible. We might be able to provide a wrapper struct but I see `libp2p-core` as a set of fundamental building blocks that optimizes for flexibility and thus, reading docs and proper use of the APIs is essential IMO.

thomaseizinger:
> I suggest we wait for #2289 and #2622 to adapt to the latest `StreamMuxer` changes before merging here. That way we can better predict the impact on these implementations. @thomaseizinger what do you think?

Hmm, I'd prefer to merge this early to be honest. Happy to wait for an assessment from @kpp, @elenaf9 and @melekes whether merging first would make things easier.

melekes:
How does the new interface translates when it comes to `OutboundSubstream`

```rust
    type OutboundSubstream = BoxFuture<'static, Result<Arc<DetachedDataChannel>, Self::Error>>;
```

In webrtc transport, I create a future in `open_substream` and poll it in `poll_outbound`. Where am I supposed to create a future now? Note the code I'm using is async.

thomaseizinger:
The new interface provides a `poll_outbound` function which should eventually return a `Substream`.

If your implementation is based on async, you'd probably want to have an `Option<BoxFuture>` which you fill with a new one when it is empty or poll the inner one when it is present.

thomaseizinger:
Currently, this interface still relies on internal mutability but I am planning to make it `Pin<&mut Self>` in a follow-up PR.

mxinden:
> > I suggest we wait for #2289 and #2622 to adapt to the latest `StreamMuxer` changes before merging here. That way we can better predict the impact on these implementations. @thomaseizinger what do you think?
> 
> Hmm, I'd prefer to merge this early to be honest. Happy to wait for an assessment from @kpp, @elenaf9 and @melekes whether merging first would make things easier.

Fine by me as well. Mind resolving the conflicts @thomaseizinger?

thomaseizinger:
Conflicts resolved. Please check if I updated the changelog and manifest versions correctly @mxinden !

thomaseizinger:
@mxinden @elenaf9 This is ready for a review again.

swarm/src/connection.rs
```diff
@@ -138,38 +149,38 @@ where
...
+                        .pop_front()
+                        .expect("`open_info` is not empty");
                     let endpoint = SubstreamEndpoint::Dialer(user_data);
                     self.handler.inject_substream(substream, endpoint);
-                    continue;
-                }
-                Poll::Ready(SubstreamEvent::AddressChange(address)) => {
-                    self.handler.inject_address_change(&address);
-                    return Poll::Ready(Ok(Event::AddressChange(address)));
+                    continue; // Go back to the top, handler can potentially make progress again.
```
mxinden, line 168: Side note, I like this style of:

- When one made progress
- but one does not return
- start from the top via `continue` as other sub components might be able to make progress now.

mxinden (approved):
Thanks for the thorough work! Great to have you here.

swarm/src/connection.rs
thomaseizinger: I do too although it is a pattern I had to get used to and it is not trivial to understand when you get started with async Rust.

PR #2725, opened 2022-06-24
transports/tcp/: Call take_error on tokio TcpStream

mxinden:
# Description

See e04e95a for the rational.

With tokio `v1.19.0` released, `TcStream` exposes `take_error`.

This commit applies the same fix from e04e95a to the tokio TCP provider.

Corresponding merged pull request for the async-std side: https://github.com/libp2p/rust-libp2p/pull/2458/

Tokio pull request: https://github.com/tokio-rs/tokio/pull/4739

## Change checklist

### DIFF
--- a/transports/tcp/src/provider/tokio.rs
+++ b/transports/tcp/src/provider/tokio.rs
@@ -64,8 +64,22 @@ impl Provider for Tcp {
 
     fn new_stream(s: net::TcpStream) -> BoxFuture<'static, io::Result<Self::Stream>> {
         async move {
+            // Taken from [`tokio_crate::net::TcpStream::connect_mio`].
+
             let stream = tokio_crate::net::TcpStream::try_from(s)?;
+
+            // Once we've connected, wait for the stream to be writable as
+            // that's when the actual connection has been initiated. Once we're
+            // writable we check for `take_socket_error` to see if the connect
+            // actually hit an error or not.
+            //
+            // If all that succeeded then we ship everything on up.
             stream.writable().await?;
+
+            if let Some(e) = stream.take_error()? {
+                return Err(e);
+            }
+
             Ok(TcpStream(stream))
         }
         .boxed()
--- a/transports/tcp/CHANGELOG.md
+++ b/transports/tcp/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 - Update to `libp2p-core` `v0.34.0`.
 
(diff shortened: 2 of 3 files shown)


transports/tcp/src/provider/tokio.rs
```diff
@@ -64,8 +64,22 @@ impl Provider for Tcp {
...
+
             let stream = tokio_crate::net::TcpStream::try_from(s)?;
+
+            // Once we've connected, wait for the stream to be writable as
+            // that's when the actual connection has been initiated. Once we're
+            // writable we check for `take_socket_error` to see if the connect
+            // actually hit an error or not.
+            //
+            // If all that succeeded then we ship everything on up.
             stream.writable().await?;
```
dignifiedquire, line 77: looking at `fn writable` this seems to be doing sth different than what `connect_mio` calls under the hood

```rust
poll_fn(|cx| stream.io.registration().poll_write_ready(cx)).await?;
```
vs
```rust
 pub async fn writable(&self) -> io::Result<()> {
    self.ready(Interest::WRITABLE).await?;
    Ok(())
}
```

mxinden: Correct. I was following the recommendation in the last paragraph of the `poll_write_ready` docs:

```

    /// Polls for write readiness.
    ///
    /// If the tcp stream is not currently ready for writing, this method will
    /// store a clone of the `Waker` from the provided `Context`. When the tcp
    /// stream becomes ready for writing, `Waker::wake` will be called on the
    /// waker.
    ///
    /// Note that on multiple calls to `poll_write_ready` or `poll_write`, only
    /// the `Waker` from the `Context` passed to the most recent call is
    /// scheduled to receive a wakeup. (However, `poll_read_ready` retains a
    /// second, independent waker.)
    ///
    /// This function is intended for cases where creating and pinning a future
    /// via [`writable`] is not feasible. Where possible, using [`writable`] is
    /// preferred, as this supports polling from multiple tasks at once.
```

https://github.com/tokio-rs/tokio/blob/55078ffec3bba78803ff646f3933a23834789e4e/tokio/src/net/tcp/stream.rs#L793-L807

Do you see any problems with that @dignifiedquire?

dignifiedquire: I see, that makes sense, seems like using writable is the right call to make here 👍

dignifiedquire (approved):

PR #2726, opened 2022-06-24
protocols/identify: Extend log message on second identify push

mxinden:
# Description

Print remote peer ID when seeing a second identify push stream coming in.

Most of the changes are needed to make `IdentifyHandler` aware of the peer ID of its remote.

## Change checklist

### DIFF
--- a/protocols/identify/src/handler.rs
+++ b/protocols/identify/src/handler.rs
@@ -27,13 +27,40 @@ use futures::prelude::*;
 use futures_timer::Delay;
 use libp2p_core::either::{EitherError, EitherOutput};
 use libp2p_core::upgrade::{EitherUpgrade, InboundUpgrade, OutboundUpgrade, SelectUpgrade};
+use libp2p_core::{ConnectedPoint, PeerId};
 use libp2p_swarm::{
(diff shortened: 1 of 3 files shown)


mxinden:
//CC @BigLep this is the last pull request, making sure log messages on exceeding limits are useful.

protocols/identify/src/handler.rs
```diff
@@ -120,8 +149,9 @@ impl ConnectionHandler for IdentifyHandler {
             EitherOutput::Second(fut) => {
                 if self.inbound_identify_push.replace(fut).is_some() {
                     warn!(
```
thomaseizinger, line 151: Unrelated: We might want to upgrade to `tracing` at some point so we can emit these values in a structured way.

Or we wait until `log` has a structured key-value support.

thomaseizinger (approved):

protocols/identify/src/handler.rs
mxinden: :+1: for structured logging, though not something I will focus on any time soon. Don't have an opinion on crates providing structured logging.

BigLep (commented):
Thanks @mxinden .  Looks good from a messaging regard.

PR #2734, opened 2022-06-28
misc/metrics: Track # connected nodes supporting specific protocol 

mxinden:
# Description

An example metric exposed with this patch:

```
libp2p_identify_protocols{protocol="/ipfs/ping/1.0.0"} 10
```

This implies that 10 of the currently connected nodes support the ping protocol.

To achieve the above, this pull request also:

- protocols/relay: Expose HOP_PROTOCOL_NAME and STOP_PROTOCOL_NAME
- protocols/ping: Expose PROTOCOL_NAME
- protocols/identify: Expose PROTOCOL_NAME and PUSH_PROTOCOL_NAME
- protocols/dcutr: Expose PROTOCOL_NAME
- misc/metrics: Explicitly delegate event recording to each recorder
    
    This allows delegating a single event to multiple `Recorder`s. That enables e.g. the
    `identify::Metrics` `Recorder` to act both on `IdentifyEvent` and `SwarmEvent`. The latter enables
    it to garbage collect per peer data on disconnects.

## Change checklist


misc/metrics/CHANGELOG.md
```diff
@@ -14,6 +14,8 @@
 
 - Update to `libp2p-kad` `v0.38.0`.
 
+- Track number of connected nodes supporting a specific protocol via the identify protocol.
```
mxinden:
```suggestion
- Track number of connected nodes supporting a specific protocol via the identify protocol. See [PR 2734].

[PR 2734]: https://github.com/libp2p/rust-libp2p/pull/2734/
```

protocols/dcutr/CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
 
 - Update to `libp2p-swarm` `v0.37.0`.
 
+- Expose `PROTOCOL_NAME`.
```
mxinden:
```suggestion
- Expose `PROTOCOL_NAME`. See [PR 2734].

[PR 2734]: https://github.com/libp2p/rust-libp2p/pull/2734/
```

protocols/identify/CHANGELOG.md
```diff
@@ -6,6 +6,8 @@
 
 - Extend log message on second identify push stream with peer ID.
 
+- Expose `PROTOCOL_NAME` and `PUSH_PROTOCOL_NAME`.
```
mxinden:
```suggestion
- Expose `PROTOCOL_NAME` and `PUSH_PROTOCOL_NAME`. See [PR 2734].

[PR 2734]: https://github.com/libp2p/rust-libp2p/pull/2734/
```

protocols/ping/CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
 
 - Update to `libp2p-swarm` `v0.37.0`.
 
+- Expose `PROTOCOL_NAME`.
```
mxinden:
```suggestion
- Expose `PROTOCOL_NAME`. See [PR 2734].

[PR 2734]: https://github.com/libp2p/rust-libp2p/pull/2734/
```

protocols/relay/CHANGELOG.md
```diff
@@ -6,6 +6,8 @@
 
 - Do not duplicate the p2p/xxx component with the relay PeerId when a client requests a reservation. See [PR 2701].
 
+- Expose `HOP_PROTOCOL_NAME` and `STOP_PROTOCOL_NAME`.
```
mxinden:
```suggestion
- Expose `HOP_PROTOCOL_NAME` and `STOP_PROTOCOL_NAME`. See [PR 2734].

[PR 2734]: https://github.com/libp2p/rust-libp2p/pull/2734/
```

misc/metrics/src/identify.rs
```diff
@@ -96,27 +110,129 @@ impl Metrics {
...
-            libp2p_identify::IdentifyEvent::Received { info, .. } => {
-                self.identify.received.inc();
-                self.identify
-                    .received_info_protocols
+            libp2p_identify::IdentifyEvent::Received { peer_id, info, .. } => {
+                {
+                    let mut protocols: Vec<String> = info
+                        .protocols
+                        .iter()
+                        .filter(|p| {
```
thomaseizinger, line 129: What is the rationale for the filtering?

thomaseizinger (commented):
Concept ACK but I think we shouldn't filter the supported protocols.

misc/metrics/src/identify.rs
mxinden: A remote node can advertise arbitrary protocol names in its _libp2p-identify_ information. Each of these protocols advertised by the remote will result in a new metric as Prometheus treats each metric _name_ and _label_ combination as a new metric.

A malicious remote node can trigger a [cardinality explosion](https://www.robustperception.io/cardinality-is-key/) through advertising a new set of arbitrary protocols on each _libp2p-identify_ exchange. To prevent this, we filter by known protocols, thus restricting the cardinality of the metric.

Does that make sense @thomaseizinger?

Long term, we could allow the user to whitelist certain custom protocols, though I think we can solve that use-case once it actually comes up.

thomaseizinger: > A remote node can advertise arbitrary protocol names in its _libp2p-identify_ information. Each of these protocols advertised by the remote will result in a new metric as Prometheus treats each metric _name_ and _label_ combination as a new metric.
> 
> A malicious remote node can trigger a [cardinality explosion](https://www.robustperception.io/cardinality-is-key/) through advertising a new set of arbitrary protocols on each _libp2p-identify_ exchange. To prevent this, we filter by known protocols, thus restricting the cardinality of the metric.
> 
> Does that make sense @thomaseizinger?

Is there another way we can mitigate these attacks?

Can we just cap the total number of metrics to something like 1000?

This is a really subtle, non-obvious behaviour that people will only find by digging in the source code which is not ideal IMO.

> Long term, we could allow the user to whitelist certain custom protocols, though I think we can solve that use-case once it actually comes up.

I'd prefer doing this from the get-go. At least with a "loud constructor" that makes people aware of the filtering.

mxinden: > Can we just cap the total number of metrics to something like 1000?

While that would prevent the amount of metrics exposed on each scrape, it does not restrict the amount of unique metrics exposed over time, unless we track the protocols over time, potentially across restarts.

We follow this pattern in `libp2p-gossipsub`. I would consider this a hack.

https://github.com/libp2p/rust-libp2p/blob/d0da3a0973431babba94fffd3ffd7723e0b09837/protocols/gossipsub/src/metrics.rs#L43-L54


> > Long term, we could allow the user to whitelist certain custom protocols, though I think we can solve that use-case once it actually comes up.
> 
> I'd prefer doing this from the get-go. At least with a "loud constructor" that makes people aware of the filtering.

I am fine with a "loud constructor" (good name) though I am not yet sure how exactly that would look like. E.g. would a user pass it to the top level `Metric::new` constructor? What if the `libp2p-identify` feature is not enabled?

How about, instead of a "loud constructor", exposing the number of unrecognized protocols as well @thomaseizinger?

Something along the lines of:

```
libp2p_identify_protocols{protocol = "/ipfs/id/1.0.0"} 10
libp2p_identify_protocols{protocol = "unrecognized"} 10
```

thomaseizinger: > > > Long term, we could allow the user to whitelist certain custom protocols, though I think we can solve that use-case once it actually comes up.
> > 
> > I'd prefer doing this from the get-go. At least with a "loud constructor" that makes people aware of the filtering.
> 
> I am fine with a "loud constructor" (good name) though I am not yet sure how exactly that would look like. E.g. would a user pass it to the top level `Metric::new` constructor? What if the `libp2p-identify` feature is not enabled?

`identify::Metric::new` would need to be renamed to `identify::Metric::with_default_protocol_whitelist` or something similar.

Either now or later, a 2nd constructor can be added that allows users to pass their own whitelist of protocols.

> How about, instead of a "loud constructor", exposing the number of unrecognized protocols as well @thomaseizinger?
> 
> Something along the lines of:
> 
> ```
> libp2p_identify_protocols{protocol = "/ipfs/id/1.0.0"} 10
> libp2p_identify_protocols{protocol = "unrecognized"} 10
> ```

I think that would be a good addition in any case.

mxinden: > `identify::Metric::new` would need to be renamed to `identify::Metric::with_default_protocol_whitelist` or something similar.

But I envision users to never instantiate `identify::Metrics` themselves, but instead the top level `Metrics` object.

E.g. see:

https://github.com/mxinden/rust-libp2p-server/blob/04d3fbcfd8ebbcb594ee0609a41cc4aa0408f8d9/src/main.rs#L106

thomaseizinger: I see. We never seem to abstract over `Recorder` so I am wondering why we need this as a trait and it cannot just be a regular function on `Metrics`?

If it were a regular function, we could make the `record` function loud: `record_with_default_protocol_whitelist`.

mxinden: > I see. We never seem to abstract over `Recorder` so I am wondering why we need this as a trait and it cannot just be a regular function on `Metrics`?

With `Metrics` do you refer to the top level `Metrics` or `identify::Metrics`? On the latter, yes, it could be a method.

> If it were a regular function, we could make the `record` function loud: `record_with_default_protocol_whitelist`.

So users would call `metrics.identify.record_with_default_protocol_whitelist` and provide the list for each event?

thomaseizinger: > > I see. We never seem to abstract over `Recorder` so I am wondering why we need this as a trait and it cannot just be a regular function on `Metrics`?
> 
> With `Metrics` do you refer to the top level `Metrics` or `identify::Metrics`? On the latter, yes, it could be a method.

Ah I guess on the top-level one, we can't because Rust doesn't support method overloading. Do we necessarily need the top-level `record` function call?

If users would generally call through to `metrics.identify` and other "sub" metrics, the "loud" function call for identify would feel less out of place.

> > If it were a regular function, we could make the `record` function loud: `record_with_default_protocol_whitelist`.
> 
> So users would call `metrics.identify.record_with_default_protocol_whitelist` and provide the list for each event?

For the `default_` one, no list needs to be provided. For an alternative one, yes one would have to pass the list (probably as reference).

mxinden: > > How about, instead of a "loud constructor", exposing the number of unrecognized protocols as well @thomaseizinger?
> > Something along the lines of:
> > ```
> > libp2p_identify_protocols{protocol = "/ipfs/id/1.0.0"} 10
> > libp2p_identify_protocols{protocol = "unrecognized"} 10
> > ```
> 
> I think that would be a good addition in any case.

I added the "unrecognized" label value in https://github.com/libp2p/rust-libp2p/pull/2734/commits/70076f428b230e25e3e4e4297a85020999d56bb7 and also adjusted the `# HELP` text to:

```
            "Number of connected nodes supporting a specific protocol, with \
             \"unrecognized\" for each peer supporting one or more unrecognized \
             protocols",
```

mxinden: > > > I see. We never seem to abstract over `Recorder` so I am wondering why we need this as a trait and it cannot just be a regular function on `Metrics`?
> > 
> > 
> > With `Metrics` do you refer to the top level `Metrics` or `identify::Metrics`? On the latter, yes, it could be a method.
> 
> Ah I guess on the top-level one, we can't because Rust doesn't support method overloading. Do we necessarily need the top-level `record` function call?
> 
> If users would generally call through to `metrics.identify` and other "sub" metrics, the "loud" function call for identify would feel less out of place.

In the current setup we **do** need the top level `record` method, as some events need to be passed to multiple `Recorder` implementations. This is the case for `SwarmEvent`s today, passed both to the `Swarm` recorder and the `Identify` recorder. The latter uses the `SwarmEvent`s to do garbage collection when a peer disconnects.

I am not 100% happy with the current ergonomics of the `libp2p-metrics` interface. I don't think having the user use the lower-level `record` methods would be an improvement as the user then needs to ensure to e.g. pass `SwarmEvent` to both the `Swarm` and the `Identify` `Recorder` implementation. Otherwise they would have a memory leak.

thomaseizinger: That's fair.

Let's just leave as is then.

mxinden: :+1: 

It bugs me that I haven't found a better abstraction so far.

misc/metrics/src/identify.rs
```diff
@@ -96,27 +112,136 @@ impl Metrics {
...
+                        .collect();
+
+                    // Signal via an additional label value that one or more
+                    // protocols of the remote peer have not been recognized.
+                    if protocols.len() < info.protocols.len() {
+                        protocols.push("unrecognized".to_string());
+                    }
+
+                    protocols.sort_unstable();
+                    protocols.dedup();
```
thomaseizinger, line 160: This will flatten all `unrecognized` into 1, regardless of how many unrecognized protocols there are.

Would it make sense to increment a separate counter for the number of unrecognized protocols?

thomaseizinger (approved):

misc/metrics/src/identify.rs
mxinden: > Would it make sense to increment a separate counter for the number of unrecognized protocols?

We would then need to track the number of unrecognized protocols per peer. On each new identify message, we would have to subtract the previous amount and add the new amount to the counter. I am not saying this is impossible, just adds a bit of complexity which I am not sure is worth the benefit. In other words, is it relevant for a user to know how many, potentially duplicate, unrecognized protocols all connected peers offer as a sum?

I was hoping for the addition to the `# HELP` text to resolve any confusion:

> with "unrecognized" for each peer supporting one or more unrecognized protocols

thomaseizinger: > > Would it make sense to increment a separate counter for the number of unrecognized protocols?
> 
> We would then need to track the number of unrecognized protocols per peer. On each new identify message, we would have to subtract the previous amount and add the new amount to the counter.

I think this would only apply if we would track it as a gauge?

A counter always increases anyway and you need to use functions like `rate` to see how it changes over time.

A spike in the unrecognized protocols _rate_ could e.g. hint at a spam attack by a peer.

I don't feel strongly about it though, happy to go either way :)

dignifiedquire:
@mxinden it would be really nice to have this be extendable in some form, is there an easy way to do this if I say have two other protocols in my impl I want to track in this same system?

misc/metrics/src/identify.rs
mxinden: > > > Would it make sense to increment a separate counter for the number of unrecognized protocols?
> > 
> > 
> > We would then need to track the number of unrecognized protocols per peer. On each new identify message, we would have to subtract the previous amount and add the new amount to the counter.
> 
> I think this would only apply if we would track it as a gauge?
> 
> A counter always increases anyway and you need to use functions like `rate` to see how it changes over time.

The rate on a counter would only tell how often you saw a unrecognized protocol, not how many unrecognized protocols connected peers offer. One could correlate this with the interval `libp2p-identify` requests a new identification, though that is (a) neither specified nor (b) solid with the push mechanism.

> A spike in the unrecognized protocols _rate_ could e.g. hint at a spam attack by a peer.

True, though I think we would want to prevent such attack before it happens. In addition, given that this attack would potentially bring down our Prometheus instance, who is going to alert us of the spam attack?

mxinden:
> @mxinden it would be really nice to have this be extendable in some form, is there an easy way to do this if I say have two other protocols in my impl I want to track in this same system?

Agreed. Though thus far I have been unable to come up with a good design how to allow users to whitelist a set of protocols. Any ideas @dignifiedquire?

I would argue this pull request already adds value, i.e. _tracking core libp2p protocols only_ is already helpful, thus I will proceed with this pull request.

dignifiedquire:
fully agreed on this adding value already. happy to defer the extension to a followup issue. will do some thinking about an api to extend this

misc/metrics/src/identify.rs
thomaseizinger: > > > > Would it make sense to increment a separate counter for the number of unrecognized protocols?
> > > 
> > > 
> > > We would then need to track the number of unrecognized protocols per peer. On each new identify message, we would have to subtract the previous amount and add the new amount to the counter.
> > 
> > 
> > I think this would only apply if we would track it as a gauge?
> > A counter always increases anyway and you need to use functions like `rate` to see how it changes over time.
> 
> The rate on a counter would only tell how often you saw a unrecognized protocol, not how many unrecognized protocols connected peers offer. One could correlate this with the interval `libp2p-identify` requests a new identification, though that is (a) neither specified nor (b) solid with the push mechanism.

Assuming a stable set of peers, the rate of unrecognised protocols should be stable too? Identify pushes might throw a little spanner in here and there but overall, I'd expect it to be meaningful.

> > A spike in the unrecognized protocols _rate_ could e.g. hint at a spam attack by a peer.
> 
> True, though I think we would want to prevent such attack before it happens. In addition, given that this attack would potentially bring down our Prometheus instance, who is going to alert us of the spam attack?

We are preventing the attack by only increasing a counter instead of creating a label per protocol. Surely prometheus can't be brought down just because a counter increases at a massive rate?

All I am saying is that increasing the counter by the number of unrecognised protocols should allow us to observe the attempt of an attack assuming that a non-malicious node will have a stable number of unrecognised protocols whereas an attack would likely max out the size limit of the incoming message with as many protocols as possible and thus increase the counter by more than ordinary behaviour.

It is not massively useful outside of this I think so happy to go either way.

mxinden: > > > A spike in the unrecognized protocols _rate_ could e.g. hint at a spam attack by a peer.
> > 
> > 
> > True, though I think we would want to prevent such attack before it happens. In addition, given that this attack would potentially bring down our Prometheus instance, who is going to alert us of the spam attack?
> 
> We are preventing the attack by only increasing a counter instead of creating a label per protocol. Surely prometheus can't be brought down just because a counter increases at a massive rate?

Oh error in reasoning on my end here. You are right. Sorry about that @thomaseizinger.

> It is not massively useful outside of this I think so happy to go either way.

I will go without it.

PR #2737, opened 2022-06-28
protocols/identify: fix dev dependencies required for examples/identify

elenaf9:
# Description

Fix the identify dev dependencies required for the example added in #2689.
Without them it currently does not compile. I only noticed it because I tried to run the identify tests from within the `protocols/identify` folder.  Not sure why our CI didn't complain. Maybe it's because we test the workspace instead of each package directly? @mxinden @thomaseizinger.

### DIFF
--- a/protocols/identify/Cargo.toml
+++ b/protocols/identify/Cargo.toml
@@ -25,12 +25,16 @@ thiserror = "1.0"
 void = "1.0"
 
 [dev-dependencies]
-async-std = "1.6.2"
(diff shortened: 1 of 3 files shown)


mxinden (approved):
:pray:

mxinden:
> Not sure why our CI didn't complain. Maybe it's because we test the workspace instead of each package directly?

That is my intuition as well. E.g. the below on the root does not fail:

```
cargo check --examples --workspace --no-default-features
```

Not sure how we could catch this in CI without executing `cargo check` on each individual crate.

Issue #2675, opened 2022-05-27
chat and chat-tokio examples: high CPU consumption (Ubuntu 20.04)


ddeguglielmo:
## Summary

Both chat and chat-tokio examples have very high CPU consumption (close to 100%).

## Expected behaviour

I do not expect the examples to have such an high CPU consumption when no messages are exchanged.

## Actual behaviour
I have two instances of chat (chat-tokio) example running on the same machine. After the publication of the first message by part of the first peer (that is correctly received by the second one), the CPU consumption of the processes increases significantly and reaches 100%.

## Possible Solution
When we tested the same examples on a machine running macOS, we have not observed such an high CPU usage.

## Version
- libp2p version: master branch
- OS: Linux Ubuntu 20.04

## Would you like to work on fixing this bug?
Yes.

mxinden:
Thanks for the report. https://github.com/tokio-rs/console might help here debugging the issue. Help greatly appreciated.

mxinden:
@ddeguglielmo with the latest changes to `libp2p-mdns` does the above issue still persist?

ddeguglielmo:
Hello @mxinden, 
I have just finished testing both chat and chat-tokio examples (I used the current master branch version). The examples work correctly and the CPU consumption of the processes is very low. Thus the issue is not present anymore.

mxinden:
Wonderful. Thanks @ddeguglielmo for testing. And of course @gallegogt and @thomaseizinger for fixing this issue!

Issue #2591, opened 2022-03-25
mdns: don't perform IO on main task


rkuhn:
Currently, the `mdns::Behaviour` polls the `InterfaceState` on the main task, see https://github.com/libp2p/rust-libp2p/blob/d1d358ca437c8c584486df9085f57726608d9ee3/protocols/mdns/src/behaviour.rs#L290-L306.

`InterfaceState` in turn performs IO as part of this: https://github.com/libp2p/rust-libp2p/blob/d1d358ca437c8c584486df9085f57726608d9ee3/protocols/mdns/src/behaviour/iface.rs#L206-L210

We should not be doing any form of IO on the main behaviour task because it reduces latency in the dispatch and handling of events distributed to the various behaviours.

To fix this, we should spawn a separate task per runtime and communicate with it via a channels.

rkuhn:
addendum: `lsof -nP` output showing the relevant FDs
```
actyx   29873 rkuhn   35u  netlink                           0t0 877398060 ROUTE
actyx   29873 rkuhn   36u  a_inode               0,13          0      9494 [eventpoll]
actyx   29873 rkuhn   37u  a_inode               0,13          0      9494 [eventfd]
actyx   29873 rkuhn   38u  a_inode               0,13          0      9494 [timerfd]
actyx   29873 rkuhn   39u     IPv4          877389402        0t0       UDP *:5353 
actyx   29873 rkuhn   40u     IPv4          877389403        0t0       UDP 95.217.194.97:44860 
actyx   29873 rkuhn   41u     IPv4          877389404        0t0       UDP *:5353 
actyx   29873 rkuhn   42u     IPv4          877389405        0t0       UDP 172.17.0.1:45727 
actyx   29873 rkuhn   43u     IPv6          877398061        0t0       TCP *:4001 (LISTEN)
actyx   29873 rkuhn   44u     IPv4          877398062        0t0       TCP *:4001 (LISTEN)
```

rkuhn:
okay, here’s the really weird part: with this patch to `async-io` the poll loop gets broken after some time, settling down into its proper “wait for netlink messages” state
```
diff --git a/src/reactor.rs b/src/reactor.rs
index b6ae153..7e79478 100644
--- a/src/reactor.rs
+++ b/src/reactor.rs
@@ -382,9 +382,11 @@ impl Direction {
     /// Moves all wakers into a `Vec`.
     fn drain_into(&mut self, dst: &mut Vec<Waker>) {
[...]

PR #2738, opened 2022-06-29
[protocols/mdns] Remove Async<UdpSocket> and set the sockets as nonblocking

gallegogt:
# Description

In the previous version of the code the readable and writable functions were running constantly causing high CPU consumption constantly when using tokio as runtime.

* https://github.com/libp2p/rust-libp2p/issues/2591
* https://github.com/libp2p/rust-libp2p/issues/2675

## Change checklist

### DIFF
--- a/protocols/mdns/src/behaviour/iface.rs
+++ b/protocols/mdns/src/behaviour/iface.rs
@@ -24,7 +24,7 @@ mod query;
 use self::dns::{build_query, build_query_response, build_service_discovery_response};
 use self::query::MdnsPacket;
 use crate::MdnsConfig;
-use async_io::{Async, Timer};
+use async_io::Timer;
 use futures::prelude::*;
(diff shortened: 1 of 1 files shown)


mxinden:
Thanks for the report. I was operating under the impression that `async-io` is compatible with tokio. Is that not the case?

I don't think the patch proposed here is a valid solution. E.g. using a plain `std` `UdpSocket` does not give us wake-ups whenever the socket is ready to send or receive. 

See also https://github.com/libp2p/rust-libp2p/issues/2591 //CC @rkuhn.

gallegogt:
Hi, thanks for you feedback, I'm going to close the PR, I still think it's a poor solution, I keep turning the problem around to find a better solution to the excessive CPU consumption that this module has.

I also think that using async-io when Tokyo is active is not a efficient solution since you are using two different reactors

PR #2742, opened 2022-07-01
protocols/rendezvous/src/client: Fix clippy warning let-unit-value

mxinden:
See CI failure e.g. here: https://github.com/libp2p/rust-libp2p/runs/7151073049?check_suite_focus=true

### DIFF
--- a/protocols/rendezvous/src/client.rs
+++ b/protocols/rendezvous/src/client.rs
@@ -310,7 +310,7 @@ fn handle_outbound_event(
             expiring_registrations.extend(registrations.iter().cloned().map(|registration| {
                 async move {
                     // if the timer errors we consider it expired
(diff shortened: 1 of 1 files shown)


protocols/rendezvous/src/client.rs
```diff
@@ -310,7 +310,7 @@ fn handle_outbound_event(
             expiring_registrations.extend(registrations.iter().cloned().map(|registration| {
                 async move {
                     // if the timer errors we consider it expired
-                    let _ = futures_timer::Delay::new(Duration::from_secs(registration.ttl as u64))
+                    futures_timer::Delay::new(Duration::from_secs(registration.ttl as u64))
                         .await;
```
mxinden:
```suggestion
                    futures_timer::Delay::new(Duration::from_secs(registration.ttl as u64)).await;
```

PR #2748, opened 2022-07-07
protocols/mdns: Allow users to choose between async-io and tokio runtime

gallegogt:
**Problem**

When use the Tokio library with mdns module, it is polling every time the UDP socket; causing hight CPU usage.

**Solution**

When using Tokio library, use the UppSocket provided by the Tokio library instead of the Async-io wrapper.

**Breaking Change**

After this fix exist two features ```async-io```, ```tokio``` for this module, for the general module must be use the features ```mdns-async-io``` or ```mdns-tokio``` for async-io or tokio libraries

* https://github.com/libp2p/rust-libp2p/issues/2675
* https://github.com/libp2p/rust-libp2p/issues/2591

## Change checklist


protocols/mdns/CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 - Update to `libp2p-swarm` `v0.38.0`.
 
+- fix(protocols/mdns): fix high cpu usage with tokio library.
```
thomaseizinger: Please follow the existing format of the changelog entries! :)

protocols/mdns/src/behaviour/socket.rs
```diff
@@ -0,0 +1,201 @@
...
+// FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+// DEALINGS IN THE SOFTWARE.
+
+use std::{
+    net::{SocketAddr, UdpSocket},
+    task::{Context, Poll},
+};
+
+/// Interface that must be implemented by the different runtimes to use the UdpSocket in async mode
+///
```
thomaseizinger:
```suggestion
/// Interface that must be implemented by the different runtimes to use the UdpSocket in async mode.
```

I don't think these newlines are idiomatic rustdoc.

protocols/mdns/src/behaviour.rs
```diff
@@ -46,7 +56,11 @@ pub struct Mdns {
     if_watch: IfWatcher,
 
     /// Mdns interface states.
-    iface_states: HashMap<IpAddr, InterfaceState>,
+    #[cfg(feature = "async-io")]
+    iface_states: HashMap<IpAddr, InterfaceState<AsyncUdpSocket, AsyncTimer>>,
+
+    #[cfg(feature = "tokio")]
+    iface_states: HashMap<IpAddr, InterfaceState<TokioUdpSocket, AsyncTimer>>,
```
thomaseizinger: Features should be additive in Rust but this code will break at compile-time if both are enabled.

I think we might need to go with an approach similar as in `libp2p-tcp` here: Create a dedicated `TokioMdns` and `AsyncIoMdns` struct and enable them conditionally.

thomaseizinger (changes requested):
Thanks!

I think the feature itself is worthwhile having but we need to implement it slightly different. Commands like `cargo build --all-features` should continue to work across the workspace.

gallegogt:
Hi @thomaseizinge, thanks for the feedback, I will check it

protocols/mdns/Cargo.toml
```diff
@@ -25,8 +24,26 @@ smallvec = "1.6.1"
 socket2 = { version = "0.4.0", features = ["all"] }
 void = "1.0.2"
 
+async-io-crate  = { package = "async-io", version = "1.3.1", optional = true }
+tokio-crate = { package = "tokio", version = "1.19", default-features = false, features = ["net", "time"], optional = true}
+
+[features]
+default = ["async-io"]
+tokio = ["tokio-crate"]
+async-io = ["async-io-crate"]
```
thomaseizinger: I think it would be good to use the `dep:` syntax here to hide the other two features.

protocols/mdns/src/behaviour.rs
```diff
@@ -35,18 +39,36 @@ use smallvec::SmallVec;
...
+pub type TokioMdns = GenMdns<TokioUdpSocket, TokioTimer>;
+
 /// A `NetworkBehaviour` for mDNS. Automatically discovers peers on the local network and adds
 /// them to the topology.
 #[derive(Debug)]
-pub struct Mdns {
+pub struct GenMdns<S, T>
+where
+    T: TimerBuilder,
+    S: AsyncSocket,
```
thomaseizinger: I think we can remove these and add them only at the impl block where they are needed.

protocols/mdns/src/behaviour/iface.rs
```diff
@@ -34,20 +35,24 @@ use std::{
...
 };
 
 /// An mDNS instance for a networking interface. To discover all peers when having multiple
 /// interfaces an [`InterfaceState`] is required for each interface.
 #[derive(Debug)]
-pub struct InterfaceState {
+pub struct InterfaceState<U, T>
+where
+    U: AsyncSocket,
+    T: TimerBuilder,
```
thomaseizinger: Same here.

protocols/mdns/src/behaviour/socket.rs
```diff
@@ -0,0 +1,190 @@
...
+        Self: Sized;
+
+    /// Attempts to receive a single packet on the socket from the remote address to which it is connected.
+    fn poll_receive_packet(
+        &mut self,
+        _cx: &mut Context,
+        _buf: &mut [u8],
+    ) -> Poll<Option<(usize, SocketAddr)>> {
+        Poll::Pending
+    }
```
thomaseizinger: Not sure about these default implementations? Why have them?

protocols/mdns/src/behaviour/timer.rs
```diff
@@ -0,0 +1,131 @@
...
+    fn at(instant: Instant) -> Self;
+
+    /// Creates a timer that emits events periodically.
+    fn interval(duration: Duration) -> Self;
+
+    /// Creates a timer that emits events periodically, starting at start.
+    fn interval_at(start: Instant, duration: Duration) -> Self;
+
+    /// Poll the timer
+    fn poll_tick(&mut self, cx: &mut Context) -> Poll<Self::Item>;
```
thomaseizinger: Instead of this fn here, what about implementing `Stream`?

This would make the trait naming better as this is the only none-builder function atm.

examples/chat-tokio.rs
```diff
@@ -33,15 +33,19 @@
...
 
 use futures::StreamExt;
 use libp2p::{
     core::upgrade,
     floodsub::{self, Floodsub, FloodsubEvent},
     identity,
-    mdns::{Mdns, MdnsEvent},
+    mdns::{
+        // `MdnsTokio` is available through the `mdns-tokio` feature.
+        TokioMdns,
```
thomaseizinger: Typo.


```suggestion
        // `TokioMdns` is available through the `mdns-tokio` feature.
        TokioMdns, 
```

protocols/mdns/CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 - Update to `libp2p-swarm` `v0.38.0`.
 
+- Fix high CPU usage with Tokio library (see [PR 2748]).
```
thomaseizinger, line 9: Missing link to PR.

protocols/mdns/src/behaviour.rs
```diff
@@ -19,10 +19,14 @@
 // DEALINGS IN THE SOFTWARE.
 
 mod iface;
+mod socket;
+mod timer;
 
 use self::iface::InterfaceState;
+
+use crate::behaviour::{socket::AsyncSocket, timer::TimerBuilder};
+
```
thomaseizinger: Why these newlines?

protocols/mdns/src/behaviour/iface.rs
```diff
@@ -24,8 +24,9 @@ mod query;
 use self::dns::{build_query, build_query_response, build_service_discovery_response};
 use self::query::MdnsPacket;
 use crate::MdnsConfig;
-use async_io::{Async, Timer};
-use futures::prelude::*;
+
+use crate::behaviour::{socket::AsyncSocket, timer::TimerBuilder};
+
```
thomaseizinger: These newlines are unnecessary too.

protocols/mdns/src/behaviour/socket.rs
```diff
@@ -0,0 +1,190 @@
...
+        }
+
+        /// Attempts to send data on the socket to a given address.
+        fn poll_send_packet(
+            &mut self,
+            cx: &mut Context,
+            packet: &[u8],
+            to: SocketAddr,
+        ) -> Poll<()> {
+            if self.poll_writable(cx).is_ready() {
```
thomaseizinger: Nit: Could use the `ready` macro here (and in a few other places).

protocols/mdns/src/behaviour/iface.rs
```diff
@@ -198,43 +205,40 @@ impl InterfaceState {
...
+        while let Some(packet) = self.send_buffer.pop_front() {
+            match self.send_socket.poll_send_packet(
+                cx,
+                &packet,
+                SocketAddr::new(self.multicast_addr, 5353),
+            ) {
+                Poll::Ready(_) => log::trace!("sent packet on iface {}", self.addr),
+                Poll::Pending => s_buffer.push_front(packet),
             }
         }
```
thomaseizinger: Couldn't we already break from the loop on the first `Pending`?

Seems wasteful to empty the entire send buffer.

protocols/mdns/src/behaviour/socket.rs
```diff
@@ -0,0 +1,190 @@
...
+                    Err(err) if err.kind() == std::io::ErrorKind::WouldBlock => {
+                        return Poll::Ready(());
+                    }
+                    Err(err) => {
+                        log::error!("Failed reading datagram: {}", err);
+                        return Poll::Ready(());
+                    }
+                },
+                Poll::Ready(Err(e)) => {
+                    log::error!("Failed recv ready datagram: {}", e);
```
thomaseizinger: These error messages have copy paste errors.

protocols/mdns/src/behaviour/socket.rs
```diff
@@ -0,0 +1,190 @@
...
+                    Ok(_len) => {
+                        log::trace!("sent packet on iface {}", to);
+                        return Poll::Ready(());
+                    }
+                    Err(err) if err.kind() == std::io::ErrorKind::WouldBlock => {
+                        return Poll::Ready(());
+                    }
+                    Err(err) => {
+                        log::error!("Failed reading datagram: {}", err);
+                        return Poll::Ready(());
```
thomaseizinger: I am not sure about returning `Ready` here. Why not propagate the error?

thomaseizinger (changes requested):
Thanks!

I gave this a more in-depth read, please see the inline comments :)

mxinden:
Related issue on `if-watch` used in `libp2p-mdns`: https://github.com/mxinden/if-watch/issues/21

//CC @dignifiedquire

mxinden:
Thanks @gallegogt for the thorough patch and thanks @thomaseizinger for the detailed reviews!

protocols/mdns/src/behaviour/socket.rs
gallegogt: I will remove that default implementations.

protocols/mdns/Cargo.toml
gallegogt: @thomaseizinger  Do you mean use `tokio = ["dep:tokio"]` instead of `tokio = ["tokio-crate"]`?

mxinden:
@gallegogt let us know once this is ready for another review.

gallegogt:
Hi @mxinden it's ready for review

protocols/mdns/CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 - Update to `libp2p-swarm` `v0.38.0`.
 
+- Fix high CPU usage with Tokio library (see [PR 2748](https://github.com/libp2p/rust-libp2p/pull/2748) ).
```
```suggestion
- Fix high CPU usage with Tokio library. See [PR 2748].

[PR 2748]: https://github.com/libp2p/rust-libp2p/pull/2748
```

Typically, we put the link at the bottom.

protocols/mdns/src/behaviour/iface.rs
```diff
@@ -69,7 +68,11 @@ pub struct InterfaceState {
     ttl: Duration,
 }
 
-impl InterfaceState {
+impl<U, T> InterfaceState<U, T>
+where
+    U: AsyncSocket,
+    T: TimerBuilder + futures::Stream,
```
thomaseizinger: I am surprised that we can say `futures::Stream` here without specifying the `Item` type?

protocols/mdns/src/behaviour/socket.rs
```diff
@@ -0,0 +1,149 @@
...
+use std::{
+    io::Error,
+    net::{SocketAddr, UdpSocket},
+    task::{Context, Poll},
+};
+
+/// Interface that must be implemented by the different runtimes to use the UdpSocket in async mode
+pub trait AsyncSocket: Send + 'static {
+    /// Create the async socket from the ```std::net::UdpSocket```
+    fn from_socket(socket: UdpSocket) -> std::io::Result<Self>
```
thomaseizinger:
```suggestion
    fn from_std(socket: UdpSocket) -> std::io::Result<Self>
```

Just an idea, not attached to it.

protocols/mdns/src/behaviour/socket.rs
```diff
@@ -0,0 +1,149 @@
...
+        _buf: &mut [u8],
+    ) -> Poll<Result<Option<(usize, SocketAddr)>, Error>>;
+
+    /// Attempts to send data on the socket to a given address.
+    fn poll_send_packet(
+        &mut self,
+        _cx: &mut Context,
+        _packet: &[u8],
+        _to: SocketAddr,
+    ) -> Poll<Result<(), Error>>;
```
thomaseizinger:
```suggestion
    /// Attempts to receive a single packet on the socket from the remote address to which it is connected.
    fn poll_read(
        &mut self,
        _cx: &mut Context,
        _buf: &mut [u8],
    ) -> Poll<Result<Option<(usize, SocketAddr)>, Error>>;

    /// Attempts to send data on the socket to a given address.
    fn poll_write(
        &mut self,
        _cx: &mut Context,
        _packet: &[u8],
        _to: SocketAddr,
    ) -> Poll<Result<(), Error>>;
```

`poll_read` and `poll_write` are more common I think but at the same time, this might be a bit ambiguous at the usage site.

protocols/mdns/src/behaviour/socket.rs
```diff
@@ -0,0 +1,149 @@
...
+    fn from_socket(socket: UdpSocket) -> std::io::Result<Self>
+    where
+        Self: Sized;
+
+    /// Attempts to receive a single packet on the socket from the remote address to which it is connected.
+    fn poll_receive_packet(
+        &mut self,
+        _cx: &mut Context,
+        _buf: &mut [u8],
+    ) -> Poll<Result<Option<(usize, SocketAddr)>, Error>>;
```
thomaseizinger: Can we add some documentation here in what happens to the buffer in case of `Poll::Pending`? Does the caller have a guarantee that the buffer will only be touched if this returns `Poll::Ready`?

protocols/mdns/src/behaviour/socket.rs
```diff
@@ -0,0 +1,149 @@
...
+        fn poll_receive_packet(
+            &mut self,
+            cx: &mut Context,
+            buf: &mut [u8],
+        ) -> Poll<Result<Option<(usize, SocketAddr)>, Error>> {
+            match self.poll_recv_ready(cx) {
+                Poll::Ready(Ok(_)) => match self.try_recv_from(buf) {
+                    Ok((len, from)) => Poll::Ready(Ok(Some((len, from)))),
+                    Err(err) if err.kind() == std::io::ErrorKind::WouldBlock => {
+                        Poll::Ready(Ok(None))
```
thomaseizinger: The [docs](https://docs.rs/tokio/1.20.0/tokio/net/struct.UdpSocket.html#errors-2) say this can never happen so I think we could `panic!` here? That should allow us to remove the `Option` type from the return value if I am not mistaken.

protocols/mdns/src/behaviour/socket.rs
```diff
@@ -0,0 +1,149 @@
...
+        fn poll_send_packet(
+            &mut self,
+            cx: &mut Context,
+            packet: &[u8],
+            to: SocketAddr,
+        ) -> Poll<Result<(), Error>> {
+            match self.poll_send_ready(cx) {
+                Poll::Ready(Ok(_)) => match self.try_send_to(packet, to) {
+                    Ok(_len) => Poll::Ready(Ok(())),
+                    Err(err) if err.kind() == std::io::ErrorKind::WouldBlock => Poll::Ready(Ok(())),
```
thomaseizinger: The [example](https://docs.rs/tokio/1.20.0/tokio/net/struct.UdpSocket.html#example-6) uses a loop here to avoid the `WouldBlock` case. I think we should do that too.

protocols/mdns/src/behaviour/socket.rs
```diff
@@ -0,0 +1,149 @@
...
+            TokioUdpSocket::from_std(socket)
+        }
+
+        /// Attempts to receive a single packet on the socket from the remote address to which it is connected.
+        fn poll_receive_packet(
+            &mut self,
+            cx: &mut Context,
+            buf: &mut [u8],
+        ) -> Poll<Result<Option<(usize, SocketAddr)>, Error>> {
+            match self.poll_recv_ready(cx) {
```
thomaseizinger: You can use `?` here and also the `ready!` macro if you want.

```suggestion
            match futures::ready!(self.poll_recv_ready(cx)?) {
```

protocols/mdns/src/behaviour/timer.rs
```diff
@@ -0,0 +1,129 @@
...
+use std::{
+    marker::Unpin,
+    pin::Pin,
+    task::{Context, Poll},
+    time::{Duration, Instant},
+};
+
+/// Simple wrapper for the differents type of timers
+#[derive(Debug)]
+pub struct WrapTimer<T> {
```
thomaseizinger: Why not just call this `Timer`?

protocols/mdns/src/behaviour/timer.rs
```diff
@@ -0,0 +1,129 @@
...
+    timer: T,
+}
+
+/// Builder interface to homogenize the differents implementations
+pub trait TimerBuilder: Send + Unpin + 'static {
```
thomaseizinger: Could name this `Builder` and refer to it as `timer::Builder` if we want the additional context.

protocols/mdns/src/behaviour/timer.rs
```diff
@@ -0,0 +1,129 @@
...
+        /// Creates a timer that emits events periodically, starting at start.
+        fn interval_at(start: Instant, duration: Duration) -> Self {
+            WrapTimer {
+                timer: Timer::interval_at(start, duration),
+            }
+        }
+    }
+
+    impl Stream for AsyncTimer {
+        type Item = Instant;
```
thomaseizinger, line 76: Do we ever need the item type? Otherwise we could map it to `()` and thus have the same signature for both modules.

protocols/mdns/src/behaviour/timer.rs
```diff
@@ -0,0 +1,129 @@
...
+        }
+
+        /// Creates a timer that emits events periodically.
+        fn interval(duration: Duration) -> Self {
+            WrapTimer {
+                timer: time::interval(duration),
+            }
+        }
+
+        /// Creates a timer that emits events periodically, starting at start.
```
thomaseizinger: I don't think we need to copy the docs from the trait definition over.

protocols/mdns/tests/use-tokio.rs
```diff
@@ -0,0 +1,137 @@
...
+use libp2p::{
+    identity,
+    mdns::{MdnsConfig, MdnsEvent, TokioMdns},
+    swarm::{Swarm, SwarmEvent},
+    PeerId,
+};
+use std::error::Error;
+use std::time::Duration;
+
+async fn create_swarm(config: MdnsConfig) -> Result<Swarm<TokioMdns>, Box<dyn Error>> {
```
thomaseizinger, line 56: I'd prefer the tests to be at the top and the utility functions to be at the bottom of the file.

src/lib.rs
```diff
@@ -79,8 +79,7 @@ pub use libp2p_identify as identify;
 #[cfg_attr(docsrs, doc(cfg(feature = "kad")))]
 #[doc(inline)]
 pub use libp2p_kad as kad;
-#[cfg(feature = "mdns")]
-#[cfg_attr(docsrs, doc(cfg(feature = "mdns")))]
```
thomaseizinger, line 83: Seems like this line with the `doc(cfg)` got lost!

thomaseizinger (commented):
Thanks!

This looks really good now, I've left a couple more comments :)

protocols/mdns/src/behaviour/socket.rs
gallegogt: @thomaseizinger I'll check it but, but the Tokio socket implementation have a function with this name, and might cause conflict

protocols/mdns/src/behaviour/socket.rs
gallegogt: @thomaseizinger the same with this function, the Tokio socket implementation have a function with this name, and might cause conflict

protocols/mdns/src/behaviour/socket.rs
gallegogt: In Tokio case, from documentation.

[Tokio try_recv_from](https://docs.rs/tokio/1.20.0/tokio/net/struct.UdpSocket.html#method.try_recv_from)

```
// Try to recv data, this may still fail with `WouldBlock`
// if the readiness event is a false positive.
```

Throw panic when is WouldBlock, does not seem to be a good option in this case.

thomaseizinger: Ah you are right, I was mis-reading the code!

What we can then do is put all of this in a loop I think. In case we encounter `WouldBlock`, we can loop up to `poll_recv_ready` which should then return `Pending` and register a Waker. This should allow us to get rid of the `Option` return value!

Thoughts?

protocols/mdns/src/behaviour/socket.rs
thomaseizinger: If we only import the `AsyncSocket` type at the usage site, it should be unambiguous.

protocols/mdns/src/behaviour/socket.rs
thomaseizinger: The other alternative is to also just forward the error to our user. This has likely the advantage that we only have to write the loop once.

I think this is better because even with the loop here, we can't remove the `WouldBlock` case from the signature so any caller may still think it has to handle it.

protocols/mdns/src/behaviour/iface.rs
```diff
@@ -198,43 +199,49 @@ impl InterfaceState {
...
-                .now_or_never()
-            {
-                Some(Ok((len, from))) => {
-                    if let Some(packet) = MdnsPacket::new_from_bytes(&self.recv_buffer[..len], from)
-                    {
-                        self.inject_mdns_packet(packet, params);
-                    }
+        match self
+            .recv_socket
+            .poll_receive_packet(cx, &mut self.recv_buffer)
```
thomaseizinger: I think this should be a loop that reads for as long as it returns `Poll::Ready`.

protocols/mdns/src/behaviour/iface.rs
gallegogt: In this case, the `Stream::Item` is not used, so it is not required to check it at compile time.

protocols/mdns/src/behaviour/timer.rs
gallegogt: I would prefer to leave it as it is, in case the instant is required later

protocols/mdns/src/behaviour/iface.rs
gallegogt: @thomaseizinger  Is the buffer too long to require a loop?

gallegogt:
Hi @mxinden @thomaseizinger  it's ready for review

protocols/mdns/src/behaviour/iface.rs
```diff
@@ -199,37 +199,39 @@ where
...
-        match self
-            .recv_socket
-            .poll_receive_packet(cx, &mut self.recv_buffer)
-        {
-            Poll::Ready(Ok(Some((len, from)))) => {
-                if let Some(packet) = MdnsPacket::new_from_bytes(&self.recv_buffer[..len], from) {
-                    self.inject_mdns_packet(packet, params);
+        match self.recv_socket.poll_read(cx, &mut self.recv_buffer) {
+            Poll::Ready(result) => match result {
+                Ok((len, from)) => {
```
thomaseizinger: I'd prefer to not have nested `match` expressions but instead, nest the patterns as the original implementation did.

protocols/mdns/src/behaviour/iface.rs
```diff
@@ -107,7 +110,9 @@ impl InterfaceState {
                 SocketAddr::new(IpAddr::V6(Ipv6Addr::UNSPECIFIED), 0)
             }
         };
-        let send_socket = Async::new(UdpSocket::bind(bind_addr)?)?;
+        let std_socket = UdpSocket::bind(bind_addr)?;
+        let send_socket = U::from_std(std_socket)?;
```
thomaseizinger: For the sake of minimizing the diff, we could write this similar to how the old code did it:

```suggestion
        let send_socket = U::from_std(UdpSocket::bind(bind_addr)?)?;
```

protocols/mdns/src/behaviour/iface.rs
```diff
@@ -171,17 +175,14 @@ impl InterfaceState {
...
+                        let da = if let Some(new_addr) = address_translation(addr, &observed) {
+                            new_addr.clone()
+                        } else {
+                            addr.clone()
+                        };
 
-                    for addr in addrs {
-                        self.discovered
-                            .push_back((*peer.id(), addr, new_expiration));
+                        self.discovered.push_back((*peer.id(), da, new_expiration));
```
thomaseizinger: This is not functionally equivalent I think. Previously, we've pushed the translated AND untranslated address whereas now we only push one or the other, depending on whether the translation worked. What is the reason for the behaviour change?

protocols/mdns/src/behaviour/iface.rs
```diff
@@ -198,43 +199,51 @@ impl InterfaceState {
...
     ) -> Option<(PeerId, Multiaddr, Instant)> {
         // Poll receive socket.
-        while self.recv_socket.poll_readable(cx).is_ready() {
-            match self
-                .recv_socket
-                .recv_from(&mut self.recv_buffer)
-                .now_or_never()
-            {
-                Some(Ok((len, from))) => {
+        match self.recv_socket.poll_read(cx, &mut self.recv_buffer) {
```
thomaseizinger: This should be a loop similar as to before.

Previously, we read as many packets as possible until we hit `Poll::Pending`. Now we are reading at most one packet and then moving on to sending responses. We should exhaust the `recv_socket` until it returns `Pending`.

protocols/mdns/src/behaviour/iface.rs
```diff
@@ -198,43 +199,51 @@ impl InterfaceState {
...
+            match self.send_socket.poll_write(
+                cx,
+                &packet,
+                SocketAddr::new(self.multicast_addr, 5353),
+            ) {
+                Poll::Ready(data) => match data {
+                    Ok(_) => log::trace!("sent packet on iface {}", self.addr),
+                    Err(err) => {
+                        log::error!("error sending packet on iface {} {}", self.addr, err);
+                        self.send_buffer.push_front(packet);
```
thomaseizinger: This is not functionally equivalent to what we did before.

Previously, we discarded a packet that we failed to send AND moved on to the next one if the socket was still ready to accept packets.

Now we are pushing the packet back into the buffer and breaking from the loop.

thomaseizinger (changes requested):
Thank you for incorporating the feedback :)

I think this is close to being mergeable. I made a few comments where we don't seem to be doing quite the same as previously. Please comment if that is on purpose, otherwise I'd prefer to stick to the previous logic.

protocols/mdns/src/behaviour/iface.rs
gallegogt: Sorry, in this case reduce the 2 loops to 1:

```
for addr in peer.addresses() {                                                      
     if let Some(new_addr) = address_translation(addr, &observed) {                  
        self.discovered.push_back((*peer.id(), new_addr.clone(), new_expiration));  
     }                                                                               
                                                                                           
     self.discovered.push_back((*peer.id(), addr.clone(), new_expiration));                  
 } 
```

thomaseizinger: It is fine to reduce the two loops to one but we need to keep the same functionality :)

The code snipped you posted seems good!

examples/chat-tokio.rs
```diff
@@ -33,15 +33,19 @@
 //! ```sh
 //!cargo run --example chat-tokio \\
 //!    --no-default-features \\
-//!    --features="floodsub mplex noise tcp-tokio mdns"
+//!    --features="floodsub mplex noise tcp-tokio mdns-tokio"
```
mxinden, line 36: :pray:

protocols/mdns/CHANGELOG.md
```diff
@@ -2,8 +2,12 @@
 
 - Update to `libp2p-swarm` `v0.38.0`.
 
+- Fix high CPU usage with Tokio library (see [PR 2748]).
```
mxinden: Can you document the added feature flags here? What should an `async-std` and `tokio` user do when upgrading to the new version?

protocols/mdns/src/behaviour/iface.rs
```diff
@@ -198,43 +201,52 @@ impl InterfaceState {
...
+                Poll::Ready(Ok((len, from))) => {
                     if let Some(packet) = MdnsPacket::new_from_bytes(&self.recv_buffer[..len], from)
                     {
                         self.inject_mdns_packet(packet, params);
                     }
                 }
-                Some(Err(err)) => log::error!("Failed reading datagram: {}", err),
-                None => {}
+                Poll::Ready(Err(err)) if err.kind() == std::io::ErrorKind::WouldBlock => {
+                    // Not more bytes available on the socket to read
```
mxinden:
```suggestion
                    // No more bytes available on the socket to read.
```

Not a native speaker. Might be wrong.

protocols/mdns/src/behaviour/socket.rs
```diff
@@ -0,0 +1,129 @@
...
+// FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+// DEALINGS IN THE SOFTWARE.
+
+use std::{
+    io::Error,
+    net::{SocketAddr, UdpSocket},
+    task::{Context, Poll},
+};
+
+/// Interface that must be implemented by the different runtimes to use the UdpSocket in async mode
```
mxinden:
```suggestion
/// Interface that must be implemented by the different runtimes to use the [`UdpSocket`] in async mode
```

protocols/mdns/src/behaviour/socket.rs
```diff
@@ -0,0 +1,129 @@
...
+pub trait AsyncSocket: Send + 'static {
+    /// Create the async socket from the ```std::net::UdpSocket```
```
mxinden:
```suggestion
    /// Create the async socket from the [`std::net::UdpSocket`].
```

mxinden (commented):
This looks good to me. Deferring final review to @thomaseizinger.

Thanks @gallegogt and @thomaseizinger for the solid work here!

protocols/mdns/CHANGELOG.md
```diff
@@ -2,8 +2,14 @@
 
 - Update to `libp2p-swarm` `v0.38.0`.
 
+- Allow users to choose between async-io and tokio runtime
+  in the mdns protocol implementation. `async-io` is a default
+  feature, with an additional `tokyo` feature  (see [PR 2748])
```
```suggestion
  feature, with an additional `tokio` feature  (see [PR 2748])
```

Small typo :)

protocols/mdns/src/behaviour/iface.rs
```diff
@@ -198,43 +201,52 @@ impl InterfaceState {
...
         // Poll receive socket.
-        while self.recv_socket.poll_readable(cx).is_ready() {
-            match self
-                .recv_socket
-                .recv_from(&mut self.recv_buffer)
-                .now_or_never()
-            {
-                Some(Ok((len, from))) => {
+        loop {
+            match self.recv_socket.poll_read(cx, &mut self.recv_buffer) {
```
thomaseizinger: Feel free to leave this as is but this could be written as a `while let` which would avoid the `Poll::Pending => break` clause.

```rust
while let Poll::Ready(result) = self.recv_socket.poll_read(cx, &mut self.recv_buffer) {
    match result {
        // ..
    }
}
```

protocols/mdns/src/behaviour/iface.rs
```diff
@@ -198,43 +201,52 @@ impl InterfaceState {
...
-                Some(Err(err)) => log::error!("Failed reading datagram: {}", err),
-                None => {}
+                Poll::Ready(Err(err)) if err.kind() == std::io::ErrorKind::WouldBlock => {
+                    // No more bytes available on the socket to read
+                    break;
+                }
+                Poll::Ready(Err(err)) => {
+                    log::error!("failed reading datagram: {}", err);
+                    break;
+                }
```
thomaseizinger: In the previous implementation, we did not `break` on an error. Please keep it that way.

protocols/mdns/src/behaviour/iface.rs
```diff
@@ -198,43 +201,52 @@ impl InterfaceState {
...
-            match self
-                .recv_socket
-                .recv_from(&mut self.recv_buffer)
-                .now_or_never()
-            {
-                Some(Ok((len, from))) => {
+        loop {
+            match self.recv_socket.poll_read(cx, &mut self.recv_buffer) {
+                Poll::Ready(Ok((len, from))) => {
                     if let Some(packet) = MdnsPacket::new_from_bytes(&self.recv_buffer[..len], from)
```
thomaseizinger, line 209: An idea for a follow-up: We could create a wrapper that utilizes `AsyncSocket` to directly read and write `MdnsPacket`s, i.e. have a layer between this code here and the actual socket.

protocols/mdns/src/behaviour/iface.rs
```diff
@@ -198,43 +201,52 @@ impl InterfaceState {
...
+            ) {
+                Poll::Ready(Ok(_)) => log::trace!("sent packet on iface {}", self.addr),
+                Poll::Ready(Err(err)) => {
+                    log::error!("error sending packet on iface {} {}", self.addr, err);
+                    break;
+                }
+                Poll::Pending => {
+                    self.send_buffer.push_front(packet);
+                    break;
                 }
```
thomaseizinger: Here too, we are `break`ing but previously didn't.

thomaseizinger (changes requested):
Thanks!

I've found too more misalignments with the previous version. Thanks for sticking to this :)

thomaseizinger (approved):
Thanks for implementing all the feedback!

This is ready now :)

thomaseizinger:
The `test_expired_tokio` test is failing. Can you look into that @gallegogt?

gallegogt:
> The `test_expired_tokio` test is failing. Can you look into that @gallegogt?

I'll check it, but in my test environments work, I will try in other PC.

thomaseizinger:
> > The `test_expired_tokio` test is failing. Can you look into that @gallegogt?
> 
> I'll check it, but in my test environments work, I will try in other PC.

It also fails on my local computer. I am running Manjaro Linux.

gallegogt:
> > > The `test_expired_tokio` test is failing. Can you look into that @gallegogt?
> 
> > 
> 
> > I'll check it, but in my test environments work, I will try in other PC.
> 
> 
> 
> It also fails on my local computer. I am running Manjaro Linux.

@thomaseizinger , @mxinden 
My environment:

```
ARM

Description:    Ubuntu 18.04.6 LTS
Release:        18.04
Codename:       bionic
```

rustc 1.62.0 (a8314ef7d 2022-06-27)

image_2022-07-21_105521367

mxinden:
Failing on my machine as well:

```
➜  cargo test --all-features
    Finished test [unoptimized + debuginfo] target(s) in 0.17s
     Running unittests src/lib.rs (/home/mxinden/code/github.com/libp2p/rust-libp2p/target/debug/deps/libp2p_mdns-aeea3dee3d0f3ca4)

running 5 tests
test behaviour::iface::dns::tests::build_service_discovery_response_correct ... ok
test behaviour::iface::dns::tests::build_query_correct ... ok
test behaviour::iface::dns::tests::test_random_string ... ok
test behaviour::iface::query::tests::test_create_mdns_peer ... ok
test behaviour::iface::dns::tests::build_query_response_correct ... ok

test result: ok. 5 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

     Running tests/use-async-std.rs (/home/mxinden/code/github.com/libp2p/rust-libp2p/target/debug/deps/use_async_std-77e86062dfdf653f)

running 3 tests
test test_discovery_async_std_ipv4 ... ok
test test_discovery_async_std_ipv6 ... ok
test test_expired_async_std ... ok

test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 1.03s

     Running tests/use-tokio.rs (/home/mxinden/code/github.com/libp2p/rust-libp2p/target/debug/deps/use_tokio-278cda36b01406d1)

running 3 tests
test test_expired_tokio ... FAILED
test test_discovery_tokio_ipv4 ... ok
test test_discovery_tokio_ipv6 ... ok

failures:

---- test_expired_tokio stdout ----
thread 'test_expired_tokio' panicked at 'called `Result::unwrap()` on an `Err` value: Elapsed(())', protocols/mdns/tests/use-tokio.rs:55:10


failures:
    test_expired_tokio

test result: FAILED. 2 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 9.15s

error: test failed, to rerun pass '--test use-tokio'
```

gallegogt:
> Failing on my machine as well:
> 
> 
> 
> ```
> 
> ➜  cargo test --all-features
> 
>     Finished test [unoptimized + debuginfo] target(s) in 0.17s
> 
>      Running unittests src/lib.rs (/home/mxinden/code/github.com/libp2p/rust-libp2p/target/debug/deps/libp2p_mdns-aeea3dee3d0f3ca4)
> 
> 
> 
> running 5 tests
> 
> test behaviour::iface::dns::tests::build_service_discovery_response_correct ... ok
> 
> test behaviour::iface::dns::tests::build_query_correct ... ok
> 
> test behaviour::iface::dns::tests::test_random_string ... ok
> 
> test behaviour::iface::query::tests::test_create_mdns_peer ... ok
> 
> test behaviour::iface::dns::tests::build_query_response_correct ... ok
> 
> 
> 
> test result: ok. 5 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
> 
> 
> 
>      Running tests/use-async-std.rs (/home/mxinden/code/github.com/libp2p/rust-libp2p/target/debug/deps/use_async_std-77e86062dfdf653f)
> 
> 
> 
> running 3 tests
> 
> test test_discovery_async_std_ipv4 ... ok
> 
> test test_discovery_async_std_ipv6 ... ok
> 
> test test_expired_async_std ... ok
> 
> 
> 
> test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 1.03s
> 
> 
> 
>      Running tests/use-tokio.rs (/home/mxinden/code/github.com/libp2p/rust-libp2p/target/debug/deps/use_tokio-278cda36b01406d1)
> 
> 
> 
> running 3 tests
> 
> test test_expired_tokio ... FAILED
> 
> test test_discovery_tokio_ipv4 ... ok
> 
> test test_discovery_tokio_ipv6 ... ok
> 
> 
> 
> failures:
> 
> 
> 
> ---- test_expired_tokio stdout ----
> 
> thread 'test_expired_tokio' panicked at 'called `Result::unwrap()` on an `Err` value: Elapsed(())', protocols/mdns/tests/use-tokio.rs:55:10
> 
> note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
> 
> 
> 
> 
> 
> failures:
> 
>     test_expired_tokio
> 
> 
> 
> test result: FAILED. 2 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 9.15s
> 
> 
> 
> error: test failed, to rerun pass '--test use-tokio'
> 
> ```

Very rare situation, I guess tokio (explicitly non_blocking) takes longer to find the peers than AsynIO. If so, I increase the expiration time and query request a little

mxinden:
> Very rare situation, I guess tokio (explicitly non_blocking) takes longer to find the peers than AsynIO. If so, I increase the expiration time and query request a little

That would be surprising to me, given that the timeout is on the order of seconds, still something worth trying @gallegogt.

gallegogt:
Hi @mxinden and @thomaseizinger can you check if the error persist please ?

thomaseizinger:
> Hi @mxinden and @thomaseizinger can you check if the error persist please ?

Yep, CI is still failing.

thomaseizinger:
I can't reproduce it locally anymore :/

gallegogt:
The weird thing is that if I run `cargo test --workspace --all-features -j1` it doesn't give me the error. :/

thomaseizinger:
> The weird thing is that if I run `cargo test --workspace --all-features -j1` it doesn't give me the error. :/

I think GitHub actions VMs have at least two cores.

Edit: Yep, two cores: https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources

gallegogt:
Hi @thomaseizinger and @mxinden after some research I found  that the tokio implementation take some time (about 10s ) to find the peers.

I refactored the expiration test function, where the timeout check is from when the peers are found

Atte.

thomaseizinger:
> Hi @thomaseizinger and @mxinden after some research I found that the tokio implementation take some time (about 10s ) to find the peers.
> 
> I refactored the expiration test function, where the timeout check is from when the peers are found
> 
> Atte.

Nice! Thanks for digging into this. I've kicked off another test run, lets see what CI thinks :)

gallegogt:
It fails again, :(, I'll try again with a 2 cpu in docker.

We keep in touch, thanks

gallegogt:
hi @thomaseizinger, I found out that **tokio::interval** works different than **AsyncIO::Timer::interval**, the first one fires the tick now and then during the period and the second one fires after the duration, passed as a parameter.

I checked in my fork actions that the lastest changes works

thomaseizinger:
> hi @thomaseizinger, I found out that **tokio::interval** works different than **AsyncIO::Timer::interval**, the first one fires the tick now and then during the period and the second one fires after the duration, passed as a parameter.
> 
> I checked in my fork actions that the lastest changes works

That is quite interesting, thanks for working this out! I kicked CI again :)

I'll give it a review too but currently travelling so please bear with me :D

.github/workflows/testing.yml
```diff
@@ -0,0 +1,134 @@
+name: Pre testing
```
thomaseizinger: Are you intending to merge this workflow as well or is this a leftover from testing?

thomaseizinger (approved):
Thanks a lot for this contribution! Great to see better support for multiple runtimes!

One last question regarding the workflow, otherwise this is good to be merged from my perspective!

.github/workflows/testing.yml
gallegogt: sorry it's a remnant of the tests, I'll delete it

gallegogt:
@thomaseizinger done with last comments.

thomaseizinger:
> @thomaseizinger done with last comments.

Thanks!

Just one note for future PRs, we tend to not force-push on this repository because PRs will be squash-merged in the end so regular merge commits of e.g. `master` are completely fine :)

thomaseizinger:
I've kicked CI again and will merge once it passes!

mxinden (approved):
:pray: Thanks @gallegogt and @thomaseizinger!

tomaka:
> When use the Tokio library with mdns module, it is polling every time the UDP socket; causing hight CPU usage.

I've just discovered this PR. Did we ever find out what causes this? If not, it seems to me that this PR was done to bypass the issue, not fix it.

rkuhn:
See #2591 for my investigation into this: the short story is that `epoll` immediately returns.

PR #2760, opened 2022-07-19
transports/tcp: Bump to v0.35.0

mxinden:
Follow up on https://github.com/libp2p/rust-libp2p/pull/2724/. Given that
libp2p-core is bumped to v0.35.0, libp2p-tcp needs to be bumped as well.


thomaseizinger (approved):
Whoops. That is what you get from doing it manually :)

It would be nice to have a bot for this.

mxinden:
Happy for improvement suggestions on the release process. Thus far I haven't been annoyed enough in order to change something about it myself.

PR #2763, opened 2022-07-20
core/tests: Remove unnecessary `Arc`

thomaseizinger:
# Description

## Change checklist

- [x] I have performed a self-review of my own code
~- [ ] I have made corresponding changes to the documentation~
~- [ ] I have added tests that prove my fix is effective or that my feature works~
~- [ ] A changelog entry has been made in the appropriate crates~

PR #2765, opened 2022-07-20
core/muxing: Have functions on `StreamMuxer` take `Pin<&mut Self>`

thomaseizinger:
# Description

This is the idiomatic way of doing `poll` interface in Rust.

https://github.com/libp2p/rust-libp2p/issues/2722
Depends on https://github.com/libp2p/rust-libp2p/pull/2763.
Depends on https://github.com/libp2p/rust-libp2p/pull/2764.
Depends on https://github.com/libp2p/rust-libp2p/pull/2775.
Depends on https://github.com/libp2p/rust-libp2p/pull/2776.

## Change checklist


core/CHANGELOG.md
```diff
@@ -2,8 +2,10 @@
 
 - Remove `StreamMuxer::poll_event` in favor of individual functions: `poll_inbound`, `poll_outbound`
   and `poll_address_change`. Consequently, `StreamMuxerEvent` is also removed. See [PR 2724].
+- Use `Pin<&mut Self>` as the receiver type for all `StreamMuxer` poll functions. See [PR XXXX].
 
 [PR 2724]: https://github.com/libp2p/rust-libp2p/pull/2724
+[PR XXXX]: https://github.com/libp2p/rust-libp2p/pull/XXXX
```
elenaf9:
```suggestion
- Use `Pin<&mut Self>` as the receiver type for all `StreamMuxer` poll functions. See [PR 2765].

[PR 2724]: https://github.com/libp2p/rust-libp2p/pull/2724
[PR 2765]: https://github.com/libp2p/rust-libp2p/pull/XXXX
```

core/src/muxing.rs
```diff
@@ -93,5 +104,132 @@ pub trait StreamMuxer {
     /// >           that the remote is properly informed of the shutdown. However, apart from
     /// >           properly informing the remote, there is no difference between this and
     /// >           immediately dropping the muxer.
-    fn poll_close(&self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>>;
+    fn poll_close(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>>;
+}
+
+/// Extension trait for [`StreamMuxer`].
+pub trait StreamMuxerExt: StreamMuxer + Sized {
```
elenaf9, line 111: Nice!

swarm/src/connection.rs
```diff
@@ -131,10 +131,10 @@ where
 
     /// Begins an orderly shutdown of the connection, returning the connection
     /// handler and a `Future` that resolves when connection shutdown is complete.
-    pub fn close(self) -> (THandler, impl Future<Output = io::Result<()>>) {
+    pub fn close(mut self) -> (THandler, impl Future<Output = io::Result<()>>) {
         (
             self.handler.into_connection_handler(),
-            poll_fn(move |cx| self.muxing.poll_close(cx)),
+            poll_fn(move |cx| self.muxing.poll_close_unpin(cx)),
```
elenaf9:
```suggestion
            self.muxing.close(),
```
Can we use the `StreamExt::close` here?

thomaseizinger: I tried but unfortunately, we are moving `self` here and `close` doesn't move `self`. It could be made to move it but I am not sure if that isn't too restrictive?

I guess there isn't really a use for accessing the muxer after closing so we can move it into the future.

core/src/muxing.rs
thomaseizinger: Glad you like it! I wasn't sure if it isn't over-engineered :)

swarm/src/connection.rs
thomaseizinger: I've changed `StreamMuxerExt::close` to consume `self` which allows us to use it here.

core/src/muxing.rs
mxinden, line 111: Fancy

core/src/muxing.rs
```diff
@@ -93,5 +104,132 @@ pub trait StreamMuxer {
...
+    fn next_outbound(&mut self) -> NextOutbound<'_, Self>;
+
+    fn close(self) -> Close<Self>;
+}
+
+impl<S> StreamMuxerExt for S
+where
+    S: StreamMuxer,
+{
+    fn poll_inbound_unpin(
```
mxinden, line 113: Why not provide this as a default implementation on the trait definition itself?

mxinden (commented):
:pray: thanks for the patch using `Pin`.

core/src/muxing.rs
thomaseizinger: I guess we could because we own the trait?

Default impls can be overridden and these are not really meant too. I think that is the only argument for a separate trait.

mxinden: No strong opinion here. One datapoint is that `StreamExt` implements the methods directly:

https://docs.rs/futures/latest/futures/prelude/stream/trait.StreamExt.html

thomaseizinger: Oh, now I understand what you mean! I thought we are discussing having the extension trait vs putting the functions as defaults on `StreamMuxer`.

With the extension trait, I think I prefer default functions too because it co-locates things and should also result in less LoC.

mxinden: Great. To make sure we are on the same page, with the above in mind, this `trait` `impl` would go away @thomaseizinger?

thomaseizinger: The `impl` block would be there but it will be a one-liner as all the implementations are in the trait definition.

One alternative idea would be to remove `StreamMuxerExt` entirely and add the default functions to `StreamMuxer` directly.

I think I mildly prefer the extension trait though because:

- It makes it impossible for implementations to overwrite them, thus guaranteeing that they are the same for all.
- It makes `StreamMuxer` a more coherent trait and we could draw a dependency boundary between `StreamMuxer` and its extension trait if we wanted to.

thomaseizinger:
@elenaf9 @mxinden This is ready for another review.

I believe this is useful despite the on-going discussion on the overall design because we will be using Pin anyway for `poll` function.

core/src/muxing.rs
```diff
@@ -93,5 +104,105 @@ pub trait StreamMuxer {
...
+
+pub struct NextInbound<'a, S>(&'a mut S);
+
+pub struct NextOutbound<'a, S>(&'a mut S);
+
+pub struct Close<S>(S);
+
+impl<'a, S> Future for NextInbound<'a, S>
+where
+    S: StreamMuxer + Unpin,
```
thomaseizinger, line 179: I am not sure about this `Unpin` bound and whether we can make it go away or whether it doesn't actually hurt at all.

mxinden: Played around with this as well. I am not sure how one would do this without the `Unpin` bound.

I think this is similar to `StreamExt::next` which as well requires the inner object to be `Unpin`:

> Note that because next doesn’t take ownership over the stream, the [Stream](https://docs.rs/futures/latest/futures/stream/trait.Stream.html) type must be [Unpin](https://doc.rust-lang.org/nightly/core/marker/trait.Unpin.html). If you want to use next with a [!Unpin](https://doc.rust-lang.org/nightly/core/marker/trait.Unpin.html) stream, you’ll first have to pin the stream. This can be done by boxing the stream using [Box::pin](https://doc.rust-lang.org/nightly/alloc/boxed/struct.Box.html#method.pin) or pinning it to the stack using the pin_mut! macro from the pin_utils crate.

https://docs.rs/futures/latest/futures/prelude/stream/trait.StreamExt.html#method.next

muxers/yamux/src/lib.rs
```diff
@@ -39,21 +38,19 @@ use thiserror::Error;
...
+    control: yamux::Control,
+}
 
 impl<S> fmt::Debug for Yamux<S> {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         f.write_str("Yamux")
     }
 }
 
-struct Inner<S> {
```
mxinden, line 50: :tada:

mxinden (approved):
Wonderful to see this happening! :pray:

mxinden:
@elenaf9 any objections to me merging here with the discussion on https://github.com/libp2p/rust-libp2p/issues/2722#issuecomment-1200447624 in mind?

I think we want to take `Pin<&mut Self>` no matter whether we go with 3 `poll_` methods (`poll_inbound`, `poll_outbound`, `poll_close`) or one (`poll`).

muxers/yamux/src/lib.rs
mxinden: To expand on this: I am very happy for the `Mutex` around `Inner` to go away. Thanks @thomaseizinger!

melekes (approved):
👍

muxers/yamux/src/lib.rs
thomaseizinger: Me too 😊

core/src/muxing.rs
thomaseizinger: Cool, thanks for confirming. I'll leave it as is then!

muxers/mplex/src/lib.rs
```diff
@@ -85,25 +85,34 @@ where
     type Substream = Substream<C>;
     type Error = io::Error;
 
-    fn poll_inbound(&self, cx: &mut Context<'_>) -> Poll<Result<Self::Substream, Self::Error>> {
+    fn poll_inbound(
+        self: Pin<&mut Self>,
+        cx: &mut Context<'_>,
+    ) -> Poll<Result<Self::Substream, Self::Error>> {
         self.io
             .lock()
```
elenaf9, line 93: Couldn't we now get rid of the `Mutex` in `self.io`?

thomaseizinger: Unfortunately no because we need to share the struct between the substreams and the `StreamMuxer` implementation.

elenaf9 (approved):

PR #2766, opened 2022-07-21
README.md: Add crates.io and docs.rs badges

LesnyRumcajs:
# Description
I find those useful when browsing dependencies, I guess adding them won't hurt. :smile:


thomaseizinger (approved):
This is fine yeah, thank you!

PR #2762, opened 2022-07-20
core/muxing: Drop `Unpin` requirement from `SubstreamBox`

thomaseizinger:
# Description

We are already boxing the type, might as well pin it.

## Links to any relevant issues

https://github.com/libp2p/rust-libp2p/issues/2722.

## Change checklist

- [x] I have performed a self-review of my own code
- [x] I have made corresponding changes to the documentation
~- [ ] I have added tests that prove my fix is effective or that my feature works~
- [x] A changelog entry has been made in the appropriate crates

PR #2769, opened 2022-07-23
core: fix PR number in changelog entry

elenaf9:
Addendum to #2762.
Replace placeholder in changelog entry with actual PR number


thomaseizinger (approved):
Thanks :)

PR #2770, opened 2022-07-24
core/identity/ecdsa: fix clippy::op-ref, clippy::needless-borrow

elenaf9:
# Description

Fix clippy::op-ref, clippy::needless-borrow, which cause an error if you run `cargo custom-clippy` directly in the `/core` package or depend on it. See https://github.com/kpp/rust-libp2p/runs/7489162901?check_suite_focus=true.

Does this need a changelog entry?


thomaseizinger (approved):
No changelog entry needed IMO :)

But curious why this isn't flagged by CI?

elenaf9:
> But curious why this isn't flagged by CI?

I was also confused. The same happened in #2737. Took me actually a while to figure it out, though in hindsight the answer is rather obvious:
Our `custom-clippy` just runs clippy in the root package with default features, and in our root Cargo.toml we disable default features for dependencies like libp2p-core. Since ecdsa is behind a features flag it was not checked. 
I think it would make more sense to check all packages with all features. Opened #2771.

mxinden (approved):
Thanks for the fix and thanks for looking into why this wasn't caught before.

Also don't think this needs a changelog.

PR #2771, opened 2022-07-24
.cargo: check all features in custom-clippy

elenaf9:
# Description

Our `custom-clippy` currently just runs clippy with the default features. The result of this is that a) not all packages are checked, and b) all features in those packages are disabled. This caused issues like #2770 and #2737.
Extended it now to check all features.

Depends on:
- [x] #2770 
- [x] #2772 
- [x] #2773 
- [x] #2774


thomaseizinger (approved):

elenaf9:
As an aside: while reading through the cargo book i stumbled across the following note on rust workspaces: https://doc.rust-lang.org/cargo/reference/workspaces.html#the-workspace-section
> All [path dependencies](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-path-dependencies) residing in the workspace directory automatically become members.

Apart from some of the packages in `/misc`,  I think all of our `[workspace]` `members` that we currently explicitly list are already path dependencies. I don't really have a strong opinion in this, but theoretically we could remove them from the `members` list, which makes it one place less to worry about when adding a new package. Wdyt @thomaseizinger @mxinden? Are there any other packages in our workspace that are not path dependencies?

thomaseizinger:
Interesting, I didn't know that. Happy to remove them from the members list then :)

mxinden (approved):
:pray: Thanks for investigating.

mxinden:
@elenaf9 would you mind addressing the remaining clippy failures?

elenaf9:
> @elenaf9 would you mind addressing the remaining clippy failures?

@mxinden  #2772 🙂

mxinden:
Oh, sorry for missing https://github.com/libp2p/rust-libp2p/pull/2772 @elenaf9.

PR #2772, opened 2022-07-24
dcutr: fix clippy lints

elenaf9:
# Description

Fix clippy warnings in libp2p-dcutr: `new-without-default`, `single-match`, `large-enum-variant`, `redundant-pattern-matching` and `redundant-closure`.


protocols/dcutr/src/behaviour.rs
```diff
@@ -324,7 +320,6 @@ impl NetworkBehaviour for Behaviour {
 
 /// A [`NetworkBehaviourAction`], either complete, or still requiring data from [`PollParameters`]
 /// before being returned in [`Behaviour::poll`].
-#[allow(clippy::large_enum_variant)]
```
thomaseizinger, line 327: Personally, I always allow this one because the Box is quite the ergonomic hit when pattern matching and the lint desc. also says that one would need to measure the performance benefit gained from boxing.

thomaseizinger (approved):
ACK modulo one thing.

protocols/dcutr/src/behaviour.rs
elenaf9: I agree that in some cases it makes more sense to allow this lint rather than using a `Box`. 
That being said, in this case here we pass around the `inbound_connect` quite a bit without using it, and for all enums it appeared on this error was thrown. We can wrap it once in a `Box` and pass that around instead, without any ergonomic hit. The only place where `inbound_connect` is used is in 
https://github.com/libp2p/rust-libp2p/blob/f85a9909aca659f10f6b024fa0ed581199dddf90/protocols/dcutr/src/handler/relayed.rs#L371-L373
 where we have to use `AsMut` anyway to access the inner future.

mxinden (approved):
Thanks Elena!

PR #2777, opened 2022-07-25
Add a test to create a swarm with websocket transport

kpp:
I try to catch this error: https://github.com/nimiq/core-rs-albatross/issues/732

@mxinden @tomaka do you have any ideas how to emulate it?


mxinden:
Unfortunately I don't. I think having more logs right before the panic will be helpful.

kpp:
Ok, it didn't work.

PR #2780, opened 2022-07-28
docs/coding-guidelines: Add document

mxinden:
# Description

Add document outlining a set of coding guidelines followed and to be followed across the rust-libp2p code base.

My goal is not to impose these guidelines onto anyone, nor to be dogmatic about them. Instead I would like to discuss their value, eventually establishing a shared mindset, helping newcomers getting started faster.

For now, this is a draft only. Lots of the wording still need polishing. Suggestions / feedback / disagreement is welcome anytime.

This might be a helpful resource for discussions like https://github.com/libp2p/rust-libp2p/pull/2622#discussion_r929649318 in the future.

## Change checklist


docs/coding-guidelines.md
```diff
@@ -0,0 +1,269 @@
...
+
+<!-- markdown-toc end -->
+
+
+Below is a set of coding guidelines followed across the rust-libp2p code base.
+
+## Hierarchical State Machines
+
+If you sqint, rust-libp2p is just a big hierarchy of [state
+machines](https://en.wikipedia.org/wiki/Finite-state_machine) where parents pass
```
thomaseizinger, line 30: Personally, I am a fan of using a "one-line per sentence" rule for version-controlled markdown documents.

Git works on a line-by-line basis and sentences in natural language tend to change as a whole too. Using one sentence per line makes it really easy to use GitHub's suggestions to collaborate on documents like this and diffs are more meaningful too.

Just an idea :)

docs/coding-guidelines.md
```diff
@@ -0,0 +1,269 @@
...
+|PingBehaviour|  |IdentifyBehaviour|   |KademliaBehaviour|
+|-------------|  |-----------------|   |-----------------|
+|poll()       |  |poll()           |   |poll()           |
+`-------------'  `-----------------'   `-----------------'
+```
+
+Using hierarchical state machines is a deliberate choice throughout the
+rust-libp2p code base. It makes reasoning about control and data flow easy and
+works well with Rust's `Future` model. The archetecture pattern of
+hierarchical state machines should be used wherever possible.
```
thomaseizinger: Another really important point is determinism. You can't control the order in which futures in an executor are polled but you can control the order in which you compose state machines.

docs/coding-guidelines.md
```diff
@@ -0,0 +1,269 @@
...
+
+### Conventions for `poll` implementations
+
+The `poll` method of a single state machine can be complex especially when that
+state machine itself `poll`s many child state machines. The patterns shown below
+have proven useful and should be followed across the code base.
+
+``` rust
+fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output>{
+  loop {
```
thomaseizinger, line 78: Make a comment here about looping vs recursing.

docs/coding-guidelines.md
```diff
@@ -0,0 +1,269 @@
...
+      }
+
+      // The child did not make progress. It has registered the waker for a
+      // later wake up. Proceed with the other childs.
+      Poll::Ready(_) => {
+    }
+
+    match self.child_2.poll(cx) {
+      // As above.
+    }
```
thomaseizinger: I think it is important to mention that using continue in here will go back to child1 which could make progress again.

There is a subtle difference between looping over the entire thing and exhausting each child until it returns Pending. We want the former, not the latter.

docs/coding-guidelines.md
```diff
@@ -0,0 +1,269 @@
...
+
+    // Second priority is finishing local work, i.e. sending on the socket.
+    if let Poll::Ready(()) = socket.poll_ready(cx) {
+      todo!("Send messages")
+    }
+
+    // Last priority is accepting new work, i.e. reading from the socket.
+    if let Some(work_item) = socket.poll_next(cx) {
+      todo!("Start work on new work item")
+    }
```
thomaseizinger: I'd suggest we loop here too.

docs/coding-guidelines.md
```diff
@@ -0,0 +1,269 @@
...
+
+## Bound everything
+
+The concept of unboundedness is an illusion. Use bounded mechanisms to prevent
+unbounded memory growth and high latencies.
+
+### Channels
+
+When using channels (e.g. `futures::channel::mpsc` or `std::sync::mpsc`)
+always use the bounded variant, never use the unbounded variant. When using a
```
thomaseizinger, line 181: We should use clippy's mechanism of banning these functions!

https://rust-lang.github.io/rust-clippy/master/index.html#disallowed_methods

thomaseizinger (commented):
Excellent document, lots of good knowledge here :)

Left some comments and ignored typos for now, we can deal with those once we settle on the content.

docs/coding-guidelines.md
melekes: Are there downsides to this design? If so, should we list them?

docs/coding-guidelines.md
```diff
@@ -0,0 +1,269 @@
...
+        // but NEVER move to the next child if the current child made progress. Given
+        // that the current child might be able to make more progress, it did not yet
+        // register the waker in order for the root task to be woken up later on. Moving
+        // on to the next child might result in the larger `Future` task to stall as it
+        // assumes that there is no more progress to be made.
+      }
+
+      // The child did not make progress. It has registered the waker for a
+      // later wake up. Proceed with the other childs.
+      Poll::Ready(_) => {
```
melekes:
```suggestion
      Poll::Pending => {
```

?

melekes (commented):
👍

docs/coding-guidelines.md
mxinden: Agreed, though not sure how to put it into words.

There is this line already:

``` rust
        // or `continue`, thus polling `child_1` again. `child_1` can potentially make more progress:
        continue
```

Could you make a suggestion @thomaseizinger ?

docs/coding-guidelines.md
thomaseizinger: In my experience, the main downsides are:

- It _feels_ more verbose but I think it is not actually in reality when you would compare two solutions side by side.
- The mix of control flow (`loop`, `return`, `break`, `continue`) in `poll` functions together with the asynchronous and thus decoupled communication via events can be very hard to understand.

Both are a form of complexity that we are trading for correctness and performance which seems typical in Rust land.

docs/coding-guidelines.md
thomaseizinger: I'd suggest to just fill in the `match` here too and make a comment next to the `continue` of the `Poll::Pending` branch.

thomaseizinger:
This blogpost might also go onto a further reading list: https://sans-io.readthedocs.io/how-to-sans-io.html

docs/coding-guidelines.md
mxinden: > I'd suggest to just fill in the `match` here too

As in the same as the above? What would be the benefit of that? Isn't that just duplicating the same comments once more?

mxinden:
> This blogpost might also go onto a further reading list: https://sans-io.readthedocs.io/how-to-sans-io.html

I am a great fan of this coding style. It makes testing so easy.

That said, I don't think we apply this coding style anywhere within the rust-libp2p code-base. E.g. a `Transport` like `libp2p-dns` wrapps an I/O source, `multistream-select` wraps an I/O source, a `ConnectionHandler` has direct access to an I/O source. Thus I am having difficulties relating it to these coding guidelines.

docs/coding-guidelines.md
mxinden: :+1: Good point. Documented in 7714a30f.

docs/coding-guidelines.md
mxinden: :+1: Added in 3b07e71c.

docs/coding-guidelines.md
mxinden: Sounds good to me. Done in 995fd8cd.

docs/coding-guidelines.md
mxinden: :pray: 2b40afa4

docs/coding-guidelines.md
mxinden: Good idea. Though I am failing to have clippy pick up my `Clippy.toml` when I save it in the root of the repository.

``` toml
disallowed-methods = [
  { path = "futures::sync::mpsc::channel", reason = "does not enforce backpressure" },
]
```

@thomaseizinger does this work for you?

docs/coding-guidelines.md
thomaseizinger: I wouldn't duplicate the entire explanation but perhaps have a complete example of the control flow and make one comment at the bottom that highlights that we always `continue` to go back to the top. Even though it should be clear from the written text, I find that seeing the entire control flow is better.

Maybe also build in the case where the event from `child2` is dispatched to `child1`?

thomaseizinger:
> > This blogpost might also go onto a further reading list: [sans-io.readthedocs.io/how-to-sans-io.html](https://sans-io.readthedocs.io/how-to-sans-io.html)
> 
> I am a great fan of this coding style. It makes testing so easy.
> 
> That said, I don't think we apply this coding style anywhere within the rust-libp2p code-base. E.g. a `Transport` like `libp2p-dns` wrapps an I/O source, `multistream-select` wraps an I/O source, a `ConnectionHandler` has direct access to an I/O source. Thus I am having difficulties relating it to these coding guidelines.

Well, it does apply to some degree. `NetworkBehaviour` is basically a sans-IO state machine right? `Swarm` is passing events between the network and the behaviour with the minor exception that `Substream` has a boxed muxer inside that directly connects to the IO source. But given that we don't follow it fully, it is maybe not the best blog post to reference.

docs/coding-guidelines.md
thomaseizinger: It reliably works for me after a `cargo clean`.

docs/coding-guidelines.md
thomaseizinger: But I have to admit I am also in the "use an editor with word wrap enabled"-camp :D

docs/coding-guidelines.md
mxinden: I added https://github.com/libp2p/rust-libp2p/pull/2780/commits/ed4d6330487e5a0e8bf8dc61bc9e14367b32c5c8. Let me know what you think.

docs/coding-guidelines.md
mxinden: I documented the above in https://github.com/libp2p/rust-libp2p/pull/2780/commits/0ef322a27252bd53d87205419969fcdcefdceade. @thomaseizinger I really liked your wording which I mostly adopted. Hope you don't mind.

@melekes in case you can think of other downsides, please comment and let's get them included in the discussion and the document.

docs/coding-guidelines.md
```diff
@@ -0,0 +1,287 @@
+# Coding Guidelines
+
+<!-- markdown-toc start - Don't edit this section. Run M-x markdown-toc-refresh-toc -->
+**Table of Contents**
```
thomaseizinger, line 4: GitHub Markdown has a built-in ToC:

image

It is a bit hidden so I don't know we want to rely on people using it if they want a ToC.

docs/coding-guidelines.md
```diff
@@ -0,0 +1,287 @@
...
+
+Below is a set of coding guidelines followed across the rust-libp2p code base.
+
+## Hierarchical State Machines
+
+If you sqint, rust-libp2p is just a big hierarchy of [state
+machines](https://en.wikipedia.org/wiki/Finite-state_machine) where parents pass
+events down to their children and children pass events up to their parents.
+
+![Architecture](architecture.svg)
```
thomaseizinger, line 33: Instead of embedding the file, you could reference it planttext.com: https://www.planttext.com/api/plantuml/svg/SoWkIImgAStDuL80Wk3onA9S1PiQNLs5eFpy4gVKZCIopFpI8dH9v_oylDJaaipyl83yV0oj8KM9USK5-KKbO0aKWo0A0ZBpqb7DngB8CoKrhoGphPAWGk_4bDJSdCn4RAwrKYZ8pydHqCIY5qKArKEeSd3le0gZU09L2iFfgD8O3aP8EwJcfG0Z0m00

The same token also works as an "edit" link: https://www.planttext.com/?text=SoWkIImgAStDuL80Wk3onA9S1PiQNLs5eFpy4gVKZCIopFpI8dH9v_oylDJaaipyl83yV0oj8KM9USK5-KKbO0aKWo0A0ZBpqb7DngB8CoKrhoGphPAWGk_4bDJSdCn4RAwrKYZ8pydHqCIY5qKArKEeSd3le0gZU09L2iFfgD8O3aP8EwJcfG0Z0m00

docs/coding-guidelines.md
```diff
@@ -0,0 +1,287 @@
...
+rust-libp2p code base. It makes reasoning about control and data flow simple. It
+works well with Rust's `Future` model. It allows fine-grain control e.g. on the
+order child state machines are polled.
+
+The above comes with downsides. It feels more verbose. The mix of control flow (`loop`, `return`,
+`break`, `continue`) in `poll` functions together with the asynchronous and thus decoupled
+communication via events can be very hard to understand. Both are a form of complexity that we are
+trading for correctness and performance which aligns with Rust's and rust-libp2p's goals.
+
+The archetecture pattern of hierarchical state machines should be used wherever possible.
```
thomaseizinger:
```suggestion
The architecture pattern of hierarchical state machines should be used wherever possible.
```

docs/coding-guidelines.md
thomaseizinger: @mxinden Any comment on this? We don't have to resolve it here. Should I open a separate discussion?

thomaseizinger (approved):

docs/coding-guidelines.md
mxinden: Neat. Though as you say, quite hidden. I suggest we keep the additional ToC, though that is not a strong opinion.

docs/coding-guidelines.md
mxinden: I am not familiar with planttext. Is it a service we can rely on? What if they prune their database?

docs/coding-guidelines.md
mxinden: Like this, correct @thomaseizinger? https://github.com/mxinden/rust-libp2p/commit/ecf86c81dd54fb6180749131cc6cb630883bfecd

With GitHubs diff highlighting showing what within a line has changed, I don't think there is a great benefit for it. As a downside, I think it adds complexity, e.g. having to explain this to newcomers.

> Just an idea :)

I don't feel strongly about it. It seems like you don't feel strongly about it either. Thus I will proceed without the change.

That said, feel free to propose a new pull request with the commit above.

---

> But I have to admit I am also in the "use an editor with word wrap enabled"-camp :D

Same here. I even stopped truncating my emails to 80 characters :exploding_head:

docs/coding-guidelines.md
thomaseizinger: I've been using it for years without issues.

There is no database, I am pretty sure the diagram source code is embedded in the URL in some form of encoding.

docs/coding-guidelines.md
mxinden: Will be introduced with https://github.com/libp2p/rust-libp2p/pull/2823.

docs/coding-guidelines.md
mxinden: That makes sense. I would argue though that there is benefit in being able to access the diagram and thus read the document without internet access.

docs/coding-guidelines.md
thomaseizinger: If I get to choose, I'd pick one sentence per line.

We don't have to enforce it though (I am yet to find a tool that reliably does).

Are you happy with doing it on a personal preference basis?

mxinden: > Are you happy with doing it on a personal preference basis?

For markdown files on rust-libp2p? Yes, that is fine by me.

PR #2784, opened 2022-07-31
swarm/src/behaviour: Deprecate NetworkBehaviourEventProcess

mxinden:
# Description

In preparation for https://github.com/libp2p/rust-libp2p/pull/2751.

## Change checklist


swarm/src/behaviour.rs
```diff
@@ -79,16 +79,11 @@ pub(crate) type THandlerOutEvent<THandler> =
...
-/// * Alternatively, users that need access to the root [`NetworkBehaviour`] implementation while
-/// processing emitted events, can specify `#[behaviour(event_process = true)]` (default is false).
-/// Events generated by the behaviour's struct members are delegated to [`NetworkBehaviourEventProcess`]
-/// trait implementations. Those must be provided by the user on the type that [`NetworkBehaviour`]
-/// is derived on.
+/// Events ([`NetworkBehaviour::OutEvent`]) returned by each `struct` member are wrapped in a new
+/// `enum` event, with an `enum` variant for each `struct` member. Users can define this event
+/// `enum` themselves and provide the name to the derive macro via `#[behaviour(out_event =
+/// "MyCustomOutEvent")]`. If the user does not specify an `out_event`, the derive macro generates
+/// the event definition itself, naming it `<STRUCT_NAME>Event`.
```
elenaf9: This is not true yet, as the changes of the `swarm_derive::NetworkBehaviour` proc macro are not included in this PR.

swarm/src/behaviour.rs
```diff
@@ -340,6 +327,13 @@ pub trait PollParameters {
 ///
 /// You can opt out of this behaviour through `#[behaviour(event_process = false)]`. See the
 /// documentation of [`NetworkBehaviour`] for details.
+#[deprecated(
+    since = "0.38.0",
+    note = "Use `#[behaviour(out_event = \"MyBehaviourEvent\")]` instead. See \
+            https://github.com/libp2p/rust-libp2p/blob/master/swarm/CHANGELOG.md#0380 \
```
thomaseizinger, line 332: I think it would be better to link to a heading rather than a line. This line will change as we add more entries to it because we add to the top, meaning the reference will point to the wrong thing.

thomaseizinger (approved):
Concept ACK, just one suggestion.

elenaf9 (commented):
If I understand it correctly, the plan is to publish a release that includes this changes here (deprecating `NetworkBehaviourEventProcess`), but do a separate release for the main changes of #2751 (changing the proc macro, removing `NetworkBehaviourEventProcess` completely). 

I am not sure if it make sense to deprecate `NetworkBehaviourEventProcess` without publishing the changes in the proc macro. User that want to upgrade can not use the automatically generated `OutEvent` yet. Imo ideally the first release should already include all changes of #2751, but not yet remove support for `NetworkBehaviourEventProcess` / <s>`event_process=false`</s> `event_process=true`.
<s>The problem that I see here is what the default should be (= What should be generated if a struct is only annotated with `#[derive(NetworkBehavour)]` and nothing else?).
 In the old version the default was to expect `NetworkBehaviourEventProcess`, in the new version the default will be to generate the `OutEvent` enum.</s>
What do you think about the following:
- Include all changes of #2751, but only deprecate `NetworkBehaviourEventProcess` instead of removing it and keep support for <s>`event_process=false`</s> `event_process=true`  (which means that for this interim period the macro would have to support both, `event_process` and the new generated `OutEvent`)
- <s>Change default for `event_process` to `true` instead of `false`. User that would like to keep using it would have to explicitly add `#[event_process=false]` rather than it being the default</s>
- Per default generate the `OutEvent` enum
- In the following release then just remove `NetworkBehaviourEventProcess` and  `event_process`.

We'd have to explicitly point that out in our Changelog, but I think it would be more convenient for the majority of our users.

mxinden:
> Change default for `event_process` to `true` instead of `false`. User that would like to keep using it would have to explicitly add `#[event_process=false]` rather than it being the default

I think there is a confusion here. Today, by default, `event_process` is `false` and thereby the derive macro uses the `out_event` mechanism instead of the `NetworkBehaviourEventProcess` mechanism. See https://github.com/libp2p/rust-libp2p/pull/2214.

https://github.com/libp2p/rust-libp2p/blob/eaf3f3a7fb4aba228ebd15366ab79b64cf37832c/swarm-derive/src/lib.rs#L73-L75

Am I missing something @elenaf9?

swarm/src/behaviour.rs
mxinden: Good catch! Thanks. Addressed in https://github.com/libp2p/rust-libp2p/pull/2784/commits/8d5f6a5098ac6eb9e12ec9376e29832c72121621.

elenaf9:
> > Change default for `event_process` to `true` instead of `false`. User that would like to keep using it would have to explicitly add `#[event_process=false]` rather than it being the default
> 
> I think there is a confusion here. Today, by default, `event_process` is `false` and thereby the derive macro uses the `out_event` mechanism instead of the `NetworkBehaviourEventProcess` mechanism. See #2214.
> 
> https://github.com/libp2p/rust-libp2p/blob/eaf3f3a7fb4aba228ebd15366ab79b64cf37832c/swarm-derive/src/lib.rs#L73-L75
> 
> Am I missing something @elenaf9?

You are right, I mixed it up what  `event_process=false` means. But since default `event_process=false` already means that `out_event` is used it makes it even easier. So to rephrase my idea above: 
Instead of just deprecating `NetworkBehaviourEventProcess`, in my opinion the first release should already include the changes in the proc macro (and the changes in our examples and other behaviours, so effectively everything from #2751), just that we don't remove the support for `event_process = true` and  `NetworkBehaviourEventProcess` yet. For users that want to stick with the deprecated `event_process` nothing would change. The benefit would be that users that do want to switch can then already use the generated enum. Any reason why this would not make sense?

thomaseizinger:
I think that is a really good idea actually!

swarm/src/behaviour.rs
mxinden: `#0380` is the header once we remove `[unreleased]` from it. Right now it is:

```
https://github.com/libp2p/rust-libp2p/blob/master/swarm/CHANGELOG.md#0380-unreleased
```

Does that make sense @thomaseizinger?

mxinden:
@elenaf9 @thomaseizinger with https://github.com/libp2p/rust-libp2p/pull/2792 merged, would you mind taking another look?

swarm/src/behaviour/either.rs
```diff
@@ -237,6 +236,7 @@ where
     }
 }
 
+#[allow(deprecated)]
 impl<TEvent, TBehaviourLeft, TBehaviourRight> NetworkBehaviourEventProcess<TEvent>
     for Either<TBehaviourLeft, TBehaviourRight>
 where
```
elenaf9, line 242: What is the usage case for `NetworkBehaviourEventProcess` on `Either` (or `Toggle` below)?
As far as I understood it, the purpose of `NetworkBehaviourEventProcess` is only to be used in combination with the `NetworkBehaviour` derive macro.

mxinden: I guess when you use `Either` or `Toggle` when building a nested `NetworkBehaviour`. See for example:

https://github.com/libp2p/rust-libp2p/blob/a4110a2b6939d5b460f11df6dd158308f517becf/swarm-derive/tests/test.rs#L354-L369

mxinden: Here with `NetworkBehaviourEventProcess`:

https://github.com/libp2p/rust-libp2p/blob/2f2b7cbec127a48d869d87ad52d69a4c4c29ddb5/swarm-derive/tests/test.rs#L328-L352

elenaf9: But you wouldn't call the `NetworkBehaviourEventProcess` implementation of an inner behaviour, no? Instead you implement `NetworkBehaviourEventProcess` on the parent (in the above case in `Foo`) and handle the event there. Why would anyone want to inject the event back into an inner behaviour?

None of our behaviours currently implement `NetworkBehaviourProcess`, so the below trait bound for it would not apply anyway:
https://github.com/libp2p/rust-libp2p/blob/a4110a2b6939d5b460f11df6dd158308f517becf/swarm/src/behaviour/toggle.rs#L237-L239
So `libp2p::ping::Ping` does not implement `NetworkBehaviourEventProcess`, therefore `Toggle<libp2p::ping::Ping>` will also not implement it.

elenaf9: Unless I am missing something, there was never really a use-case for `NetworkBehaviourEventProcess` on those two behaviours in the first place. However, that's not really a concern of this PR. 
So I guess I am fine with for now with keeping it the way it is,  since this trait will be remove anyway in the next release.

swarm/CHANGELOG.md
```diff
@@ -1,8 +1,72 @@
 # 0.38.0 [unreleased]
 
-- Update dial address concurrency factor to `8`, thus dialing up to 8 addresses concurrently for a single connection attempt. See `Swarm::dial_concurrency_factor` and [PR 2741].
+- Remove `NetworkBehaviourEventProcess`. When deriving `NetworkBehaviour` on a custom `struct` users
```
elenaf9:
```suggestion
- Deprecate `NetworkBehaviourEventProcess`. When deriving `NetworkBehaviour` on a custom `struct` users
```

swarm/CHANGELOG.md
```diff
@@ -1,8 +1,72 @@
...
+  documentation](https://docs.rs/libp2p/latest/libp2p/swarm/trait.NetworkBehaviour.html) for
+  details.
+
+  Previously
+
+  ``` rust
+  #[derive(NetworkBehaviour)]
+  #[behaviour(event_process = true)]
+  struct MyBehaviour {
+      floodsub: Gossipsub,
```
elenaf9:
```suggestion
      gossipsub: Gossipsub,
```

swarm/CHANGELOG.md
```diff
@@ -1,8 +1,72 @@
...
+  }
+  ```
+
+  Now
+
+  ``` rust
+  #[derive(NetworkBehaviour)]
+  #[behaviour(out_event = "MyBehaviourEvent")]
+  struct MyBehaviour {
+      floodsub: Gossipsub,
```
elenaf9:
```suggestion
      gossipsub: Gossipsub,
```

swarm/CHANGELOG.md
```diff
@@ -1,8 +1,72 @@
...
+      mdns: Mdns,
+  }
+
+  enum MyBehaviourEvent {
+      Floodsub(FloodsubEvent),
```
elenaf9:
```suggestion
      Gossipsub(GossipsubEvent),
```

swarm/CHANGELOG.md
```diff
@@ -1,8 +1,72 @@
...
-- Update dial address concurrency factor to `8`, thus dialing up to 8 addresses concurrently for a single connection attempt. See `Swarm::dial_concurrency_factor` and [PR 2741].
+- Remove `NetworkBehaviourEventProcess`. When deriving `NetworkBehaviour` on a custom `struct` users
+  should either bring their own `OutEvent` via `#[behaviour(out_event = "MyBehaviourEvent")]` or,
+  when not specified, have the derive macro generate one for the user.
 
-- Update to `libp2p-core` `v0.35.0`.
+  See [`NetworkBehaviour`
+  documentation](https://docs.rs/libp2p/latest/libp2p/swarm/trait.NetworkBehaviour.html) for
+  details.
+
```
elenaf9: Was the PR not linked on purpose (same for the changelog below)?

elenaf9 (approved):

swarm/CHANGELOG.md
```diff
@@ -13,6 +77,10 @@
   documentation](https://docs.rs/libp2p/latest/libp2p/swarm/trait.NetworkBehaviour.html) for
   details.
 
+- Update dial address concurrency factor to `8`, thus dialing up to 8 addresses concurrently for a single connection attempt. See `Swarm::dial_concurrency_factor` and [PR 2741].
+
+- Update to `libp2p-core` `v0.35.0`.
+
```
thomaseizinger: Can we add the changelog entry above so this doesn't come up in the diff?

swarm/src/behaviour.rs
thomaseizinger: Ah yes, I thought this was the line number.

thomaseizinger (approved):
ACK modulo mine and @elenaf9 's comments :)

elenaf9:
One more thing: I think if we are deprecating `NetworkBehaviourEventProcess` we should also already change our examples & tests to not use it anymore (`examples/distributed-key-value-store.rs`, `examples/chat.rs`, ...). Can be done in a separate PR, but imo  it should be part of the same release.

swarm/CHANGELOG.md
mxinden: Thanks for the catch.

swarm/CHANGELOG.md
mxinden: :+1: linked with b5cd98da

swarm/CHANGELOG.md
mxinden: I ordered the entries by significance. I don't think users care much about the `libp2p-core` update. I do think users care about how to upgrade from `NetworkBehaviourEventProcess`. Does that reasoning make sense @thomaseizinger?

mxinden:
> One more thing: I think if we are deprecating `NetworkBehaviourEventProcess` we should also already change our examples & tests to not use it anymore (`examples/distributed-key-value-store.rs`, `examples/chat.rs`, ...). Can be done in a separate PR, but imo it should be part of the same release.

That is a good point. https://github.com/libp2p/rust-libp2p/pull/2784/commits/0f3ef691ff1499c3fb85f96c21a26b44fea28e23 does the change.

mxinden:
As far as I can tell all comments are addressed and everyone approved. Thus I am merging here. Happy to do any follow-ups in case I missed something.

As always, thanks for all the input @elenaf9 and @thomaseizinger!

PR #2785, opened 2022-07-31
swarm/src/connection: Test max_negotiating_inbound_streams

mxinden:
# Description

Test that `HandlerWrapper` upholds the provided
`max_negotiating_inbound_streams` limit.

Feature introduced in #2697.

## Change checklist

- ~~[ ] I have made corresponding changes to the documentation~~
- ~~[ ] I have added tests that prove my fix is effective or that my feature works~~
- ~~[ ] A changelog entry has been made in the appropriate crates~~


mxinden:
@thomaseizinger @elenaf9 could one of you give this pull request a review?

swarm/src/connection/handler_wrapper.rs
```diff
@@ -440,3 +440,84 @@ pub enum Event<TOutboundOpenInfo, TCustom> {
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::handler::PendingConnectionHandler;
+    use quickcheck::*;
+    use std::num::NonZeroU8;
+    use std::sync::Arc;
+
+    struct DummySubstream(Arc<()>);
```
thomaseizinger:
```suggestion
    struct PendingSubstream(Arc<()>);
```

thomaseizinger: Plus, I'd prefer such kind of infrastructure to be at the bottom of the module. More generally, items in a module / file should be sorted by priority / likelihood of wanting to see them as you open the file and read it top to bottom.

Having imports at the top somewhat contradicts this rule but I think that is too controversial to adopt :D

swarm/src/connection/handler_wrapper.rs
```diff
@@ -440,3 +440,84 @@ pub enum Event<TOutboundOpenInfo, TCustom> {
...
+            wrapper.inject_substream(substream, SubstreamEndpoint::Listener);
+
+            assert_eq!(
+                Arc::strong_count(&alive_substreams_counter),
+                max_negotiating_inbound_streams + 1,
+                "Expect substream exceeding the limit to be dropped."
+            );
+        }
+
+        QuickCheck::new().quickcheck(prop as fn(_));
```
thomaseizinger, line 489: Have you ever considered using the [`#[quickcheck]`](https://github.com/BurntSushi/quickcheck#the-quickcheck-attribute) macro?

swarm/src/connection/handler_wrapper.rs
```diff
@@ -440,3 +440,84 @@ pub enum Event<TOutboundOpenInfo, TCustom> {
...
+        }
+
+        fn poll_close(self: Pin<&mut Self>, _cx: &mut Context<'_>) -> Poll<std::io::Result<()>> {
+            Poll::Pending
+        }
+    }
+
+    #[test]
+    fn max_negotiating_inbound_streams() {
+        fn prop(max_negotiating_inbound_streams: NonZeroU8) {
```
thomaseizinger: This is only a `NonZeroU8` so that quickcheck does not generate `0`?

swarm/src/connection/handler_wrapper.rs
```diff
@@ -440,3 +440,84 @@ pub enum Event<TOutboundOpenInfo, TCustom> {
...
+            let max_negotiating_inbound_streams: usize =
+                max_negotiating_inbound_streams.get().into();
+            let handler = PendingConnectionHandler::new("test".to_string());
```
thomaseizinger: We could inline this variable.

swarm/src/connection/handler_wrapper.rs
thomaseizinger: If the zero case is not interesting, shouldn't we also use a `NonZero` type in `HandlerWrapper`?

core/src/upgrade/pending.rs
```diff
@@ -0,0 +1,76 @@
...
+
+use crate::upgrade::{InboundUpgrade, OutboundUpgrade, ProtocolName, UpgradeInfo};
+use futures::future;
+use std::iter;
+use void::Void;
+
+/// Implementation of [`UpgradeInfo`], [`InboundUpgrade`] and [`OutboundUpgrade`] that always
+/// returns a pending upgrade.
+#[derive(Debug, Copy, Clone)]
+pub struct PendingUpgrade<P> {
```
thomaseizinger, line 30: Why do we need this to be generic?

swarm/src/handler/pending.rs
```diff
@@ -0,0 +1,116 @@
...
+    fn listen_protocol(&self) -> SubstreamProtocol<Self::InboundProtocol, Self::InboundOpenInfo> {
+        SubstreamProtocol::new(PendingUpgrade::new(self.protocol_name.clone()), ())
+    }
+
+    fn inject_fully_negotiated_inbound(
+        &mut self,
+        _: <Self::InboundProtocol as InboundUpgrade<NegotiatedSubstream>>::Output,
+        _: Self::InboundOpenInfo,
+    ) {
+        unreachable!("`PendingUpgrade` is never successful.");
```
thomaseizinger: Why is this different from `inject_fully_negotiated_outbound`?

thomaseizinger (approved):
LGTM overall, a few questions.

swarm/src/connection/handler_wrapper.rs
mxinden: I haven't. Looks handy to me. Though I would prefer being consistent, i.e. move the entire code base to `#[quickcheck]` at once instead of diverging on a couple of pull requests.

swarm/src/connection/handler_wrapper.rs
mxinden: Good point. I think we should support the use-case of `0`. That is, I think we should support the use-case of blocking all inbound streams.

With that in mind, the test should cover it. https://github.com/libp2p/rust-libp2p/pull/2785/commits/87b9096bf9d987cb366909b93469cb55144579a8 does the adjustments.

Thanks @thomaseizinger.

core/src/upgrade/pending.rs
mxinden: Are you suggesting to use a `String` instead? What if a user wants to use it with a type other than `String` that implements `ProtocolName`?

swarm/src/handler/pending.rs
mxinden: In `inject_fully_negotiated_outbound` we can leverage the `OutboundOpenInfo` to carry an uninhabited type (in this case `Void`), thus proving it to be impossible at compile time. for `inject_fully_negotiated_inbound` we don't have a corresponding `InboundOpenInfo` mechanism, and can thus not apply this trick.

swarm/src/connection/handler_wrapper.rs
mxinden: :+1: 8c1090cd

swarm/src/connection/handler_wrapper.rs
mxinden: :+1: 62f5e4c9

mxinden:
Thanks for the review @thomaseizinger. I addressed all your comments. Let me know what you think.

core/src/upgrade/pending.rs
mxinden: With "user" I mean am thinking of potential users beyond the test introduced in this pull request.

swarm/src/handler/pending.rs
thomaseizinger: But we do or what else is line 62? 🤔

core/src/upgrade/pending.rs
thomaseizinger: > Are you suggesting to use a `String` instead? What if a user wants to use it with a type other than `String` that implements `ProtocolName`?

`&'static str` would probably be my preference and I'd say we can make it generic once needed? For tests, a static protocol name is probably good enough for many if not all cases.

mxinden: `libp2p-core` has the `ProtocolName` abstraction. Thus I would suggest all of `libp2p-core` should use that abstraction. I am happy to design an alternative approach. Though I think that should happen in a different pull request.

swarm/src/handler/pending.rs
mxinden: > for `inject_fully_negotiated_inbound` we don't have a corresponding `InboundOpenInfo` mechanism, and can thus not apply this trick.

Sorry. This is wrong. We do have the `InboundOpenInfo` mechanism. The problem is that in `ConnectionHandler::listen_protocol` and within that the call to `SubstreamProtocol::new` we have to provide the `InboundOpenInfo`. We can not provide an instance of `Void`, thus we can not use `Void` and `InboundOpenInfo` to make this state impossible at compile time.

@thomaseizinger does that make sense?

thomaseizinger: Ah yes that makes perfect sense!

Couldn't we use the actual substream variable though? The upgrade uses pending so we don't have to instantiate `void` but it would be "resolved" here, correct?

mxinden: Good point. Thanks @thomaseizinger. What do you think of https://github.com/libp2p/rust-libp2p/pull/2785/commits/a0960bb091b8d923c05baa298b241c4e2dc08aeb?

swarm/src/handler/pending.rs
```diff
@@ -0,0 +1,120 @@
...
+    fn inject_fully_negotiated_outbound(
+        &mut self,
+        protocol: <Self::OutboundProtocol as OutboundUpgrade<NegotiatedSubstream>>::Output,
+        _info: Self::OutboundOpenInfo,
+    ) {
+        void::unreachable(protocol);
+        #[allow(unreachable_code)]
+        {
+            void::unreachable(_info);
+        }
```
thomaseizinger, line 76:
```suggestion
        _: Self::OutboundOpenInfo,
    ) {
        void::unreachable(protocol);
```

I'd almost suggest to not use `void::unreachable` for the `info` part if we have to use `#[allow(unreachable_code)]` for it to work.

thomaseizinger (approved):
One suggestion but don't feel strongly about it.

swarm/src/handler/pending.rs
mxinden: I prefer the additional `void::unreachable`, just in case we refactor this part of the codebase in the future.

mxinden:
Thanks for the help @thomaseizinger!

PR #2786, opened 2022-08-01
enable sharing the binary file and use a larger buffer

qidu:
# Description
Update an examle for trasnfering a binary file

## Change checklist

~~I have added tests that prove my fix is effective or that my feature works~~
~~A changelog entry has been made in the appropriate crates~~


examples/file-sharing.rs
```diff
@@ -196,6 +208,8 @@ enum CliArgument {
         name: String,
     },
     Get {
+        #[clap(long)]
+        path: PathBuf,
```
mxinden: Instead of providing a file path for the retrieved file to be written to, how about leveraging unix pipelines?

E.g. one would do:

``` bash
cargo run -- get [...] > my_file
```

What do you think?

mxinden (commented):
Thanks for the pull request.

I am in favor of the move to `Vec<u8>`. I suggest we don't write to a file but continue to write to stdout.

examples/file-sharing.rs
qidu: Good idea! Thanks for your advices. I've changed the output to stdout().

examples/file-sharing.rs
```diff
@@ -134,8 +136,17 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
+                            let mut reader = BufReader::new(File::open(&path)?);
+                            let mut vec = Vec::<u8>::new();
+                            let mut buffer = reader.fill_buf()?;
+                            let mut len = buffer.len();
+                            while len > 0 {
+                                vec.extend_from_slice(buffer);
+                                reader.consume(len);
+                                buffer = reader.fill_buf()?;
+                                len = buffer.len();
+                            }
```
mxinden: Given that this is reading the entire file into `vec`, what is the difference to the previous version, i.e. `std::fs::read_to_string`? If you want to read an entire file as `u8`, why not use the below?

https://doc.rust-lang.org/std/fs/fn.read.html

qidu: OK

qidu: OK. fixed.

examples/file-sharing.rs
```diff
@@ -134,8 +135,7 @@ async fn main() -> Result<(), Box<dyn Error>> {
                     // Reply with the content of the file on incoming requests.
                     Some(network::Event::InboundRequest { request, channel }) => {
                         if request == name {
-                            let file_content = std::fs::read_to_string(&path)?;
-                            network_client.respond_file(file_content, channel).await;
+                            network_client.respond_file(&std::fs::read(&path)?, channel).await;
```
mxinden:
```suggestion
                            network_client.respond_file(std::fs::read(&path)?, channel).await;
```

Why borrow here when we clone again later on?

qidu: but expected `&Vec<u8>`  here.  any better choice?

examples/file-sharing.rs
```diff
@@ -351,9 +351,9 @@ mod network {
         }
 
         /// Respond with the provided file content to the given request.
-        pub async fn respond_file(&mut self, file: String, channel: ResponseChannel<FileResponse>) {
+        pub async fn respond_file(&mut self, file: &Vec<u8>, channel: ResponseChannel<FileResponse>) {
             self.sender
-                .send(Command::RespondFile { file, channel })
+                .send(Command::RespondFile { file: file.to_vec(), channel })
```
thomaseizinger: Nit: I think the reference on line 354 is unnecessary if we are going to create a copy anyway and it will remove the diff on line 356.

thomaseizinger (approved):
LGTM

One small suggestion.

examples/file-sharing.rs
qidu: OK, done. Thanks~

examples/file-sharing.rs
```diff
@@ -351,9 +351,9 @@ mod network {
         }
 
         /// Respond with the provided file content to the given request.
-        pub async fn respond_file(&mut self, file: String, channel: ResponseChannel<FileResponse>) {
+        pub async fn respond_file(&mut self, file: Vec<u8>, channel: ResponseChannel<FileResponse>) {
             self.sender
-                .send(Command::RespondFile { file, channel })
+                .send(Command::RespondFile { file: file, channel })
```
thomaseizinger:
```suggestion
                .send(Command::RespondFile { file, channel })
```

thomaseizinger (approved):

mxinden (approved):
Looks good to me. Thanks.

mxinden (approved):
Looks good to me. Thanks for the follow ups.

qidu:
> 



> 

Thanks~

PR #2790, opened 2022-08-01
examples/README.md: Fix tutorial link

lukehinds:
Currently points to a file that no longer exists, producing a 404

Instead link to `/src/tutorials/ping.rs`


mxinden (approved):
Thanks @lukehinds for the fix.

Hope the tutorial is helpful.

PR #2792, opened 2022-08-02
swarm-derive/: Generate OutEvent if not provided

mxinden:
# Description

Generate `NetworkBehaviour::OutEvent` if not provided through
`#[behaviour(out_event = "MyOutEvent")]` and event processing is
disabled (default).

Extracted from https://github.com/libp2p/rust-libp2p/pull/2751.

As suggested by @elenaf9 in https://github.com/libp2p/rust-libp2p/pull/2784#pullrequestreview-1056551989 we would do the following:

1. Merge this pull request.
2. Merge the pull request deprecating `NetworkBehaviourEventProcess`.
3. Cut a new release.
4. Remove `NetworkBehaviourEventProcess` via https://github.com/libp2p/rust-libp2p/pull/2751.

What do folks think?

- Initial pull request https://github.com/libp2p/rust-libp2p/pull/2751
- Deprecation pull request https://github.com/libp2p/rust-libp2p/pull/2784

## Change checklist


swarm/src/behaviour.rs
```diff
@@ -79,16 +79,11 @@ pub(crate) type THandlerOutEvent<THandler> =
 /// it will delegate to each `struct` member and return a concatenated array of all addresses
 /// returned by the struct members.
 ///
-/// When creating a custom [`NetworkBehaviour`], you must choose one of two methods to respond to
```
mxinden, line 82: Note that the documentation on `NetworkBehaviour` now no longer documents `NetworkBehaviourEventProcess`. My rational being that we will deprecate it anyways, thus there is no use in pointing newcomers to it.

Any objections?

thomaseizinger: Nope.

thomaseizinger (approved):

elenaf9 (approved):
Maybe I overlooked it, but is there a test that validates for a generated `OutEvent` that it looks as expected? Could be done by implementing a simple trait for the generated OutEvent or something, e.g. `impl TryInto<libp2p::kad::KademliaEvent> for FooEvent` (for each inner behaviour).

thomaseizinger:
> Maybe I overlooked it, but is there a test that validates for a generated `OutEvent` that it looks as expected? Could be done by implementing a simple trait for the generated OutEvent or something, e.g. `impl TryInto<libp2p::kad::KademliaEvent> for FooEvent` (for each inner behaviour).

We could also write a function that matches on an instance of the enum. Should also validate the enum structure and the variant names.

mxinden:
@elenaf9 @thomaseizinger good idea. What do you think of https://github.com/libp2p/rust-libp2p/pull/2792/commits/b4e32ed7a908aded7678f8e6a9d30ab29ee34ec1?

elenaf9 (approved):
Lgtm!

swarm-derive/tests/test.rs
```diff
@@ -43,8 +43,14 @@ fn one_field() {
...
 
     #[allow(dead_code)]
+    #[allow(unreachable_code)]
     fn foo() {
-        require_net_behaviour::<Foo>();
+        let _out_event: <Foo as NetworkBehaviour>::OutEvent = unimplemented!();
+        match _out_event {
+            FooEvent::Ping(event) => {
+                let _: libp2p::ping::Event = event;
+            }
```
thomaseizinger: Why not just extend the pattern match?

```suggestion
            FooEvent::Ping(libp2p::ping::Event { .. }) => { }
```

thomaseizinger (approved):

swarm-derive/tests/test.rs
mxinden: Done in https://github.com/libp2p/rust-libp2p/pull/2792/commits/ff636810156a54fd9eb63566ed1093352ef650bb. Though unfortunately I don't see a way to do the same for the identify and kademlia event given that they are enums.

Issue #2767, opened 2022-07-21
Replace `ConnectedPoint` with `Dialer` in `libp2p_swarm::DialError::WrongPeerId`


dmitry-markin:
## Description

Change the [`endpoint`](https://github.com/libp2p/rust-libp2p/blob/62622a1bad70381e56237b4da4d9abcf2b1a0e24/swarm/src/lib.rs#L1444) type to `Dialer`, extracting it from [`ConnectedPoint`](https://github.com/libp2p/rust-libp2p/blob/62622a1bad70381e56237b4da4d9abcf2b1a0e24/core/src/connection.rs#L116).

## Motivation

Currently in case of [`DialError`](https://github.com/libp2p/rust-libp2p/blob/62622a1bad70381e56237b4da4d9abcf2b1a0e24/swarm/src/lib.rs#L1423) with `WrongPeerId` the [`ConnectedPoint`](https://github.com/libp2p/rust-libp2p/blob/62622a1bad70381e56237b4da4d9abcf2b1a0e24/core/src/connection.rs#L116) is returned, which can contain `Listener`, which can't happen, because we are the dialing side. Replacing `ConnectedPoint` with `Dialer` can help to resolve this ambiguity and eliminate client code assertions.

`ConnectedPoint` endpoint type in `WrongPeerId` was introduced in [this PR](https://github.com/libp2p/rust-libp2p/pull/2428#discussion_r786066319), which was merged [here](https://github.com/libp2p/rust-libp2p/pull/2441).

The original discussion that lead to this issue is in [the substrate PR](https://github.com/paritytech/substrate/pull/11872#discussion_r926094382) that changes the error printed when we connect to a bootnode that provides a different than expected peer id.

## Are you planning to do it yourself in a pull request?

~~Maybe.~~ Yes.

mxinden:
That sounds reasonable to me. Thanks @dmitry-markin for the detailed report. 

I am not going to work on this any time soon. I am happy to help you implement it though. Given your "maybe", are you still interested?

dmitry-markin:
Yes, I think I can look into it in a week or so. The one thing I worry about is that these changes will definitely break the API, and I don't know how to handle it. I.e, I can update substrate, but we need to somehow notify other clients.

mxinden:
Given that this is not a subtle change, but a change at the type system, I think users will (a) notice early and (b) can consult the changelog to get more advice. In other words, I am not worried about the breaking API change @dmitry-markin.

[...]

PR #2793, opened 2022-08-02
Use `Multiaddr` instead of `ConnectedPoint` in `DialError::WrongPeerId`

dmitry-markin:
# Description

`Listener` variant of `ConnectedPoint` is meaningless in `DialError::WrongPeerId`, so this PR replaces `ConnectedPoint` with `Multiaddr` (previously returned in the `Dialer` variant).

This change is proposed in https://github.com/libp2p/rust-libp2p/issues/2767.

## Open Questions

I'm unsure whether it makes sense to also include `role_override` into `WrongPeerId` error. This PR just drops it.

## Change checklist


swarm/src/lib.rs
```diff
@@ -1454,9 +1454,15 @@ impl From<PendingOutboundConnectionError<io::Error>> for DialError {
...
             PendingConnectionError::Aborted => DialError::Aborted,
-            PendingConnectionError::WrongPeerId { obtained, endpoint } => {
-                DialError::WrongPeerId { obtained, endpoint }
-            }
+            PendingConnectionError::WrongPeerId { obtained, endpoint } => match endpoint {
+                ConnectedPoint::Dialer {
+                    address,
+                    role_override: _,
+                } => DialError::WrongPeerId { obtained, address },
+                ConnectedPoint::Listener { .. } => panic!(
```
mxinden, line 1462: Instead of a `panic` at runtime I would prefer making this state impossible at compile time. 

Would splitting `PendingConnectionError` be an option @dmitry-markin?

dmitry-markin: Thanks, this seems reasonable, I'll try to do it.

dmitry-markin: We can replace https://github.com/libp2p/rust-libp2p/blob/a4110a2b6939d5b460f11df6dd158308f517becf/swarm/src/connection/pool.rs#L622-L731 with two big arms for `PendingPoint::Dialer` & `PendingPoint::Listener`, duplicating common code on lines 651-706 for each branch. This way there will be no ambiguity whether we are handling `Dialer` or `Listener` related errors. We'll also need to duplicate code in https://github.com/libp2p/rust-libp2p/blob/a4110a2b6939d5b460f11df6dd158308f517becf/swarm/src/connection/error.rs for `PendingInboundConnectionError` & `PendingOutboundConnectionError`, which won't have the common base after these changes.

I can't weigh the code duplication vs panicking in impossible branch in this case. Should I proceed in the way described above @mxinden?

mxinden: > with two big arms for `PendingPoint::Dialer` & `PendingPoint::Listener`, duplicating common code on lines 651-706 for each branch. This way there will be no ambiguity whether we are handling `Dialer` or `Listener` related errors.

I would hope that we could do without the duplication here. Though I would have to play around with it myself.

> We'll also need to duplicate code in https://github.com/libp2p/rust-libp2p/blob/a4110a2b6939d5b460f11df6dd158308f517becf/swarm/src/connection/error.rs for `PendingInboundConnectionError` & `PendingOutboundConnectionError`, which won't have the common base after these changes.

:+1: Off the top of my head, that seems reasonable to me.

---

Note that some of this might change in the near future with https://github.com/libp2p/rust-libp2p/issues/2824. One potential implementation may be https://github.com/libp2p/rust-libp2p/pull/2828, though as you can see that is still work in progress.

dmitry-markin: Should I wait for https://github.com/libp2p/rust-libp2p/pull/2828 to be merged before implementing the approach proposed above?

mxinden: Yes, I think that is a good idea. :pray: 

In case you are interested in contributing in the meantime, many of the [help wanted](https://github.com/libp2p/rust-libp2p/issues?q=is%3Aopen+is%3Aissue+label%3A%22help+wanted%22) issues don't conflict.

thomaseizinger: https://github.com/libp2p/rust-libp2p/issues/2824 is now resolved although with a different approach than in #2828 in case you want to revive this PR @dmitry-markin.

jxs:
thanks @dmitry-markin, superseeded by https://github.com/libp2p/rust-libp2p/pull/5861

Issue #2217, opened 2021-09-07
Tracking issue for renaming symbols across the repository


thomaseizinger:
Discussed in https://github.com/libp2p/rust-libp2p/discussions/2174.

- [x] https://github.com/libp2p/rust-libp2p/pull/2215

thomaseizinger:
I followed the same approach outlined here in https://github.com/libp2p/rust-libp2p/pull/2859 and introduced top-level `dummy` and `keep_alive` modules in `libp2p-swarm`. I think eventually, it would be great if we can also introduce these top-level modules in `libp2p` and re-export these kind of modules there. That would allow us to group things like `dummy::Behaviour` and `dummy::ConnectionHandler` together under `libp2p::dummy` which would be consistent with the re-export of sub-crates like `libp2p::ping` or `libp2p::kad`.

momoshell:
@jxs @thomaseizinger Hey guys, can I help with the UDS, if that one is still needed?

jxs:
hi @momoshell thanks for the interest, yeah, go for it, I think it's only the `UdsConfig` type

momoshell:
@jxs and done! ✅

> hi [@momoshell](https://github.com/momoshell) thanks for the interest...

My pleasure!
Since I've been using this library in production for a few years now, it's only natural to give something back.
Will be jumping on other issues, if you guys let me, that is 😊

Issue #897, opened 2019-01-29
RFC: inlining


jamesray1:
This library has lots of inline attributes on functions. It seems like the pattern is that they are added when a function is only a few lines. However, from https://internals.rust-lang.org/t/when-should-i-use-inline/598/8 it seems like it is better to not use it in a library, let the compiler decide, and let library users use link-time optimisation if they want to have a super-fast binary. Any thoughts?

tomaka:
I didn't comment on this earlier because the answer is that I don't know.

As far as I know, putting `#[inline]` makes it possible to inline function calls across crates by embedding the source code of the inlined function within the `.rlib`.

The counter-argument is that link-time optimisation does the same anyway, except at a different step of the compilation, and that therefore we don't need to inline across crates.

jamesray1:
Closing as it seems we agree that inlining isn't needed.

tomaka:
Well, I don't know if what I said it's true.
The only real answer would be to experiment with/without `#[inline]` and see the consequence on the bytecode.

romanb:
I think we do have a rough consensus by now that explicit `#[inline]` should be the exception in cases where benchmarking confirmed a bottleneck or hot code path that should be optimised through inlining across crates even in cases without link-time optimisation.

PR #2797, opened 2022-08-03
core/muxing: Generalise `StreamMuxer::poll_address_change` to `poll`

thomaseizinger:
# Description

This is to allow general-purpose background work to be performed
by implementations.

https://github.com/libp2p/rust-libp2p/issues/2722

## Change checklist


core/CHANGELOG.md
```diff
@@ -5,12 +5,14 @@
 - Drop `Unpin` requirement from `SubstreamBox`. See [PR 2762] and [PR 2776].
 - Drop `Sync` requirement on `StreamMuxer` for constructing `StreamMuxerBox`. See [PR 2775].
 - Use `Pin<&mut Self>` as the receiver type for all `StreamMuxer` poll functions. See [PR 2765].
+- Generalise `StreamMuxer::poll_address_change` to `StreamMuxer::poll_event`. See [PR XXXX].
```
thomaseizinger: This is a weird changelog entry because we haven't even released the top one yet but we are re-introducing a function here that was removed in a PR that is contained in the same release. How would you like to handle this @mxinden?

melekes (approved):
👍

core/src/muxing.rs
```diff
@@ -105,6 +97,22 @@ pub trait StreamMuxer {
     /// >           properly informing the remote, there is no difference between this and
     /// >           immediately dropping the muxer.
     fn poll_close(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>>;
+
+    /// Poll for an event of the underlying connection.
+    ///
+    /// In addition to returning an event, this function may be used to perform any kind of background
+    /// work that needs to happen for the muxer to do its work. Implementations can rely on this
+    /// function to be called regularly and unconditionally.
+    fn poll_event(
```
elenaf9: Not a strong opinion, but what do you think about just calling it `poll`?
- It emphasizes that this method can / should be used to drive background work
- It's consistent with `Swarm::poll`, `Transport::poll`, etc. which all serve the same functionality (driving background work and returning events that may happen)

elenaf9 (approved):
👍

core/src/muxing.rs
mxinden: With the additional functionality (i.e. _do generic work_) I find `poll` more intuitive than `poll_event` as well.

core/CHANGELOG.md
mxinden: I would suggest we update the entry introduced in https://github.com/libp2p/rust-libp2p/pull/2724/.

core/src/muxing.rs
thomaseizinger: I am not strongly attached to `poll_event`. I decided for it to have consistency among the functions in `StreamMuxer`. Currently all of them have a qualifier that expresses, what they are polling for. If it is just called `poll` that is lost.

Maybe the asymmetry is good because the functions shouldn't be treated the same way? Thoughts?

thomaseizinger: Maybe the docs also need some rewriting. I think the fact that this function can be used to do background work is more a conclusion rather than a requirement. What needs to be documented well is that this function will be called unconditionally, whereas all the others may not be.

thomaseizinger: > Not a strong opinion, but what do you think about just calling it `poll`?
> 
>  * It emphasizes that this method can / should be used to drive background work

Not that I necessarily disagree but how does it emphasize that more than the current name?

>  * It's consistent with `Swarm::poll`, `Transport::poll`, etc. which all serve the same functionality (driving background work and returning events that may happen)

All other interfaces also don't have any other `poll` functions so I am not sure that is such a good comparison.

elenaf9: > > Not a strong opinion, but what do you think about just calling it `poll`?
> > 
> > * It emphasizes that this method can / should be used to drive background work
> 
> Not that I necessarily disagree but how does it emphasize that more than the current name?
> 
> > * It's consistent with `Swarm::poll`, `Transport::poll`, etc. which all serve the same functionality (driving background work and returning events that may happen)
> 
> All other interfaces also don't have any other `poll` functions so I am not sure that is such a good comparison.

I read `poll_<xyz>` always as "poll for xyz", as in "poll this method if you are interested in xyz". So callers may call this method, but if they are not interested in xyz they don't have to.
`poll` on the other hand (at least how we use it in `Swarm`, `Transport`) is used to drive background work. Even if the user is not interested in the produced events, they still must call it continuously.

thomaseizinger: Hmm, that does make sense. Let's go with `poll` then :)

core/src/muxing.rs
```diff
@@ -86,25 +90,33 @@ pub trait StreamMuxer {
...
     ///
     /// > **Note**: You are encouraged to call this method and wait for it to return `Ready`, so
     /// >           that the remote is properly informed of the shutdown. However, apart from
     /// >           properly informing the remote, there is no difference between this and
     /// >           immediately dropping the muxer.
     fn poll_close(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>>;
+
+    /// Poll to allow the underlying connection to make progress.
+    ///
+    /// In contrast to all other `poll`-functions on [`StreamMuxer`], this function must be called
```
thomaseizinger:
```suggestion
    /// In contrast to all other `poll`-functions on [`StreamMuxer`], this function MUST be called
```

core/src/muxing.rs
```diff
@@ -86,25 +90,33 @@ pub trait StreamMuxer {
...
+    /// implementations to return events about the underlying connection that the caller MUST deal
+    /// with.
+    fn poll(
+        self: Pin<&mut Self>,
+        cx: &mut Context<'_>,
+    ) -> Poll<Result<StreamMuxerEvent, Self::Error>>;
+}
+
+/// An event produced by a [`StreamMuxer`].
+pub enum StreamMuxerEvent {
```
thomaseizinger, line 117: Any thoughts on just naming this `Event` and referring to is as `muxing::Event`?

I think there was a loose consensus around https://github.com/libp2p/rust-libp2p/issues/2217 at some point but we haven't really made progress on this front.

elenaf9: 👍 Sounds good to me.

core/src/muxing/boxed.rs
```diff
@@ -65,14 +59,16 @@ where
             .map_err(into_io_error)
     }
 
-    fn poll_address_change(
+    #[inline]
```
elenaf9, line 62: Why inline this method but not the other ones, and why only inline it in this muxer? Just asking out of curiosity because I am not really familiar with `#[inline]`.

elenaf9 (approved):
LGTM!

core/src/muxing/boxed.rs
thomaseizinger: Not on purpose to be honest. I thought I left everything as it was before.

I am not too familiar with inlining either but the rough advice I got was that the compiler tends to be smarter on when it is needed :)

core/src/muxing.rs
mxinden: > I read `poll_<xyz>` always as "poll for xyz", as in "poll this method if you are interested in xyz". So callers may call this method, but if they are not interested in xyz they don't have to.
> `poll` on the other hand (at least how we use it in `Swarm`, `Transport`) is used to drive background work. Even if the user is not interested in the produced events, they still must call it continuously.

Same mindset for me. Well putten into words.

core/src/muxing.rs
mxinden: I am in favor of just `Event` as long as we don't do `use muxing::Event` in other files, but instead just `use muxing` and then refer to it as `muxing::Event`.

core/src/muxing/boxed.rs
mxinden: I think the many `#[inline]` attributes are from past premature optimizations. See https://github.com/libp2p/rust-libp2p/issues/897.

> the rough advice I got was that the compiler tends to be smarter on when it is needed :)

Yes. Unless proven through a benchmark, let the compiler make the decision and thus don't use `#[inline]`.

mxinden: > Not on purpose to be honest. I thought I left everything as it was before.

The method was marked as `#[inline]` before this patch as well. I am fine with either removing it here or in a pull request in the future.

mxinden:
If I am not mistaken this only requires the rename from `StreamMuxerEvent` to `Event`, right? 

https://github.com/libp2p/rust-libp2p/pull/2797#discussion_r941110145

Also happy for this to happen in a different pull request if you prefer @thomaseizinger.

(Sorry for the merge, thought this was already ready to go.)

thomaseizinger:
> If I am not mistaken this only requires the rename from `StreamMuxerEvent` to `Event`, right? 
> 
> https://github.com/libp2p/rust-libp2p/pull/2797#discussion_r941110145
> 
> Also happy for this to happen in a different pull request if you prefer @thomaseizinger.
> 
> (Sorry for the merge, thought this was already ready to go.)

Yes. I am also happy for this to happen in a different PR :)

mxinden (approved):

Issue #2650, opened 2022-05-16
Drop custom `Either` types


thomaseizinger:
## Description

We have a lot of custom `Either` types within `libp2p-core`, some of which (I think) could go away by implementing our traits directly on the `Either` type that we are already depending on.

## Motivation

Less code to maintain.

## Checklist

- [x] `EitherTransport`: https://github.com/libp2p/rust-libp2p/pull/3338
- [x] `EitherError`: https://github.com/libp2p/rust-libp2p/pull/3337
- [x] `EitherOutput`: https://github.com/libp2p/rust-libp2p/pull/3341
- [x] `EitherUpgrade`: https://github.com/libp2p/rust-libp2p/pull/3339
- [x] `EitherName`: Cannot be removed at the moment because it would conflict with the blanket implementation of `ProtocolName` over `T`. This will however be resolved as soon as https://github.com/libp2p/rust-libp2p/issues/2831 is fixed.
- [ ] `EitherFuture`:
	- In the current state of the codebase, we cannot remove this because we need an implementation of `Future` that returns `Either` and does not force the same `Output` on both types.
	- We can remove some of the usages with https://github.com/libp2p/rust-libp2p/issues/3347.
	- The other usage is within `SelectUpgrade`. That one I believe we cannot remove because `SelectUpgrade` is for example used to choose between two muxers which is functionality we definitely want. However, in order to make https://github.com/libp2p/rust-libp2p/issues/3271 easier, we can at least move this dedicated `EitherFuture` type into the same crate as `SelectUpgrade` because it has only one user.
- [x] `EitherIter`: Cannot be removed because `Either` does not implement `IntoIterator`. If we still need this after https://github.com/libp2p/rust-libp2p/issues/2831 is merged, we can request it upstream.

## Are you planning to do it yourself in a pull request?

No / Maybe. Could be a good first issue. What do you think @mxinden?

mxinden:
Sounds good to me. Also, should we be using `either::Either` or `futures::future::Either` or both?

thomaseizinger:
> Sounds good to me. Also, should we be using `either::Either` or `futures::future::Either` or both?

I am not sure how well maintained `either::Either` is but feels like the more appropriate one, given that `Future` is part of std now?

I've had a brief look and for example the `Error` impl in `either::Either` is not overriding all methods so it may need a bit of upstream work until we can completely replace our types.

[...]

PR #2798, opened 2022-08-04
core/either: Remove `EitherName` in favor of `Either`

thomaseizinger:
# Description

#2650

## Change checklist


core/CHANGELOG.md
```diff
@@ -5,12 +5,14 @@
...
 - Drop `Sync` requirement on `StreamMuxer` for constructing `StreamMuxerBox`. See [PR 2775].
 - Use `Pin<&mut Self>` as the receiver type for all `StreamMuxer` poll functions. See [PR 2765].
+- Remove `EitherName` in favor of `Either` from the `either` crate. See [PR XXXX].
 
 [PR 2724]: https://github.com/libp2p/rust-libp2p/pull/2724
 [PR 2762]: https://github.com/libp2p/rust-libp2p/pull/2762
 [PR 2775]: https://github.com/libp2p/rust-libp2p/pull/2775
 [PR 2776]: https://github.com/libp2p/rust-libp2p/pull/2776
 [PR 2765]: https://github.com/libp2p/rust-libp2p/pull/2765
+[PR XXXX]: https://github.com/libp2p/rust-libp2p/pull/XXXX
```
```suggestion
- Remove `EitherName` in favor of `Either` from the `either` crate. See [PR 2798].

[PR 2724]: https://github.com/libp2p/rust-libp2p/pull/2724
[PR 2762]: https://github.com/libp2p/rust-libp2p/pull/2762
[PR 2775]: https://github.com/libp2p/rust-libp2p/pull/2775
[PR 2776]: https://github.com/libp2p/rust-libp2p/pull/2776
[PR 2765]: https://github.com/libp2p/rust-libp2p/pull/2765
[PR 2798]: https://github.com/libp2p/rust-libp2p/pull/2798
```

mxinden:
> This does not compile because the blanket impl for `ProtocolName` over anything that implements `AsRef<u8>` is conflicting with the implementation on `Either`.

I don't have a strong opinion on this. Given the complication above I am inclined to just leave it as is. Any additional benefits of this patch beyond consistency?

thomaseizinger:
> > This does not compile because the blanket impl for `ProtocolName` over anything that implements `AsRef<u8>` is conflicting with the implementation on `Either`.
> 
> I don't have a strong opinion on this. Given the complication above I am inclined to just leave it as is. Any additional benefits of this patch beyond consistency?

I'll give it a bit more thought but I don't think so.

Instead of just adding impls that we need, I can see to refactor some code to streamline, how we generate `ProtocolName`s.

It is a bit weird that we allow so many things to be `ProtocolName`s and on the trait, we define two requirements that aren't really enforced:

- Max length of 140
- Must start with `/`

Maybe we should introduce a `ProtocolName` newtype?

mxinden:
On the meta level, I think we could do better modelling protocol names.

Part of the problem is the conflict between multistream select and libp2p-identify on the specification level. One accepts a `[u8]`, the other expects a `String`.

> Maybe we should introduce a ProtocolName newtype?

Or a unit type for each protocol name, each implementing `ProtocolName`.

thomaseizinger:
> On the meta level, I think we could do better modelling protocol names.
> 
> Part of the problem is the conflict between multistream select and libp2p-identify on the specification level. One accepts a `[u8]`, the other expects a `String`.

Damn. Of course it wasn't that easy to fix :)

> > Maybe we should introduce a ProtocolName newtype?
> 
> Or a unit type for each protocol name, each implementing `ProtocolName`.

Nice idea but it won't allow us to enforce invariants, right?

mxinden:
> Nice idea but it won't allow us to enforce invariants, right?

True. Would be great to enforce these at runtime. Would be wonderful to enforce these at compile time :innocent:

thomaseizinger:
> > Nice idea but it won't allow us to enforce invariants, right?
> 
> True. Would be great to enforce these at runtime. Would be wonderful to enforce these at compile time :innocent: 

Something for a future issue? I have some ideas :)

To progress on this issue, would it be okay to transition all/most usages to std types like `&static [u8]` and `String`?

mxinden:
> > > Nice idea but it won't allow us to enforce invariants, right?
> > 
> > 
> > True. Would be great to enforce these at runtime. Would be wonderful to enforce these at compile time innocent
> 
> Something for a future issue? I have some ideas :)

:+1: 

> To progress on this issue, would it be okay to transition all/most usages to std types like `&static [u8]` and `String`?

Given that `Either` also implements `AsRef<[u8]>` when the two subtypes implement `AsRef<[u8]>`, is the implementation of `ProtocolName` on `Either` still needed?

https://docs.rs/either/latest/either/enum.Either.html#impl-AsRef%3C%5BTarget%5D%3E

thomaseizinger:
> > To progress on this issue, would it be okay to transition all/most usages to std types like `&static [u8]` and `String`?
> 
> Given that `Either` also implements `AsRef<[u8]>` when the two subtypes implement `AsRef<[u8]>`, is the implementation of `ProtocolName` on `Either` still needed?
> 
> https://docs.rs/either/latest/either/enum.Either.html#impl-AsRef%3C%5BTarget%5D%3E

(Un)fortunately yes. I remember trying this and it is not working.

I think the reason is in Rust's explicit strong-typing system. Just because something implements all super traits of a trait A does not mean it implements trait A by itself unless it is actively declared as such. In Golang, this would work :)

mxinden:
> To progress on this issue, would it be okay to transition all/most usages to std types like `&static [u8]` and `String`?

I am fine with that.

core/CHANGELOG.md
```diff
@@ -6,13 +6,16 @@
 - Change `StreamMuxer` interface to be entirely poll-based. All functions on `StreamMuxer` now
   require a `Context` and return `Poll`. This gives callers fine-grained control over what they
   would like to make progress on. See [PR 2724] and [PR 2797].
+- Remove `EitherName` in favor of `Either` from the `either` crate. See [PR 2798].
+- Replace blanket `ProtocolName` impl with specific ones on `&'static [u8]` and `Cow<'static, [u8]>`. See [PR 2798].
```
thomaseizinger, line 10: We could split this out into a different PR if you want.

mxinden: I am fine with a single pull request only.

core/src/upgrade.rs
```diff
@@ -126,12 +128,28 @@ pub trait ProtocolName {
...
     }
 }
 
+impl<A, B> ProtocolName for Either<A, B>
+where
+    A: ProtocolName,
+    B: ProtocolName,
+{
+    fn protocol_name(&self) -> &[u8] {
+        ::either::for_both!(self, inner => inner.protocol_name())
```
mxinden, line 149: Neat!

core/CHANGELOG.md
mxinden, line 10: Why not implement `ProtocolName` for `String` and `str`?

mxinden (commented):
I would actually suggest making `String` the default protocol representation. See https://github.com/libp2p/rust-libp2p/issues/2831. What do you think @thomaseizinger?

thomaseizinger:
> I would actually suggest making `String` the default protocol representation. See #2831. What do you think @thomaseizinger?

Yeah we can move to string! Should we do that as part of this PR or make another one? It is getting a bit OT from the original goal haha.

I am happy to fix #2831 first and then come back to this!

mxinden:
I think the following order makes most sense:

1. #2831 
2. Move `ProtocolName` to `String`
3. Do the actual change here, namely to use `Either`

That said, I don't think we have to stick with that order.

thomaseizinger:
> I think the following order makes most sense:
> 
>     1. [misc/multistream-select: Treat protocol as `String` and not `[u8]` #2831](https://github.com/libp2p/rust-libp2p/issues/2831)
> 
>     2. Move `ProtocolName` to `String`
> 
>     3. Do the actual change here, namely to use `Either`
> 
> 
> That said, I don't think we have to stick with that order.

I agree, leave it with me 😊

thomaseizinger:
Closing in favor of #2966.

PR #2799, opened 2022-08-05
README: Point to security@libp2p.io

mxinden:
libp2p now has its own security mail address, namely security@libp2p.io.

No need to rely on security@ipfs.io any longer.


thomaseizinger (approved):

elenaf9 (changes requested):
Nice! The bug-report template also contains the security e-mail and needs to be changed.
https://github.com/libp2p/rust-libp2p/blob/028decec6913f33b49bffb8cbfa7202629b2c22f/.github/ISSUE_TEMPLATE/bug_report.md?plain=1#L8

mxinden:
Good catch. Thanks @elenaf9.

elenaf9 (approved):

PR #2802, opened 2022-08-06
protocols/: Remove passing default variant to `WithPeerId::condition`

K0UR05H:
# Description
Since `PeerCondition::Disconnected` is the default variant and is used [here](https://github.com/libp2p/rust-libp2p/blob/028decec6913f33b49bffb8cbfa7202629b2c22f/swarm/src/dial_opts.rs#L54), there is no need to pass it again to `WithPeerId::condition`.

## Change checklist


elenaf9:
Thank you for contributing @k0ur0x!
I am not sure whether it's a good idea to remove the explicit `.condition(dial_opts::PeerCondition::Disconnected)`. Being explicit about things imo generally makes maintaining the code easier. It saves trouble in the future when you look back at some code wondering "Was using the default value a conscious choice?". Especially if we ever decided to change the default value.

mxinden (approved):
Thanks for the contribution @k0ur0x.

I don't have a strong opinion here.

I think the choice of using `PeerCondition::Disconnected` is reasonably obvious in each of these cases, thus I don't think there is any need to be explicit and thus I think we can just depend on the default behavior as proposed in this pull request.

elenaf9 (approved):
> I don't have a strong opinion here.
> 
> I think the choice of using PeerCondition::Disconnected is reasonably obvious in each of these cases, thus I don't think there is any need to be explicit and thus I think we can just depend on the default behavior as proposed in this pull request.

Makes sense. No strong opinion on my side either.

PR #2803, opened 2022-08-06
.github/workflows: Split advisory issues from PR workflows using `cargo-deny`

pinkforest:
# Description

actions-rs/check-audit is not maintained

https://github.com/rust-secure-code/wg/issues/46

## Open Questions

It is currently complaining about `owning_ref` - we added advisory few days ago

Please review to make sure this is addressed e.g. potentially supressing / migration etc.

I've also added this into yaml:

```yaml
    # Prevent sudden announcement of a new advisory from failing ci:
    continue-on-error: ${{ matrix.checks == 'advisories' }}
```

### Note: Please review licenses
Cargo deny does license check and I had to add the `ring` crate specific license and Unicode `unicode-ident` as special cases

## Change checklist

- [X] A changelog entry has been made in the appropriate crates - N/A


deny.toml
```diff
@@ -0,0 +1,144 @@
...
+# not have its license(s) checked
+#registries = [
+#]
+
+# This section is considered when running `cargo deny check bans`.
+# More documentation about the 'bans' section can be found here:
+# https://embarkstudios.github.io/cargo-deny/checks/bans/cfg.html
+[bans]
+# Lint level for when multiple versions of the same crate are detected
+multiple-versions = "warn"
```
thomaseizinger, line 106: I'd love to ban this actually but I think that is gonna be a fair bit of work.

thomaseizinger (approved):
I am in favor of this.

thomaseizinger:
The actual issue is already being addressed: https://github.com/prometheus/client_rust/pull/78

thomaseizinger:
Friendly note @pinkforest : We squash merge here and favor merges over force-pushes. Thank you :)

pinkforest:
Oh that was just github's catch-up rebase, didn't want to bother with merge from main commits noise - but noted thanks :)

p.s. needs CONTRIBUTING.md :woman_shrugging: 

EDIT: p.p.s. a CONTRIBUTING.md I eventually found - https://github.com/ipfs/community/blob/master/CONTRIBUTING.md#rebasing instructs `push -f`

.github/workflows/cargo-audit.yml
```diff
@@ -11,10 +11,19 @@ on:
...
+  cargo-deny:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        checks:
+          - advisories
+          - bans licenses sources
+
+    # Prevent sudden announcement of a new advisory from failing ci:
+    continue-on-error: ${{ matrix.checks == 'advisories' }}
```
mxinden: I don't read the warnings whenever the overall run succeeds. Thus I would have missed https://github.com/libp2p/rust-libp2p/issues/2794 right? How about failing instead? I know it is more noise, but I would argue it is worth it.

deny.toml
```diff
@@ -0,0 +1,144 @@
+# This section is considered when running `cargo deny check advisories`
```
mxinden, line 1: Is this mostly the default? How do we keep this updated when new options are added upstream?

deny.toml
mxinden: Out of curiosity, where are we at today?

I think the best we can do is be a good citizen and merge @dependabot pull requests quickly.

mxinden (commented):
Thanks @pinkforest and @thomaseizinger for the help on this!

.github/workflows/cargo-audit.yml
pinkforest: Hm. I was going to ask whether we should keep the cron job and it should be kept separate and PR still has checks for licenses and other things - the cron - a separate workflow action - would still raise separately Issue about any advisories - or do we want to keep the CI failing in this case?

deny.toml
pinkforest: These rarely / never change as it's type of check. If there are new types of check then it is a feature and dependabot will pester about upgrading the action and it is then when you can discover any new type of checks - 

Nonetheless RustSec advisory categories may change but it is covered under advisories regardless.

pinkforest:
multiple-versions - I didn't get any warnings but there are some weird SemVer stuff that sometimes it's impossible to deal with so it's best to leave as warn - reality of dealing with rust ecosystem.

I will send a commit around getting the separate cron-check for advisories and will separate it from the PR CI workflow

pinkforest:
I just switched the PR to be different flow with deny from periodic one that uses audit

The periodic will catch any advisories and beep about them as usual and the PR / push will catch license / shady dependencies

.github/workflows/cargo-audit.yml
pinkforest: So I changed now that the PR is now on cargo deny upon PR and will not fail straight from advisories whilst there is still periodic cron job to raise any advisory failures separately via audit - the periodic schedule can be always changed

mxinden:
@pinkforest why do we need both `cargo audit` and `cargo deny`?

pinkforest:
Audit is for the advisories on scanning periodically. 

Deny is for the PR scanning which includes license checks.

As you can see other is on PR / push and the other one is for cron basis.

It's better to use both.

deny.toml
thomaseizinger: > Out of curiosity, where are we at today?

Something like 20 duplicates I think.

> I think the best we can do is be a good citizen and merge @dependabot pull requests quickly.

We can do better and regularly check for duplicate dependencies and investigate, where they come from. We can also audit, which dependencies we have and try to get rid of some. Another way is to audit for set but unused features or entire dependencies altogether.

For example, we pull in an old version of rand (0.7) via `ed25519-dalek`.

Several of the windows specific dependencies are also duplicates I think.

thomaseizinger:
> @pinkforest why do we need both `cargo audit` and `cargo deny`?

`cargo deny` doesn't open issues I think although we could potentially script that with the pre-installed `gh` CLI.

thomaseizinger:
> I just switched the PR to be different flow with deny from periodic one that uses audit
> 
> The periodic will catch any advisories and beep about them as usual and the PR / push will catch license / shady dependencies

Thanks!

Can we split this PR into two then? Changing cargo audit to only run on a schedule and using cargo deny are two different things IMO.

I think the latter is pretty uncontroversial whereas the former may need a bit of discussion.

I am okay with moving away from a PR-based audit with `cargo audit`. We will still get the open issue and can prioritise work to fix them whilst being able to merge other PRs.

For `cargo deny`, we could remove the `continue-on-error` config and just not mark it as mandatory for merge.

pinkforest:
I am not changing your existing cargo audit except modifying it to exclude PR pushes which is being switched to cargo deny.

So I don't think there is much to gain to separate PRs at this point.

Happy to take the continue-on-error out.

deny.toml
pinkforest: ed25519-dalek is not maintained.
and there is another concern about it which doesn't break anything but it's not good either
https://github.com/rustsec/advisory-db/issues/1360
I don't think any of this stops this PR tho ?

.github/workflows/cargo-deny-pr.yml
```diff
@@ -0,0 +1,27 @@
...
+  cargo-deny:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        checks:
+          - advisories
+          - bans licenses sources
+
+    # Prevent sudden announcement of a new advisory from failing ci:
+    continue-on-error: ${{ matrix.checks == 'advisories' }}
```
mxinden: I suggest to remove this.

mxinden (commented):
One comment. Otherwise looks good to me. In the future, agreed with Thomas that separate pull requests are easier.

.github/workflows/cargo-deny-pr.yml
pinkforest: Removed

.github/workflows/cargo-deny-pr.yml
```diff
@@ -0,0 +1,24 @@
...
+      - '**/Cargo.toml'
+      - '**/Cargo.lock'
+jobs:
+  cargo-deny:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        checks:
+          - advisories
+          - bans licenses sources
```
mxinden, line 16: Given that we no longer treat `advisories` special, is the matrix still needed? Also do we need to list the commands any longer at all? Wouldn't a `cargo deny check` suffice to run them all?

pinkforest: I personally prefer to be explicit as if there is change in the future you can gradually adopt more check features.

If the preference is to be implicit non-exhaustive then we can change this but prepare for surprise(s) if the tooling changes.

Also I personally prefer matrix to separate the workflow streams so I can catch / view errors in context e.g. I want to have different mindset when dealing with licenses vs advisories - plus it's faster as silver lining.

If the preference is to bundle them to same flow then that can be certainly changed.

Let me know and I can change.

deny.toml
thomaseizinger: It doesn't no, sorry for the OT.

.github/workflows/cargo-deny-pr.yml
thomaseizinger: I'd like to keep it separate actually so we can temporarily ignore a failing advisory check for example. It also allows us to mark some of them as required via GH for merging and leave others optional.

.github/workflows/cargo-deny-pr.yml
```diff
@@ -0,0 +1,24 @@
+name: cargo deny
+on:
+  push:
+    paths:
+      - '**/Cargo.toml'
+      - '**/Cargo.lock'
```
thomaseizinger: We don't have this checked into Git so I think it doesn't do anything?

pinkforest: Correct - Not at the moment no, but I would recommend that we should put lock into Git after this so it can be used to quick screen every new update that comes through via dependabot and also that means that you can easily continuously test all updates what the downstream consumers are actually using via libp2p as transient dep which also means that the --locked version is tested. It also is useful in changelog to see if e.g. a library that was implementation detail that went up a SemVer or especially non-implementation detail which might affect pub APIs.

thomaseizinger: We are a library though so I am not sure what benefit we are getting from checking in `Cargo.lock`?

On the contrary, I'd say it is better to not pin specific versions. A new user will get the same dependencies as our CI runs if they add `rust-libp2p` to their `Cargo.toml`, meaning in case the builds breaks due to a bad patch semver upgrade f.e., we should know earlier than if we were to check in a `Cargo.lock`.

Upgrades outside of the allowed semver range are performed by dependabot and if they contain publicly exposed types, we already note these in our changelog. (At the same time, we try to limit these cases to avoid unnecessary breaking changes for our users.)

To get back on topic, I think we should remove these triggers because they don't do anything for us and are effectively dead configuration.

thomaseizinger:
@mxinden @pinkforest I tried to summarize what this PR does in a new title. Improvements welcome!

.github/workflows/cargo-deny-pr.yml
pinkforest: Cargo.lock ensures that you test with all locked versions - as a library - and be aware of changes from downstream that the library still works for anyone given Cargo.lock - think your Cargo.lock as some manifest that you've tested and you're actively supporting.

It also ensures that you get security advisories that are triggered from Cargo.lock and not from Cargo.toml which may not reflect the full reality of the usage - .lock also shows all the conflciting versions e.g. if you use time 0.1 due to using default feature-set f.ex. that would be easy to fix once it's visible.

thomaseizinger: I am not quite sure I follow.

Assume a security advisory in some crate `foo` that is vulnerable in version: `<= 0.1.10` and `rust-libp2p` specifies `0.1.10`. Let there be a security fix in `0.1.11`. With the requirement `0.1.10`, we only express that we need _at least_ `0.1.10` but any greater patch version is fine too. Thus any binary can freely update to `0.1.11` on their end. I don't see how `rust-libp2p` is vulnerable here or needs any change to its manifest. Any security advisory issue opened on our project is basically a "can't fix" situation.

Now, in the time between the advisory being published and the fix being available, `cargo audit` will still open an issue on our repository but there isn't a whole lot we can do. I don't see how checking in a `Cargo.lock` file improves anything in this situation. Instead of not having to do anything, we need to explicitly bump the version in a PR that will likely only touch the lockfile: something that is irrelevant for our users. (To my knowledge, `cargo` ignores lockfiles in dependencies.)

On a more general note: I don't think it is good practice to only support a locked range of versions. On the contrary, we should strive to be compatible with any version combination that is possible given our `Cargo.toml`. I don't see how a versioned lockfile is helping us with that.

pinkforest: .lock would make sure that you always test with locked set of dependencies and ensures that every time there is immediate or transitive (not just transitive) dependency update you know that your software .. or library.. still works.. instead of users (of the library) reporting it - despite this not (always) leaving with you any control at least you are aware of it and can possibility mitigate e.g. if crate did a breaking change where they didn't bump semver - this happens quite a bit ... esp on direct dependencies as .toml would just shift to the latest version but Cargo.lock would give granularity to immediate deps too. 

Yeah sure you can still know your library doesn't work with only .toml but there is always a lot of udpates between and then taking RCA takes longer and it's messier IMO

e.g. TL;DR Even if cargo ignores the dependencies you get the granular visibility and notice breaking changes instead of waiting someone to report it when one dependency went haywire - and you can isolate it quick from dependabot PR's on their own CI runs that bump the bumps in organised manner.

Something to ponder about.

pinkforest: I basically - even as a library - screen / gate all .lock bumps to make sure everything still works. It's additional chore but imo worth it when I just do @dependabot merge for all of them and gives great visibility around what is going on.

.lock also tends to make new crates visible in the dependency tree and you get notification that hey there is some new crate we are now dependening on .. and goes via PR so it's informed and organised bump

pinkforest: Anyways I addressed your .lock out so can always add that back if you choose to add Cargo.lock

thomaseizinger: Thanks for the input! Definitely something to consider for the future. I'll open an issue to track it as a separate discussion!

thomaseizinger (approved):
LGTM! Thanks for the contribution and sorry for the delay in getting this reviewed1

thomaseizinger:
@mxinden Feel free to merge if you also agree with the changes.

mxinden (approved):
:pray:

PR #2805, opened 2022-08-08
*: Remove `inject_connected` / `inject_disconnected` from docs

K0UR05H:
# Description
Both `inject_connected` and `inject_disconnected` are removed in dc8433e3.

## Change checklist


swarm/src/behaviour.rs
```diff
@@ -230,7 +230,7 @@ pub trait NetworkBehaviour: 'static {
     /// Informs the behaviour about an event generated by the handler dedicated to the peer identified by `peer_id`.
     /// for the behaviour.
     ///
-    /// The `peer_id` is guaranteed to be in a connected state. In other words, `inject_connected`
+    /// The `peer_id` is guaranteed to be in a connected state. In other words, `inject_connection_established`
```
thomaseizinger: Lets change this and other rustdoc appearances to an actual reference with `[``]`.

That will yield a compile-error if we ever hit this again.

thomaseizinger (commented):
Thank you!

One suggestion.

swarm/src/behaviour.rs
K0UR05H: Thanks for your review. I changed them.

mxinden (commented):
Thanks for the help here. Please wrap the lines at 80 or 120 characters. Otherwise, this looks good to me.

mxinden (approved):
:pray: Thank you for the help!

PR #402, opened 2018-08-07
Use tk_listen to absorb errors on the TcpListener

tomaka:
[The documentation of TcpListener](https://docs.rs/tokio-tcp/0.1.1/tokio_tcp/struct.TcpListener.html#errors) suggests doing what this PR does.
The `TcpListener` can sometimes produce an error which would just close the listener.

I think handling this directly in the TCP transport is the right location. There might be valid reasons why a listener produces an error, therefore the upper layers should continue to drop the listener if an error happens.

PR #2813, opened 2022-08-11
transports/tcp: simplify IfWatcher integration

elenaf9:
# Description

Update to `if-watch` version `v2.0.0`, which allows us to simplify our integration in the tcp transport quite a bit:
- Since `IfWatcher::new` is not async anymore we do not need the `IfWatch` wrapper and related logic anymore
- Use `if_watch::IfWatcher` for both runtimes. According to this discussion, this should be ok now: https://github.com/libp2p/rust-libp2p/pull/2289#discussion_r917523292. 
However, I just saw now that for `cfg(not(any(target_os = "ios", target_os = "linux", target_os = "macos", target_os = "windows")))` `if-watch` still uses `async-io`. Given that this won't apply for the vast majority of users I'd say this is acceptable. Alternative is to do https://github.com/mxinden/if-watch/issues/21 before continuing here.
- Remove `InAddr` and instead manage the `IfWatcher` directly in the `TcpListenStream`. In case of `InAddr::One` directly push an `TransportEvent::NewAddress` into our event queue when creating the listener.

Related discussions: https://github.com/libp2p/rust-libp2p/pull/2289#discussion_r917403346.

## Open Questions

- <s>We could remove `InAddr` completely and in case of `InAddr::One` just directly push an `TransportEvent::NewAddress` into our event queue when creating the listener. See https://github.com/elenaf9/rust-libp2p/commit/c4f2124185d9de9ac0d2dd1dbeb3c919fae6bce2. But not sure if that really simplifies things. I guess for other transports that don't have whole port-reuse logic this might make more sense.</s>
Edit: Done now in 64a3165.

## Change checklist


transports/tcp/src/lib.rs
```diff
@@ -743,68 +729,46 @@ where
...
-                                    {
-                                        let ma = ip_to_multiaddr(ip, me.listen_addr.port());
-                                        log::debug!("Expired listen address: {}", ma);
-                                        me.port_reuse.unregister(ip, me.listen_addr.port());
-                                        return Poll::Ready(Some(Ok(
-                                            TcpListenerEvent::AddressExpired(ma),
-                                        )));
-                                    }
+                InAddr::Any(if_watcher) => {
+                    while let Poll::Ready(ev) = IfWatcher::poll_next(Pin::new(if_watcher), cx) {
```
thomaseizinger: You could add a `poll_next_unpin` convenience function to `IfWatcher`.

thomaseizinger: Actually, it is statically `Unpin` so we could hide the pinning internally?

thomaseizinger (commented):
This looks like a really cool simplification!

mxinden:
> * However, I just saw now that for `cfg(not(any(target_os = "ios", target_os = "linux", target_os = "macos", target_os = "windows")))` `if-watch` still uses `async-io`. Given that this won't apply for the vast majority of users I'd say this is acceptable. Alternative is to do fix

This seems fine to me.

transports/tcp/src/lib.rs
elenaf9: Good idea. Done with mxinden/if-watch#25 and a139f76.

transports/tcp/src/lib.rs
```diff
@@ -605,23 +607,12 @@ pub enum TcpListenerEvent<S> {
...
-
 /// The listening addresses of a [`TcpListenStream`].
-enum InAddr<TIfWatcher> {
+enum InAddr {
     /// The stream accepts connections on a single interface.
-    One {
-        addr: IpAddr,
-        out: Option<Multiaddr>,
-    },
+    One(Option<Multiaddr>),
```
thomaseizinger: The use of `Option` here is pretty subtle. Would it make sense to introduce a 3rd variant that represents the state of already having reported the multiaddr once? This would also allow us to use `mem::replace` so we don't need to `take` the `Multiaddr` from the `Option`.

transports/tcp/src/lib.rs
```diff
@@ -734,116 +721,88 @@ where
...
-            match &mut me.in_addr {
-                InAddr::Any { if_watch, addrs } => match if_watch {
-                    // If we listen on all interfaces, wait for `if-watch` to be ready.
-                    IfWatch::Pending(f) => match ready!(Pin::new(f).poll(cx)) {
-                        Ok(w) => {
-                            *if_watch = IfWatch::Ready(w);
-                            continue;
+        match &mut me.in_addr {
+            InAddr::Any(if_watcher) => {
+                while let Poll::Ready(ev) = if_watcher.as_mut().poll_if_event(cx) {
```
thomaseizinger: I don't think we need `as_mut` here.

```suggestion
                while let Poll::Ready(ev) = if_watcher.poll_if_event(cx) {
```

transports/tcp/src/lib.rs
```diff
@@ -734,116 +721,88 @@ where
...
-            let incoming = match T::poll_accept(&mut me.listener, cx) {
-                Poll::Pending => return Poll::Pending,
-                Poll::Ready(Ok(incoming)) => incoming,
-                Poll::Ready(Err(e)) => {
-                    // These errors are non-fatal for the listener stream.
-                    log::error!("error accepting incoming connection: {}", e);
-                    me.pause = Some(Delay::new(me.sleep_on_error));
-                    return Poll::Ready(Some(Ok(TcpListenerEvent::Error(e))));
+        if let Some(mut pause) = me.pause.take() {
+            match Pin::new(&mut pause).poll(cx) {
```
thomaseizinger: We should be able to use `poll_unpin` here from `FutureExt`.

```suggestion
            match pause.poll_unpin(cx) {
```

transports/tcp/src/lib.rs
```diff
@@ -734,116 +721,88 @@ where
...
-                remote_addr,
-            })));
-        }
+        log::debug!("Incoming connection from {} at {}", remote_addr, local_addr);
+
+        Poll::Ready(Some(Ok(TcpListenerEvent::Upgrade {
+            upgrade: future::ok(incoming.stream),
+            local_addr,
+            remote_addr,
+        })))
```
thomaseizinger: This is a bit of stylistic break from how we usually do things.

In `poll` functions with multiple sub futures, we tend to return `Poll::Pending` at the bottom and return early with `Poll::Ready` from `match` blocks.

This way, all sections of the the `poll` functions look the same.

transports/tcp/src/lib.rs
thomaseizinger: If we would use a `loop` here, there would be only one `match` that includes the `Poll` which again, would be slightly more consistent with the remaining blocks in this function.

thomaseizinger (approved):
This is a really great improvement!

I've added some stylistic comments.

thomaseizinger:
>  * We could remove `InAddr` completely and in case of `InAddr::One` just directly push an `TransportEvent::NewAddress` into our event queue when creating the listener. See [elenaf9@c4f2124](https://github.com/elenaf9/rust-libp2p/commit/c4f2124185d9de9ac0d2dd1dbeb3c919fae6bce2). But not sure if that really simplifies things. I guess for other transports that don't have whole port-reuse logic this might make more sense.

I've only just noticed now. I think that is a good idea!

thomaseizinger:
Note: I've edited your description so we don't actually close https://github.com/mxinden/if-watch/issues/21 with merging this.

transports/tcp/src/lib.rs
elenaf9: Fixed with 8d16ad7. Was that what you meant? :)

transports/tcp/src/lib.rs
elenaf9: Done with 8d16ad7936ef3fb74bd58161761a4139f9550fc5.

transports/tcp/src/lib.rs
elenaf9: Removed as suggested in https://github.com/libp2p/rust-libp2p/pull/2813#issuecomment-1220538958, hence this is obsolete.

transports/tcp/src/lib.rs
```diff
@@ -734,116 +710,84 @@ where
...
-                    if let Some(multiaddr) = out.take() {
-                        me.port_reuse.register(*addr, me.listen_addr.port());
-                        return Poll::Ready(Some(Ok(TcpListenerEvent::NewAddress(multiaddr))));
-                    }
+        if let Some(mut pause) = me.pause.take() {
+            match pause.poll_unpin(cx) {
+                Poll::Ready(_) => {}
+                Poll::Pending => {
+                    me.pause = Some(pause);
+                    return Poll::Pending;
```
elenaf9, line 733: I moved this section to the top so that the pause is polled before polling the `if_watcher`. 
If polling the `if_watcher` errored, we set a delay before returning. From the logged error messages _"Failure polling interfaces: {..}. Scheduling retry."_ I assume that the if_watcher should be polled again **after** this delay. Thus on the next iteration we should first poll the pause. Is this assumption correct?

transports/tcp/src/lib.rs
```diff
@@ -368,7 +371,20 @@ where
...
+        } else {
+            self.port_reuse.register(local_addr.ip(), local_addr.port());
+            let listen_addr = ip_to_multiaddr(local_addr.ip(), local_addr.port());
+            self.pending_events.push_back(TransportEvent::NewAddress {
+                listener_id: id,
+                listen_addr,
+            });
+            None
+        };
+        TcpListenStream::<T>::new(id, listener, if_watcher, self.port_reuse.clone())
```
thomaseizinger: Perhaps two different constructors would be a good idea here?

transports/tcp/src/lib.rs
```diff
@@ -734,116 +714,84 @@ where
...
+        }
 
-            if let Some(mut pause) = me.pause.take() {
-                match Pin::new(&mut pause).poll(cx) {
-                    Poll::Ready(_) => {}
-                    Poll::Pending => {
-                        me.pause = Some(pause);
-                        return Poll::Pending;
+        if let Some(if_watcher) = me.if_watcher.as_mut() {
+            loop {
```
thomaseizinger: In our guidelines we usually put the loop around everything, not just one single child. Would that make sense here?

transports/tcp/src/lib.rs
mxinden: I am confused why we have the delay here in the first place. If I am not mistaken, this was introduced in https://github.com/libp2p/rust-libp2p/pull/402. 

> [sleep_on_error](https://docs.rs/tk-listen/latest/tk_listen/trait.ListenExt.html#method.sleep_on_error) -- filters Stream of accepted sockets for errors. Simple errors like ConnectionReset are just ignored. Severe errors like Too many files open will delay next accept() call for the delay specified, effectively allowing other connections to be processed and release resources for new ones. [Replaces code like this][2].

https://docs.rs/tk-listen/latest/tk_listen/

I am failing to understand why this was necessary back then. I am also failing to understand why this is needed today.

Does anyone have further insight? If not, I am happy to look deeper into this.

transports/tcp/src/lib.rs
```diff
@@ -734,116 +714,84 @@ where
...
+                        if me.listen_addr.is_ipv4() == ip.is_ipv4() {
+                            let ma = ip_to_multiaddr(ip, me.listen_addr.port());
+                            log::debug!("Expired listen address: {}", ma);
+                            me.port_reuse.unregister(ip, me.listen_addr.port());
+                            return Poll::Ready(Some(Ok(TcpListenerEvent::AddressExpired(ma))));
+                        }
+                    }
+                    Poll::Ready(Err(err)) => {
+                        log::debug! {
+                            "Failure polling interfaces: {:?}. Scheduling retry.",
```
thomaseizinger: I know that this is existing code so we don't need to fix it here but I typically consider logging AND returning an error an anti-pattern.

thomaseizinger: It will lead to printing the same error multiple times.

transports/tcp/src/lib.rs
thomaseizinger: Seems reasonable to me but then we should maybe put the loop around everything and poll this more often?

transports/tcp/src/lib.rs
elenaf9: I don't think it makes sense here. Apart from the `IfWatcher`, all other sections directly return. There is never a case where a child makes progress and then we want to `continue` from the top. 
We only really need the loop in the first place because the `IfWatcher` may return Ipv6 address events that we want to skip if we listen on an ipv4 address and the other way around. In that case we ignore the event and poll the `IfWatcher` again.  
Actually, now that I think about it again I think I even prefer the previous `while let Poll::Ready(event) = if_watcher.poll_if_event(cx)`. 

I think the whole loop pattern makes sense in implementations where we drive multiple futures that are interdependent, i.e. if one child makes progress and we handle its events, it could be that now previously polled children can make progress again. But this is not the case here.

transports/tcp/src/lib.rs
elenaf9: @mxinden 
> Does anyone have further insight? If not, I am happy to look deeper into this.

I was wondering about that as well, but I also don't have any further insights.

@thomaseizinger 
> Seems reasonable to me but then we should maybe put the loop around everything and poll this more often?

Each time the delay is set we directly `return`. So there is no case where we would `continue` the loop and `pause` is `Some`.

transports/tcp/src/lib.rs
thomaseizinger: Fair enough. The only argument against would be that consistency makes it easier to read / skim the code, even if we would be able to write it in a different style here.

I don't mind much though :)

transports/tcp/src/lib.rs
elenaf9: Generally agree. Though in this specific case we only return a very general `TransportError`. The debug output gives some more insight on where the error happened (polling the IfWatcher vs polling the socket). Alternatively we could introduce a new Error enum for the two possible errors.
I am leaning towards just leaving it as it is, but I don't feel strongly about it. Happy to change it if you insist :)

transports/tcp/src/lib.rs
elenaf9: Like `TcpListenStream::new_single_interface` and `TcpListenStream::new_all_interfaces`?
They only differ by one line, so not sure if it makes sense.
(The above logic for enqueuing a `NewAddress` event can not be moved into the constructor).
Or did I misunderstand you?

transports/tcp/src/lib.rs
mxinden: Let's move the discussion on whether the delay is needed in the first place to https://github.com/libp2p/rust-libp2p/pull/2849.

transports/tcp/src/lib.rs
thomaseizinger: It would enable an early return but other than that, it is equivalent yeah! Up to you :)

transports/tcp/src/lib.rs
thomaseizinger: It would be nice to not print and return. Once we have proper backtraces, tracing the error back should be no issue.

Until then: Does the return error (`err`) not indicate that it is from the ifwatcher?

mxinden:
@elenaf9 let me know if you want me to take another look here.

transports/tcp/src/lib.rs
elenaf9: Not sure if I understand what you mean. @thomaseizinger wdyt of 4a63d84?

transports/tcp/src/lib.rs
elenaf9: Good point(s). Removed with 036f56f.

transports/tcp/src/lib.rs
thomaseizinger: Looks good to me, thank you!

thomaseizinger (approved):
Great improvement!

mxinden (commented):
Looks good to me.

Have you tested this pull request with both `tokio` and `async-std` as a runtime using a wildcard IP address?

If yes, this is ready to merge from my end.

transports/tcp/src/lib.rs
```diff
@@ -368,7 +371,25 @@ where
         socket.bind(&socket_addr.into())?;
         socket.listen(self.config.backlog as _)?;
         socket.set_nonblocking(true)?;
-        TcpListenStream::<T>::new(id, socket.into(), self.port_reuse.clone())
+        let listener: TcpListener = socket.into();
+        let local_addr = listener.local_addr()?;
+
+        if local_addr.ip().is_unspecified() {
```
melekes, line 377: shouldn't we check if port is unspecified too?

mxinden: That was my first thought as well. But while an unspecified IP results in one IP per (potentially future) interface, an unspecified port results in a **single** randomly chosen port.

@elenaf9 @melekes please correct me in case I am missing something here.

mxinden: > an unspecified port results in a single randomly chosen port.

Well, I am not sure what the OS does when it can not find a unique port number that is free on all interfaces. I am guessing that it will choose a different port number per interface. Though still **only one** per interface.

elenaf9: > That was my first thought as well. But while an unspecified IP results in one IP per (potentially future) interface, an unspecified port results in a single randomly chosen port.

Yes exactly.

elenaf9:
> Have you tested this pull request with both tokio and async-std as a runtime using a wildcard IP address?

Yes, though on my local machine (linux). But our CI also tests both runtimes in the tcp test, which includes tests with wildcard addresses. But those tests only test the tcp transport directly and not within a swarm. 
Lmk if you want me to add tests that integrate the TCP transport in the swarm.

mxinden:
> > Have you tested this pull request with both tokio and async-std as a runtime using a wildcard IP address?
> 
> Yes, though on my local machine (linux). But our CI also tests both runtimes in the tcp test, which includes tests with wildcard addresses. But those tests only test the tcp transport directly and not within a swarm. Lmk if you want me to add tests that integrate the TCP transport in the swarm.

I think this is fine as is. Thanks.

mxinden (approved):
Thanks for thinking this all the way through!

mxinden:
CI failure is unrelated. https://github.com/libp2p/rust-libp2p/pull/2869 released v0.48.0. Thus this pull request updates the version to v0.49.0. Problem is, that https://github.com/libp2p/test-plans still operates with v0.48.0.

As far as I can tell this is a chicken-and-egg problem. We can't merge with a green CI here as the Testground tests are failing. We can not merge a patch to libp2p/test-plans as the version of libp2p/rust-libp2p is not yet bumped to v0.49.0.

I don't think this is a big issue. Will need to think about how to solve it long term. //CC @laurentsenta in case you have an idea.

I will merge here. I will create a pull request against libp2p/test-plans next.

PR #2817, opened 2022-08-13
transports/noise: Migrate away from deprecated `sodiumoxide` for tests

### LINKED ISSUE #2811: RUSTSEC-2021-0137: sodiumoxide is deprecated
> sodiumoxide is deprecated

| Details             |                                                |
| ------------------- | ---------------------------------------------- |
| Status              | unmaintained                |
| Package             | `sodiumoxide`                      |
| Version             | `0.2.7`                   |
| URL                 | [https://github.com/sodiumoxide/sodiumoxide/commit/5bb1dfd2578539b89ffb0cbea25f21f00cfb963e](https://github.com/sodiumoxide/sodiumoxide/commit/5bb1dfd2578539b89ffb0cbea25f21f00cfb963e) |
| Date                | 2021-10-22                         |

Alternatives may be found - not in any specific order:

- [libsodium-sys-stable](https://crates.io/crates/libsodium-sys-stable)
- [dryoc](https://crates.io/crates/dryoc)
- [RustCrypto/nacl-compat](https://github.com/RustCrypto/nacl-compat) (`crypto_box`, `crypto_kx`, `crypto_secretstream`)
- [RustCrypto/xsalsa20poly1305](https://github.com/RustCrypto/AEADs/tree/master/xsalsa20poly1305) (`crypto_secretbox`)
- [Signatory](https://crates.io/crates/signatory)
- [ed25519-compact](https://crates.io/crates/ed25519-compact)
- [ed25519-dalek](https://github.com/dalek-cryptography/ed25519-dalek)
- [ring](https://github.com/briansmith/ring)

Recommendations can be also found from:

- [Awesome Rust Cryptography](https://github.com/The-DevX-Initiative/RCIG_Coordination_Repo/blob/main/Awesome_Rust_Cryptography.md) 

No direct maintained fork exists.

See [advisory page](https://rustsec.org/advisories/RUSTSEC-2021-0137.html) for additional details.

pinkforest:
# Description

Sodiumoxide is used for testing the legacy noise protocol crypto

Sodiumoxide is deprecated which brings for testing comparison in noise proto context;
 - sodiumoxide for ed25519 for Secret | Public -Key test ref
 - libsodium-sys 0.2.7 bindings (about 1 year ago)
 - libsodium C: crypto_sign_ed25519_pk_to_curve25519
 - libsodium C: crypto_sign_ed25519_sk_to_curve25519

This change migrates from sodiumoxide to:
 - [libsodium-sys-stable](https://crates.io/crates/libsodium-sys-stable) fork 1.19.22 (4 months ago)
 - [ed25519-compact](https://crates.io/crates/ed25519-compact) 1.0.11 (6 months ago) for Secret | Public -Key test ref

Fixes #2811 

## Change checklist

- Migrate testing from against sodiumoxide to libsodium-sys-stable and ed25519-compact

- [X] A changelog entry has been made in the appropriate crates


transports/noise/src/protocol/x25519.rs
```diff
@@ -327,29 +327,25 @@ mod tests {
...
-    }
-
-    pub fn ed25519_pk_to_curve25519(k: &sign::PublicKey) -> Option<[u8; 32]> {
+    pub fn ed25519_pk_to_curve25519(k: &test_sign::PublicKey) -> Option<[u8; 32]> {
         let mut out = [0u8; 32];
         unsafe {
-            if crypto_sign_ed25519_pk_to_curve25519(out.as_mut_ptr(), (&k.0).as_ptr()) == 0 {
+            if libsodium_sys::crypto_sign_ed25519_pk_to_curve25519(out.as_mut_ptr(), k.as_ptr())
+                == 0
+            {
```
thomaseizinger: Most of this diff would go away if we import the `crypto_sign` function directly I think.

Thoughts?

transports/noise/src/protocol/x25519.rs
```diff
@@ -278,10 +278,9 @@ impl snow::types::Dh for Keypair<X25519> {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use ed25519_compact as test_sign;
```
thomaseizinger: By renaming the import from `sign` to `test_sign`, we are already causing more diff than necessary but at the same time, the name `test_sign` is not ideal IMO.

I think we should drop this import rename and fully qualify the `ed25519_compact` types.

transports/noise/Cargo.toml
```diff
@@ -33,7 +33,8 @@ async-io = "1.2.0"
 env_logger = "0.9.0"
 libp2p-tcp = { path = "../../transports/tcp" }
 quickcheck = "0.9.0"
-sodiumoxide = "0.2.5"
+libsodium-sys-stable = { version = "1.19.22", features = ["fetch-latest"] }
```
mxinden, line 36:
```suggestion
libsodium-sys-stable = { version = "1.19.22" }
```

Could we do without this feature? I.e. would it be fine to not use the latest libsodium?

transports/noise/CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 - Update to `libp2p-core` `v0.35.0`.
 
+- Migrate testing from against sodiumoxide to libsodium-sys-stable and ed25519-compact
```
mxinden: I don't think this is relevant for a user, thus I suggest not mentioning this in the changelog.

mxinden (commented):
Thanks for looking into this @pinkforest!

transports/noise/CHANGELOG.md
pinkforest: Users are developers and I as a user of this library would like to know what crypto impl(s) is / are being used for what

transports/noise/Cargo.toml
pinkforest: It's the default - I am just explicitly mentioning here as to the intent - and it fetches the latest stable.

Otherwise it's optimized or minimal - I recommend the stable

Cargo features:

    fetch-latest: download the latest stable version.
    optimized: build a version optimized for the current platform.
    minimal: do not build deprecated APIs.

transports/noise/CHANGELOG.md
mxinden: I don't think the majority of folks cares what libraries are being used for testing.

pinkforest: Up to you - was a worth a try :) I'll change it

pinkforest: Done - Reverted that commit.

thomaseizinger (commented):
Thanks! Two ideas from my end :)

transports/noise/src/protocol/x25519.rs
pinkforest: This just confuses people and is redundant - in tests it's better to be explicit to ensure we know what we are using to test and what we are actually using.

transports/noise/src/protocol/x25519.rs
pinkforest: Problem is it is test signature not the signature we are testing - it is better to be explicit about that it is used for testing.

The types often collide and it is best to use types that signify how those types are being used - it already confused me what was supposed to be what and I hate that kind of confusion :/

thomaseizinger: I agree with the idea but does `test_sign` really communicate that any better? To make this really clear, we could add a comment on these tests, explaining that we are using a different impl of ed25519 to be better guarded against bugs in the production implementation.

transports/noise/src/protocol/x25519.rs
thomaseizinger: Agree in principle but even as it is now, this is all pretty subtle and better explained in comments in my opinion. I don't feel strongly about the diff.

transports/noise/src/protocol/x25519.rs
pinkforest: Ok. How does that look now ? I've changed to FQN ed25519_compact now.

transports/noise/src/protocol/x25519.rs
pinkforest: Don't feel strong either in the end. I've changed it :)

mxinden (approved):
This looks good to me. Thanks for the follow ups @pinkforest.

Will wait for another review from @thomaseizinger before merging.

thomaseizinger (approved):
Sorry for the long delay and thanks for fixing this!

PR #2818, opened 2022-08-13
*: fix `clippy::derive-partial-eq-without-eq`

elenaf9:
# Description

Fix clippy lint [derive_partial_eq_without_eq](https://rust-lang.github.io/rust-clippy/master/index.html#derive_partial_eq_without_eq) introduced with clippy `0.1.63`. 
Within `prost` this lint can/ will not be fixed (see https://github.com/tokio-rs/prost/issues/661 and https://github.com/tokio-rs/prost/pull/678), therefore we need to add an explicit `#[allow(clippy::derive_partial_eq_without_eq)]` flag for code generated by prost.

Edit: The reason why it can not be fixed within `prost` is because the lint returns some false positives. See https://github.com/tokio-rs/prost/runs/7192713936?check_suite_focus=true. According to https://github.com/rust-lang/rust-clippy/issues/8970 this is already fixed on nightly, so once a new clippy version is published we could remove the explicit ``#[allow(..)]` again.

### DIFF
--- a/core/src/lib.rs
+++ b/core/src/lib.rs
@@ -38,6 +38,7 @@
 #[cfg(feature = "serde")]
 extern crate _serde as serde;
 
+#[allow(clippy::derive_partial_eq_without_eq)]
 mod keys_proto {
     include!(concat!(env!("OUT_DIR"), "/keys_proto.rs"));
 }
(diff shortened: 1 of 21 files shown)


mxinden (approved):
Thanks for the fix. And thanks for researching and documenting the `prost` related issues.

PR #2819, opened 2022-08-15
swarm-derive/: Add where clause of behaviour to generated out event

mxinden:
# Description

When generating the `OutEvent` for a `NetworkBehaviour`, add the `where` clause of the
`NetworkBehaviour` `struct` to the generated `enum`.

Follow up to https://github.com/libp2p/rust-libp2p/pull/2792

## Change checklist


thomaseizinger (approved):

elenaf9:
Won't this cause an error in case of
```rust
    #[derive(NetworkBehaviour)]
    struct Foo<T>
    where
        T: Copy
    {
        ping: libp2p::ping::Ping,
        #[behaviour(ignore)]
        bar: T,
    }
```
which will generate 

```rust
pub enum FooEvent<T> // Error: unused Parameter
where
    T: Copy
{
    Ping(<libp2p::ping::Ping as NetworkBehaviour>::OutEvent)
}
```
Am I missing something?

mxinden:
@elenaf9 that is correct. Good catch. Though why would someone use the `OutEvent` mechanism in combination with `#[behaviour(ignore)]`?

Once we remove `NetworkBehaviourEventProcess` (https://github.com/libp2p/rust-libp2p/pull/2751) I don't think there is any use for `poll_method` nor `ignore` at all. See https://github.com/libp2p/rust-libp2p/pull/2751#discussion_r930451630.

With that in mind, I would leave this as is. Do you agree @elenaf9?

elenaf9:
> Though why would someone use the OutEvent mechanism in combination with #[behaviour(ignore)]?
> 
> Once we remove NetworkBehaviourEventProcess (https://github.com/libp2p/rust-libp2p/pull/2751) I don't think there is any use for poll_method nor ignore at all. See https://github.com/libp2p/rust-libp2p/pull/2751#discussion_r930451630.

I am a bit confused. Why would we not need `#[behaviour(ignore)]` anymore? Isn't it still needed to signal the derive macro that this field is not a network behaviour? E.g. when implementing the methods of the `NetworkBehaviour` trait and we pass the event to our inner behaviours?
And with `OutEvent` I would think we stilll need it to mark that this field won't generate an event:
- If the `OutEvent` is provided by the user, we should not need the trait bound `From<<FIELD as NetworkBehaviour>::OutEvent>` for this field
- If the `OutEvent` is generated, we should not generated an enum variant for this field. See above code example.

@mxinden  what am I missing here?

mxinden:
> I am a bit confused. Why would we not need `#[behaviour(ignore)]` anymore?

Let me rephrase: 

With `NetworkBehaviourEventProcess` users sometimes needed an `events: VecDeque` which they could push events into from `NetworkBehaviourEventProcess` and pop events from within their custom `poll` method. `events` would not implement `NetworkBehaviour` and thus needed the `#[behaviour(ignore)]`.

With the removal of `NetworkBehaviourEventProcess` there is no more need for fields on the root `NetworkBehaviour` that do not themselves implement `NetworkBehaviour`, at least I can not think of a use-case. 

Can you think of a use case for `#[behaviour(ignore)]` and a custom `poll` method once `NetworkBehaviourEventProcess` is removed @elenaf9?

elenaf9:
Thanks for elaborating @mxinden!  Makes sense now. 
> Can you think of a use case for #[behaviour(ignore)] and a custom poll method once NetworkBehaviourEventProcess is removed @elenaf9?

I guess the only use case is if the wrapping `NetworkBehaviour` wants to implement some additional logic on top of the inner behaviours and / or hide some of the implementation details about them. But this can also be implemented on top of the `Swarm`, so I agree that we don't need `#[behaviour(ignore)]` or `#[behaviour(poll =  ..)]` anymore. 

> With that in mind, I would leave this as is. Do you agree @elenaf9?

Yep :)

Issue #2824, opened 2022-08-17
swarm/: Support generic connection management


mxinden:
## Description

> Today a user can:
> 
> 1. Set a limit for incoming/outgoing/pending connections, global and per peer.
> 2. Close a connection via `NetworkBehaviour`.
> 3. Keep a connection alive via `ConnectionHandler::keep_alive`.
> 
> A user is restricted in (1) as they can not do anything beyond setting upper bound limits. E.g. they can not decide whether to accept or deny an incoming connection based on the IP, the `PeerId` or on their current CPU or memory utilization.

(Taken from https://github.com/libp2p/rust-libp2p/discussions/2118#discussioncomment-3134752.)

I see 3 ways how we can enable users to do advanced connection management:

1. Make the [`pool::ConnectionCounter`](https://github.com/libp2p/rust-libp2p/blob/475289c19ca3f79fe0cbad075963f4269cd125a8/swarm/src/connection/pool.rs#L926-L939) generic and allow users to provide their own.
    - Yet another moving piece.
    - When not `Box`ed requires an additional generic parameter.
    - Can only do connection management. Not as powerful as the `NetworkBehaviour` trait and thus limited in its decision making.
3. Extend `NetworkBehaviour` with methods called to review a pending or established connection.
    - Yet another set of methods on `NetworkBehaviour`.
    - Has knowledge of the full state, thus allows powerful decision making.
    - Users are already familiar with it.
5. Emit a `SwarmEvent::ReviewConnection` requiring the user to manually accept each pending/established connection via e.g. `Swarm::accept()`. 
    - Complicates the getting-started experience. User has to explicitly call a method to make progress. Not intuitive.

## Motivation

Allows downstream users to do advanced connection management and can simplify existing implementations working around this today.

- https://github.com/sigp/lighthouse/tree/stable/beacon_node/lighthouse_network/src/peer_manager
- https://github.com/paritytech/substrate/blob/master/client/peerset/src/lib.rs
- https://github.com/n0-computer/iroh/tree/main/iroh-p2p

## Are you planning to do it yourself in a pull request?

Yes

mxinden:
I am currently working on proposal (2), i.e. extending `NetworkBehaviour`. Hope to publish a first proof-of-concept this week.

thomaseizinger:
How is (2) going to work in terms of composing behaviours? One `NetworkBehaviour` declining a connection means it will be denied? i.e. Need full agreement across all `NetworkBehaviour`s?

[...]

PR #2828, opened 2022-08-19
 swarm/: Support generic connection management through NetworkBehaviour

mxinden:
# Description

This pull request enables generic connection management by implementing option (2) of https://github.com/libp2p/rust-libp2p/issues/2824:

> Extend `NetworkBehaviour` with methods called to review a pending or established connection.

Very much work in progress at this point.

## Change checklist


Cargo.toml
```diff
@@ -13,6 +13,8 @@ categories = ["network-programming", "asynchronous"]
 [features]
 default = [
     "autonat",
+    # TODO: Should this really be a default?
+    "connection-limit",
```
thomaseizinger, line 17: Long term, I would say no because of https://github.com/libp2p/rust-libp2p/issues/2173. For consistency with the current way of doing things, probably yes.

protocols/connection-limit/Cargo.toml
```diff
@@ -0,0 +1,24 @@
+[package]
+name = "libp2p-connection-limit"
```
thomaseizinger, line 2: Does this have to be its own crate? I think this could also live in `libp2p-swarm`.

It can still be feature-flagged there if we want but looking at the dependencies of this crate, we don't really gain much at the moment from separating this out.

protocols/connection-limit/src/lib.rs
```diff
@@ -0,0 +1,356 @@
...
+    pub fn new(limits: ConnectionLimits) -> Self {
+        Behaviour {
+            counters: ConnectionCounters::new(limits),
+        }
+    }
+}
+
+impl NetworkBehaviour for Behaviour {
+    type ConnectionHandler = DummyConnectionHandler;
+    type OutEvent = ();
```
thomaseizinger, line 44: Perhaps use `Void` instead?

swarm/src/behaviour.rs
```diff
@@ -778,3 +808,9 @@ impl Default for CloseConnection {
         CloseConnection::All
     }
 }
+
+// TODO: Needed in the first place? Are there some common errors that we want?
+#[derive(Debug)]
+pub enum ReviewDenied {
+    Error(Box<dyn Error>),
```
thomaseizinger, line 815: This should definitely have a `'static` bound.

swarm/src/behaviour.rs
```diff
@@ -778,3 +808,9 @@ impl Default for CloseConnection {
         CloseConnection::All
     }
 }
+
+// TODO: Needed in the first place? Are there some common errors that we want?
```
thomaseizinger, line 812: Perhaps just having these function return `Box<dyn Error + 'static>` is enough?

protocols/connection-limit/src/lib.rs
```diff
@@ -0,0 +1,356 @@
...
+        _peer_id: Option<PeerId>,
+        // TODO: Maybe an iterator is better?
+        _addresses: &[Multiaddr],
+        endpoint: Endpoint,
+    ) -> Result<(), ReviewDenied> {
+        match endpoint {
+            Endpoint::Dialer => self.counters.check_max_pending_outgoing(),
+            Endpoint::Listener => self.counters.check_max_pending_incoming(),
+        }
+        .map_err(|e| ReviewDenied::Error(e.into()))
```
thomaseizinger, line 61: We could use `?` here if you add a `From` implementation.

protocols/connection-limit/src/lib.rs
```diff
@@ -0,0 +1,356 @@
...
+        _cx: &mut Context<'_>,
+        _params: &mut impl PollParameters,
+    ) -> Poll<NetworkBehaviourAction<Self::OutEvent, Self::ConnectionHandler>> {
+        Poll::Pending
+    }
+}
+
+/// Network connection information.
+#[derive(Debug, Clone)]
+pub struct ConnectionCounters {
```
thomaseizinger, line 126: 2 things:

1. Do we need this indirection?
2. If yes, does it have to be `pub`?

protocols/connection-limit/tests/lib.rs
```diff
@@ -0,0 +1,198 @@
...
+        let limits = ConnectionLimits::default().with_max_pending_outgoing(Some(outbound.into()));
+
+        let mut swarm = make_swarm(limits);
+
+        let addr: Multiaddr = "/memory/1234".parse().unwrap();
+
+        for _ in 0..outbound {
+            swarm
+                .dial(
+                    DialOpts::peer_id(PeerId::random())
```
thomaseizinger, line 48: Does `PeerId::random` only work here because we never poll the swarm?

I think this is misleading, why not dial without `PeerId` instead?

protocols/connection-limit/tests/lib.rs
```diff
@@ -0,0 +1,198 @@
...
+use libp2p_swarm::SwarmBuilder;
+use libp2p_swarm::SwarmEvent;
+use libp2p_yamux::YamuxConfig;
+use quickcheck::QuickCheck;
+
+// TODO: Test that pending limit is decreased.
+
+#[test]
+fn enforces_pending_outbound_connection_limit() {
+    fn prop(outbound: u8) {
```
thomaseizinger, line 38: Sorry for OT but we have so many different ways of testing swarms in our repo, it makes me uneasy every time :sweat_smile:

thomaseizinger: I'll put this on my list to work on from September forward :D

swarm/src/lib.rs
```diff
@@ -282,6 +285,7 @@ where
     /// similar mechanisms.
     external_addrs: Addresses,
 
+    // TODO: I think this should move into the connection manager behaviour as well.
```
thomaseizinger, line 288: Into _the_ behaviour or into _a_ behaviour. I was thinking the same thing when I thought about this topic. Keeping a record of banned peers can likely be a `NetworkBehaviour` itself IMO.

swarm/src/behaviour.rs
```diff
@@ -191,6 +192,35 @@ pub trait NetworkBehaviour: 'static {
         vec![]
     }
 
+    // TODO: Make sure this and all the methods below are really implemented across all behaviours.
+    fn review_pending_connection(
+        &mut self,
+        _peer_id: Option<PeerId>,
+        // TODO: Maybe an iterator is better?
+        _addresses: &[Multiaddr],
+        _endpoint: Endpoint,
```
thomaseizinger, line 201: Instead of passing the endpoint, would it make sense to have two different callbacks, one for incoming and one for outgoing?

swarm/src/behaviour.rs
```diff
@@ -191,6 +192,35 @@ pub trait NetworkBehaviour: 'static {
         vec![]
     }
 
+    // TODO: Make sure this and all the methods below are really implemented across all behaviours.
+    fn review_pending_connection(
```
thomaseizinger, line 196: I am not fully sold on the "review" terminology but I also can't think of something better off the top of my head.

thomaseizinger (commented):
Hope you don't mind the early, unsolicited review :)
I've added a few thoughts but overall, this looks really interesting. I like how we move more and more to a plugin system and `libp2p-swarm` is becoming a kind of "executor" for these plugins.

protocols/connection-limit/Cargo.toml
mxinden: Separate crate forces us to deliver on the promise of being able to implement ones own.

That said, agreed that we don't gain much more and instead add complexity.

protocols/connection-limit/src/lib.rs
mxinden: Does not. Good catch. Just copied this from the previous connection limiting mechanism.

swarm/src/behaviour.rs
mxinden: Not a fan of "review" either. Grateful for naming suggestions!

swarm/src/behaviour.rs
dignifiedquire: Something like `LimitReached` seems to very common to me, so would be nice to have that as a specifc one, maybe with an optional message for customizability

thomaseizinger:
I am closing this in favor of https://github.com/libp2p/rust-libp2p/pull/3099.

Issue #2833, opened 2022-08-20
Error building libp2p with dcutr feature


haider-rs:
## Summary
Cannot build libp2p with dcutr feature.
Created a new cargo project
Added a dependency of libp2p 
`libp2p = { version = "0.46.1", features = ["tcp-tokio", "dns-tokio"] }`
run cargo check, runs fine, change the dependency 
`libp2p = { version = "0.46.1", features = ["tcp-tokio", "dns-tokio", "dcutr"] }`
throws error

## Expected behaviour
should return Finished dev

## Actual behaviour
returns 
```
error: failed to select a version for `signature`.
    ... required by package `ecdsa v0.13.3`
    ... which satisfies dependency `ecdsa-core = "^0.13"` of package `p256 v0.10.0`
    ... which satisfies dependency `p256 = "^0.10.0"` of package `libp2p-core v0.34.0`
    ... which satisfies dependency `libp2p-core = "^0.34.0"` (locked to 0.34.0) of package `libp2p v0.46.1`
    ... which satisfies dependency `libp2p = "^0.46.1"` (locked to 0.46.1) of package `testing_env v0.1.0 (/Users/mp-haidera-pyse-403/Desktop/testing_env)`
versions that meet the requirements `>=1.3.1, <1.5` are: 1.4.0, 1.3.2, 1.3.1

all possible versions conflict with previously selected packages.

  previously selected package `signature v1.6.0`
    ... which satisfies dependency `signature = ">=1.3.1"` (locked to 1.6.0) of package `ed25519 v1.5.2`
    ... which satisfies dependency `ed25519 = "^1"` (locked to 1.5.2) of package `ed25519-dalek v1.0.1`
    ... which satisfies dependency `ed25519-dalek = "^1.0.1"` (locked to 1.0.1) of package `libp2p-core v0.34.0`
    ... which satisfies dependency `libp2p-core = "^0.34.0"` (locked to 0.34.0) of package `libp2p v0.46.1`
    ... which satisfies dependency `libp2p = "^0.46.1"` (locked to 0.46.1) of package `testing_env v0.1.0 (/Users/mp-haidera-pyse-403/Desktop/testing_env)`

failed to select a version for `signature` which could resolve this conflict
```

<details><summary>Debug Output</summary>
<p>

```
<output>
```
</p>
</details>

## Version

- libp2p version (version number, commit, or branch):
0.44.0
0.46.0
0.46.1

elenaf9:
[...]

PR #2836, opened 2022-08-22
protocols/dcutr: disable `libp2p-core` default features

elenaf9:
# Description

Fix `libp2p-dcutr` is leaking the default features for `libp2p-core`.

Workaround for #2833.


elenaf9:
@mxinden do I also need to bump the dependency version in the root `Cargo.toml` and bump the `libp2p` version or is this not necessary because its only a patch version bump?

protocols/dcutr/CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+# 0.5.1 [unreleased]
```
mxinden:
```suggestion
# 0.5.1
```

I will release this right away.

mxinden (approved):
Thanks for the fix! Will cut a new release right away.

> @mxinden do I also need to bump the dependency version in the root Cargo.toml and bump the libp2p version or is this not necessary because its only a patch version bump?

No need for the bump in the root `Cargo.toml`.

mxinden:
[Tagged](https://github.com/libp2p/rust-libp2p/releases/tag/libp2p-dcutr-0.5.1) and [published](https://crates.io/crates/libp2p-dcutr/0.5.1).

PR #2842, opened 2022-08-24
swarm-derive/: Remove support for ignoring fields on struct

mxinden:
# Description

With the removal of `NetworkBehaviourEventProcess` there is no more need for
ignoring fields.

Any objections? Anyone can still see a use-case for fields that need to be ignored?

Deprecation of `NetworkBehaviourEventProcess` https://github.com/libp2p/rust-libp2p/pull/2840 and removal https://github.com/libp2p/rust-libp2p/pull/2840.

Past discussions on the removal of `poll` https://github.com/libp2p/rust-libp2p/pull/2751#discussion_r930451630 and https://github.com/libp2p/rust-libp2p/pull/2819#issuecomment-1221249932

## Change checklist


thomaseizinger (approved):
No objections!

dariusc93:
If im reading this right, we can have different fields in the struct without it being specific to any behaviour since originally those fields that didnt implement such behaviour would need to be ignored to be there?

mxinden:
I am not sure I am following @dariusc93. With this change you can no longer have any fields on your main `struct` that do not implement `NetworkBehaviour`. Do you foresee any issues with that?

dariusc93:
Sorry I wasnt clear (misread the pr when looking over it last night). Ignoring fields do have its benefits. In one of my use case would be to have the struct contain a registry to track `QueryId` and check against that when receiving a `KademliaEvent::OutboundQueryCompleted` event and return the results. I know rust-ipfs does the same (though might not be a good example since it havent seen much activities on that side). I might have some workarounds though.

mxinden:
> Ignoring fields do have its benefits. In one of my use case would be to have the struct contain a registry to track `QueryId` and check against that when receiving a `KademliaEvent::OutboundQueryCompleted` event and return the results.

Why not store the "registry to track `QueryId`" in the object (e.g. `struct`) that is polling the `Swarm`?

PR #2850, opened 2022-08-26
add IDE(idea/vscode/vim/emacs) specific gitignore.

SuccinctPaul:
# Description

## Change checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,14 @@
 target
 Cargo.lock
+
+# VSCode & Idea settings
+.vscode/
+.idea/
+# Vim swap files
+*.swp
+*.swo
+# Emacs backup files
+*~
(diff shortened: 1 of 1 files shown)


mxinden:
I don't think rust-libp2p should have any editor specific configuration. Why not use the global `.gitignore` on your local machine?

SuccinctPaul:
It's kind of validating, as most projects do.

Because there always be someone doesn't set global .gitignore on their machine.

mxinden:
> Because there always be someone doesn't set global .gitignore on their machine.

Those cases would be caught in code reviews, which has not happened in the recent past. With that in mind, I don't think this is necessary.

I am closing here. I appreciate your contribution. Don't let this discourage you from contributing in the future :)

Issue #2848, opened 2022-08-25
prost version upgrade brakes core build scripts


divagant-martian:
## Summary
the upgrade to prost v0.11.0 includes some breaking changes, including prost no longer installing protoc. As a consequence the last version requires protoc to be explicitely/manually installed. See https://github.com/tokio-rs/prost/releases/tag/v0.11.0 and https://github.com/sigp/lighthouse/runs/8005207211?check_suite_focus=true#step:6:1020 for the error that now we get building libp2p's latests version

## Expected behaviour

Either libp2p should continue to build or the new requirement should be noted in the release notes and installation guide

## Actual behaviour

upgrading libp2p unexpectedly fails to build. See below copied error from CI

<details><summary>Debug Output</summary>
<p>

```
error: failed to run custom build command for `libp2p-core v0.35.1`

Caused by:
  process didn't exit successfully: `/home/runner/work/lighthouse/lighthouse/target/release/build/libp2p-core-730d1467ea8eddc3/build-script-build` (exit status: 101)
  --- stderr
  thread 'main' panicked at '
  Could not find `protoc` installation and this build crate cannot proceed without
  this knowledge. If `protoc` is installed and this crate had trouble finding
  it, you can set the `PROTOC` environment variable with the specific path to your
  installed `protoc` binary.

  For more information: https://docs.rs/prost-build/#sourcing-protoc
  ', /home/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/prost-build-0.11.1/src/lib.rs:1227:10
  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
warning: build failed, waiting for other jobs to finish...
error: failed to compile `lighthouse v3.0.0 (/home/runner/work/lighthouse/lighthouse/lighthouse)`, intermediate artifacts can be found at `/home/runner/work/lighthouse/lighthouse/target`
make: *** [install] Error 101
Makefile:28: recipe for target 'install' failed
```
</p>
</details>

## Possible Solution

ideally install protoc, but I assume there is a reason behind prost removing this from their build script. If this is not possible, protoc should be noted as a build dependency in the docs.

## Version

v0.47.0

## Would you like to work on fixing this bug?

Maybe but I'm not sure why the change was done in prost in the first place

mxinden:
Thanks for calling this out @divagant-martian!

I updated the release notes:

https://github.com/libp2p/rust-libp2p/releases/tag/v0.47.0

[...]

PR #2851, opened 2022-08-26
Update changelogs for prost dep update

divagant-martian:
# Description
Patch changelogs for #2788

#2848 

## Open Questions
na

## Change checklist

- [ ] I have made corresponding changes to the documentation **does not apply**
- [ ] I have added tests that prove my fix is effective or that my feature works **does not apply**

### DIFF
--- a/core/CHANGELOG.md
+++ b/core/CHANGELOG.md
@@ -6,6 +6,8 @@
 
 # 0.35.0
 
+- Update prost requirement from 0.10 to 0.11 which no longer installs the protoc Protobuf compiler.
+  Thus you will need protoc installed locally. See [PR 2788].
 - Drop `Unpin` requirement from `SubstreamBox`. See [PR 2762] and [PR 2776].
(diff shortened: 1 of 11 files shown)


mxinden (approved):
Thank you @divagant-martian!

dignifiedquire:
thanks for the update. just a general note this is really painful when using libp2p now, as all ci flows have to be updated to manually install protoc, really unfortunate for downstream users

mxinden:
Agreed @dignifiedquire, though I don't see a solution. Do you? 

I don't think this is libp2p specific, but applies to any application using `prost`.

dignifiedquire:
agree it is a general prost problem, one idea I had is to make a rust crate that brings protoc with it, so one could include in its build deps

jxs:
It seems https://github.com/tokio-rs/prost/pull/620 addresses that, but has been stagnant since https://github.com/tokio-rs/prost/pull/657

Issue #2832, opened 2022-08-20
swarm/behaviour: Inject events via single method


mxinden:
## Description

Replace the various `NetworkBehaviour::inject_xxx` methods with a single `NetworkBehaviour::on_event` method and a `InEvent` `enum`.

Example replacing `NetworkBehaviour::inject_connection_established`:

``` diff
diff --git a/swarm/src/behaviour.rs b/swarm/src/behaviour.rs
index c0ac5976..c10adb7d 100644
--- a/swarm/src/behaviour.rs
+++ b/swarm/src/behaviour.rs
@@ -191,7 +191,13 @@ pub trait NetworkBehaviour: 'static {
         vec![]
     }
 
+    fn on_event(&mut self, _event: InEvent) {}
+
     /// Informs the behaviour about a newly established connection to a peer.
+    #[deprecated(
+        since = "0.39.0",
+        note = "Handle `InEvent::ConnectionEstablished` in `NetworkBehaviour::on_event` instead."
+    )]
     fn inject_connection_established(
         &mut self,
         _peer_id: &PeerId,
@@ -778,3 +784,15 @@ impl Default for CloseConnection {
         CloseConnection::All
     }
 }
+
+pub enum InEvent<'a> {
+    /// Informs the behaviour about a newly established connection to a peer.
+    ConnectionEstablished {
+        peer_id: PeerId,
+        connection_id: ConnectionId,
+        endpoint: &'a ConnectedPoint,
+        // TODO: Would a slice not be better?
+        failed_addresses: Option<&'a Vec<Multiaddr>>,
+        other_established: usize,
+    },
+}
diff --git a/swarm/src/lib.rs b/swarm/src/lib.rs
index 38df855c..dfefb050 100644
--- a/swarm/src/lib.rs
+++ b/swarm/src/lib.rs
@@ -710,6 +710,14 @@ where
                         failed_addresses.as_ref(),
                         non_banned_established,
                     );
+                    self.behaviour
+                        .on_event(behaviour::InEvent::ConnectionEstablished {
+                            peer_id,
+                            connection_id: id,
+                            endpoint: &endpoint,
+                            failed_addresses: failed_addresses.as_ref(),
+                            other_established: non_banned_established,
+                        });
                     return Some(SwarmEvent::ConnectionEstablished {
                         peer_id,
                         num_established,
```

## Motivation

[...]

Issue #2778, opened 2022-07-26
swarm/src/behaviour: Add KeepAlive Behaviour


mxinden:
## Description

When getting started with rust-libp2p, it is confusing that connections over time through the `KeepAlive` mechanism.

See e.g. https://github.com/libp2p/rust-libp2p/issues/2768 https://github.com/libp2p/rust-libp2p/pull/2679

Today, users can import the `Ping` `NetworkBehaviour` and set `with_keep_alive`.

https://github.com/libp2p/rust-libp2p/blob/ce963dfcaa0a99e4b9bef27cdfe29e48946d823a/protocols/ping/src/handler.rs#L103-L116

I think there is value in a simple `KeepAlive` `NetworkBehaviour` implementation, keeping the connection alive, without the need to import `Ping`. The `ConnectionHandler` of the `KeepAlive` `NetworkBehaviour` could be exposed as well, thus allowing it to be used without its parent `NetworkBehaviour`.

## Are you planning to do it yourself in a pull request?

No

K0UR05H:
It seems like it's possible to keep a connection open using `DummyBehaviour::with_keep_alive` but I don't know if it's good enough.

mxinden:
To be honest I haven't been aware of `DummyBehaviour::with_keep_alive`. Thanks @k0ur0x.

I am closing here for now. I would still consider a pure `KeepAlive` `NetworkBehaviour` useful. That said, I doubt it is worth the additional work.

thomaseizinger:
I am reopening this because:

- the orthogonality of the static keep alive mechanism with the remaining ping functionality is highlighted once again in https://github.com/libp2p/rust-libp2p/pull/2852
- I find `DummyBehaviour` to be a rather strange component too and thus I think it isn't a particularly good solution to the problem discussed here. Also see https://github.com/libp2p/rust-libp2p/pull/2859.

Issue #2657, opened 2022-05-17
protocols/stream: Add a streaming-response protocol to rust-libp2p


b5:
## Description

Add support for streaming multiple responses, instead of a 1-1 request-response model in rust-libp2p.

## Motivation

* This has been explored in #1947
* we've taken a stab at an implementation for Iroh: https://github.com/n0-computer/iroh/pull/26/commits/b55e00e445644941e17c7e31b811657eb164cb0f 
* our use case was to use libp2p in place of a gRPC style RPC/IPC framework
* we had to abandon this effort due to [time constraints](https://www.notion.so/iroh-docs/Switch-from-libp2p-streaming-to-gRPC-for-IPC-4f42299f5705440dba7a253f5479f550)
* @carsonfarmer is interested in [similar functionality](https://discord.com/channels/806902334369824788/847994211923525652/975835290369392731)
* substrate has a [similar generic notifications protocol](https://github.com/paritytech/substrate/tree/master/client/network/src/protocol/notifications). It isn't very usable by third parties, but they clearly have a usecase for it. This use case isn't the same as ours, but may be worth considering

## Requirements

* would be great to gather other's use cases before fleshing out strict requirements

## Open questions

Would this constitute a change to the libp2p spec? Interop story across implementations is an open question. Our use case only requires rust support this feature.

## Are you planning to do it yourself in a pull request?

**Maybe**. We've taken a stab at it, and may be able to help in a joint effort starting sometime after July 20th. I'd be happy to help with planning & coordination in the meantime.

carsonfarmer:
Thanks for kicking this off @b5. I'm also willing to help with planning, and potentially contribute some code and testing.

rkuhn:
We (Actyx) have used our own implementation based on OneShot requests in our control plane (e.g. the admin accessing the node for settings and debugging). This turned out to be a performance bottleneck for high rate data transfers (e.g. dump/restore of a node’s whole data), which led me to try my hand at a protocol based on persistent substreams. Results for our use-case are very promising, but it is not yet battle-hardened (other features of higher priority intervened). You can find the current code in [this gist](https://gist.github.com/rkuhn/413aa0cb4f7415bbb10c3cddd1fa0615).

The API is as simple as I could make it:

- `request()` takes a request message and a channel for transmitting the responses
- the behaviour emits requests paired with a channel for sending the responses

What do you think?

mxinden:
> * `request()` takes a request message and a channel for transmitting the responses
[...]

PR #2852, opened 2022-08-28
swarm: Add `FromFn` `ConnectionHandler`

thomaseizinger:
## Description

The `from_fn` handler is a building block for describing protocols that have little or no connection-specific state. It encourages you to implement your protocol as a function of data without suspending the IO stream. This allows you to use regular `async` functions to express whatever message handshake you want.

To allow sharing of data between connections, `from_fn` uses an abstraction called `Shared`. It implements the observer pattern and pushes new versions of this data to all connections every time it gets updated. Each new stream (inbound or outbound) gets a copy of this state when created.

The stream handlers can either complete with a single event or emit several events as they progress. Thus, this abstraction should also help with the much requested "streaming response" protocol.

With `from_fn`, you still need to implement your own `NetworkBehaviour`. This is considered to be relatively easy though. Not having to implement a `ConnectionHandler` is a major step towards better usability.

## Open Questions

- I think we need a better name. Any suggestions?
	- `async_fn_handler`?
- Should AutoNat also be converted?
- Should identify also be converted?
- Should Kademlia be converted? As long as can fit the DHT into memory, this would be possible.
- What do we do with `libp2p-request-response`?

## Open tasks

- [x] Propose `ReadyUpgrade` as separate PR: https://github.com/libp2p/rust-libp2p/pull/2855
- [x] Add a utility for notifying all handlers
- [x] Implement limit for max incoming substreams (const generic vs regular number?)
- [x] Implement limit for max pending dials?
- [ ] Fix TODOs in code
- [x] Experiment with this handler in other protocols (ping? rendezvous? request-response?)

## Change checklist


thomaseizinger:
One problem we currently have is that with `NotifyHandler` on `NetworkBehaviourAction`, we cannot notify all connections which makes it cumbersome to use the `InEvent::UpdateState` variant.

@mxinden @elenaf9 What do you think of adding `NotifyHandler::All`?

thomaseizinger:
I looked at using this implementation for the ping protocol and I think it would be doable apart from I thing: Overriding `keep_alive`. The current ping `ConnectionHandler` allows the user the configure whether the connection should be kept alive. We could drop this functionality in favor of the `DummyBehaviour` (should we rename that one to `StaticKeepAliveBehaviour` to be easier to discover? I think overriding `keep_alive` is orthogonal to the ping behaviour.)

Thoughts?

mxinden:
> I think overriding `keep_alive` is orthogonal to the ping behaviour.

Agreed. See https://github.com/libp2p/rust-libp2p/issues/2778.

swarm/src/handler/from_fn.rs
```diff
@@ -0,0 +1,349 @@
...
+    outbound_streams: FuturesUnordered<BoxFuture<'static, TOutbound>>,
+
+    on_new_inbound:
+        Box<dyn Fn(NegotiatedSubstream, &mut TState) -> BoxFuture<'static, TInbound> + Send>,
+    on_new_outbound: Box<
+        dyn Fn(NegotiatedSubstream, &mut TState, TOutboundInfo) -> BoxFuture<'static, TOutbound>
+            + Send,
+    >,
+
+    idle_waker: Option<Waker>,
```
mxinden: Currently implementations of `ConnectionHandler` depend on being `poll`ed after any of the `inject_xxx` methods are called. I don't think that implicit assumption is great. I prefer the approach you take here, making it explicit via `idle_waker`. The only downside is the additional complexity.

swarm/src/handler/from_fn.rs
```diff
@@ -0,0 +1,349 @@
...
+        inbound_streams_limit,
+        pending_dial_limit,
+        inbound_streams: FuturesUnordered::default(),
+        outbound_streams: FuturesUnordered::default(),
+        on_new_inbound: Box::new(move |stream, state| on_new_inbound(stream, state).boxed()),
+        on_new_outbound: Box::new(move |stream, state, info| {
+            on_new_outbound(stream, state, info).boxed()
+        }),
+        idle_waker: None,
+        pending_dials: VecDeque::default(),
```
mxinden:
```suggestion
        pending_outbound: VecDeque::default(),
```

It took me a bit to understand that this is an outbound stream and not connection. I associate _dial_ with a connection.

mxinden (commented):
I have to give this abstraction more thought. I find this abstraction to be very similar to `libp2p-request-response`. I have to think about whether it makes sense to have this next to `libp2p-request-response`, have this replace `libp2p-request-response` or have `libp2p-request-response` build on top of this.

swarm/src/handler/from_fn.rs
thomaseizinger: Makes sense, I'll change it!

swarm/src/handler/from_fn.rs
thomaseizinger: Right, I didn't even think about it but of course, with our `loop`s, we implicitly poll components again after we have called functions on them.

`ConnectionHandler` is not going to be used outside of `NetworkBehaviour` so I think it is fine to depend on this and remove the complexity from here.

thomaseizinger:
> I have to give this abstraction more thought. I find this abstraction to be very similar to `libp2p-request-response`. I have to think about whether it makes sense to have this next to `libp2p-request-response`, have this replace `libp2p-request-response` or have `libp2p-request-response` build on top of this.

My preference would be to first see if we can successfully use this abstraction within our monorepo, e.g. have `libp2p-request-response` build on top of this.

In addition, I'd like to somehow survey people to see how they use it / whether they know about this abstraction. Perhaps more examples can also help with that.

swarm/src/handler/from_fn.rs
```diff
@@ -0,0 +1,349 @@
...
+    pending_dial_limit: usize,
+    on_new_inbound: impl Fn(NegotiatedSubstream, &mut TState) -> TInboundFuture + Send + 'static,
+    on_new_outbound: impl Fn(NegotiatedSubstream, &mut TState, TOutboundOpenInfo) -> TOutboundFuture
+        + Send
+        + 'static,
+) -> FromFn<TInbound, TOutbound, TOutboundOpenInfo, TState>
+where
+    TInboundFuture: Future<Output = TInbound> + Send + 'static,
+    TOutboundFuture: Future<Output = TOutbound> + Send + 'static,
+    TState: Default,
```
elenaf9: What do you think about having a `TState` passed as parameter instead of using `Default`? Use case would be config parameters like timeouts, or for example the next_request_id `Arc` in the request-response protocol:
https://github.com/libp2p/rust-libp2p/blob/89f898c69f6e58944e746565c3f8f8a3bbf3324e/protocols/request-response/src/handler.rs#L95-L96

What I think we could then also do is make `on_new_inbound` and `on_new_outbound` function pointers instead of closures. Anything that they need to capture from the environment can be passed in the initial state.

swarm/src/handler/from_fn.rs
```diff
@@ -0,0 +1,349 @@
...
+        }),
+        idle_waker: None,
+        pending_dials: VecDeque::default(),
+        failed_open: VecDeque::default(),
+        state: TState::default(),
+    }
+}
+
+#[derive(Debug)]
+pub enum OutEvent<I, O, OpenInfo> {
```
elenaf9, line 219: Related to #2854: rename `OutEvent` -> `ToBehaviour` and `InEvent` -> `FromBehaviour` / `ToHandler` (the latter would describe it better from the behaviour's / Users PoV) ?

swarm/src/handler/from_fn.rs
```diff
@@ -0,0 +1,349 @@
...
+        }
+
+        if let Some(outbound_open_info) = self.pending_dials.pop_front() {
+            return Poll::Ready(ConnectionHandlerEvent::OutboundSubstreamRequest {
+                protocol: SubstreamProtocol::new(
+                    ReadyUpgrade::new(self.protocol),
+                    outbound_open_info,
+                ),
+            });
+        }
```
elenaf9: Following our coding guidelines i think we should move this block to the bottom so that we first return finished local work (finished upgrades on existing substreams) before we open new substreams.

swarm/src/handler/from_fn.rs
```diff
@@ -0,0 +1,349 @@
...
+
+    inbound_streams: FuturesUnordered<BoxFuture<'static, TInbound>>,
+    outbound_streams: FuturesUnordered<BoxFuture<'static, TOutbound>>,
+
+    on_new_inbound:
+        Box<dyn Fn(NegotiatedSubstream, &mut TState) -> BoxFuture<'static, TInbound> + Send>,
+    on_new_outbound: Box<
+        dyn Fn(NegotiatedSubstream, &mut TState, TOutboundInfo) -> BoxFuture<'static, TOutbound>
+            + Send,
+    >,
```
elenaf9: If we use a separate `FromFnProto` and implement `IntoConnectionHandler`, we would know the `PeerId` and `ConnectedPoint` of the remote and make it an parameter of the user's function. 
So it would then be `Fn(NegociatedSubstream, &mut TState, PeerId, ConnectedPoint)`.
Use case would be e.g. if we have different upgrades depending on whether a connection is relayed not, or depending on the PeerId, etc..

elenaf9 (commented):
👍 I like the general idea. 

> I tried using this locally for rendezvous and we would have to use this mechanism to push the registration data from the behaviour down to the connections.

Did you manage to implement the rendezvous or any other protocol completely with this handler?

> Is the name any good?

No strong opinion. Maybe `FromUpgrade` would describe it a bit better?

> @mxinden @elenaf9 What do you think of adding `NotifyHandler::All`?

I don't see a reason not to add it, and updating the state of all handlers sounds like a valid use-case, so I'm in favor of it.

swarm/src/handler/from_fn.rs
thomaseizinger: Fair enough. I wanted to reduce the number of parameters but having this configurable seems reasonable!

thomaseizinger: Wouldn't function pointers be unnecessarily limiting? What if the user wants to close over certain variables?

Is your argument that they should capture it in `TState` instead?

elenaf9: > Is your argument that they should capture it in TState instead?

Yes exactly (provided that we decide to make `TState` a parameter). I don't think we would limit the user with this in any case, and we could reduce complexity a bit. Can you think of any case that would not be possible with this?

thomaseizinger: I think everything would still be possible but it is just not as ergonomic. I'd assume that users are used to closures and having to think how to "emulate" them creates complexity.

Why would it be simpler to have a function pointer?

thomaseizinger: To be clear, I agree making `TState` a parameter but I'd keep the closures.

swarm/src/handler/from_fn.rs
thomaseizinger: Unless we agree on https://github.com/libp2p/rust-libp2p/issues/2854, I'd refrain from renaming it here and rather rename them all in one go.

swarm/src/handler/from_fn.rs
thomaseizinger: That sounds like a good addition!

swarm/src/handler/from_fn.rs
elenaf9: > I think everything would still be possible but it is just not as ergonomic. I'd assume that users are used to closures and having to think how to "emulate" them creates complexity.
> 
> Why would it be simpler to have a function pointer?


We would not need to box them and it would improve readability a bit. But I also see your point, so I am fine with sticking with the generic `Fn`.

thomaseizinger: Thanks for the suggestion! The boxing is internal so I don't think it concerns users much. I think the best of both worlds is allowing users to pass in `TState` but also allow closures.

Now that I think of it, we can actually make them `FnMut`s to allow even more use cases.

thomaseizinger:
> > @mxinden @elenaf9 What do you think of adding `NotifyHandler::All`?
> 
> I don't see a reason not to add it, and updating the state of all handlers sounds like a valid use-case, so I'm in favor of it.

Cool, I'll open a separate PR for it then :)

thomaseizinger:
> > I tried using this locally for rendezvous and we would have to use this mechanism to push the registration data from the behaviour down to the connections.
> 
> Did you manage to implement the rendezvous or any other protocol completely with this handler?

I don't have a working example yet:

- Rendezvous is tricky because we kind of want `NotifyHandler::All` to continuously push the state down to each handler.
- Ping is tricky because it currently exposes a "keep connection alive" mechanism that doesn't really have anything to do with ping. I've opened https://github.com/libp2p/rust-libp2p/pull/2859 as a first step of resolving this.
- I haven't tried using it for `libp2p-request-response` but that one should be fairly easy I think.

mxinden:
> > > @mxinden @elenaf9 What do you think of adding `NotifyHandler::All`?
> > 
> > 
> > I don't see a reason not to add it, and updating the state of all handlers sounds like a valid use-case, so I'm in favor of it.
> 
> Cool, I'll open a separate PR for it then :)

Note that this would require `ConnectionHandler::InEvent` to be `Clone`. See https://github.com/libp2p/rust-libp2p/pull/1880.

thomaseizinger:
> > > > @mxinden @elenaf9 What do you think of adding `NotifyHandler::All`?
> > > 
> > > 
> > > I don't see a reason not to add it, and updating the state of all handlers sounds like a valid use-case, so I'm in favor of it.
> > 
> > 
> > Cool, I'll open a separate PR for it then :)
> 
> Note that this would require `ConnectionHandler::InEvent` to be `Clone`. See #1880.

Thanks for linking this. I guess we've come full circle. I was thinking that it will require `Clone` but I hadn't yet explored, what the implications of such a bound are. I was hoping to constrain it somehow to just that one variant but I think that is probably only possible with unsafe code.

To make this handler really useful, we'd need some kind of helper that sends an action to all connections.

swarm/src/handler/from_fn.rs
thomaseizinger: I think this will allow for a really neat implementation of the identify protocol because we just have the observed address available in here :)

thomaseizinger:
@mxinden @elenaf9 Please have another look at this! I also added a dirty patch that converts `libp2p-rendezous` to use the new `FromFn` handler.

I also added a `Shared` abstraction that will "automatically" push state updates to the handler any time the state is modified. We also have a passing test now wthin the `from_fn` module.

Looking forward to your feedback. I think we might be onto something here. Pushing state to the connection handlers makes a many things a lot easier!

protocols/rendezvous/src/server.rs
```diff
@@ -109,190 +113,169 @@ pub enum Event {
...
+                }
+                Err(e) => {
+                    substream
+                        .send(Message::DiscoverResponse(Err(ErrorCode::InvalidCookie)))
+                        .await?;
+                    substream.close().await?;
+
+                    None
+                }
+            },
```
thomaseizinger: This is where we harvest the fruits of this work! Because the `ConnectionHandler` has access to `RegistrationData`, we can "answer" the remote's query inline which allows us to use `async/await` to write the protocol. In the end, it is just a simple request-response but we couldn't do this before because we needed to suspend the stream and wait for the data coming back from the behaviour!

I think this could be a good fit for any protocol where all the data nicely fits into memory.

thomaseizinger:
Another update on this:

I've introduced a builder pattern to builder the initial handler. This makes it much more convenient to configure a handler that doesn't accept inbound streams for example or will never make outbound streams. A lot of this is now also enforced in the type system.

I am gonna try to full convert `libp2p-rendezous` at some point and see how it works. For any protocol where the substream finishes eventually, I believe this is a great fit.

At the moment, we can't convert `libp2p-ping` because it reuses the substream for future pings. I might look into using `Stream` as an abstraction though instead of `Future` to support protocols that keep the substream open for longer.

thomaseizinger:
The latest patches introduce the notion of a `Stream` handler which allows us to implement protocols such as `libp2p-ping` that reuse their stream and emit events as they go along.

I think this is getting close to fulfilling the requirements of #2657.

protocols/rendezvous/src/server.rs
thomaseizinger: > I think this could be a good fit for any protocol where all the data nicely fits into memory.

Note that with the latest patches, the data is shared between all connections with an `Arc` and thus not actually duplicated in-memory.

swarm/src/handler/from_fn.rs
```diff
@@ -0,0 +1,928 @@
...
+        }
+    }
+
+    pub fn register_connection(&mut self, peer_id: PeerId, id: ConnectionId) {
+        self.connections.insert((peer_id, id));
+    }
+
+    pub fn unregister_connection(&mut self, peer_id: PeerId, id: ConnectionId) {
+        self.connections.remove(&(peer_id, id));
+    }
```
thomaseizinger: Once https://github.com/libp2p/rust-libp2p/issues/2832 is fixed, we can make this more ergonomic by just requiring a reference to the `FromSwarm` event and we can pick out the data as we need.

thomaseizinger:
Rendezvous is now also fully migrated and functional.

protocols/rendezvous/src/server.rs
```diff
@@ -109,192 +149,249 @@ pub enum Event {
...
+                        error,
+                    },
+                )]
+            }
+            from_fn::OutEvent::InboundEmitted(Ok(Some(InboundOutEvent::Unregister(namespace)))) => {
+                self.registrations.remove(namespace.clone(), peer);
+
+                vec![NetworkBehaviourAction::GenerateEvent(
+                    Event::PeerUnregistered { peer, namespace },
+                )]
```
thomaseizinger: A lot of code here is only present to make things observable from the outside. We don't really need to emit all these events in order for the protocol to be functional.

thomaseizinger:
I think this is ready for a proper review @mxinden @jxs @elenaf9!

I've updated the PR description.

protocols/ping/src/lib.rs
```diff
@@ -118,30 +232,159 @@ impl NetworkBehaviour for Behaviour {
...
+                            Either::Right((Ok((stream, rtt)), _)) => {
+                                Ok(Some((Success::Ping { rtt }, stream)))
+                            }
+                            Either::Right((Err(e), _)) => {
+                                Err(Failure::Other { error: Box::new(e) })
+                            }
+                        }
+                    })
+                }
+            })
```
rkuhn, line 267: yes, this looks pretty neat!

swarm/src/handler/from_fn.rs
```diff
@@ -0,0 +1,927 @@
...
+            },
+        }
+    }
+}
+
+impl<TState> Builder<WantInboundHandler<TState>> {
+    pub fn with_inbound_handler<TInbound, TInboundStream>(
+        self,
+        max_inbound: usize,
+        handler: impl Fn(NegotiatedSubstream, PeerId, ConnectedPoint, Arc<TState>) -> TInboundStream
```
rkuhn, line 107: Please include the protocol name (also for outbound).

swarm/src/handler/from_fn.rs
```diff
@@ -0,0 +1,927 @@
...
+pub struct WantInboundHandler<TState> {
+    state: Arc<TState>,
+}
+
+pub struct WantOutboundHandler<TState, TInbound> {
+    state: Arc<TState>,
+    max_inbound: usize,
+    inbound_handler: Option<
+        Box<
+            dyn Fn(
```
rkuhn, line 69: AFAICS this restricts the part of the handler that constructs the stream processor to shared access to its closure — is there a reason for this? If not I’d prefer FnMut.

swarm/src/handler/from_fn.rs
```diff
@@ -0,0 +1,927 @@
...
+/// Inbound substreams may be opened at any time by the remote. To facilitate this one and more usecases,
+/// the supplied callbacks for inbound and outbound substream are given access to the handler's `state`
+/// field. State can be shared between the [`NetworkBehaviour`] and a [`FromFn`] [`ConnectionHandler`]
+/// via the [`Shared`] abstraction.
+///
+/// [`Shared`] tracks a piece of state and updates all registered [`ConnectionHandler`]s whenever the
+/// state changes.
+///
+/// [`NetworkBehaviour`]: crate::NetworkBehaviour
+pub fn from_fn(protocol: &'static str) -> Builder<WantState> {
```
rkuhn, line 46: please allow multiple protocols, like
```rust
pub fn from_fn<I>(protocols: I) -> Builder<I, WantState>
where
    I: IntoIterator + Clone + Send + 'static
    I::Item: ProtocolName
{ ... }
```

swarm/src/handler/from_fn.rs
```diff
@@ -0,0 +1,927 @@
...
+            });
+        }
+
+        if self.inbound_streams.is_empty()
+            && self.outbound_streams.is_empty()
+            && self.pending_outbound_streams.is_empty()
+        {
+            if self.keep_alive.is_yes() {
+                // TODO: Make timeout configurable
+                self.keep_alive = KeepAlive::Until(Instant::now() + Duration::from_secs(10))
```
rkuhn, line 677: It feels weird that this grace period is implemented on the level of ConnectionHandler — shouldn’t it be a NetworkBehaviour setting that applies whenever KeepAlive toggles from Yes to No?

swarm/src/handler/from_fn.rs
```diff
@@ -0,0 +1,927 @@
...
+    outbound_streams: SelectAll<BoxStream<'static, TOutbound>>,
+
+    on_new_inbound: Option<
+        Box<
+            dyn Fn(
+                    NegotiatedSubstream,
+                    PeerId,
+                    ConnectedPoint,
+                    Arc<TState>,
+                ) -> BoxStream<'static, TInbound>
```
rkuhn, line 474: I’d prefer type aliases for these function types

swarm/src/handler/from_fn.rs
```diff
@@ -0,0 +1,927 @@
...
+        THandler: IntoConnectionHandler,
+    {
+        if self.dirty {
+            self.shared = Arc::new(self.inner.clone());
+
+            self.pending_update_events = self
+                .connections
+                .iter()
+                .map(|(peer_id, conn_id)| (*peer_id, *conn_id, self.shared.clone()))
+                .collect();
```
rkuhn, line 312: This erases previously undelivered wake-ups — which is fine unless it happens in a tight loop.

rkuhn (commented):
Yes, this does indeed look quite promising!

swarm/src/handler/from_fn.rs
thomaseizinger: No particular reason, I guess this could be useful if you want to track connection-local state across all inbound/outbound streams?

That would only be shared across one closure though, meaning you wouldn't be able to share state between the inbound and outbound factory function.

I think we might be able to change this interface to allow for a single factory function that gets called for both inbound and outbound streams which would allow you to share state. Is that what you would be after?

swarm/src/handler/from_fn.rs
thomaseizinger: Yeah that is doable!

swarm/src/handler/from_fn.rs
thomaseizinger: I am not happy with this either but I couldn't think of a better way to handle this. `keep_alive` is per connection. My initial implementation would have been a `with_keep_alive_timeout` configuration parameter on the builder that defaults to 30s or something.

swarm/src/handler/from_fn.rs
thomaseizinger: What do you mean exactly? As a user, you shouldn't need to come in contact this this `BoxStream` per se. Where would you like to see a type alias?

swarm/src/handler/from_fn.rs
thomaseizinger: The old events are useless though because they contain stale state. I am not sure if there is a better way?

swarm/src/handler/from_fn.rs
thomaseizinger: At the moment, there is only one protocol possible so it will always be that one if that function is called.

rkuhn: I know, and I’d like to [change that](https://github.com/rkuhn/blog/blob/master/08_peer-to-peer_internet.md). All prefabricated upgrades today ignore the negotiated protocol name, which to my mind leaves the second most important value provided by libp2p on the table (the first one being the ability to dial a PeerId).

swarm/src/handler/from_fn.rs
rkuhn: Sharing state is not my concern because doing that always proceeds via shared references (of which Arc is a special case). `Fn` does not allow captured values to be modified (more precisely: mutably borrowed), even if they are owned.

swarm/src/handler/from_fn.rs
rkuhn: I only mean the readability of this source module by itself.

swarm/src/handler/from_fn.rs
rkuhn: Right, old events are useless. Perhaps I am overthinking this, my worry is that there may be a situation where only the first (few) connections are ever notified because updates keep coming in — a fairness concern. This would be solved by randomising the notification order, or by always finishing one batch of notifications (with values updated during the process!) before starting the next. The second is deterministic but more work to implement.

Getting these things wrong usually has low chance of severe impact, but when the impact happens it is really difficult for downstream code to find the bug.

thomaseizinger: I think that we currently drain events from the `NetworkBehaviour` before we inject new ones into it so I don't think that this would happen but I'd have to double check that.

swarm/src/handler/from_fn.rs
thomaseizinger: > `Fn` does not allow captured values to be modified (more precisely: mutably borrowed), even if they are owned.

If you want to mutate a captured value, you are effectively sharing the state within a connection across all stream handler factory invocations.

I think we are saying the same thing and I agree that this capability is useful :)

rkuhn: Ah, you mean “sharing along the timeline”, not “sharing in different places”. Yes, now we both know that we are in agreement :-)

thomaseizinger:
I think we can close this as won't-merge. It was an interesting exploration though!

PR #2855, opened 2022-08-30
core/upgrade/: Add `ReadyUpgrade`

thomaseizinger:
# Description

This is the counterpart to `PendingUpgrade`.

Extracted out of https://github.com/libp2p/rust-libp2p/pull/2852.

## Change checklist


core/CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+# 0.35.2 [unreleased]
+
+- Add `ReadyUpgrade`. See [PR XXXX].
+
+[PR XXXX]: https://github.com/libp2p/rust-libp2p/pull/XXXX
```
```suggestion
- Add `ReadyUpgrade`. See [PR 2855].

[PR 2855]: https://github.com/libp2p/rust-libp2p/pull/2855
```

thomaseizinger:
I am planning to use this throughout the codebase where applicable. Should I do this as part of this PR or send separate ones @mxinden? It could for example be used in the ping protocol.

mxinden (commented):
I am in favor of introducing a `ReadyUpgrade`.

Note that long term I would like to get rid of the upgrade mechanism in `ConnectionHandler`, though that still needs more thought.

> I am planning to use this throughout the codebase where applicable. Should I do this as part of this PR or send separate ones @mxinden? It could for example be used in the ping protocol.

I think every pull request should be:

- As small as possible.
- Atomic.
- Leave `master` in the same or better state.

For the sake of the latter, concretely here for the sake of code deduplication, I think this pull request should include the changes across the many use-cases.

thomaseizinger:
@mxinden The only place where it was directly applicable was the ping protocol.

protocols/ping/src/protocol.rs
```diff
@@ -52,35 +49,6 @@ pub struct Ping;
...
-impl OutboundUpgrade<NegotiatedSubstream> for Ping {
-    type Output = NegotiatedSubstream;
-    type Error = Void;
-    type Future = future::Ready<Result<Self::Output, Self::Error>>;
-
-    fn upgrade_outbound(self, stream: NegotiatedSubstream, _: Self::Info) -> Self::Future {
-        future::ok(stream)
-    }
-}
-
```
mxinden, line 83: _Emphasizing this, though I think we are of the same opinion:_ While not de-duplicating given that `ReadyUpgrade` is only used once, I do think these kind of pull requests do a great job simplifying the business logic, here `libp2p-ping`, moving auxiliary not-directly-related logic to common places, here `libp2p-core`.

protocols/ping/src/handler.rs
```diff
@@ -18,10 +18,11 @@
 // FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
 // DEALINGS IN THE SOFTWARE.
 
-use crate::protocol;
+use crate::{protocol, PROTOCOL_NAME};
 use futures::future::BoxFuture;
 use futures::prelude::*;
 use futures_timer::Delay;
+use libp2p_core::upgrade::ReadyUpgrade;
```
mxinden, line 25: Given that `libp2p-ping` now depends on `libp2p-core` `v0.35.2`, the `libp2p-ping` `Cargo.toml` needs to be adjusted, right?

mxinden (commented):
Thanks for the follow up!

protocols/ping/src/protocol.rs
thomaseizinger: Yes, strong agreement here.

protocols/ping/src/handler.rs
thomaseizinger: Good catch!

thomaseizinger:
Comments addressed, ready to merge from my end @mxinden!

mxinden (approved):

PR #2857, opened 2022-08-30
build(deps): Bump rand to 0.8 and quickcheck to 1

### LINKED ISSUE #2732: Consider replacing quickcheck with an alternative
Libp2p is still on rand `0.7` (sometimes `0.7.2`, sometimes `0.7.3`). In an effort to reduce the number of duplicate dependencies in our project, we'd like to bump this to rand `0.8`.

It looks like this is stuck on the [quickcheck](https://github.com/libp2p/rust-libp2p/pull/1929#issuecomment-770762317) bump from `0.9` to `1.0`. This change removed `rand` as a public dependency, but has caused the `.gen_range()` method in `quickcheck` to become hidden.

There have been a couple of issues on the quickcheck repo requesting a new `gen_range` function: https://github.com/BurntSushi/quickcheck/issues/279, https://github.com/BurntSushi/quickcheck/issues/267

The crate author has suggested that at one point quickcheck will expose the seed of the rng, so a crate can use that to seed their own generator: https://github.com/BurntSushi/quickcheck/issues/279#issuecomment-800639203. This seems like a complicated way to implement `Arbitrary` for structs that contain ranges, but it is technically possible. This change has not been implemented at moment.

Someone [suggested](https://github.com/BurntSushi/quickcheck/issues/279#issuecomment-1154894360) a workaround:

```rs
pub fn range(gen: &mut Gen, range: Range<u64>) -> u64 {
    u64::arbitrary(gen) % (range.end - range.start) + range.start
}
```

However there does not seem to be any progress on this, or any updates to quickcheck, for over a year now.

This issue is intended to start the conversation on either updating `quickcheck` or replacing it, so the `rand` dependency can be updated.

If we have a plan forward I'm willing to make a PR to implement it.

### LINKED ISSUE #2847: rand:wasm-bindgen missing feature
https://github.com/libp2p/rust-libp2p/blob/d92cab8581b634511a0aa54e661570b4dad31830/Cargo.toml#L61

```
$ cargo outdated
...
error: failed to select a version for `rand`.
    ... required by package `libp2p v0.47.0 (/tmp/cargo-outdatedcGcclr)`
    ... which satisfies path dependency `libp2p` of package `libp2p-swarm v0.38.0 (/tmp/cargo-outdatedcGcclr/swarm)`
    ... which satisfies path dependency `libp2p-swarm` of package `multistream-select v0.11.0 (/tmp/cargo-outdatedcGcclr/misc/multistream-select)`
    ... which satisfies path dependency `multistream-select` of package `libp2p-core v0.35.1 (/tmp/cargo-outdatedcGcclr/core)`
    ... which satisfies path dependency `libp2p-core` of package `libp2p-mplex v0.35.0 (/tmp/cargo-outdatedcGcclr/muxers/mplex)`
versions that meet the requirements `^0.8.5` are: 0.8.5

the package `libp2p` depends on `rand`, with features: `wasm-bindgen` but `rand` does not have these features.
```

kpp:
Fixes https://github.com/libp2p/rust-libp2p/issues/2847
Fixes https://github.com/libp2p/rust-libp2p/issues/2732

It would be nice if you can help me with or at least give an advise:
 
- [x] ~~Some tests fail~~
- [x] ~~I didn't implement my own gen_range but used modulus instead~~
- [x] ~~There is a lot of leaky abstractions of `rand` in `kad`~~


Cargo.toml
```diff
@@ -100,7 +100,7 @@ libp2p-yamux = { version = "0.39.0", path = "muxers/yamux", optional = true }
 multiaddr = { version = "0.14.0" }
 parking_lot = "0.12.0"
 pin-project = "1.0.0"
-rand = "0.7.3" # Explicit dependency to be used in `wasm-bindgen` feature
+rand = "0.8" # Explicit dependency to be used in `wasm-bindgen` feature
```
thomaseizinger: Given the comment, do we still need it now?

thomaseizinger: Compiles without it on my machine.

protocols/kad/src/kbucket.rs
```diff
@@ -574,7 +573,9 @@ mod tests {
     }
 
     #[test]
+    #[ignore]
     fn bucket_contains_range() {
+        return todo!("test fails randomly");
```
thomaseizinger: The overflow is caused by line 595. I am not sure what that means for the test itself but commenting that line out makes things work again.

protocols/kad/src/kbucket.rs
```diff
@@ -598,7 +599,9 @@ mod tests {
     }
 
     #[test]
+    #[ignore]
     fn rand_distance() {
+        return todo!("test fails randomly");
```
thomaseizinger: Here, the overflow is caused by the `+ 1` on line 611.

thomaseizinger: I suspect that in both of these cases, this now surfaces because the randomness returned by `thread_rng` is completely different than what `Gen` from quickcheck used to give us.

protocols/kad/src/query/peers/closest/disjoint.rs
```diff
@@ -536,8 +539,8 @@ mod tests {
     struct Target(KeyBytes);
 
     impl Arbitrary for Target {
-        fn arbitrary<G: Gen>(g: &mut G) -> Self {
-            Target(Key::from(random_peers(1, g).pop().unwrap()).into())
+        fn arbitrary(_g: &mut Gen) -> Self {
+            Target(Key::from(random_peers(1, &mut rand::thread_rng()).pop().unwrap()).into())
```
thomaseizinger: Given that we always pass the same randomness now, we can probably get rid of this argument?

protocols/kad/src/query/peers/closest/disjoint.rs
```diff
@@ -866,7 +873,9 @@ mod tests {
 
     /// Ensure [`ClosestPeersIter`] and [`ClosestDisjointPeersIter`] yield same closest peers.
     #[test]
+    #[ignore]
     fn closest_and_disjoint_closest_yield_same_result() {
+        return todo!("test fails randomly");
```
thomaseizinger: This one passes on my machine.

swarm/src/registry.rs
```diff
@@ -379,7 +378,9 @@ mod tests {
     }
 
     #[test]
+    #[ignore]
     fn score_retention() {
+        return todo!("This test fails randomly");
```
thomaseizinger: Passes for me locally.

thomaseizinger (commented):
Thank you! This is something that has been overdue for a while!

swarm/src/registry.rs
kpp:
```
---- registry::tests::score_retention stdout ----
thread 'registry::tests::score_retention' panicked at 'assertion failed: exists', swarm/src/registry.rs:407:22
thread 'registry::tests::score_retention' panicked at '[quickcheck] TEST FAILED (runtime error). Arguments: (AddressRecord { addr: "/tcp/181", score: Finite(0) }, AddressRecord { addr: "/tcp/67", score: Infinite })
Error: "assertion failed: exists"', /home/kpp/.cargo/registry/src/github.com-1ecc6299db9ec823/quickcheck-1.0.3/src/tester.rs:165:28
```

protocols/kad/src/query/peers/closest/disjoint.rs
kpp: Well, the question is: do we need to get the randomness from Gen or from Rng? If from Rng, then most of that `gen_range` changes can be reverted.

Cargo.toml
kpp: Thanks! Fixed

protocols/kad/src/query/peers/closest/disjoint.rs
thomaseizinger: I think for `random_peers`, it is completely fine to get it from `thread_rng`.

kpp: Done

swarm/src/registry.rs
thomaseizinger: Weird, let me check again!

thomaseizinger: Fails for me now too, maybe a false-positive before.

thomaseizinger: The following patch fixes it:

```patch
Index: swarm/src/registry.rs
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/swarm/src/registry.rs b/swarm/src/registry.rs
--- a/swarm/src/registry.rs	(revision cff5192c82a4f0c2039bba3ea65efa27a8af8694)
+++ b/swarm/src/registry.rs	(date 1662043501670)
@@ -380,11 +380,14 @@
     #[test]
     #[ignore]
     fn score_retention() {
-        return todo!("This test fails randomly");
         fn prop(first: AddressRecord, other: AddressRecord) -> TestResult {
             if first.addr == other.addr {
                 return TestResult::discard();
             }
+
+            if first.score == AddressScore::Finite(0) {
+                return TestResult::discard();
+            }
 
             let mut addresses = Addresses::default();
 
```

Looking at the source, I think what happens is that an `Address` with a score 0 gets added initially but it dropped once _any_ new address is added, regardless of its score. I am not sure if that is the intended behaviour and the test is wrong or vice versa.

protocols/kad/src/query/peers/closest/disjoint.rs
kpp:
```
failures:

---- query::peers::closest::disjoint::tests::closest_and_disjoint_closest_yield_same_result stdout ----
thread 'query::peers::closest::disjoint::tests::closest_and_disjoint_closest_yield_same_result' panicked at 'called `Option::unwrap()` on a `None` value', protocols/kad/src/query/peers/closest/disjoint.rs:600:87
```

kpp: Fixed

swarm/src/registry.rs
kpp: Okay, so the issue is in first:finite and other:infinite. This test checks this condition:

```
// Only finite scores push out other finite scores.
```

Which is not what we see here. For some reason infinite score pushed out a finite score.

thomaseizinger: > Which is not what we see here. For some reason infinite score pushed out a finite score.

Not really. The zero score pushes it itself out so to speak because zero scores are always removed, regardless of what you add to `Addresses`.

https://github.com/libp2p/rust-libp2p/blob/36a27738618bfa0e90c285f1d379f078d2572730/swarm/src/registry.rs#L204-L215

thomaseizinger: It may be that the earlier version of quickcheck omitted the `0` case for some reason?

kpp: > It may be that the earlier version of quickcheck omitted the 0 case for some reason?

I ran 100 tests on master with `assert!(!first.score.is_zero());` and they never fail. Seems this is an unlikely path for the earlier version. Shall we discard the test on finite=0?

thomaseizinger: Given that we pretty explicitly remove entries where the score is 0, I assume this is an oversight in the test design rather than a bug so yes, I think discarding the input for `score == 0` should be okay.

We might want to add a test that manifests the "remove 0 scores" behaviour.

kpp: Fixed

kpp:
Ready for review! @thomaseizinger thank you for helping me with this.

protocols/kad/src/kbucket.rs
```diff
@@ -590,8 +589,13 @@ mod tests {
             assert!(bucket_ref.contains(&min));
             assert!(bucket_ref.contains(&max));
 
-            assert!(!bucket_ref.contains(&Distance(min.0 - 1)));
-            assert!(!bucket_ref.contains(&Distance(max.0 + 1)));
+            if min != Distance(0.into()) {
```
thomaseizinger, line 592:
```suggestion
            if min != Distance(0.into()) { // avoid underflow
```

protocols/kad/src/kbucket.rs
```diff
@@ -590,8 +589,13 @@ mod tests {
...
+                assert!(!bucket_ref.contains(&Distance(min.0 - 1)));
+            }
+
+            if max != Distance(U256::max_value()) {
```
thomaseizinger, line 597:
```suggestion
            if max != Distance(U256::max_value()) { // avoid overflow
```

protocols/kad/src/query/peers/closest/disjoint.rs
```diff
@@ -743,11 +744,12 @@ mod tests {
...
 
     impl Arbitrary for Graph {
-        fn arbitrary<G: Gen>(g: &mut G) -> Self {
-            let mut peer_ids = random_peers(g.gen_range(K_VALUE.get(), 200), g)
-                .into_iter()
-                .map(|peer_id| (peer_id.clone(), Key::from(peer_id)))
-                .collect::<Vec<_>>();
+        fn arbitrary(g: &mut Gen) -> Self {
+            let mut peer_ids =
+                random_peers(K_VALUE.get() + (usize::arbitrary(g) % (200 - K_VALUE.get())))
```
thomaseizinger: This change makes sense now but I am a bit worried that it will be hard to follow later without the context of this version upgrade.

Given that this only affects test code, what do you think of adding a `quickcheck-ext` crate to the workspace that is set to `publish = false` where we define a utility function (Perhaps with an extension trait?) to regain the functionality of `gen_range`?

thomaseizinger (commented):
Thank you for pushing this forward!

protocols/kad/src/query/peers/closest/disjoint.rs
kpp: Well, I can do that on Monday if nobody does that today.

thomaseizinger: That would be much appreciated!

kpp: Done

kpp:
So I added an implementation of `gen_range`, however it is too generic and `<T>` sometimes is treated as `integer` which is then substituted as `i32`, so I have to fix T with something like:

```diff
-            match g.gen_range(0..5) {
+            match g.gen_range(0..5u8) {
                 0 => Message::Header(HeaderLine::V1),
                 1 => Message::NotAvailable,
                 2 => Message::ListProtocols,
```

I added a T: [Unsigned](https://docs.rs/num-traits/latest/num_traits/sign/trait.Unsigned.html) bound to catch this kind of errors at compile time.

kpp:
Also this is a very strange failed test https://github.com/libp2p/rust-libp2p/runs/8207467155?check_suite_focus=true, issued in https://github.com/libp2p/rust-libp2p/issues/2874.

core/Cargo.toml
```diff
@@ -50,8 +50,7 @@ libp2p-mplex = { path = "../muxers/mplex" }
 libp2p-noise = { path = "../transports/noise" }
 libp2p-tcp = { path = "../transports/tcp" }
 multihash = { version = "0.16", default-features = false, features = ["arb"] }
-quickcheck = "0.9.0"
-rand07 = { package = "rand", version = "0.7" }
+quickcheck = { package = "quickcheck-ext", path = "../misc/quickcheck-ext" }
```
thomaseizinger, line 51: Nice idea with reexporting `quickcheck` from the extension crate!

thomaseizinger (approved):
LGTM!

@mxinden Can you have a look at this too?

thomaseizinger:
Excuse the bad merge commit, I've fixed the build error in [25f87c5](https://github.com/libp2p/rust-libp2p/pull/2857/commits/25f87c5232a39a313aa343396c963015a9b8fc12).

muxers/mplex/Cargo.toml
```diff
@@ -18,7 +18,7 @@ libp2p-core = { version = "0.36.0", path = "../../core", default-features = fals
 log = "0.4"
 nohash-hasher = "0.2"
 parking_lot = "0.12"
-rand = "0.7"
+rand = "0.8"
```
mxinden, line 21: Given that this is a `dependency` and not a `dev-dependency` and given that the current `libp2p-mplex` version (`v0.36.0`) is released, this pull request needs to bump the patch version of `libp2p-mplex` to `v0.36.1`.

Please do the following:

- Bump version in `muxers/mplex/Cargo.toml`
- Add entry to `muxers/mplex/CHANGELOG.md
- Update `libp2p-mplex` in `Cargo.toml`
- Add changelog entry in `CHANGELOG.md`

protocols/kad/Cargo.toml
```diff
@@ -21,7 +21,7 @@ log = "0.4"
 libp2p-core = { version = "0.36.0", path = "../../core", default-features = false }
 libp2p-swarm = { version = "0.39.0", path = "../../swarm" }
 prost = "0.11"
-rand = "0.7.2"
+rand = "0.8"
```
mxinden, line 24: Pull request needs to bump `libp2p-kad` to `v0.40.1`. See comment on `libp2p-mplex` above.

protocols/kad/src/query/peers/closest/disjoint.rs
```diff
@@ -773,11 +774,11 @@ mod tests {
 
             // Make each peer aware of a random set of other peers within the graph.
             for (peer_id, peer) in peers.iter_mut() {
-                peer_ids.shuffle(g);
+                peer_ids.shuffle(&mut rand::thread_rng());
```
mxinden: This makes the `Arbitrary` implementation on `Graph` no longer deterministic based on `Gen`, correct? Can we stay deterministic?

protocols/ping/Cargo.toml
```diff
@@ -17,7 +17,7 @@ instant = "0.1.11"
 libp2p-core = { version = "0.36.0", path = "../../core", default-features = false }
 libp2p-swarm = { version = "0.39.0", path = "../../swarm" }
 log = "0.4.1"
-rand = "0.7.2"
+rand = "0.8"
```
mxinden, line 20: `libp2p-ping` version bump required. See comments above.

protocols/request-response/Cargo.toml
```diff
@@ -18,7 +18,7 @@ instant = "0.1.11"
 libp2p-core = { version = "0.36.0", path = "../../core", default-features = false  }
 libp2p-swarm = { version = "0.39.0", path = "../../swarm" }
 log = "0.4.11"
-rand = "0.7"
+rand = "0.8"
```
mxinden, line 21: Version bump required. See comments above.

swarm/Cargo.toml
```diff
@@ -19,7 +19,7 @@ instant = "0.1.11"
 libp2p-core = { version = "0.36.0", path = "../core", default-features = false }
 log = "0.4"
 pin-project = "1.0.0"
-rand = "0.7"
+rand = "0.8"
```
mxinden, line 22: Version bump required. See comments above.

transports/pnet/Cargo.toml
```diff
@@ -15,8 +15,8 @@ futures = "0.3.1"
 log = "0.4.8"
 salsa20 = "0.9"
 sha3 = "0.10"
-rand = "0.7"
+rand = "0.8"
```
mxinden, line 18: Version bump required. See comments above.

mxinden (commented):
Solid work. Thanks @kpp for solving this long standing issue in such a clean way.

kpp:
@mxinden I bumped the versions of crates and added CHANGELOG entries. pnet changelog should reflect the date of publishing.

protocols/kad/src/query/peers/closest/disjoint.rs
kpp: Still there is a `random_peers` with rand. So maybe there is a way to stay deterministic, but I have no idea how to achieve this in an easy way.

muxers/mplex/CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+# 0.36.1
```
```suggestion
# 0.36.1 [unreleased]
```

protocols/floodsub/Cargo.toml
```diff
@@ -18,7 +18,7 @@ libp2p-core = { version = "0.36.0", path = "../../core", default-features = fals
 libp2p-swarm = { version = "0.39.0", path = "../../swarm" }
 log = "0.4"
 prost = "0.11"
-rand = "0.7"
+rand = "0.8"
```
mxinden, line 21: As far as I can tell, this is missing a version bump and a changelog entry, right?

protocols/gossipsub/Cargo.toml
```diff
@@ -17,7 +17,7 @@ bytes = "1.0"
 byteorder = "1.3.4"
 fnv = "1.0.7"
 futures = "0.3.5"
-rand = "0.7.3"
+rand = "0.8"
```
mxinden, line 20: Needs a version bump and a changelog entry, otherwise this might not be released any time soon.

protocols/kad/CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+# 0.40.1
```
```suggestion
# 0.40.1 [unreleased]
```

protocols/kad/src/kbucket.rs
```diff
@@ -600,6 +606,7 @@ mod tests {
     #[test]
     fn rand_distance() {
         fn prop(ix: u8) -> bool {
+            let ix = ix.min(254); // avoid overflow in 2^ix
```
mxinden: The overflow is not in `2^ix` but in `2^(ix +1)`, right? How about enforcing the bound below, i.e. in case it is `2^(255+1) - 1` do `u256::max` instead?

protocols/kad/src/query/peers/closest.rs
```diff
@@ -483,7 +483,8 @@ mod tests {
     use rand::{rngs::StdRng, Rng, SeedableRng};
     use std::{iter, time::Duration};
 
-    fn random_peers<R: Rng>(n: usize, g: &mut R) -> Vec<PeerId> {
+    fn random_peers(n: usize) -> Vec<PeerId> {
+        let mut g = rand::thread_rng();
```
mxinden: With this change the `random_peers` function is no longer deterministic over the provided random number generator, thus making debugging the tests below harder.

How about replacing the `g.gen::<[u8; 32>` below, or seeding a `rand::Rng` with a deterministic number from `g: Gen`?

protocols/kad/src/query/peers/closest/disjoint.rs
```diff
@@ -452,16 +452,16 @@ mod tests {
...
+            let num_closest_iters = g.gen_range(0..20 + 1);
+            let peers = random_peers(g.gen_range(0..20 * num_closest_iters + 1));
 
             let iters: Vec<_> = (0..num_closest_iters)
                 .map(|_| {
-                    let num_peers = g.gen_range(0, 20 + 1);
+                    let num_peers = g.gen_range(0..20 + 1);
                     let mut peers = peers
-                        .choose_multiple(g, num_peers)
+                        .choose_multiple(&mut rand::thread_rng(), num_peers)
```
mxinden: Would `Gen::choose` work here? That way the `Arbitrary` implementation is still deterministic.

protocols/kad/src/query/peers/closest/disjoint.rs
```diff
@@ -536,12 +536,13 @@ mod tests {
...
-        fn arbitrary<G: Gen>(g: &mut G) -> Self {
-            Target(Key::from(random_peers(1, g).pop().unwrap()).into())
+        fn arbitrary(_g: &mut Gen) -> Self {
+            Target(Key::from(random_peers(1).pop().unwrap()).into())
         }
     }
 
-    fn random_peers<R: Rng>(n: usize, g: &mut R) -> Vec<PeerId> {
+    fn random_peers(n: usize) -> Vec<PeerId> {
+        let mut g = rand::thread_rng();
```
mxinden: Same as above, can this be done deterministically?

protocols/kad/src/query/peers/closest/disjoint.rs
mxinden: `random_peers` just needs a `[u8; 32]`, right? That can be generated with `Gen`, right?

protocols/ping/CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+# 0.39.1
```
mxinden:
```suggestion
# 0.39.1 [unreleased]
```

protocols/request-response/CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+# 0.21.1
```
mxinden:
```suggestion
# 0.21.1 [unreleased]
```

swarm/CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+# 0.39.1
```
```suggestion
# 0.39.1 [unreleased]
```

transports/pnet/CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+# 0.22.1 [Unreleased]
```
mxinden:
```suggestion
# 0.22.1 [unreleased]
```

mxinden (commented):
Thanks for the follow ups. Still a couple of comments. Keeping the `libp2p-kad` tests deterministic would ease debugging in the future by a lot.

protocols/kad/src/query/peers/closest/disjoint.rs
kpp: Alright, this can be done deterministically, let me play with this a little bit more. I didn't expect this issue to take so long but it looks like I have to choice

protocols/kad/src/query/peers/closest/disjoint.rs
kpp: We need a non-repeating random list.

protocols/kad/src/kbucket.rs
kpp: Sorry, I don't get it.

kpp:
@mxinden I made kad tests deterministic.

protocols/floodsub/CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+# 0.39.1 [unreleased]
+
+- Bump rand to 0.8 and quickcheck to 1. See [PR 2857].
```
thomaseizinger, line 3: In previous versions, we just had an entry `Update dependencies`. See v0.33.0 for example. Given that most of your update happen through dependabot, I think it is fine to stay on such a generic level.

Having said that, feel free to also leave this if you don't want to change it back again.

@mxinden What do you think?

protocols/kad/src/query/peers/closest/disjoint.rs
```diff
@@ -447,21 +447,21 @@ mod tests {
...
 
             let iters: Vec<_> = (0..num_closest_iters)
                 .map(|_| {
-                    let num_peers = g.gen_range(0, 20 + 1);
-                    let mut peers = peers
-                        .choose_multiple(g, num_peers)
+                    let num_peers = g.gen_range(0..20 + 1);
+                    let mut peers = g
+                        .choose_multiple(&peers, num_peers)
+                        .into_iter()
```
thomaseizinger: In `choose_multiple` we already had an iterator. Perhaps we can expose that iterator directly instead of `collect`ing first and then making an iterator again?

thomaseizinger (approved):
LGTM!

thomaseizinger (approved):
This is good to merge from my side. Thank you @kpp!

I think the only open point is https://github.com/libp2p/rust-libp2p/pull/2857#discussion_r972171518, plus there is now a merge conflict with master.

protocols/kad/src/kbucket.rs
mxinden: Would the following do the trick @kpp?

``` diff
modified   protocols/kad/src/kbucket.rs
@@ -605,7 +605,10 @@ mod tests {
             let b = U256::from(2);
             let e = U256::from(ix);
             let lower = b.pow(e);
-            let upper = b.pow(e + U256::from(1)) - U256::from(1);
+            let upper = b
+                .checked_pow(e + U256::from(1))
+                .map(|x| x - U256::from(1))
+                .unwrap_or(U256::MAX);
             lower <= n && n <= upper
         }
         quickcheck(prop as fn(_) -> _);
```

protocols/floodsub/CHANGELOG.md
mxinden: I am fine either way.

protocols/kad/src/kbucket.rs
kpp: Now I see. Thank you, done.

thomaseizinger:
All comments are resolved and tests are passing. Going ahead with merging this. Interoperability tests fail because of https://github.com/libp2p/test-plans/pull/41.

PR #2859, opened 2022-08-31
swarm: Split off "keep alive" functionality from `DummyConnectionHandler`

### LINKED ISSUE #2778: swarm/src/behaviour: Add KeepAlive Behaviour

thomaseizinger:
# Description

Previously, the `DummyConnectionHandler` offered a "keep alive" functionality, i.e. it allowed users to set the value of what is returned from `ConnectionHandler::keep_alive`. This handler is primarily used in tests or `NetworkBehaviour`s that don't open any connections (like mDNS). In all of these cases, it is statically known whether we want to keep connections alive. As such, this functionality is better represented by a static `KeepAliveConnectionHandler` that always returns `KeepAlive::Yes` and a `DummyConnectionHandler` that always returns `KeepAlive::No`.

To follow the naming conventions described in https://github.com/libp2p/rust-libp2p/issues/2217, we introduce a top-level `keep_alive` and `dummy` behaviour in `libp2p-swarm` that contains both the `NetworkBehaviour` and `ConnectionHandler` implementation for either case.

- https://github.com/libp2p/rust-libp2p/pull/2852: To actually replace "Ping"'s `ConnectionHandler` with a `FromFn` `ConnectionHandler`, we need to decouple the "Keep Alive" functionality from the Ping protocol. If this PR is accepted, I'd send a follow-up to remove the keep-alive functionality from ping in favor of using the `KeepAliveBehaviour`.

## Open Questions

- ~Implementations on `()` are not as easy to discover. Should we retain `DummyBehaviour` and `DummyConnectionHandler` but strip the keep-alive logic from them?~
- ~Should we rename the `KeepAlive` to `KeepAlivePolicy` (or something else) to avoid ambiguity with `behaviour::KeepAlive`?~
- ~Should we strip the keep alive functionality from `libp2p-ping` as part of this PR or in a follow-up?~

## Change checklist


mxinden:
> Implementations on `()` are not as easy to discover. 

I have to give this more thought. I agree that it makes it harder to discover. Also I don't find the fact that `()` implements a _dummy_ `NetworkBehaviour` intuitive.

> Should we retain `DummyBehaviour` and `DummyConnectionHandler` but strip the keep-alive logic from them?

I think this is a good idea.

thomaseizinger:
> > Should we retain `DummyBehaviour` and `DummyConnectionHandler` but strip the keep-alive logic from them?
> 
> I think this is a good idea.

I'll progress in this direction then!

thomaseizinger:
This is ready for another review @mxinden !

swarm/src/lib.rs
```diff
@@ -1508,38 +1510,44 @@ impl error::Error for DialError {
...
+        _: &mut impl PollParameters,
+    ) -> Poll<NetworkBehaviourAction<Self::OutEvent, Self::ConnectionHandler>> {
+        Poll::Pending
     }
 }
 
-impl NetworkBehaviour for DummyBehaviour {
-    type ConnectionHandler = handler::DummyConnectionHandler;
+/// Implementation of [`NetworkBehaviour`] that doesn't do anything other than keep all connections alive.
+pub struct DummyNetworkBehaviour;
```
mxinden: Given that we call the `Toggle` `NetworkBehaviour` implementation just `Toggle`, how about moving this one to `swarm/src/behaviour/dummy.rs` and calling it just `Dummy`? That would be in line with our naming conventions, right?

swarm/src/lib.rs
```diff
@@ -1508,38 +1510,44 @@ impl error::Error for DialError {
     }
 }
 
-/// Dummy implementation of [`NetworkBehaviour`] that doesn't do anything.
-#[derive(Clone)]
-pub struct DummyBehaviour {
-    keep_alive: KeepAlive,
-}
+/// Implementation of [`NetworkBehaviour`] that doesn't do anything other than keep all connections alive.
+pub struct KeepAliveBehaviour;
```
mxinden: Same as the comment on `DummyNetworkBehaviour`, how about calling this just `KeepAlive` and moving it into `swarm/src/behaviour/keep_alive.rs`?

swarm/src/lib.rs
```diff
@@ -1508,38 +1510,44 @@ impl error::Error for DialError {
...
+
+impl NetworkBehaviour for KeepAliveBehaviour {
+    type ConnectionHandler = KeepAliveConnectionHandler;
+    type OutEvent = void::Void;
 
-impl DummyBehaviour {
-    pub fn with_keep_alive(keep_alive: KeepAlive) -> Self {
```
mxinden, line 1524: Just for the record, I do **not** think we need to role this out in two steps, (1) deprecating it and (2) removing it.

mxinden (commented):
Thanks for cleaning this up. Small suggestions on the naming of the two `NetworkBehaviour` implementations.

mxinden:
Can you update the pull request description?

swarm/src/lib.rs
thomaseizinger: Our naming conventions aren't crystal clear here :)

There is somewhat of a naming clash between `behaviour::KeepAlive` and the actual `KeepAlive` enum now. I think we can rename the latter one to `KeepAlivePolicy` to make that clearer. This would also allow us to rename the `KeepAliveConnectionHandler` to just `KeepAlive` and have the module name differentiate between the two.

thomaseizinger:
Should we strip `libp2p-ping`'s keep-alive functionality as part of this PR too or do that in a follow-up PR?

thomaseizinger:
> Can you update the pull request description?

Updated. Also added new questions.

mxinden:
> Should we strip the keep alive functionality from `libp2p-ping` as part of this PR or in a follow-up?

Shall we follow the two step deprecation path as we did in the recent path?

swarm/src/lib.rs
mxinden: Good point on the clash between `KeepAlive` _policy_ and _behaviour_.

I don't feel strongly about this. I tend towards keeping the names as is. `KeepAlive` is widely used, the `KeepAliveBehaviour` is only used in test code. Thus I don't think this justifies a breaking change.

mxinden (dismissed):
Looks good to me once the merge conflicts are resolved.

thomaseizinger:
> > Should we strip the keep alive functionality from `libp2p-ping` as part of this PR or in a follow-up?
> 
> Shall we follow the two step deprecation path as we did in the recent path?

Yeah that is a good idea!

examples/ping.rs
```diff
@@ -82,3 +75,13 @@ async fn main() -> Result<(), Box<dyn Error>> {
         }
     }
 }
+
+/// Our network behaviour.
+///
+/// For illustrative purposes, this includes the `KeepAlive` behaviour so a continuous sequence of
```
mxinden:
```suggestion
/// For illustrative purposes, this includes the [`KeepAlive`](behaviour::KeepAlive) behaviour so a continuous sequence of
```

To make sure this doesn't become outdated in the future.

misc/metrics/examples/metrics/main.rs
```diff
@@ -104,3 +103,13 @@ fn main() -> Result<(), Box<dyn Error>> {
     });
     Ok(())
 }
+
+/// Our network behaviour.
+///
+/// For illustrative purposes, this includes the `KeepAlive` behaviour so the ping actually happen
```
mxinden: Same here.

protocols/ping/src/handler.rs
```diff
@@ -111,6 +111,10 @@ impl Config {
     /// If the maximum number of allowed ping failures is reached, the
     /// connection is always terminated as a result of [`ConnectionHandler::poll`]
     /// returning an error, regardless of the keep-alive setting.
+    #[deprecated(
+        since = "0.40.0",
+        note = "Use `libp2p::swarm::behaviour::KeepAlive` if you need to keep connections alive unconditionally."
+    )]
```
mxinden, line 117: :pray:

mxinden (approved):
Thanks for the follow-ups. Apart from the merge conflicts, this looks great to me.

examples/ping.rs
thomaseizinger: I don't think docs run on examples but we can still do this :)

protocols/ping/tests/ping.rs
```diff
@@ -266,3 +268,18 @@ impl Arbitrary for MuxerChoice {
...
+struct Behaviour {
+    keep_alive: behaviour::KeepAlive,
+    ping: ping::Behaviour,
+}
+
+impl Behaviour {
+    fn new(config: ping::Config) -> Self {
+        Self {
+            keep_alive: behaviour::KeepAlive,
+            ping: ping::Behaviour::new(config),
```
thomaseizinger: Hmm, I just realised that the naming here is inverted. Should it perhaps be `keep_alive::Behaviour` and `dummy::Behaviour`?

thomaseizinger: Overall, the modules would be: `libp2p::swarm::dummy::{Behaviour,ConnectionHandler}` and possibly more. I think that would work well because it groups all "dummy" things together which is how they are likely to be used instead of grouping them by "layer", i.e. behaviour, connection handler etc.

thomaseizinger: Perhaps we can even have a `libp2p-dummy` crate? :)

thomaseizinger: > Perhaps we can even have a `libp2p-dummy` crate? :)

For now we could emulate this in the root crate by re-exporting `libp2p::swarm::dummy` as `libp2p::dummy`.

mxinden: I prefer `libp2p_swarm::behaviour::keep_alive::Behaviour` and `libp2p_ping::Behaviour`.

That said, I don't feel strongly about it. I mostly just care about consistency.

mxinden (approved):
This is ready to merge from my end. Feel free to merge in case you want to keep the naming as is or want that to happen in a follow-up @thomaseizinger.

protocols/ping/tests/ping.rs
thomaseizinger: Might as well do it properly then :)

swarm/src/lib.rs
thomaseizinger: This is resolved with the latest rename.

thomaseizinger:
> This is ready to merge from my end. Feel free to merge in case you want to keep the naming as is or want that to happen in a follow-up @thomaseizinger.

I ended up introducing top-level `dummy` and `keep_alive` modules in `libp2p_swarm` that export both, the `ConnectionHandler` and the corresponding `NetworkBehaviour`.

thomaseizinger:
Changelog and PR description are up to date. This is ready to be merged from my end.

mxinden:
@thomaseizinger can you take a look at the clippy failure?

```
error: unused import: `NetworkBehaviour`
Error:   --> examples/ping.rs:44:21
   |
44 | use libp2p::swarm::{NetworkBehaviour, Swarm, SwarmEvent};
   |                     ^^^^^^^^^^^^^^^^
   |
   = note: `-D unused-imports` implied by `-D warnings`
   ```

**Comment by thomaseizinger** [comment]
> @thomaseizinger can you take a look at the clippy failure?
> 
> ```
> error: unused import: `NetworkBehaviour`
> Error:   --> examples/ping.rs:44:21
>    |
> 44 | use libp2p::swarm::{NetworkBehaviour, Swarm, SwarmEvent};
>    |                     ^^^^^^^^^^^^^^^^
>    |
>    = note: `-D unused-imports` implied by `-D warnings`
> ```

Ah yes, I expected that one to happen eventually. I originally got the warning because https://github.com/libp2p/rust-libp2p/pull/2932 wasn't yet fixed!

File: swarm/src/dummy.rs [hunk 8]
```diff
@@ -0,0 +1,104 @@
...
+use crate::handler::{InboundUpgradeSend, OutboundUpgradeSend};
+use crate::{ConnectionHandlerEvent, ConnectionHandlerUpgrErr, KeepAlive, SubstreamProtocol};
+use libp2p_core::connection::ConnectionId;
+use libp2p_core::upgrade::DeniedUpgrade;
+use libp2p_core::PeerId;
+use libp2p_core::UpgradeError;
+use std::task::{Context, Poll};
+use void::Void;
+
+/// Implementation of [`NetworkBehaviour`] that doesn't do anything other than keep all connections alive.
```
mxinden: It does not keep connections alive, right?

mxinden (approved):
Ready to merge from my end otherwise.

swarm/src/dummy.rs
thomaseizinger: Absolutely correct. Thanks for spotting the typo!

Issue #1396, opened 2020-01-19
ring is a bad dependency; get rid of it


leo-lb:
Hello,
I'd like to suggest the removal of `ring` as a dependency all-together. That also includes rustls and webpki. (rustls is used by the websocket transport)

Why:
- Because I don't see a point in using that one vs others when it mostly is a wrapper to Assembly code: 
![Screenshot Capture - 2020-01-19 - 04-09-14](https://user-images.githubusercontent.com/44934014/72673965-85e03400-3a71-11ea-948d-822fe38721d7.png)
- The maintainer is problematic: https://github.com/briansmith/ring/issues/774 and doesnt work in the spirit of Free Software.
- There is no portable fallback for crypto primitives, which means it wont compile on less common platforms, at the mercy of the problematic maintainer.
- The build system of `ring` is really a mess.

I suggest using the `openssl` crate instead. That supports both OpenSSL and LibreSSL and any other library that offers an openssl-compatible interface. Or individual crates that implement the primitives needed by `libp2p` such as: https://github.com/RustCrypto -- but I don't think they're that mature yet.

I started porting `libp2p` crates to `openssl` but it seems that `ring` types are exposed in public interfaces, so that would be a breaking change. Also, usage of `ring` really doesnt seem contained, so it's not so trivial to do so either.

burdges:
We're a couple years away from any viable ring and rustls alternatives I think, assuming some qualified crypto devs begin work tomorrow. 

Avoid openssl.  It's poorly written and dangerous.  I'd expect some yanked ring versions prevent vulnerabilities, but maybe nobody really knows. 

We need TLS for transport layer encryption and authentication, which means rustls.  SECIO is far worse and should be turned off asap.  We need UDP for our parachain piece distribution eventually, maybe some analog of Bittorrent's uTP.  We'll want QUIC for UDP protocols, meaning more TLS.  I'd think all three Rust QUIC implementations use rustls, but the better one quinn definitely does.

RustCrypto provides only symmetric primitives, but not asymmetric primitives, and definitely not the painful and dangerous "agile" crypto demanzded by TLS.  All interfaces in RustCrypto require const generics before stabilization.  I'm unsure about side channel protections and performance in RustCrypto, well all that crypto gets written in assemble for good reasons.  I hear Mozilla should produce RFCs this year for Rust and LLVM support for constant-time code, not sure about the performance impact.

[...]

PR #2860, opened 2022-08-31
core/: Introduce `rsa` feature flag to avoid `ring` dependency

GamePad64:
# Description

I've tried to build libp2p for router, running mips cpu, but the build has failed because of ring dependency. Ring is used mainly for RSA in libp2p-code, so I've made rsa as an optional feature

#1396
#1975

## Change checklist


core/CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+# 0.35.2
```
thomaseizinger: Please also change the version in `Cargo.toml`! :)

thomaseizinger:
```suggestion
# 0.35.2 [unreleased]
```

thomaseizinger: Plus, I think this is a breaking change because someone using `libp2p-core` with `default-features = false` may now have a broken build if they match exhaustively on the `Keypair` enum.

thomaseizinger (commented):
Generally, I am happy to land this change. Please see inline comments :)

core/CHANGELOG.md
GamePad64: Yes, this is a breaking change, so I've bumped libp2p-core version to 0.36.0. Also, fixed tests.

core/src/identity/error.rs
```diff
@@ -67,13 +67,15 @@ pub struct SigningError {
 
 /// An error during encoding of key material.
 impl SigningError {
+    #[allow(dead_code)]
```
thomaseizinger: When does this trigger? Do we only use this function with the `rsa` feature?

thomaseizinger (approved):
Thank you for doing all the version updates and changelog entries!

One more question and I'd like to have @mxinden's review on this as well before we merge it.

thomaseizinger:
> I've tried to build libp2p for router, running mips cpu, but the build has failed because of ring dependency. Ring is used mainly for RSA in libp2p-code, so I've made rsa as an optional feature

One more note on this, I am pretty sure `libp2p-quic` - once merged - will also bring in a dependency on `ring` again. Using QUIC is obviously optional but I just wanted to mention that RSA keys are not the only thing that ties us to `ring`.

core/src/identity/error.rs
GamePad64: SigningError used in `rsa` and `secp256k1` features. So with these features disabled, these functions become dead code. Probably, `#[cfg(any(feature = "secp256k1", feature = "rsa"))]` would be better here.

GamePad64:
`libp2p-core` is a central part of libp2p, so removing this problematic dependency from it would be much better. `libp2p-quic` is not a required transport, so we can just continue to use `libp2p-tcp` :wink: 
The biggest problem with `ring`, that it supports only x86 and arm in 32- and 64-bit variants. Using it denies building your apps for embedded targets, supported in Rust (like routers and such).

core/src/identity/error.rs
thomaseizinger: I think so too, can you change it to the above feature combination?

GamePad64: Ok, replaced `allow(dead_code)` with `cfg(feature=...)`

core/Cargo.toml
```diff
@@ -59,9 +59,10 @@ serde_json = "1.0"
 prost-build = "0.11"
 
 [features]
-default = [ "secp256k1", "ecdsa" ]
+default = [ "secp256k1", "ecdsa", "rsa" ]
```
thomaseizinger: In the changelog, we say people need to enable the feature for `rsa` support but we enable it as a default feature here. Given that we are already making a breaking change, I think I'd prefer to not enable this as the default.

core/Cargo.toml
```diff
@@ -59,9 +59,10 @@ serde_json = "1.0"
...
 secp256k1 = [ "libsecp256k1" ]
 ecdsa = [ "p256" ]
+rsa = [ "ring" ]
```
thomaseizinger:
```suggestion
rsa = [ "dep:ring" ]
```

This should hide the `ring` feature that we are exposing through the new optional dependency.

transports/noise/src/protocol/x25519.rs
```diff
@@ -120,6 +120,7 @@ impl Protocol<X25519> for X25519 {
         Ok(PublicKey(X25519(pk)))
     }
 
+    #[allow(irrefutable_let_patterns)]
```
thomaseizinger, line 123: This is unfortunate :/

thomaseizinger (approved):
Thanks!

One more note regarding using the new `dep` syntax.

core/Cargo.toml
GamePad64: Ok, removed

core/Cargo.toml
GamePad64: Done!

transports/noise/src/protocol/x25519.rs
GamePad64: Yes, but I don't know how to change this without a massive rewrite :(

thomaseizinger:
Thanks for adopting the suggestions so quickly. Just waiting for @mxinden now :)

mxinden (commented):
Thanks for the patch and thanks for following this all the way through, updating the many versions :pray:.

This looks good to me apart from one thing.

The `libp2p-core` feature `secp256k1`  is passed through to `libp2p`. In addition the `secp256k1` is part of the `libp2p` `default` feature.

According to the libp2p specification:

> Implementations MUST support Ed25519. Implementations SHOULD support RSA if they wish to interoperate with the mainline IPFS DHT and the default IPFS bootstrap nodes. Implementations MAY support Secp256k1 and ECDSA, but nodes using those keys may not be able to connect to all other nodes.

https://github.com/libp2p/specs/blob/master/peer-ids/peer-ids.md#key-types

I read this as `rsa` taking a higher priority as `secp256k1`. Thus I suggest passing the `rsa` feature through `libp2p` and adding `rsa` to the `default` feature on `libp2p`. That would as well remove the breaking change, i.e. the removal of default support for decoding `rsa` keys.

@GamePad64 @thomaseizinger does the above reasoning make sense?

GamePad64:
Yes, this makes sense to me :) Added `rsa` feature to base libp2p crate and added it to `default`.

thomaseizinger (approved):
@mxinden That sounds good to me.

It is a breaking change regardless though. A user depending on the crate with disabled default features will not get RSA support whereas previously, they did.

mxinden (approved):
Thanks for the follow-ups!

mxinden:
> It is a breaking change regardless though. A user depending on the crate with disabled default features will not get RSA support whereas previously, they did.

:+1: I will point this out in the release notes on https://github.com/libp2p/rust-libp2p/pull/2869.

mxinden:
:tada: first pull request running the interop tests https://github.com/libp2p/rust-libp2p/pull/2835 //CC @laurentsenta.

PR #2861, opened 2022-08-31
swarm/connection: Enforce limit on inbound substreams via `StreamMuxer`

### LINKED ISSUE #2796: Enforce back-pressure on incoming substreams using new `StreamMuxer` interface
Once https://github.com/libp2p/rust-libp2p/issues/2722 is complete, I would suggest we inline the `HandlerWrapper` component into the `swarm::Connection` struct.

This will resolve some indirection which:

- Allows us to use the incoming substream limit to enforce back-pressure
- Simplifies understanding of the code because `StreamMuxer` and `ConnectionHandler` are polled by the same component (`Connection`) and directly pass data between each other instead of going through another layer.

thomaseizinger:
# Description

Inlines the `HandlerWrapper` component into `Connection` which allows us to enforce the limit of inbound streams directly down to the `StreamMuxer` level by not calling `poll_inbound` if we would exceed the limit. Depending on the muxer implementation, this may enforce backpressure all the way to the other node.

Due to this backpressure mechanism, a remote may now be slow in accepting new outbound streams. To prevent unbounded growth of "pending" streams, we change the implementation such that the "upgrade timeout" for outbound streams starts ticking from the moment the stream is requested instead of only from when the stream is opened and we start the upgrade.

Fixes #2796.

## Open tasks

- [x] Find out why we had to adjust the `Connection::poll` implementation
- [x] Start substream upgrade timeout at the time of substream _request_ and not when the upgrade starts
- [x] Add changelog entry to yamux & bump version

## Open Questions

- How many streams should we buffer at most in `libp2p-yamux::Yamux`?
- Should the change to `libp2p-yamux` be in a separate PR?

## Change checklist


swarm/src/connection.rs
```diff
@@ -74,10 +72,39 @@ where
...
+    ///
+    /// Note: This only enforces a limit on the number of concurrently
+    /// negotiating inbound streams. The total number of inbound streams on a
+    /// connection is the sum of negotiating and negotiated streams. A limit on
+    /// the total number of streams can be enforced at the
+    /// [`StreamMuxerBox`](libp2p_core::muxing::StreamMuxerBox) level.
+    max_negotiating_inbound_streams: usize,
+    /// For each outbound substream request, how to upgrade it.
+    pending_dial_upgrades:
+        VecDeque<SubstreamProtocol<THandler::OutboundProtocol, THandler::OutboundOpenInfo>>,
```
elenaf9: I think we should also have some sort of limit for pending outbound substreams so that this can not grow unbounded if the remote never accepts an inbound substream.

Consider the following setup: We have a (malicous) node listening on the relay, that doesn't poll any inbound substreams. If now a lot of (malicous) nodes try to connect to this node via the relay, the relay will push a new `SubstreamProtocol` with a `CircuitReq` to this channel each time, resulting in unbounded growth.

E.g. we could have `max_negotiating_outbound_streams` and if we exceed the limit we pop the front item and call `inject_dial_upgrade_error`. Wdyt?

thomaseizinger: To clarify, the scenario you are thinking of is that `StreamMuxer::poll_outbound` returns `Poll::Pending` basically all the time and thus the timeout we have on the upgrades never triggers because we never even start the upgrade?

For muxers like yamux and mplex, this won't happen I believe because the remote can't enforce backpressure but with QUIC and later qmux, this could indeed happen.

We could directly check this limit on line 176, no need to push/pop items.

elenaf9: Yes that's what I meant :)

> We could directly check this limit on line 176, no need to push/pop items.

Depends on whether on a full queue we want to drop old pending substreams in favor of new ones or not. 
I am leaning towards the former since old substreams will most likely have timeout-ed anyway even if then the remote starts accepting streams.

Alternative idea: the `pending_dial_upgrades`  could also already be a `FuturesUnordered` where just the `timeout` is polled. 
Since `FuturesUnordered` implements `Iterator` we can still just pop the first item if a new outbound substream is ready like we currently do. But additionally we can then also poll `pending_dial_upgrades` to remove any `SubstreamUpgrade`s that may have already timed out.

thomaseizinger: > > We could directly check this limit on line 176, no need to push/pop items.
> 
> Depends on whether on a full queue we want to drop old pending substreams in favor of new ones or not. I am leaning towards the former since old substreams will most likely have timeout-ed anyway even if then the remote starts accepting streams.

What do you mean with "has timed out"? Currently, rust-libp2p has a timeout on _upgrading_ streams, but no timeout on actually opening the underlying stream. I think it is worth considering to extend this timeout to also cover the initial opening part, especially because there has been talk about getting rid of the entire `Upgrade` machinery too.

> Alternative idea: the `pending_dial_upgrades` could also already be a `FuturesUnordered` where just the `timeout` is polled. Since `FuturesUnordered` implements `Iterator` we can still just pop the first item if a new outbound substream is ready like we currently do. But additionally we can then also poll `pending_dial_upgrades` to remove any `SubstreamUpgrade`s that may have already timed out.

If we have the timeout include the opening phase of the stream, I think it would be better to not allow new streams to be opened instead of dropping the oldest one. That would allow a friendly but simply slow remote to still open a substream eventually, otherwise the upgrade process could be partially complete and simply drop it half-way through.

elenaf9: Ah you are right. I did not read the code carefully enough and thought the queued `SubstreamProtocol` already creates the timeout `Delay` and that it is just not polled until we start the upgrading. Sorry for the confusion.

Don't really have an opinion on whether this should be changes to that the inital opening part is included in the timeout or not.

thomaseizinger: As it stands, this PR doesn't make the situation any worse on that front but provides an improvement on how we back-pressure inbound streams.

The timeout on stream upgrades is generally misleading as we've identified in this discussion (also see https://github.com/libp2p/rust-libp2p/issues/2863#issuecomment-1235656143) and I think deserves some thinking on its own but I don't want to stall this PR on it.

elenaf9: > As it stands, this PR doesn't make the situation any worse on that front but provides an improvement on how we back-pressure inbound streams.

I think it makes the situation worse because now we have a way of back-pressuring streams to the remote, but the remote does not have a way of handling this back-pressure and instead just fills up an unbounded queue.
See previous example of how an attacker could exploit this on a relay server. I think we should at least add the previously mentioned limit on pending dial upgrades to avoid this.

thomaseizinger: Thanks for outlining this again. I agree!

My suggestion would be to change the implementation such that the stream upgrade timeout already starts ticking the moment we wait for an inbound stream.

swarm/src/connection.rs
```diff
@@ -19,7 +19,6 @@
 // DEALINGS IN THE SOFTWARE.
 
 mod error;
-mod handler_wrapper;
```
mxinden, line 22: :tada:

swarm/src/connection.rs
mxinden: > My suggestion would be to change the implementation such that the stream upgrade timeout already starts ticking the moment we wait for an inbound stream.

You meant to write "we wait for an **outbound** stream", right @thomaseizinger?

swarm/src/connection.rs
```diff
@@ -161,27 +251,37 @@ where
...
+                    Poll::Ready(substream) => {
+                        let protocol = self.handler.listen_protocol();
+
+                        self.negotiating_in
+                            .push(SubstreamUpgrade::new_inbound(substream, protocol));
+
+                        continue; // Go back to the top, handler can potentially make progress again.
+                    }
                 }
             }
```
mxinden: I am afraid this leads to a deadlock with the current Yamux implementation.

`<Yamux as StreamMuxer>::poll` currently only returns `Poll::Pending`, but does not drive the underlying state machine.

https://github.com/libp2p/rust-libp2p/blob/b8c3b282eee5d6387a9cf2f64af0253e764c1172/muxers/yamux/src/lib.rs#L126-L131

The only `StreamMuxer` method driving the Yamux state machine is `<Yamux as StreamMuxer>::poll_inbound`:

https://github.com/libp2p/rust-libp2p/blob/b8c3b282eee5d6387a9cf2f64af0253e764c1172/muxers/yamux/src/lib.rs#L104-L115

Note that the Yamux state machine is the component reading and writing from/to the underlying I/O resource.

Thus, once the `self.max_negotiating_inbound_streams` limit is hit, established streams and pending outbound streams will no longer make progress.

---

**Suggestion**

Yamux does not support back-pressure on stream creation at the protocol level. Thus, the only thing a node can do is drop inbound streams. The question is where to drop inbound streams.

Ideally I would like to drop streams within https://github.com/libp2p/rust-yamux once a certain limit is hit. We could as well buffer streams in `libp2p-yamux::Yamux` and drop streams exceeding the limit there.

mxinden (commented):
Wonderful to see. I think there is more work needed before we can land this though.

mxinden:
Looking at the many simplifications that @thomaseizinger and @elenaf9  landed in the last months, I already foresee the day where rust-libp2p is a single `struct` only :smile:

swarm/src/connection.rs
thomaseizinger: Yes!

swarm/src/connection.rs
thomaseizinger: > Ideally I would like to drop streams within [libp2p/rust-yamux](https://github.com/libp2p/rust-yamux) once a certain limit is hit. We could as well buffer streams in `libp2p-yamux::Yamux` and drop streams exceeding the limit there.

We don't know about the limit in the yamux multiplexer unfortunately and it would be rather convoluted to pass the actual limit in. What do you think would an appropriate limit be that we could hardcode?

thomaseizinger: Actually, we don't need a separate limit. We can move the general `poll` to the bottom of the `poll` loop in `Connection`. That way, we can simply drop all streams that we get in `StreamMuxer::poll`. If we get there and still receive streams, it means we have hit the limit on how many inbound streams we can process.

Does this make sense?

thomaseizinger: However, this doesn't follow the convention of "prefer local over remote work".

thomaseizinger: > However, this doesn't follow the convention of "prefer local over remote work".

This makes me prefer the separate limit again. I think we can set this to something rather high because it should not be hit in normal operation.

thomaseizinger: Let me know what you think of [6b98c50](https://github.com/libp2p/rust-libp2p/pull/2861/commits/6b98c50209b59957e5b85413bc2397c2bdb83f5d).

Happy to adjust the limit!

thomaseizinger:
I had to adjust the implemenation of `Connection::poll` for the autonat (???) tests to work again. I've debugged it for a bit but I can't explain why. Any pointers are welcome.

The current `Connection::poll` implementation is not ideal because it does prioritise local over remote work.

thomaseizinger:
Hmm, this yamux change seems to break quite a few tests @mxinden :/

thomaseizinger:
This is ready for a full review @mxinden @elenaf9! PR description is up to date, please have a look.

muxers/yamux/src/lib.rs
```diff
@@ -42,8 +44,20 @@ pub struct Yamux<S> {
     incoming: S,
     /// Handle to control the connection.
     control: yamux::Control,
+    /// Temporarily buffers inbound streams in case our node is performing backpressure on the remote.
+    ///
+    /// The only way how yamux can make progress is by driving the [`Incoming`] stream. However, the
+    /// [`StreamMuxer`] interface is designed to allow a caller to selectively make progress via
+    /// [`StreamMuxer::poll_inbound`] and [`StreamMuxer::poll_outbound`] whilst the move general
```
mxinden:
```suggestion
    /// [`StreamMuxer::poll_inbound`] and [`StreamMuxer::poll_outbound`] whilst the more general
```

muxers/yamux/src/lib.rs
```diff
@@ -125,9 +139,21 @@ where
...
     ) -> Poll<Result<StreamMuxerEvent, Self::Error>> {
-        Poll::Pending
+        let this = self.get_mut();
+
+        loop {
+            let inbound_stream = ready!(this.poll_inner(cx))?;
+
+            if this.inbound_stream_buffer.len() >= MAX_BUFFERED_INBOUND_STREAMS {
+                log::warn!("dropping {inbound_stream} because buffer is full");
+                drop(inbound_stream);
```
mxinden, line 147: Just want to make sure we are making an informed decision here. Instead of tail dropping, we could as well do head dropping.

I would have to put more thoughts into this before having an opinion. This does maintain the status quo, thus we can consider it a non-change.

muxers/yamux/src/lib.rs
```diff
@@ -42,8 +44,20 @@ pub struct Yamux<S> {
...
+    /// [`StreamMuxer`] interface is designed to allow a caller to selectively make progress via
+    /// [`StreamMuxer::poll_inbound`] and [`StreamMuxer::poll_outbound`] whilst the move general
+    /// [`StreamMuxer::poll`] is designed to make progress on existing streams etc.
+    ///
+    /// This buffer stores inbound streams that are created whilst [`StreamMuxer::poll`] is called.
+    /// Once the buffer is full, new inbound streams are dropped.
+    inbound_stream_buffer: VecDeque<yamux::Stream>,
 }
 
+const MAX_BUFFERED_INBOUND_STREAMS: usize = 25;
```
mxinden, line 59: If I am not mistaken, the additional buffer of 25 streams increases the overall number of possible inflight inbound streams, correct? In other words, while we previously only supported `max_negotiating_inbound_streams` before dropping streams, we now allow up to `max_negotiating_inbound_streams + 25`.

Is this correct?

mxinden (commented):
Thanks for the follow ups on this Thomas!

muxers/yamux/src/lib.rs
thomaseizinger: With this change, the timeout for opening a new stream already begins as the request is queued. I believe head dropping would negatively affect the number of successfully negotiated streams because we might drop a stream that was about to be processed by the local node. A dropped stream will trigger a `ConnectionHandler::inject_dial_upgrade_error` with the provided "open info" on the remote node.

In a scenario where a node opens 3 streams that are all pending in this buffer, I think it may be confusing if stream 1 fails but 2 and 3 succeed. It feels more natural to process this in FIFO order.

muxers/yamux/src/lib.rs
thomaseizinger: Yep, that sounds correct to me! Like I said, this number is completely arbitrary and we can change it to something else :)

Cargo.toml
```diff
@@ -99,7 +99,7 @@ libp2p-swarm = { version = "0.39.0", path = "swarm" }
 libp2p-swarm-derive = { version = "0.30.1", path = "swarm-derive" }
 libp2p-uds = { version = "0.35.0", path = "transports/uds", optional = true }
 libp2p-wasm-ext = { version = "0.36.0", path = "transports/wasm-ext", default-features = false, optional = true }
-libp2p-yamux = { version = "0.40.0", path = "muxers/yamux", optional = true }
+libp2p-yamux = { version = "0.40.1", path = "muxers/yamux", optional = true }
```
thomaseizinger:
```suggestion
libp2p-yamux = { version = "0.40.0", path = "muxers/yamux", optional = true }
```

muxers/yamux/Cargo.toml
```diff
@@ -3,7 +3,7 @@ name = "libp2p-yamux"
 edition = "2021"
 rust-version = "1.56.1"
 description = "Yamux multiplexing protocol for libp2p"
-version = "0.40.0"
+version = "0.40.1"
```
thomaseizinger:
```suggestion
version = "0.40.0"
```

swarm/src/connection.rs
```diff
@@ -225,4 +349,456 @@ impl fmt::Display for ConnectionLimit {
...
+        user_data: UserData,
+        timeout: Delay,
+        upgrade: Upgrade,
+        /// A waker to notify our [`FuturesUnordered`] that we have extracted the data.
+        ///
+        /// This will ensure that we will get polled again in the next iteration which allows us to
+        /// resolve with `Ok(())` and be removed from the [`FuturesUnordered`].
+        extracted_waker: Option<Waker>,
+    },
+    Done,
```
elenaf9, line 470: Why is the 'Done' variant needed? Couldn't we just pop off the `SubstreamRequest` from our queue once a new outbound substream is ready?

elenaf9 (commented):
Great work!

> I had to adjust the implemenation of Connection::poll for the autonat (???) tests to work again. I've debugged it for a bit but I can't explain why. Any pointers are welcome.

I can look into that (I am the author of those tests). Those test expect a strict order in which events should occur, which is maybe not ideal. On the other hand they prove that it does make a difference in which order we poll our state machines and this can help decide what order makes the most sense.

thomaseizinger:
> > I had to adjust the implemenation of Connection::poll for the autonat (???) tests to work again. I've debugged it for a bit but I can't explain why. Any pointers are welcome.
> 
> I can look into that (I am the author of those tests). Those test expect a strict order in which events should occur, which is maybe not ideal. On the other hand they prove that it does make a difference in which order we poll our state machines and this can help decide what order makes the most sense.

That is outdated and has since been fixed. No adjustment was needed, I just had a stupid bug :D

swarm/src/connection.rs
thomaseizinger: I need this so `extract` can take `&mut self` and still return ownership without cloning. The bigger problem is that you can't remove futures from a `FuturesUnordered` so I need to modify them and make sure the return `Poll::Ready` which will get them removed!

mxinden (approved):
:pray:

mxinden:
Interoperability failures due to https://github.com/libp2p/test-plans/pull/41. Merging here.

PR #2862, opened 2022-08-31
protocols/gossipsub: Allow publishing to anything that implements `Into<TopicHash>` 

GamePad64:
# Description

I store TopicHashes for gossipsub topics in my app for performance reasons. And rehashing Topics every time I want to publish seems redundant to me.

## Change checklist

### DIFF
--- a/protocols/gossipsub/src/behaviour.rs
+++ b/protocols/gossipsub/src/behaviour.rs
@@ -586,22 +586,14 @@ where
         Ok(true)
     }
 
-    /// Publishes a message to a topic in the network.
-    pub fn publish<H: Hasher>(
+    /// Publishes a message with multiple topics to the network.
+    pub fn publish(
         &mut self,
-        topic: Topic<H>,
-        data: impl Into<Vec<u8>>,
-    ) -> Result<MessageId, PublishError> {
-        self.publish_to_topichash(topic.into(), data)
-    }
-
-    /// Publishes a message to a topic in the network using TopicHash.
-    pub fn publish_to_topichash(
-        &mut self,
-        topic: TopicHash,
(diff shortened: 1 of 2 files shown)


protocols/gossipsub/src/behaviour.rs
```diff
@@ -586,20 +586,29 @@ where
         Ok(true)
     }
 
-    /// Publishes a message with multiple topics to the network.
+    /// Publishes a message to a topic in the network.
     pub fn publish<H: Hasher>(
         &mut self,
         topic: Topic<H>,
```
thomaseizinger:
```suggestion
        topic: impl Into<TopicHash>,
```

Would it make sense to generalise the API to this instead of duplicating the function?
This would allow you to pass either a topic or a topic hash.

thomaseizinger (commented):
Thanks for contributing!

I think this is okay but I'd like to also have @mxinden 's opinion on this.

Can you also make a changelog entry please? (Depending on which API we settle on of course.)

protocols/gossipsub/src/behaviour.rs
thomaseizinger: cc @AgeManning What do you think of this API change?

GamePad64: I like this approach, as it extends the existing API without overcomplicating it. Changed this PR to match this comment and added a changelog item.

thomaseizinger (approved):
Looks good from my end!

Going to wait a little longer to see if we can get a 2nd review in :)

AgeManning (approved):
Yeah this looks good to me also. Thanks for the contribution

mxinden (approved):
Looks good to me. Thanks for the patch.

Issue #2854, opened 2022-08-30
swarm/handler: Rename `InEvent` and `OutEvent` to `FromBehaviour` and `ToBehaviour`


thomaseizinger:
## Description

Rename the associated types of `ConnectionHandler` from `InEvent` to `FromBehaviour` and `OutEvent` to `ToBehaviour`.

## Motivation

Using a `ConnectionHandler` outside of the `NetworkBehaviour` abstraction is generally possible put very unlikely. To make it easier for users to understand, what these types do, we can rename them to not just be an "in" and "out" but to include where the event is going to be handled or where it is coming from.

## Are you planning to do it yourself in a pull request?

Maybe.

mxinden:
I find this more intuitive looking at `ConnectionHandler` only. Though in the greater picture, including `NetworkBehaviour`, it is inconsistent with `NetworkBehaviour::OutEvent`.

thomaseizinger:
> I find this more intuitive looking at `ConnectionHandler` only. Though in the greater picture, including `NetworkBehaviour`, it is inconsistent with `NetworkBehaviour::OutEvent`.

We can name that one `ToSwarm`? :)

elenaf9:
+1 in favor of using more concrete names for associated types. 

Maybe we could instead of `FromBehaviour` name `InEvent` -> `ToHandler`. If we then also follow @thomaseizinger proposal to rename `NetworkBehaviour::OutEvent` to `ToSwarm` we'd have consistend names `ToHandler`, `ToBehaviour`, `ToSwarm`, etc. 
Similar naming is also used in the relay v2 protocol for events between handler and listeners.

elenaf9:
Related to this: What do you think about renaming `NetworkBehaviour::inject_event` -> `::inject_handler_msg` and `ConnectionHandler::inject_event` -> `inject_behaviour_msg` (or something similar).

When first working with `NetworkBehaviour` and `Handler` as a user I struggled quite a bit with understanding how behaviour and handler are related and exchange messages. As @thomaseizinger  wrote above, there is no real use-case for a `ConnectionHandler` outside of a `NetworkBehaviour`, so I think we can be a bit less abstract with our names here.

thomaseizinger:
> When first working with `NetworkBehaviour` and `Handler` as a user I struggled quite a bit with understanding how behaviour and handler are related and exchange messages.

[...]

PR #2867, opened 2022-09-04
swarm/behaviour: Replace inject_* with on_event

### LINKED ISSUE #2832: swarm/behaviour: Inject events via single method

mxinden:
# Description

Replace the various `NetworkBehaviour::inject_*` methods with a single
`NetworkBehaviour::on_event` method and a `InEvent` `enum`.

I want to open this for early feedback before I update the many `NetworkBehaviour` implementations in `/protocols`.

Fixes https://github.com/libp2p/rust-libp2p/issues/2832.

## Change checklist


swarm/src/behaviour/either.rs
```diff
@@ -49,170 +47,24 @@ where
...
+                |e| e.unwrap_left(),
+            )),
+            Either::Right(b) => b.on_event(event.map_handler(
+                |h| h.unwrap_right(),
+                |h| match h {
+                    Either::Right(h) => h,
+                    Either::Left(_) => unreachable!(),
+                },
+                |e| e.unwrap_right(),
+            )),
```
mxinden: While not the actual goal of this effort, this is a nice simplification for the `Either` and `Toggle` `NetworkBehaviour`.

In general, I expect this pull request to simplify many of these general purpose `NetworkBehaviour` implementations. (Not that we have that many of them.)

MarcoPolo:
Exciting!

swarm/src/behaviour.rs
```diff
@@ -152,7 +152,15 @@ pub trait NetworkBehaviour: 'static {
         vec![]
     }
 
+    /// Informs the behaviour about an event from the [Transport](libp2p_core::Transport) or the
```
thomaseizinger:
```suggestion
    /// Informs the behaviour about an event from the [`Transport`](libp2p_core::Transport) or the
```

swarm/src/behaviour.rs
```diff
@@ -163,21 +171,31 @@ pub trait NetworkBehaviour: 'static {
...
-    fn inject_connection_closed(
-        &mut self,
-        _: &PeerId,
-        _: &ConnectionId,
-        _: &ConnectedPoint,
-        _: <Self::ConnectionHandler as IntoConnectionHandler>::Handler,
-        _remaining_established: usize,
-    ) {
-    }
+    // TODO: Have to remove this as it takes ownership of `Handler`.
```
thomaseizinger: You could call `on_event` in the default implementation and construct the corresponding variant of `InEvent`.

swarm/src/behaviour.rs
```diff
@@ -187,61 +205,99 @@ pub trait NetworkBehaviour: 'static {
...
-    /// This can include, for example, an error during the handshake of the encryption layer, or the
-    /// connection unexpectedly closed.
-    fn inject_listen_failure(
-        &mut self,
-        _local_addr: &Multiaddr,
-        _send_back_addr: &Multiaddr,
-        _handler: Self::ConnectionHandler,
-    ) {
-    }
+    // TODO: Can not deprecate as it takes ownership of handler event.
```
thomaseizinger: We can deprecate them + make a default implementation where we call `on_event`. Thus, the moment users migrate, they should receive the event via `on_event`.

swarm/src/behaviour.rs
```diff
@@ -721,3 +777,214 @@ impl Default for CloseConnection {
...
+        peer_id: PeerId,
+        connection_id: ConnectionId,
+        endpoint: &'a ConnectedPoint,
+        // TODO: Would a slice not be better?
+        failed_addresses: Option<&'a Vec<Multiaddr>>,
+        other_established: usize,
+    },
+    /// Informs the behaviour about a closed connection to a peer.
+    ///
+    /// This event is always paired with an earlier call to
```
thomaseizinger: "call" isn't really good terminology here I'd say.

swarm/src/behaviour.rs
```diff
@@ -721,3 +777,214 @@ impl Default for CloseConnection {
...
+    /// [`InEvent::ConnectionEstablished`] with the same peer ID, connection ID
+    /// and endpoint.
```
thomaseizinger:
```suggestion
    /// This event is always paired with an earlier
    /// [`InEvent::ConnectionEstablished`] with the same peer ID, connection ID
    /// and endpoint.
```

swarm/src/handler/either.rs
```diff
@@ -64,6 +64,29 @@ where
     }
 }
 
+// Taken from https://github.com/bluss/either.
+impl<L, R> IntoEitherHandler<L, R> {
```
thomaseizinger, line 68: Do we still need `IntoEitherHandler`?

swarm/src/lib.rs
```diff
@@ -1542,13 +1622,63 @@ impl NetworkBehaviour for DummyBehaviour {
...
+                connection: _,
+                event,
+            } => void::unreachable(event),
+            InEvent::ConnectionEstablished {
+                peer_id: _,
+                connection_id: _,
+                endpoint: _,
+                failed_addresses: _,
+                other_established: _,
+            } => {}
```
thomaseizinger, line 1574: Why match on all of these in here?

swarm/src/behaviour/either.rs
```diff
@@ -49,170 +47,24 @@ where
...
-
-    fn inject_listener_error(&mut self, id: ListenerId, err: &(dyn std::error::Error + 'static)) {
-        match self {
-            Either::Left(a) => a.inject_listener_error(id, err),
-            Either::Right(b) => b.inject_listener_error(id, err),
-        }
-    }
-
-    fn inject_listener_closed(&mut self, id: ListenerId, reason: Result<(), &std::io::Error>) {
+    fn on_event(&mut self, event: super::InEvent<Self::ConnectionHandler>) {
```
thomaseizinger: Instead of `super`, I'd prefer importing these from the crate root, potentially prefixing them with the module they are contained in if we need to differentiate with module-local types.

thomaseizinger (commented):
Did a first pass, looks good overall!

Had an idea how we could deprecate more functions.

swarm/src/behaviour.rs
mxinden: Genius @thomaseizinger! A lot better. Done with https://github.com/libp2p/rust-libp2p/pull/2867/commits/de5c46dbe198c756c27de467196324084dee86b8.

swarm/src/behaviour.rs
mxinden: :+1: wonderful. Done with https://github.com/libp2p/rust-libp2p/pull/2867/commits/de5c46dbe198c756c27de467196324084dee86b8.

swarm/src/handler/either.rs
mxinden: How else would the `IntoConnectionHandler::into_handler` methods of the left and right branch be called?

swarm/src/lib.rs
mxinden: To have a compile time error when any of them change, forcing us to revisit each implementation. I guess it is a bit overkill for this one. Let me know if you prefer this to be changed.

swarm/src/behaviour/either.rs
mxinden: Imported as `behaviour::InEvent`. Let me know if that works for you.

mxinden:
Thanks @thomaseizinger for the helpful review. Mind taking another look?

mxinden:
Once everyone is happy with the changes in `libp2p-swarm` I will start porting the many users of `libp2p-swarm` in `/protocols` to the new `NetworkBehaviour::on_event` methods. I plan to do so in this pull request, unless folks prefer separate pull request(s).

swarm/src/behaviour.rs
```diff
@@ -721,3 +838,214 @@ impl Default for CloseConnection {
...
+    /// Informs the behaviour about an event generated by the handler dedicated to the peer
+    /// identified by `peer_id`. for the behaviour.
+    ///
+    /// The `peer_id` is guaranteed to be in a connected state. In other words,
+    /// [`InEvent::ConnectionEstablished`] has previously been received with this `PeerId`.
+    ConnectionHandler {
+        peer_id: PeerId,
+        connection: ConnectionId,
+        event: <<Handler as IntoConnectionHandler>::Handler as ConnectionHandler>::OutEvent,
+    },
```
elenaf9: Referencing a suggestion from @thomaseizinger  in https://github.com/libp2p/rust-libp2p/issues/2854#issuecomment-1237339147: what do you think about splitting this into two methods `on_swarm_event` and `on_handler_event`? 
Imo this would make it easier to understand the kind of event, its source, and how to handle it. 
One major difference between the two types of events is for example that events from the handler are something that the behaviour should definitely handle and that differ depending on the type of behaviour, whereas swarm events are more "informative" in the sense that the behaviour may handle them, but doesn't have to if they are not relevant.

swarm/src/handler/either.rs
thomaseizinger: What I was getting at is: Can we replace it with `Either`?

swarm/src/lib.rs
thomaseizinger: I think it is a good idea but I'd prefer if we do that in a more prominent place. For example, we could have a dedicated `public_api.rs` in `swarm/tests` that concerns itself with such things. I find it a bit odd that this particular impl should take care of that.

Thoughts?

swarm/src/behaviour.rs
```diff
@@ -721,3 +838,214 @@ impl Default for CloseConnection {
...
     }
 }
+
+pub enum InEvent<'a, Handler: IntoConnectionHandler> {
+    /// Informs the behaviour about a newly established connection to a peer.
+    ConnectionEstablished {
+        peer_id: PeerId,
+        connection_id: ConnectionId,
+        endpoint: &'a ConnectedPoint,
+        // TODO: Would a slice not be better?
```
thomaseizinger: Is there a semantic difference between `None` and `Some([])`?

swarm/src/behaviour.rs
thomaseizinger: I am also in favor of my suggestion :relaxed:

swarm/src/behaviour.rs
```diff
@@ -721,3 +838,214 @@ impl Default for CloseConnection {
...
+    {
+        self.try_map_handler(
+            |h| Some(map_into_handler(h)),
+            |h| Some(map_handler(h)),
+            |e| Some(map_event(e)),
+        )
+        .expect("To return Some as all closures return Some.")
+    }
+
+    fn try_map_handler<NewHandler>(
```
thomaseizinger: Nit: IMO `try_` implies failure, meaning this should return a `Result` instead.

The standard library also follows this: https://doc.rust-lang.org/std/?search=try_

thomaseizinger: If we want it to return an `Option`, perhaps `maybe` might be a better prefix?

thomaseizinger:
> Once everyone is happy with the changes in `libp2p-swarm` I will start porting the many users of `libp2p-swarm` in `/protocols` to the new `NetworkBehaviour::on_event` methods. I plan to do so in this pull request, unless folks prefer separate pull request(s).

We would have to litter the rest of the codebase with `allow(deprecated)` attributes to make CI pass. I think porting the protocols in this PR is better.

swarm/src/handler/either.rs
mxinden: I think we discussed this in the past, see https://github.com/libp2p/rust-libp2p/pull/2192#issuecomment-900213835 @thomaseizinger.

swarm/src/behaviour.rs
mxinden: Done via https://github.com/libp2p/rust-libp2p/pull/2867/commits/07a4f201b78978cb56a6f24fd1426de5664447cc.

swarm/src/behaviour.rs
mxinden: Good point. I don't think there is. 99a4b64f removes the `Option` and moves to a `&[]`.

swarm/src/behaviour.rs
mxinden: :+1: done via 46f20fab.

mxinden:
https://github.com/libp2p/rust-libp2p/pull/2867/commits/63b70fe60d723ce75510dc0cfea81302b92a7f58 provides default implementations for `NetworkBehaviour::on_swarm_event` and `NetworkBehaviour::on_connection_handler_event`, thus this patch set can be released as a non-breaking change.

Once we remove the deprecated `NetworkBehaviour::inject_*` methods, I think we should remove the default implementations on  `NetworkBehaviour::on_swarm_event` and `NetworkBehaviour::on_connection_handler_event` as well.

swarm/src/handler/either.rs
thomaseizinger: You've got much better memory than I do :D

thomaseizinger:
> [63b70fe](https://github.com/libp2p/rust-libp2p/commit/63b70fe60d723ce75510dc0cfea81302b92a7f58) provides default implementations for `NetworkBehaviour::on_swarm_event` and `NetworkBehaviour::on_connection_handler_event`, thus this patch set can be released as a non-breaking change.
> 
> Once we remove the deprecated `NetworkBehaviour::inject_*` methods, I think we should remove the default implementations on `NetworkBehaviour::on_swarm_event` and `NetworkBehaviour::on_connection_handler_event` as well.

Agreed!

swarm/src/behaviour.rs
```diff
@@ -152,97 +152,226 @@ pub trait NetworkBehaviour: 'static {
...
+    /// peer identified by `peer_id`. for the behaviour.
+    ///
+    /// The [`PeerId`] is guaranteed to be in a connected state. In other words,
+    /// [`InEvent::ConnectionEstablished`] has previously been received with this [`PeerId`].
+    fn on_connection_handler_event(
+        &mut self,
+        _peer_id: PeerId,
+        _connection_id: ConnectionId,
+        // TODO: Instead of the type alias THandlerOutEvent the full definition might be more
+        // intuitive for users outside of libp2p-swarm.
```
thomaseizinger, line 168: Yep, esp. because the type alias is not public!

swarm/src/behaviour.rs
```diff
@@ -152,97 +152,226 @@ pub trait NetworkBehaviour: 'static {
...
-        _: <Self::ConnectionHandler as IntoConnectionHandler>::Handler,
-        _remaining_established: usize,
+        peer_id: &PeerId,
+        connection_id: &ConnectionId,
+        endpoint: &ConnectedPoint,
+        handler: <Self::ConnectionHandler as IntoConnectionHandler>::Handler,
+        remaining_established: usize,
     ) {
+        self.on_swarm_event(InEvent::ConnectionClosed {
+            peer_id: *peer_id,
```
thomaseizinger, line 214: Nice that we can fix the references here!

swarm/src/behaviour.rs
```diff
@@ -721,3 +850,182 @@ impl Default for CloseConnection {
         CloseConnection::All
     }
 }
+
+// TODO: Rename to FromSwarm
```
thomaseizinger, line 854: Should we pool the rename together with the release of this change as well?

thomaseizinger: We could also name it `FromTransport` maybe? I am beginning to more and more take on the notion of `libp2p-swarm` being a kind of "runtime" that actually doesn't do anything by itself but passes data between other components so in a way, the swarm itself would never produce an event.

swarm/src/handler/either.rs
thomaseizinger: What do you think about adding a comment about it?

thomaseizinger (commented):
It is nice that we can release this as a non-breaking change.

Couple of questions:

- https://github.com/libp2p/rust-libp2p/pull/2867/files#r965651079 is not yet resolved
- Are we converting all behaviours in this repository to the new APIs?

swarm/src/behaviour.rs
elenaf9: In favor of naming it `FromSwarm`. `FromTransport` imo could be confusing because its not an event emitted by a transport **implementation** (e.g. in the relay protocol we do have events that are actually produced by the relay client transport). Also, some events are still caused by the swarm and not the transport, e.g. a `DialFailure` because of swarm connection limits.
That being said, I do generally like the conception of swarm being some kind of "runtime".

mxinden:
Closing here in favor of https://github.com/libp2p/rust-libp2p/pull/3011.

Issue #2865, opened 2022-09-02
Should `StreamMuxerBox` track how many "active" substreams are still around?


thomaseizinger:
## Description

Whilst thinking about https://github.com/libp2p/rust-libp2p/issues/2863, I noticed that it is very easy to (accidentally) circumvent the limit on inbound substreams by "leaking" them to the `ConnectionHandler`. We do this in various places within `rust-libp2p` and I am sure it happens in within code-bases of ours users too.

`StreamMuxerBox` already boxes up each substream as a `SubstreamBox`. We could extend `SubstreamBox` with a `Weak<()>` where `StreamMuxerBox` owns the corresponding `Arc<>`. This would make `Arc::weak_count` effectively be the count of all active (i.e. not dropped) substreams the muxer has given out.

Within `Connection`, we know about `StreamMuxerBox` so we could expose functions on it to give us the current count of active inbound and outbound streams, which would allow us to actually enforce a limit on those.

## Motivation

Not limiting the number of inbound streams can cause unexpected memory and CPU growth and in the worst case, is exploitable via a DoS attack.

## Downsides

Rolling out this change may negatively affect users because they will suddenly experience a limit where there wasn't one before.

## Are you planning to do it yourself in a pull request?

Yes.

thomaseizinger:
Tagging as "help wanted" because I'd like to get people's input on this idea.

mxinden:
I like the idea. Though instead of using the `Arc::weak_count` to enforce a limit, I suggest using it to build a Prometheus metric. I would imagine a `Connection` emits an event whenever a new stream is created. That event would as well contain the overall count of streams on that connection.

I am against using `Arc::weak_count` as a limit as I find it unintuitive. We pass ownership of the stream to the `ConnectionHandler` implementation, thus I think it is the `ConnectionHandler` that has to manage the resource impact.

thomaseizinger:
> I like the idea. Though instead of using the `Arc::weak_count` to enforce a limit, I suggest using it to build a Prometheus metric. I would imagine a `Connection` emits an event whenever a new stream is created. That event would as well contain the overall count of streams on that connection.
> 
> I am against using `Arc::weak_count` as a limit as I find it unintuitive. We pass ownership of the stream to the `ConnectionHandler` implementation, thus I think it is the `ConnectionHandler` that has to manage the resource impact.

[...]

Issue #2863, opened 2022-09-01
Simplify `ConnectionHandler` trait by removing as many associated types as possible


thomaseizinger:
Simplify the `ConnectionHandler` interface. This is one of the first interfaces users get in touch with when the implement their own protocols. Making it as simple as possible without removing functionality should be the goal. All but `{To,From}Behaviour` should be removed and we should directly hand the negotiated stream to the `ConnectionHandler` without any form of "upgrade".

```[tasklist]
### Sub-epics
- [ ] https://github.com/libp2p/rust-libp2p/issues/4306
- [ ] https://github.com/libp2p/rust-libp2p/issues/3268
- [ ] https://github.com/libp2p/rust-libp2p/discussions/4585
- [ ] https://github.com/libp2p/rust-libp2p/issues/3591
```
```[tasklist]
### Use `ReadyUpgrade` everywhere
- [ ] https://github.com/libp2p/rust-libp2p/issues/4075
- [ ] https://github.com/libp2p/rust-libp2p/pull/4563
- [ ] https://github.com/libp2p/rust-libp2p/pull/3914
- [ ] https://github.com/libp2p/rust-libp2p/issues/4649
- [ ] refactor(dcutr): use `ReadyUpgrade` for `{In,Out}boundUpgrade`
- [ ] refactor(gossipsub): use `ReadyUpgrade` for `{In,Out}boundUpgrade`
```
```[tasklist]
### Migration plan
- [ ] https://github.com/libp2p/rust-libp2p/issues/4307
- [ ] https://github.com/libp2p/rust-libp2p/issues/4521
- [ ] https://github.com/libp2p/rust-libp2p/issues/4697
- [ ] https://github.com/libp2p/rust-libp2p/issues/4790
```

thomaseizinger:
I did a bit more exploration and we will need to change a few more things to make this possible.

Concretely, trying to apply `ReadyUpgrade` (https://github.com/libp2p/rust-libp2p/pull/2855) in as many places as possible will help us with this change. Protocols like identify currently use the upgrade mechanism to directly read a message from the stream and close it afterwards. This is convenient because it means the handler is directly given the event / response without having to manually poll an `OptionFuture` or `FuturesUnordered`.

That being said, many protocols do much more complicated things within those upgrade traits (looking at you kademlia and relay) and I think it would overall be more consistent if we just move all of that into the `ConnectionHandler`.

[...]

PR #2878, opened 2022-09-08
core/muxing: Use total number of alive inbound streams for backpressure

thomaseizinger:
# Description

A PoC implementation for what is described in https://github.com/libp2p/rust-libp2p/issues/2865.

- https://github.com/libp2p/rust-libp2p/issues/2863

## Open tasks

- [x] Merge https://github.com/libp2p/rust-libp2p/pull/2861
- [x] Remove `SubstreamBox::new` from public API of `libp2p-core`
- [x] Bump version of `libp2p-core` to 0.37.0
- [x] Bump version of `libp2p-swarm` to 0.40.0

## Open Questions

- ~Should `libp2p-kad` make use of the new limit and https://github.com/libp2p/rust-libp2p/blob/cce296e55eab361fffb7c5304eb504e45cb3ba59/protocols/kad/src/handler.rs#L42 be refactored?~
	- I ended up using it.

## Change checklist


mxinden (commented):
Thanks for writing this POC!

From what I can see here, I am in favor of this proposal.

I would like to only merge here once we have at least one user of this API. I would guess that the best user for this would be `libp2p-metrics`.

While we could emit a new `SwarmEvent` on each new stream, I wonder how we could inform `libp2p-metrics` once a stream closed.

Next to the number of substreams, it would be wonderful to expose the protocol that is being negotiated on the substream as well. That would enable us to expose the following metric:

```
libp2p_swarm_stream { direction = "inbound", protocol = "/libp2p/kad/1.0.0" } 10
```

thomaseizinger:
> Thanks for writing this POC!
> 
> From what I can see here, I am in favor of this proposal.
> 
> I would like to only merge here once we have at least one user of this API. I would guess that the best user for this would be `libp2p-metrics`.

Yes, makes sense. I see the first user in `swarm::Connection` but for that #2861 needs to merge first. Together with #2861, this will give us actual backpressure on the number of open streams and not just the ones that are currently negotiating.

> While we could emit a new `SwarmEvent` on each new stream, I wonder how we could inform `libp2p-metrics` once a stream closed.
> 
> Next to the number of substreams, it would be wonderful to expose the protocol that is being negotiated on the substream as well. That would enable us to expose the following metric:
> 
> ```
> libp2p_swarm_stream { direction = "inbound", protocol = "/libp2p/kad/1.0.0" } 10
> ```

I did something like this [in the past](https://github.com/itchysats/itchysats/blob/master/xtra-libp2p/src/substream.rs). If we make `swarm::Connection` aware of metrics, it is fairly easily doable.

swarm/CHANGELOG.md
```diff
@@ -1,3 +1,16 @@
+# 0.40.0 [unreleased]
+
+- Use the total number of alive inbound streams for back-pressure. This can have a BIG impact on your application
+  depending on how it uses `libp2p`. Previously, the limit for inbound streams per connection only applied to the
+  _upgrade_ phase, i.e. for the time `InboundUpgrade` was running. Any stream being returned from `InboundUpgrade` and
+  given to the `ConnectionHandler` did not count towards that limit, essentially mitigating the back-pressure mechanism.
+  With this release, substreams count towards that limit until they are dropped and thus we actually enforce, how many
+  inbound streams can be active at one time _per connection_. `libp2p` will not accept any more incoming streams once
+  that limit is hit. If you experience stalls or unaccepted streams in your application, consider upping the limit via
+  `SwarmBuilder::max_negotiating_inbound_streams`. See [PR 2878].
```
thomaseizinger: Not sure if I am giving this too much attention but it feels like a really important change to me.

mxinden: I have to give this more thought. Previously I was under the impression that this will be  feature for reporting only, but not actually enforcing a limit.

As of today I am undecided whether the number of established inbound streams should have a global connection maximum, or whether it should only have a maximum per `ConnectionHandler` implementation, potentially coordinated with the `NetworkBehaviour` and thus a maximum per `NetworkBehaviour` implementation. I am not sure the global application has enough knowledge to set a suitable limit for all protocols on a single connection, nor do I think a static limit per connection is a good idea in the first place.

Ideally I would like to enforce a limit for connections only. With https://github.com/libp2p/rust-libp2p/pull/2828 a connection limit could e.g. be dynamic based on the current memory utilization. Protocols, e.g. like Kademlia, would enforce a streams-per-connection limit in their `ConnectionHandler` implementation. That limit would be a value of maximum expected parallelization, e.g. 16 (as in "we don't expect an implementation to be handle more than 16 requests in parallel").

What do you think @thomaseizinger?

thomaseizinger: If we want the limit to be dynamic, all we need to do is add a function to `ConnectionHandler` that allows us to query the max number of allowed streams:

```rust
trait ConnectionHandler {
	fn max_inbound_streams(&self) -> usize {
		128 // This is today's default.
	}
}
```

We can then use this limit on every iteration of `Connection::poll` to check if we should poll for more inbound streams.

If we move forward with https://github.com/libp2p/rust-libp2p/issues/2863 then the notion of upgrades will be gone and this limit does not mean anything anymore. Thus, I am inclined to say we should remove this from the `Swarm` and `Pool` altogether and only have the `ConnectionHandler` decide.

mxinden: > If we move forward with #2863 then the notion of upgrades will be gone and this limit does not mean anything anymore. Thus, I am inclined to say we should remove this from the `Swarm` and `Pool` altogether and only have the `ConnectionHandler` decide.

That would be my preference.

thomaseizinger: LMK what you think of [`40ab076` (#2878)](https://github.com/libp2p/rust-libp2p/pull/2878/commits/40ab076ac88a16ec7e9c647935725b92a764753a).

thomaseizinger:
@mxinden Updated the PR description with an open question.

thomaseizinger:
@mxinden In https://github.com/libp2p/rust-libp2p/discussions/2957, a usecase for tracking the number of substreams in metrics came up.

I can't think of a way of recording this though if we don't want to take in a dependency on `prometheus-client`. If we had a metric instance inside `Connection`, I could update it on every loop iteration. We could emit events every time we open or accept a new substream?

thomaseizinger:
@AgeManning I'd be curious what you think of this PR in regards to what it changes in gossipsub ([`534c3f6` (#2878)](https://github.com/libp2p/rust-libp2p/pull/2878/commits/534c3f6ed9d8bc43f6fdc8a9313df811343fe7dd)).

The gist is that we can now enforce, how many substreams to accept from the remote. Given that GossipSub should only ever have one inbound stream, I set this to one.

AgeManning:
Yeah the changes look fine to me.  We only have one inbound substream as you've pointed out, so I dont see any issue here

thomaseizinger:
I can't reproduce [this](https://github.com/libp2p/rust-libp2p/actions/runs/3249249130/jobs/5331434048) error on my machine. Anyone else got any luck?

Also, can you have another look at this @mxinden? I implemented a _per connection_-configurable limit now as requested in https://github.com/libp2p/rust-libp2p/pull/2878#discussion_r983655065 :)

swarm/src/handler.rs
```diff
@@ -115,6 +117,13 @@ pub trait ConnectionHandler: Send + 'static {
     /// >           This allows a remote to put the list of supported protocols in a cache.
     fn listen_protocol(&self) -> SubstreamProtocol<Self::InboundProtocol, Self::InboundOpenInfo>;
 
+    /// The maximum number of inbound substreams allowed on the underlying connection.
+    ///
+    /// Once this limit is hit, we will stop accepting new inbound streams from the remote.
+    fn max_inbound_streams(&self) -> usize {
+        DEFAULT_MAX_INBOUND_STREAMS
+    }
```
mxinden, line 125: In case an implementation of `ConnectionHandler` does not implement this method, but accepts `DEFAULT_MAX_INBOUND_STREAMS` streams, it would starve all other implementations of `ConnectionHandler` from receiving more inbound streams, correct?

swarm/src/handler/multi.rs
```diff
@@ -121,6 +121,14 @@ where
         SubstreamProtocol::new(upgrade, info).with_timeout(timeout)
     }
 
+    fn max_inbound_streams(&self) -> usize {
+        self.handlers
+            .values()
+            .map(|h| h.max_inbound_streams())
+            .max()
```
mxinden, line 128: Shouldn't this be the sum? I.e. the sum of all limits equals the total limit.

swarm/src/handler/multi.rs
```diff
@@ -121,6 +121,14 @@ where
...
+            .unwrap_or(0) // No handlers? No substreams.
```
mxinden, line 129: :D

mxinden (commented):
I think long term this is something we should be doing. I am not yet sure on whether the current implementation allows collaborative usage of these limits across `ConnectionHandler` implementations.

In my eyes, before we do this, we should tackle https://github.com/libp2p/rust-libp2p/issues/2863 first, thus simplifying the sets (pending-multistream, pending-upgrade, negotiated) of inbound streams.

Another thought would be to redesign `max_inbound_streams` to be closer to `Sink::poll_ready`:

```
fn poll_new_inbound_stream_ready(&self, cx) -> Poll<()> {
```

A `ConnectionHandler` could thus signal in a async way whether it is willing to accept new streams.

Also adding this to the agenda for the next rust-libp2p community call.

swarm/src/handler.rs
thomaseizinger: From the connection task PoV, there is only one `ConnectionHandler`. I think if we get the combinators right, this shouldn't be an issue?

thomaseizinger:
> I think long term this is something we should be doing. I am not yet sure on whether the current implementation allows collaborative usage of these limits across `ConnectionHandler` implementations.

Do you think the proposed implementation is an improvement?

I think it is but I agree with you that it can be even better.

> In my eyes, before we do this, we should tackle https://github.com/libp2p/rust-libp2p/issues/2863 first, thus simplifying the sets (pending-multistream, pending-upgrade, negotiated) of inbound streams.

Funny, I see this a step towards that solution. With this PR, a substream counts towards a limit as soon as it exists and not just during the upgrade phase.

The next idea (but could be done in parallel) for #2863 is to make less use of upgrades in other protocols but build a `FuturesUnordered`-like primitive that implementations can use to do the upgrade themselves.

> Another thought would be to redesign `max_inbound_streams` to be closer to `Sink::poll_ready`:
> 
> ```
> fn poll_new_inbound_stream_ready(&self, cx) -> Poll<()> {
> ```
> 
> A `ConnectionHandler` could thus signal in a async way whether it is willing to accept new streams.

This is an interesting idea! Should a `ConnectionHandler` have multiple `poll` functions perhaps?

- `poll_ready_inbound`
- `poll_new_outbound`
- `poll_event`

> Also adding this to the agenda for the next rust-libp2p community call.

👍

thomaseizinger:
Setting to draft until we reach a decision.

thomaseizinger:
Closing because stale.

Issue #2884, opened 2022-09-12
Introduce `libp2p-swarm-test` crate


thomaseizinger:
## Description

Introduce a crate that contains reusable bits that make it easy to correctly test `NetworkBehaviour` implemenations. In particular, the crate should support the setting up and connecting a number of swarms, driving them correctly and asserting the emitted events.

## Motivation

For code to be high quality it should be tested. If testing is easier, users are more likely to do it, thus yielding higher quality code.

## Are you planning to do it yourself in a pull request?

Yes.

PR #2886, opened 2022-09-12
*: Unfiy how we depend on crates across the workspace

### LINKED ISSUE #2753: Unify how crates within the workspace depend on each other
1. Should all workspace dependencies express a version number? We currently sometimes include and sometimes we don't.
2. We mix two styles of depending on workspace crates:
a. Depending on the `libp2p` meta crate and activating certain features
b. Depending on the crate directly
c. In some cases, we even do [both](https://github.com/libp2p/rust-libp2p/blob/d4f8ec2d48bca3274296d92a1c74ccbfc9ce4276/swarm/Cargo.toml#L30-L33).

Tasks of this issue:

1. Figure out, how this should ideally be done. Take into account release process work, how easy it is detect breaking changes etc
2. Document it in this issue.
3. Implement the solution consistently across the workspace.

cc @mxinden I know you have a release script. Would be good to know which assumptions that one makes f.e.

Folks who have experience in maintaining big workspaces are also welcome to comment.

thomaseizinger:
# Description

Whilst working on #2884, I came across our not so consistent dev-dependencies again which this PR is fixing!

1. For regular dependencies, depend on the concrete subcrate.
2. For dev-dependencies, depend on the root `libp2p` crate.

Fixes #2753.

## Open Questions

I used a local script to check that each crate compiles on its own with the provided feature set:

```sh
cargo metadata --format-version=1 --no-deps | jq -c '.packages | map(.name) | .[]' | xargs -L 1 cargo check --all-targets --package
```

This could fairly easily be made into a dynamic GitHub workflow where we run tests or just cargo check on each crate individually instead of on the entire workspace.
The failure case I can see with the current setup is that - because cargo unifies features across the dependency graph - a missing feature in the dependency declaration for one crate may not be detected by our current CI. This however only affects tests so I am not sure if it is worth it.
Another bonus point could be improved CI times by running the tests / checks for all crates in parallel.

- [ ] See if we can enforce this convention with a script in CI

## Change checklist

~- [ ] A changelog entry has been made in the appropriate crates~


mxinden (approved):
:pray: Thank you for cleaning this up! Much appreciated.

mxinden:
> The failure case I can see with the current setup is that - because cargo unifies features across the dependency graph - a missing feature in the dependency declaration for one crate may not be detected by our current CI. This however only affects tests so I am not sure if it is worth it.

I am fine with that, though given that the failures are only on tests, I am not sure we should prioritize it, nor am I sure it is worth the added complexity.

> Another bonus point could be improved CI times by running the tests / checks for all crates in parallel.

Worth exploring. I am guessing that the GitHub CI runners don't come with many CPU cores and thus I would guess that this does not have a big impact. Also I expect the Testground tests to be our longest CI run from now on, thus I am not sure how much this optimization would actually help us.

mxinden:
What do you think of adding a short paragraph describing this convention to https://github.com/libp2p/rust-libp2p/blob/master/docs/coding-guidelines.md?

I am sure future Max would be very grateful.

thomaseizinger:
> > Another bonus point could be improved CI times by running the tests / checks for all crates in parallel.
> 
> Worth exploring. I am guessing that the GitHub CI runners don't come with many CPU cores and thus I would guess that this does not have a big impact.

I think the opposite is the case. Each job run will get its own container / resources AFAIK, meaning while we are limited to cargo's parallelization now, running them in parallel through multiple jobs should get us quite the speed-up. Not to mention that tests in different crates are run sequentially by cargo whereas with multiple jobs, they'd run in parallel.

> Also I expect the Testground tests to be our longest CI run from now on, thus I am not sure how much this optimization would actually help us.

Probably. I still wouldn't mind if the other jobs are faster :D

thomaseizinger:
> What do you think of adding a short paragraph describing this convention to [https://github.com/libp2p/rust-libp2p/blob/master/docs/coding-guidelines.md](https://github.com/libp2p/rust-libp2p/blob/master/docs/coding-guidelines.md?rgh-link-date=2022-09-15T14%3A24%3A05Z)?
> 
> I am sure future Max would be very grateful.

Can do but I think this will be dead documentation eventually. I'll look into how this can be enforced with a script perhaps, so we just have it as part of CI.

mxinden:
> > > Another bonus point could be improved CI times by running the tests / checks for all crates in parallel.
> > 
> > 
> > Worth exploring. I am guessing that the GitHub CI runners don't come with many CPU cores and thus I would guess that this does not have a big impact.
> 
> I think the opposite is the case. Each job run will get its own container / resources AFAIK, meaning while we are limited to cargo's parallelization now, running them in parallel through multiple jobs should get us quite the speed-up. Not to mention that tests in different crates are run sequentially by cargo whereas with multiple jobs, they'd run in parallel.

Ah, yes, you are absolutely right. Thanks for clarifying.

mxinden:
> > What do you think of adding a short paragraph describing this convention to [https://github.com/libp2p/rust-libp2p/blob/master/docs/coding-guidelines.md](https://github.com/libp2p/rust-libp2p/blob/master/docs/coding-guidelines.md?rgh-link-date=2022-09-15T14%3A24%3A05Z)?
> > I am sure future Max would be very grateful.
> 
> Can do but I think this will be dead documentation eventually. I'll look into how this can be enforced with a script perhaps, so we just have it as part of CI.

Even better!

mxinden:
>  Continuous integration / Build and test (--no-default-features) (pull_request) Failing after 9m 

Once this one is green, I am happy for this to be merged.

thomaseizinger:
> > Continuous integration / Build and test (--no-default-features) (pull_request) Failing after 9m
> 
> Once this one is green, I am happy for this to be merged.

Sweet! Having it compile would help I guess :)

thomaseizinger:
> > > What do you think of adding a short paragraph describing this convention to [https://github.com/libp2p/rust-libp2p/blob/master/docs/coding-guidelines.md](https://github.com/libp2p/rust-libp2p/blob/master/docs/coding-guidelines.md?rgh-link-date=2022-09-16T09%3A09%3A03Z)?
> > > I am sure future Max would be very grateful.
> > 
> > 
> > Can do but I think this will be dead documentation eventually. I'll look into how this can be enforced with a script perhaps, so we just have it as part of CI.
> 
> Even better!

I'll look into it as a follow-up!

PR #2888, opened 2022-09-12
test: introduce `libp2p-swarm-test`

### LINKED ISSUE #2884: Introduce `libp2p-swarm-test` crate

thomaseizinger:
## Description

This patch-set introduces `libp2p-swarm-test`. It provides utilities for quick and safe bootstrapping of tests for `NetworkBehaviour`s. The main design features are:

- Everything has timeouts
- APIs don't get in your way
- Minimal boilerplate

Closes #2884.

## Notes

I did not port `libp2p-kad` because the testing logic was too complicated. I also did not add tests for protocols that don't already have some. That is out of scope for this PR.

## Open Questions

- ~Should this testing crate also pull together utilities like [`ArbitraryPeerId`](https://github.com/kpp/rust-libp2p/blob/bc77431620f952dd8a9e0054bba73bb1c9638bb7/protocols/kad/src/query/peers/closest/disjoint.rs#L535-L546)?~
	- No. That should go into `quickcheck-ext`.
- ~Can we make the `Swarm::new_ephemeral` API any better? At the moment it only allows to configure the underlying transport but the authentication and muxer are hardcoded. I don't want to introduce any more parameters because it will get too verbose and there wouldn't be a benefit in having the function in the first place. Should we perhaps introduce a dedicated `SwarmBuilder` for our tests that start with some reasonable test defaults, like a `MemoryTransport`?~
- ~Trying to port the new harness to `libp2p-ping`, I noticed that we'd lose the `MuxerChoice` as part of the prop test. If we go forward with the `SwarmBuilder` extension idea, we could make a function there that cycles through various muxers and authentication algorithms to make sure things work across various stacks. I am not sure it is worth it though to be honest.~

## Open tasks

- ~Port to more crates~

## Change checklist

~- [ ] A changelog entry has been made in the appropriate crates~


melekes (commented):
I'm in great support of this ❤️

swarm-test/src/lib.rs
```diff
@@ -0,0 +1,226 @@
...
+    /// Establishes a connection to the given [`Swarm`], polling both of them until the connection is established.
+    async fn block_on_connection<T>(&mut self, other: &mut Swarm<T>)
+    where
+        T: NetworkBehaviour + Send,
+        <T as NetworkBehaviour>::OutEvent: Debug;
+
+    /// Listens on a random memory address, polling the [`Swarm`] until the transport is ready to accept connections.
+    async fn listen_on_random_memory_address(&mut self) -> Multiaddr;
+
+    /// Listens on a random memory address, polling the [`Swarm`] until the transport is ready to accept connections.
```
mxinden:
```suggestion
    /// Listens on a random tcp address, polling the [`Swarm`] until the transport is ready to accept connections.
```

swarm-test/src/lib.rs
```diff
@@ -0,0 +1,226 @@
...
+    async fn listen_on_random_localhost_tcp_port(&mut self) -> Multiaddr;
+
+    async fn next_within(
+        &mut self,
+        seconds: u64,
```
mxinden: Having to provide a timeout seems off to me. How would a user decide on the right value? Can't we just rely on the overall test timeout?

swarm/src/lib.rs
```diff
@@ -249,6 +249,16 @@ pub enum SwarmEvent<TBehaviourOutEvent, THandlerErr> {
     Dialing(PeerId),
 }
 
+impl<TBehaviourOutEvent, THandlerErr> SwarmEvent<TBehaviourOutEvent, THandlerErr> {
+    /// Extract the `TBehaviourOutEvent` from this [`SwarmEvent`] in case it is the `Behaviour` variant, otherwise fail.
+    pub fn try_into_behaviour_event(self) -> Result<TBehaviourOutEvent, Self> {
```
mxinden, line 302:
```suggestion
    pub fn try_into_behaviour_event(self) -> Result<TBehaviourOutEvent, Self> {
```

Neat return signature!

swarm-test/src/lib.rs
```diff
@@ -0,0 +1,226 @@
...
+    where
+        T: Transport + Send + 'static + Unpin,
+        T::Dial: Send + 'static,
+        T::ListenerUpgrade: Send + 'static,
+        T::Error: Send + Sync,
+        T::Output: AsyncRead + AsyncWrite + Send + 'static + Unpin,
+        Self: Sized;
+
+    /// Establishes a connection to the given [`Swarm`], polling both of them until the connection is established.
+    async fn block_on_connection<T>(&mut self, other: &mut Swarm<T>)
```
mxinden:
```suggestion
    async fn connect<T>(&mut self, other: &mut Swarm<T>)
```

- `block_on` reminds me of `futures::executor::block_on` which this doesn't use.
- `block_on_connection` does not imply actually triggering the dial

What do you think of the new name?

mxinden (commented):
Direction looks good to me. Thanks for pushing this forward.

mxinden:
> Should this testing crate also pull together utilities like [`ArbitraryPeerId`](https://github.com/kpp/rust-libp2p/blob/bc77431620f952dd8a9e0054bba73bb1c9638bb7/protocols/kad/src/query/peers/closest/disjoint.rs#L535-L546)?

I am guessing that defining `Arbitrary` on `PeerId` directly under the `cfg[test]` doesn't expose outside of the crate. Is that correct?

mxinden:
> * Can we make the `Swarm::new_ephemeral` API any better? At the moment it only allows to configure the underlying transport but the authentication and muxer are hardcoded. I don't want to introduce any more parameters because it will get too verbose and there wouldn't be a benefit in having the function in the first place. Should we perhaps introduce a dedicated `SwarmBuilder` for our tests that start with some reasonable test defaults, like a `MemoryTransport`?

I think all our non `Transport` tests should use the `MemoryTransport`, thus I am in favor of this default.

mxinden:
> Trying to port the new harness to `libp2p-ping`, I noticed that we'd lose the `MuxerChoice` as part of the prop test. If we go forward with the `SwarmBuilder` extension idea, we could make a function there that cycles through various muxers and authentication algorithms to make sure things work across various stacks. I am not sure it is worth it though to be honest.

I still have to put more thoughts into this, though as an intuition I suggest that a test only tests one thing. E.g. the `libp2p-ping` tests test `libp2p-ping` and not some permutation of different transports and muxers. I think that should happen in a separate dedicated test.

swarm-test/src/lib.rs
thomaseizinger: Tests don't have an overall timeout unless the user wraps it all in one which is boilerplate and thus doesn't happen very often. Plus, compared to a test timeout, having the test failure in here provides a nice stacktrace that is easier to debug on where the test is failing.

The reason we can't have a timeout default is because we don't know about the test case, only the dev writing the test can decide that :)

swarm/src/lib.rs
thomaseizinger: :pray:

swarm-test/src/lib.rs
thomaseizinger: Yeah sounds good and it is shorter too :)

thomaseizinger:
> > Should this testing crate also pull together utilities like [`ArbitraryPeerId`](https://github.com/kpp/rust-libp2p/blob/bc77431620f952dd8a9e0054bba73bb1c9638bb7/protocols/kad/src/query/peers/closest/disjoint.rs#L535-L546)?
> 
> I am guessing that defining `Arbitrary` on `PeerId` directly under the `cfg[test]` doesn't expose outside of the crate. Is that correct?

That is correct.

> > * Can we make the `Swarm::new_ephemeral` API any better? At the moment it only allows to configure the underlying transport but the authentication and muxer are hardcoded. I don't want to introduce any more parameters because it will get too verbose and there wouldn't be a benefit in having the function in the first place. Should we perhaps introduce a dedicated `SwarmBuilder` for our tests that start with some reasonable test defaults, like a `MemoryTransport`?
> 
> I think all our non `Transport` tests should use the `MemoryTransport`, thus I am in favor of this default.

Cool, that makes this simpler :)

> > Trying to port the new harness to `libp2p-ping`, I noticed that we'd lose the `MuxerChoice` as part of the prop test. If we go forward with the `SwarmBuilder` extension idea, we could make a function there that cycles through various muxers and authentication algorithms to make sure things work across various stacks. I am not sure it is worth it though to be honest.
> 
> I still have to put more thoughts into this, though as an intuition I suggest that a test only tests one thing. E.g. the `libp2p-ping` tests test `libp2p-ping` and not some permutation of different transports and muxers. I think that should happen in a separate dedicated test.

Yeah I am tending towards the same. We should have a good muxer test harness instead!

swarm-test/src/lib.rs
mxinden: > Tests don't have an overall timeout unless the user wraps it all in one which is boilerplate and thus doesn't happen very often. Plus, compared to a test timeout, having the test failure in here provides a nice stacktrace that is easier to debug on where the test is failing.

I wasn't aware of this. Thanks.

Still, I would consider concrete timeouts in tests a code smell. Users of this framework would have to make assumptions on the environment this is run on. E.g. they need to take into account that their test might run on a GitHub CI or some high performant developer machine.

I can not think of a way without the timeout while still having a strack trace. Is the strack trace important? I guess this boils down to "should one be able to see the root cause purely from the unit test failure, or do we expect a red test to need debugging anyways".

I don't expect any unit test to take longer than 10 seconds. In case we want to keep the stack traces, how about setting the timeout to a fixed `10s`?

mxinden:
> > > Should this testing crate also pull together utilities like [`ArbitraryPeerId`](https://github.com/kpp/rust-libp2p/blob/bc77431620f952dd8a9e0054bba73bb1c9638bb7/protocols/kad/src/query/peers/closest/disjoint.rs#L535-L546)?
> > 
> > 
> > I am guessing that defining `Arbitrary` on `PeerId` directly under the `cfg[test]` doesn't expose outside of the crate. Is that correct?
> 
> That is correct.

That is unfortunate. In that case, I am in favor of providing an `ArbitraryPeerId` from this new crate.

swarm-test/src/lib.rs
```diff
@@ -0,0 +1,183 @@
...
+    }
+}
+
+fn get_rand_memory_address() -> Multiaddr {
+    let address_port = rand::random::<u64>();
+    let addr = format!("/memory/{}", address_port)
+        .parse::<Multiaddr>()
+        .unwrap();
+
+    addr
```
elenaf9:
```suggestion
    Protocol::Memory(0).into()
```
The memory transport picks a random port if the port is 0.

swarm-test/src/lib.rs
```diff
@@ -0,0 +1,183 @@
...
+#[async_trait]
+pub trait SwarmExt {
+    type NB: NetworkBehaviour;
+
+    /// Create a new [`Swarm`] with an ephemeral identity.
+    ///
+    /// The swarm will use a [`MemoryTransport`] together with a noise authentication layer and
+    /// yamux as the multiplexer. However, these details should not be relied upon by the test
+    /// and may change at any time.
+    fn new_ephemeral(behaviour_fn: impl FnOnce(PeerId, Keypair) -> Self::NB) -> Self
```
elenaf9:
```suggestion
    fn new_ephemeral(behaviour_fn: impl FnOnce(Keypair) -> Self::NB) -> Self
```
Since the `PeerId` can be derived from the `Keypair`, what do you think about removing it as parameter?

swarm/src/lib.rs
elenaf9, line 302: Is there a case where a user would call this method and not unwrap the result? If not, what do you think about panicking here instead of returning an Err?

elenaf9 (approved):
Nice!

swarm-test/src/lib.rs
thomaseizinger: > I don't expect any unit test to take longer than 10 seconds. In case we want to keep the stack traces, how about setting the timeout to a fixed `10s`?

I guess for any situation where the timeout doesn't fit, the user can always fall back to the `StreamExt::next` function which doesn't have a timeout so this is probably a good idea?

What we need to balance with these test abstractions is for them to flexible enough to handle a lot of situations but also providing enough value in making test code concise. I think a timeout 10s is a good default and if it doesn't fit, you can always just code it yourself and it is not too verbose. If it happens a lot, we can look into adjusting the timeout or finding other solutions.

swarm-test/src/lib.rs
thomaseizinger: That is so much clearer, thank you!

swarm-test/src/lib.rs
thomaseizinger: We could actually, yeah. It is a bit verbose though and many times, behaviours only need to know the `PeerId` and not the `Keypair`. Perhaps I can make some clever abstractions here so the user can choose ...

swarm/src/lib.rs
thomaseizinger: Given that this is a function in production code, I don't think panicking is a good option here.

elenaf9: Woops, I overlooked that it is in `swarm` and not `swarm-test`. Agree with you.

thomaseizinger:
> > > > Should this testing crate also pull together utilities like [`ArbitraryPeerId`](https://github.com/kpp/rust-libp2p/blob/bc77431620f952dd8a9e0054bba73bb1c9638bb7/protocols/kad/src/query/peers/closest/disjoint.rs#L535-L546)?
> > > 
> > > 
> > > I am guessing that defining `Arbitrary` on `PeerId` directly under the `cfg[test]` doesn't expose outside of the crate. Is that correct?
> > 
> > 
> > That is correct.
> 
> That is unfortunate. In that case, I am in favor of providing an `ArbitraryPeerId` from this new crate.

Now that we have `quickcheck-ext`, should it perhaps go in there?

mxinden:
@thomaseizinger let me know once I should be taking another look here.

swarm-test/src/lib.rs
thomaseizinger: Addressed in [`c6b670d` (#2888)](https://github.com/libp2p/rust-libp2p/pull/2888/commits/c6b670d19b4a4c27cab660bc12688aa6578dd69c).

thomaseizinger:
@libp2p/rust-libp2p-maintainers Please have another look to make sure this goes into the right direction! :)

mxinden:
> @libp2p/rust-libp2p-maintainers Please have another look to make sure this goes into the right direction! :)

Right direction from my side.

mxinden:
Still interested to get this merged @thomaseizinger! Writing a tiny protocol right now where this would come in handy for testing.

thomaseizinger:
> Still interested to get this merged @thomaseizinger! Writing a tiny protocol right now where this would come in handy for testing.

Ah yes, me too. I'd prefer if we port the entire workspace over to this though which means there is still quite a bit of work to be done before we can merge it.

thomaseizinger:
I'll see to spend some cycles on it soon.

swarm-test/src/lib.rs
```diff
@@ -0,0 +1,352 @@
...
+///
+/// # Difference to [`futures::future::join`]
+///
+/// This function is similar to joining two futures with two crucial differences:
+/// 1. As described above, it allows you to obtain more than a single event.
+/// 2. More importantly, it will continue to poll the [`Swarm`]s **even if they already has emitted all expected events**.
+///
+/// Especially (2) is crucial for our usage of this function.
+/// If a [`Swarm`] is not polled, nothing within it makes progress.
+/// This can "starve" the other swarm which for example may wait for another message to be sent on a connection.
```
thomaseizinger, line 120: I spent a good two hours of debugging timeouts in the test and unwoken tasks until it dawned on me why things were not working. @elenaf9 I believe this might also be the reason for your timeouts in #3345.

This function may seem like a beast but after trying various solutions locally, I am convinced it makes sense to add this to the test-harness and use this instead of `future::join` for testing `Swarm`s.

swarm-test/src/lib.rs
```diff
@@ -0,0 +1,353 @@
+use async_trait::async_trait;
```
mxinden, line 21: Please add a license header here :)

swarm-test/src/lib.rs
```diff
@@ -0,0 +1,353 @@
...
+/// This allows the compiler to infer how many events you are expecting based on how you use this function.
+/// For example, if you expect the first [`Swarm`] to emit 2 events, you should assign the first variable of the returned tuple value to an array of size 2.
+/// This works especially well if you directly pattern-match on the return value.
+///
+/// ## Type of event
+///
+/// This function utilizes the [`TryIntoOutput`] trait.
+/// Similar as to the number of expected events, the type of event is inferred based on your usage.
+/// If you match against a [`SwarmEvent`], the first [`SwarmEvent`] will be returned.
+/// If you match against your [`NetworkBehaviour::OutEvent`] type, [`SwarmEvent`]s which are not [`SwarmEvent::Behaviour`] will be skipped until the [`Swarm`] returns a behaviour event.
```
mxinden, line 108: :exploding_head:

swarm-test/src/lib.rs
```diff
@@ -0,0 +1,353 @@
...
+                    res1.push(o1);
+                }
+            }
+            Either::Right((o2, _)) => {
+                if let Ok(o2) = o2.try_into_output() {
+                    res2.push(o2);
+                }
+            }
+        }
+    }
```
mxinden, line 160: Long term it would be nice to detect missed wake-ups in a `NetworkBehaviour` implementation with `libp2p-swarm-test`. If I am not mistaken, that is not possible with the above `while` `select!` loop as long as the other swarm triggers a wake-up.

Lets defer to a future pull request.

swarm-test/src/lib.rs
```diff
@@ -0,0 +1,353 @@
...
+use std::time::Duration;
+
+/// An extension trait for [`Swarm`] that makes it easier to set up a network of [`Swarm`]s for tests.
+#[async_trait]
+pub trait SwarmExt {
+    type NB: NetworkBehaviour;
+
+    /// Create a new [`Swarm`] with an ephemeral identity.
+    ///
+    /// The swarm will use a [`MemoryTransport`] together with a noise authentication layer and
```
mxinden:
```suggestion
    /// The swarm will use a [`MemoryTransport`] together with [`Plaintext2Config`] authentication layer and
```

If I am not mistaken, this framework doesn't actually use noise.

swarm-test/src/lib.rs
```diff
@@ -0,0 +1,353 @@
...
+                        "Ignoring {:?} while waiting for listening to succeed",
+                        other
+                    );
+                    None
+                }
+            })
+            .await;
+
+        // Memory addresses are externally reachable because they all share the same memory-space.
+        self.add_external_address(memory_multiaddr.clone(), AddressScore::Infinite);
```
mxinden, line 314: :+1: good idea

swarm-test/src/lib.rs
```diff
@@ -0,0 +1,353 @@
...
+
+    fn new_ephemeral(behaviour_fn: impl FnOnce(Keypair) -> Self::NB) -> Self
+    where
+        Self: Sized,
+    {
+        let identity = Keypair::generate_ed25519();
+        let peer_id = PeerId::from(identity.public());
+
+        let transport = MemoryTransport::default()
+            .or_transport(libp2p_tcp::async_io::Transport::default())
```
mxinden, line 211: Why do we need the TCP transport? Is this needed for `libp2p-autonat`?

protocols/request-response/tests/ping.rs
```diff
@@ -156,154 +160,127 @@ fn ping_protocol() {
...
+                    continue;
+                }
+            },
+            event = swarm2.select_next_some() => {
+                if let SwarmEvent::Behaviour(ev) = event {
+                    break ev;
+                }
+            },
+        )
+    };
```
mxinden, line 276: Can this not be done via the new shiny `drive` function?

mxinden (approved):
Just looking at the diff, this is already a great step forward.

A couple of comments. Nothing major. Good to merge from my end.

swarm-test/src/lib.rs
thomaseizinger: Thanks! I ended up doing it this way because it allowed to create the `connect` abstraction which can then rely on retrieving the addresses via the external addresses.

swarm-test/src/lib.rs
thomaseizinger: Yes, and `dcutr` needs it too.

protocols/request-response/tests/ping.rs
thomaseizinger: Probably, I've introduced that one only at the very end!

thomaseizinger: I am not sure actually, this one is a bit different because the two peers operate autonomously. `drive` is useful if you want to "step" through a test in a more linear fashion because it gives you access to the `Swarm`s inbetween. Here, we can just create to tasks that operate independently.

swarm-test/src/lib.rs
thomaseizinger: This is kind of a conflicting requirement. I added this because it is very easy to poll a `Swarm` in a wrong way in a test, thus it is often the _test_ that is buggy, not the `NetworkBehaviour` implementation.

Issue #708, opened 2018-11-29
Add an onion transport


tomaka:

shivankgarg98:
Hello @tomaka @gnunicorn ! I'm resolving this issue.

umgefahren:
Is it revolved?

I think it could be implemented quite easily with the [tor_stream](https://crates.io/crates/tor-stream) crate. However this seems rather inelegant, because it relies on a running Tor Proxy in the background. I would wait for a mature version of [Arti](https://gitlab.torproject.org/tpo/core/arti). This would be more cross platform and probably more performant. However doing this now would be quite a hassle, because Arti will probably change it‘s API a lot.

thomaseizinger:
We have experimented with libp2p and Tor a lot and landed on the following:

1. It is quite easy to do a dial-only transport: https://github.com/comit-network/xmr-btc-swap/blob/master/swap/src/network/tor_transport.rs
2. In our experience, Tor users want to manage their Tor daemon and onion addresses themselves. Thus, if I had to integrate a libp2p application with Tor again, I would give them tools to easily setup a hidden services and build a libp2p transport that reads the Tor configuration file and automatically listens on the respective TCP port to receive data.

olanod:
Perhaps relevant https://blog.torproject.org/arti_100_released the Tor project released Arti v1, their new Rust based Tor implementation.

thomaseizinger:
> Perhaps relevant [blog.torproject.org/arti_100_released](https://blog.torproject.org/arti_100_released) the Tor project released Arti v1, their new Rust based Tor implementation.

That is big news indeed. Reading through the blogpost under the "What's coming next?" section, it says that onion services will come with `arti` version `1.2.0`. Listening will require the creation of an onion service so we will have to wait for that to build a full fledged transport.

It should be possible to build one that can only dial a Tor address though and simply fail unconditionally for listening. PRs welcome :)

umgefahren:
I started working on this, however I have issues dealing with Multiaddress correctly. As far as I can tell Multiaddress comes with some support for Onion out of the box, but it's somewhat opaque what this means in terms of Arti's TorAddress.

For now I just extract a SocketAddr (using TCP for the port), since Onion services are not supported anyway.

thomaseizinger:
[...]

PR #2887, opened 2022-09-12
protocols/noise: Introduce `NoiseAuthenticated::xx` constructor with X25519 DH key exchange

thomaseizinger:
# Description

This is the most common way of doing noise in libp2p and thus deserves a convenience ctor. I came across this while working on #2884, trying to reduce the noise (no pun intended) in setting up swarms for testing.

## Change checklist

- [x] I have performed a self-review of my own code
- [x] I have made corresponding changes to the documentation
~- [ ] I have added tests that prove my fix is effective or that my feature works~
- [x] A changelog entry has been made in the appropriate crates

PR #2899, opened 2022-09-13
transports/onion: Add dial-only implementation of `Transport`

umgefahren:
# Description

Adding a libp2p transport built on top of the newly released [Arti](https://gitlab.torproject.org/tpo/core/arti). Currently only dialing is supported, since Onion Services are not supported by Arti yet and it will likely take some time until it's supported, since it's not funded yet.

The implementation should be considered as a first step towards a stable Tor based transport.

This implementation follows the discussion in #708 and should close this issue.

## Open Questions

~~I added libp2p-onion to the default features. However this might be a bad idea, since arti has a lot of dependencies and therefore increases the build time of libp2p greatly.~~ (There are no default features anymore)

Since ~~there is no documentation yet and~~ it's very untested (because of the transports nature automated testing is hard) ~~libp2p-onion maybe shouldn't be integrated in libp2p at all now~~.

## Change checklist


examples/ping-onion.rs
```diff
@@ -0,0 +1,121 @@
...
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS
+// OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+// FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+// DEALINGS IN THE SOFTWARE.
```
umgefahren: I'm not quite sure this copyright notice is correct.

umgefahren (commented):
Documentation is missing, but I will add that, once the code itself is approved.

examples/ping-onion.rs
```diff
@@ -0,0 +1,122 @@
...
+    libp2p_core::transport::Boxed<(PeerId, libp2p_core::muxing::StreamMuxerBox)>,
+    Box<dyn Error>,
+> {
+    use std::time::Duration;
+
+    let transport = {
+        let dns_tcp = dns::DnsConfig::system(tcp::TcpTransport::new(
+            tcp::GenTcpConfig::new().nodelay(true),
+        ))
+        .await?;
```
thomaseizinger: Why do we need this to showcase a dial-only onion transport?

examples/ping-onion.rs
```diff
@@ -0,0 +1,122 @@
...
+    // For illustrative purposes, the ping protocol is configured to
+    // keep the connection alive, so a continuous sequence of pings
+    // can be observed.
+    let behaviour = ping::Behaviour::new(ping::Config::new().with_keep_alive(true));
+
+    let mut swarm = Swarm::new(transport, behaviour, local_peer_id);
+
+    // Tell the swarm to listen on all interfaces and a random, OS-assigned
+    // port.
+    swarm.listen_on("/ip4/0.0.0.0/tcp/0".parse()?)?;
```
thomaseizinger: That seems unnecessary?

transports/onion/src/address.rs
```diff
@@ -0,0 +1,34 @@
...
+    let tcp_port = tcp_port_opt.ok_or(TorAddrError::NoPort)?;
+    Ok(SocketAddr::new(ip, tcp_port))
+}
+
+pub(super) fn dangerous_extract_tor_address(
+    multiaddr: &Multiaddr,
+) -> Result<TorAddr, TorAddrError> {
+    let socket_addr = try_extract_socket_addr(multiaddr)?;
+    TorAddr::dangerously_from(socket_addr)
+}
```
thomaseizinger: From reaching the docs of `dangerously_from`, I think we should only be supporting `/dns` multiaddr because we can't guarantee where the IP comes from and whether it will have leaked the user's IP through a DNS resolution.

transports/onion/tests/mod.rs
```diff
@@ -0,0 +1,88 @@
...
+
+use libp2p::{
+    dns, identity, mplex, noise,
+    ping::{self, Behaviour},
+    swarm::SwarmEvent,
+    tcp, yamux, PeerId, Swarm, Transport,
+};
+use libp2p_onion::OnionClient;
+
+#[cfg(feature = "async-std")]
```
thomaseizinger: Instead of this, you can define the test in `Cargo.toml` explicitly and say it requires certain features. See here for example: https://github.com/Restioson/xtra/blob/b538a5a41130fbfeb3e14eb816e1010f9167d17a/Cargo.toml#L114-L116

transports/onion/tests/mod.rs
```diff
@@ -0,0 +1,88 @@
...
+> {
+    use std::time::Duration;
+
+    use libp2p_core::upgrade::{SelectUpgrade, Version};
+
+    let transport = {
+        let dns_tcp = dns::DnsConfig::system(tcp::TcpTransport::new(
+            tcp::GenTcpConfig::new().nodelay(true),
+        ))
+        .await?;
```
thomaseizinger: Let's not use this transport here.

transports/onion/tests/mod.rs
```diff
@@ -0,0 +1,88 @@
...
+    loop {
+        match swarm.select_next_some().await {
+            SwarmEvent::NewListenAddr { address, .. } => {
+                println!("Listening on {:?}", address);
+                break;
+            }
+            _ => {}
+        }
+    }
+}
```
thomaseizinger: This doesn't test anything about the onion transport itself but just the embedded TCP transport.

transports/onion/src/lib.rs
```diff
@@ -0,0 +1,123 @@
...
+    #[error("onion services are not implented yet, since arti doesn't support it. (awaiting Arti 1.2.0)")]
+    OnionServiceUnimplemented,
+}
+
+#[derive(Clone)]
+pub struct OnionClient {
+    // client is in an Arc, because wihtout it the Transport::Dial method can't be implemented,
+    // due to lifetime issues. With the, eventual, stabilization of static async traits this issue
+    // will be resolved.
+    client: Arc<TorClient<PreferredRuntime>>,
```
thomaseizinger: Although I think the Tor people did a great job with this abstraction, I the way he handled this in libp2p so far is that we expose multiple transports, one for each runtime.

See `libp2p-tcp` as an example.

transports/onion/src/lib.rs
```diff
@@ -0,0 +1,123 @@
...
+#[derive(Debug, Clone, Copy, Default)]
+pub struct AlwaysErrorListenerUpgrade;
+
+impl core::future::Future for AlwaysErrorListenerUpgrade {
+    type Output = Result<OnionStream, OnionError>;
+    fn poll(
+        self: std::pin::Pin<&mut Self>,
+        _cx: &mut std::task::Context<'_>,
+    ) -> std::task::Poll<Self::Output> {
+        core::task::Poll::Ready(Err(OnionError::OnionServiceUnimplemented))
```
thomaseizinger: You could also panic here because it will never be called, that would remove an error variant from the public API.

transports/onion/src/lib.rs
```diff
@@ -0,0 +1,123 @@
...
+
+#[derive(Clone)]
+pub struct OnionClient {
+    // client is in an Arc, because wihtout it the Transport::Dial method can't be implemented,
+    // due to lifetime issues. With the, eventual, stabilization of static async traits this issue
+    // will be resolved.
+    client: Arc<TorClient<PreferredRuntime>>,
+}
+
+pub type OnionBuilder = TorClientBuilder<PreferredRuntime>;
```
thomaseizinger: This is leaking the `arti_client` dependency to our consumers, meaning it affects semver when we upgrade the `arti_client` dependency.

I think that is okay for now.

transports/onion/src/provider.rs
```diff
@@ -0,0 +1,132 @@
+use std::pin::Pin;
+
+use arti_client::DataStream;
+use futures::{AsyncRead, AsyncWrite};
+
+#[derive(Debug)]
+pub struct OnionStream {
+    inner: DataStream,
```
thomaseizinger, line 34: Can we just expose `DataStream` directly? Or does that not work because something further down the stack requires `futures::Async{Read,Write}` even with the `tokio` feature`?

thomaseizinger: I am not happy with these `cfg`s in here. Will need to give this more thought :)

transports/onion/tests/mod.rs
thomaseizinger: I am not sure if we can provide _good_ tests here. In the end, all the heavy lifting is done by arti which we can trust in. All we are doing is a bit of plumbing and that is mostly secured by the type system so I am not sure we need tests for this crate on this level.

Obviously, a manual test might not hurt :)

thomaseizinger (commented):
Did an initial pass, thanks for tackling this!

examples/ping-onion.rs
umgefahren: We actually don't. We could just connect to the normal examples/ping, I just thought it would be easier that way.

examples/ping-onion.rs
umgefahren: I think it's necessary in order to connect to the node.

transports/onion/src/address.rs
umgefahren: I think we should allow connections from IP addresses but also warn in the doc. I will implement the changes.

transports/onion/tests/mod.rs
umgefahren: I will delete the test, since it seems such a bad idea.

transports/onion/src/provider.rs
umgefahren: > Can we just expose `DataStream` directly? Or does that not work because something further down the stack requires `futures::Async{Read,Write}` even with the `tokio` feature`?

I'm actually not quite sure. Although this part is, along with the cfg's, ported from libp2p-tcp.

transports/onion/src/lib.rs
umgefahren: Resolved in latest commit.

transports/onion/src/lib.rs
umgefahren: Resolved in latest commit.

transports/onion/tests/mod.rs
umgefahren: Resolved in latest commit.

transports/onion/tests/mod.rs
umgefahren: Resolved in latest commit.

transports/onion/tests/mod.rs
umgefahren: Resolved in latest commit.

transports/onion/src/address.rs
umgefahren: Resolved in latest commit.

examples/ping-onion.rs
thomaseizinger: Good, because I think it would be a bad showcase as it is too easy to users to copy this example, thinking they are now using Tor and instead leak IPs everywhere. Arguably a very naive way of doing things but I believe we should still prevent it.

examples/ping-onion.rs
thomaseizinger: Listening is only necessary if we want other peers to connect to us.

transports/onion/src/address.rs
```diff
@@ -32,3 +32,49 @@ pub(super) fn dangerous_extract_tor_address(
...
+        if let Some(dns_s) = $dns {
+            match (dns_s.as_ref(), $tcp_port).into_tor_addr() {
+                Ok(tor_addr) => return Ok(tor_addr),
+                Err(e) => $tor_addr_error = Some(e),
+            }
+        }
+    };
+}
+
+pub(super) fn safe_extract_tor_address(multiaddr: &Multiaddr) -> Result<TorAddr, TorAddrError> {
```
thomaseizinger: Can we have some unit tests for this function please?

transports/onion/src/lib.rs
```diff
@@ -22,37 +23,95 @@ pub enum OnionError {
...
 
-pub type OnionBuilder = TorClientBuilder<PreferredRuntime>;
+pub type OnionBuilder<R> = TorClientBuilder<R>;
 
-impl OnionClient {
+impl<R: Runtime> OnionClient<R> {
     #[inline]
-    pub fn builder() -> OnionBuilder {
-        TorClient::builder()
+    pub fn from_builder(builder: OnionBuilder<R>, safe_mode: bool) -> Result<Self, OnionError> {
```
thomaseizinger: Instead of a bool, can we use an enum here? That would be much more self-documenting. I am thinking something like:

```rust
pub enum AddressConversion {
    DnsOnly,
    IpAndDns
}
```

The rustdoc on this enum can then link to Tor's documentation of `DangerouslyIntoTorAddress` etc which will explain the privacy caveats.

transports/onion/src/lib.rs
```diff
@@ -22,37 +23,95 @@ pub enum OnionError {
...
 
+#[cfg(all(feature = "rustls", feature = "tokio"))]
+#[cfg_attr(docsrs, doc(cfg(all(feature = "rustls", feature = "tokio"))))]
+impl OnionClient<tor_rtcompat::tokio::TokioRustlsRuntime> {
+    pub fn builder() -> OnionBuilder<tor_rtcompat::tokio::TokioRustlsRuntime> {
+        let runtime = tor_rtcompat::tokio::TokioRustlsRuntime::current()
+            .expect("Couldn't get the current tokio rustls runtime");
+        TorClient::with_runtime(runtime)
+    }
+}
```
thomaseizinger: Can we have some convenience constructors here that do everything in one go with sensible defaults?

- Bootstrap the client
- Enable safe mode

transports/onion/src/address.rs
```diff
@@ -0,0 +1,80 @@
...
+            }
+            _ => {}
+        }
+    }
+    let ip = ip_4.or(ip_6).ok_or(TorAddrError::InvalidHostname)?;
+    let tcp_port = tcp_port_opt.ok_or(TorAddrError::NoPort)?;
+    Ok(SocketAddr::new(ip, tcp_port))
+}
+
+pub(super) fn dangerous_extract_tor_address(
```
thomaseizinger: `address` is a private module so this can be simplified to `pub fn`,

transports/onion/src/address.rs
```diff
@@ -0,0 +1,80 @@
+use std::net::{IpAddr, SocketAddr};
+
+use arti_client::{IntoTorAddr, TorAddr, TorAddrError};
+use libp2p_core::{multiaddr::Protocol, Multiaddr};
+
+fn try_extract_socket_addr(multiaddr: &Multiaddr) -> Result<SocketAddr, TorAddrError> {
```
thomaseizinger: Can we order the functions in this file according to their visibility / usage please?

Public functions first, followed by utilities in the order they are used as you read code from top to bottom. Thanks!

transports/onion/src/provider.rs
thomaseizinger: > > Can we just expose `DataStream` directly? Or does that not work because something further down the stack requires `futures::Async{Read,Write}` even with the `tokio` feature`?
> 
> I'm actually not quite sure. Although this part is, along with the cfg's, ported from libp2p-tcp.

Oh interesting, I'll need to take another look at what is going on there then!

transports/onion/src/lib.rs
```diff
@@ -0,0 +1,193 @@
...
+    /// Always returns `TransportError::MultiaddrNotSupported`
+    fn listen_on(
+        &mut self,
+        addr: libp2p_core::Multiaddr,
+    ) -> Result<
+        libp2p_core::transport::ListenerId,
+        libp2p_core::transport::TransportError<Self::Error>,
+    > {
+        // although this address might be supported, this is returned in order to not provoke an
+        // error when trying to listen on this transport.
```
thomaseizinger, line 207: This comment is not quite true I think because at the moment, we simply don't support listening on anything because we haven't even defined what listening on an onion transport is meant to do.

transports/onion/src/lib.rs
```diff
@@ -0,0 +1,193 @@
...
+        None
+    }
+
+    /// always returns pending
+    fn poll(
+        self: Pin<&mut Self>,
+        _cx: &mut std::task::Context<'_>,
+    ) -> std::task::Poll<libp2p_core::transport::TransportEvent<Self::ListenerUpgrade, Self::Error>>
+    {
+        // pending is returned here, because this won't panic an OrTransport.
```
thomaseizinger:
```suggestion
        // pending is returned here because this transport doesn't support listening
```

transports/onion/src/lib.rs
```diff
@@ -0,0 +1,193 @@
...
+    ) -> Result<
+        libp2p_core::transport::ListenerId,
+        libp2p_core::transport::TransportError<Self::Error>,
+    > {
+        // although this address might be supported, this is returned in order to not provoke an
+        // error when trying to listen on this transport.
+        Err(TransportError::MultiaddrNotSupported(addr))
+    }
+
+    /// Always returns false
```
thomaseizinger:
```suggestion
```

Don't think we need that :)

transports/onion/src/lib.rs
```diff
@@ -0,0 +1,193 @@
...
+        self.dial(addr)
+    }
+
+    /// always returns `None`
+    fn address_translation(&self, _listen: &Multiaddr, _observed: &Multiaddr) -> Option<Multiaddr> {
+        None
+    }
+
+    /// always returns pending
+    fn poll(
```
thomaseizinger: Same here.

```suggestion
    fn dial_as_listener(
        &mut self,
        addr: Multiaddr,
    ) -> Result<Self::Dial, TransportError<Self::Error>> {
        self.dial(addr)
    }

    fn address_translation(&self, _listen: &Multiaddr, _observed: &Multiaddr) -> Option<Multiaddr> {
        None
    }

    fn poll(
```

transports/onion/src/address.rs
```diff
@@ -0,0 +1,80 @@
...
+            }
+            Protocol::Dnsaddr(s) => {
+                dns_addr = Some(s);
+            }
+            Protocol::Tcp(p) => {
+                tcp_port_opt = Some(p);
+            }
+            _ => {}
+        }
+    }
```
thomaseizinger: I am not sure if looping towards the end of the address is the best approach here.

I'd suggest we instead pop off two items from the iterator, ignore the rest and match the first one for any of the DNS variants and the second one on `Tcp`. If they are anything else, bail with `MultiAddrNotSuppored` (this is currently done wrong in the `dial` implementation which means the next transport will not get selected in a chain of transports).

This should allow you to not use mutability but just directly assign the variables.

transports/onion/src/address.rs
```diff
@@ -0,0 +1,80 @@
...
+            }
+            Protocol::Dns4(s) => {
+                dns_4 = Some(s);
+            }
+            Protocol::Dns6(s) => {
+                dns_6 = Some(s);
+            }
+            Protocol::Dnsaddr(s) => {
+                dns_addr = Some(s);
+            }
```
thomaseizinger: `/dnsaddr` is a specific protocol that requires resolution of TXT records which won't work with Tor so we shouldn't be using this here!

examples/ping-onion.rs
thomaseizinger: We'll have to think of a different example to actually showcase Tor. The mix of TCP, DNS and Tor is quite confusing here. I'd prefer something that just uses Tor.

If we want to have a separate listening service, I think it would be good to split this into two, where the listening service doesn't know anything about Tor and the dialing service only uses Tor.

Cargo.toml
```diff
@@ -68,6 +71,11 @@ yamux = ["dep:libp2p-yamux"]
 secp256k1 = ["libp2p-core/secp256k1"]
 rsa = ["libp2p-core/rsa"]
 serde = ["libp2p-core/serde", "libp2p-kad?/serde", "libp2p-gossipsub?/serde"]
+onion = ["dep:libp2p-onion"]
```
thomaseizinger: Is this feature-flag by itself useful?

transports/onion/src/lib.rs
```diff
@@ -0,0 +1,193 @@
...
+    }
+
+    fn dial(&mut self, addr: Multiaddr) -> Result<Self::Dial, TransportError<Self::Error>> {
+        let tor_address = if !self.safe_mode {
+            safe_extract_tor_address(&addr).or_else(|_| dangerous_extract_tor_address(&addr))
+        } else {
+            safe_extract_tor_address(&addr)
+        }
+        .map_err(OnionError::from)
+        .map_err(TransportError::Other)?;
```
thomaseizinger: Here we should return a `TransportError::MultiaddrNotSupported` in case we can't dial the address.

thomaseizinger (commented):
Thanks! I've left a few more comments with a mix of nits and other feedback :)

examples/ping-onion.rs
mxinden: You would replace the top line with your name:

```
Copyright 2022 XXX
```

Alternatively you can use:

```
Copyright 2022 Protocol Labs
```

mxinden:
Very happy to see this happening! Thanks @umgefahren for the work and thanks @thomaseizinger for the reviews. Please ping me in case you need my input anywhere.

mxinden:
Cross-referencing related effort on the nim-libp2p side: https://github.com/status-im/nim-libp2p/pull/765

//CC @Menduist and @diegomrsantos

Cargo.toml
umgefahren: No. I will remove it.

umgefahren:
The example seems to be the cause of a lot of difficulties. I would propose postponing the example.

transports/onion/src/lib.rs
umgefahren: Resolved in latest commit.

transports/onion/src/address.rs
umgefahren: Oh I see

examples/ping-onion.rs
```diff
@@ -0,0 +1,119 @@
...
+        println!("bootstrapped!");
+        onion
+    };
+
+    let noise_keys = noise::Keypair::<noise::X25519Spec>::new()
+        .into_authentic(&keypair)
+        .expect("Signing libp2p-noise static DH keypair failed.");
+    Ok(transport
+        .upgrade(upgrade::Version::V1)
+        .authenticate(noise::NoiseConfig::xx(noise_keys).into_authenticated())
```
thomaseizinger: A [recent PR](https://github.com/libp2p/rust-libp2p/pull/2887) simplified the construction of a noise authentication layer. Please use that here too for consistency.

transports/onion/src/address.rs
```diff
@@ -0,0 +1,107 @@
...
+                a.iter().for_each(|e| {
+                    new.push(e);
+                });
+                b.iter().for_each(|e| {
+                    new.push(e);
+                });
+                let new_addr_res = safe_extract_tor_address(&mut new);
+                assert!(new_addr_res.is_ok());
+            }
+        }
```
thomaseizinger: What are we testing here?

transports/onion/src/address.rs
```diff
@@ -0,0 +1,107 @@
...
+
+        let dns_4_test_addr = multiaddr!(Dns4(Cow::Borrowed("dns.ip4.tld")), Tcp(11u16));
+        let dns_4_test_addr_res = safe_extract_tor_address(&mut dns_4_test_addr.clone());
+        assert!(dns_4_test_addr_res.is_ok());
+        assert!(dns_4_test_addr_res.unwrap().is_ip_address().not());
+
+        let dns_6_test_addr = multiaddr!(Dns6(Cow::Borrowed("dns.ip6.tld")), Tcp(12u16));
+        let dns_6_test_addr_res = safe_extract_tor_address(&mut dns_6_test_addr.clone());
+        assert!(dns_6_test_addr_res.is_ok());
+        assert!(dns_6_test_addr_res.unwrap().is_ip_address().not());
```
thomaseizinger: I think this would be clearer if we follow the schema of:

- arrange (construct a list of multiaddresses)
- act (try and convert them into tor addresses)
- assert (that they are all `Ok`)

Currently, there is a lot of text which makes it hard to see, where the input parameters to the test are and what is actually being asserted.

transports/onion/src/address.rs
```diff
@@ -0,0 +1,107 @@
...
+mod tests {
+    use std::{borrow::Cow, net::Ipv4Addr, ops::Not};
+
+    use libp2p_core::{multiaddr::multiaddr, Multiaddr};
+
+    use super::safe_extract_tor_address;
+
+    #[test]
+    fn extract_correct_address() {
+        let dns_test_addr = multiaddr!(Dns(Cow::Borrowed("ip.tld")), Tcp(10u16));
```
thomaseizinger: You may find that constructing the mutliaddress from a string and parsing it is easier to read, i.e. this line could be written as:

```suggestion
        let dns_test_addr = "/dns/ip.tld/tcp/10".parse::<Multiaddr>().unwrap();
```

transports/onion/src/lib.rs
```diff
@@ -0,0 +1,207 @@
...
+}
+
+pub type OnionBuilder<R> = TorClientBuilder<R>;
+
+#[derive(Debug, Clone, Copy, Hash, Default, PartialEq, Eq, PartialOrd, Ord)]
+pub enum AddressConversion {
+    /// uses only dns for address resolution
+    #[default]
+    DnsOnly,
+    /// uses ip and dns for addresses
```
thomaseizinger, line 85: Please link to the docs `arti` docs of `TorAddr::dangerously_from` here.

transports/onion/src/lib.rs
```diff
@@ -0,0 +1,207 @@
...
+    }
+
+    pub async fn bootstrap(&self) -> Result<(), OnionError> {
+        self.client.bootstrap().await.map_err(OnionError::ArtiErr)
+    }
+}
+
+macro_rules! default_constructor {
+    () => {
+        pub async fn default() -> Result<Self, OnionError> {
```
thomaseizinger: Nit: I think people associate `default` with a very particular signature, the one from the `Default` trait. Maybe this is better as `new` or perhaps `bootstrapped`?

transports/onion/src/lib.rs
```diff
@@ -0,0 +1,207 @@
...
+    ArtiErr(#[from] arti_client::Error),
+}
+
+#[derive(Clone)]
+pub struct OnionClient<R: Runtime> {
+    // client is in an Arc, because wihtout it the Transport::Dial method can't be implemented,
+    // due to lifetime issues. With the, eventual, stabilization of static async traits this issue
+    // will be resolved.
+    client: Arc<TorClient<R>>,
+    pub conversion_mode: AddressConversion,
```
thomaseizinger, line 72: Instead of making this public, what do you think about a `with_address_conversion` function? That would allow people to write:

```rust
let client = OnionClient::default().await?.with_address_conversion(AddressConversion::IpAndDns);
```

transports/onion/src/lib.rs
```diff
@@ -0,0 +1,207 @@
...
+mod address;
+mod provider;
+
+#[doc(inline)]
+pub use provider::OnionStream;
+
+#[derive(Debug, thiserror::Error)]
+pub enum OnionError {
+    #[error("error during address translation")]
+    AddrErr(#[from] TorAddrError),
```
thomaseizinger: Is this variant used anywhere, now that we convert that to `MultiaddrNotSupported`?

examples/ping-onion.rs
```diff
@@ -0,0 +1,119 @@
...
+//! "/ip4/0.0.0.0/tcp/24915"`
+//!
+//! In the second terminal window, start a new instance of the example with:
+//!
+//! ```sh
+//! cargo run --example ping-onion -- /ip4/127.0.0.1/tcp/24915
+//! ```
+//!
+//! The two nodes establish a connection, negotiate the ping protocol
+//! and begin pinging each other.
```
thomaseizinger: This example text needs to be updated.

For this to work with the original ping example, the user will need to listen on an external interface and open a port so that their IP is reachable from the internet and thus the Tor network.

Then they can run the ping-onion example against that IP.

examples/ping-onion.rs
```diff
@@ -0,0 +1,119 @@
...
+> {
+    use std::time::Duration;
+
+    let transport = {
+        let onion = onion::OnionClient::from_builder(onion::OnionClient::builder(), false)?;
+        println!("bootstrapping...");
+        onion.bootstrap().await?;
+        println!("bootstrapped!");
+        onion
+    };
```
thomaseizinger: Could make use of our new constructor here.

transports/onion/src/address.rs
```diff
@@ -0,0 +1,107 @@
+use std::net::{IpAddr, SocketAddr};
+
+use arti_client::{IntoTorAddr, TorAddr, TorAddrError};
+use libp2p_core::{multiaddr::Protocol, Multiaddr};
+
+pub fn dangerous_extract_tor_address(multiaddr: &Multiaddr) -> Result<TorAddr, TorAddrError> {
+    let socket_addr = try_extract_socket_addr(&mut multiaddr.clone())?;
```
thomaseizinger: If we end up cloning it, we may as well require ownership of the address directly.

transports/onion/src/address.rs
```diff
@@ -0,0 +1,107 @@
...
+    TorAddr::dangerously_from(socket_addr)
+}
+
+pub fn safe_extract_tor_address(multiaddr: &mut Multiaddr) -> Result<TorAddr, TorAddrError> {
```
thomaseizinger: To align the signatures, this should probably take an owned version too.

transports/onion/src/lib.rs
```diff
@@ -0,0 +1,207 @@
...
+    #[error("error in arti")]
+    ArtiErr(#[from] arti_client::Error),
+}
+
+#[derive(Clone)]
+pub struct OnionClient<R: Runtime> {
```
thomaseizinger:
```suggestion
pub struct OnionTransport<R: Runtime> {
```

This follows the naming convention from the remaining crates.

transports/onion/src/provider.rs
thomaseizinger: I had a look at what we currently do in `libp2p-tcp` and I think it is worth doing the same thing here:

1. For `async-io`, we can directly expose `DataStream` because it implements the traits directly, I don't think it is necessary to wrap that again.
2. For `tokio`, we can wrap `DataStream` and use the adaptors to implement `futures::Async{Read,Write}`.

That way, it is clear which types are exposed for which features and enabling all of them is not a problem and does not require any boolean operations on the feature set.

To realize this, we may a `Provider` abstraction, similar to what we do in `libp2p-tcp`. It should be fairly simple though because we only need to adapt what we do with the return value from `TorClient::connect`.

transports/onion/src/lib.rs
thomaseizinger: I think this may allow us to just use `arti_client::Error` and not need `OnionError` at all?

thomaseizinger (commented):
Thanks for the continued work here :blush: 

I left a few more comments. Looking forward to merging this!

transports/onion/src/address.rs
umgefahren: Possibly uncovered combinations and longer multiaddresses. I can delete it though.

thomaseizinger: Despite multi-addresses allowing for a lot of nesting, I am not sure an address in the form of `/dns4/domain.tld/tcp/8080/dns4/other.tld` makes any sense :/

transports/onion/src/address.rs
umgefahren: I agree, I just always wanted to try that macro 🙃

transports/onion/src/lib.rs
umgefahren: `new` would fit very good in rust conventions, however `bootstrapped` would be better. I chose `default` exactly because it resembles the `Default` trait, which I would chose, if it would allow for async constructors.

transports/onion/src/lib.rs
umgefahren: sounds good and convenient, however I don't think it conflicts with making conversion mode public. There might be reasons why someone would like to change the address conversion mode later after construction.

transports/onion/src/address.rs
umgefahren: I could do that but that might be incorrect because, as you said:
> I'd suggest we instead pop off two items from the iterator, ignore the rest and match the first one for any of the DNS variants and the second one on `Tcp`. If they are anything else, bail with `MultiAddrNotSuppored` (this is currently done wrong in the `dial` implementation which means the next transport will not get selected in a chain of transports).

If it's not mutable we don't get the changes to the multiaddress. However I think there is an easy solution.

umgefahren: Nevermind, I don't see an ideal solution. If we want to make the pop operations count i consider `&mut` a way better solution. I tried it with owned versions, however this would require way more clones. If you have a better idea, I would be glad.

examples/ping-onion.rs
umgefahren: Thanks for the suggestion. I went for the first option :)

transports/onion/src/lib.rs
thomaseizinger: The `with_` function does allow for that though, right? The only thing that is not possible is reading the mode but given that it is static, I am not sure why one would need that.

Another though I just had is that we could embed it in the type signature too. That could allow client code to require a "safe" OnionTransport in a type-safe way. Might be overkill though!

umgefahren: > The with_ function does allow for that though, right? The only thing that is not possible is reading the mode but given that it is static, I am not sure why one would need that.

For sure, I just considered it less natural. I will however remove the `pub` modifier if thats whats necessary.

> Another though I just had is that we could embed it in the type signature too. That could allow client code to require a "safe" OnionTransport in a type-safe way. Might be overkill though!
It might be possible, but I think it's not necessary. The API is pretty obfuscated now anyway and I don't see a real benefit.

examples/ping-onion.rs
```diff
@@ -0,0 +1,114 @@
...
+//! ```
+//!
+//! It will print the PeerId and the listening addresses, e.g. `Listening on
+//! "/ip4/0.0.0.0/tcp/24915"`
+//!
+//! Make sure that the first computer is reachable under one of these ip addresses and port.
+//!
+//! On the second computer run:
+//! ```sh
+//! cargo run --example ping-onion -- /ip4/127.0.0.1/tcp/24915
```
thomaseizinger: Nit: 127.0.0.1 will never be reachable from the internet :upside_down_face:

transports/onion/src/lib.rs
```diff
@@ -0,0 +1,232 @@
+#![doc(html_logo_url = "https://libp2p.io/img/logo_small.png")]
+#![doc(html_favicon_url = "https://libp2p.io/img/favicon.png")]
+#![cfg_attr(docsrs, feature(doc_cfg))]
+//! Tor based transport for libp2p. Connect through the Tor network to TCP listeners.
+//!
+//! Main entrypoint of the crate: [OnionTransport]
```
thomaseizinger:
```suggestion
//! Main entrypoint of the crate: [`OnionTransport`]
```

This will make sure rustdoc flags it as a bad link if it is ever renamed.

examples/ping-onion.rs
```diff
@@ -0,0 +1,114 @@
...
+
+async fn onion_transport(
+    keypair: identity::Keypair,
+) -> Result<
+    libp2p_core::transport::Boxed<(PeerId, libp2p_core::muxing::StreamMuxerBox)>,
+    Box<dyn Error>,
+> {
+    use std::time::Duration;
+
+    let transport = onion::OnionAsyncStdNativeTlsTransport::bootstrapped().await?;
```
thomaseizinger: I believe for this example to work we will need to adjust the `AddressConversion` mode here, otherwise we can dial an IP address.

transports/onion/src/address.rs
thomaseizinger: I've pushed 8c20c301, let me know what you think :)

transports/onion/src/lib.rs
thomaseizinger: I think this is a good compromise. Having the field pub allows users to read it and the `with` function makes it possible to chain it onto an existing instance.

transports/onion/src/lib.rs
```diff
@@ -0,0 +1,232 @@
...
+            conversion_mode,
+            phantom: PhantomData::default(),
+        })
+    }
+
+    pub async fn bootstrap(&self) -> Result<(), OnionError> {
+        self.client.bootstrap().await
+    }
+
+    pub fn with_address_conversion(&mut self, conversion_mode: AddressConversion) -> &mut Self {
```
thomaseizinger: I am never quite sure what the Rust-idiomatic fn signature for builder functions is. I typically go for `self` because it allows usages like:

```rust
let client = OnionTransport::bootstrapped().await?.with_address_conversion(AddressConversion::IpAndDns);
```

Using `&mut self` forces an intermediary variable but I've had people [complain](https://github.com/testcontainers/testcontainers-rs/issues/382) about that in the past. Thoughts?

transports/onion/src/provider.rs
```diff
@@ -0,0 +1,84 @@
+#[cfg(feature = "tokio")]
+use std::pin::Pin;
+
+use arti_client::DataStream;
+use futures::{AsyncRead, AsyncWrite};
+
+pub trait OnionStream: AsyncRead + AsyncWrite + From<DataStream> {}
+
+impl OnionStream for DataStream {}
```
thomaseizinger, line 26: Nice idea!

transports/onion/src/lib.rs
```diff
@@ -0,0 +1,232 @@
...
+        let runtime = tor_rtcompat::tokio::TokioRustlsRuntime::current()
+            .expect("Couldn't get the current tokio rustls runtime");
+        TorClient::with_runtime(runtime)
+    }
+    default_constructor!();
+}
+
+#[cfg(all(feature = "native-tls", feature = "async-std"))]
+#[cfg_attr(docsrs, doc(cfg(all(feature = "native-tls", feature = "async-std"))))]
+pub type OnionAsyncStdNativeTlsTransport =
```
thomaseizinger: Nit: I find it a bit odd that we are splitting `Onion` and `Transport` apart here. How about:

```suggestion
pub type AsyncStdNativeTlsOnionTransport =
```

thomaseizinger (commented):
Thanks!

Let me know what you think of the commit I pushed. A few minor doc fixes, otherwise happy to merge this :)

Would you mind doing a smoke test with the example to check that we aren't shipping something outright broken?

thomaseizinger:
It seems we have a CI failure from `cargo deny` because it can't interpret some license properly :/

thomaseizinger:
> I added libp2p-onion to the default features. However this might be a bad idea, since arti has a lot of dependencies and therefore increases the build time of libp2p greatly.

There is consensus to remove all the default features and I've finally gotten around to implementing that in https://github.com/libp2p/rust-libp2p/pull/2918 so this shouldn't be a concern really :)

umgefahren (commented):
Very nice solution :)

transports/onion/src/lib.rs
umgefahren: Yes it's actually quite difficult and there is no really elegant solution in my opinion. Using Self is less flexible and requires assignments on the other hand it's more clean. I preferred `&mut self` because it allows for changing the conversion mode later on, while Self would be only really appropriate during creation. Since we added this to allow conversion during creation however `Self` would be better.

umgefahren:
> It seems we have a CI failure from `cargo deny` because it can't interpret some license properly :/

It seems like the license of [bounded-vec-deque](https://crates.io/crates/bounded-vec-deque) and the license of [tinystr](https://crates.io/crates/tinystr) seem to have problems. The latter is licensed under Unicode-DFS-2016 which seems like a pretty permisse MIT-like license, but I'm not a lawyer. Although I'm no particular expert, I think we could just add this license to the allowed ones. `bounded-vec-deque` on the other hand causes me a bit of headache. Not only because the repository linked to, seems to be dead. I could file an issue with Arti describing the issues... although this would probably require writing an alternative.

transports/onion/src/lib.rs
thomaseizinger: Yeah, there isn't really a usecase where you hold on to a `Transport` in libp2p. You construct them and pass them to the `Swarm`, at which point it is moved.

thomaseizinger:
> > It seems we have a CI failure from `cargo deny` because it can't interpret some license properly :/
> 
> It seems like the license of [bounded-vec-deque](https://crates.io/crates/bounded-vec-deque) and the license of [tinystr](https://crates.io/crates/tinystr) seem to have problems. The latter is licensed under Unicode-DFS-2016 which seems like a pretty permisse MIT-like license, but I'm not a lawyer. Although I'm no particular expert, I think we could just add this license to the allowed ones. `bounded-vec-deque` on the other hand causes me a bit of headache. Not only because the repository linked to, seems to be dead. I could file an issue with Arti describing the issues... although this would probably require writing an alternative.

One can inspect the actual source code here: https://docs.rs/crate/bounded-vec-deque/0.1.1/source/

It is dual-licensed under GPL3 and BSD. You should be able to somehow tell `cargo deny`, what license it actually is. I think we had to do that for `ring` too but I am no expert in this tool, sorry :(

Regarding `tinystr`, I'd like to have @mxinden's thoughts.

In both cases, it may be worth reaching out to the arti developers and letting them know that these two crates are being flagged here. Not necessarily with the goal of having them replaced but it is likely good to know. They don't seem to use `cargo deny` so they probably just don't know about these crates :)

mxinden:
> I could file an issue with Arti describing the issues... although this would probably require writing an alternative.

> In both cases, it may be worth reaching out to the arti developers and letting them know that these two crates are being flagged here. Not necessarily with the goal of having them replaced but it is likely good to know. They don't seem to use `cargo deny` so they probably just don't know about these crates :)

That would be my preference. @umgefahren would you mind reaching out to them?

In case they decide to stick with the two crates, which of course is very much a valid option, I can try to organize some legal advice to learn more about the special licenses. I myself don't feel knowledgeable enough to make a decision here.

examples/ping-onion.rs
```diff
@@ -0,0 +1,117 @@
+// Copyright 2022 Hannes Furmans
```
mxinden, line 1: I would suggest moving this example to `transports/onion/examples`. Not a strong opinion though.

Either way, I think it should be listed on `examples/README.md`.

transports/onion/Cargo.toml
```diff
@@ -0,0 +1,33 @@
+[package]
+name = "libp2p-onion"
+version = "0.1.0"
+edition = "2021"
+license = "MIT"
+resolver = "2"
+
+
```
mxinden:
```suggestion
```

Same below.

transports/onion/Cargo.toml
```diff
@@ -0,0 +1,33 @@
...
+[dev-dependencies]
+libp2p = { version = "0.49", path = "../../" }
+tokio-crate = { package = "tokio", version = "1", features = ["rt", "macros"] }
+async-std-crate = { package = "async-std", version = "1", features = ["attributes"] }
+
+
+[features]
+default = ["async-std", "native-tls"]
+tokio = ["arti-client/tokio", "dep:tokio-crate"]
+async-std = ["arti-client/async-std"]
```
mxinden:
```suggestion
async-std = ["arti-client/async-std", "dep:async-std-crate"]
```

I might be missing something here.

transports/onion/Cargo.toml
```diff
@@ -0,0 +1,33 @@
...
+native-tls = ["arti-client/native-tls"]
+rustls = ["arti-client/rustls"]
+
+[package.metadata.docs.rs]
+rustdoc-args = ["--cfg", "docsrs"]
+rustc-args = ["--cfg", "docsrs"]
```
mxinden, line 30: What is this needed for?

transports/onion/src/address.rs
```diff
@@ -0,0 +1,147 @@
+use arti_client::{DangerouslyIntoTorAddr, IntoTorAddr, TorAddr};
```
mxinden, line 21: Missing license header.

transports/onion/src/lib.rs
```diff
@@ -0,0 +1,233 @@
+#![doc(html_logo_url = "https://libp2p.io/img/logo_small.png")]
```
mxinden, line 21: Missing license header.

transports/onion/src/lib.rs
```diff
@@ -0,0 +1,233 @@
...
+//! # use async_std_crate as async_std;
+//! # use libp2p_core::Transport;
+//! # async fn test_func() -> Result<(), Box<dyn std::error::Error>> {
+//! let address = "/dns/www.torproject.org/tcp/1000".parse()?;
+//! let mut transport = libp2p_onion::AsyncStdNativeTlsOnionTransport::bootstrapped().await?;
+//! // we have achieved tor connection
+//! let _conn = transport.dial(address)?.await?;
+//! # Ok(())
+//! # }
+//! # async_std::task::block_on(async { test_func().await.unwrap() });
```
mxinden, line 39: Would you mind fixing the indentation here? As far as I know it will propagate to the rendered HTML.

transports/onion/src/lib.rs
```diff
@@ -0,0 +1,233 @@
...
+#[doc(inline)]
+pub use provider::OnionTokioStream;
+
+use provider::OnionStream;
+
+pub type OnionError = arti_client::Error;
+
+#[derive(Clone)]
+pub struct OnionTransport<R: Runtime, S> {
+    // client is in an Arc, because wihtout it the Transport::Dial method can't be implemented,
```
mxinden:
```suggestion
    // client is in an Arc, because without it the [`Transport::dial`] method can't be implemented,
```

transports/onion/src/lib.rs
```diff
@@ -0,0 +1,233 @@
...
+    phantom: PhantomData<S>,
+}
+
+/// Configure the onion transport from here.
+pub type OnionBuilder<R> = TorClientBuilder<R>;
+
+/// Mode of address conversion. Refer tor [arti_client::TorAddr](https://docs.rs/arti-client/latest/arti_client/struct.TorAddr.html) for details.
+#[derive(Debug, Clone, Copy, Hash, Default, PartialEq, Eq, PartialOrd, Ord)]
+pub enum AddressConversion {
+    /// uses only dns for address resolution (default)
```
mxinden:
```suggestion
    /// Uses only dns for address resolution (default).
```

For the sake of consistency, I would prefer comments to (a) be full sentences and (b) start with a capital letter.

mxinden (commented):
Thanks for the work here!

How about reaching out to the Arti maintainers, asking whether they would be so kind and give this a review as well? @umgefahren are you in touch with them already? I have been in touch with them in the past, thus happy to do the reach out.

umgefahren:
> How about reaching out to the Arti maintainers, asking whether they would be so kind and give this a review as well? @umgefahren are you in touch with them already? I have been in touch with them in the past, thus happy to do the reach out.

I tried to get in touch with them, however I wasn't successful or didn't try hard enough. So it would be nice, if you could reach out to them for now.

transports/onion/Cargo.toml
umgefahren: I can't really try this, however I used this in my other crates in order to enable the display of features in the docs.

examples/ping-onion.rs
umgefahren: I think we should leave it there, since all examples are in the examples/ directory and not in the corresponding parts. If we change that for all, we should change it for libp2p-onion too, however I think this is more concise.

transports/onion/src/lib.rs
umgefahren: image

I don't see any indentation issues, if you see some, i will correct them.

transports/onion/Cargo.toml
umgefahren: I might be missing something too, but it doesn't seem to change anything.

transports/onion/src/address.rs
umgefahren: Resolved in latest commit.

transports/onion/src/lib.rs
umgefahren: Resolved in latest commit.

transports/onion/Cargo.toml
thomaseizinger: What this does is tell docs.rs to pass the `docsrs` conditional compilation flag to rustc, meaning any code that is guarded with `cfg(docsrs)` will now be included in the build.

We use this in the `libp2p` root crate to conditionally include the `doc` attribute: https://github.com/libp2p/rust-libp2p/blob/04b07c4d14a0a6ec350fc21261a33d859fc71289/src/lib.rs#L53

We need to do this conditionally because the `#[doc(cfg)]` attribute is [not yet stable](https://github.com/rust-lang/rust/issues/43781). docs.rs runs all their build pipelines on nightly so we can use it there.

Now, interestingly, we don't set this in our `libp2p` crate so all of these `cfg_attr` attributes have been doing nothing all this time :)

In this crate, it is used here: https://github.com/umgefahren/rust-libp2p/blob/6f4bca5bbfd141bc18d9d931c5d28b50a212f0fe/transports/onion/src/provider.rs#L31-L36

And will tell docs.rs to include a little "only available with feature `tokio`" tag next to the item.

examples/ping-onion.rs
thomaseizinger: > I think we should leave it there, since all examples are in the examples/ directory and not in the corresponding parts. If we change that for all, we should change it for libp2p-onion too, however I think this is more concise.

IMO the examples in `/examples` should showcase use of the root `libp2p` crate which usually involves multiple different protocols (identify + kad + ping etc). Most protocols in `/protocols` have their own examples that showcase specific aspects of them.

With this reasoning, the `ping` example shouldn't be there though and neither should `mdns-passive-discovery` and `gossipsub-chat`.

transports/onion/Cargo.toml
thomaseizinger: This is actually something that should be fixed. I'll open an issue about it.

umgefahren:
Is there anything left to do for me?

transports/onion/Cargo.toml
thomaseizinger: Have you tried running with `--no-default-features --feature async-std`? That should trigger a problem I think.

thomaseizinger:
> Is there anything left to do for me?

Yes, sorry for the delay!

- Please fix the merge conflicts
- Please adopt our new `feature` handling (no more default features). See #2918.
- Please add an override of some sort for cargo deny so that it is happy :)
- Please add a changelog file and the necessary entries to the root `CHANGELOG.md`

umgefahren:
I implemented all changes requested. I hope you are pleased with my resolution of the issues with `cargo deny`.

CHANGELOG.md
```diff
@@ -48,6 +49,9 @@
 - Remove default features. You need to enable required features explicitly now. As a quick workaround, you may want to use the
   new `full` feature which activates all features. See [PR 2918].
 
+- New `libp2p-onion` crate
+  ([PR 2899](https://github.com/libp2p/rust-libp2p/pull/2899)).
+
```
umgefahren: I added a comment to the changelog for 0.49.0. Although I'm not actually sure this pull request will be merged in time and it's not mentioned in #2931

thomaseizinger:
> I implemented all changes requested. I hope you are pleased with my resolution of the issues with `cargo deny`.

Yep, looks all good! I just pushed a few commits with minor nits, hope you don't mind :)
This looks good to be merged!

I did notice that `libp2p-onion` does not compile without default features because of an issue in `arti_client`. It references `PreferredRuntime` without a feature gate but the symbol itself is feature-gated.

This shouldn't be an issue for us though as we only allow users to depend on the crate by selecting one of the two runtimes.

thomaseizinger (approved):
This is ready to be merged!
Thank you @umgefahren :)

@mxinden Please go ahead with merging if it is also ready from your end.

umgefahren:
Are there any issues left for me to resolve? @mxinden

mxinden:
:wave: @nmathewson and ijackson,

I hope you don't mind me pinging you here. First off, thank you for your work on https://gitlab.torproject.org/tpo/core/arti. You might remember a past call with me during the early days of Arti.

As you can tell from the pull request description, @umgefahren is integrating Arti with rust-libp2p, thus enabling a rust-libp2p node to dial a remote endpoint through the Tor network.

I am tagging you here for the following reasons:

1. Might be cool for you to see where Arti is being used.
2. In case you have time, your input on this pull request would be very valuable. No worries if you don't have time.
3. Getting your opinion on the license issues we are facing with the import of Arti. More specifically the crates `bounded-vec-deque` and `tinystr`. See https://github.com/libp2p/rust-libp2p/pull/2899#issuecomment-1254760056. Again, no worries if you don't have time.

thomaseizinger:
Can you fix the clippy error please? See https://github.com/libp2p/rust-libp2p/pull/2859 for reference.

umgefahren:
Done

umgefahren:
Why did the CI fail?

thomaseizinger:
> Why did the CI fail?

Some download issue during setup of the jobs. I re-triggered them!

mxinden:
I did not forget about this. I am sorry for the delay.

CHANGELOG.md
```diff
@@ -48,6 +49,8 @@
 - Remove default features. You need to enable required features explicitly now. As a quick workaround, you may want to use the
   new `full` feature which activates all features. See [PR 2918].
 
+
+- New `libp2p-onion` crate. See [PR 2899].
```
thomaseizinger: This needs to go into a new section of the changelog :)

umgefahren: I moved it, I thought it would make it into the 0.49.0 release. Didn't expect the delays.

Cargo.toml
```diff
@@ -49,6 +49,10 @@ full = [
     "wasm-ext-websocket",
     "websocket",
     "yamux",
+    "onion-async-std",
+    "onion-tokio",
+    "onion-native-tls",
+    "onion-rustls",
```
thomaseizinger: Please keep this list alphabetically sorted!

Cargo.toml
```diff
@@ -85,7 +89,11 @@ secp256k1 = ["libp2p-core/secp256k1"]
 rsa = ["libp2p-core/rsa"]
 ecdsa = ["libp2p-core/ecdsa"]
 serde = ["libp2p-core/serde", "libp2p-kad?/serde", "libp2p-gossipsub?/serde"]
-tokio = ["libp2p-mdns?/tokio", "libp2p-tcp?/tokio", "libp2p-dns?/tokio"]
+onion-async-std = ["dep:libp2p-onion", "libp2p-onion?/async-std"]
+onion-tokio = ["dep:libp2p-onion", "libp2p-onion?/tokio"]
+onion-native-tls = ["dep:libp2p-onion", "libp2p-onion?/native-tls"]
+onion-rustls = ["dep:libp2p-onion", "libp2p-onion?/rustls"]
+tokio = ["libp2p-mdns?/tokio", "libp2p-tcp?/tokio", "libp2p-dns?/tokio", "libp2p-onion?/tokio"]
 async-std = ["libp2p-mdns?/async-io", "libp2p-tcp?/async-io", "libp2p-dns?/async-std"]
```
thomaseizinger: This feature should activate the `async-std` feature in `libp2p-onion`, same for the tokio one.

Cargo.toml
```diff
@@ -85,7 +89,11 @@ secp256k1 = ["libp2p-core/secp256k1"]
 rsa = ["libp2p-core/rsa"]
 ecdsa = ["libp2p-core/ecdsa"]
 serde = ["libp2p-core/serde", "libp2p-kad?/serde", "libp2p-gossipsub?/serde"]
-tokio = ["libp2p-mdns?/tokio", "libp2p-tcp?/tokio", "libp2p-dns?/tokio"]
+onion-async-std = ["dep:libp2p-onion", "libp2p-onion?/async-std"]
+onion-tokio = ["dep:libp2p-onion", "libp2p-onion?/tokio"]
```
thomaseizinger: Please remove these in favor of a single `onion` one and forwarding of the tokio and async-std ones in the `tokio` and `async-std` feature.

CHANGELOG.md
```diff
@@ -43,11 +44,18 @@
...
+
+[PR 2899]: https://github.com/libp2p/rust-libp2p/pull/2899
+
 # 0.49.0
 
 - Remove default features. You need to enable required features explicitly now. As a quick workaround, you may want to use the
   new `full` feature which activates all features. See [PR 2918].
 
+
```
```suggestion
```

thomaseizinger (approved):
LGTM!

A few minor changes please.
@mxinden are you happy to merge this?

umgefahren:
@thomaseizinger could you add the hacktoberfest-accepted label :)

transports/onion/Cargo.toml
```diff
@@ -0,0 +1,30 @@
+[package]
+name = "libp2p-onion"
+version = "0.1.0"
```
mxinden:
```suggestion
version = "0.1.0-alpha"
```

Like with https://github.com/libp2p/rust-libp2p/pull/2622 and https://github.com/libp2p/rust-libp2p/pull/2289 I suggest releasing this new crate as an `alpha` for now. Objections?

mxinden (approved):
Merging here sounds good to me.

I reached out to @willscott on the licensing matters. He has no objections. I will reach out to our legal team as well, just in case. I don't think we need to block on this.

I suggest we wait for https://github.com/libp2p/test-plans/issues/59 and merge with a green CI though.

transports/onion/Cargo.toml
umgefahren: Sounds good to me.

mxinden: :exploding_head: wow that was fast. Thanks @umgefahren.

umgefahren:
Is it ok that the interoperability check always fails? And since the semver-checks were introduced in #2635 the CI doesn't get green anymore sadly.

mxinden:
> I will reach out to our legal team as well, just in case. I don't think we need to block on this.

No objections from our legal team.

mxinden:
> Is it ok that the interoperability check always fails? And since the semver-checks were introduced in #2635 the CI doesn't get green anymore sadly.

Interoperability tests are back to green with https://github.com/libp2p/test-plans/pull/60/.

thomaseizinger:
> > I will reach out to our legal team as well, just in case. I don't think we need to block on this.
> 
> No objections from our legal team.

On what in particular? Removing the licence headers in each file?

mxinden:
> > > I will reach out to our legal team as well, just in case. I don't think we need to block on this.
> > 
> > 
> > No objections from our legal team.
> 
> On what in particular? Removing the licence headers in each file?

No. Regarding the two crates with licensing concerns.

marten-seemann:
I'd be hesitant to merge this code into the main repository, for multiple reasons:
1. False sense of security: libp2p, and applications built on top of libp2p, broadcast their IDs and multiaddresses at many different places. It's really easy to accidentally reveal your identity, probably even if you turn off the obvious candidates like Identify, AutoNAT and DCuTR. libp2p has not been designed with privacy goals as strict as Tor's in mind, and making sure that the stack has the privacy properties that people expect when they hear that "libp2p over Tor is possible" would be a _massive_ engineering effort. There's probably a lot more ways to compromise yourself other than AutoNAT just publishing your non-Tor addresses.
In all likelihood, adding a Tor transport would not yield the privacy properties that people are looking for. Keep in mind that Tor is used by dissidents all around the world, who are not experts in network engineering. It is our responsibility to not endanger their safety by overpromising on the privacy properties of the software we publish.
3. Usability: DHT lookups are already slow. Adding the Tor hops to the equation will make things even more painful. There might be optimizations one could achieve from a deeper integration between libp2p and Tor (after all, Tor is running a DHT as well), but that would be a larger effort and require a really in-depth understanding of both IPFS and Tor.
4. Ethical: If the libp2p Tor transport gained a lot of adoption, this might pose a problem for the Tor network. It's a fairly small network with less than 10k relays, handling important traffic of people who actually rely on the privacy properties of the network.

Merging a Tor transport into a PL-maintained libp2p implementation sends a signal that this transport is properly tested and vetted, no matter how many disclaimers we include. As a modular stack, it should be possible to have this code live outside of rust-libp2p (and outside the libp2p org). This would make it possible for people to play around and experiment with this code, without giving it the blessing it would receive from being a part of rust-libp2p.

marten-seemann:
Adding more context here. The ask for a Tor transport is as old as IPFS / libp2p itself. See for example this discussion from 2016 (!): https://github.com/libp2p/go-libp2p/pull/79

Menduist:
I'll reply to your points from nim-libp2p point of view, since we'll soon merge our Tor transport. As always, you are free to do want you want, I'm just giving another perspective :)

1. We are going to recommend (or even force) application developers to create a second switch for the Tor transport only. Trying to isolate the Tor transport from other transports properly in a single switch is very tricky and will probably fail, since a lot of state is shared (conn manager, identity, etc)
2. The application developer will then have to balance, and think about which operations are safe to put on the regular switch, and which one are privacy risk and should go on the Tor switch (at the cost of latency & bandwidth lost etc). For instance, participating in a gossipsub mesh is probably fine on the regular switch, but publishing on it should go on the Tor switch
3. I mean, the Tor network exists to be used. Of course people can abuse it and use a lot of resources from it without actually needing to. But they don't need our help for that. For me, Tor transport is a first step in privacy protection in libp2p at the network level. If we start to see a lot of adoption, it may be the sign that we need to build onion routing (or equivalent) in libp2p networks themselves, instead of leaching another network. But today we can't know if that appetite for network level privacy exist


EDIT: and just to clarify: the goal of the Tor transport is not to say "you just enable this flag and your application becomes magically private", it's here to say "if you are building an application with privacy in mind, you can use this flag as one of the tool used to keep it secure". But it's so easy to leak informations at every layer, the whole application has to be built with privacy in mind

diegomrsantos:
We are considering:

- Making the TorTransport private
- Creating a TorSwitch where it's not possible to add other transports and name resolving is disabled (see Tor design paper section 4.3 Opening and closing streams). It will force the use of a new dedicated switch with a new peer id.

umgefahren:
First, I want to thank for all the thoughts and comments on this PR. I see pretty much every point made here, but I have some additional things to say, and I want to offer a way to go forward.

## Regarding @marten-seemann comments:
- Of course, "libp2p over Tor" may imply more privacy than this offers. But I don’t think it’s that big of an issue. If we are honest about the caveats that lay here, we can somewhat trust developers to handle the transport with the care it requires. After all, libp2p is not faced towards direct users, so the responsibility is somewhat with the libp2p consumers and thus not directly with the users of Tor. If we include some notices in the documentation and continue our effort here, I consider it a hand able risk.
- Usability is an issue though. I can’t really argue against that. Although there might some usage profiles where the drawbacks are worth the benefits. Maybe the Tor transport is only used to retrieve data directly from a peer via Request-Response. Something like this would require the "switch" described later.
- While I understand your ethical considerations, I also agree with @Menduist. While we don’t make abuse easier, we make it a lot easier to write privacy focused software and possibly help the groups under attack we want to protect.

## Regarding @Menduist:
I pretty much agree with all you said. I think we in this repository should look at methods on how we can steer behaviors towards using a given transport and also take a look at nim-libp2p here. But I think this is beyond the scope of this PR. I will open an issue about it, as soon as this is merged.

Finally, I want to leave some comments of my own here.

The reason why I wrote this transport was because I’m really interested in bringing transport level privacy to libp2p. The new iteration of the Dione project I wrote would benefit heavily from it and I could think of a lot of other applications where transport level privacy would be good. I recognize it will be quite hard to properly integrate that into libp2p, but we should do it anyway. I would also recommend looking into writing a generic (in customizable) but tested and capable onion router as libp2p behavior and transport. I would say it’s well in the scope of the project and very doable. In fact, there are Onion Routers written with libp2p (i.e. [hoprnet](https://github.com/hoprnet/hoprnet)).

## Recommendations
If it’s fine by @mxinden and @thomaseizinger, I’m still in favor of merging here, although I would like to take some steps previously.
1.	Include a privacy notice in the documentation.
2.	Rename everything onion to Tor (we could add other onion routers later and libp2p-onion would fit better if we were ever to develop our own onion router).

We should monitor how this transport develops and is being used. Throwing it out is of course always an option. However, I would do that, once we have better alternatives, transport level privacy seems to be a concern to some people.

marten-seemann:
The requests for Tor support in libp2p / IPFS are as old as the project itself. Over the years, the IPFS and libp2p team has repeatedly decided to not merge any Tor transport into the main project.

Where is the push to merge this into the main repository coming from? Why does this transport need the implicit blessing that comes with living in the main repo? One of the advantages of libp2p being a modular networking stack is that it is possible to include your own transports from repositories outside the main project.

umgefahren:
You are right, of course it could live outside of rust-libp2p. If this doesn't get merged, I will extract it into it's own repository. But I don't want to be the person making that call.

thomaseizinger:
> Where is the push to merge this into the main repository coming from? Why does this transport need the implicit blessing that comes with living in the main repo? One of the advantages of libp2p being a modular networking stack is that it is possible to include your own transports from repositories outside the main project.

From a maintainer PoV, I definitely see the appeal of certain transports living outside of the monorepo. This does however require us (@libp2p/rust-libp2p-maintainers) to make a commitment to API stability and reducing the number of breaking changes each version. Otherwise, consuming those out-of-tree transports is just a pain.

What if we introduce a `libp2p/rust-libp2p-community`? This could be a place for community-driven transports and behaviours where people such as @umgefahren can maintain their crates without requiring our blessing.

thomaseizinger:
> This does however require us (@libp2p/rust-libp2p-maintainers) to make a commitment to API stability and reducing the number of breaking changes each version. 

I opened a separate discussion for this here: https://github.com/libp2p/rust-libp2p/discussions/3072

umgefahren:
>What if we introduce a `libp2p/rust-libp2p-community`? This could be a place for community-driven transports and behaviours where people such as @umgefahren can maintain their crates without requiring our blessing.

That sounds like a great compromise to me. Should I close this PR and open a new one at the new repository?

thomaseizinger:
> > What if we introduce a `libp2p/rust-libp2p-community`? This could be a place for community-driven transports and behaviours where people such as @umgefahren can maintain their crates without requiring our blessing.
> 
> That sounds like a great compromise to me. Should I close this PR and open a new one at the new repository?

Let's wait for what the other maintainers have to say about this :)

thomaseizinger:
> > Where is the push to merge this into the main repository coming from? Why does this transport need the implicit blessing that comes with living in the main repo? One of the advantages of libp2p being a modular networking stack is that it is possible to include your own transports from repositories outside the main project.
> 
> From a maintainer PoV, I definitely see the appeal of certain transports living outside of the monorepo. This does however require us (@libp2p/rust-libp2p-maintainers) to make a commitment to API stability and reducing the number of breaking changes each version. Otherwise, consuming those out-of-tree transports is just a pain.
> 
> What if we introduce a `libp2p/rust-libp2p-community`? This could be a place for community-driven transports and behaviours where people such as @umgefahren can maintain their crates without requiring our blessing.

Friendly ping @mxinden @jxs @elenaf9 :)

thomaseizinger:
I think for the time being, it is better to leave this out of tree. Let's discuss more over at #3072 how we can stabilize the APIs to make maintaining transports and behaviours out of tree more pleasant.

@umgefahren Feel free to tag me or any other maintainer any time you want feedback or if you are struggling with an upgrade.

In the meantime, we can extend our README with a link to your repository to give it some visibility. Also, `libp2p-tor` and `libp2p-onion` are still available on crates.io :)

umgefahren:
Thanks, I will publish the work I've done here on my own repository.

umgefahren:
If someone is interested: https://github.com/umgefahren/libp2p-tor

PR #2900, opened 2022-09-13
*: Fix various clippy warnings

umgefahren:
# Description

I just implemented some changes suggested by clippy.

## Change checklist


thomaseizinger (approved):
Thank you!

We do run clippy in CI but with a reduced set of lints. Can we perhaps narrow that set down, now that we have fixed a couple more?

umgefahren:
Yes we could do that. I'm wondering why clippy doesn't pick that up now. I think it might be because the actions still use actions-rs which is not really necessary anymore, since Rust comes with the runners now and actions-rs might be outdated. 

I could make an additional PR updating the Workflows.

Many projects also include some lints from the [restriction group](https://rust-lang.github.io/rust-clippy/master/). This could be considered.

thomaseizinger:
> Yes we could do that. I'm wondering why clippy doesn't pick that up now. I think it might be because the actions still use actions-rs which is not really necessary anymore, since Rust comes with the runners now and actions-rs might be outdated.
> 
> I could make an additional PR updating the Workflows.

Feel free, I am always in favor of using tooling better.

> Many projects also include some lints from the [restriction group](https://rust-lang.github.io/rust-clippy/master/). This could be considered.

Feel free to open an issue with a list of suggestions so we can discuss it :)

mxinden (approved):
Thanks @umgefahren.

I will merge here, assuming that the above follow-ups happen in separate pull requests.

mxinden:
Ignoring Testground failures for now. See https://github.com/libp2p/test-plans/pull/41.

PR #2903, opened 2022-09-14
protocols/noise: Add `NoiseConfig::with_prologue`

### LINKED ISSUE #2791: transports/noise: Expose Prologue mechanism
## Description

Enable users to provide a [Noise Prologue](https://noiseprotocol.org/noise.html#prologue) via [Builder::prologue](https://docs.rs/snow/latest/snow/struct.Builder.html#method.prologue).

## Motivation

Needed for https://github.com/libp2p/rust-libp2p/pull/2622, more specifically to validate the remote's TLS fingerprints. See https://github.com/libp2p/specs/pull/412#discussion_r930582351.

Maybe. In case someone wants to pick this up instead, please comment to make sure we are not duplicating work.

thomaseizinger:
# Description

Allows users to set the noise prologue of the handshake.

- #2791
- #2622

## Change checklist

- [X] A changelog entry has been made in the appropriate crates


thomaseizinger:
That was surprisingly simple :)

melekes (approved):
👍

transports/noise/src/lib.rs
```diff
@@ -88,6 +89,15 @@ impl<H, C: Zeroize, R> NoiseConfig<H, C, R> {
...
     }
 
+    /// Set the noise prologue.
+    ///
+    /// The prologue can contain arbitrary data and will be hashed into the noise handshake.
+    /// For the handshake to succeed, both parties must set the same prologue.
+    pub fn with_prologue(&mut self, prologue: Vec<u8>) -> &mut Self {
+        self.prologue = prologue;
+        self
+    }
```
thomaseizinger: I made this a `with_` function instead of exposing it in every constructor because it is often not needed and this way, we can keep the entire prologue functionality hidden from users that just want to use regular noise for their connections.

Having said that, if we agree to merge https://github.com/libp2p/rust-libp2p/pull/2887, then most users don't need to touch `NoiseConfig` anyway so perhaps it should just be a regular constructor parameter?

transports/noise/src/lib.rs
```diff
@@ -88,6 +89,15 @@ impl<H, C: Zeroize, R> NoiseConfig<H, C, R> {
         NoiseAuthenticated { config: self }
     }
 
+    /// Set the noise prologue.
+    ///
+    /// The prologue can contain arbitrary data and will be hashed into the noise handshake.
+    /// For the handshake to succeed, both parties must set the same prologue.
```
mxinden: How about linking to https://noiseprotocol.org/noise.html#prologue here?

transports/noise/src/lib.rs
```diff
@@ -185,6 +199,7 @@ where
         let session = self
             .params
             .into_builder()
+            .prologue(self.prologue.as_ref())
```
mxinden: Do I understand correctly that setting no prologue is equal to setting an empty prologue?

mxinden (commented):
Thanks for looking into this!

I don't have an opinion on the concrete API.

Can you include a unit test in this pull request? Especially since this is very security critical for the WebRTC protocol.

transports/noise/src/lib.rs
thomaseizinger: Yes, that is my understanding. The prologue is hashed into the protocol. Appending an empty slice vs appending nothing is not going to change the final hash.

thomaseizinger:
> Can you include a unit test in this pull request? Especially since this is very security critical for the WebRTC protocol.

What would you like the test to do? I am not against it but perhaps it would be better to have a test on the WebRTC level where we check that the handshake fails if we provide different certificates. That is the functionality we want. Doing that through the noise prologue is an implementation detail so it can of feels like we'd have a test there anyway, making this one redundant?

thomaseizinger:
> Can you include a unit test in this pull request? Especially since this is very security critical for the WebRTC protocol.

I've tried to create an abstraction in-between that tests that we are passing the prologue on correctly. While doing that, I noticed some code duplication between the various protocols. That got me thinking on why we even support so many handshake types when the specification only mentions a particular one.

What do you think of removing the others?

transports/noise/src/lib.rs
```diff
@@ -174,25 +221,22 @@ impl<T, C> InboundUpgrade<T> for NoiseConfig<IX, C>
...
 
     fn upgrade_inbound(self, socket: T, _: Self::Info) -> Self::Future {
-        let session = self
-            .params
-            .into_builder()
-            .local_private_key(self.dh_keys.secret().as_ref())
-            .build_responder()
-            .map_err(NoiseError::from);
+        let config = self.legacy;
+        let identity = self.dh_keys.clone().into_identity();
```
thomaseizinger, line 232: I had to clone the keys here to make the borrow-checker happy. Not ideal but it removes a bit of duplication. I am taking suggestions on how to do it better :)

thomaseizinger:
> > Can you include a unit test in this pull request? Especially since this is very security critical for the WebRTC protocol.
> 
> I've tried to create an abstraction in-between that tests that we are passing the prologue on correctly. While doing that, I noticed some code duplication between the various protocols. That got me thinking on why we even support so many handshake types when the specification only mentions a particular one.
> 
> What do you think of removing the others?

I am getting the impression that the abstractions within `libp2p-noise` are not quite ideal. For example, we have this entire layer of `handshake::*` functions that are all only used once (apart from `rt1_initiator` which is used twice) and they all almost correspond 1-to-1 to implementations of `{In,Out}boundUpgrade` on `NoiseConfig`.

I think it would make sense to simplify this, even if it means there will be a slight duplication between the `IX` and `IK` upgrades.

melekes (approved):

transports/noise/src/lib.rs
mxinden: How about implementing `into_responder` on `ProtocolParams` and not on `NoiseConfig`?

mxinden:
> That got me thinking on why we even support so many handshake types when the specification only mentions a particular one.

I think this dates back to the initial Noise specification days where it wasn't quite clear yet which handshake pattern to use. I don't have a strong opinion here, though an intuition to keep supporting them.

mxinden:
> [Add unit tests for handshake hashes](https://github.com/libp2p/rust-libp2p/pull/2903/commits/5b0c4cbdb85021ad2d9b77b596742833df0a8272)

:pray: test looks good to me.

transports/noise/src/lib.rs
thomaseizinger: We'd have to pass the prologue and keys as an argument then. I guess we can do that :)

thomaseizinger: Unfortunately, that doesn't work well with the tests I've written. Passing the prologue in breaks the abstraction layer and the tests will no longer reflect the same usage as the code.

If we decide to merge https://github.com/libp2p/rust-libp2p/pull/2909, more optimizations might be possible!

thomaseizinger:
@mxinden Do you think we can merge this as is (i.e. with the `.clone()`) or would like me to find a different way? If so, I'd advocate for merging #2887 first because the removed indirection will make it easier to optimize the data flow.

mxinden (approved):
I am fine merging here despite the `clone`. I doubt this would ever be a performance issue.

Thanks @thomaseizinger for the work. Sorry @melekes for the delay for https://github.com/libp2p/rust-libp2p/pull/2622.

mxinden:
Interoperability test failures are due to https://github.com/libp2p/test-plans/pull/41. Merging here.

PR #2915, opened 2022-09-19
core/identity: Implement `Hash` and `Ord` for PublicKey

### LINKED ISSUE #2910: Ord, Eq, Hash instances for PublicKey for use with collections (sets, maps)
## Description

Implement those traits so that `PublicKey` can be used as a key in `BTreeSet`, `HashSet`, `BTreeMap` and `HashMap`.

## Motivation

I want to store a set of public keys for authorization. I can not get away with only storing `PeerId`s because I need to verify signatures.  

## Current Implementation

No implementation for `PublicKey`, but there are implementations for `PeerId`.

Yes if this enhancement is wanted by maintainers.

futpib:
# Description

Implemented `Ord` and `Hash` instances for `PublicKey` to enable putting them in a `HashSet`/`BTreeSet`.

Fixes #2910

## Change checklist


core/CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+# 0.36.1 - [unreleased]
+
+- Added `Hash` and `Ord` instances for `PublicKey`. See [PR 2915].
```
```suggestion
- Implement `Hash` and `Ord` for `PublicKey`. See [PR 2915].
```

core/src/identity.rs
```diff
@@ -379,4 +379,56 @@ mod tests {
 
         assert_eq!(expected_peer_id, peer_id);
     }
+
+    #[test]
+    fn public_key_hash() {
```
thomaseizinger: I think a better test would be something like:

```rust
#[test]
fn impls_hash() {
	assert_impls_hash::<PublicKey>();
}

fn assert_impls_hash<T: Hash>() { }
```

We don't need to test an automatically generated implementation for its behaviour.

core/src/identity/ecdsa.rs
```diff
@@ -222,6 +223,12 @@ impl fmt::Debug for PublicKey {
     }
 }
 
+impl hash::Hash for PublicKey {
+    fn hash<H: hash::Hasher>(&self, state: &mut H) {
+        self.to_bytes().hash(state);
+    }
+}
```
thomaseizinger, line 237: This will likely trigger this clippy lint: https://rust-lang.github.io/rust-clippy/master/#derive_hash_xor_eq

Please add a test that the property documented in the lint holds, ideally with a prop test :)

core/src/identity/ed25519.rs
```diff
@@ -126,6 +128,26 @@ impl fmt::Debug for PublicKey {
     }
 }
 
+// Remove (derive instead) when this PR is merged:
+// https://github.com/dalek-cryptography/ed25519-dalek/pull/176
+impl hash::Hash for PublicKey {
+    fn hash<H: hash::Hasher>(&self, state: &mut H) {
+        self.0.as_bytes().hash(state);
+    }
+}
```
thomaseizinger: Same story as for ecdsa.

core/src/identity/secp256k1.rs
```diff
@@ -163,6 +165,24 @@ impl fmt::Debug for PublicKey {
     }
 }
 
+impl hash::Hash for PublicKey {
+    fn hash<H: hash::Hasher>(&self, state: &mut H) {
+        self.encode().hash(state);
+    }
+}
```
thomaseizinger, line 178: Same as for ecdsa.

thomaseizinger (commented):
Thank you for working on this!

I've left a few comments :)

core/src/identity/ecdsa.rs
```diff
@@ -249,4 +268,13 @@ mod tests {
         let invalid_msg = "h3ll0 w0rld".as_bytes();
         assert!(!pk.verify(invalid_msg, &sig));
     }
+
+    #[test]
+    fn ecdsa_public_key_eq_implies_hash() {
+        fn prop(SomePublicKey(pub1): SomePublicKey, SomePublicKey(pub2): SomePublicKey) -> bool {
+            pub1 != pub2 || hash(&pub1) == hash(&pub2)
```
thomaseizinger: Is this property correct?

Having two equal pub keys will short-circuit this I believe and we never compare the hash but that is exactly the property we want to check.

I just tried the following and with release mode, that takes about 15 seconds to run. I wasn't patient enough to run it in Debug mode. With 10_000 tests, this fails because it can't generate enough tests to satisfy the `min_tests_passed` property (which defaults to 0).

```rust
    #[test]
    fn ecdsa_public_key_eq_implies_hash() {
        fn prop(SomePublicKey(pub1): SomePublicKey, SomePublicKey(pub2): SomePublicKey) -> TestResult {
            if pub1 != pub2 {
                return TestResult::discard()
            }

            if hash(&pub1) == hash(&pub2) {
                return TestResult::passed()
            } else {
                return TestResult::failed()
            }
        }

        QuickCheck::new()
			.tests(100_000)
            .min_tests_passed(1)
            .quickcheck(prop as fn(_, _) -> _);
    }
```

Thanks for implementing this but I am no longer sure that this is actually the best way to check this. Sorry for leading you down this path.

What if we instead just remove the derived `PartialEq` and implement it manually ourselves by comparing the bytes? That would remove the lint warning, the need for the test and gives us peace of mind that this property actually holds.

thomaseizinger (changes requested):
Thanks for following up with this! Please see inline comment.

core/src/identity/ecdsa.rs
futpib: The property is correct for sure, exactly what clippy and rust docs show https://doc.rust-lang.org/std/hash/trait.Hash.html#hash-and-eq.

No surprise that it's hard to generate 100000 equal keys. If you want to generate and cover equal pairs as thoroughly as unequal pairs, I would tweak the arbitrary implementation (or just write another test that generates one key and clones it). I just did not think it was the goal.

thomaseizinger: > The property is correct for sure, exactly what clippy and rust docs show https://doc.rust-lang.org/std/hash/trait.Hash.html#hash-and-eq.

I thought it through again and it does hold but I have found it hard to understand because it is not expressed the same way as the docs say it :)

The whole point of the property is to check if equal keys have equal hashes. Non equal keys are not interesting so we can discard those test inputs.

Cloning one key and comparing `PartialEq` and `Hash` is not useful by itself. What the lint wants to guard against is that the `PartialEq` impl f.e. uses different fields from the `Hash` impl, thus two differently generated keys could violate the property. Unless you have a non-deterministic Hash impl, the hash of the same key will always be the same so cloning the key doesn't necessarily prove that we are not violating the property.

Like I said above, I don't think a proptest is the right tool here, sorry for leading you down that road earlier.

I'd like to use implement `PartialEq` manually. That will silence the lint and we can see from the impl that we are using the internal byte representation for both.

core/src/identity/rsa.rs
```diff
@@ -307,28 +307,30 @@ mod tests {
...
 
+    #[derive(Clone, Debug)]
+    struct SomePublicKey(PublicKey);
+
+    impl Arbitrary for SomePublicKey {
+        fn arbitrary<G: Gen>(g: &mut G) -> SomePublicKey {
+            SomePublicKey(SomeKeypair::arbitrary(g).0.public())
+        }
+    }
+
```
thomaseizinger:
```suggestion
```

I think we are no longer using this now, correct?

thomaseizinger (approved):
Thank you! :)

I've triggered CI. One last comment, otherwise this should be good to go in!

mxinden (approved):
Thanks @futpib for the patch and @thomaseizinger for the guidance.

mxinden:
Interoperability test failures due to https://github.com/libp2p/test-plans/pull/41. Merging here.

core/CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+# 0.36.1 - [unreleased]
```
mxinden, line 1: Oh, missed this one. @futpib can you please:

1. Update the version in `core/Cargo.toml` to `0.36.1`
2. Update `libp2p-core` to `0.36.1` in the root `Cargo.toml`
3. And reference the bump in the root `Changelog`.

See also https://github.com/libp2p/rust-libp2p/blob/master/docs/release.md#development-between-releases

mxinden (commented):
Oh, missed one thing.

thomaseizinger:
Fixed the merge conflicts :)
Let's get this in once CI passes!

mxinden:
Interoperability tests failing due to https://github.com/libp2p/test-plans/pull/41. Merging here. Thanks @futpib!

PR #2917, opened 2022-09-19
Slimmified CI 

umgefahren:
# Description

I removed some steps from the CI to improve the peformance. This is possible, because GitHub Actions runners now come with Rust support.

Origin: #2900 

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,7 +20,7 @@ jobs:
     steps:
 
     - name: Cancel Previous Runs
-      uses: styfle/cancel-workflow-action@bb6001c4ea612bf59c3abfc4756fbceee4f870c7 # 0.10.0
+      uses: styfle/cancel-workflow-action@0.10.0 # 0.10.0
       with:
         access_token: ${{ github.token }}
 
@@ -55,7 +55,7 @@ jobs:
     steps:
 
     - name: Cancel Previous Runs
-      uses: styfle/cancel-workflow-action@bb6001c4ea612bf59c3abfc4756fbceee4f870c7 # 0.10.0
+      uses: styfle/cancel-workflow-action@0.10.0 # 0.10.0
       with:
         access_token: ${{ github.token }}
 
@@ -83,7 +83,7 @@ jobs:
     steps:
 
     - name: Cancel Previous Runs
-      uses: styfle/cancel-workflow-action@bb6001c4ea612bf59c3abfc4756fbceee4f870c7 # 0.10.0
+      uses: styfle/cancel-workflow-action@0.10.0 # 0.10.0
       with:
         access_token: ${{ github.token }}
 
@@ -92,7 +92,7 @@ jobs:
 
     - uses: actions/checkout@v3
 
-    - uses: Swatinem/rust-cache@6720f05bc48b77f96918929a9019fb2203ff71f8 # v2.0.0
+    - uses: Swatinem/rust-cache@v2 # v2.0.0
 
     - name: Check rustdoc links
       run: RUSTDOCFLAGS="--deny broken_intra_doc_links" cargo doc --verbose --workspace --no-deps --document-private-items --all-features
@@ -102,7 +102,7 @@ jobs:
     steps:
 
     - name: Cancel Previous Runs
-      uses: styfle/cancel-workflow-action@bb6001c4ea612bf59c3abfc4756fbceee4f870c7 # 0.10.0
+      uses: styfle/cancel-workflow-action@0.10.0 # 0.10.0
       with:
         access_token: ${{ github.token }}
 
@@ -111,7 +111,7 @@ jobs:
 
     - uses: actions/checkout@v3
 
-    - uses: Swatinem/rust-cache@6720f05bc48b77f96918929a9019fb2203ff71f8 # v2.0.0
+    - uses: Swatinem/rust-cache@v2 # v2.0.0
 
     - name: Run clippy
       run: cargo clippy
@@ -122,7 +122,7 @@ jobs:
     steps:
 
     - name: Cancel Previous Runs
-      uses: styfle/cancel-workflow-action@bb6001c4ea612bf59c3abfc4756fbceee4f870c7 # 0.10.0
+      uses: styfle/cancel-workflow-action@0.10.0 # 0.10.0
       with:
         access_token: ${{ github.token }}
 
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -127,19 +111,10 @@ jobs:
...
-        toolchain: stable
-        override: true
-        components: clippy
-
     - uses: Swatinem/rust-cache@6720f05bc48b77f96918929a9019fb2203ff71f8 # v2.0.0
 
-    - name: Run cargo clippy
-      uses: actions-rs/cargo@844f36862e911db73fe0815f00a4a2602c279505 # v1.0.3
-      with:
-        command: custom-clippy # cargo alias to allow reuse of config locally
```
thomaseizinger, line 142: We are running an alias here, please retain that (with the comment!).

.github/workflows/ci.yml
```diff
@@ -156,12 +131,6 @@ jobs:
 
     - uses: actions/checkout@v3
 
-    - uses: actions-rs/toolchain@16499b5e05bf2e26879000db0c1d13f7e13fa3af # v1.0.7
-      with:
-        profile: minimal
-        toolchain: stable
-        override: true
-
     - uses: Swatinem/rust-cache@6720f05bc48b77f96918929a9019fb2203ff71f8 # v2.0.0
```
thomaseizinger: Can we bump to this v2 as well then?

.github/workflows/ci.yml
```diff
@@ -127,19 +111,10 @@ jobs:
...
     - uses: actions/checkout@v3
 
-    - uses: actions-rs/toolchain@16499b5e05bf2e26879000db0c1d13f7e13fa3af # v1.0.7
-      with:
-        profile: minimal
-        toolchain: stable
-        override: true
-        components: clippy
-
     - uses: Swatinem/rust-cache@6720f05bc48b77f96918929a9019fb2203ff71f8 # v2.0.0
```
thomaseizinger: Here too!

.github/workflows/ci.yml
```diff
@@ -102,12 +92,6 @@ jobs:
 
     - uses: actions/checkout@v3
 
-    - uses: actions-rs/toolchain@16499b5e05bf2e26879000db0c1d13f7e13fa3af # v1.0.7
-      with:
-        profile: minimal
-        toolchain: stable
-        override: true
-
     - uses: Swatinem/rust-cache@6720f05bc48b77f96918929a9019fb2203ff71f8 # v2.0.0
```
thomaseizinger: And here.

thomaseizinger (commented):
Thanks! I am in favor of this.

I don't know why we are referencing actions by hash instead of version. @mxinden Do you know?

mxinden:
> I don't know why we are referencing actions by hash instead of version. @mxinden Do you know?

Reasoning is here:

> I pinned the third-party action versions to the exact commit SHAs. We follow the same strategy in protocol go repositories to ensure the code we use doesn't change.

https://github.com/libp2p/rust-libp2p/pull/2424

I don't have a strong opinion here. Maybe @galargh does?

galargh:
Yes, exactly! Pinning actions by hash is a mitigation for a potential supply chain attack. We try to do that for all actions that are neither from PL orgs nor https://github.com/actions.

thomaseizinger:
Okay, thanks for clarifying! Let's go back to commit hashes then please @umgefahren!

Plus, https://github.com/libp2p/rust-libp2p/pull/2917#discussion_r974884841 needs to be addressed before we can merge this!

.github/workflows/ci.yml
umgefahren: But why are you running an alias here? I don't quite understand the assignment.

thomaseizinger: The alias is a workaround for clippy still not having a config file.

It is defined here: https://github.com/libp2p/rust-libp2p/blob/master/.cargo/config.toml#L3

This allows us to allow certain lints across the codebase without having to duplicate this in a bunch of places :)

.github/workflows/ci.yml
```diff
@@ -20,16 +20,16 @@ jobs:
...
         access_token: ${{ github.token }}
 
     - name: Install Protoc
-      uses: arduino/setup-protoc@v1
+      uses: arduino/setup-protoc@31a047f6ff1b725b36c561770f81012e2f165926
 
     - uses: actions/checkout@v3
 
-    - uses: Swatinem/rust-cache@6720f05bc48b77f96918929a9019fb2203ff71f8 # v2.0.0
+    - uses: Swatinem/rust-cache@76686c56f2b581d1bb5bda44b51f7e24bd9b8b8e
```
thomaseizinger, line 32: As far as I can see, this hash is not yet part of a released version. I'd prefer if we just not touch this line at all and revert back to exactly what we had :)

.github/workflows/ci.yml
```diff
@@ -20,16 +20,16 @@ jobs:
     steps:
 
     - name: Cancel Previous Runs
-      uses: styfle/cancel-workflow-action@bb6001c4ea612bf59c3abfc4756fbceee4f870c7 # 0.10.0
+      uses: styfle/cancel-workflow-action@0.10.0 # 0.10.0
       with:
         access_token: ${{ github.token }}
 
     - name: Install Protoc
-      uses: arduino/setup-protoc@v1
```
thomaseizinger, line 28: It is my understanding that [`v1`](https://github.com/arduino/setup-protoc/tree/v1) would actually correspond to https://github.com/arduino/setup-protoc/commit/64c0c85d18e984422218383b81c52f8b077404d3.

.github/workflows/ci.yml
```diff
@@ -20,16 +20,16 @@ jobs:
     steps:
 
     - name: Cancel Previous Runs
-      uses: styfle/cancel-workflow-action@bb6001c4ea612bf59c3abfc4756fbceee4f870c7 # 0.10.0
```
thomaseizinger, line 23: Let's revert this too please.

thomaseizinger (changes requested):
Thank you!

I think the original idea of the PR - to get rid of `actions-rs/toolchain` is good. To stay focused on that, can we revert the other changes to what they were? I think the old commit hashes been set exactly to the hash of the releases before and now we are pointing to some commit on master.

umgefahren:
I will take another go at this later. Closing for now.

Issue #2902, opened 2022-09-14
*: Automate various parts of the release process


thomaseizinger:
## Description

Most tasks around releasing crates in our monorepo are currently performed manually:

- Adding changelog entries (happens in PRs usually)
- Pull through version numbers of workspace dependencies (also happens in PRs)
- Actually releasing crates to crates.io (done by @mxinden)

## Motivation

Repetitive tasks consume unnecessary time and should be automated.

## Requirements

1. I'd like to have the pulling through of version bumps somehow automated, i.e. if I bump `libp2p-core` by a minor version, all dependent crates should update their dependency automatically and bump their on version as required.
2. Adding a changelog entry about the above would be useful.
3. Automatically lint against broken links in markdown. This is currently not possible because the root `CHANGELOG` links to what the header _will be_ after the release, meaning until then, they are all broken because the contain the `-unreleased` postfix.

It would be nice to trigger these operations through a bot  / CI so contributors don't have to learn how to do it.

## Are you planning to do it yourself in a pull request?

Maybe.

thomaseizinger:
I would suggest we make use of the `cargo-xtask` pattern: https://github.com/matklad/cargo-xtask/

BigLep:
It's great to see this issue.  

I think the first step here is to document today what we do.  I don't best (automation) to be the enemy of better (document what we do).  Basically I view the first step is "repeatability".  Then we can move to automating more and more of it.  We need to make sure this isn't trible knowledge that only lives in a limited number of peoples' heads.

@mxinden : in https://github.com/libp2p/test-plans/pull/41#issuecomment-1249508288, you raised the question on the priority of this.  I'd like to prevent this from being an "either or" conversation.  I think it should be a "both and".  Something is better than nothing on this front.  

A couple of options for moving forward:
1. the next time you do a release, have someone else shepherd you the whole time.  Maybe even do a screen recording.  It's their responsibility to then write up the steps if you can't do it.  The next time you reverse shadow to see where the gaps are.
2. (better I think) You take first stab at writing the steps.  For the next release, you have someone else attempt to follow them.  You are shadowing the process in case they get stuck.  The release is then responsible for fixing any issues in the steps that see.  

I think option 2 is better because it gets us validating the correctness/completeness of the release process quicker, but I defer to the team.

Other ideas:
1. @p-shahi may be able to help, but I think it's also great if rust-libp2p maintainers move the ball forward independently as well.  
[...]

PR #2918, opened 2022-09-20
*: Remove default features from all crates

thomaseizinger:
# Description

This patch-set removes our use of `default` features. We also introduce a new `full` feature that activates all features.

- https://github.com/libp2p/rust-libp2p/pull/2173
- https://github.com/libp2p/rust-libp2p/pull/2146

## Change checklist


CHANGELOG.md
```diff
@@ -45,6 +45,9 @@
 
 # 0.49.0 - [unreleased]
 
+- Remove default features. You need to enable all features explicitly now. As a quick workaround, you may want to use the
```
```suggestion
- Remove default features. You need to enable required features explicitly now. As a quick workaround, you may want to use the
```

I think otherwise folks would think that they just need to enable **all** features manually from now on.

core/Cargo.toml
```diff
@@ -46,7 +46,7 @@ ring = { version = "0.16.9", features = ["alloc", "std"], default-features = fal
 async-std = { version = "1.6.2", features = ["attributes"] }
 base64 = "0.13.0"
 criterion = "0.4"
-libp2p = { path = "../", default-features = false, features = ["mplex", "noise"] }
+libp2p = { path = "..", features = ["full"] }
```
mxinden, line 49: Say I want to run the `libp2p-core` unit tests locally, wouldn't it be nicer to only compile the `libp2p` features that I need? Is this use-case worth the additional effort?

mxinden (commented):
Thanks for tackling this.

As mentioned elsewhere, I am in favor of this change.

Given that this is a large breaking change for users, is there anyone objecting to this change?

core/Cargo.toml
thomaseizinger: It is a pain though to keep track of the features you need to activate :/

Plus, with all crates requesting different features, I actually think we have to compile things multiple times to satisfy each combination. If the convention is to enable `full` everywhere, we only need to compile every crate once to and the compiled artifact can be used for every crate's tests[^1].

I think the convenience of just saying "full" is worth it. Locally you have caches of everything so you will most likely just pay in linker time and if you modify your code between test executions, you gotta relink the test binary anyway so I am not sure it will matter in practice.

Can we try it out and if it negatively affects anything, figure out how to make it better?

[^1]: This is an unproven hypothesis.

thomaseizinger:
> Given that this is a large breaking change for users, is there anyone objecting to this change?

Is it that large a change? Activate a couple of feature flags and your back in. No functionality is changing!

core/Cargo.toml
mxinden: Thanks for elaborating. Using `full` sounds good to me.

mxinden:
> > Given that this is a large breaking change for users, is there anyone objecting to this change?
> 
> Is it that large a change? Activate a couple of feature flags and your back in. No functionality is changing!

I remember `tokio` moving to this convention taking me a couple of h. I am not saying it is hard. Just saying it is tedious. Never the less I am in favor of this change.

@thomaseizinger do you want this to be included in `v0.49.0` (https://github.com/libp2p/rust-libp2p/pull/2931)?

mxinden (approved):

thomaseizinger:
> > > Given that this is a large breaking change for users, is there anyone objecting to this change?
> > 
> > 
> > Is it that large a change? Activate a couple of feature flags and your back in. No functionality is changing!
> 
> I remember `tokio` moving to this convention taking me a couple of h. I am not saying it is hard. Just saying it is tedious. Never the less I am in favor of this change.
> 
> @thomaseizinger do you want this to be included in `v0.49.0` (#2931)?

That would be great, yeah!

dignifiedquire:
I would be great to have a short section "How to upgrade" for the release, so folks now how to adjust their `feature` usage.

thomaseizinger:
> I would be great to have a short section "How to upgrade" for the release, so folks now how to adjust their `feature` usage.

Is the first sentence in the changelog entry not good enough? https://github.com/libp2p/rust-libp2p/blob/2173-no-default-features/CHANGELOG.md#0490---unreleased

dignifiedquire:
> Is the first sentence in the changelog entry not good enough?

I somehow missed that when looking through the diff. LGTM 👍

mxinden:
@thomaseizinger would you mind resolving the merge conflicts here? No rush, just a ping.

thomaseizinger:
> @thomaseizinger would you mind resolving the merge conflicts here? No rush, just a ping.

Sure! I just didn't realize there are any :/
I opened https://github.com/libp2p/rust-libp2p/issues/2948.

thomaseizinger:
@mxinden Resolved.

thomaseizinger:
Wooooo! This is so exciting!

umgefahren:
As long as #2950 is still open, this could result in a very confusing experience for a beginner. We should definitely fix that.

core/Cargo.toml
thomaseizinger: I changed my mind. Esp. working on `libp2p-core` is quite painful now because every change there invalidates `libp2p-core` artifact which causes a re-compile of every other crate.

mxinden: :+1: mind proposing a patch @thomaseizinger?

thomaseizinger: So I think the best way to resolve this is to actually stop the circular dependency. This will allow for proper caching of the build artifacts. It will also solve the issue I am currently facing with `release-please` (#2902).

PR #2934, opened 2022-09-23
multistream-select: Remove parallel dialing optimization

### LINKED ISSUE #2925: Remove usage of the `ls` command in multistream-select
## Summary

Currently, if there are more than `3` protocols listed, `multistream-select` uses a parallel approach which relies on the `ls` command. 
This command has been removed from `go-multistream` and so all dials to kubo or other nodes using newer `go-libp2p` will fail in this case.

### References
- https://github.com/ipfs/kubo/issues/8640#-ls--requests-for---multistream-100--are-removed
- https://github.com/libp2p/rust-libp2p/blob/master/misc/multistream-select/src/dialer_select.rs#L319
- https://github.com/multiformats/go-multistream/pull/76

## Possible Solution

Make it either configurable or remove the usage of the `ls` command.
I have a patch that just removes the parallel version, which is the simplest version. I am not sure how it could be avoided to not use `ls`. 

Yes

dignifiedquire:
This is to avoid the usage of the now optional `ls` command, and stay compatible with go-multistream.

Closes #2925

## Change checklist


misc/multistream-select/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# Unreleased
```
thomaseizinger:
```suggestion
# 0.12.0 [Unreleased]
```

misc/multistream-select/src/dialer_select.rs
```diff
@@ -55,44 +49,13 @@ pub fn dialer_select_proto<R, I>(
...
     R: AsyncRead + AsyncWrite,
     I: IntoIterator,
     I::Item: AsRef<[u8]>,
 {
     let protocols = protocols.into_iter().peekable();
-    DialerSelectSeq {
+    DialerSelectFuture {
         version,
         protocols,
         state: SeqState::SendHeader {
```
thomaseizinger: We could also rename `SeqState` to `State` now.

misc/multistream-select/src/tests.rs
```diff
@@ -22,7 +22,6 @@
 
 #![cfg(test)]
```
thomaseizinger: Not quite sure why the tests are in a separate module here. Can we move them to `tests/`?

misc/multistream-select/src/tests.rs
```diff
@@ -182,38 +181,6 @@ fn negotiation_failed() {
...
-        server.await;
-        client.await;
-    }
-
-    async_std::task::block_on(run(Version::V1));
-    async_std::task::block_on(run(Version::V1Lazy));
-}
-
 #[test]
 fn select_proto_serial() {
```
thomaseizinger: Is there a value in having this test over `select_proto_basic`?

thomaseizinger: `select_proto_basic` seems to be a strict super set of this test.

thomaseizinger (approved):
Thanks!

I am okay with this change :)

thomaseizinger:
I think we should also straight-up bump the version in the manifest file and use this new implementation in `libp2p-core`.

misc/multistream-select/CHANGELOG.md
dignifiedquire: fixed

misc/multistream-select/src/dialer_select.rs
dignifiedquire: fixed

misc/multistream-select/src/tests.rs
dignifiedquire: removed

misc/multistream-select/src/tests.rs
dignifiedquire: done

dignifiedquire:
> I think we should also straight-up bump the version in the manifest file and use this new implementation in `libp2p-core`.

done

misc/multistream-select/src/tests.rs
dignifiedquire: They used private methods before, to explicitly execute the two different code paths.

misc/multistream-select/tests/dialer_select.rs
```diff
@@ -22,12 +22,9 @@
 
 #![cfg(test)]
```
thomaseizinger: I think this is obsolete now.

```suggestion
```

thomaseizinger (approved):
LGTM, thank you!

misc/multistream-select/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# 0.12.0 [Unreleased]
+
+- Remove parallel dialing optimization, to avoid requiring the use of the `ls` command.
```
thomaseizinger:
```suggestion
- Remove parallel dialing optimization, to avoid requiring the use of the `ls` command. See [PR 2934].

[PR 2934]: https://github.com/libp2p/rust-libp2p/pull/2934
```

misc/multistream-select/tests/dialer_select.rs
```diff
@@ -20,14 +20,10 @@
 
 //! Integration tests for protocol negotiation.
```
thomaseizinger, line 22:
```suggestion
```

dignifiedquire:
@mxinden anyhing blocking from merging this PR? would be great to have this in

misc/multistream-select/CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+# 0.12.0 [Unreleased]
```
mxinden:
```suggestion
# 0.12.0 [unreleased]
```

mxinden (approved):
Thanks @dignifiedquire for debugging, reporting and fixing this!

mxinden:
Not sure why the interoperability tests are failing. I assume due to:

```
Sep 28 11:02:14.311833	INFO	1.1452s       FAIL << go-latest-v0.22[000] (2e992c) >> failed to dial 12D3KooWBK4ZGwm1kKpPucpiD57hGjvzuDvU4D6eyoLj5EQnNdcm:
  * [/ip4/16.0.0.4/tcp/1234] failed to negotiate security protocol: protocol not supported
```

I will look further into this.

PR #2936, opened 2022-09-23
.github/workflows: Deny cargo doc warnings in CI

### LINKED ISSUE #2914: Fix warnings produced by `cargo doc --workspace`
## Description

Currently, `cargo doc` is emitting a number of warnings. We should fix those and amend the CI configuration to fail as soon as any warnings are emitted.

## Motivation

Warnings are (often) no good.

Maybe.

ddimaria:
# Description

- [x] Update CI to fail on warnings
- [x] Fix all warnings

Fixes https://github.com/libp2p/rust-libp2p/issues/2914

## Open Questions

I'm assuming that it's OK to document private items as it was already specified in the `RUSTDOCFLAGS` `--document-private-items` directive.  Warnings are still generated with this turned on.  To turn off the warnings (in `swarm`), I added `#![allow(rustdoc::private_intra_doc_links)]` to `lib.rs`.

```shell
warning: public documentation for `max_negotiating_inbound_streams` links to private item `PoolConfig::with_max_negotiating_inbound_streams`
    --> swarm/src/lib.rs:1373:15
     |
1373 |     /// See [`PoolConfig::with_max_negotiating_inbound_streams`].
     |               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ this item is private
     |
     = note: this link will resolve properly if you pass `--document-private-items`
```

## Change checklist


swarm/src/lib.rs
```diff
@@ -51,7 +51,7 @@
 //! The [`ConnectionHandler`] trait defines how each active connection to a
 //! remote should behave: how to handle incoming substreams, which protocols
 //! are supported, when to open a new outbound substream, etc.
-//!
+#![allow(rustdoc::private_intra_doc_links)]
```
thomaseizinger: I am not sure if it makes sense to allow this. Do the offending links work on `docs.rs`? Because if not, then there isn't really much of a point I think.

I'd be in favor of rewriting the docs instead.

@mxinden Thoughts?

thomaseizinger (commented):
Thanks for working on this!

One question.

swarm/src/lib.rs
mxinden: > I'd be in favor of rewriting the docs instead.

I would be in favor of this as well.

mxinden (commented):
Thanks @ddimaria for the help here.

swarm/src/lib.rs
ddimaria: The links show up in `docs.rs`, though the links do not work because the `connection` and `connection/pool` modules are private, so completely agree on the doc rewrite (assuming making `connection` and `connection/pool` public isn't a good option).  Links to the offending documentation:

* https://github.com/libp2p/rust-libp2p/blob/master/swarm/src/dial_opts.rs#L124-L129
* https://github.com/libp2p/rust-libp2p/blob/master/swarm/src/dial_opts.rs#L198-L203
* https://github.com/libp2p/rust-libp2p/blob/master/swarm/src/lib.rs#L1371-L1377

I can lift the doc comments from the referenced functions and replace the private links with them, though this is not DRY, so it adds to the doc-update burden.  Thoughts @thomaseizinger ?

thomaseizinger: I'd suggest we simply remove the docs from the non-public items and put them on the public ones. In other words, inline the documentation we are currently referring to and deleting it from there!

.github/workflows/ci.yml
```diff
@@ -111,7 +111,7 @@ jobs:
     - uses: Swatinem/rust-cache@6720f05bc48b77f96918929a9019fb2203ff71f8 # v2.0.0
 
     - name: Check rustdoc links
-      run: RUSTDOCFLAGS="--deny broken_intra_doc_links" cargo doc --verbose --workspace --no-deps --document-private-items --all-features
+      run: RUSTDOCFLAGS="--deny rustdoc::broken_intra_doc_links --deny warnings" cargo doc --verbose --workspace --no-deps --all-features
```
ddimaria, line 114: Removed `--document-private-items` as we don't want references to non-public items.

swarm/src/lib.rs
ddimaria: This is done, though I left the documentation in the private functions alone as it looked out of place with the other items documented.  It may be out of scope for this PR, but it's worth considering changing `///` to `//` documentation in private items.

thomaseizinger (approved):
Awesome, thank you!

ddimaria:
> Awesome, thank you!

Thank you @thomaseizinger! Do you need to approve the workflow (first time contributor)?

> First-time contributors need a maintainer to approve running workflows. [Learn more.](https://docs.github.com/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks) 9 expected checks

PR #2939, opened 2022-09-24
protocols/mdns: Optimise `InterfaceState::poll` for low latency

thomaseizinger:
# Description

This is now in accordance with our coding guidelines. Inspired by @gallegogt's findings.

- https://github.com/libp2p/rust-libp2p/discussions/2916

## Change checklist

### DIFF
--- a/protocols/mdns/CHANGELOG.md
+++ b/protocols/mdns/CHANGELOG.md
@@ -1,3 +1,9 @@
+# 0.40.1 [unreleased]
+
+- Fix a bug that could cause a delay of ~10s until peers would get discovered when using the tokio runtime. See [PR XXXX].
+
+[PR XXXX]: https://github.com/libp2p/rust-libp2p/pull/XXXX
+
 # 0.40.0
 
 - Update to `libp2p-swarm` `v0.39.0`.
--- a/protocols/mdns/CHANGELOG.md
+++ b/protocols/mdns/CHANGELOG.md
@@ -1,8 +1,8 @@
 # 0.40.1 [unreleased]
 
-- Fix a bug that could cause a delay of ~10s until peers would get discovered when using the tokio runtime. See [PR XXXX].
+- Fix a bug that could cause a delay of ~10s until peers would get discovered when using the tokio runtime. See [PR 2939].
 
-[PR XXXX]: https://github.com/libp2p/rust-libp2p/pull/XXXX
+[PR 2939]: https://github.com/libp2p/rust-libp2p/pull/2939
 
 # 0.40.0
 
--- a/protocols/mdns/CHANGELOG.md
+++ b/protocols/mdns/CHANGELOG.md
@@ -1,16 +1,15 @@
 # 0.41.0 [unreleased]
 
-- Fix a bug that could cause a delay of ~10s until peers would get discovered when using the tokio runtime. See [PR 2939].
-
 - Remove default features. If you previously depended on `async-io` you need to enable this explicitly now. See [PR 2918].
 
 - Update to `libp2p-core` `v0.37.0`.
 
 - Update to `libp2p-swarm` `v0.40.0`.
 
+- Fix a bug that could cause a delay of ~10s until peers would get discovered when using the tokio runtime. See [PR 2939].
 
-[PR 2939]: https://github.com/libp2p/rust-libp2p/pull/2939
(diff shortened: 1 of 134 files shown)


protocols/mdns/CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+# 0.40.1 [unreleased]
+
+- Fix a bug that could cause a delay of ~10s until peers would get discovered when using the tokio runtime. See [PR XXXX].
+
+[PR XXXX]: https://github.com/libp2p/rust-libp2p/pull/XXXX
```
```suggestion
- Fix a bug that could cause a delay of ~10s until peers would get discovered when using the tokio runtime. See [PR 2939].

[PR 2939]: https://github.com/libp2p/rust-libp2p/pull/2939
```

protocols/mdns/src/behaviour/iface.rs
```diff
@@ -200,51 +200,58 @@ where
...
-                    {
-                        self.inject_mdns_packet(packet, params);
+        loop {
+            // 1st priority: Low latency: Create packet ASAP after timeout.
+            if Pin::new(&mut self.timeout).poll_next(cx).is_ready() {
+                log::trace!("sending query on iface {}", self.addr);
+                self.send_buffer.push_back(build_query());
+            }
+
+            // 2nd priority: Keep local buffers small: Empty buffers ASAP.
```
mxinden: Shouldn't we first try to send existing work (`self.send_buffer.pop_front`) and then add more work (`self.send_buffer.push_back`)? In case we do a `push_back`, we can do a `continue` to follow with a `pop_front`.

protocols/mdns/src/behaviour/iface.rs
```diff
@@ -200,51 +200,58 @@ where
...
+                        log::error!("failed reading datagram: {}", err);
+                    }
                 }
             }
-        }
 
-        if Pin::new(&mut self.timeout).poll_next(cx).is_ready() {
-            log::trace!("sending query on iface {}", self.addr);
-            self.send_buffer.push_back(build_query());
+            // 4th priority: Emit event only if nothing else can make progress.
```
mxinden: Would it not make more sense to do (4) before (3)? I.e. return discovered addresses before checking whether there are other to discover? That would keep the buffer `self.discovered` small.

mxinden (commented):
Thanks for looking into this.

I don't yet understand what the issue was. Was `self.timeout` not polled early on and thus no waker registered?

protocols/mdns/src/behaviour/iface.rs
```diff
@@ -200,51 +200,58 @@ where
         cx: &mut Context,
         params: &impl PollParameters,
     ) -> Option<(PeerId, Multiaddr, Instant)> {
```
mxinden: Not directly related. Why do we return an `Option` and not a `Poll`?

thomaseizinger:
> I don't yet understand what the issue was. Was `self.timeout` not polled early on and thus no waker registered?

The issue was that we pushed into `send_buffer` but then returned from the function, despite having just created more work that we could have done (i.e. send the item we just pushed in the buffer) :)

protocols/mdns/src/behaviour/iface.rs
thomaseizinger: I don't know why.

protocols/mdns/src/behaviour/iface.rs
thomaseizinger: I am not sure if it makes a big difference. The timer is our only local work, i.e. nothing else will push to the buffer. When we say "prefer local work over remote work", does that not imply that whatever generates the local work has the highest priority, followed by trying to send it to the remote?

protocols/mdns/src/behaviour/iface.rs
thomaseizinger: Yeah that would make sense I think!

protocols/mdns/src/behaviour/iface.rs
thomaseizinger: Fixed it!

thomaseizinger:
@mxinden I addressed your comments and also went on a bit of a refactoring detour because I found some of the code hard to understand. I hope it is better now :blush: 

You should be able to review patch-by-patch!

mxinden (approved):
Thank you @thomaseizinger.

mxinden:
> You should be able to review patch-by-patch!

That helped. Thanks.

PR #1175, opened 2019-06-13
[WIP] TLS transport upgrade

g-r-a-n-t:
Addressing #1109

This is nowhere near completion. At the moment, it is just an empty upgrade on a given transport.

PR #2945, opened 2022-09-26
transports/tls: Add `libp2p-tls` as per spec

### LINKED ISSUE #1109: Add TLS Handshake
The TLS handshake as described here https://github.com/libp2p/specs/blob/1a5e6eb805fc6c902128288ae1575945bcdc6e2a/tls/design%20considerations.md should be implemented.

thomaseizinger:
## Description

This code is mostly extracted from #2289. I've tagged everyone who made contributions to that PR as co-authors (cc @kpp @dvc94ch @elenaf9 @mxinden) in the first commit.

On top of that, this also adds an implementation of the `Upgrade{Inbound,Outbound}` traits so that TLS can be used with other transports. Finally, this PR also makes an effort towards https://github.com/libp2p/specs/issues/459 by including a few static tests (cc @marten-seemann).

- https://github.com/libp2p/specs/issues/459
- https://github.com/libp2p/rust-libp2p/pull/2289
- https://github.com/libp2p/rust-libp2p/pull/1175
- https://github.com/ChainSafe/forest/issues/974 (cc @austinabell)
- Fixes https://github.com/libp2p/rust-libp2p/issues/1109.

## Open tasks

- [ ] Reference https://github.com/DemiMarie as co-author: `Co-authored-by:  Demi Marie Obenour <demiobenour@gmail.com>`
- [ ] Reference https://github.com/tomaka as co-author: `Co-authored-by: Pierre Krieger <pierre.krieger1708@gmail.com>`

## Change checklist


thomaseizinger:
Unfortunately, it seems `ring` is not WASI friendly https://github.com/libp2p/rust-libp2p/actions/runs/3125423794/jobs/5069777479#step:9:36.

kpp:
It seems like a nice idea but you erased the authorship of the code (I mean `git blame`).

thomaseizinger:
> It seems like a nice idea but you erased the authorship of the code (I mean `git blame`).

The first commit contains `Co-authored-by` entries for all people who contributed. I believe GitHub will carry those over once we squash-merge so the authorship should be retained. I think that should be transparent enough? I am open to other ideas :)

thomaseizinger:
I could take all existing commits and put another one on top that deletes all the QUIC code. Given that we squash-merge, I am not sure it makes a big difference?

kpp:
Can you also apply this patch https://github.com/kpp/rust-libp2p/pull/20? It's a little bit outdated, I was waiting for the QUIC branch to adapt the changes from master.

thomaseizinger:
> Can you also apply this patch [kpp#20](https://github.com/kpp/rust-libp2p/pull/20)? It's a little bit outdated, I was waiting for the QUIC branch to adapt the changes from master.

Done!

transports/tls/src/certificate.rs
```diff
@@ -0,0 +1,547 @@
...
+
+#[derive(Debug, thiserror::Error)]
+#[error(transparent)]
+pub struct VerificationError(#[from] pub(crate) webpki::Error);
+
+/// Attempts to parse the provided bytes as a [`P2pCertificate`].
+///
+/// For this to succeed, the certificate must contain the specified extension and the signature must
+/// match the embedded public key.
+pub fn parse(der_input: &[u8]) -> Result<P2pCertificate, ParseError> {
```
kpp: parse_verify?

transports/tls/src/verifier.rs
```diff
@@ -0,0 +1,217 @@
...
+            SignatureScheme::RSA_PSS_SHA256,
+            SignatureScheme::RSA_PKCS1_SHA512,
+            SignatureScheme::RSA_PKCS1_SHA384,
+            SignatureScheme::RSA_PKCS1_SHA256,
+        ]
+    }
+}
+
+impl ServerCertVerifier for Libp2pCertificateVerifier {
+    fn verify_server_cert(
```
kpp, line 105: So we can dial with a PK and without. If we dial with a PK then we should get a cert.extension.public_key and check whether they are equal. This code was mistakenly removed in 3b1ff348999da1543.

cc @marten-seemann

kpp: From specs:

```
The public host key allows the peer to calculate the peer ID of the peer it is connecting to. Clients MUST verify that the peer ID derived from the certificate matches the peer ID they intended to connect to, and MUST abort the connection if there is a mismatch.
```

transports/tls/Cargo.toml
```diff
@@ -0,0 +1,23 @@
+[package]
```
kpp, line 1: Should exclude publishing test_assets

transports/tls/src/certificate.rs
```diff
@@ -43,9 +43,7 @@ static P2P_SIGNATURE_ALGORITHM: &rcgen::SignatureAlgorithm = &rcgen::PKCS_ECDSA_
 
 /// Generates a self-signed TLS certificate that includes a libp2p-specific
 /// certificate extension containing the public key of the given keypair.
-pub fn generate(
-    identity_keypair: &identity::Keypair,
-) -> Result<rcgen::Certificate, rcgen::RcgenError> {
+pub fn generate(identity_keypair: &identity::Keypair) -> Result<rcgen::Certificate, GenError> {
```
kpp: Still `rcgen::Certificate` leaks?

transports/tls/tests/smoke.rs
```diff
@@ -0,0 +1,73 @@
...
+            match swarm1.next().await.unwrap() {
+                SwarmEvent::NewListenAddr {
+                    address,
+                    listener_id,
+                } if listener_id == expected_listener_id => break address,
+                _ => continue,
+            };
+        }
+    };
+    swarm2.dial(listen_address).unwrap();
```
kpp, line 27: This one dials something like `/memory/2698462084451899396` (without PK). It would be nice to have two more tests: with a correct explicit PK and with incorrect PK.

transports/tls/src/verifier.rs
marten-seemann: > So we can dial with a PK and without. If we dial with a PK then we should get a cert.extension.public_key and check whether they are equal. This code was mistakenly removed in https://github.com/libp2p/rust-libp2p/commit/3b1ff348999da1543b7c394f4db9f7364e061508.
>
> cc @marten-seemann

Not sure I understand the question, can you please clarify? PK = public key? private key?
The libp2p extension needs to be present in the certificate in any case.

kpp: PK = public key.

> The libp2p extension needs to be present in the certificate in any case.

Yes but if we dial with a known PK we MUST verify that the PK in the extension equals to the PK we are dialing to. Is it correct?

marten-seemann: That's correct. Although I'd expect the more common case to be that you know the Peer ID in advance, not the public key (for peer IDs where the key is not inlined).

transports/tls/tests/smoke.rs
thomaseizinger: Verifying whether we dialed the correct peer in case the multiaddr has a peer ID inlined happens on the `Swarm` level in `rust-libp2p` and is thus transport independent. I don't think we need to add a test for this!

transports/tls/src/certificate.rs
thomaseizinger: You are right, well spotted!

We can make this one `pub(crate)` though because it is only used by internal APIs.

transports/tls/Cargo.toml
thomaseizinger: Yep, good catch, thank you!

transports/tls/src/verifier.rs
thomaseizinger: Verification that we connected to the intended peer (`/p2p`) happens in the `Swarm`, independent of the given `Transport` in `rust-libp2p`. I think we are covering this requirement for any `Transport` as long as people are using the `Swarm` abstraction[^1].


[^1]: Most users are as that is what a lot of functionality is built on top of but in theory, you could use it without `libp2p-swarm`. I know of one project that does and there, the check is implemented manually.

transports/tls/src/certificate.rs
thomaseizinger: I purposely I named this one `parse` and the other one `parse_unverified` because I think verification / validation should be part of parsing to make good use of the type system. Also see [this blogpost](https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/).

transports/tls/src/certificate.rs
thomaseizinger: Resolved in [`e1f34ae` (#2945)](https://github.com/libp2p/rust-libp2p/pull/2945/commits/e1f34aeafb9a543465e52d6bfffdfeb620cd9843).

transports/tls/tests/smoke.rs
marten-seemann: That works, but it’s not a very nice behavior. Certificates should be verified during the handshake, as this allows you to abort the connection attempt during TLS handshake.
Verifying the peer ID after the handshake leads to the successful establishment of a connection on the server side, just to have that connection killed 1 RTT later. Even worse, while aborting during the handshake lets the peer know why the handshake failed (as this is communicated by a TLS alert), aborting after handshake completion leaves the peer guessing what went wrong.

thomaseizinger: Yeah that makes sense. It is a historical decision that would require a fair bit of work to change. I'll open an issue about it but would prefer to not block this PR on it.

kpp: > but would prefer to not block this PR on it.

And we will merge master and use a not full implementation of the tls spec in 2 PRs. I can do that if you like.

kpp: https://github.com/libp2p/rust-libp2p/pull/2947

transports/tls/src/certificate.rs
```diff
@@ -0,0 +1,552 @@
...
+pub struct ParseError(#[from] pub(crate) webpki::Error);
+
+#[derive(Debug, thiserror::Error)]
+#[error(transparent)]
+pub struct VerificationError(#[from] pub(crate) webpki::Error);
+
+/// Internal function that only parses but does not verify the certificate.
+///
+/// Useful for testing but unsuitable for production.
+fn parse_unverified(der_input: &[u8]) -> Result<P2pCertificate, webpki::Error> {
```
mxinden, line 122:
```suggestion
#[cfg(test)]
fn parse_unverified(der_input: &[u8]) -> Result<P2pCertificate, webpki::Error> {
```

Not an optimization, but rather a safety measure to prevent accidental use. What do you think?

transports/tls/tests/smoke.rs
```diff
@@ -0,0 +1,73 @@
+use futures::{future, StreamExt};
```
mxinden, line 1: Please add a license header here :innocent: and in all other files.

mxinden (commented):
Wonderful to see this happening. Thanks @thomaseizinger and everyone involved.

I have not yet taken a deeper look.

@thomaseizinger have you been able to test this against a go-libp2p node? E.g. against one of `/dnsaddr/bootstrap.libp2p.io/`, e.g. via https://github.com/mxinden/libp2p-lookup?

transports/tls/src/certificate.rs
thomaseizinger: Can't do that because it is used by the `parse` function to avoid code duplication.

It is private so I don't think we need to be worried about accidential use.

thomaseizinger:
> Wonderful to see this happening. Thanks @thomaseizinger and everyone involved.
> 
> I have not yet taken a deeper look.
> 
> @thomaseizinger have you been able to test this against a go-libp2p node? E.g. against one of `/dnsaddr/bootstrap.libp2p.io/`, e.g. via https://github.com/mxinden/libp2p-lookup?

I have not yet tested it against the go-impl but can do :)

thomaseizinger:
> > Wonderful to see this happening. Thanks @thomaseizinger and everyone involved.
> > I have not yet taken a deeper look.
> > @thomaseizinger have you been able to test this against a go-libp2p node? E.g. against one of `/dnsaddr/bootstrap.libp2p.io/`, e.g. via [mxinden/libp2p-lookup](https://github.com/mxinden/libp2p-lookup)?
> 
> I have not yet tested it against the go-impl but can do :)

So that doesn't seem to work at the moment :sweat_smile: 

```
❯ cargo run --release -- direct --address /dnsaddr/bootstrap.libp2p.io/p2p/QmQCU2EcMqAqQPR2i9bChDtGNJchTbq5TbXJJ16u19uLTa
    Finished release [optimized] target(s) in 0.17s
     Running `target/release/libp2p-lookup direct --address /dnsaddr/bootstrap.libp2p.io/p2p/QmQCU2EcMqAqQPR2i9bChDtGNJchTbq5TbXJJ16u19uLTa`
[2022-09-28T22:52:35Z ERROR libp2p_lookup] Lookup failed: FailedToDialPeer { error: Transport([("/dnsaddr/bootstrap.libp2p.io/p2p/QmQCU2EcMqAqQPR2i9bChDtGNJchTbq5TbXJJ16u19uLTa", Other(Custom { kind: Other, error: Custom { kind: Other, error: Other(A(B(Apply(ClientUpgrade(Custom { kind: InvalidData, error: InvalidCertificateData("invalid peer certificate: UnknownIssuer") }))))) } }))]) }.
```

But noise also doesn't work :thinking: 

```
❯ cargo run --release -- direct --address /dnsaddr/bootstrap.libp2p.io/p2p/QmQCU2EcMqAqQPR2i9bChDtGNJchTbq5TbXJJ16u19uLTa
    Finished release [optimized] target(s) in 1.25s
     Running `target/release/libp2p-lookup direct --address /dnsaddr/bootstrap.libp2p.io/p2p/QmQCU2EcMqAqQPR2i9bChDtGNJchTbq5TbXJJ16u19uLTa`
[2022-09-28T22:53:31Z ERROR libp2p_lookup] Lookup failed: FailedToDialPeer { error: Transport([("/dnsaddr/bootstrap.libp2p.io/p2p/QmQCU2EcMqAqQPR2i9bChDtGNJchTbq5TbXJJ16u19uLTa", Other(Custom { kind: Other, error: Custom { kind: Other, error: Other(A(B(Apply(InvalidKey)))) } }))]) }.
```

kpp:
> have you been able to test this against a go-libp2p node?

The crypto should work because I tested it against go-libp2p-quic.

thomaseizinger:
> > have you been able to test this against a go-libp2p node?
> 
> The crypto should work because I tested it against go-libp2p-quic.

Your implementation or does that include my refactorings already? :)

kpp:
Haha nice catch. Let me check it with your implementation.

mxinden:
> But noise also doesn't work thinking
> 
> ```
> ❯ cargo run --release -- direct --address /dnsaddr/bootstrap.libp2p.io/p2p/QmQCU2EcMqAqQPR2i9bChDtGNJchTbq5TbXJJ16u19uLTa
>     Finished release [optimized] target(s) in 1.25s
>      Running `target/release/libp2p-lookup direct --address /dnsaddr/bootstrap.libp2p.io/p2p/QmQCU2EcMqAqQPR2i9bChDtGNJchTbq5TbXJJ16u19uLTa`
> [2022-09-28T22:53:31Z ERROR libp2p_lookup] Lookup failed: FailedToDialPeer { error: Transport([("/dnsaddr/bootstrap.libp2p.io/p2p/QmQCU2EcMqAqQPR2i9bChDtGNJchTbq5TbXJJ16u19uLTa", Other(Custom { kind: Other, error: Custom { kind: Other, error: Other(A(B(Apply(InvalidKey)))) } }))]) }.
> ```

Have you enabled the `rsa` feature @thomaseizinger?

kpp:
It works on my side (https://github.com/kpp/rust-libp2p/pull/24). I was able to reach go nodes via QUIC with your branch.

thomaseizinger:
> > But noise also doesn't work thinking
> > ```
> > ❯ cargo run --release -- direct --address /dnsaddr/bootstrap.libp2p.io/p2p/QmQCU2EcMqAqQPR2i9bChDtGNJchTbq5TbXJJ16u19uLTa
> >     Finished release [optimized] target(s) in 1.25s
> >      Running `target/release/libp2p-lookup direct --address /dnsaddr/bootstrap.libp2p.io/p2p/QmQCU2EcMqAqQPR2i9bChDtGNJchTbq5TbXJJ16u19uLTa`
> > [2022-09-28T22:53:31Z ERROR libp2p_lookup] Lookup failed: FailedToDialPeer { error: Transport([("/dnsaddr/bootstrap.libp2p.io/p2p/QmQCU2EcMqAqQPR2i9bChDtGNJchTbq5TbXJJ16u19uLTa", Other(Custom { kind: Other, error: Custom { kind: Other, error: Other(A(B(Apply(InvalidKey)))) } }))]) }.
> > ```
> 
> Have you enabled the `rsa` feature @thomaseizinger?

I haven't, just used the code from master branch as is :)
Adding the feature did the trick, which makes sense when looking at the key format. I just didn't think of it ...

thomaseizinger:
Lookup also succeeded with TLS authentication! Exciting! :partying_face:

thomaseizinger:
This is ready for a final review :)

transports/tls/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+license = "MIT"
+exclude = ["src/test_assets"]
+
+[dependencies]
+futures = { version = "0.3.24", default-features = false }
+futures-rustls = "0.22.2"
+libp2p-core = { version = "0.37.0", path = "../../core" }
+rcgen = "0.9.2"
+ring = "0.16.20"
+rustls = { version = "0.20.2", default-features = false, features = ["dangerous_configuration"] } # Exposed dependency! Bump major version on any change.
```
mxinden: Can you document why it is safe here to enable the `dangerous_configuration` feature?

mxinden (commented):
This looks good to me. Thanks @thomaseizinger for the follow-ups. Thanks @kpp for https://github.com/libp2p/rust-libp2p/pull/2945/commits/79a464a780e67f04f952e90440cfd6e69a9b184e.

Given that the logic here is critical to get right, I would appreciate additional eyes. @marten-seemann do you have some time to give this a review? Happy to help with any Rust related questions.

transports/tls/src/verifier.rs
```diff
@@ -0,0 +1,242 @@
...
+    fn verify_tls12_signature(
+        &self,
+        _message: &[u8],
+        _cert: &Certificate,
+        _dss: &DigitallySignedStruct,
+    ) -> Result<HandshakeSignatureValid, rustls::Error> {
+        // The libp2p handshake uses TLS 1.3 (and higher).
+        // Endpoints MUST NOT negotiate lower TLS versions.
+        Err(rustls::Error::PeerIncompatibleError(
+            "Only TLS 1.3 certificates are supported".to_string(),
```
marten-seemann: Why is this check done when verifying the certificate? The TLS version is negotiated before certificates are exchanged.

I also believe that the term "TLS 1.3 certificate" is misleading. Certificates are not bound to the TLS version.

transports/tls/src/certificate.rs
```diff
@@ -0,0 +1,552 @@
...
+        if !user_owns_sk {
+            return Err(Error::UnknownIssuer);
+        }
+
+        Ok(())
+    }
+
+    /// Return the signature scheme corresponding to [`AlgorithmIdentifier`]s
+    /// of `subject_pki` and `signature_algorithm`
+    /// according to `<https://tools.ietf.org/id/draft-ietf-tls-tls13-21.html#rfc.section.4.2.3>`.
```
marten-seemann: This should link to RFC 8446.

marten-seemann (approved):
Not sure my review actually adds a lot of value, since my Rust knowledge is very limited.

I was surprised to see so much code related to different signature algorithms, is there really no way to have the TLS library handle that for you?

transports/tls/src/verifier.rs
kpp: > The TLS version is negotiated before certificates are exchanged.

That's right and this is how it is done right now. So this is a no-op just in case.

transports/tls/Cargo.toml
thomaseizinger: Sure thing!

transports/tls/src/verifier.rs
thomaseizinger: > Why is this check done when verifying the certificate? The TLS version is negotiated before certificates are exchanged.

The TLS library at hand supports multiple TLS versions but they are all combined into one interface. Meaning at compile-time, we don't know which TLS version will be used and thus we need to satisfy this interface and provide an implementation for `verify_tls12_signature`. This function will should never be called because we disable anything other than 1.3 at configuration time. Due to TLS' flexibility, this is hard to express statically so I assume the authors of the TLS library didn't bother in modelling this using the type system.

> I also believe that the term "TLS 1.3 certificate" is misleading. Certificates are not bound to the TLS version.

I'll change the error message.

thomaseizinger:
> I was surprised to see so much code related to different signature algorithms, is there really no way to have the TLS library handle that for you?

I was surprised by that too! Note that I just extracted most of this code from #2289 so I am not actually the author of it.

I tried to use the existing verification logic from rustls but because our extension is marked as critical, it fails somewhere deep down in the verification chain. As a result, we have to do the parsing ourselves which requires mapping of the signature scheme to the verification algorithm:

https://github.com/libp2p/rust-libp2p/blob/a3af0173c4c894175341469f12b467fd6a28e8e6/transports/tls/src/certificate.rs#L265-L289

The mapping looks sound to me but we may be handling too many cases. I don't know enough about TLS1.3 to say this for sure but the default verification logic e.g. only handles a subset of these:

https://github.com/rustls/rustls/blob/16e38681574f5d7f4e33ac2a65255c7f587e9d68/rustls/src/verify.rs#L714-L731

I am pretty sure that some code here can be optimised. For example, we run all of our parsing logic for both callbacks: `verify_presented_certs` and `verify_tls13_signature`. That seems unnecessary but I don't understand enough about TLS and rustls to safely change this which is why I'd rather stay conservative. It would be good to get this audited though by someone who knows rustls very well.

kpp (approved):

kpp:
> The mapping looks sound to me but we may be handling too many cases.

I had to dig into RFC for TLS 1.3 to get the list of algorithms. These are not "too many", these are actual list of supported algos.

transports/tls/src/certificate.rs
```diff
@@ -1,4 +1,4 @@
-// Copyright 2021 Parity Technologies (UK) Ltd.
+// Copyright 2022 Protocol Labs.
```
bkchr: Just removing the copyright doesn't look right to me. I'm not a lawyer, but the original copyright should stay?

thomaseizinger:
> > The mapping looks sound to me but we may be handling too many cases.
> 
> I had to dig into RFC for TLS 1.3 to get the list of algorithms. These are not "too many", these are actual list of supported algos.

Yep, I've been down that road too when working on the code :)
What are your thoughts in regards to the snippet from rustls that I linked above? That only handles a much smaller set of algos.

transports/tls/src/certificate.rs
thomaseizinger: That was an oversight, thanks for pointing that out!

kpp:
> What are your thoughts in regards to the snippet from rustls that I linked above?

I could use https://github.com/barebones-x509/barebones-x509/ and reuse their code to minify the code in the tls module. But when I tried to dial to a go implementation it didn't work. (the crate was/is OK, we didn't use it properly).

So I carefully rewrote it controlling every part I could and trying not to bloat the code but to keep it reasonable. It took me a month or so to figure out what's broken. So the code can be compatible with the go impl and does not break on a weird but acceptable by RFC configuration.

You are free to refactor as you like leaving the authorship and copyrights. Feel free to ping me for an extra code review. I will be happy to assist.

> I could take all existing commits and put another one on top that deletes all the QUIC code. Given that we squash-merge, I am not sure it makes a big difference?

I don't know what's the point of co-authorship with squash-merge. The code was originally written by @Demi-Marie in https://github.com/libp2p/rust-libp2p/pull/1334 and rewritten by me looking at their code and at the barebones-x509 crate. So their co-authorship is not an open question.

thomaseizinger:
> > What are your thoughts in regards to the snippet from rustls that I linked above?
> 
> I could use [barebones-x509/barebones-x509](https://github.com/barebones-x509/barebones-x509/) and reuse their code to minify the code in the tls module. But when I tried to dial to a go implementation it didn't work. (the crate was/is OK, we didn't use it properly).
> 
> So I carefully rewrote it controlling every part I could and trying not to bloat the code but to keep it reasonable. It took me a month or so to figure out what's broken. So the code can be compatible with the go impl and does not break on a weird but acceptable by RFC configuration.
> 
> You are free to refactor as you like leaving the authorship and copyrights. Feel free to ping me for an extra code review. I will be happy to assist.

I am not planning to as I don't think I have the necessary knowledge at this point. Thanks for your prior work!

> > I could take all existing commits and put another one on top that deletes all the QUIC code. Given that we squash-merge, I am not sure it makes a big difference?
> 
> I don't know what's the point of co-authorship with squash-merge. The code was originally written by @Demi-Marie in #1334 and rewritten by me looking at their code and at the barebones-x509 crate. So their co-authorship is not an open question.

I am planning to attribute co-authorship to all people involved in this code on the squash-merge commit. I think that is fair. Please let me know if you disagree!

In particular, am planning on including the following (in alphabetical order):

```
Co-authored-by: David Craven <david@craven.ch>
Co-authored-by: Demi Marie Obenour <demiobenour@gmail.com>
Co-authored-by: Elena Frank <elena.frank@protonmail.com>
Co-authored-by: Max Inden <mail@max-inden.de
Co-authored-by: Pierre Krieger <pierre.krieger1708@gmail.com>
Co-authored-by: Roman Proskuryakov <r.proskuryakoff@gmail.com>
```

thomaseizinger:
This is ready to merge from my end. @mxinden @elenaf9 Waiting for your input / approval :)

transports/tls/Cargo.toml
```diff
@@ -0,0 +1,29 @@
+[package]
+name = "libp2p-tls"
+version = "0.1.0"
```
mxinden:
```suggestion
version = "0.1.0-alpha"
```

To the best of my knowledge, the code below has not run in any production systems.

To communicate expectations, I suggest releasing this as an alpha first.

See also same suggestion on QUIC pull request https://github.com/libp2p/rust-libp2p/pull/2289#discussion_r983395958

mxinden (approved):
This is good to merge from my end. Please go ahead @thomaseizinger unless @elenaf9 has any objections.

I am in favor of adding the above list of co-authors to the squash commit.

transports/tls/Cargo.toml
thomaseizinger: On it!

Most users depend on `libp2p` though so they are not going to see this version. IMO we are effectively removing that "alpha" label again through the re-export.

Do you think we should perhaps not re-export the TLS implementation yet so users have to depend on it manually?

mxinden: > Most users depend on `libp2p` though so they are not going to see this version. IMO we are effectively removing that "alpha" label again through the re-export.

That is true. Though I would explicitly call it out in the changelog.

> Do you think we should perhaps not re-export the TLS implementation yet so users have to depend on it manually?

I don't have an opinion here. I think we should be consistent with `libp2p-quic`. Chatted with @elenaf9 quickly. She prefers not removing `libp2p-quic` from `libp2p`.

elenaf9: What do you think about annotating the re-export with the [`unstable` attribute macro](https://rustc-dev-guide.rust-lang.org/stability.html#unstable) @thomaseizinger ?

thomaseizinger: > What do you think about annotating the re-export with the [`unstable` attribute macro](https://rustc-dev-guide.rust-lang.org/stability.html#unstable) @thomaseizinger ?

That is specific to the Rust compiler from what I know, I don't think you can use that actually?

thomaseizinger:
Currently blocked on https://github.com/libp2p/test-plans/pull/60.

transports/tls/Cargo.toml
elenaf9: Ah yes you are right! Don't have a strong on opinion on whether to re-export tls and quic or not. I did not consider that when re-exporting we "hide" the alpha-flag, so maybe not re-exporting is the best solution after all.

transports/tls/src/lib.rs
```diff
@@ -0,0 +1,78 @@
...
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+// FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+// DEALINGS IN THE SOFTWARE.
+
+//! TLS configuration based on libp2p TLS specs.
+//!
+//! See <https://github.com/libp2p/specs/blob/master/tls/tls.md>.
+
```
elenaf9, line 25:
```suggestion

#![cfg_attr(docsrs, feature(doc_cfg, doc_auto_cfg))]

```

transports/tls/Cargo.toml
thomaseizinger: I think not exporting it would be the better solution but I don't want to block this PR on it. We can always change that until the next release.

PR #2949, opened 2022-09-28
.cargo: Run `clippy` on ALL the source files

thomaseizinger:
## Description

image

I turns out you need a PhD in clippology to figure out how to have it lint every source file in a repository. Previously, we were not passing `--workspace` to clippy which means it didn't lint a lot of source files. I am not sure which subset it was linting and which one it skipped but with this PR, we should be linting every source file in this repository.

## Review guide

This PR is best reviewed patch-by-patch. The massive diff comes from the gossipsub tests but most of that is just removing one layer of modules which is done in a separate commit.

- The first commit is completely auto-generated by running `clippy --fix`. Probably not worth reviewing. If you run the command mentioned in the commit locally and follow it with `cargo fmt`, you should get the same result!
- The second commit is manual changes. Those should probably be reviewed.
- The third and forth commit just moves everything from the `tests` module one layer up, fixing the last clippy lint. I tried to split this up so the diff shows how minimal the changes are but Git chokes on it and likes to see unchanged newlines instead of just showing the whitespace diff. You can verify locally that it is just whitespace changes by running: `git diff 9443c7a4 9443c7a4^ --ignore-all-space`
- The fifth commit runs `cargo fmt`. You can again verify this locally if you want. Splitting this out makes the previous diff command work.
- The sixth commit enforces clippy across the entire workspace.

## Change checklist


protocols/gossipsub/src/peer_score/tests.rs
```diff
@@ -217,12 +225,14 @@ fn test_score_first_message_deliveries_cap() {
...
-    let mut topic_params = TopicScoreParams::default();
-    topic_params.topic_weight = 1.0;
-    topic_params.first_message_deliveries_weight = 1.0;
-    topic_params.first_message_deliveries_decay = 1.0; // test without decay
-    topic_params.first_message_deliveries_cap = 50.0;
-    topic_params.time_in_mesh_weight = 0.0;
+    let topic_params = TopicScoreParams {
+        topic_weight: 1.0,
+        first_message_deliveries_weight: 1.0,
+        first_message_deliveries_decay: 1.0,
```
thomaseizinger:
```suggestion
        first_message_deliveries_decay: 1.0, // test without decay
```

protocols/gossipsub/src/peer_score/tests.rs
```diff
@@ -261,12 +271,14 @@ fn test_score_first_message_deliveries_decay() {
...
-    let mut topic_params = TopicScoreParams::default();
-    topic_params.topic_weight = 1.0;
-    topic_params.first_message_deliveries_weight = 1.0;
-    topic_params.first_message_deliveries_decay = 0.9; // decay 10% per decay interval
-    topic_params.first_message_deliveries_cap = 2000.0;
-    topic_params.time_in_mesh_weight = 0.0;
+    let topic_params = TopicScoreParams {
+        topic_weight: 1.0,
+        first_message_deliveries_weight: 1.0,
+        first_message_deliveries_decay: 0.9,
```
thomaseizinger:
```suggestion
        first_message_deliveries_decay: 0.9, // decay 10% per decay interval
```

protocols/gossipsub/src/behaviour/tests.rs
```diff
@@ -3376,16 +3343,18 @@ mod tests {
...
-        topic_params.mesh_message_deliveries_threshold = 5.0;
-        topic_params.mesh_message_deliveries_activation = Duration::from_secs(1);
-        topic_params.mesh_message_deliveries_window = Duration::from_millis(100);
-        topic_params.topic_weight = 0.7;
+        let topic_params = TopicScoreParams {
+            time_in_mesh_weight: 0.0,             //deactivate time in mesh
+            first_message_deliveries_weight: 0.0, //deactivate first time deliveries
+            mesh_message_deliveries_weight: -2.0,
+            mesh_message_deliveries_decay: 0.9,
+            mesh_message_deliveries_cap: 0.0,
```
thomaseizinger:
```suggestion
            mesh_message_deliveries_cap: 10.0,
```

thomaseizinger (commented):
Made some mistakes apparently when fixing the lints.

protocols/gossipsub/tests/smoke.rs
```diff
@@ -187,7 +187,7 @@ fn multi_hop_propagation() {
     let _ = env_logger::try_init();
 
     fn prop(num_nodes: u8, seed: u64) -> TestResult {
-        if num_nodes < 2 || num_nodes > 50 {
+        if !(2..=50).contains(&num_nodes) {
```
mxinden, line 190: Not that it is important, though I am curious, do folks find this easier to read?

mxinden (approved):
First off, thanks for doing this!

Thanks for making it easy to review.

Thanks for making it very entertaining!

Goodbye to all those `.clone()` calls.

protocols/gossipsub/tests/smoke.rs
thomaseizinger: Pretty sure this one was done by clippy, I don't remember writing that.

If I were to write this test, I would probably split it into two `if`s:

```rust
if num_nodes < 2 {
	return TestResult::discard();
}

if num_nodes > 50 {
	return TestResult::discard();
}
```

thomaseizinger: Or make a newtype that implements `Arbitrary` that only generates numbers between 3 and 49. We can call it `NonZeroOneTwoU6Minus15`!

protocols/autonat/tests/test_client.rs
```diff
@@ -497,7 +489,8 @@ async fn test_outbound_failure() {
         }
 
         let inactive = servers.split_off(1);
-        // Drop the handles of the inactive servers to kill them.
+
+        #[allow(clippy::needless_collect)] // Drop the handles of the inactive servers to kill them.
```
jxs, line 493: out of curiosity to understand in this case why clippy was complaining I commented the `allow` and ran  the `cargo +nightly clippy --all-features --all-targets -- -A clippy::type_complexity -A clippy::pedantic -D warnings` command. But clippy didn't complain, is it not required?

jxs (approved):
LGTM, great effort Thomas!

protocols/gossipsub/tests/smoke.rs
mxinden: :+1: on the above suggestions in general, though I think in this particular case leaving it as is is just fine.

protocols/autonat/tests/test_client.rs
thomaseizinger: You need the `--workspace` flag! My best guess is that this is because our root Cargo.toml is a crate _and_ a workspace declaration.

I wonder whether we could avoid this if we put the meta crate in a directory too.

Thoughts @mxinden?

jxs: yeah, was running outdated rust version :sweat_smile: sorry. Btw with the latest nightly there are new lints triggered for [`large_enum_variant`](https://rust-lang.github.io/rust-clippy/master/index.html#large_enum_variant).

thomaseizinger: > yeah, was running outdated rust version 😅 sorry. Btw with the latest nightly there are new lints triggered for [`large_enum_variant`](https://rust-lang.github.io/rust-clippy/master/index.html#large_enum_variant).

Yeah I know :)

I opened https://github.com/libp2p/rust-libp2p/issues/2951 for this recently!

PR #2952, opened 2022-09-29
muxers: Add test harness for `StreamMuxer` implementations

### LINKED ISSUE #508: Add a crate that provides a test suit for an implementation of StreamMuxer

thomaseizinger:
# Description

This adds a basic test harness for stream muxers. It covers everything that was covered by the old mplex tests but additionally runs them also for yamux.

yamux doesn't pass one of them. I don't understand yet whether it is a harness bug or an actual bug in yamux. In any case, this is a strict improvement over what we currently have so I would like to not block this PR on that. The harness will be used to make progress on the WebRTC implementation. See https://github.com/melekes/rust-libp2p/pull/9#issuecomment-1258944682.

As part of writing this harness, I realised that `StreamMuxerExt::next_{inbound,outbound}` were a bad idea because they don't uphold the `StreamMuxer` contract of calling `StreamMuxer::poll` so I went ahead and deprecated them.

cc @melekes

- Fixes #508.
- https://github.com/melekes/rust-libp2p/pull/10

## Open tasks

- [x] Extract deprecation of `StreamMuxerExt` functions into separate PR: https://github.com/libp2p/rust-libp2p/pull/3002
- [x] Delete unnecessary `write_to_remote_closed` test
- [x] ~Add test for resetting stream on drop~

## Change checklist


melekes (approved):
👍

muxers/test-harness/src/lib.rs
```diff
@@ -0,0 +1,361 @@
...
+
+            stream.write_all(b"PONG").await.unwrap();
+            stream.close().await.unwrap();
+        },
+    )
+    .await;
+}
+
+/// Verifies that we can write to a substream that has been half-closed by the remote.
+pub async fn write_to_remote_closed<A, B, S, E>(alice: A, bob: B)
```
mxinden: How does this differ to `read_after_close`. In `read_after_close` Bob is writing to a substream that Alice closed, right?

muxers/test-harness/src/lib.rs
```diff
@@ -0,0 +1,361 @@
...
+
+            assert_eq!(buf, b"PING");
+        },
+        |mut stream| async move {
+            stream.write_all(b"PING").await.unwrap();
+            stream.close().await.unwrap();
+        },
+    )
+    .await;
+}
```
mxinden, line 106: This is the official winner in the competition of most-concise-test-case within the rust-libp2p repository!

muxers/yamux/tests/compliance.rs
```diff
@@ -0,0 +1,38 @@
+use libp2p_yamux::YamuxConfig;
+
+#[async_std::test]
+async fn close_implies_flush() {
```
mxinden, line 4: To deduplicate, would it make sense for `libp2p_muxer_test_harness` to offer a `run_all` method, taking a muxer constructor? That would prevent us adding a test to the harness but not to all the muxer implementations.

muxers/yamux/tests/compliance.rs
```diff
@@ -0,0 +1,38 @@
...
+    let (alice, bob) =
+        libp2p_muxer_test_harness::connected_muxers_on_memory_transport::<YamuxConfig, _, _>()
+            .await;
+
+    libp2p_muxer_test_harness::close_implies_flush(alice, bob).await;
+}
+
+#[async_std::test]
+#[ignore] // Hangs forever, is this a harness bug? It passes if we try to write to the stream.
```
mxinden, line 13: May it be the case that the Rust yamux implementation only announces the stream to the remote once it writes the first byte onto it?

core/src/muxing.rs
```diff
@@ -160,11 +160,19 @@ pub trait StreamMuxerExt: StreamMuxer + Sized {
     }
 
     /// Returns a future that resolves to the next inbound `Substream` opened by the remote.
+    #[deprecated(
```
mxinden: Good catch. Would you mind extracting this change into a separate pull request?

mxinden (commented):
Very much in favor of this change. Thanks @thomaseizinger for the excellent design.

muxers/yamux/tests/compliance.rs
thomaseizinger: I decided against it because I wanted to have individual tests in the output. Could be solved with a macro I guess but I don't think the current design is too much of a problem to keep in sync and it doesn't involve any magic.

muxers/test-harness/src/lib.rs
thomaseizinger: Yeah you are right. I can probably remove this one :)

muxers/test-harness/src/lib.rs
thomaseizinger: When it comes to tests, every character of noise counts. You don't want to read details that are not important :)

I'd like to get rid of the trait bounds put I don't think it is possible.

muxers/yamux/tests/compliance.rs
thomaseizinger: Could be yeah.

Because of `multistream-select`, we will always write something but it is still a bit odd.

core/src/muxing.rs
thomaseizinger: Done: https://github.com/libp2p/rust-libp2p/pull/3002

thomaseizinger:
I managed to adapt the mplex implementation so I got the reset test case to pass.

Going to work on yamux next.

thomaseizinger:
> I managed to adapt the mplex implementation so I got the reset test case to pass.

We should probably extract that into a separate PR. It might break something and that will make bisecting easier.

thomaseizinger:
Given that both our muxers, `mplex` and `yamux` need work to actually conform to the "send reset on drop" test, I suggest we merge this without all of that and just the first couple of tests. That is already an improvement.

We can then fix the muxer implementations in a separate PR!

thomaseizinger:
Removed the mplex changes as per comment above and chat with @mxinden out-of-band. This is ready for another review :)

mxinden (approved):
Again, wonderful work! Thanks Thomas!

thomaseizinger:
Merging because semver checks are a false-positive error.

PR #2954, opened 2022-09-29
transports/noise: Have NoiseAuthenticated::xx use X25519Spec

mxinden:
# Description

We support two versions of the Noise XX handshake with X25519, but only one of them is compliant with the specification and thus compliant with other implementations. We should always default to the spec compliant handshake.

Caught via [the libp2p interoperability tests](https://github.com/libp2p/test-plans). A node running `master` would no longer connect to nodes running older versions given that older versions would use the spec compliant Noise handshake.

Fixes bug introduced in https://github.com/libp2p/rust-libp2p/pull/2887/

## Open Questions

As a follow up, we might consider dropping support for the non-spec compliant version.

## Change checklist

  - Given that https://github.com/libp2p/rust-libp2p/pull/2887/ has not yet been released, we don't need a changelog entry.

### DIFF
--- a/transports/noise/src/lib.rs
+++ b/transports/noise/src/lib.rs
@@ -390,12 +390,12 @@ pub struct NoiseAuthenticated<P, C: Zeroize, R> {
     config: NoiseConfig<P, C, R>,
 }
 
-impl NoiseAuthenticated<XX, X25519, ()> {
+impl NoiseAuthenticated<XX, X25519Spec, ()> {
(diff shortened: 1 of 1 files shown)


jxs (approved):

mxinden:
Thanks @jxs.

I am merging here to unblock other pull requests. Please still give this a review @thomaseizinger and @elenaf9. I will do a follow-up pull requests for any suggestions you raise here.

thomaseizinger:
Thanks for the fix!

PR #2955, opened 2022-09-29
Don't panic on missing PendingConnectionInfo in swarm.pending

kpp:
# Description

Replace .expect with log::error + continue

https://github.com/paritytech/polkadot-sdk/issues/773

## Change checklist

- [ ] Do I need to replace all other similar patterns of `.remove().expect()`?
- [ ] Is there another solution?


swarm/src/connection/pool.rs
```diff
@@ -608,15 +608,20 @@ where
                     output: (obtained_peer_id, mut muxer),
                     outgoing,
                 } => {
+                    let pending_connection =
+                        if let Some(pending_connection) = self.pending.remove(&id) {
+                            pending_connection
+                        } else {
+                            log::error!("Failed to get an entry in `self.pending` for {:?}", id,);
```
eskimor, line 615: Do we know, how this was possible to happen?

mxinden: Pending connections should only be _aborted_ but never removed before the connection specific task shuts down.

https://github.com/libp2p/rust-libp2p/blob/f359c2ce5ce741f913e7d6e8d884552e7aead7ef/swarm/src/connection/pool.rs#L844-L851

I am saying _should_ as it might not be the case.

@kpp have you been able to further debug this? Can you add more details on the panic and the versions being used?

kpp: > have you been able to further debug this?

No I could not catch it.

libp2p 0.47.0. 

> Can you add more details on the panic

No, the issue above is all we have

thomaseizinger: I am taking this on to investigate.

thomaseizinger:
Setting to draft until we proceed.

thomaseizinger:
Many things around connection management have changed since this was opened. Let me know if you end up still encountering this in a recent release.

PR #2961, opened 2022-09-30
transports/tcp: Unify symbol naming

thomaseizinger:
# Description

This follows the effort described in https://github.com/libp2p/rust-libp2p/issues/2217. Instead of exporting one `TcpTransport` and a `TokioTcpTransport`, we use the already existing modules to define a `Transport` type alias.

The recommended way of referencing these new transports from the root `libp2p` crate is:

1. `use libp2p::tcp;`
2. Reference transports as `tcp::async_io::Transport` or `tcp::tokio::Transport`.

- #2217

## Change checklist


transports/tcp/CHANGELOG.md
```diff
@@ -7,8 +7,13 @@
 
 - Remove default features. If you previously depended on `async-std` you need to enable this explicitly now. See [PR 2918].
 
+- Deprecate types with `Tcp` prefix (`GenTcpConfig`, `TcpTransport` and `TokioTcpTransport`) in favor of referencing them by module / crate. See [PR XXXX].
+
+- Remove `TcpListenStream` and `TcpListenerEvent` from public API. See [PR XXXX].
+
 [PR 2813]: https://github.com/libp2p/rust-libp2p/pull/2813
 [PR 2918]: https://github.com/libp2p/rust-libp2p/pull/2918
+[PR XXXX]: https://github.com/libp2p/rust-libp2p/pull/XXXX
```
thomaseizinger:
```suggestion
- Deprecate types with `Tcp` prefix (`GenTcpConfig`, `TcpTransport` and `TokioTcpTransport`) in favor of referencing them by module / crate. See [PR XXXX].

- Remove `TcpListenStream` and `TcpListenerEvent` from public API. See [PR 2961].

[PR 2813]: https://github.com/libp2p/rust-libp2p/pull/2813
[PR 2918]: https://github.com/libp2p/rust-libp2p/pull/2918
[PR 2961]: https://github.com/libp2p/rust-libp2p/pull/2961
```

transports/tcp/CHANGELOG.md
```diff
@@ -7,8 +7,13 @@
 
 - Remove default features. If you previously depended on `async-std` you need to enable this explicitly now. See [PR 2918].
 
+- Deprecate types with `Tcp` prefix (`GenTcpConfig`, `TcpTransport` and `TokioTcpTransport`) in favor of referencing them by module / crate. See [PR XXXX].
```
elenaf9:
```suggestion
- Deprecate types with `Tcp` prefix (`GenTcpConfig`, `TcpTransport` and `TokioTcpTransport`) in favor of referencing them by module / crate. See [PR 2961].
```

elenaf9 (approved):
👍

protocols/relay/examples/relay_v2.rs
```diff
@@ -25,12 +25,13 @@ use futures::stream::StreamExt;
 use libp2p::core::upgrade;
 use libp2p::identify::{Identify, IdentifyConfig, IdentifyEvent};
 use libp2p::multiaddr::Protocol;
+use libp2p::ping;
 use libp2p::relay::v2::relay::{self, Relay};
 use libp2p::swarm::{Swarm, SwarmEvent};
-use libp2p::tcp::TcpTransport;
```
jxs, line 30: unrelated with this PR, should we setup a `rustfmt.toml` and configure [`imports_granularity`](https://rust-lang.github.io/rustfmt/?version=v1.5.1&search=#imports_granularity) to maintain import consistency across the repo?

jxs (approved):
LGMT :+1:

protocols/relay/examples/relay_v2.rs
thomaseizinger: That would be great but unfortunately, the setting is not yet stable so we can't use it.

You can hack around that but defining an alias same as we do for `clippy`. See [here](https://github.com/Restioson/xtra/blob/master/.cargo/config.toml) for example.

I'd be in favor of that :)
I think I tried at some point but not sure why I didn't go through with it :thinking:

elenaf9 (changes requested):
Sorry, two things just came to my mind when I though about following the same naming for the QuicTransport; 
1. How will `tokio::Transport` and `async_io::Transport` be show on docs.rs if `lib::Transport` is annotated with `#[doc(hdiden)]`? I believe if we mark this as hidden user will see the type aliases for each provider, but not be able to the see the public interface of `Transport`, thus also not what method it has and what traits are implemented on it. Is this correct? if so, I'd be strongly in favor of not hiding `Transport`.
2. Why are we in general not exporting `GenTransport` and `Provider` to the user, if they (for whatever reason) want to implement a custom provider? We already have this abstraction, why not expose it?

thomaseizinger:
> Sorry, two things just came to my mind when I though about following the same naming for the QuicTransport;
> 
>     1. How will `tokio::Transport` and `async_io::Transport` be show on docs.rs if `lib::Transport` is annotated with `#[doc(hdiden)]`? I believe if we mark this as hidden user will see the type aliases for each provider, but not be able to the see the public interface of `Transport`, thus also not what method it has and what traits are implemented on it. Is this correct? if so, I'd be strongly in favor of not hiding `Transport`.

I am linking to the `libp2p_core::Transport` trait in the docs. Is that good enough?
I can also add an example on how to use it!

>     2. Why are we in general not exporting `GenTransport` and `Provider` to the user, if they (for whatever reason) want to implement a custom provider? We already have this abstraction, why not expose it?

We are exporting `GenTransport` (now called `Transport`). We have to because "it appears in a public interface". The reason I'd rather like to hide it is because it is completely uninteresting to users IMO. The interesting interface is the `libp2p_core::Transport` implementation.

Regarding `Provider`: I'd rather not expose it. If it is an implementation detail we can make changes to it without breaking backwards compatibility. Support for new executors will ideally be upstreamed to us.

thomaseizinger:
This is what it looks like now:

image

thomaseizinger:
I ended up removing the `hidden` attribute again because I couldn't make it work to show the `new` function otherwise.

mxinden:
@thomaseizinger is this ready for another review? If so @elenaf9 can you take another look and check whether your concerns are addressed?

elenaf9 (approved):
Sorry for the delay. LGTM!

thomaseizinger:
> @thomaseizinger is this ready for another review? If so @elenaf9 can you take another look and check whether your concerns are addressed?

Yep this is ready to review. Will fix the merge conflicts.

mxinden (approved):
As mentioned before, big fan of the first-deprecate-then-remove path now. Thanks for making `libp2p-tcp` consistent with the recent renames.

I am in favor of merging here. Though I think we should tackle https://github.com/libp2p/test-plans/issues/59 first.

PR #2962, opened 2022-09-30
Cargo.toml: Deprecate executor specific features for sub-crates

thomaseizinger:
# Description

Typically, an application will only use one kind of executor like `tokio` or `async-std`. Currently, users of libp2p need to activate a separate feature for each sub-crate that is specific to the executor they want to use, i.e. `tcp-tokio`, `dns-tokio`, etc.

With this PR, we deprecate these features in favor of a single `tokio` and `async-std` feature. Thanks to cargo's support for weak-features, we can forward the respective executor feature to each sub-crate without pulling them in as dependencies unless the user also activated the corresponding protocol (`dns`, `tcp` etc).

Deprecation warnings don't work on re-exports so I had to (temporarily) define a new module in `lib.rs`. This module should be removed again once we actually remove the deprecated features.

Here is a screenshot of the warning being active:

image

This is a backwards-compatible change.

The warning goes away as soon as we stop using the deprecated feature:

image

## Change checklist


CHANGELOG.md
```diff
@@ -75,7 +85,8 @@
     - Update to [`libp2p-websocket` `v0.39.0`](transports/websocket/CHANGELOG.md#0390).
     - Update to [`libp2p-yamux` `v0.41.0`](muxers/mplex/CHANGELOG.md#0410).
 
-- [PR 2918]: https://github.com/libp2p/rust-libp2p/pull/2918
+[PR 2918]: https://github.com/libp2p/rust-libp2p/pull/2918
+[PR XXXX]: https://github.com/libp2p/rust-libp2p/pull/XXXX
```
```suggestion
[PR 2962]: https://github.com/libp2p/rust-libp2p/pull/2962
```

CHANGELOG.md
```diff
@@ -48,6 +48,16 @@
...
 
+- Introduce `tokio` and `async-std` features and deprecate the following ones:
+  - `tcp-tokio` in favor of `tcp` + `tokio`
+  - `mdns-tokio` in favor of `mdns` + `tokio`
+  - `dns-tokio` in favor of `dns` + `tokio`
+  - `tcp-async-io` in favor of `tcp` + `async-std`
+  - `mdns-async-io` in favor of `mdns` + `async-std`
+  - `dns-async-std` in favor of `dns` + `async-std`
+  
+  See [PR XXXX].
```
```suggestion
  See [PR 2962].
```

elenaf9:
In favor of this! 
But I wonder if it is really necessary to deprecate the old features, rather than just removing them directly. Upgrading to the new feature set is very easy, whereas the deprecation logic added in this PR is quite complex.

thomaseizinger:
> In favor of this! But I wonder if it is really necessary to deprecate the old features, rather than just removing them directly. Upgrading to the new feature set is very easy, whereas the deprecation logic added in this PR is quite complex.

The annoying thing with removing features is that `cargo` completely chokes and gives up compiling altogether if it can't find the feature with a given name. The upgrade might be easy but it looks completely broken which might delay looking at a dependabot PR in detail.

The logic is a bit complex but it is only temporary until the release is out. We can remove it immediately after so I wouldn't be too worried about it :)

mxinden (approved):
I am in favor the change in general.

I am in favor of the two step (deprecation + removal) process. While not directly related, I think https://github.com/libp2p/rust-libp2p/issues/2971 is a good data point for the argument that configuring rust-libp2p features is hard. Thus I think making it easier for users at the expense of complexity on our end makes sense.

Before we merge here, can you create a second pull request removing the deprecated modules. Once this is merged and released, we can then merge the second pull request.

thomaseizinger:
> Before we merge here, can you create a second pull request removing the deprecated modules. Once this is merged and released, we can then merge the second pull request.

Done here: https://github.com/libp2p/rust-libp2p/pull/3001

PR #2963, opened 2022-09-30
.github/worksflows/ci: Fix bad use of bash

thomaseizinger:
# Description

If only my Bash-fu wouldn't be so bad ...

## Change checklist


mxinden (approved):
Took me a bit to find the mistake :detective: 

In general, I am open to switching to a better scripting language, e.g. Python, if folks like.

Not that my Python skills are better than my Bash skills. Much rather that I think Python helps me make less mistakes.

PR #2970, opened 2022-10-03
transports/tcp: only translate tcp addresses

elenaf9:
# Description

Return `None` in in `<GenTcpTransport as Transport>::address_translation` if the address is not a tcp address.
Relevant if in case of something like `OrTransport<TcpTransport, QuicTransport>`, where tcp would currently perform the address translation for quic addresses. 
Raise by @mxinden for quic in https://github.com/libp2p/rust-libp2p/pull/2289#discussion_r983428741, so I think the same should also be applied for the tcp transport.

## Change checklist


transports/tcp/src/lib.rs
```diff
@@ -829,13 +832,28 @@ fn ip_to_multiaddr(ip: IpAddr, port: u16) -> Multiaddr {
...
+        iter.next(),
+        Some(Protocol::Ip4(_))
+            | Some(Protocol::Ip6(_))
+            | Some(Protocol::Dns(_))
+            | Some(Protocol::Dns4(_))
+            | Some(Protocol::Dns6(_))
+    );
+    is_tcp &= matches!(iter.next(), Some(Protocol::Tcp(_)));
+    is_tcp
+}
```
thomaseizinger:
```suggestion
fn is_tcp_addr(addr: &Multiaddr) -> bool {
    use Protocol::*;

    let mut iter = addr.iter();

    let first = match iter.next() {
        None => return false,
        Some(p) => p,
    };
    let second = match iter.next() {
        None => return false,
        Some(p) => p,
    };

    matches!(first, Ip4(_) | Ip6(_) | Dns(_) | Dns4(_) | Dns6(_)) && matches!(second, Tcp(_))
}
```

Thoughts?

transports/tcp/src/lib.rs
```diff
@@ -1240,4 +1258,43 @@ mod tests {
...
+    fn test_address_translation() {
+        #[cfg(feature = "async-io")]
+        let transport = TcpTransport::new(GenTcpConfig::new());
+        #[cfg(all(feature = "tokio", not(feature = "async-io")))]
+        let transport = TokioTcpTransport::new(GenTcpConfig::new());
+
+        let port = 42;
+        let tcp_listen_addr = Multiaddr::empty()
+            .with(Protocol::Ip4(Ipv4Addr::new(127, 0, 0, 1)))
+            .with(Protocol::Tcp(port));
```
thomaseizinger, line 1276: I always find these easier to read if we parse them from a string.

thomaseizinger (approved):
LGTM!

transports/tcp/src/lib.rs
elenaf9: I prefer using types where possible, avoiding any unwraps that would be needed when parsing from string :)

transports/tcp/src/lib.rs
elenaf9: No strong opinion on my side. Added your suggestion :)

mxinden (approved):
Thank you for the fix @elenaf9!

Issue #2950, opened 2022-09-29
Properly set rustdoc attributes for docs.rs


thomaseizinger:
On the topic of:

```toml
[package.metadata.docs.rs]
rustdoc-args = ["--cfg", "docsrs"]
rustc-args = ["--cfg", "docsrs"]
```

What this does is tell docs.rs to pass the `docsrs` conditional compilation flag to rustc, meaning any code that is guarded with `cfg(docsrs)` will now be included in the build.

We use this in the `libp2p` root crate to conditionally include the `doc` attribute: https://github.com/libp2p/rust-libp2p/blob/04b07c4d14a0a6ec350fc21261a33d859fc71289/src/lib.rs#L53

We need to do this conditionally because the `#[doc(cfg)]` attribute is [not yet stable](https://github.com/rust-lang/rust/issues/43781). docs.rs runs all their build pipelines on nightly so we can use it there.

Now, interestingly, we don't set this in our `libp2p` crate so all of these `cfg_attr` attributes have been doing nothing all this time :)

In this crate, it is used here: https://github.com/umgefahren/rust-libp2p/blob/6f4bca5bbfd141bc18d9d931c5d28b50a212f0fe/transports/onion/src/provider.rs#L31-L36

And will tell docs.rs to include a little "only available with feature `tokio`" tag next to the item.

_Originally posted by @thomaseizinger in https://github.com/libp2p/rust-libp2p/pull/2899#discussion_r982971842_

umgefahren:
I'm currently working on it, however it is a lot of effort, since I have to go through nearly every single line of code here.

I will make a pull request with the changes I applied to libp2p-core, so you can judge wether the changes are good.

PR #2975, opened 2022-10-04
Properly set rustdoc attr. Closing #2950

umgefahren:
# Description

I added the necessary rustdoc flags in docs.rs in all crates. It was just easier to do it on all Cargo.toml files, we can remove the addition if there are any subcrates which don't have any cfg's in them.

This is the solution for #2950, which was detected during development of #2899. Since we remove all default features in #2918 this might me more relevant.

## Open Questions

Are you happy with the changes I made to libp2p-core? I will continue, once this is approved. We also might consider, that there is already a stabilization PR for the feature (rust-lang/rust#100883). However this doesn't makes this work here obsolete. We could just do a regex on all code and remove the `cfg_attr(docsrs` part and the `#![cfg_attr(docsrs, feature(doc_cfg))]`. Or we just wait

## Change checklist

Quite unsure what to do about the changelog here...


umgefahren:
In order to build the doc like docs.rs would do it (hopefully):
```bash
RUSTDOCFLAGS="--cfg docsrs" RUSTFLAGS="--cfg docsrs" cargo doc --open --workspace --all-features
```
The `--open` flag is optional.

umgefahren:
We might consider adding a something to the CI to run this.

core/src/identity.rs
```diff
@@ -32,12 +32,15 @@
 //! (e.g. [ed25519 binary format](https://datatracker.ietf.org/doc/html/rfc8032#section-5.1.5)).
 //! All key types have functions to enable conversion to/from their binary representations.
 
-#[cfg(feature = "ecdsa")]
+#[cfg(any(feature = "ecdsa", docsrs))]
```
thomaseizinger: Why are we combining these with `any`?

Shouldn't the regular feature be enough?

core/src/lib.rs
```diff
@@ -35,6 +35,8 @@
 //!   define how to upgrade each individual substream to use a protocol.
 //!   See the `upgrade` module.
 
+#![cfg_attr(docsrs, feature(doc_cfg))]
```
thomaseizinger, line 38: These seems like it is a left over from another attempt at solving this.

thomaseizinger (commented):
Thanks for working on this!

It is a bit messy but there isn't actually a whole lot of cfg code in the repository. `libp2p-core` is probably the biggest user of it.

Are all the items you are annotating actually public API? Because we don't need to put it on the ones that aren't.

core/src/identity.rs
umgefahren: Yes, you are right. Only if it's compiled with all features though, but yes.

core/src/lib.rs
umgefahren: No I don't think so. This turns the required feature gate on. I.e. `#[doc(cfg(...))]`.

thomaseizinger: Ah yes, my bad! Thanks for clarifying!

Cargo.toml
```diff
@@ -201,3 +198,8 @@ required-features = ["full"]
 [[example]]
 name = "distributed-key-value-store"
 required-features = ["full"]
+
+[package.metadata.docs.rs]
+all-features = true
+rustdoc-args = ["--cfg", "docsrs"]
+rustc-args = ["--cfg", "docsrs"]
```
thomaseizinger, line 205: Can we add a newline so this icon goes away? :innocent:

core/src/identity.rs
```diff
@@ -33,11 +33,14 @@
 //! All key types have functions to enable conversion to/from their binary representations.
 
 #[cfg(feature = "ecdsa")]
+#[cfg_attr(docsrs, doc(cfg(feature = "ecdsa")))]
```
thomaseizinger, line 36: I wonder whether it is worth it using a macro to remove this duplication. I can see a couple of downsides, not sure how critical they actually are:

- Less flexibility in specifying the rendered feature combination.
- Increased compile-time because the macro needs to run before all our other crates.
- Newcomers will be unfamiliar with the solution.

protocols/mdns/Cargo.toml
```diff
@@ -46,3 +46,7 @@ required-features = ["async-io"]
 name = "use-tokio"
 required-features = ["tokio"]
 
+[package.metadata.docs.rs]
+all-features = true
+rustdoc-args = ["--cfg", "docsrs"]
+rustc-args = ["--cfg", "docsrs"]
```
thomaseizinger, line 52: I think it may be worthwhile to just copy-paste this into all manifests? Otherwise, if we add features later we may forget about this.

We can include a link to [this](https://docs.rs/about/builds#cross-compiling) documentation to make this a bit easier to understand.

core/src/identity.rs
thomaseizinger: I just found `doc_auto_cfg` which seems to be useful to automate all of this. The magic line seems to be:

```rust
```

See https://github.com/tokio-rs/valuable/pull/80/files for example.

thomaseizinger (commented):
Thanks for working on this!

I'd like us to explore the use of `doc_auto_cfg` and see what the output looks like! :)

core/src/identity.rs
umgefahren: I will try that, although if we decide to go with this, it would be much easier for me to do that in a new branch (and pull request).

umgefahren:
`#![feature(doc_cfg, doc_cfg_auto)]` seems like the way to go. Since this PR is really cluttered now, I will open a new one.

PR #2976, opened 2022-10-04
core/transport: Improve docs for `Transport::address_translation`

elenaf9:
# Description

- de05313d4a2629e918378271dd67bc4b55b2e297: enhance docs for `Transport::address_translation`
- <s>3785dd8c5aec63017e2dbbd6d79b5f313af99673: proposal: add default implementation for `Transport::address_translation` that returns `None`.</s>

Motivated by a recent discussion in https://github.com/libp2p/rust-libp2p/pull/2289#discussion_r983428741 and a follow-up out of band discussion with @mxinden , that showed that the usage and relevance of this method is currently not well documented.

## Open Questions

<s>Does a default `None` implementation for `Transport::address_translation` make sense? It might help emphasizing that this is not relevant for all transports? Not sure about this myself, happy to revert that commit.</s>

## Change checklist


core/src/transport.rs
```diff
@@ -149,10 +149,24 @@ pub trait Transport {
...
+    /// Within [`libp2p_swarm::Swarm`][libp2p_swarm::Swarm] this is used when extending the
+    /// listening addresses of the local peer with external addresses observed by remote peers.
+    /// On transports where this is not relevant (i.e. no NATs are present) `None` should be
+    /// returned for the sake of de-duplication.
+    ///
+    /// Note: if the listen or observed address is not a valid address of this transport,
+    /// `None` should be returned as well.
+    fn address_translation(&self, _listen: &Multiaddr, _observed: &Multiaddr) -> Option<Multiaddr> {
+        None
+    }
```
mxinden:
```suggestion
    fn address_translation(&self, _listen: &Multiaddr, _observed: &Multiaddr) -> Option<Multiaddr>;
```

Implementing `address_translation` for combinator `Transport` implementations like `OrTransport` is relevant. I suggest we don't provide a default implementation. In case we write another combinator `Transport` implementation like `OrTransport`, we won't miss implementing it.

mxinden (commented):
Very much in favor of the updates docs. I think we should do without the default implementation. Let me know in case the reasoning makes sense.

core/src/transport.rs
thomaseizinger: Agree that explicit is better. If `Transport` were something that is implemented a lot outside of this repo it'd be different but it isn't really so it seems like a footgun to me.

elenaf9:
Force-pushed to drop 3785dd8c5aec63017e2dbbd6d79b5f313af99673.

I believe a changelog entry is not needed given that I just updated the docs?

core/src/transport.rs
elenaf9: Dropped the commit.

core/src/transport.rs
```diff
@@ -149,9 +149,22 @@ pub trait Transport {
...
-    /// the local node that may be reachable for other peers.
+    /// Performs a transport-specific mapping of an address `observed` by a remote onto a
+    /// local `listen` address to yield an address for the local node that may be reachable
+    /// for other peers.
+    ///
+    /// This is relevant for transports where Network Address Translation (NAT) can occur
+    /// so that e.g. the peer is observed at a different IP than the IP of the local
+    /// listening address. See also [`address_translation`][crate::address_translation].
+    ///
+    /// Within [`libp2p::Swarm`](<https://docs.rs/libp2p/latest/libp2p/struct.Swarm.html>) this is
```
elenaf9, line 160: Is there a better way to link to items from crates that are not a dependency?

mxinden: I am not aware of a way. Maybe @thomaseizinger?

mxinden (approved):

core/src/transport.rs
thomaseizinger: Me neither but I also think it is kind of an odd thing to do. This crate shouldn't "know" about `Swarm`. Do we have to include this sentence here?

thomaseizinger (approved):

core/src/transport.rs
elenaf9: > Me neither but I also think it is kind of an odd thing to do. This crate shouldn't "know" about `Swarm`. Do we have to include this sentence here?

Good point. But in practice transports are commonly used in a swarm, so I think its okay to mention this here. When I was looking at it for Quic it helped me to know how it is used in the swarm.

PR #2977, opened 2022-10-04
protocols/mdns: Removed obsolete lazy static use from mdns

umgefahren:
# Description

While implementing #2975 I discovered an unnecessary use of lazy_static. It can be replaced with a `const`, since there is const support for the constructor of Ipv4Addr and Ipv6Addr since Rust version `1.32.0`. This also removes an unnecessary dependency and might result in better performance, although very, very marginal.

None

## Open Questions

None.

## Change checklist


umgefahren (commented):
Seems correct.

mxinden (approved):
Wonderful to see. Thanks for the patch @umgefahren!

thomaseizinger (approved):
Nice, thanks!

PR #2982, opened 2022-10-04
misc/metrics: Add `protocols` label to address-specific metrics

### LINKED ISSUE #2758: misc/metrics/: Track address stack
## Description

In order to answer questions like:

- Do most of the connections to remote peers use IPv4 or IPv6? TCP, Quic or WebRTC? ...
- What protocols do remote peers support? IPv4 or IPv6? TCP, Quic or WebRTC?

I think we should expose metrics like:

```
libp2p_swarm_connections_incoming{address_stack="/ip4/tcp/"} 42
libp2p_swarm_connections_incoming{address_stack="/ip4/udp/quic/"} 42
```

and 

```
# HELP: Number of connected nodes supporting a specific listen address stack, with "unrecognized\" for each peer supporting one or more unrecognized protocols.
libp2p_identify_listen_addresses{address_stack="/ip4/tcp/"} 42
```

## Requirements

1. A way to strip a `Multiaddr` of the variable parts (e.g. concrete IP or concrete port) to prevent a cardinality explosion. See example implementation https://github.com/mxinden/kademlia-exporter/blob/7c6ced0c857152f8798a54d89648c54f65d47817/src/exporter/node_store.rs#L318-L358

Not any time soon. Happy to help contributors.

John-LittleBearLabs:
## Description

Previously, we would only track the metrics like the number of open connections. With this patch, we extend these metrics with a `protocols` label that contains a "protocol stack". A protocol stack is a multi-address with all variable parts removed. For example, `/ip4/127.0.0.1/tcp/1234` turns into `/ip4/tcp`.

Resolves https://github.com/libp2p/rust-libp2p/issues/2758.

## Open Questions

~The use of Family appends `_total` to the metric name. Is this desirable?~

## Change checklist


misc/metrics/Cargo.toml
```diff
@@ -21,11 +21,11 @@ dcutr = ["libp2p-dcutr"]
 [dependencies]
 libp2p-core = { version = "0.37.0", path = "../../core" }
 libp2p-dcutr =  { version = "0.7.0", path = "../../protocols/dcutr", optional = true }
-libp2p-identify = { version = "0.40.0", path = "../../protocols/identify", optional = true }
+libp2p-identify = { path = "../../protocols/identify", optional = true }
```
thomaseizinger: Please retain the versions here.

Cargo.toml
```diff
@@ -109,11 +109,12 @@ libp2p-swarm-derive = { version = "0.30.1", path = "swarm-derive" }
 libp2p-uds = { version = "0.36.0", path = "transports/uds", optional = true }
 libp2p-wasm-ext = { version = "0.37.0", path = "transports/wasm-ext", optional = true }
 libp2p-yamux = { version = "0.41.0", path = "muxers/yamux", optional = true }
-multiaddr = { version = "0.14.0" }
+multiaddr = { path = "../rust-multiaddr" }
```
thomaseizinger: Can we for now have the `protocol_stack` function in here? Upstreaming it right away doesn't feel right as we may have to iterate on it for a bit.

thomaseizinger (commented):
Thanks for working on this!

misc/metrics/Cargo.toml
```diff
@@ -23,8 +23,8 @@ libp2p-core = { version = "0.37.0", path = "../../core" }
 libp2p-dcutr =  { version = "0.7.0", path = "../../protocols/dcutr", optional = true }
 libp2p-identify = { version = "0.40.0", path = "../../protocols/identify", optional = true }
 libp2p-kad = { version = "0.41.0", path = "../../protocols/kad", optional = true }
-libp2p-ping = { version = "0.40.0", path = "../../protocols/ping", optional = true }
-libp2p-relay =  { version = "0.13.0", path = "../../protocols/relay", optional = true }
+libp2p-ping = { version = "0.40.0",  path = "../../protocols/ping", optional = true }
+libp2p-relay =  {  version = "0.13.0", path = "../../protocols/relay", optional = true }
```
thomaseizinger: Please undo this formatting :)

misc/metrics/examples/metrics/main.rs
```diff
@@ -109,8 +113,23 @@ fn main() -> Result<(), Box<dyn Error>> {
 ///
 /// For illustrative purposes, this includes the [`keep_alive::Behaviour`]) behaviour so the ping actually happen
 /// and can be observed via the metrics.
-#[derive(NetworkBehaviour, Default)]
+#[derive(NetworkBehaviour)]
+// #[behaviour(out_event = "Event")]
```
thomaseizinger:
```suggestion
```

misc/metrics/src/identify.rs
```diff
@@ -27,6 +27,8 @@ use prometheus_client::registry::Registry;
 use std::collections::HashMap;
 use std::iter;
 use std::sync::{Arc, Mutex};
+use prometheus_client::metrics::family::Family;
+use super::protocol_stack::protocol_stack;
```
thomaseizinger: Please use `crate::` for importing crate-local functions.

misc/metrics/src/protocol_stack.rs
```diff
@@ -0,0 +1,13 @@
...
+
+//TODO: remove this file and replace with calls to Multiaddr::protocol_stack
+//  once that lands : https://github.com/multiformats/rust-multiaddr/pull/60
+//  This does imply changing the keys from Vec<String> to Vec<'static str>
+
+fn tag(proto: Protocol) -> Option<String> {
+    format!("{}", proto).split("/").skip(1).next().map(str::to_owned)
+}
+
+pub(crate) fn protocol_stack(ma: &Multiaddr) -> Vec<String> {
```
thomaseizinger:
```suggestion
pub fn protocol_stack(ma: &Multiaddr) -> Vec<String> {
```

The entire module is crate-private so we don't need this! :)

misc/metrics/src/protocol_stack.rs
```diff
@@ -0,0 +1,13 @@
+use libp2p_core::multiaddr::{Multiaddr,Protocol};
+
+//TODO: remove this file and replace with calls to Multiaddr::protocol_stack
+//  once that lands : https://github.com/multiformats/rust-multiaddr/pull/60
+//  This does imply changing the keys from Vec<String> to Vec<'static str>
+
+fn tag(proto: Protocol) -> Option<String> {
+    format!("{}", proto).split("/").skip(1).next().map(str::to_owned)
+}
```
thomaseizinger: Can we have helpers at the bottom of the file please?

misc/metrics/src/protocol_stack.rs
thomaseizinger: Can we have a few tests for this please?

misc/metrics/src/protocol_stack.rs
```diff
@@ -0,0 +1,13 @@
+use libp2p_core::multiaddr::{Multiaddr,Protocol};
+
+//TODO: remove this file and replace with calls to Multiaddr::protocol_stack
+//  once that lands : https://github.com/multiformats/rust-multiaddr/pull/60
+//  This does imply changing the keys from Vec<String> to Vec<'static str>
+
+fn tag(proto: Protocol) -> Option<String> {
+    format!("{}", proto).split("/").skip(1).next().map(str::to_owned)
```
thomaseizinger: It would be great to avoid these allocations here. It creates some duplication but we could `match` on all variants and return a `&'static str` instead of using `Display`.

Metrics should be cheap to record so I think this is worth it.

misc/metrics/src/protocol_stack.rs
```diff
@@ -0,0 +1,13 @@
+use libp2p_core::multiaddr::{Multiaddr,Protocol};
+
+//TODO: remove this file and replace with calls to Multiaddr::protocol_stack
+//  once that lands : https://github.com/multiformats/rust-multiaddr/pull/60
+//  This does imply changing the keys from Vec<String> to Vec<'static str>
```
thomaseizinger: To signal that this is function has upstream potential, we could model it as an extension trait:

```rust
trait MultiaddrExt {
	fn protocol_stack(&self) -> String;
}
```

misc/metrics/src/swarm.rs
```diff
@@ -156,8 +157,9 @@ impl<TBvEv, THandleErr> super::Recorder<libp2p_swarm::SwarmEvent<TBvEv, THandleE
                     })
                     .inc();
             }
-            libp2p_swarm::SwarmEvent::IncomingConnection { .. } => {
-                self.connections_incoming.inc();
+            libp2p_swarm::SwarmEvent::IncomingConnection { send_back_addr, .. } => {
+                let labels = protocol_stack(send_back_addr);
+                self.connections_incoming.get_or_create(&labels).inc();
```
thomaseizinger: I don't think we want to create one label per protocol. Instead we want one label per entire stack, i.e. the labels here should be all concatenated.

This is probably best to happen inside the `protocol_stack` function so it only returns a `String`.

misc/metrics/src/swarm.rs
```diff
@@ -246,6 +248,7 @@ impl<TBvEv, THandleErr> super::Recorder<libp2p_swarm::SwarmEvent<TBvEv, THandleE
 #[derive(Encode, Hash, Clone, Eq, PartialEq)]
 struct ConnectionEstablishedLabels {
     role: Role,
+
```
thomaseizinger: Please remove this unneeded reformatting.

thomaseizinger (changes requested):
Thanks! A few comments.

misc/metrics/src/protocol_stack.rs
John-LittleBearLabs: Honestly I was trying to go minimal here because in my mind the "real" implementation is still [in multiaddr](https://github.com/multiformats/rust-multiaddr/pull/60). But yeah I could put a test here too.

misc/metrics/src/protocol_stack.rs
John-LittleBearLabs: Yeah the rust-multiaddr implementation is an `Iterator<Item = &'static str>` because I had the same basic concern. I can copy-paste more.

misc/metrics/src/protocol_stack.rs
John-LittleBearLabs: I'm not sure how that would look. We'd be writing an `impl MultiaddrExt for Multiaddr` in both repos? Or a helper struct in this repo that wraps around Multiaddr?

misc/metrics/src/swarm.rs
John-LittleBearLabs: That actually raises a risk of collision (not right now, but a potential future tag) unless a delimiter is used. The most obvious delimiter would be "/".

After we go to `&str` it should be a similar number of allocations between `Vec<&'static str>` and `String`, correct? Are you thinking comparisons and hashing is faster with String or something of that nature?

John-LittleBearLabs:
With the change in key the metrics look slightly different now

````
libp2p_swarm_connections_incoming_total{ip4/tcp} 1
libp2p_identify_listen_addresses_total{ip4/tcp} 5
````

File: misc/metrics/src/protocol_stack.rs (hunk 9 above)
Comment (thomaseizinger) [reply misc/metrics/src/protocol_stack.rs]: Not quite. The idea I wanted to convey is that from a user PoV, an extension trait makes the interface look like it is already on `Multiaddr`, i.e. we are _extending_ `Multiaddr` with functionality that we would like to have there. Once you do `impl MultiaddrExt for Multiaddr`, you can just call `multiaddr.protocol_stack()` as if `Multiaddr` would natively provide that function.

That way, call-sites don't have to change once we do actually upstream it. At that point, the trait can be deleted and the function just goes onto a regular `impl Multiaddr` block.

Semantically, it doesn't make a difference but I find that an extension trait conveys a different intent compared to a module-private utility function :)

I don't feel strongly about it though so feel free to go with whatever!

File: misc/metrics/src/swarm.rs (hunk 10 above)
Comment (thomaseizinger) [reply misc/metrics/src/swarm.rs]: > That actually raises a risk of collision (not right now, but a potential future tag) unless a delimiter is used. The most obvious delimiter would be "/".

Yeah, the original issue uses `/` as a delimiter: https://github.com/libp2p/rust-libp2p/issues/2758

> After we go to `&str` it should be a similar number of allocations between `Vec<&'static str>` and `String`, correct? Are you thinking comparisons and hashing is faster with String or something of that nature?

The allocations I was concerned with were the ones were we allocate a new string _for each_ protocol in the multiaddr. Constructing a new, single string will require one allocation. I don't think we can get rid of that one.

This should do the trick[^1] (note the leading `/` at `/tcp`:

```rust
let stack = addr
	.iter()
	.map(|p| match p {
		Protocol::Tcp(_) => "/tcp",
	 	// etc
	})
	.collect::<String>();
```

[^1]: Looking at the `collect` impl, we will still allocate multiple times depending on the length of the multiaddr though: https://doc.rust-lang.org/src/alloc/string.rs.html#2073

thomaseizinger:
> With the change in key the metrics look slightly different now
> 
> ```
> libp2p_swarm_connections_incoming_total{ip4/tcp} 1
> libp2p_identify_listen_addresses_total{ip4/tcp} 5
> ```

Yep, that is almost what we want (see https://github.com/libp2p/rust-libp2p/issues/2758)! :)

A leading `/` would be great and instead of creating a new label for each one, we should have one label `address_stack` with multiple values!

misc/metrics/src/protocol_stack.rs
thomaseizinger: Oh I wasn't aware of that PR, sorry about that!

We'd still probably want a test that shows that we construct the correct stack as it is described in the issue. It makes review easier because I can see the output of the function without having to run or brain-compile it :)

misc/metrics/src/protocol_stack.rs
thomaseizinger: Don't mind the above, I didn't realize you were already creating the function upstream. All of this was so that we don't _need_ to create it upstream but can instead iterate on it here first.

John-LittleBearLabs:
> > With the change in key the metrics look slightly different now
> > ```
> > libp2p_swarm_connections_incoming_total{ip4/tcp} 1
> > libp2p_identify_listen_addresses_total{ip4/tcp} 5
> > ```
> 
> Yep, that is almost what we want (see #2758)! :)

Yeah that's where I got the names. The _total suffix appears to come from the `Family`. Not sure what to do about that.
 
> A leading `/` would be great and

That should be easy enough. 

> instead of creating a new label for each one, we should have one label `address_stack` with multiple values!

Not sure I follow exactly. I'm guessing instead of  
````
libp2p_identify_listen_addresses_total{/ip4/tcp} 5
libp2p_identify_listen_addresses_total{/ip4/udp} 1
````
We'd have something like
````
libp2p_identify_listen_addresses /ip4/tcp=5 /ip4/udp=1
````
? If so I'd have to look into how something like that would be done (glancing at prometheus-client does not make it obvious).
Or do you mean just have one metric to count both identify listen addresses and also swarm connections incoming summed together?

**Comment by thomaseizinger** [comment]
> > instead of creating a new label for each one, we should have one label `address_stack` with multiple values!
> 
> Not sure I follow exactly. I'm guessing instead of
> 
> ```
> libp2p_identify_listen_addresses_total{/ip4/tcp} 5
> libp2p_identify_listen_addresses_total{/ip4/udp} 1
> ```
> 
> We'd have something like
> 
> ```
> libp2p_identify_listen_addresses /ip4/tcp=5 /ip4/udp=1
> ```
> 
> ? If so I'd have to look into how something like that would be done (glancing at prometheus-client does not make it obvious). Or do you mean just have one metric to count both identify listen addresses and also swarm connections incoming summed together?

It is actually quite easy. Take a look at the example at the root of the prometheus-client docs: https://docs.rs/prometheus-client/0.18.0/prometheus_client/

If you define a custom `Labels` type, the `Encode` trait will render them as `key=value`, e.g. in our case `address_stack=/ip4/tcp/` etc.
If you then ingest these metrics into something like grafana, it will recognise the label key (`address_stack`) and allows you perform queries on it, visualize it etc. For example, you could see the individual value for each value of `address_stack` in a graph or group them by particular values.

Under the hood, prometheus will actually create a separate metric per unique label combination. Labels are just a kind of syntax-sugar on top to allow grouping / filter etc. Plus, (I think) you automatically get a `_total` metric that represents the sum of all of them.

**Comment by John-LittleBearLabs** [comment]
> If you define a custom `Labels` type, the `Encode` trait will render them as `key=value`, e.g. in our case `address_stack=/ip4/tcp/` etc. 

Not 100% sure this is what you meant, but here's where I"m at now (this output is from the metrics example).

````
$curl -s http://127.0.0.1:36827/metrics | grep -E 'identify_listen_addresses|swarm_connections_incoming'
# HELP libp2p_identify_listen_addresses Number of listen addresses for remote peer per protocol stack.
# TYPE libp2p_identify_listen_addresses counter
libp2p_identify_listen_addresses_total{address_stack="/ip4/tcp"} 5
# HELP libp2p_swarm_connections_incoming Number of incoming connections per address stack.
# TYPE libp2p_swarm_connections_incoming counter
libp2p_swarm_connections_incoming_total{address_stack="/ip4/tcp"} 1
# HELP libp2p_swarm_connections_incoming_error Number of incoming connection errors.
# TYPE libp2p_swarm_connections_incoming_error counter

lbl@chomp:~lbl/code/rust-multiaddr/tests
$curl -s http://127.0.0.1:35883/metrics | grep -E 'identify_listen_addresses|swarm_connections_incoming'
# HELP libp2p_identify_listen_addresses Number of listen addresses for remote peer per protocol stack.
# TYPE libp2p_identify_listen_addresses counter
libp2p_identify_listen_addresses_total{address_stack="/ip4/tcp"} 6
# HELP libp2p_swarm_connections_incoming Number of incoming connections per address stack.
# TYPE libp2p_swarm_connections_incoming counter
# HELP libp2p_swarm_connections_incoming_error Number of incoming connection errors.
# TYPE libp2p_swarm_connections_incoming_error counter
````

**Comment by thomaseizinger** [comment]
> > If you define a custom `Labels` type, the `Encode` trait will render them as `key=value`, e.g. in our case `address_stack=/ip4/tcp/` etc.
> 
> Not 100% sure this is what you meant, but here's where I"m at now (this output is from the metrics example).
> 
> ```
> $curl -s http://127.0.0.1:36827/metrics | grep -E 'identify_listen_addresses|swarm_connections_incoming'
> # HELP libp2p_identify_listen_addresses Number of listen addresses for remote peer per protocol stack.
> # TYPE libp2p_identify_listen_addresses counter
> libp2p_identify_listen_addresses_total{address_stack="/ip4/tcp"} 5
> # HELP libp2p_swarm_connections_incoming Number of incoming connections per address stack.
> # TYPE libp2p_swarm_connections_incoming counter
> libp2p_swarm_connections_incoming_total{address_stack="/ip4/tcp"} 1
> # HELP libp2p_swarm_connections_incoming_error Number of incoming connection errors.
> # TYPE libp2p_swarm_connections_incoming_error counter
> 
> lbl@chomp:~lbl/code/rust-multiaddr/tests
> $curl -s http://127.0.0.1:35883/metrics | grep -E 'identify_listen_addresses|swarm_connections_incoming'
> # HELP libp2p_identify_listen_addresses Number of listen addresses for remote peer per protocol stack.
> # TYPE libp2p_identify_listen_addresses counter
> libp2p_identify_listen_addresses_total{address_stack="/ip4/tcp"} 6
> # HELP libp2p_swarm_connections_incoming Number of incoming connections per address stack.
> # TYPE libp2p_swarm_connections_incoming counter
> # HELP libp2p_swarm_connections_incoming_error Number of incoming connection errors.
> # TYPE libp2p_swarm_connections_incoming_error counter
> ```

Yep that looks correct, thanks! :blush:

File: misc/metrics/src/protocol_stack.rs [hunk 12]
```diff
@@ -0,0 +1,89 @@
...
+    fn protocol_stack(&self) -> String {
+        // Has potential to allocate multiple times, but this line expresses the intent here.
+        //  std::iter::once("/").chain(ma.iter().map(tag).intersperse("/")).collect()
+        let len = self.iter().fold(0, |acc, proto| acc + tag(proto).len() + 1);
+        let mut result = String::with_capacity(len);
+        for proto_tag in self.iter().map(tag) {
+            result.push('/');
+            result.push_str(proto_tag);
+        }
+        result
```
thomaseizinger: Nice work!

misc/metrics/src/protocol_stack.rs
```diff
@@ -0,0 +1,89 @@
...
+impl From<Multiaddr> for Label {
+    fn from(ma: Multiaddr) -> Self {
+       Self:: from(&ma)
+    }
+}
+impl From<&Multiaddr> for Label {
+     fn from(ma: &Multiaddr) -> Self {
+        Self { address_stack: ma.protocol_stack() }
+    }
+}
```
thomaseizinger: I don't find these to be particularly intuitive. Just because something can be expressed with `From` doesn't mean it should use `From`.

I'd be in favor of just saying:

```rust
impl Label {
	pub fn for_multi_address(addr: &Multiaddr) -> Self {}
}
```

This makes it a lot more clear on the call-site, what is happening. `into()` is very in-transparent. One of the best uses of `From` and `Into` (as with all traits actually) is to create APIs that are generic over these capabilities, like:

```rust
fn do_something(value: impl Into<Label>) {
	let label = value.into();

	// Do something with `Label`
}
```

If you have to call `into()` yourself - as we do in `identify.rs` - the value of using a standardized trait to do your conversion is practically zero. I'd actually claim it is harmful because it obfuscates what is happening which is why I'd prefer a more concrete function that conveys more intent.

Hope that makes sense :)

misc/metrics/src/protocol_stack.rs
```diff
@@ -0,0 +1,89 @@
...
+
+    #[test]
+    fn ip6_tcp_wss_p2p() {
+        let ma = Multiaddr::try_from("/ip6/2001:8a0:7ac5:4201:3ac9:86ff:fe31:7095/tcp/8000/wss/p2p/QmcgpsyWgH8Y8ajJz1Cu72KnS5uo2Aa2LpzU7kinSupNKC").expect("testbad");
+        let actual : Label = ma.into();
+        assert_eq!(actual.address_stack,"/ip6/tcp/wss/p2p");
+        let mut buf = Vec::new();
+        actual.encode(&mut buf).expect("encode failed");
+        let actual = String::from_utf8(buf).expect("invalid utf-8?");
+        assert_eq!(actual, "address_stack=\"/ip6/tcp/wss/p2p\"");
```
thomaseizinger: You can use raw strings (`r#""#`) to avoid the escaping.

misc/metrics/src/protocol_stack.rs
```diff
@@ -0,0 +1,89 @@
...
+            result.push_str(proto_tag);
+        }
+        result
+    }
+}
+
+#[derive(Encode, Hash, Clone, Eq, PartialEq)]
+pub struct Label {
+    address_stack: String
+}
```
thomaseizinger: Given that this is our new public API, can it go at the top of the file please?

thomaseizinger (commented):
A few more comments but otherwise LGTM.

This is almost ready to merge, thank you! :)

thomaseizinger (approved):
LGTM!

I'll let @mxinden also have a look as he wrote the original issue!

thomaseizinger:
Actually, can you add a changelog entry please? :)

misc/metrics/CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 # 0.10.0 [unreleased]
 
+- Added `protocol_stack` metrics.
```
thomaseizinger: All entries apart from dependency bumps should also mention the PR they come from. See some of the other examples in this file :)

Plus, Max just pushed out 0.49, so this will need an update to the master branch and an according version bump (patch-level) please!

misc/metrics/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# 0.10.1
```
thomaseizinger:
```suggestion
# 0.10.1 [unreleased]
```

misc/metrics/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
...
+
+- Added `protocol_stack` metrics. See [PR 2982].
```
thomaseizinger: Missing link for this reference.

thomaseizinger (approved):
Nice work!

@mxinden Thoughts from your end?

misc/metrics/examples/metrics/main.rs
```diff
@@ -109,8 +113,22 @@ fn main() -> Result<(), Box<dyn Error>> {
...
-#[derive(NetworkBehaviour, Default)]
+#[derive(NetworkBehaviour)]
 struct Behaviour {
+    identify: identify::Behaviour,
     keep_alive: keep_alive::Behaviour,
     ping: ping::Behaviour,
 }
+
+impl Behaviour {
+    fn new(local_key: libp2p::identity::Keypair) -> Self {
```
mxinden: How about passing the public key only here?

misc/metrics/src/protocol_stack.rs
```diff
@@ -0,0 +1,86 @@
+use libp2p_core::multiaddr::{Multiaddr, Protocol};
+use prometheus_client::encoding::text::Encode;
+
+#[derive(Encode, Hash, Clone, Eq, PartialEq)]
+pub struct Label {
+    address_stack: String,
+}
+impl Label {
+    pub fn for_multi_address(ma: &Multiaddr) -> Self {
```
mxinden:
```suggestion
    pub fn from_multiaddress(ma: &Multiaddr) -> Self {
```

I don't think using `for_` is a convention. Am I missing something?

misc/metrics/src/protocol_stack.rs
```diff
@@ -0,0 +1,86 @@
+use libp2p_core::multiaddr::{Multiaddr, Protocol};
+use prometheus_client::encoding::text::Encode;
+
+#[derive(Encode, Hash, Clone, Eq, PartialEq)]
+pub struct Label {
+    address_stack: String,
+}
+impl Label {
```
mxinden: Why not implement `From` for `Label`?

``` rust
impl From<&Multiaddr> for Label {
```

misc/metrics/src/protocol_stack.rs
```diff
@@ -0,0 +1,86 @@
+use libp2p_core::multiaddr::{Multiaddr, Protocol};
+use prometheus_client::encoding::text::Encode;
+
+#[derive(Encode, Hash, Clone, Eq, PartialEq)]
+pub struct Label {
```
mxinden:
```suggestion
pub struct Labels {
```
Nit pick, a `struct` is a set of labels.

misc/metrics/src/swarm.rs
```diff
@@ -156,8 +157,10 @@ impl<TBvEv, THandleErr> super::Recorder<libp2p_swarm::SwarmEvent<TBvEv, THandleE
                     })
                     .inc();
             }
-            libp2p_swarm::SwarmEvent::IncomingConnection { .. } => {
-                self.connections_incoming.inc();
+            libp2p_swarm::SwarmEvent::IncomingConnection { send_back_addr, .. } => {
```
mxinden, line 162: How about doing the same for all other `SwarmEvent`s? E.g. `ConnectionEstablished`?

misc/metrics/src/protocol_stack.rs
```diff
@@ -0,0 +1,86 @@
...
+    }
+}
+
+//TODO: remove this trait and tag() and replace with calls to the upstream method
+//  once that lands : https://github.com/multiformats/rust-multiaddr/pull/60
+// In the meantime there is no _ case in the match so one can easily detect mismatch in supported
+//  protocols when dependency version changes.
+trait MultiaddrExt {
+    fn protocol_stack(&self) -> String;
+}
```
mxinden: I still need to cut a new release of `multiaddr`. Sorry for the delay.

mxinden (dismissed):
This is already proven useful, see https://github.com/libp2p/rust-libp2p/pull/2289#issuecomment-1299304244.

Couple of comments. How about expanding this beyond the two metrics you currently improve?

misc/metrics/src/protocol_stack.rs
John-LittleBearLabs: No, you're right that that is more idiomatic for rust and so it makes more sense. I'll be implementing From as mentioned below.

I think I landed on "for" without really thinking about it just because if I were describing the result/purpose of this method in English prose I would've likely used "for", but I'm not writing English prose here - I'm writing rust.

misc/metrics/src/swarm.rs
John-LittleBearLabs: I added ConnectionEstablished, but do you really want to do _all_ of them, even things like `connected_to_banned_peer` ?

````
$for u in http://127.0.0.1:37857/metrics http://127.0.0.1:34293/metrics ; do curl -s "${u}" | grep tcp ; done
libp2p_identify_listen_addresses_total{address_stack="/ip4/tcp"} 5
libp2p_swarm_connections_incoming_total{address_stack="/ip4/tcp"} 1
libp2p_swarm_connections_established_total{role="Listener",proto_stack="/ip4/tcp"} 1
libp2p_identify_listen_addresses_total{address_stack="/ip4/tcp"} 6
libp2p_swarm_connections_established_total{role="Dialer",proto_stack="/ip4/tcp"} 1
````

File: misc/metrics/src/protocol_stack.rs (hunk 21 above)
Comment (thomaseizinger) [reply misc/metrics/src/protocol_stack.rs]: I think we had this at some point and I recommended to move away from it. Unless we are actually using `From` as an abstraction somewhere, using a generic function has downsides in terms of clarity. It also encourages the use of `.into` at callsites which makes it hard to see, what this is actually being converted into.

File: misc/metrics/src/swarm.rs (hunk 23 above)
Comment (John-LittleBearLabs) [reply misc/metrics/src/swarm.rs:162]: Added what I easily can, but these are triggered by events that don't seem to carry the an obvious multiaddr to key off of in the event struct itself:

dial_attempt
outgoing_connection_error
listener_error

These copy-pasted outputs are from the example running in my branch. Using exclusively tcp multiaddrs, hence the grep to find metrics keyed off protocol.
````
$for u in http://127.0.0.1:{37961,33459}/metrics  ; do curl -s "${u}" | grep tcp ; done
libp2p_identify_listen_addresses_total{address_stack="/ip4/tcp"} 5
libp2p_swarm_connections_incoming_total{address_stack="/ip4/tcp"} 1
libp2p_swarm_new_listen_addr_total{address_stack="/ip4/tcp"} 5
libp2p_swarm_connections_established_total{role="Listener",proto_stack="/ip4/tcp"} 1
libp2p_identify_listen_addresses_total{address_stack="/ip4/tcp"} 5
libp2p_swarm_new_listen_addr_total{address_stack="/ip4/tcp"} 5
libp2p_swarm_connections_established_total{role="Dialer",proto_stack="/ip4/tcp"} 1
````

File: misc/metrics/src/protocol_stack.rs (hunk 21 above)
Comment (John-LittleBearLabs) [reply misc/metrics/src/protocol_stack.rs]: Ah yes, that sounds familiar. I probably resolved & hid the comment when I committed the recommended change, but it's probably still above somewhere.

Personally I think I prefer the From approach in this case - since the struct has such an incredibly narrow usage I don't think there's too much confusion about what `into` is doing. But of course I'll go with whatever the consensus is.

Comment (thomaseizinger) [reply misc/metrics/src/protocol_stack.rs]: `From` would be nice if we could change the signature of `get_or_create` to accept an `impl Into<S>` so you don't have to call `.into`.

Unfortunately we can't do that without requiring ownership on _every_callsite which would require allocations even though we only need ownership inside the `get_or_create` function for the first call. Meaning a lot of these allocations would be wasted.

I'd be in favor of not obfuscating what we are converting here:

1. Using `.into` makes it hard to find all usages of `Labels`. Currently, I have to delete the `From` impl and see where the compiler fails.
2. I find this:
```rust
                    self.listen_addresses
                        .get_or_create(&protocol_stack::Labels::new(&listen_addr))
                        .inc();
```
much clearer to read than this:
```rust
                    self.listen_addresses
                        .get_or_create(&listen_addr.into())
                        .inc();
```
The former I can glance over whereas with the latter, my brain stops and is like: "Hang on, `into()` what?". And I have to recall the function signature of `get_or_create`, check the type of `listen_addr` and search for `From` implementations to understand what is happening here.

3. If we ever need more data for the labels, `From` would need to be using a tuple at which point we need a constructor anyway.
4. Out of principle, I think we should not be depending on abstractions (`From`), if we don't use them. Traits are only useful in generic code, so unless we can make a function that uses `T: Into` where we can pass a `&Multiaddr`, I think this doesn't qualify.

misc/metrics/src/swarm.rs
thomaseizinger: > Added what I easily can, but these are triggered by events that don't seem to carry the an obvious multiaddr to key off of in the event struct itself:
> 
> - dial_attempt
> - outgoing_connection_error
> - listener_error

Yeah that is fine.

> These copy-pasted outputs are from the example running in my branch. Using exclusively tcp multiaddrs, hence the grep to find metrics keyed off protocol.
> 
> ```
> $for u in http://127.0.0.1:{37961,33459}/metrics  ; do curl -s "${u}" | grep tcp ; done
> libp2p_identify_listen_addresses_total{address_stack="/ip4/tcp"} 5
> libp2p_swarm_connections_incoming_total{address_stack="/ip4/tcp"} 1
> libp2p_swarm_new_listen_addr_total{address_stack="/ip4/tcp"} 5
> libp2p_swarm_connections_established_total{role="Listener",proto_stack="/ip4/tcp"} 1
> libp2p_identify_listen_addresses_total{address_stack="/ip4/tcp"} 5
> libp2p_swarm_new_listen_addr_total{address_stack="/ip4/tcp"} 5
> libp2p_swarm_connections_established_total{role="Dialer",proto_stack="/ip4/tcp"} 1
> ```

We don't seem to be consistent in the naming. I'd actually prefer the "protocol"-inspired naming. How about we just call this label `protocols`:

```bash
$ for u in http://127.0.0.1:{37961,33459}/metrics  ; do curl -s "${u}" | grep tcp ; done

libp2p_identify_listen_addresses_total{protocols="/ip4/tcp"} 5
libp2p_swarm_connections_incoming_total{protocols="/ip4/tcp"} 1
libp2p_swarm_new_listen_addr_total{protocols="/ip4/tcp"} 5
libp2p_swarm_connections_established_total{role="Listener",protocols="/ip4/tcp"} 1
libp2p_identify_listen_addresses_total{protocols="/ip4/tcp"} 5
libp2p_swarm_new_listen_addr_total{protocols="/ip4/tcp"} 5
libp2p_swarm_connections_established_total{role="Dialer",protocols="/ip4/tcp"} 1
```

John-LittleBearLabs: I agree, @thomaseizinger . Now have:

```bash
$for u in http://127.0.0.1:{39507,42089}/metrics  ; do curl -s "${u}" | grep tcp ; done
libp2p_identify_listen_addresses_total{protocols="/ip4/tcp"} 5
libp2p_swarm_connections_incoming_total{protocols="/ip4/tcp"} 1
libp2p_swarm_new_listen_addr_total{protocols="/ip4/tcp"} 5
libp2p_swarm_connections_established_total{role="Listener",protocols="/ip4/tcp"} 1
libp2p_identify_listen_addresses_total{protocols="/ip4/tcp"} 5
libp2p_swarm_new_listen_addr_total{protocols="/ip4/tcp"} 5
libp2p_swarm_connections_established_total{role="Dialer",protocols="/ip4/tcp"} 1
```

misc/metrics/CHANGELOG.md
```diff
@@ -16,6 +16,10 @@
 
 - Update to `libp2p-gossipsub` `v0.43.0`.
 
+- Added `protocol_stack` metrics. See [PR 2982].
```
thomaseizinger:
```suggestion
- Add `protocol_stack` metrics. See [PR 2982].
```

misc/metrics/src/protocol_stack.rs
```diff
@@ -0,0 +1,89 @@
+use libp2p_core::multiaddr::{Multiaddr, Protocol};
+use prometheus_client::encoding::text::Encode;
+
+#[derive(Encode, Hash, Clone, Eq, PartialEq)]
+pub struct Labels {
+    protocols: String,
+}
+impl Labels {
```
thomaseizinger: I am surprised the formatter doesn't put a newline here?

misc/metrics/src/protocol_stack.rs
```diff
@@ -0,0 +1,89 @@
...
+}
+
+//TODO: remove this trait and tag() and replace with calls to the upstream method
+//  once that lands : https://github.com/multiformats/rust-multiaddr/pull/60
+// In the meantime there is no _ case in the match so one can easily detect mismatch in supported
+//  protocols when dependency version changes.
+pub trait MultiaddrExt {
+    fn protocol_stack(&self) -> String;
+}
+impl MultiaddrExt for Multiaddr {
```
thomaseizinger: Same here.

misc/metrics/src/protocol_stack.rs
```diff
@@ -0,0 +1,89 @@
...
+        let mut result = String::with_capacity(len);
+        for proto_tag in self.iter().map(tag) {
+            result.push('/');
+            result.push_str(proto_tag);
+        }
+        result
+    }
+}
+
+pub fn tag(proto: Protocol) -> &'static str {
```
thomaseizinger:
```suggestion
fn tag(proto: Protocol) -> &'static str {
```

thomaseizinger (approved):
Lets merge it! Thanks for sticking to this and sorry for the back and forth in the review!

I think clippy still wants to have a word with you, otherwise, I am happy to merge this :)

thomaseizinger:
@mxinden Do you want to wait for https://github.com/multiformats/rust-multiaddr/pull/62 and upgrade in this PR right away?

misc/metrics/src/protocol_stack.rs
John-LittleBearLabs: Double checked and the formatter doesn't. :shrug: 

Not sure these two blocks are distinct & separate enough to conceptually justify a blank line, but if that's how folks are used to reading it I'll do the consistent thing so it doesn't stand out (as it does not deserve to stand out).

misc/metrics/src/protocol_stack.rs
John-LittleBearLabs: Added one.

mxinden:
> @mxinden Do you want to wait for [multiformats/rust-multiaddr#62](https://github.com/multiformats/rust-multiaddr/pull/62) and upgrade in this PR right away?

Yes. I don't think we are in a rush here and thus it's fine to wait.

mxinden (commented):
Thanks for the follow-ups here @John-LittleBearLabs.

Once https://github.com/multiformats/rust-multiaddr/pull/62 is merged, and released and used in this pull request, this is good to go from my end.

thomaseizinger:
I adjusted the PR title and description for a better commit message, plus did some minor tidy up in the code. Based on [this](https://github.com/libp2p/rust-libp2p/pull/2982#pullrequestreview-1172873819) comment, I am going ahead in merging this!

mxinden (approved):
Thanks! Good to go from my side.

mxinden:
> [Merge branch 'master' into addrstak_metric](https://github.com/libp2p/rust-libp2p/pull/2982/commits/7176f143ce612641bee5a1542950f41f821b26d0)

@thomaseizinger in which cases do we manually merge `master` and in which cases do we depend on mergify to merge `master`?

mxinden:
For those interested, this is now deployed on kademlia-exporter.max-inden.de and the graphs are updated accordingly.

https://kademlia-exporter.max-inden.de/d/Pfr0Fj6Mk/rust-libp2p?orgId=1&from=now-2d&to=now&var-data_source=Prometheus&var-instance=kademlia-exporter-ipfs:8080

thomaseizinger:
> > [Merge branch 'master' into addrstak_metric](https://github.com/libp2p/rust-libp2p/pull/2982/commits/7176f143ce612641bee5a1542950f41f821b26d0)
> 
> @thomaseizinger in which cases do we manually merge `master` and in which cases do we depend on mergify to merge `master`?

I am not sure I remember why I merged master in but mergify is only going to update the branch when the PR is already queued for merging.

It will only queue it once all checks are passing. One of those is the `send-it` label. All others are inferred from the branch protection settings. If you have a flaky CI check, mergify won't trigger another merge so maybe that is what happened here?

PR #3006, opened 2022-10-10
libp2p_swarm_derive: Add `crate` parameter to configure `libp2p` crate name

stormshield-pj50:
# Description

When importing libp2p with a use statement and a custom alias name, the code generated by the NetworkBehaviour derive macro does not use this alias and does not compile. This PR enhances the derive macro with a new parameter "crate" that can be used to specify the custom alias name.

## Open Questions
None

## Change checklist

### DIFF
--- a/swarm-derive/src/lib.rs
+++ b/swarm-derive/src/lib.rs
@@ -46,21 +46,43 @@ fn build(ast: &DeriveInput) -> TokenStream {
 fn build_struct(ast: &DeriveInput, data_struct: &DataStruct) -> TokenStream {
     let name = &ast.ident;
     let (_, ty_generics, where_clause) = ast.generics.split_for_impl();
-    let multiaddr = quote! {::libp2p::core::Multiaddr};
-    let trait_to_impl = quote! {::libp2p::swarm::NetworkBehaviour};
-    let either_ident = quote! {::libp2p::core::either::EitherOutput};
-    let network_behaviour_action = quote! {::libp2p::swarm::NetworkBehaviourAction};
-    let into_connection_handler = quote! {::libp2p::swarm::IntoConnectionHandler};
-    let connection_handler = quote! {::libp2p::swarm::ConnectionHandler};
-    let into_proto_select_ident = quote! {::libp2p::swarm::IntoConnectionHandlerSelect};
-    let peer_id = quote! {::libp2p::core::PeerId};
-    let connection_id = quote! {::libp2p::core::connection::ConnectionId};
-    let dial_errors = quote! {Option<&Vec<::libp2p::core::Multiaddr>>};
-    let connected_point = quote! {::libp2p::core::ConnectedPoint};
-    let listener_id = quote! {::libp2p::core::transport::ListenerId};
-    let dial_error = quote! {::libp2p::swarm::DialError};
 
-    let poll_parameters = quote! {::libp2p::swarm::PollParameters};
+    let user_provided_libp2p_prefix = ast
+        .attrs
+        .iter()
+        .filter_map(get_meta_items)
+        .flatten()
+        .filter_map(|meta_item| {
+            if let syn::NestedMeta::Meta(syn::Meta::NameValue(ref m)) = meta_item {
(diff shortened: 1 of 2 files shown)


swarm-derive/tests/test.rs
```diff
@@ -133,6 +133,23 @@ fn custom_event() {
     }
 }
 
+#[test]
+fn custom_prefix() {
+    use libp2p as mylibp2p;
+    #[allow(dead_code)]
+    #[derive(NetworkBehaviour)]
+    #[behaviour(libp2p_prefix = "mylibp2p")]
```
thomaseizinger: Serde calls this parameter `crate`: https://serde.rs/container-attrs.html#crate

I'd be in favor of following that here too.

stormshield-pj50: OK, done

thomaseizinger (approved):
LGTM, thanks!

swarm-derive/src/lib.rs
```diff
@@ -46,21 +46,43 @@ fn build(ast: &DeriveInput) -> TokenStream {
...
+            if let syn::NestedMeta::Meta(syn::Meta::NameValue(ref m)) = meta_item {
+                if m.path.is_ident("crate") {
+                    if let syn::Lit::Str(ref s) = m.lit {
+                        return Some(syn::parse_str::<syn::TypePath>(&s.value()).unwrap());
+                    }
+                }
+            }
+            None
+        })
+        .next();
```
mxinden, line 65: This is mostly the same as the logic for `out_event`, right? What do you think of moving it into a function called in both places?

mxinden (approved):
One suggestion. Otherwise looks good to me.

swarm-derive/tests/test.rs
```diff
@@ -133,6 +133,23 @@ fn custom_event() {
...
+    struct Foo {
+        ping: ping::Behaviour,
+        identify: identify::Behaviour,
+    }
+
+    #[allow(dead_code)]
+    fn foo() {
+        require_net_behaviour::<Foo>();
+    }
+}
```
elenaf9, line 151: This test also already passes with current master (if you remove line 141). 
Correct me if I am wrong, but I believe the current `::libp2p` prefix imports libp2p from the crate level, which means that aliases in a module should not be an issue? @mxinden?

I think it is only a problem if the crate is imported in the `Cargo.toml` with a different name.
`serde` on the other hand states for their `crate` attribute that: 
> This is normally only applicable when invoking re-exported Serde derives from a public macro in a different crate.

I am not against adding this parameter, but could we add a test that fails on current master and would be fixed with this? 🙂

thomaseizinger: As far as I understand, the issue is not clashing symbols but renaming the crate in your manifest.

To test that, we'd have to make a separate test crate with a renamed import.

This make me wonder, should we perhaps just require certain symbols to be present for the macro to work?

That would make https://github.com/libp2p/rust-libp2p/pull/3023 easier to implement.

Or perhaps we could make a seperate "derive" prelude? Like `libp2p::swarm_derive_prelude` that includes all the required symbols. Then we can add a configuration knob here that allows you to change where that prelude is coming from.

thomaseizinger: > This make me wonder, should we perhaps just require certain symbols to be present for the macro to work?
> 
> That would make #3023 easier to implement.
> 
> Or perhaps we could make a seperate "derive" prelude? Like `libp2p::swarm_derive_prelude` that includes all the required symbols. Then we can add a configuration knob here that allows you to change where that prelude is coming from.

Actually, don't worry about it, I already have a solution to that.

thomaseizinger:
If we end up shipping https://github.com/libp2p/rust-libp2p/pull/3055, then this PR is obsolete as the `prelude` option allows you to achieve the same thing via:

```rust
#[behaviour(prelude = "mylibp2p::swarm::derive_prelude")
```

thomaseizinger:
Closing in favor of https://github.com/libp2p/rust-libp2p/pull/3055 which is about to be merged.

PR #3011, opened 2022-10-11
swarm/behaviour: Replace `inject_*` with `on_event`

### LINKED ISSUE #2832: swarm/behaviour: Inject events via single method

jxs:
# Description

Superseeds #2867.

Fixes #2832.

## Change checklist


swarm/src/behaviour.rs
```diff
@@ -152,97 +152,224 @@ pub trait NetworkBehaviour: 'static {
...
+    /// Informs the behaviour about an event generated by the [`ConnectionHandler`] dedicated to the
+    /// peer identified by `peer_id`. for the behaviour.
+    ///
+    /// The [`PeerId`] is guaranteed to be in a connected state. In other words,
+    /// [`FromSwarm::ConnectionEstablished`] has previously been received with this [`PeerId`].
+    fn on_connection_handler_event(
+        &mut self,
+        _peer_id: PeerId,
+        _connection_id: ConnectionId,
+        _event: crate::THandlerOutEvent<Self>,
```
jxs: @thomaseizinger @mxinden regarding https://github.com/libp2p/rust-libp2p/pull/2867/files#r973871955,  doc replaces the alias with the full definition, do you still want to change it?

image

swarm/src/behaviour.rs
```diff
@@ -721,3 +848,183 @@ impl Default for CloseConnection {
         CloseConnection::All
     }
 }
+
+pub enum FromSwarm<'a, Handler: IntoConnectionHandler> {
```
jxs, line 856: @thomaseizinger @elenaf9 as per https://github.com/libp2p/rust-libp2p/pull/2867/files#r974184169, renamed `InEvent` to `FromSwarm`

swarm/src/handler/either.rs
```diff
@@ -29,6 +29,8 @@ use libp2p_core::upgrade::{EitherUpgrade, UpgradeError};
 use libp2p_core::{ConnectedPoint, Multiaddr, PeerId};
 use std::task::{Context, Poll};
 
+/// Auxiliary type to allow implementing [`IntoConnectionHandler`]. As [`IntoConnectionHandler`] is
```
jxs, line 32: as per https://github.com/libp2p/rust-libp2p/pull/2867/files#r973874109 added a comment explaining existance of the type

swarm/tests/public_api.rs
```diff
@@ -0,0 +1,64 @@
+use libp2p::core::transport::ListenerId;
+use libp2p::swarm::behaviour::FromSwarm;
+use libp2p::swarm::handler::DummyConnectionHandler;
+
+#[test]
+// test to break compilation everytime a variant changes,
+// forcing us to revisit each implementation
+fn swarm_event_variants() {
```
jxs: @thomaseizinger following your suggestion on https://github.com/libp2p/rust-libp2p/pull/2867/files#r965651079 moved the test here

jxs (commented):
**Question**:
- In the `NetworkBehaviour` doc, some `inject_*` (now on_*) refer **Indicates to the behaviour** others **informs the behaviour** should we have them all on the same wording, for consistency? 

- updated the `autonat` behaviour, if that looks ok to you will follow to the other protocols

mxinden:
> * In the `NetworkBehaviour` doc, some `inject_*` (now on_*) refer **Indicates to the behaviour** others **informs the behaviour** should we have them all on the same wording, for consistency?

Consistent wording sounds good to me. Thanks for catching that. I suggest only changing the wording on the new APIs.

swarm/src/behaviour.rs
mxinden: No strong opinion here. Intuition tells me that the full definition is cleaner and easier to understand when reading the code itself. Will let @thomaseizinger decide.

protocols/autonat/src/behaviour.rs
```diff
@@ -351,77 +348,155 @@ impl NetworkBehaviour for Behaviour {
...
+        conn: &ConnectionId,
+        endpoint: &ConnectedPoint,
+        failed_addresses: Option<&Vec<Multiaddr>>,
+        other_established: usize,
+    ) {
+        let failed_addresses = failed_addresses
+            .map(|v| v.as_slice())
+            .unwrap_or_else(|| &[]);
+        self.on_connection_established(*peer, *conn, endpoint, failed_addresses, other_established)
+    }
```
mxinden: `Swarm` still calls the old `NetworkBehaviour::inject_*` methods. The `NetworkBehaviour::inject_*` method default implementations delegate to the `NetworkBehaviour::on_*` methods.

I suggest that this pull request ports `libp2p-autonat` to implement `NetworkBehaviour::on_*` only, and no more implement the `NetworkBehaviour::inject_*` methods.  For now `Swarm` will be calling `NetworkBehaviour::inject_*` whose default implementations would then call the `libp2p-autonat` `<Behaviour as NetworkBehaviour>::on_event` methods. Thus there is no need for `Behaviour` to implement `NetworkBehaviour::inject_*`.

Long term we want to (a) get rid of the `NetworkBehaviour::inject_*` methods and (b) call the `NetworkBehaviour::on_*` methods in `Swarm` directly.

Does the above make sense @jxs?

jxs: yeah ofc! :) missed that, thanks Max. updated

jxs:
> > * In the `NetworkBehaviour` doc, some `inject_*` (now on_*) refer **Indicates to the behaviour** others **informs the behaviour** should we have them all on the same wording, for consistency?
> 
> Consistent wording sounds good to me. Thanks for catching that. I suggest only changing the wording on the new APIs.

sorry, forgot to ask which one should prevail then

swarm/src/behaviour.rs
thomaseizinger: I am in favor of the full definition unless we make the type alias public :)

thomaseizinger:
> > > * In the `NetworkBehaviour` doc, some `inject_*` (now on_*) refer **Indicates to the behaviour** others **informs the behaviour** should we have them all on the same wording, for consistency?
> > 
> > 
> > Consistent wording sounds good to me. Thanks for catching that. I suggest only changing the wording on the new APIs.
> 
> sorry, forgot to ask which one should prevail then

"inform" sounds better to me but I don't have a strong opinion here.

protocols/autonat/src/behaviour.rs
```diff
@@ -478,35 +479,74 @@ impl NetworkBehaviour for Behaviour {
...
-    fn inject_listener_closed(&mut self, id: ListenerId, reason: Result<(), &std::io::Error>) {
-        self.inner.inject_listener_closed(id, reason)
+    fn on_swarm_event(&mut self, event: FromSwarm<Self::ConnectionHandler>) {
+        match event {
+            FromSwarm::ConnectionEstablished {
+                peer_id,
+                connection_id,
+                endpoint,
+                failed_addresses,
+                other_established,
```
thomaseizinger: We might want to consider introduced structs for each enum variant so we can just pass along the struct here instead of each field.

For example:

```rust
enum FromSwarm {
	ConnectionEstablished(ConnectionEstablished)
}

struct ConnectionEstablished {
	peer_id: PeerId,
	// etc
}
```

Thoughts?

swarm/src/behaviour.rs
```diff
@@ -721,3 +848,183 @@ impl Default for CloseConnection {
...
+    /// Indicates to the behaviour that a multiaddr we were listening on has expired,
+    /// which means that we are no longer listening on it.
+    ExpiredListenAddr {
+        listener_id: ListenerId,
+        addr: &'a Multiaddr,
+    },
+    /// A listener experienced an error.
+    ListenerError {
+        listener_id: ListenerId,
+        err: &'a (dyn std::error::Error + 'static),
```
thomaseizinger: Why are all these here references? An `Error` is not clonable typically so this API is not as useful as it could be.

protocols/autonat/src/behaviour.rs
jxs: yeah, makes sense to me also, wdyt @mxinden?

swarm/src/behaviour.rs
jxs: I think it was to maintain the previous function signatures, so you suggest removing the references in `FromSwarm` Thomas?  
But won't that imply a breaking change in `inject_listener_error` for example? It's expecting `&'a (dyn std::error::Error + 'static)`

swarm/src/behaviour.rs
jxs: yeah agreed it's cleaner and easier to understand, updated!  
Btw Thomas, even if the type alias is not public, [Rust seems to be able to point to the aliased one](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=d67bb8a886212da10623213eeabc3f80) was there another reason you suggest the type alias be public?

protocols/autonat/src/behaviour.rs
mxinden: Works for me. If I am not mistaken one can pattern match these within the `match` arms and thus there is no downside for users that don't delegate to specific methods, right? Something along the lines of:

``` rust
match event {
  FromSwarm::ConnectionEstablished(ConnectionEstablished { peer_id, .. }) => { todo!() },
```

swarm/src/behaviour.rs
mxinden: > Why are all these here references? An `Error` is not clonable typically so this API is not as useful as it could be.

This is due to use passing this error both to the root `NetworkBehaviour` **and** to the user as a `SwarmEvent`. I don't see  a way how to change that.

https://github.com/libp2p/rust-libp2p/blob/3371d7ceab242440216ae9ab99829631fa418f3b/swarm/src/lib.rs#L925-L928

protocols/autonat/src/behaviour.rs
thomaseizinger: Yep that is possible but a bit more verbose.

swarm/src/behaviour.rs
thomaseizinger: Oh I didn't know we need it in two places. Don't mind me then :)

rkuhn: This is one of the worst parts of the Rust ecosystem: the (missing) properties of the Error trait. Since there is so little one can do with it, and since we do have a much more precise type than `dyn Error` available, may I propose to just put it into an `Arc` and thus make it fully accessible and cloneable?

thomaseizinger: I wonder if we even need to pass it along here? A `NetworkBehaviour` has no control over the listeners. I don't understand why it would be interested in knowing that there is a new one etc.

What is it going to do with the error of a failed listener? Those errors aren't even fatal, the listener will keep going afterwards.

I did a quick audit of the codebase and the only callbacks we actually use are `inject_new_listen_addr` and `inject_expired_listen_addr`. All other listener-based ones are unused. We could also remove them and only tell the `NetworkBehaviour`s about new addresses that we are listening on / no longer listening on.

The original PR that introduced this was #2011. Unfortunately, it doesn't mention a motivation for why it was done. @mxinden Do you remember?

thomaseizinger: I had a quick [browse](https://github.com/search?l=&q=%22fn+inject_new_listener%22+-filename%3Atests.rs+language%3ARust&type=code) around GitHub and all usages I could find are either delegating to inner implementations, are in test files or empty.

Even the usage in [`ipfs-embed`](https://github.com/ipfs-rust/ipfs-embed/blob/87ac5acc1e8c8a0dad9e74560fdfea22e02ba98d/src/net/peers.rs#L804-L808) (which I believe is the source for the contribution), seems a bit weird to me: 

- Metrics for general stuff like listeners should probably be collected outside of the behaviour to make sure it happens in a 
canonical place and not in multiple places. What makes the `Peers` behaviour priviledged to collect stats on the number of listeners compared to any other behaviour?
- Logging could happen outside too and would probably lead to less duplication: What if every behaviour starts to log these? If we want these logs, having them [here](https://github.com/libp2p/rust-libp2p/blob/bdf9209824e364bb4e4047f7a32de45479d997e8/swarm/src/lib.rs#L325-L334) would probably be better.
- I am not sure I understand the value of wrapping and returning the event. The information is already available to the caller by means of polling the swarm itself. Exposing it again through a stream seems like duplicated work.

With these learnings in mind, my suggestion would be to remove these callbacks and thus get of the problem of errors that can only be passed around by reference. My strongest argument for this is that behaviours don't have control over listeners, so they can't take corrective action in case of an error. The caller of `Swarm` is responsible for the listener lifecycle and the error is already returned to them.

protocols/autonat/src/behaviour.rs
jxs: but if we are using structs we should define getters and setters to not expose its inner fields right? and therefore struct fields will not be able to be destructured on `match`. Would that be a problem @mxinden?

thomaseizinger: > but if we are using structs we should define getters and setters to not expose its inner fields right?

Not necessarily. Making fields of a struct private is useful if e.g. there are invariants we need to maintain in how they can be modified. In our case, these structs are just "bags" of data so the fields can easily be public IMO. Really what we would need is [enum-variant types](https://github.com/rust-lang/lang-team/issues/122).

swarm/src/behaviour.rs
jxs: makes sense, should I open a new issue with it? wdyt @mxinden?

mxinden: > The original PR that introduced this was #2011. Unfortunately, it doesn't mention a motivation for why it was done. @mxinden Do you remember?

I don't unfortunately.

> makes sense, should I open a new issue with it? wdyt @mxinden?

Yes. I think that is the right way forward. In my eyes removing the methods is orthogonal to this pull request.

rkuhn: The only important notion from the `ipfs-embed` perspective is that errors like “port already taken” need to be bubbled up so that they can be presented to the software admin — which is the case via the result of `Swarm::listen_on()`.

jxs: opened https://github.com/libp2p/rust-libp2p/issues/3040

protocols/autonat/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+# 0.8.1 - [unreleased]
+
+- replace `Behaviour`'s NetworkBehaviour implemention `inject_*` methods with the new `on_*` methods.
+  See [PR 3011].
```
thomaseizinger:
```suggestion
- Replace `Behaviour`'s NetworkBehaviour implemention `inject_*` methods with the new `on_*` methods.
  See [PR 3011].
```

protocols/autonat/src/behaviour.rs
```diff
@@ -298,43 +302,44 @@ impl Behaviour {
...
     }
-}
-
-impl NetworkBehaviour for Behaviour {
-    type ConnectionHandler = <RequestResponse<AutoNatCodec> as NetworkBehaviour>::ConnectionHandler;
-    type OutEvent = Event;
 
-    fn inject_connection_established(
+    /// Called on the [`FromSwarm::ConnectionEstablished`]
+    /// [event](`NetworkBehaviour::on_swarm_event`).
```
thomaseizinger: Not sure if we need these? An IDE will tell you that :)

Don't feel strongly though.

protocols/autonat/src/behaviour.rs
```diff
@@ -478,35 +504,59 @@ impl NetworkBehaviour for Behaviour {
...
     }
 
-    fn inject_event(
-        &mut self,
-        peer_id: PeerId,
-        conn: ConnectionId,
-        event: RequestResponseHandlerEvent<AutoNatCodec>,
-    ) {
-        self.inner.inject_event(peer_id, conn, event)
+    fn on_swarm_event(&mut self, event: FromSwarm<Self::ConnectionHandler>) {
```
thomaseizinger, line 478: You need to raise the minimum version of `libp2p-swarm` specified in the `Cargo.toml` here to the patch version that includes this function, otherwise we might get funny dependency errors! :)

protocols/autonat/src/behaviour.rs
```diff
@@ -351,79 +356,100 @@ impl NetworkBehaviour for Behaviour {
...
+                new,
+            }));
 
         if old.is_relayed() && new.is_relayed() {
             return;
         }
-        let connections = self.connected.get_mut(peer).expect("Peer is connected.");
+        let connections = self
+            .connected
+            .get_mut(&peer_id)
```
thomaseizinger: It would be good to reduce diff here by renaming `peer_id` to `peer` and perhaps `connection_id` to `conn`?

protocols/autonat/src/behaviour.rs
```diff
@@ -298,43 +302,44 @@ impl Behaviour {
...
+        }: ConnectionEstablished,
+    ) {
+        self.inner
+            .on_swarm_event(FromSwarm::ConnectionEstablished(ConnectionEstablished {
+                peer_id,
+                connection_id,
+                endpoint,
+                failed_addresses,
+                other_established,
+            }));
```
thomaseizinger: Can we not generically do this in the new `on_swarm_event` callback?

protocols/autonat/src/behaviour.rs
```diff
@@ -298,43 +302,44 @@ impl Behaviour {
...
+
+        let connections = self.connected.entry(peer_id).or_default();
```
thomaseizinger: Same here as mentioned before, I'd be great if we can avoid the diff here that is only caused by renames.

protocols/dcutr/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+# 0.7.1 - [unreleased]
+
+- replace `Behaviour`'s NetworkBehaviour implemention `inject_*` methods with the new `on_*` methods.
```
thomaseizinger:
```suggestion
- Replace `Behaviour`'s NetworkBehaviour implemention `inject_*` methods with the new `on_*` methods.
```

protocols/dcutr/src/behaviour.rs
```diff
@@ -134,17 +125,45 @@ impl NetworkBehaviour for Behaviour {
...
+                .expect("Peer of direct connection to be tracked.");
+            connections
+                .remove(&connection_id)
+                .then(|| ())
+                .expect("Direct connection to be tracked.");
+            if connections.is_empty() {
+                self.direct_connections.remove(&peer_id);
+            }
         }
     }
```
thomaseizinger: Can we preserve the order here to minimise the diff? i.e. do `connection_closed` after `dial_failure`?

protocols/dcutr/src/behaviour.rs
```diff
@@ -177,45 +196,36 @@ impl NetworkBehaviour for Behaviour {
...
+    type OutEvent = Event;
+
+    fn new_handler(&mut self) -> Self::ConnectionHandler {
+        handler::Prototype::UnknownConnection
     }
 
-    fn inject_event(
+    fn addresses_of_peer(&mut self, _peer_id: &PeerId) -> Vec<Multiaddr> {
+        vec![]
+    }
```
thomaseizinger: Nit: This is the default and can be omitted. Could do in a separate PR :)

protocols/floodsub/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+# 0.40.1 - [unreleased]
+
+- replace `Floodsub`'s NetworkBehaviour implemention `inject_*` methods with the new `on_*` methods.
+  See [PR 3011].
```
thomaseizinger, line 8: You need to actually add the link to, otherwise this reference is dead :)

thomaseizinger (commented):
Did a first pass, still need to review the remaining code! :)

protocols/autonat/CHANGELOG.md
jxs: Thanks Thomas, updated :)

protocols/autonat/src/behaviour.rs
jxs: yeah you are right, removed them :)

protocols/dcutr/CHANGELOG.md
jxs: updated :)

protocols/autonat/src/behaviour.rs
jxs: we can for the variants that can be `Copy` which is all but `ConnectionClosed`. Did it

protocols/autonat/src/behaviour.rs
jxs: done

protocols/autonat/src/behaviour.rs
jxs: done

protocols/autonat/src/behaviour.rs
jxs: thanks Thomas! updated :)

swarm/src/behaviour.rs
thomaseizinger: Yes. When overriding the callback, users will copy the signature (or IntelliJ / Rust-analyzer will do it for them). If they can't reference the type alias, that is an additional friction point.

protocols/autonat/src/behaviour.rs
thomaseizinger: > we can for the variants that can be `Copy` which is all but `ConnectionClosed`. Did it

Hmm, it would be really nice if the entire event is `Clone` but as you said, `ConnetionClosed` can't be because of `Handler`. That is quite unfortunate. I wonder if we can work around this somehow. Having to unpack every `FromSwarm` because you want to delegate to an inner implementation is really annoying. The fact that this composition is not completely smooth may bite us in the future :/

I'll track this separately, no need to block this PR here :)

protocols/floodsub/CHANGELOG.md
thomaseizinger: We typically have a link section at the bottom the changelog entry, see other versions for reference!

protocols/dcutr/src/behaviour.rs
jxs: ah right, thanks! Since it's just this case I removed it, no need to open a PR just for a small change wdyt?

protocols/floodsub/CHANGELOG.md
jxs: Ah right! :see_no_evil:   sorry Thomas fixed!

protocols/dcutr/src/behaviour.rs
jxs: yeah no worries, did it also for the `request-response` case :)

swarm/src/behaviour.rs
jxs: yeah, noticed it recently. Thanks Thomas

protocols/autonat/Cargo.toml
```diff
@@ -19,7 +19,7 @@ futures = "0.3"
 futures-timer = "3.0"
```
mxinden:
``` diff
- version = "0.8.0"
+ version = "0.8.1"
```

protocols/autonat/src/behaviour.rs
```diff
@@ -478,35 +475,65 @@ impl NetworkBehaviour for Behaviour {
...
-        peer_id: PeerId,
-        conn: ConnectionId,
-        event: RequestResponseHandlerEvent<AutoNatCodec>,
-    ) {
-        self.inner.inject_event(peer_id, conn, event)
+    fn on_swarm_event(&mut self, event: FromSwarm<Self::ConnectionHandler>) {
+        match event {
+            FromSwarm::ConnectionEstablished(connection_established) => {
+                self.inner
+                    .on_swarm_event(FromSwarm::ConnectionEstablished(connection_established));
```
mxinden, line 482: Why only forward some of the `SwarmEvent`s to the underlying `libp2p-request-response` `NetworkBehaviour`? Why not forward all of them before the `match event {` above?

protocols/dcutr/Cargo.toml
```diff
@@ -18,7 +18,7 @@ futures = "0.3.1"
 futures-timer = "3.0"
```
mxinden: Same here, you would need to bump the `version` field.

protocols/dcutr/src/behaviour.rs
```diff
@@ -316,6 +316,19 @@ impl NetworkBehaviour for Behaviour {
...
+    fn on_swarm_event(&mut self, event: FromSwarm<Self::ConnectionHandler>) {
+        match event {
+            FromSwarm::ConnectionEstablished(connection_established) => {
+                self.on_connection_established(connection_established)
+            }
+            FromSwarm::ConnectionClosed(connection_closed) => {
+                self.on_connection_closed(connection_closed)
+            }
+            FromSwarm::DialFailure(dial_failure) => self.on_dial_failure(dial_failure),
+            _ => {}
```
mxinden: I would be in favor of not using a wildcard (i.e. `_`) here. In case we ever introduce a new `FromSwarm` event variant, I would appreciate the compiler error message. Instead I think we should write out the events here and have the match arm be a simple `{}`. I don't think that would be too verbose.

What do folks think?

protocols/gossipsub/src/behaviour/tests.rs
```diff
@@ -2351,9 +2353,10 @@ fn test_add_outbound_peers_if_min_is_not_satisfied() {
 
 //TODO add a test that ensures that new outbound connections are recognized as such.
 // This is at the moment done in behaviour with relying on the fact that the call to
-// `inject_connection_established` for the first connection is done before `inject_connected`
-// gets called. For all further connections `inject_connection_established` should get called
-// after `inject_connected`.
+// `NetworkBehaviour::on_swarm_event` with `FromSwarm::ConnectionEstablished`
+// for the first connection is done before `inject_connected`
```
mxinden: The mention of `inject_connected` is outdated here, right?

swarm/CHANGELOG.md
```diff
@@ -1,3 +1,15 @@
+# 0.40.1 - [unreleased]
+
+- Add new `on_swarm_event` method to `NetworkBehaviour` that accepts a `FromSwarm` enum and update
+  `inject_*` methods to call `on_swarm_event` with the respective `FromSwarm` variant and deprecate
+  them. See [PR 3011].
```
mxinden: Can you add an additional paragraph telling users what they need to do, e.g. something along the lines of:

> To migrate, replace your `NetworkBehaviour::inject_*` calls with a single implementation of `NetworkBehaviour::on_swarm_event`.

(My wording is likely not ideal.)

swarm/tests/public_api.rs
```diff
@@ -0,0 +1,70 @@
...
+use libp2p_swarm::behaviour::{
+    AddressChange, ConnectionClosed, ConnectionEstablished, DialFailure, ExpiredExternalAddr,
+    ExpiredListenAddr, ListenFailure, ListenerClosed, ListenerError, NewExternalAddr,
+    NewListenAddr, NewListener,
+};
+
+#[test]
+// test to break compilation everytime a variant changes,
+// forcing us to revisit each implementation
+fn swarm_event_variants() {
```
mxinden:
```suggestion
// Test to break compilation everytime a variant changes,
// forcing us to revisit each implementation
#[test]
fn swarm_event_variants() {
```

protocols/dcutr/src/behaviour.rs
mxinden: Just saw that you introduced the `swarm_event_variants`. I would prefer the my above suggestion. I don't think it is too verbose. I trust the compiler more than my regex-search skills across the repository.

I am guessing that you find this too verbose @thomaseizinger and @jxs?

mxinden (commented):
Thanks for taking this all the way across all the crates @jxs. Thanks @thomaseizinger for the help.

swarm/tests/public_api.rs
jxs: updated, thanks Max :)

swarm/CHANGELOG.md
jxs: yeah makes sense. Updated Max, see what do you think of the wording

protocols/gossipsub/src/behaviour/tests.rs
jxs: yeah I think so too, I can't find any method called `inject_connected` is the `TODO` probably also outdated?

protocols/dcutr/src/behaviour.rs
jxs: I don't have a strong opinion, and yeah your reasoning regarding compiler vs manual regex skills makes total sense. So if you and  @thomaseizinger agree I can update to that :)

protocols/autonat/src/behaviour.rs
jxs: We can't easily, some variants are not `Copy` so if we just pass the `event` received to the inner `on_swarm_event` we move it, Thomas opened https://github.com/libp2p/rust-libp2p/issues/3046 due to a similar issue.

protocols/autonat/Cargo.toml
jxs: Thanks Max, updated!

protocols/dcutr/Cargo.toml
jxs: thanks, also did it for the remaining crates where the version was not bumped yet.

protocols/dcutr/src/behaviour.rs
thomaseizinger: Happy to match exhaustively yeah!

jxs: updated :)

swarm/CHANGELOG.md
```diff
@@ -1,3 +1,19 @@
+# 0.40.1 - [unreleased]
+
+- Add new `on_swarm_event` method to `NetworkBehaviour` that accepts a `FromSwarm` enum and update
+  `inject_*` methods to call `on_swarm_event` with the respective `FromSwarm` variant and deprecate
+  them.
```
```suggestion
  `inject_*`.
```

Would be easier to read for me.

swarm/CHANGELOG.md
```diff
@@ -1,3 +1,19 @@
+# 0.40.1 - [unreleased]
```
mxinden: `Cargo.toml` bump on `libp2p-swarm` is missing.

mxinden (approved):
This looks good to me, minus the two comments. @thomaseizinger ready to merge from your end as well?

swarm/CHANGELOG.md
jxs: right, thanks for reminding. There is a version mismatch between the released version of `libp2p-swarm` and it's entry on the `CHANGELOG.MD`. Bumped to `0.40.2` and created https://github.com/libp2p/rust-libp2p/pull/3063 to address it, probably should be merged first.

swarm/CHANGELOG.md
jxs: addressed, thanks Max.

protocols/dcutr/src/behaviour.rs
```diff
@@ -81,30 +82,18 @@ impl Behaviour {
...
+            peer_id,
+            connection_id,
+            endpoint,
+            ..
+        }: ConnectionEstablished,
     ) {
-        if connected_point.is_relayed() {
-            if connected_point.is_listener() && !self.direct_connections.contains_key(peer_id) {
+        if endpoint.is_relayed() {
+            if endpoint.is_listener() && !self.direct_connections.contains_key(&peer_id) {
```
thomaseizinger: Still a few opportunities to reduce the diff here but not strictly necessary IMO :)

protocols/kad/src/behaviour.rs
```diff
@@ -1853,38 +1811,98 @@ where
...
+                dial_opts::PeerCondition::Disconnected | dial_opts::PeerCondition::NotDialing,
+            ) => {
+                // We might (still) be connected, or about to be connected, thus do not report the
+                // failure to the queries.
+            }
+            DialError::DialPeerConditionFalse(dial_opts::PeerCondition::Always) => {
+                unreachable!("DialPeerCondition::Always can not trigger DialPeerConditionFalse.");
+            }
+        }
+    }
```
thomaseizinger: It would be great to retain the original order here.

protocols/ping/src/lib.rs
```diff
@@ -119,7 +119,7 @@ impl NetworkBehaviour for Behaviour {
         Handler::new(self.config.clone())
     }
 
-    fn inject_event(&mut self, peer: PeerId, _: ConnectionId, result: Result) {
+    fn on_connection_handler_event(&mut self, peer: PeerId, _: ConnectionId, result: Result) {
         self.events.push_front(Event { peer, result })
     }
```
thomaseizinger, line 128: Is the new `on_swarm_event` not mandatory? Should it be?

swarm/src/behaviour.rs
```diff
@@ -721,3 +849,264 @@ impl Default for CloseConnection {
...
+    /// Informs the behaviour that an external address was removed.
+    ExpiredExternalAddr(ExpiredExternalAddr<'a>),
+}
+
+/// [`FromSwarm`] variant that informs the behaviour about a newly established connection to a peer.
+#[derive(Clone, Copy)]
+pub struct ConnectionEstablished<'a> {
+    pub peer_id: PeerId,
+    pub connection_id: ConnectionId,
+    pub endpoint: &'a ConnectedPoint,
```
thomaseizinger, line 900: I am not quite sure what to think of these references.

We could just clone these values. In a very big behaviour tree, that could result in quite a few allocations though (one per behaviour).

But having only some fields be references and not others is also weird. I'd expect a `NetworkBehaviour` to either fully consume an event or not at all.

Would it make more sense to store a reference to the entire event in the `FromSwarm` enum instead? A particular `NetworkBehaviour`would then clone it on-demand.

thomaseizinger:
> This looks good to me, minus the two comments. @thomaseizinger ready to merge from your end as well?

Just doing a final review, give me another day please.

swarm/src/behaviour.rs
```diff
@@ -721,3 +849,264 @@ impl Default for CloseConnection {
...
+/// [`FromSwarm`] variant that informs the behaviour about a closed connection to a peer.
+///
+/// This event is always paired with an earlier
+/// [`FromSwarm::ConnectionEstablished`] with the same peer ID, connection ID
+/// and endpoint.
+pub struct ConnectionClosed<'a, Handler: IntoConnectionHandler> {
+    pub peer_id: PeerId,
+    pub connection_id: ConnectionId,
+    pub endpoint: &'a ConnectedPoint,
+    pub handler: <Handler as IntoConnectionHandler>::Handler,
```
thomaseizinger, line 914: The variants which have the `Handler` also prevent our enum from being `cloned`. Aren't these semantically more an event that is coming from the connection handler and should thus be merged with the `on_connection_handler_event` callback?

thomaseizinger (commented):
Thanks and sorry for the delay!

Please see the inline comments. Also, I think we should adjust `swarm-derive` as part of this PR.

protocols/dcutr/src/behaviour.rs
jxs: yeah, I assumed for this cases that a nearby diff was already being shown, using the new field names would be preferable, updated :) 
Btw Thomas, why do you prefer less diff over less name duplication?

protocols/kad/src/behaviour.rs
jxs: yeah, thanks Thomas. Addressed

protocols/ping/src/lib.rs
jxs: AFAIR I think @mxinden plans it to be after `inject_*` methods are deprecated.  
Implemented here `on_swarm_event` with all the events matched and noop'ed. Thanks

swarm/src/behaviour.rs
jxs: >Would it make more sense to store a reference to the entire event in the FromSwarm enum instead? A particular NetworkBehaviourwould then clone it on-demand.

yeah, it makes sense, I think it's ergonomically more elegant. There's only one problem with the `DialFailure` variant, `DialError` not being `clone` bytes us again, it would need to be breaking cause in `inject_dial_failure` we receive `&DialError`.

swarm/src/behaviour.rs
jxs: And how would we merge these variants with each `ConnectionHandler::OutEvent`?

jxs:
> Thanks and sorry for the delay!
> 
> Please see the inline comments. Also, I think we should adjust `swarm-derive` as part of this PR.

yeah, I assumed `swarm-derive` was to be done when deprecating the `inject_*` methods, but you are right can be done now. Updated ptal Thomas. 
Will solve the conflicts with the `CHANGELOG` and `Cargo` files

protocols/ping/src/lib.rs
mxinden: > Is the new `on_swarm_event` not mandatory? Should it be?

Are you suggesting to not provide a default implementation of `on_swarm_event` on the `Transport` `trait` definition @thomaseizinger? Would that not be in conflict with the idea of rolling this whole thing out in a non-breaking way?

swarm/src/behaviour.rs
mxinden: On a first look I agree that this is less intuitive. Taking a deeper look, given that only the `Copy` types are non-references, I think this is fine. In addition, I would be in favor of doing this as a follow-up, if we decide for it.

protocols/autonat/CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+# 0.8.1 - [unreleased]
+
+- Replace `Behaviour`'s NetworkBehaviour implemention `inject_*` methods with the new `on_*` methods.
```
mxinden:
```suggestion
- Replace `Behaviour`'s `NetworkBehaviour` implemention `inject_*` methods with the new `on_*` methods.
```

Nit pick. For the sake of consistency.

protocols/gossipsub/src/behaviour/tests.rs
mxinden: Yes, `TODO` is outdated since https://github.com/libp2p/rust-libp2p/pull/2445. Good catch @jxs. Please remove.

swarm/CHANGELOG.md
```diff
@@ -1,3 +1,19 @@
...
+- Add new `on_swarm_event` method to `NetworkBehaviour` that accepts a `FromSwarm` enum and update
+  `inject_*` methods to call `on_swarm_event` with the respective `FromSwarm` variant and deprecate
+  `inject_*`.
+  To migrate, users should replace the `NetworkBehaviour::inject_*` calls with a single
+  implementation of `NetworkBehaviour::on_swarm_event` treating each `FromSwarm` variant in
+  the same way its corresponding `inject_*` call was treated.
+  See [PR 3011].
+
+- Add new `on_connection_handler_event` method with the same signature as `inject_event`, make the
+  default implementation of `inject_event` call `on_connection_handler_event` and deprecate it.
```
mxinden, line 14: Can you add a "To migrate, users ..." section here as well?

swarm/src/behaviour.rs
mxinden: Off the top of my head, I would see these events to better fit `on_swarm_event`. An event coming from the `ConnectionHandler` containing a `ConnectionHandler` seems off to me.

protocols/dcutr/src/behaviour.rs
thomaseizinger: > Btw Thomas, why do you prefer less diff over less name duplication?

I find it easier to review if the diff is as minimal as possible, especially for refactorings of this scale.

We can always send another PR later that just renames those variables again! :)

protocols/ping/src/lib.rs
thomaseizinger: > > Is the new `on_swarm_event` not mandatory? Should it be?
> 
> Are you suggesting to not provide a default implementation of `on_swarm_event` on the `Transport` `trait` definition @thomaseizinger? Would that not be in conflict with the idea of rolling this whole thing out in a non-breaking way?

Ah yeah, silly me! Thanks for clarifying!

swarm/src/behaviour.rs
thomaseizinger: > Off the top of my head, I would see these events to better fit `on_swarm_event`. An event coming from the `ConnectionHandler` containing a `ConnectionHandler` seems off to me.

What if we have a third callback? If we look at the swarm-derive closely, we can see that there is a difference here.

Some events are given as is to all behaviours. For events that contain a connection handler, each behaviour only receives the handler it originally created.

If we split these into different callbacks (and assuming we can get rid of the listener error), then the non-handler events could just be cloned which would make forwarding much easier!

mxinden: > Some events are given as is to all behaviours. For events that contain a connection handler, each behaviour only receives the handler it originally created.

Agreed that there is a difference here.

> What if we have a third callback? If we look at the swarm-derive closely, we can see that there is a difference here.

I would need to give this more thought. Off the top of my head the above does not justify an additional hook.

I would prefer this being explored in a separate pull request. Would that be OK for you @thomaseizinger?

thomaseizinger: > > What if we have a third callback? If we look at the swarm-derive closely, we can see that there is a difference here.
> 
> I would need to give this more thought. Off the top of my head the above does not justify an additional hook.
> 
> I would prefer this being explored in a separate pull request. Would that be OK for you @thomaseizinger?

Yeah we can explore that on top of this for sure.

swarm/CHANGELOG.md
jxs: yes, done Max :)

protocols/autonat/CHANGELOG.md
jxs: right, thanks Max did it for all protocol CHANGELOGs

protocols/gossipsub/src/behaviour/tests.rs
jxs: done :)

protocols/dcutr/src/behaviour.rs
jxs: yeah makes sense, thanks for explaining Thomas. :blush:

swarm/CHANGELOG.md
mxinden: Looks good to me. Thanks.

swarm-derive/src/lib.rs
```diff
@@ -181,35 +190,44 @@ fn build_struct(ast: &DeriveInput, data_struct: &DataStruct) -> TokenStream {
...
+            .map(|(field_n, field)| match field.ident {
+                Some(ref i) => quote! { self.#i.on_swarm_event(#from_swarm::AddressChange(#address_change { peer_id, connection_id, old, new }));},
+                None => quote! { self.#field_n.on_swarm_event(#from_swarm::AddressChange(#address_change { peer_id, connection_id, old, new }));},
+            })
     };
 
-    // Build the list of statements to put in the body of `inject_connection_closed()`.
-    let inject_connection_closed_stmts = {
+    // Build the list of statements to put in the body of `on_swarm_event()`
+    // for the FromSwarm::ConnectionClosed variant.
```
mxinden:
```suggestion
    // for the `FromSwarm::ConnectionClosed` variant.
```

Nit pick, for the sake of consistency.

mxinden (approved):
This looks good to me. I suggest we wait for another review from @thomaseizinger before merging here.

swarm-derive/src/lib.rs
jxs: thanks Max! addressed this and the remaining cases.

jxs:
addressed https://github.com/libp2p/rust-libp2p/issues/2854#issuecomment-1303607189 on the last 2 commits, you probably want to take a look again @mxinden and @thomaseizinger

thomaseizinger:
> addressed https://github.com/libp2p/rust-libp2p/issues/2854#issuecomment-1303607189 on the last 2 commits, you probably want to take a look again @mxinden and @thomaseizinger 

I think there was a misunderstanding, sorry for being ambiguous in my answer there.

I think we should definitely soft-deprecate first, i.e. make it a backwards-compatible change and do the rename of the associated type at a later point. The rename is easy to fix but I think it might be quite a big upgrade if too many things change at a time!

jxs:
yeah, sorry I misread from Max's comment :see_no_evil:. Removed last commit :)

CHANGELOG.md
```diff
@@ -76,6 +79,21 @@
...
+  the same way its corresponding `inject_*` call was treated.
+  See [PR 3011].
+
+- Add new `on_connection_handler_event` method with the same signature as `inject_event`, make the
+  default implementation of `inject_event` call `on_connection_handler_event` and deprecate it.
+  To migrate, users should replace the `NetworkBehaviour::inject_event` calls
+  with `NetworkBehaviour::on_connection_handler_event`.
+  See [PR 3011].
+
+[PR 3011]: https://github.com/libp2p/rust-libp2p/pull/3011
```
thomaseizinger: This is duplicated with the `swarm` CHANGELOG. Not sure if needs to be? We typically don't call out sub-changelogs, right?

thomaseizinger (approved):
Excellent work @jxs! Thanks for bearing with all my nitpicks :)

CHANGELOG.md
jxs: Thanks Thomas, addressed :)

mxinden:
I suggest we merge this together with https://github.com/libp2p/rust-libp2p/pull/3085 to make sure both land in the same release. Objections?

thomaseizinger:
> I suggest we merge this together with https://github.com/libp2p/rust-libp2p/pull/3085 to make sure both land in the same release. Objections?

That would be my preference too!

swarm/src/behaviour/either.rs
```diff
@@ -49,173 +47,42 @@ where
...
-            (Either::Right(behaviour), Either::Right(handler)) => behaviour
-                .inject_connection_closed(
-                    peer_id,
-                    connection,
-                    endpoint,
-                    handler,
-                    remaining_established,
-                ),
-            _ => unreachable!(),
+            Either::Left(b) => b.on_swarm_event(event.map_handler(
```
mxinden: @thomaseizinger comment on https://github.com/libp2p/rust-libp2p/pull/3085#discussion_r1015994812 sparked a thought.

What if `b` implements the `inject_*` style methods and not the `on_*` style methods? In such case, this call would be lost, right?

Unless I am missing something combinators like `Either` need to call the `inject_*` style methods until we remove them.

jxs: No no, I noticed it when replacing `Inject_*` on combinators for #3085 but with the deprecation/removal discussion forgot to address it here, thanks for reminding @mxinden! Updated to call `inject` methods on those structs. Also merged from master and fixed conflicts ptal.

swarm-derive/src/lib.rs
```diff
@@ -171,35 +179,44 @@ fn build_struct(ast: &DeriveInput, data_struct: &DataStruct) -> TokenStream {
...
-        })
+    // Build the list of statements to put in the body of `on_swarm_event()`
+    // for the `FromSwarm::ConnectionEstablished` variant.
+    let on_connection_established_stmts = {
+        data_struct
+            .fields
+            .iter()
+            .enumerate()
+            .map(|(field_n, field)| match field.ident {
+                Some(ref i) => quote! { self.#i.on_swarm_event(
```
mxinden: `swarm-derive` will need to delegate to the `inject_*` methods of the corresponding `NetworkBehaviour` implementations, right? Otherwise our non-breaking deprecation doesn't work.

mxinden (commented):
I am afraid there is still one larger work item on `swarm-derive`. Otherwise this pull request looks good to me.

swarm-derive/src/lib.rs
jxs: yeah right, thanks for noticing Max, I think I addressed all the cases but ptal.

mxinden (approved):

mxinden:
Thanks for the follow-ups. Once the last comments in https://github.com/libp2p/rust-libp2p/pull/3085#pullrequestreview-1180620311 are addressed, we can merge both.

mxinden:
https://github.com/libp2p/rust-libp2p/pull/3085 is all approved. Adding `send-it` here.

swarm/src/behaviour.rs
```diff
@@ -722,3 +850,364 @@ impl Default for CloseConnection {
         CloseConnection::All
     }
 }
+
+/// Enumeration with the list of the possible events
+/// to pass to [`on_swarm_event`](NetworkBehaviour::on_swarm_event).
+pub enum FromSwarm<'a, Handler: IntoConnectionHandler> {
```
nazar-pc, line 856: Why isn't there `Clone` derive? I need to forward this to underlying protocols and it seems like I need to match on all variants just to be able to clone and wrap it back into `FromSwarm` or am I missing something?

thomaseizinger: There are multiple variants that contain a `ConnectionHandler` which can't be cloned.

See https://github.com/libp2p/rust-libp2p/issues/3046.

rkuhn: Sometimes it feels like we're going through a lot of painful contortions just to avoid saying `Arc`.

thomaseizinger: > Sometimes it feels like we're going through a lot of painful contortions just to avoid saying `Arc`.

In this case, we would also have to say `try_unwrap`: https://doc.rust-lang.org/std/sync/struct.Arc.html#method.try_unwrap

Possible but doesn't feel very clean. I don't particularly like both solutions but I think we are getting one step closer with the new event into making composition easier.

https://github.com/libp2p/rust-libp2p/pull/3099 will help with this too because we are removing a few more mentions of the handler there.

PR #3023, opened 2022-10-14
fix: Remove circular dependencies across workspace

### LINKED ISSUE #3053: Avoid circular dependencies in workspace
Currently we have circular dependencies in our workspace.

This is problematic because:

- It messes with the build cache during incremental builds: https://github.com/libp2p/rust-libp2p/pull/2918#discussion_r976514245
- It makes tools like "release-please" choke: https://github.com/libp2p/rust-libp2p/issues/2902#issuecomment-1260557546

Getting rid of the circular dependencies is almost easily solvable, except for the `NetworkBehaviour` custom derive.

That one emits code that assumes that a crate `libp2p` is in scope where it can import items from. That is an issue because it means we can't use the macro in our workspace without depending on the `libp2p` meta crate which is a circular dependency again.

The real solution here is being discussed over at rust-lang/rust: https://github.com/rust-lang/rust/issues/54363.

We can't wait for that to be implemented though, so we need to find a workaround.

### LINKED ISSUE #3223: webrtc/tests: data channel opening took longer than 10 seconds
## Summary

Test `smoke` times out with:

```
test smoke ... FAILED

failures:

---- smoke stdout ----
thread 'smoke' panicked at 'called `Result::unwrap()` on an `Err` value: Custom { kind: Other, error: Internal("data channel opening took longer than 10 seconds (see logs)") }', transports/webrtc/tests/smoke.rs:270:42

error: test failed, to rerun pass `-p libp2p-webrtc --test smoke`

failures:
    smoke

test result: FAILED. 1 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 10.01s
```

https://github.com/libp2p/rust-libp2p/actions/runs/3659637005/jobs/6186244264 on https://github.com/libp2p/rust-libp2p/pull/3023

## Expected behaviour

Test to succeed.

No. Also let's see whether this fails more often before investigating.

thomaseizinger:
## Description

Circular dependencies are problematic in several ways:

- They result in cognitive overhead for developers, in trying to figure out what depends on what.
- They present `cargo` with limits in what order the crates can be compiled in.
- They invalidate build caches unnecessarily thus forcing `cargo` to rebuild certain crates.
- They cause problems with tooling such as `release-please`.

To actually break the circular dependencies, this patch inlines the uses of `development_transport` in the examples and tests for all sub-crates. This is only meant to be a short-term fix until https://github.com/libp2p/rust-libp2p/issues/3111 and https://github.com/libp2p/rust-libp2p/pull/2888 are fixed.

To ensure we don't accidentally reintroduce this dependency, we add a basic CI that queries `cargo metadata` using `jq`.

Resolves https://github.com/libp2p/rust-libp2p/issues/3053.
Fixes https://github.com/libp2p/rust-libp2p/issues/3223.
Related: https://github.com/libp2p/rust-libp2p/pull/2918#discussion_r976514245
Related: https://github.com/googleapis/release-please/issues/1662

- Blocked by https://github.com/libp2p/rust-libp2p/pull/3056.
- Blocked by https://github.com/libp2p/rust-libp2p/pull/3055.
- Depends-On: https://github.com/libp2p/rust-libp2p/pull/3131
- Depends-On: https://github.com/libp2p/rust-libp2p/pull/3142

## Change checklist


thomaseizinger:
Actually, #3006 is not going to quite fix it because we depend on symbols there that are re-exported from `libp2p_core` :(

galargh:
> This is a draft PR as the current build process is a bit of a mess and doesn't catch all problems. Basically what happens is that when we build with --all-features, all the features in the libp2p crate also get activated so all crates that depend on the libp2p for the derive macro can just happily import their stuff through libp2p:: which is then a circular dependency and invalidates build-caches.

That's interesting! Do you know how other projects which have a similar setup are performing builds?

thomaseizinger:
> > This is a draft PR as the current build process is a bit of a mess and doesn't catch all problems. Basically what happens is that when we build with --all-features, all the features in the libp2p crate also get activated so all crates that depend on the libp2p for the derive macro can just happily import their stuff through libp2p:: which is then a circular dependency and invalidates build-caches.
> 
> That's interesting! Do you know how other projects which have a similar setup are performing builds?

Sorry, that description is outdated as of the latest commits.

Nope, I haven't looked at any other projects yet.

thomaseizinger:
@mxinden @jxs A review would be welcome here. I opened https://github.com/libp2p/rust-libp2p/issues/3111 so we can move forward here without blocking on the de-duplication of `development_transport`.

Note that this is currently also blocked by https://github.com/libp2p/rust-libp2p/pull/3131. I expressed this in the PR description so in case everything else is approvable, we can let mergify take care of merging this in the correct order.

src/lib.rs
```diff
@@ -184,6 +184,10 @@ pub use self::transport_ext::TransportExt;
         note = "The `tcp-async-io` and `dns-async-std` features are deprecated. Use the new `tcp` and `dns` features together with the `async-std` feature."
     )
 )]
+#[deprecated(
+    since = "0.50.0",
+    note = "Compose your own transport, specific to your needs."
+)]
 pub async fn development_transport(
```
mxinden: Needs a changelog entry.

mxinden (approved):
Can you provide a patch to our coding guidelines stating that sub-crates should not import `libp2p` for testing? (Or a CI step, but I guess that is more involved.) I don't expect newcomers to read the document. But this way we can point them to it for them to discover the reasoning.

Thanks for doing the toil!

Looks good to me from my side.

src/lib.rs
thomaseizinger: Oh I thought I deleted that!

thomaseizinger:
> Or a CI step, but I guess that is more involved.

Good idea! I hacked together a `jq` query :upside_down_face:

thomaseizinger:
@mxinden Can you give this a re-approval? A few things changed since you last reviewed this! @jxs I wouldn't mind getting a review from you too!

I think this is a non-breaking change because I only touch tests and examples but let me know if you can spot anything either way!

thomaseizinger:
> @mxinden Can you give this a re-approval? A few things changed since you last reviewed this! @jxs I wouldn't mind getting a review from you too!
> 
> I think this is a non-breaking change because I only touch tests and examples but let me know if you can spot anything either way!

The motivation for landing this is (among other things) to make `release-please` work which - together with semantic commits - avoids the manual bumping of manifests and writing of changelogs.

.github/workflows/ci.yml
```diff
@@ -73,6 +73,11 @@ jobs:
           cargo install cargo-semver-checks --locked
           cargo semver-checks check-release -p ${{ matrix.crate }}
 
+      - name: Enforce no dependency on meta crate
+        run: |
+          cargo metadata --format-version=1 --no-deps | \
+          jq -e -r '.packages[] | select(.name == "${{ matrix.crate }}") | .dependencies | all(.name != "libp2p")'
```
mxinden, line 81: Very neat. Thanks @thomaseizinger!

mxinden (approved):
This looks good to me. Thanks especially for the developer experience improvement.

jxs (approved):

thomaseizinger:
CI error should be fixed now, please merge whenever you are ready @mxinden!

mxinden:
WebRTC failure seems to be consistent (i.e. non-flaky as assumed in https://github.com/libp2p/rust-libp2p/issues/3223). @thomaseizinger can you take a look?

thomaseizinger:
> WebRTC failure seems to be consistent (i.e. non-flaky as assumed in #3223). @thomaseizinger can you take a look?

Can confirm, it is not a flake. The new "swarmless" test doesn't work for WebRTC because it is not polling the `Transport` whilst upgrading connections.

thomaseizinger:
Okay, test should pass now. The solution is a bit verbose. I think we should transition `libp2p-muxer-test-harness` into `libp2p-core-test` and use that for all transports and muxers so we can cut down on the duplication and complexity of testing transports.

thomaseizinger:
It's been a long journey but I am very happy that this is merged 🎉

PR #2931, opened 2022-09-22
*: Prepare v0.49.0

mxinden:
# Description

In preparation for the `libp2p` `v0.49.0` release.

Waiting on:

- https://github.com/libp2p/rust-libp2p/pull/2928
- https://github.com/libp2p/test-plans/pull/41
- https://github.com/libp2p/rust-libp2p/pull/2932
- https://github.com/libp2p/rust-libp2p/pull/2934
- https://github.com/libp2p/rust-libp2p/pull/2974
- https://github.com/libp2p/rust-libp2p/pull/2960

## Open Questions

Anything else folks would like to block this release on?

## Change checklist

- [x] I have performed a self-review of my own code
- [ ] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] A changelog entry has been made in the appropriate crates

PR #3027, opened 2022-10-14
transports/dns/: Don't feature flag std::io import

mxinden:
# Description

Functions like `parse_dnsaddr_txt` depend on the `std::io` import. Given that the function is not feature flagged, compilation without features fails.

Discovered while releasing `v0.49.0`. Would like to get this in before continuing the `v0.49.0` release. https://github.com/libp2p/rust-libp2p/pull/2931

## Open Questions

Even though `libp2p-dns` is useless without either the async-io or tokio feature flag, I do still think that it should compile without. Objections?

## Change checklist

### DIFF
--- a/transports/dns/src/lib.rs
+++ b/transports/dns/src/lib.rs
@@ -65,7 +65,6 @@ use libp2p_core::{
 };
 use parking_lot::Mutex;
 use smallvec::SmallVec;
-#[cfg(any(feature = "async-std", feature = "tokio"))]
 use std::io;
 use std::{
     convert::TryFrom,


mxinden:
I think we can ignore the failing semver checks, given that we are in between releases right now. See https://github.com/libp2p/rust-libp2p/pull/2931.

mxinden:
Given that this is a very minor change, I will go ahead and merge, thus unblocking in half-published `v0.49.0` release.

Please still review this patch. I will take care of any requested follow-ups.

thomaseizinger:
Why did CI not catch this?

mxinden:
> Why did CI not catch this?

https://github.com/libp2p/rust-libp2p/pull/3044 reverting this is not failing.

My suspicion would be that the `tokio` or `async-std` feature is enabled through some other crate and thus e.g. the "Build and test (--no-default-features) (pull_request)" CI step succeeds. One such crate would e.g. be `libp2p-mdns`:

https://github.com/libp2p/rust-libp2p/blob/4d4833f8c2622018b5b775896905a534b8624d3c/protocols/mdns/Cargo.toml#L40-L46

@thomaseizinger does that suspicion make sense?

PR #3052, opened 2022-10-21
Update noise proto to match go-libp2p

ckousik:
# Description
This updates the noise protobuf to match the js and go-libp2p repositories. It replaces the data field with NoiseExtensions.

## Change checklist

### DIFF
--- a/transports/noise/src/io/handshake/payload.proto
+++ b/transports/noise/src/io/handshake/payload.proto
@@ -2,10 +2,12 @@ syntax = "proto3";
 
 package payload.proto;
 
-// Payloads for Noise handshake messages.
+message NoiseExtensions {
+    repeated bytes webtransport_certhashes = 1;
+}
 
 message NoiseHandshakePayload {
-    bytes identity_key = 1;
-    bytes identity_sig = 2;
-    bytes data         = 3;
+  bytes identity_key = 1;
+  bytes identity_sig = 2;
+  optional NoiseExtensions extensions = 4;
 }


rkuhn:
It seems from a casual inspection that this breaks network compatibility with any current user of the `data` field. In general, protobuf evolution proceeds only by adding optional or repeated fields.

thomaseizinger:
Thanks!

We haven't even shipped QUIC yet so WebTransport is still a bit away. I am not sure this is necessary at the moment, also given the churn that is currently going on with proto 2 & 3.

ckousik:
@thomaseizinger This is needed for the webrtc interop between go-libp2p. Still need to flesh it out so I've left it as a draft.

thomaseizinger:
> @thomaseizinger This is needed for the webrtc interop between go-libp2p. Still need to flesh it out so I've left it as a draft.

Only once we implement `WebTransport` right? It shouldn't be required for regular noise handshakes.

ckousik:
We had a similar issue where the noise handshake payload had been upgraded in Go and was still on a previous version in js. We use the noise handshake in webrtc, so the authentication handshake was failing. It got fixed after the payload proto in the js repository was upgraded. You may be right however, since I cannot get the current rust implementation to work with a js client even with the updated payload.

thomaseizinger:
> We had a similar issue where the noise handshake payload had been upgraded in Go and was still on a previous version in js. We use the noise handshake in webrtc, so the authentication handshake was failing. It got fixed after the payload proto in the js repository was upgraded. You may be right however, since I cannot get the current rust implementation to work with a js client even with the updated payload.

Is this WebRTC? The roles are reversed there as per the spec but it is a bit hard to spot. Perhaps that is your issue :)

PR #3055, opened 2022-10-24
swarm-derive: Add `prelude` configuration option to `NetworkBehaviour` macro

thomaseizinger:
# Description

I recommend patch-by-patch review.

- #3053: This PR does not yet remove all circular dependency in our workspace but only demonstrates that it is possible with the tests in `libp2p_swarm`. I'd rather ship the new `prelude` feature first and later use it across the workspace as the latter might create a lot of merge conflicts because basically all tests and manifests need to be touched. The tests are also an implementation detail whereas some of the changes to the macro are user-facing.
- https://github.com/libp2p/rust-libp2p/pull/3006
- https://github.com/rust-lang/rust/issues/54363: This PR is basically a workaround for a missing feature in cargo/rust.
- https://github.com/libp2p/test-plans/pull/72: We can't use the derive macro with a renamed crate until we merge this feature.

Depends-On: https://github.com/libp2p/rust-libp2p/pull/3081

## Change checklist

## Commit message body

Currently, our `NetworkBehaviour` derive macro depends on the `libp2p` crate to be in scope. This prevents standalone usage which forces us to depend on `libp2p` in all our tests where we want to derive a `NetworkBehaviour`.

This PR introduces a `prelude` option that - by default - points to `libp2p::swarm::derive_prelude`, a new module added to `libp2p_swarm`. With this config option, users of `libp2p_swarm` can now refer to the macro without depending on `libp2p`, breaking the circular dependency in our workspace. For consistency with the ecosystem, the macro is now also re-exported by `libp2p_swarm` instead of `libp2p` at the same position as the trait that it implements.

Lastly, we introduce an off-by-default `macros` feature flag that shrinks the dependency tree for users that don't need the derive macro.


thomaseizinger:
Friendly ping @mxinden @jxs @elenaf9 :)

jxs (approved):
LGTM Thomas!

thomaseizinger:
Note that this also turns out to be a requirement for https://github.com/libp2p/test-plans/pull/72.

swarm-derive/src/lib.rs
```diff
@@ -635,6 +625,30 @@ fn build_struct(ast: &DeriveInput, data_struct: &DataStruct) -> TokenStream {
     final_quote.into()
 }
 
+/// Parses the `value` of a key=value pair in the `#[behaviour]` attribute into the requested type.
+///
+/// Only `String` values are supported, e.g. `#[behaviour(foo="bar")]`.
+fn parse_attribute_value_by_key<T>(ast: &DeriveInput, key: &str) -> Option<T>
```
mxinden, line 631: :pray:

swarm/src/lib.rs
```diff
@@ -67,6 +67,25 @@ pub mod dummy;
 pub mod handler;
 pub mod keep_alive;
 
+/// Bundles all symbols required for the `libp2p_swarm_derive::NetworkBehavior` macro.
```
```suggestion
/// Bundles all symbols required for the [`libp2p_swarm_derive::NetworkBehavior`] macro.
```

mxinden (approved):
Tiny nit pick. Otherwise this looks good to me.

swarm/src/lib.rs
thomaseizinger: This broke CI because it is not actually a link 🙃

swarm/src/lib.rs
```diff
@@ -67,6 +67,25 @@ pub mod dummy;
 pub mod handler;
 pub mod keep_alive;
 
+/// Bundles all symbols required for the [`libp2p_swarm_derive::NetworkBehavior`] macro.
```
```suggestion
/// Bundles all symbols required for the [`libp2p_swarm_derive::NetworkBehaviour`] macro.
```

swarm/Cargo.toml
```diff
@@ -17,20 +17,36 @@ futures = "0.3.1"
...
 async-std = { version = "1.6.2", features = ["attributes"] }
+either = "1.6.0"
 env_logger = "0.9"
-libp2p = { path = "..", features = ["full"] }
+futures = "0.3.1"
+libp2p-identify = { path = "../protocols/identify" }
+libp2p-kad = { path = "../protocols/kad" }
+libp2p-ping = { path = "../protocols/ping" }
+libp2p-plaintext = { path = "../transports/plaintext" }
+libp2p-swarm-derive = { version = "0.30.2", path = "../swarm-derive" }
```
mxinden, line 40: @thomaseizinger why define `libp2p-swarm-derive` with a version here?

thomaseizinger: Copy-paste error :)

PR #3056, opened 2022-10-24
*: Remove use of `development_transport` in examples and tests

thomaseizinger:
# Description

These create circular dependencies to the root libp2p crate in tests.

- https://github.com/libp2p/rust-libp2p/issues/3053
- https://github.com/libp2p/rust-libp2p/pull/2888: Once we have the shared test abstraction, more duplication in test setup will go away again. For the examples, I think we should not be using the `development_transport` but be as close to a real-world application as possible.
- https://github.com/libp2p/rust-libp2p/pull/3023: To actually break the circular dependencies between our workspace crates, we can't use the `development_transport` inside our workspace.

## Change checklist


protocols/mdns/tests/use-tokio.rs
```diff
@@ -56,7 +58,11 @@ async fn test_expired_tokio() -> Result<(), Box<dyn Error>> {
 async fn create_swarm(config: MdnsConfig) -> Result<Swarm<TokioMdns>, Box<dyn Error>> {
     let id_keys = identity::Keypair::generate_ed25519();
     let peer_id = PeerId::from(id_keys.public());
-    let transport = libp2p::tokio_development_transport(id_keys)?;
+    let transport = tcp::TcpTransport::default()
```
jxs:
```suggestion
    let transport = tcp::tokio::Transport::default()
```

jxs (commented):
Hi Thomas, LGTM appart from the `tcp::TcpTransport` -> `tcp::async_io::Transport` and `tokio` on the `use-tokio` `mdns` test

protocols/rendezvous/examples/register.rs
```diff
@@ -39,14 +42,22 @@ async fn main() {
...
+            .multiplex(yamux::YamuxConfig::default())
+            .boxed(),
         MyBehaviour {
             rendezvous: rendezvous::client::Behaviour::new(identity.clone()),
             ping: ping::Behaviour::new(ping::Config::new().with_interval(Duration::from_secs(1))),
         },
         PeerId::from(identity.public()),
-    );
+    )
+    .executor(Box::new(|f| {
```
jxs, line 55: Curious Thomas, why do we need it here but not when using `tcp::async_io::Tcp` ?

jxs (approved):
LGTM!

jxs:
just noticed https://github.com/libp2p/rust-libp2p/blob/master/src/tutorials/ping.rs also needs to be addressed

protocols/rendezvous/examples/register.rs
thomaseizinger: `async-std` implicitly creates a reactor on each thread that an IO component is polled on. tokio only creates a reactor on worker threads and thus requires explicit control over where tasks are spawned.

By default, a `Swarm` creates a `futures::executor::Threadpool` for these background tasks and that doesn't work for tokio. Also see #3068.

thomaseizinger:
How should we move forward here? Should I remove the deprecation and just not use those transports for the tests and crate-specific examples? That will break the circular dependencies which is what I am after.

We can then decide separately, what we should rename those transport constructors to.

@mxinden Input please :)

thomaseizinger:
Closing in favor of https://github.com/libp2p/rust-libp2p/pull/3023.

PR #3066, opened 2022-10-26
 Migrate from protoc to quick-protobuf

### LINKED ISSUE #3024: Migrate from `protoc` to `quick-protobuf` and remove the buildscripts
## Description

Migrate our codebase from `protoc` to `quick-protobuf`.
We also want to get rid of the buildscripts and instead check the generated code into Git and add a CI step that verifies that they are directly generated from the protobuf files.

## Motivation

- Users are irritated about the dependencies required to build rust-libp2p: https://github.com/libp2p/rust-libp2p/issues/2645
- It causes bloat in our CI workflow: Every job needs to install protoc.
- We sometimes get CI failures in installing protoc: https://github.com/libp2p/rust-libp2p/actions/runs/3243659893/jobs/5327292307

## Current Implementation

We depend on protoc being installed because prost no longer ships with it.

No.

kckeiks:
# Description

According to [perftest](https://github.com/libp2p/rust-libp2p/pull/3050#issuecomment-1291206200), `quick-proto` is faster than `protobuf`. Since Protobuf is fundamental to `libp2p`, the performance of the library we choose is important. Also, the `quick-protobuf` generated API is much simpler than `protobuf`'s.

## Open Questions

[Pros and cons](https://github.com/tafia/quick-protobuf#why-not-rust-protobuf) of `quick-protobuf`.

[List of unsupported features.](https://github.com/tafia/quick-protobuf/issues/12)

Is it likely that we will need any of those unsupported features?

Any other concerns?

## Change checklist


core/src/identity.rs
```diff
@@ -174,42 +173,37 @@ impl Keypair {
             }
         };
 
-        Ok(pk.encode_to_vec())
+        Ok(quick_protobuf::serialize_into_vec(&pk).expect("Encoding to succeed."))
```
kckeiks, line 164: I believe this is infallible by looking at the [code](https://docs.rs/quick-protobuf/latest/src/quick_protobuf/writer.rs.html#311-319). Unfortunately, I have not found a concrete answer in the docs or issues.

kckeiks:
If we want to use `quick-protobuf`, I can start pushing patches.

core/src/identity.rs
thomaseizinger: I believe so too. One error I could find is https://docs.rs/quick-protobuf/latest/src/quick_protobuf/writer.rs.html#379 which can't happen if the `Vec` is allocated correctly.

The other error path comes from lines like https://github.com/tafia/quick-protobuf/blob/d977371e05170a016f03a80512b2a925468e3a1a/quick-protobuf/examples/pb_rs/data_types.rs#L199 which end up being the same error case as above so I think we are good.

I believe this function could actually be made infallible in `quick-protobuf` but I am not sure.

core/build.rs
```diff
@@ -18,14 +18,32 @@
...
+    }
+
+    // Delete all old generated files before re-generating new ones
+    if out_dir.exists() {
+        std::fs::remove_dir_all(&out_dir).unwrap();
+    }
+
+    std::fs::DirBuilder::new().create(&out_dir).unwrap();
+    let config_builder = ConfigBuilder::new(&protos, None, Some(&out_dir), &[in_dir]).unwrap();
+    FileDescriptor::run(&config_builder.build()).unwrap()
```
thomaseizinger: This is kind of annoying boilerplate but I could live with that. Worst case we could always make a little crate that abstracts things away for us.

core/src/identity.rs
```diff
@@ -174,42 +173,37 @@ impl Keypair {
...
 
     /// Decode a private key from a protobuf structure and parse it as a [`Keypair`].
     pub fn from_protobuf_encoding(bytes: &[u8]) -> Result<Keypair, DecodingError> {
-        use prost::Message;
-
-        let mut private_key = keys_proto::PrivateKey::decode(bytes)
-            .map_err(|e| DecodingError::new("Protobuf").source(e))
-            .map(zeroize::Zeroizing::new)?;
+        let mut private_key: zeroize::Zeroizing<keys_proto::PrivateKey> =
+            quick_protobuf::deserialize_from_slice(bytes)
```
thomaseizinger: Nit: I'd prefer using turbo-fish here.

```suggestion
        let mut private_key =
            quick_protobuf::deserialize_from_slice::<keys_proto::PrivateKey>(bytes)
```

core/src/identity.rs
```diff
@@ -279,55 +265,52 @@ impl PublicKey {
...
 
-impl From<&PublicKey> for keys_proto::PublicKey {
+impl From<&PublicKey> for keys_proto::PublicKey<'_> {
     fn from(key: &PublicKey) -> Self {
         match key {
             PublicKey::Ed25519(key) => keys_proto::PublicKey {
-                r#type: keys_proto::KeyType::Ed25519 as i32,
-                data: key.encode().to_vec(),
+                Type: keys_proto::KeyType::Ed25519,
+                Data: Cow::from(key.encode().to_vec()),
```
thomaseizinger, line 255: I think the usage of `Cow` here would actually allow us to remove an allocation. A `Ed25519` key in particular can also be viewed as a byte-slice so we could use `Cow::Borrowed` here.

core/src/signed_envelope.rs
```diff
@@ -73,37 +75,28 @@ impl SignedEnvelope {
...
-        use prost::Message;
-
         let envelope = crate::envelope_proto::Envelope {
             public_key: Some((&self.key).into()),
-            payload_type: self.payload_type,
-            payload: self.payload,
-            signature: self.signature,
+            payload_type: Cow::from(self.payload_type),
+            payload: Cow::from(self.payload),
+            signature: Cow::from(self.signature),
```
thomaseizinger: These could allow be borrowed :)

core/src/peer_record.rs
```diff
@@ -71,21 +69,17 @@ impl PeerRecord {
...
             let record = peer_record_proto::PeerRecord {
-                peer_id: peer_id.to_bytes(),
+                peer_id: Cow::from(peer_id.to_bytes()),
                 seq,
                 addresses: addresses
                     .iter()
-                    .map(|m| peer_record_proto::peer_record::AddressInfo {
-                        multiaddr: m.to_vec(),
+                    .map(|m| peer_record_proto::mod_PeerRecord::AddressInfo {
+                        multiaddr: Cow::from(m.to_vec()),
```
thomaseizinger: This one could be borrowed too I think.

thomaseizinger (commented):
Thank you for making this patch!

> Is it likely that we will need any of those unsupported features?

I don't think so. We only use protobufs to have a language-agnostic message description language so we are unlikely to use services for example.

I am not quite sure what `option` refers too? Is that talking about `optional` fields from `proto3`?

> Any other concerns?

- The verbosity of the build script is annoying but that shouldn't block a transition.
- This is a bit concerning: https://github.com/tafia/quick-protobuf/pull/153. I am not liking the fact that there is `unsafe` in (some of) the generated code.

Overall, I like that we can use `Cow` in a few places now and overall, interacting with the generated code seems a lot simpler!

It also seems that @snproj has recently been active on the repository as a maintainer? I hope you don't mind being tagged! :)

I'd like to hear @mxinden's opinion before we move forward here. Personally I am in favor assuming the above `unsafe` problem is fixed.

dignifiedquire:
> This is a bit concerning: https://github.com/tafia/quick-protobuf/pull/153. I am not liking the fact that there is unsafe in (some of) the generated code.

Looks like there was some good activity and this PR and a follow up got merged and released in the last days

snproj:
> It also seems that @snproj has recently been active on the repository as a maintainer? I hope you don't mind being tagged! :)

Hi! Didn't see this earlier 😅 Just popping in to say yep, my [org](https://github.com/ghpr-asia) intends to maintain `quick-protobuf`. Let us know if you've got any concerns if you choose to migrate!

Since you mentioned the generated API, could I ~~use that as an excuse to~~ ask if you've got any opinions on [this](https://github.com/tafia/quick-protobuf/issues/235)?

> I am not quite sure what `option` refers too? Is that talking about `optional` fields from `proto3`?

I think it refers to [this](https://developers.google.com/protocol-buffers/docs/proto3#options), which is currently being ignored by `pb-rs/src/parser.rs > option_ignore()`. Although the `proto3` distinction between `optional` and default (no keyword) fields isn't really supported yet; we have yet to implement hazzers.

kckeiks:
I'll update this PR very soon.

kckeiks:
This test is failing because `quick_protobuf::deserialize_from_slice` is returning a VarInt error when calling `from_protobuf_encoding`. I noticed that outputs from `KeyPair::to_protobuf_encoding` have the value 68 at the start then followed by 8. As an experiment, I prepended `encoded` with 68 and `deserialize_from_slice` did not fail (but the peer ids did not match).

Does anyone know what the issue might be? Will investigate later today or tomorrow....




 
```
    #[test]
    fn keypair_from_protobuf_encoding() {
        // E.g. retrieved from an IPFS config file.
        let base_64_encoded = "CAESQL6vdKQuznQosTrW7FWI9At+XX7EBf0BnZLhb6w+N+XSQSdfInl6c7U4NuxXJlhKcRBlBw9d0tj2dfBIVf6mcPA=";
        let expected_peer_id =
            PeerId::from_str("12D3KooWEChVMMMzV8acJ53mJHrw1pQ27UAGkCxWXLJutbeUMvVu").unwrap();
        let encoded = base64::decode(base_64_encoded).unwrap();
        let peer_id = keypair.public().to_peer_id();
    }
```

kckeiks:
> This test is failing because `quick_protobuf::deserialize_from_slice` is returning a VarInt error when calling `from_protobuf_encoding`. I noticed that outputs from `KeyPair::to_protobuf_encoding` have the value 68 at the start then followed by 8. As an experiment, I prepended `encoded` with 68 and `deserialize_from_slice` did not fail (but the peer ids did not match).
> 
> Does anyone know what the issue might be? Will investigate later today or tomorrow....
> 
> ```
>     #[test]
>     fn keypair_from_protobuf_encoding() {
>         // E.g. retrieved from an IPFS config file.
>         let base_64_encoded = "CAESQL6vdKQuznQosTrW7FWI9At+XX7EBf0BnZLhb6w+N+XSQSdfInl6c7U4NuxXJlhKcRBlBw9d0tj2dfBIVf6mcPA=";
>         let expected_peer_id =
>             PeerId::from_str("12D3KooWEChVMMMzV8acJ53mJHrw1pQ27UAGkCxWXLJutbeUMvVu").unwrap();
>         let encoded = base64::decode(base_64_encoded).unwrap();
>         let keypair = Keypair::from_protobuf_encoding(&encoded).unwrap();
>         let peer_id = keypair.public().to_peer_id();
>         assert_eq!(expected_peer_id, peer_id);
>     }
> ```

https://github.com/tafia/quick-protobuf/issues/202

core/src/identity.rs
kckeiks: Seem like `encode()` returns an array which creates a temporary that goes out of scope so we can't remove the allocation.

core/src/protos/envelope_proto.rs
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    fn write_message<W: WriterBackend>(&self, w: &mut Writer<W>) -> Result<()> {
+        if let Some(ref s) = self.public_key { w.write_with_tag(10, |w| w.write_message(s))?; }
+        if self.payload_type != Cow::Borrowed(b"") { w.write_with_tag(18, |w| w.write_bytes(&**&self.payload_type))?; }
+        if self.payload != Cow::Borrowed(b"") { w.write_with_tag(26, |w| w.write_bytes(&**&self.payload))?; }
+        if self.signature != Cow::Borrowed(b"") { w.write_with_tag(42, |w| w.write_bytes(&**&self.signature))?; }
+        Ok(())
+    }
+}
```
thomaseizinger, line 59: Wow this is actually so decent to look at!

kckeiks: haha yup

core/build.rs
```diff
@@ -1,49 +0,0 @@
...
-// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
-// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
-// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
-// FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
-// DEALINGS IN THE SOFTWARE.
-
-use pb_rs::{types::FileDescriptor, ConfigBuilder};
-use std::path::{Path, PathBuf};
-
-fn main() {
```
thomaseizinger, line 21: We will need a CI check that ensures that the generated code is not modified!

Ideally, the CI check can use glob patterns so we auto-detect when new `.proto` files are added to the codebase.

thomaseizinger:
> > This test is failing because `quick_protobuf::deserialize_from_slice` is returning a VarInt error when calling `from_protobuf_encoding`. I noticed that outputs from `KeyPair::to_protobuf_encoding` have the value 68 at the start then followed by 8. As an experiment, I prepended `encoded` with 68 and `deserialize_from_slice` did not fail (but the peer ids did not match).
> > Does anyone know what the issue might be? Will investigate later today or tomorrow....
> > ```
> >     #[test]
> >     fn keypair_from_protobuf_encoding() {
> >         // E.g. retrieved from an IPFS config file.
> >         let base_64_encoded = "CAESQL6vdKQuznQosTrW7FWI9At+XX7EBf0BnZLhb6w+N+XSQSdfInl6c7U4NuxXJlhKcRBlBw9d0tj2dfBIVf6mcPA=";
> >         let expected_peer_id =
> >             PeerId::from_str("12D3KooWEChVMMMzV8acJ53mJHrw1pQ27UAGkCxWXLJutbeUMvVu").unwrap();
> >         let encoded = base64::decode(base_64_encoded).unwrap();
> >         let keypair = Keypair::from_protobuf_encoding(&encoded).unwrap();
> >         let peer_id = keypair.public().to_peer_id();
> >         assert_eq!(expected_peer_id, peer_id);
> >     }
> > ```
> 
> [tafia/quick-protobuf#202](https://github.com/tafia/quick-protobuf/issues/202)

Well that is a bit of a weird one. Does the workaround mentioned in the issue solve things or are we blocked here?

kckeiks:
> > > This test is failing because `quick_protobuf::deserialize_from_slice` is returning a VarInt error when calling `from_protobuf_encoding`. I noticed that outputs from `KeyPair::to_protobuf_encoding` have the value 68 at the start then followed by 8. As an experiment, I prepended `encoded` with 68 and `deserialize_from_slice` did not fail (but the peer ids did not match).
> > > Does anyone know what the issue might be? Will investigate later today or tomorrow....
> > > ```
> > >     #[test]
> > >     fn keypair_from_protobuf_encoding() {
> > >         // E.g. retrieved from an IPFS config file.
> > >         let base_64_encoded = "CAESQL6vdKQuznQosTrW7FWI9At+XX7EBf0BnZLhb6w+N+XSQSdfInl6c7U4NuxXJlhKcRBlBw9d0tj2dfBIVf6mcPA=";
> > >         let expected_peer_id =
> > >             PeerId::from_str("12D3KooWEChVMMMzV8acJ53mJHrw1pQ27UAGkCxWXLJutbeUMvVu").unwrap();
> > >         let encoded = base64::decode(base_64_encoded).unwrap();
> > >         let keypair = Keypair::from_protobuf_encoding(&encoded).unwrap();
> > >         let peer_id = keypair.public().to_peer_id();
> > >         assert_eq!(expected_peer_id, peer_id);
> > >     }
> > > ```
> > 
> > 
> > [tafia/quick-protobuf#202](https://github.com/tafia/quick-protobuf/issues/202)
> 
> Well that is a bit of a weird one. Does the workaround mentioned in the issue solve things or are we blocked here?

Not a blocker. We're good as long as we only use `deserialize_from_slice` and `serialize_from_slice`.

core/build.rs
kckeiks: I'm not too familiar with Github actions so I have some questions. Do we want the build to fail when these files are modified? Is there a way to turn these off when we legitimately need to update these files? Also, I think when someone adds a new generated file and .proto file, the build should not fail?

Another option is to make a warning comment in the PR when proto/* files are added or modified. Another simpler approach which I've seen some projects do ([mitmproxy](https://github.com/mitmproxy/mitmproxy/tree/main/mitmproxy/contrib/kaitaistruct) for example) is to add these `proto/` directories to `.gitignore`. 

Thoughts?

Possible solutions:
- https://www.meziantou.net/executing-github-actions-jobs-or-steps-only-when-specific-files-change.htm
- https://github.com/marketplace/actions/comment-pull-request

thomaseizinger: I think we should never modify these files. I was thinking something along the lines of:

- Run protobuf compiler with some glob pattern capturing all `.proto` files
- Run `git diff --exit-code`

If anyone modified a generated file, `git diff` will fail the workflow.

This way, CI will only pass if there is a generated code file for each `.proto` file that is completely untouched, resulting in the exact same behaviour as if we were to generate it at compile-time.

core/CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+# 0.39.0 [unreleased]
```
```suggestion
# 0.38.1 [unreleased]
```

I'd hope that we can ship this as a backwards-compatible change.

transports/noise/src/lib.rs
```diff
@@ -57,12 +57,15 @@
 
 mod io;
 mod protocol;
+#[allow(clippy::derive_partial_eq_without_eq)]
```
thomaseizinger: This is also applied within the file, isn't it?

transports/noise/CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+# 0.42.0 [unreleased]
```
thomaseizinger, line 1:
```suggestion
# 0.41.1 [unreleased]
```

Same here, it should be a backwards-compatible change.

core/Cargo.toml
```diff
@@ -54,7 +54,7 @@ rmp-serde = "1.0"
 serde_json = "1.0"
 
 [build-dependencies]
-prost-build = "0.11"
+pb-rs = "0.10.0"
```
thomaseizinger: We should no longer need this, right?

thomaseizinger (commented):
Let me know when this ready for another full review!

thomaseizinger:
Please merge master instead of force-pushing. We squash merge anyway and it makes reviews easier :)

kckeiks:
It's ready for another review. Just pushed https://github.com/libp2p/rust-libp2p/pull/3066/commits/df4d1c13878bbfe94ff7586a939559650823df25 and https://github.com/libp2p/rust-libp2p/pull/3066/commits/afe818ef98bb4aeedc0681d20c61d0c4025e9ced.

core/src/identity.rs
thomaseizinger: You should be able to add `pub(crate) fn as_bytes(&self) -> &[u8]` to `ed25519::PublicKey` which then allows the use of `Cow::Borrowed`.

core/src/identity.rs
```diff
@@ -349,9 +333,9 @@ mod tests {
     #[test]
     fn keypair_from_protobuf_encoding() {
         // E.g. retrieved from an IPFS config file.
-        let base_64_encoded = "CAESQL6vdKQuznQosTrW7FWI9At+XX7EBf0BnZLhb6w+N+XSQSdfInl6c7U4NuxXJlhKcRBlBw9d0tj2dfBIVf6mcPA=";
+        let base_64_encoded = "RAgBEkC+r3SkLs50KLE61uxViPQLfl1+xAX9AZ2S4W+sPjfl0kEnXyJ5enO1ODbsVyZYSnEQZQcPXdLY9nXwSFX+pnDw";
         let expected_peer_id =
-            PeerId::from_str("12D3KooWEChVMMMzV8acJ53mJHrw1pQ27UAGkCxWXLJutbeUMvVu").unwrap();
```
thomaseizinger, line 354: Why does this test change? We should not need to change this test.

core/src/identity/error.rs
```diff
@@ -68,13 +68,6 @@ impl DecodingError {
         }
     }
 
-    pub(crate) fn unknown_key_type(key_type: i32) -> Self {
-        Self {
-            msg: format!("unknown key-type {key_type}"),
-            source: None,
-        }
-    }
```
thomaseizinger, line 76: :tada:

protocols/autonat/src/protocol.rs
```diff
@@ -18,12 +18,12 @@
 // FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
 // DEALINGS IN THE SOFTWARE.
 
-use crate::structs_proto;
+use crate::protos::structs as structs_proto;
```
thomaseizinger, line 21: I am not a fan of renaming modules on import. It would be cool if we could just access all structs under a `proto` namespace.

To achieve, would it make sense if we:

1. Rename the `protos` module to `generated`
2. Create a `proto` module in `lib.rs` as `mod proto {}`
3. Re-export all types that we need from there to flatten out the module hierarchy.

thomaseizinger (commented):
Nice, this is definitely heading in the right direction! Thank you!

Things to look out for:

- Before we convert more crates, I'd like to verify that we can reasonably easy write a CI script that checks the generated code for modifications. I've just tried it locally and `pb-rs **/*.proto` seems to do the trick! I'd like to see this rather sooner than later in the progress of this PR. I think a good next step would be to move all proto files in the directories we want, generated the according Rust code and add the CI job that verifies it all. From there, we can continue to convert more crates to actually use the newly generated code.
- We currently rely on `prost-codec` in various crates. Now that we are no longer using `prost`, that will have to renamed (and re-released) to `quick-protobuf-codec`. I'd like to verify that we can actually build the same abstraction, i.e. an implementation of `asynchronous-codec::Encoder` and `asynchronous-codec::Decoder` for structs generated using `quick-protobuf`.
- In addition to the CI script, we can remove the installation of `protoc` from all workflows.

Thanks again for tackling this, I am very excited to land this improvement. It is a big plus for our CI pipeline and removes an annoying step for people to get starting hacking on `rust-libp2p`.

core/src/identity.rs
kckeiks: This changes because of this https://github.com/libp2p/rust-libp2p/pull/3066#issuecomment-1340132206. I've been thinking about this and feel uneasy about it. Should we instead use this pattern https://github.com/tafia/quick-protobuf/issues/202#issuecomment-1093701486? That way we're consistent with the official encoding.

thomaseizinger: > Should we instead use this pattern [tafia/quick-protobuf#202 (comment)](https://github.com/tafia/quick-protobuf/issues/202#issuecomment-1093701486)? That way we're consistent with the official encoding.

I think so yes. We need to be compatible here so this test should not need changing!

kckeiks:
I will have some bandwidth at the end of this week to work on this.

thomaseizinger:
> I will have some bandwidth at the end of this week to work on this.

As a heads-up: Response times on our end might be slow over the next 2-3 weeks. Personally, I'll be pretty much off the grid for about two weeks from next Monday.

kckeiks:
Moving to https://github.com/libp2p/rust-libp2p/pull/3312

thomaseizinger:
> * We currently rely on `prost-codec` in various crates. Now that we are no longer using `prost`, that will have to renamed (and re-released) to `quick-protobuf-codec`. I'd like to verify that we can actually build the same abstraction, i.e. an implementation of `asynchronous-codec::Encoder` and `asynchronous-codec::Decoder` for structs generated using `quick-protobuf`.

@mxinden I've been thinking. Instead of releasing another crate, would you accept a PR to https://github.com/mxinden/asynchronous-codec with a feature-flag for decoding protobuf messages via `quick-protobuf`? The library already has support for JSON and CBOR so we could also add other formats to it. That would save us from maintaining and releasing one more crate.

mxinden:
> > * We currently rely on `prost-codec` in various crates. Now that we are no longer using `prost`, that will have to renamed (and re-released) to `quick-protobuf-codec`. I'd like to verify that we can actually build the same abstraction, i.e. an implementation of `asynchronous-codec::Encoder` and `asynchronous-codec::Decoder` for structs generated using `quick-protobuf`.
> 
> @mxinden I've been thinking. Instead of releasing another crate, would you accept a PR to https://github.com/mxinden/asynchronous-codec with a feature-flag for decoding protobuf messages via `quick-protobuf`? The library already has support for JSON and CBOR so we could also add other formats to it. That would save us from maintaining and releasing one more crate.

Sure thing! Happy to merge a patch into `asynchronous-codec`. I think what was holding me back back then and what made me introduce a new crate (`prost-codec`) instead, was the fact that we prefix Protobufs with a varint, which at the same was a libp2p concept to me. Thinking about it some more, might as well be a general useful thing and thus worth merging into `asynchronous-codec` directly.

thomaseizinger:
> > > * We currently rely on `prost-codec` in various crates. Now that we are no longer using `prost`, that will have to renamed (and re-released) to `quick-protobuf-codec`. I'd like to verify that we can actually build the same abstraction, i.e. an implementation of `asynchronous-codec::Encoder` and `asynchronous-codec::Decoder` for structs generated using `quick-protobuf`.
> > 
> > 
> > @mxinden I've been thinking. Instead of releasing another crate, would you accept a PR to [mxinden/asynchronous-codec](https://github.com/mxinden/asynchronous-codec) with a feature-flag for decoding protobuf messages via `quick-protobuf`? The library already has support for JSON and CBOR so we could also add other formats to it. That would save us from maintaining and releasing one more crate.
> 
> Sure thing! Happy to merge a patch into `asynchronous-codec`. I think what was holding me back back then and what made me introduce a new crate (`prost-codec`) instead, was the fact that we prefix Protobufs with a varint, which at the same was a libp2p concept to me. Thinking about it some more, might as well be a general useful thing and thus worth merging into `asynchronous-codec` directly.

I see your concern but given that protobuf serializations themselves aren't self-descriptive in that way, some strategy is needed. I think exposing a codec that always uses varint for that is a sensible choice. We can try and generalize that once someone comes along with a different usecase.

@kckeiks Would you be up for extending `asynchronous-codec` with this discussed feature?

kckeiks:
> > > > * We currently rely on `prost-codec` in various crates. Now that we are no longer using `prost`, that will have to renamed (and re-released) to `quick-protobuf-codec`. I'd like to verify that we can actually build the same abstraction, i.e. an implementation of `asynchronous-codec::Encoder` and `asynchronous-codec::Decoder` for structs generated using `quick-protobuf`.
> > > 
> > > 
> > > @mxinden I've been thinking. Instead of releasing another crate, would you accept a PR to [mxinden/asynchronous-codec](https://github.com/mxinden/asynchronous-codec) with a feature-flag for decoding protobuf messages via `quick-protobuf`? The library already has support for JSON and CBOR so we could also add other formats to it. That would save us from maintaining and releasing one more crate.
> > 
> > 
> > Sure thing! Happy to merge a patch into `asynchronous-codec`. I think what was holding me back back then and what made me introduce a new crate (`prost-codec`) instead, was the fact that we prefix Protobufs with a varint, which at the same was a libp2p concept to me. Thinking about it some more, might as well be a general useful thing and thus worth merging into `asynchronous-codec` directly.
> 
> I see your concern but given that protobuf serializations themselves aren't self-descriptive in that way, some strategy is needed. I think exposing a codec that always uses varint for that is a sensible choice. We can try and generalize that once someone comes along with a different usecase.
> 
> @kckeiks Would you be up for extending `asynchronous-codec` with this discussed feature?

Why not :)

kckeiks:
> > > * We currently rely on `prost-codec` in various crates. Now that we are no longer using `prost`, that will have to renamed (and re-released) to `quick-protobuf-codec`. I'd like to verify that we can actually build the same abstraction, i.e. an implementation of `asynchronous-codec::Encoder` and `asynchronous-codec::Decoder` for structs generated using `quick-protobuf`.
> > 
> > 
> > @mxinden I've been thinking. Instead of releasing another crate, would you accept a PR to https://github.com/mxinden/asynchronous-codec with a feature-flag for decoding protobuf messages via `quick-protobuf`? The library already has support for JSON and CBOR so we could also add other formats to it. That would save us from maintaining and releasing one more crate.
> 
> Sure thing! Happy to merge a patch into `asynchronous-codec`. I think what was holding me back back then and what made me introduce a new crate (`prost-codec`) instead, was the fact that we prefix Protobufs with a varint, which at the same was a libp2p concept to me. Thinking about it some more, might as well be a general useful thing and thus worth merging into `asynchronous-codec` directly.

@thomaseizinger @mxinden I'm running into an issue over in [mxinden/asynchronous-codec](https://github.com/mxinden/asynchronous-codec). `prost-codec::Codec` uses `UviBytes` which depends on `asynchronous-codec` so I'm running into a conflict with the traits.... I was thinking of replacing `UviBytes` with `LengthCodec` but `UviBytes` looks more robust. Any suggestions?

thomaseizinger:
> > > > * We currently rely on `prost-codec` in various crates. Now that we are no longer using `prost`, that will have to renamed (and re-released) to `quick-protobuf-codec`. I'd like to verify that we can actually build the same abstraction, i.e. an implementation of `asynchronous-codec::Encoder` and `asynchronous-codec::Decoder` for structs generated using `quick-protobuf`.
> > > 
> > > 
> > > @mxinden I've been thinking. Instead of releasing another crate, would you accept a PR to [mxinden/asynchronous-codec](https://github.com/mxinden/asynchronous-codec) with a feature-flag for decoding protobuf messages via `quick-protobuf`? The library already has support for JSON and CBOR so we could also add other formats to it. That would save us from maintaining and releasing one more crate.
> > 
> > 
> > Sure thing! Happy to merge a patch into `asynchronous-codec`. I think what was holding me back back then and what made me introduce a new crate (`prost-codec`) instead, was the fact that we prefix Protobufs with a varint, which at the same was a libp2p concept to me. Thinking about it some more, might as well be a general useful thing and thus worth merging into `asynchronous-codec` directly.
> 
> @thomaseizinger @mxinden I'm running into an issue over in [mxinden/asynchronous-codec](https://github.com/mxinden/asynchronous-codec). `prost-codec::Codec` uses `UviBytes` which depends on `asynchronous-codec` so I'm running into a conflict with the traits.... 

Ah damn. I see three solutions:

1. Don't activate the `asynchronous_codec` codec feature in `uvint` and re-implement the codec in `asynchronous-codec` so you can use it for the new protobuf-codec as an implementation detail
2. Do (1) but also make it publicly accessible and send a PR to `uvint` that removes the codec from there
3. Create a dedicated crate as we did before where we can combine the two

Looking at it from first principles, (3) is the cleanest but requires us to have another crate again. Really, `asynchronous-codec` should probably not offer any codecs that require dependencies outside of the standard library. I'd say lets go with (3) and have a crate `quick-protobuf-codec` in `misc`, replacing `prost-codec`.

> I was thinking of replacing `UviBytes` with `LengthCodec` but `UviBytes` looks more robust. Any suggestions?

We can't do that because it would not be compatible on the network layer with other peers / implementations.

kckeiks:
> > > > > * We currently rely on `prost-codec` in various crates. Now that we are no longer using `prost`, that will have to renamed (and re-released) to `quick-protobuf-codec`. I'd like to verify that we can actually build the same abstraction, i.e. an implementation of `asynchronous-codec::Encoder` and `asynchronous-codec::Decoder` for structs generated using `quick-protobuf`.
> > > > 
> > > > 
> > > > @mxinden I've been thinking. Instead of releasing another crate, would you accept a PR to [mxinden/asynchronous-codec](https://github.com/mxinden/asynchronous-codec) with a feature-flag for decoding protobuf messages via `quick-protobuf`? The library already has support for JSON and CBOR so we could also add other formats to it. That would save us from maintaining and releasing one more crate.
> > > 
> > > 
> > > Sure thing! Happy to merge a patch into `asynchronous-codec`. I think what was holding me back back then and what made me introduce a new crate (`prost-codec`) instead, was the fact that we prefix Protobufs with a varint, which at the same was a libp2p concept to me. Thinking about it some more, might as well be a general useful thing and thus worth merging into `asynchronous-codec` directly.
> > 
> > 
> > @thomaseizinger @mxinden I'm running into an issue over in [mxinden/asynchronous-codec](https://github.com/mxinden/asynchronous-codec). `prost-codec::Codec` uses `UviBytes` which depends on `asynchronous-codec` so I'm running into a conflict with the traits....
> 
> Ah damn. I see three solutions:
> 
>     1. Don't activate the `asynchronous_codec` codec feature in `uvint` and re-implement the codec in `asynchronous-codec` so you can use it for the new protobuf-codec as an implementation detail
> 
>     2. Do (1) but also make it publicly accessible and send a PR to `uvint` that removes the codec from there
> 
>     3. Create a dedicated crate as we did before where we can combine the two
> 
> 
> Looking at it from first principles, (3) is the cleanest but requires us to have another crate again. Really, `asynchronous-codec` should probably not offer any codecs that require dependencies outside of the standard library. I'd say lets go with (3) and have a crate `quick-protobuf-codec` in `misc`, replacing `prost-codec`.
> 
> > I was thinking of replacing `UviBytes` with `LengthCodec` but `UviBytes` looks more robust. Any suggestions?
> 
> We can't do that because it would not be compatible on the network layer with other peers / implementations.

Sounds good. I prefer (3) as well.

thomaseizinger:
Closing here in favor of https://github.com/libp2p/rust-libp2p/pull/3312.

PR #3070, opened 2022-10-30
protocols/gossipsub: Make use of `prost-codec`

nloadholtes:
# Description

This is to address issue #2500 (doing one PR per crate)

- For this PR, I have referenced the work done in `protocols/dcutr` and modeled this work on what was done there. I tried a few other approaches, but am submitting this one because:
  - The impact of the changes seems minimal compared to trying to get `GossipsubCodec` to "inherit" the `prost_codec::Codec` struct/traits. (Doing that seemed to touch larger number of lines/files, and I never got it to compile which I took as a warning I was going in the wrong direction.) 
  - This approach seems to work from a test/compiler-messages point of view
  - I believe this code looks more like the `protocol/dcutr` code, and I am biased towards making a code base look consistent with itself.
- I am still new to Rust, comments/suggestions/requests/opinions are welcomed :slightly_smiling_face: 
- Locally the tests pass and clippy doesn't complain

#2500 

## Change checklist
_To be completed after the above questions are answered_

- [ ] I have added tests that prove my fix is effective or that my feature works -- N/A, this is a refactor and existing tests are untouched.

Depends-On: https://github.com/libp2p/rust-libp2p/pull/3081

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -75,7 +75,6 @@
   - Update to [`libp2p-wasm-ext` `v0.38.0`](transports/wasm-ext/CHANGELOG.md#0380).
   - Update to [`libp2p-websocket` `v0.40.0`](transports/websocket/CHANGELOG.md#0400).
   - Update to [`libp2p-yamux` `v0.42.0`](muxers/yamux/CHANGELOG.md#0420).
-  - Update to [`libp2p-gossipsub`_`v0.43.0`](protocols/gossipsub/CHANGELOG.md#0430).
 
 [PR 2945]: https://github.com/libp2p/rust-libp2p/pull/2945
 [PR 3001]: https://github.com/libp2p/rust-libp2p/pull/3001
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -75,6 +75,7 @@
   - Update to [`libp2p-wasm-ext` `v0.38.0`](transports/wasm-ext/CHANGELOG.md#0380).
   - Update to [`libp2p-websocket` `v0.40.0`](transports/websocket/CHANGELOG.md#0400).
   - Update to [`libp2p-yamux` `v0.42.0`](muxers/yamux/CHANGELOG.md#0420).
+  - Update to [`libp2p-gossipsub`_`v0.43.0`](protocols/gossipsub/CHANGELOG.md#0430).
 
 [PR 2945]: https://github.com/libp2p/rust-libp2p/pull/2945
 [PR 3001]: https://github.com/libp2p/rust-libp2p/pull/3001
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -66,11 +66,7 @@ request-response = ["dep:libp2p-request-response"]
 rendezvous = ["dep:libp2p-rendezvous"]
 tcp = ["dep:libp2p-tcp"]
 uds = ["dep:libp2p-uds"]
-wasm-bindgen = [
-    "futures-timer/wasm-bindgen",
-    "instant/wasm-bindgen",
-    "getrandom/js",
-]
+wasm-bindgen = ["futures-timer/wasm-bindgen", "instant/wasm-bindgen", "getrandom/js"]
 wasm-ext = ["dep:libp2p-wasm-ext"]
 wasm-ext-websocket = ["wasm-ext", "libp2p-wasm-ext?/websocket"]
 websocket = ["dep:libp2p-websocket"]
@@ -80,18 +76,14 @@ rsa = ["libp2p-core/rsa"]
 ecdsa = ["libp2p-core/ecdsa"]
 serde = ["libp2p-core/serde", "libp2p-kad?/serde", "libp2p-gossipsub?/serde"]
 tokio = ["libp2p-mdns?/tokio", "libp2p-tcp?/tokio", "libp2p-dns?/tokio"]
-async-std = [
-    "libp2p-mdns?/async-io",
-    "libp2p-tcp?/async-io",
-    "libp2p-dns?/async-std",
-]
+async-std = ["libp2p-mdns?/async-io", "libp2p-tcp?/async-io", "libp2p-dns?/async-std"]
 
 [dependencies]
 bytes = "1"
 futures = "0.3.1"
 futures-timer = "3.0.2" # Explicit dependency to be used in `wasm-bindgen` feature
-getrandom = "0.2.3"     # Explicit dependency to be used in `wasm-bindgen` feature
-instant = "0.1.11"      # Explicit dependency to be used in `wasm-bindgen` feature
+getrandom = "0.2.3" # Explicit dependency to be used in `wasm-bindgen` feature
+instant = "0.1.11" # Explicit dependency to be used in `wasm-bindgen` feature
 lazy_static = "1.2"
 
 libp2p-autonat = { version = "0.9.0", path = "protocols/autonat", optional = true }
@@ -135,13 +127,7 @@ async-std = { version = "1.6.2", features = ["attributes"] }
 async-trait = "0.1"
 env_logger = "0.9.0"
 clap = { version = "4.0.13", features = ["derive"] }
-tokio = { version = "1.15", features = [
-    "io-util",
-    "io-std",
-    "macros",
-    "rt",
-    "rt-multi-thread",
-] }
+tokio = { version = "1.15", features = ["io-util", "io-std", "macros", "rt", "rt-multi-thread"] }
 
 [workspace]
 members = [
@@ -177,7 +163,7 @@ members = [
     "transports/tcp",
     "transports/uds",
     "transports/websocket",
-    "transports/wasm-ext",
+    "transports/wasm-ext"
 ]
 
 [[example]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -66,7 +66,11 @@ request-response = ["dep:libp2p-request-response"]
 rendezvous = ["dep:libp2p-rendezvous"]
 tcp = ["dep:libp2p-tcp"]
 uds = ["dep:libp2p-uds"]
-wasm-bindgen = ["futures-timer/wasm-bindgen", "instant/wasm-bindgen", "getrandom/js"]
+wasm-bindgen = [
+    "futures-timer/wasm-bindgen",
+    "instant/wasm-bindgen",
+    "getrandom/js",
+]
 wasm-ext = ["dep:libp2p-wasm-ext"]
 wasm-ext-websocket = ["wasm-ext", "libp2p-wasm-ext?/websocket"]
 websocket = ["dep:libp2p-websocket"]
@@ -76,14 +80,18 @@ rsa = ["libp2p-core/rsa"]
 ecdsa = ["libp2p-core/ecdsa"]
 serde = ["libp2p-core/serde", "libp2p-kad?/serde", "libp2p-gossipsub?/serde"]
 tokio = ["libp2p-mdns?/tokio", "libp2p-tcp?/tokio", "libp2p-dns?/tokio"]
-async-std = ["libp2p-mdns?/async-io", "libp2p-tcp?/async-io", "libp2p-dns?/async-std"]
+async-std = [
+    "libp2p-mdns?/async-io",
+    "libp2p-tcp?/async-io",
+    "libp2p-dns?/async-std",
+]
 
 [dependencies]
 bytes = "1"
 futures = "0.3.1"
 futures-timer = "3.0.2" # Explicit dependency to be used in `wasm-bindgen` feature
-getrandom = "0.2.3" # Explicit dependency to be used in `wasm-bindgen` feature
-instant = "0.1.11" # Explicit dependency to be used in `wasm-bindgen` feature
+getrandom = "0.2.3"     # Explicit dependency to be used in `wasm-bindgen` feature
+instant = "0.1.11"      # Explicit dependency to be used in `wasm-bindgen` feature
 lazy_static = "1.2"
 
 libp2p-autonat = { version = "0.9.0", path = "protocols/autonat", optional = true }
@@ -127,7 +135,13 @@ async-std = { version = "1.6.2", features = ["attributes"] }
 async-trait = "0.1"
 env_logger = "0.9.0"
 clap = { version = "4.0.13", features = ["derive"] }
-tokio = { version = "1.15", features = ["io-util", "io-std", "macros", "rt", "rt-multi-thread"] }
+tokio = { version = "1.15", features = [
+    "io-util",
+    "io-std",
+    "macros",
+    "rt",
+    "rt-multi-thread",
+] }
 
 [workspace]
 members = [
@@ -163,7 +177,7 @@ members = [
     "transports/tcp",
     "transports/uds",
     "transports/websocket",
-    "transports/wasm-ext"
+    "transports/wasm-ext",
 ]
 
 [[example]]
--- a/protocols/gossipsub/CHANGELOG.md
+++ b/protocols/gossipsub/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 - Update to `libp2p-swarm` `v0.41.0`.
 
+- Refactoring GossipsubCodec to use common protobuf Codec.
+
 # 0.42.0
 
 - Bump rand to 0.8 and quickcheck to 1. See [PR 2857].
--- a/protocols/gossipsub/CHANGELOG.md
+++ b/protocols/gossipsub/CHANGELOG.md
@@ -6,7 +6,9 @@
 
 - Update to `prost-codec` `v0.3.0`.
 
-- Refactoring GossipsubCodec to use common protobuf Codec.
+- Refactoring GossipsubCodec to use common protobuf Codec. See [PR 3070]
+
+[PR 3070]: https://github.com/libp2p/rust-libp2p/pull/3070
 
 # 0.42.0
 
--- a/protocols/gossipsub/CHANGELOG.md
+++ b/protocols/gossipsub/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 - Update to `libp2p-swarm` `v0.41.0`.
 
+- Update to `prost-codec` `v0.3.0`.
+
 - Refactoring GossipsubCodec to use common protobuf Codec.
 
 # 0.42.0
--- a/protocols/gossipsub/CHANGELOG.md
+++ b/protocols/gossipsub/CHANGELOG.md
@@ -6,7 +6,7 @@
 
 - Update to `prost-codec` `v0.3.0`.
 
-- Refactoring GossipsubCodec to use common protobuf Codec. See [PR 3070]
+- Refactoring GossipsubCodec to use common protobuf Codec. See [PR 3070].
 
 [PR 3070]: https://github.com/libp2p/rust-libp2p/pull/3070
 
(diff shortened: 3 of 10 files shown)


protocols/gossipsub/src/protocol.rs
```diff
@@ -285,17 +287,21 @@ impl Decoder for GossipsubCodec {
...
 
     fn decode(&mut self, src: &mut BytesMut) -> Result<Option<Self::Item>, Self::Error> {
+        // TODO: Replace with codec::decode call
+
         let packet = match self.length_codec.decode(src).map_err(|e| {
             if let std::io::ErrorKind::PermissionDenied = e.kind() {
                 GossipsubHandlerError::MaxTransmissionSize
             } else {
-                GossipsubHandlerError::Io(e)
+                GossipsubHandlerError::Codec(prost_codec::Error::Io(e))
```
nloadholtes: This is temporary code and will probably be removed once the work on the `decode` begins in earnest.

protocols/gossipsub/src/protocol.rs
```diff
@@ -267,16 +266,19 @@ impl Encoder for GossipsubCodec {
...
-        self.length_codec
-            .encode(Bytes::from(buf), dst)
-            .map_err(|_| GossipsubHandlerError::MaxTransmissionSize)
+    fn encode(
+        &mut self,
+        item: Self::Item,
+        dst: &mut BytesMut,
+    ) -> Result<(), GossipsubHandlerError> {
+        // TODO: Replace this with a call to the codec::encode
+        let mut codec: prost_codec::Codec<Self::Item> = prost_codec::Codec::new(item.encoded_len());
```
thomaseizinger: You'd want to do this in the constructor of `GossipsubCodec`.

Also the length we pass in here is mostly a DoS attack prevention. It limits how big the buffer is that we will allocate for an _incoming_ message.

This isn't used for encoding messages so passing the encoded length in here doesn't make any sense :)

In any case, this will go away once you move that to the constructor!

protocols/gossipsub/src/protocol.rs
```diff
@@ -285,17 +287,21 @@ impl Decoder for GossipsubCodec {
     type Error = GossipsubHandlerError;
 
     fn decode(&mut self, src: &mut BytesMut) -> Result<Option<Self::Item>, Self::Error> {
+        // TODO: Replace with codec::decode call
```
thomaseizinger: Once you have `prost_codec::Codec` as a field in `GossipsubCodec`, you can access its `decode` method here.

thomaseizinger (commented):
Looks good so far!

protocols/gossipsub/src/protocol.rs
```diff
@@ -183,18 +182,24 @@ where
...
 
-impl GossipsubCodec {
-    pub fn new(length_codec: codec::UviBytes, validation_mode: ValidationMode) -> Self {
+impl<In> GossipsubCodec<In> {
+    pub fn new(
+        length_codec: codec::UviBytes,
+        validation_mode: ValidationMode,
+    ) -> GossipsubCodec<In> {
+        let mut codec = prost_codec::Codec::new(0);
         GossipsubCodec {
```
nloadholtes, line 195: @thomaseizinger (When you've got time, no rush) So, this is where things went off the rails for me earlier. I believe this is how that codec field would be initialized.

But... This seems to cause a cascade of updates to everywhere `GossipsubCodec` is referenced. When I add the `<In>` to other spots (the impl for Encoder for example) it complains that it can not find the type for `In` in the scope. One thing I find weird is that it let me get this much code in without specifying a type for `In`, but yet other references that are "further out" are now being flagged as needing a type.

I found [this SO answer](https://stackoverflow.com/a/30019078/277163) that looks like it is what I need, but when I tried implementing it, even more things started showing up as build errors. (my expectation is the compiler would handle this for me, but clearly this isn't the case.)

So my question: is this on the right path for what you were suggesting? I can keep going on this, but just the sheer volume of problems I saw come up made me pause.

nloadholtes: Quick followup note: This morning I'm seeing things differently, and I'm working on this some more.

thomaseizinger: The type parameter of `prost_codec::Codec` is the message that we are deserializing. In the context of `GossipsubCodec`, that is always the same protobuf so instead forwarding the `In` type parameter to the callers of `GossipsubCodec`, you can set it to a type within the field definition of `codec`, more specifically, you can set it to `rpc_proto::Rpc`:

```rust
codec: prost_codec::Codec<rpc_proto::Rpc>,
```

protocols/gossipsub/src/protocol.rs
```diff
@@ -182,18 +182,24 @@ where
 
 /* Gossip codec for the framing */
 
-pub struct GossipsubCodec {
+pub struct GossipsubCodec<In> {
     /// Codec to encode/decode the Unsigned varint length prefix of the frames.
     length_codec: codec::UviBytes,
```
thomaseizinger: You can remove this now, this is implied by `prost_codec::Codec`.

protocols/gossipsub/src/protocol.rs
thomaseizinger: That should get rid of most compile errors :)

Try to zoom out a bit in what we are doing here (I know this can be hard when you are just learning the language):

- `prost_codec::Codec` is an abstraction that combines a `codec::UviBytes` together with a protobuf deserializer: It is a TLV-kind of encoding: We read the length from the socket, then we read the number of bytes described in the length and then we deserialize the message into a typed data structure.
- We are doing the exact same thing already in gossipsub, just all "inlined", i.e. without an abstraction. That is why we currently have a `codec::UviBytes` field in `GossipsubCodec`.
- The goal of the issue you've taken on is to remove this code duplication and use the `prost_codec::Codec` abstraction instead of implementing it all manually in here!

thomaseizinger: Most notable, the API of `GossipsubCodec` should remain completely untouched as this is a pure refactoring. Start with that as a constraint and work from there :)

nloadholtes: Awesome advice @thomaseizinger, thank you so much! I've got that in place and tests passing locally and it doesn't look nearly as rough as it did before. :sweat_smile: I'm going to revisit in the morning to double check it, etc.

> the API of GossipsubCodec should remain completely untouched as this is a pure refactoring

Thank you for that, I was wondering that but completely forgot to ask.

thomaseizinger: Sweet, looking forward to you pushing your changes!

Cargo.toml
```diff
@@ -66,7 +66,11 @@ request-response = ["dep:libp2p-request-response"]
 rendezvous = ["dep:libp2p-rendezvous"]
 tcp = ["dep:libp2p-tcp"]
 uds = ["dep:libp2p-uds"]
-wasm-bindgen = ["futures-timer/wasm-bindgen", "instant/wasm-bindgen", "getrandom/js"]
+wasm-bindgen = [
+    "futures-timer/wasm-bindgen",
+    "instant/wasm-bindgen",
+    "getrandom/js",
+]
```
thomaseizinger: Can we undo these formatting changes please?

protocols/gossipsub/CHANGELOG.md
```diff
@@ -4,6 +4,10 @@
 
 - Update to `libp2p-swarm` `v0.41.0`.
 
+- Update to `prost-codec` `v0.3.0`.
+
+- Refactoring GossipsubCodec to use common protobuf Codec.
```
thomaseizinger: Please reference the PR for this change. See other entries.

protocols/gossipsub/src/protocol.rs
```diff
@@ -267,37 +267,28 @@ impl Encoder for GossipsubCodec {
...
-            .map_err(|_| GossipsubHandlerError::MaxTransmissionSize)
+    fn encode(
+        &mut self,
+        item: Self::Item,
+        dst: &mut BytesMut,
+    ) -> Result<(), GossipsubHandlerError> {
+        match self.codec.encode(item, dst) {
+            Ok(_) => Ok(()),
+            _ => Err(GossipsubHandlerError::MaxTransmissionSize),
+        }
```
thomaseizinger: We should be able to just use `?` here and bubble up the error.

protocols/gossipsub/src/protocol.rs
```diff
@@ -267,37 +267,28 @@ impl Encoder for GossipsubCodec {
...
-            if let std::io::ErrorKind::PermissionDenied = e.kind() {
-                GossipsubHandlerError::MaxTransmissionSize
-            } else {
-                GossipsubHandlerError::Io(e)
-            }
-        })? {
-            Some(p) => p,
-            None => return Ok(None),
+    fn decode(&mut self, src: &mut BytesMut) -> Result<Option<Self::Item>, GossipsubHandlerError> {
+        let rpc = match self.codec.decode(src) {
```
thomaseizinger: Same here, please use `?`.

thomaseizinger (changes requested):

Cargo.toml
nloadholtes: Sure thing, the formatter really wanted to do that one, but I'm ok with backing it out.

thomaseizinger: We only use `rustfmt` in here so all you need is `cargo fmt`!

Unfortunately, it does not format manifest files yet so this must have been your editor :)

protocols/gossipsub/src/protocol.rs
nloadholtes:
```suggestion
        match self.codec.encode(item, dst) {
           return Ok(self.codec.encode(item, dst)?);
        }
```
Ok, after quickly reading up on `?` and conversing with the compiler, I think that change is what is needed. But... It does not like this code:

```
^ the trait `std::convert::From<prost_codec::Error>` is not implemented for `error::GossipsubHandlerError`
    = note: the question mark operation (`?`) implicitly performs a conversion on the error value using the `From` trait
    = help: the trait `std::convert::From<std::io::Error>` is implemented for `error::GossipsubHandlerError`
    = note: required because of the requirements on the impl of `std::ops::FromResidual<std::result::Result<std::convert::Infallible, 
prost_codec::Error>>` for `std::result::Result<(), error::GossipsubHandlerError>`

```
...which makes me think it is asking me to implement a new trait. I'm not sure if that is the right way to go, or changing the Result from this encode functions, or... something else I'm completely missing. :thinking: :confused:

protocols/gossipsub/src/error.rs
```diff
@@ -87,20 +88,20 @@ impl From<SigningError> for PublishError {
...
-    /// Protocol negotiation timeout.
+    #[error("Protocol negotiation timeout.")]
     NegotiationTimeout,
-    /// Protocol negotiation failed.
+    #[error("Protocol negotiation failed.")]
     NegotiationProtocolError(ProtocolError),
-    /// IO error.
-    Io(std::io::Error),
+    #[error("Failed to encode or decode")]
+    Codec(Error),
```
thomaseizinger:
```suggestion
    Codec(#[source] Error),
```

This should fix your compile error with `?`!

Cargo.toml
nloadholtes: Yes, I recently switched to vscode and everything has gone smoothly except for the formatter. It has given me problems in another file format in a python project. :rage4:

protocols/gossipsub/src/protocol.rs
nloadholtes: Hmmm, when I have `let rpc = self.codec.decode(src)?.unwrap();` the smoke tests fail with this message:

```
thread 'libp2p-swarm-task-8' panicked at 'called `Option::unwrap()` on a `None` value', protocols/gossipsub/src/protocol.rs:284:43
```

I've tried variations on `unwrap()`, but none seem be able to `return Ok(None)`.

I've settled on this code for these reasons:
1) If the old code saw a PermissionDenied error, it was setting the error to `GossipsubHandlerError::MaxTransmissionSize` which doesn't seem right.
2) Although it isn't using the `?`, it does seem to cover the cases and return None (which allows the smoke tests to pass)

CHANGELOG.md
```diff
@@ -75,6 +75,7 @@
   - Update to [`libp2p-wasm-ext` `v0.38.0`](transports/wasm-ext/CHANGELOG.md#0380).
   - Update to [`libp2p-websocket` `v0.40.0`](transports/websocket/CHANGELOG.md#0400).
   - Update to [`libp2p-yamux` `v0.42.0`](muxers/yamux/CHANGELOG.md#0420).
+  - Update to [`libp2p-gossipsub`_`v0.43.0`](protocols/gossipsub/CHANGELOG.md#0430).
```
```suggestion
```

This entry is already present further up :)

protocols/gossipsub/CHANGELOG.md
```diff
@@ -4,6 +4,12 @@
 
 - Update to `libp2p-swarm` `v0.41.0`.
 
+- Update to `prost-codec` `v0.3.0`.
+
+- Refactoring GossipsubCodec to use common protobuf Codec. See [PR 3070]
```
thomaseizinger:
```suggestion
- Refactoring GossipsubCodec to use common protobuf Codec. See [PR 3070].
```

protocols/gossipsub/src/error.rs
```diff
@@ -146,17 +147,8 @@ impl From<std::io::Error> for PublishError {
...
-impl std::error::Error for GossipsubHandlerError {
-    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
-        match self {
-            GossipsubHandlerError::Io(io) => Some(io),
-            _ => None,
-        }
+impl From<prost_codec::Error> for GossipsubHandlerError {
+    fn from(error: prost_codec::Error) -> GossipsubHandlerError {
+        GossipsubHandlerError::Codec(error)
     }
```
thomaseizinger: You can drop this if you use `#[from]` instead of `#[source]`, sorry should have said that earlier!

protocols/gossipsub/src/protocol.rs
```diff
@@ -267,37 +267,25 @@ impl Encoder for GossipsubCodec {
...
-                GossipsubHandlerError::Io(e)
-            }
-        })? {
-            Some(p) => p,
-            None => return Ok(None),
+    fn decode(&mut self, src: &mut BytesMut) -> Result<Option<Self::Item>, GossipsubHandlerError> {
+        let rpc = match self.codec.decode(src).map_err(GossipsubHandlerError::Codec) {
+            Ok(Some(p)) => p,
+            _ => return Ok(None),
         };
```
thomaseizinger: Should be able to use `?` here because you have a `From` implementation for the errors!


```suggestion
        let rpc = match self.codec.decode(src)? {
            Some(p) => p,
            None => return Ok(None),
        };
```

protocols/gossipsub/src/error.rs
```diff
@@ -136,7 +137,7 @@ impl std::error::Error for ValidationError {}
 
 impl From<std::io::Error> for GossipsubHandlerError {
     fn from(error: std::io::Error) -> GossipsubHandlerError {
-        GossipsubHandlerError::Io(error)
+        GossipsubHandlerError::Codec(prost_codec::Error::from(error))
     }
 }
```
thomaseizinger, line 141: I haven't tested but do we still need this implementation?

thomaseizinger (commented):
Thanks!

This looks really good already, just a few more minor comments :)

thomaseizinger:
No need for force-pushes btw, we squash-merge anyway :)

protocols/gossipsub/src/protocol.rs
nloadholtes: Ahhh! Ok, that makes sense now. The `Ok()` that I had there was confusing me with the error message. Now that I've tried it this way I understand what the message was trying to tell me. Thanks!

protocols/gossipsub/src/error.rs
nloadholtes: I think so. I commented this out (and also just tried the original version with the Io) and protocols.rs and handler.rs both lit up:

```
the trait bound `GossipsubHandlerError: From<std::io::Error>` is not satisfied
the trait `From<prost_codec::Error>` is implemented for `GossipsubHandlerError`rustc[E0277](https://doc.rust-lang.org/error-index.html#E0277)
decoder.rs(11, 17): required by a bound in `asynchronous_codec::Decoder::Error`
```

nloadholtes:
Thanks for the heads up about the squashes! There was a merge conflict the other day so I've been super :eyes: making sure I'm not drifting too far out of sync. (Of course since that 1 time everything has been smooth sailing :joy: )

thomaseizinger:
> Thanks for the heads up about the squashes! There was a merge conflict the other day so I've been super eyes making sure I'm not drifting too far out of sync. (Of course since that 1 time everything has been smooth sailing joy )

You can solve the merge conflicts by just merging latest master in :)

thomaseizinger (approved):
I've pushed some final patches, this LGTM!

Thanks for contributing :)

thomaseizinger:
I'll merge once CI is green!

protocols/gossipsub/src/error.rs
```diff
@@ -22,7 +22,6 @@
 
 use libp2p_core::identity::error::SigningError;
 use libp2p_core::upgrade::ProtocolError;
-use prost_codec::Error as ProstCodecError;
```
nloadholtes: Style question: in Python we do that fairly often to disambiguate names. In rust the fully qualified name is preferred? (Just want to make sure my thinking is on the right track)

nloadholtes:
Sure thing! Thank you for all your help with this. I'm going to try and tackle another protocol (and hopefully it will be smoother on the PR 😅)

protocols/gossipsub/src/error.rs
thomaseizinger: It is more a style-preference. I prefer full-qualified names because:

- It saves an import line that is otherwise just noise.
- Common crate names are easy to remember. For example, I know that we are using a crate called `prost_codec` in this repository. If I come across a reference to `prost_codec::Error` anywhere in the source-code, my brain just goes "Ah yeah, that must be the[^1] error type of that crate".
- If a crate has an `Error` type, I can pretty much assume that it is the only `Error` type, otherwise the author would have named them differently (like `EncodeError` and `DecodeError` for example).

That is a lot of useful information to extract from a single line of code :)

On the contrary, if types are renamed on import:

- I have to interrupt my code reading flow if I come across `ProstCodecError`, go the the `use` statements and check, what it actually is.
- Not all programmers will rename it in the same pattern. Meaning I might come across a name that is unfamiliar to me, only to find out it is a type I actually know, just under a different name.

In the end, it is about efficiency in being able to understand code. Especially the last point can be quite annoying :)
When you are hunting a bug f.e. your brain is usually in "spot odd things"-mode and an unfamiliar name is "odd" but in the case of renaming types, it is usually a false-positive.

It is a style thing at the end of the day so YMMV :)

[^1]: Many crates by convention just have one error type.

thomaseizinger:
Need to merge https://github.com/libp2p/rust-libp2p/pull/3081 before we can proceed here.

thomaseizinger:
> Need to merge #3081 before we can proceed here.

I am using this opportunity to test mergify's "PR dependency" feature: https://docs.mergify.com/actions/queue/#merge-depends-on

thomaseizinger:
> > Need to merge #3081 before we can proceed here.
> 
> I am using this opportunity to test mergify's "PR dependency" feature: [docs.mergify.com/actions/queue/#merge-depends-on](https://docs.mergify.com/actions/queue/#merge-depends-on)

Cool! It works like a charm:

image

@jsx @mxinden @elenaf9 Look at this! :)

Issue #3048, opened 2022-10-20
Kademlia. New inbound substream to PeerId exceeds inbound substream limit. No older substream waiting to be reused. 


shamil-gadelshin:
## Summary

I use `start_providing` Kademlia API

Here is my error after 32 requests:

```
2022-10-20T09:54:15.538792Z  WARN libp2p_kad::handler: New inbound substream to PeerId("12D3KooWNwexVxD22CAdxv5CAkxrD3QgUqhycYAE1e1NH2sLbZ7v") exceeds inbound substream limit. No older substream waiting to be reused. Dropping new substream.    
```

This error is similar to the recent discussion: https://github.com/libp2p/rust-libp2p/discussions/2957

However, I have only 2 peers (local machine setup) and 3 seconds between `start_providing` requests. 

Here is my local `inbound_streams` buffer:

```
[...]

PR #3074, opened 2022-11-02
protocols/kademlia: Refactor stream state machines

thomaseizinger:
# Description

This PR fell out of investigating #3048. It refactors the two stream state machines to implement `Stream` which then allows us to use `stream::SelectAll` to poll them all at once.

As a bonus, we get to resolve an old TODO about not closing a substream properly.

## Change checklist


protocols/kad/src/handler.rs
```diff
@@ -786,6 +786,8 @@ where
                 }
             }
         }
+
+        debug_assert!(false, "Cannot find inbound substream for {request_id:?}")
```
mxinden, line 790: Just to test my intuition, given that we never poll an underlying inbound stream in `InboundSubstreamState::WaitingUser` state it won't close in the meantime and thus it won't be removed before we call `answer_pending_request`, correct?

mxinden (approved):
This looks good to me. Thanks @thomaseizinger .

I think long term, a lot of these state machines could be replaced by `async/await`. Though I think we have bigger fish to fry in the meantime.

protocols/kad/src/handler.rs
thomaseizinger: Yes.

Issue #3035, opened 2022-10-18
Redesign `RecordStore` API / interaction


shamil-gadelshin:
## Description

Currently, `RecordStore` trait returns `Result` only for `put` and `add_provider` operations. However, custom implementations (for example DB-based) could return errors on each operation.

I wonder whether it's possible to change this trait and add `Result` on each operation to make the trait more consistent with use cases.

## Motivation

This change will allow handling errors for implementations in an idiomatic way. 

## Current Implementation

```
    /// Gets a record from the store, given its key.
    fn get(&'a self, k: &Key) -> Option<Cow<'_, Record>>;

    /// Puts a record into the store.
    fn put(&'a mut self, r: Record) -> Result<()>;

    /// Removes the record with the given key from the store.
    fn remove(&'a mut self, k: &Key);

    /// Gets an iterator over all (value-) records currently stored.
    fn records(&'a self) -> Self::RecordsIter;

.....

```

## Are you planning to do it yourself in a pull request?

Not this time. This change could affect inner Kademlia error handling.

shamil-gadelshin:
@mxinden This issue I mentioned on the community call. 

There was also a question about splitting `RecordStore` trait into separate traits for records and providers.

rklaehn:
Yes, I noticed the same thing when trying to implement a persistent record store for [iroh](https://github.com/n0-computer/iroh). I just had to unwrap the errors to be able to conform to the interface, which is not nice.

https://github.com/libp2p/rust-libp2p/pull/3021 is also somewhat related. There is not much point in having a persistent record store for a large number of records if the behaviour pulls it all into one giant `Vec<ProviderRecord>`.

mxinden:
I would suggest we move forward here and write a first proof-of-concept. Anyone volunteering?

shamil-gadelshin:
It's easy to change an interface and alter all usages for it. I could do that. However, adding correct handling for an operation result in some cases could be too invasive in the current architecture.

For example - the call of the `store.remove` in the `PutRecordJob.poll()` doesn't assume returning (or some other handling) of an error. I could probably add `log::debug` or `log::error` in such cases but I'm not sure you will accept that because it won't propagate the error to the upper layers. Will logging errors be enough when it's not possible to return an upper-level result as in `kad::Behaviour::remove_record()`?

mxinden:
I don't think we should log errors, but instead bubble them up to the user. Why is that not an option?

[...]

PR #3076, opened 2022-11-03
kad: Change RecordStore trait interface - add results to methods.

shamil-gadelshin:
# Description
This PR starts an attempt to modify `RecordStore` trait interface. Currently, some of the operations lack the `Result` and it reduces the expressiveness of the custom implementations. Here is the [discussion](https://github.com/libp2p/rust-libp2p/issues/3035).

The initial version of the PR contains a change for the single operation - `remove` and serves as example with some open questions.

## Changes
- added `Result` for the `remove` operation of the `RecordStore` trait.
- modified usages
- modified tests

Original issue: https://github.com/libp2p/rust-libp2p/issues/3035

## Open Questions

There were made four changes for use cases of the `remove method`:
- `remove_record` from Behaviour - it seemed native to just return the result
- `get_record` from Behaviour - it seemed overkill to change the returned value to result here
-  `inject_event` - the system logic doesn't seem to mean returning a result here (or using some other means to propagate an error.
- `poll` from the `PutValueJob` also doesn't seem to mean returning a result here (or using some other means to propagate an error.

Please, verify my assumptions here. 

## Change checklist

- [ ] ~I have made corresponding changes to the documentation~ Shouldn't be required for this.
- [ ] ~I have added tests that prove my fix is effective or that my feature works.~  I modified tests.

## Commit message body
Change the interface of the `RecordStore` trait. Add missed results for its operations.

### DIFF
--- a/protocols/kad/src/record/store/memory.rs
+++ b/protocols/kad/src/record/store/memory.rs
@@ -131,8 +131,10 @@ impl<'a> RecordStore<'a> for MemoryStore {
         Ok(())
     }
 
-    fn remove(&'a mut self, k: &Key) {
+    fn remove(&'a mut self, k: &Key) -> Result<()> {
         self.records.remove(k);
+
+        Ok(())
     }
 
     fn records(&'a self) -> Self::RecordsIter {
@@ -234,7 +236,9 @@ mod tests {
             let mut store = MemoryStore::new(PeerId::random());
             assert!(store.put(r.clone()).is_ok());
             assert_eq!(Some(Cow::Borrowed(&r)), store.get(&r.key));
-            store.remove(&r.key);
+            store
+                .remove(&r.key)
+                .expect("Valid response from MemoryStore.");
             assert!(store.get(&r.key).is_none());
         }
         quickcheck(prop as fn(_))
--- a/protocols/kad/CHANGELOG.md
+++ b/protocols/kad/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 - Update to `libp2p-swarm` `v0.41.0`.
 
+- Change interface of the `RecordStore` trait. Add missed results for its operations.
+
 # 0.41.0
 
 - Remove deprecated `set_protocol_name()` from `KademliaConfig` & `KademliaProtocolConfig`.
--- a/protocols/kad/src/behaviour.rs
+++ b/protocols/kad/src/behaviour.rs
@@ -46,7 +46,7 @@ use libp2p_swarm::{
     dial_opts::{self, DialOpts},
     DialError, NetworkBehaviour, NetworkBehaviourAction, NotifyHandler, PollParameters,
 };
-use log::{debug, info, warn};
+use log::{debug, error, info, warn};
 use smallvec::SmallVec;
 use std::collections::{BTreeMap, HashSet, VecDeque};
 use std::fmt;
@@ -686,7 +686,9 @@ where
 
         if let Some(record) = self.store.get(&key) {
             if record.is_expired(Instant::now()) {
-                self.store.remove(&key)
+                if let Err(ref err) = self.store.remove(&key) {
+                    warn!("Record removal failed: {:?}", err);
+                };
             } else {
                 records.push(PeerRecord {
                     peer: None,
@@ -815,12 +817,14 @@ where
     /// This is a _local_ operation. However, it also has the effect that
     /// the record will no longer be periodically re-published, allowing the
     /// record to eventually expire throughout the DHT.
-    pub fn remove_record(&mut self, key: &record::Key) {
+    pub fn remove_record(&mut self, key: &record::Key) -> store::Result<()> {
         if let Some(r) = self.store.get(key) {
             if r.publisher.as_ref() == Some(self.kbuckets.local_key().preimage()) {
-                self.store.remove(key)
+                return self.store.remove(key);
             }
         }
+
+        Ok(())
     }
 
     /// Gets a mutable reference to the record store.
@@ -2094,7 +2098,9 @@ where
                 let record = match self.store.get(&key) {
                     Some(record) => {
                         if record.is_expired(Instant::now()) {
-                            self.store.remove(&key);
+                            if let Err(ref err) = self.store.remove(&key) {
+                                error!("Record removal failed: {:?}", err);
+                            };
                             None
                         } else {
                             Some(record.into_owned())
--- a/protocols/kad/src/behaviour/test.rs
+++ b/protocols/kad/src/behaviour/test.rs
@@ -711,7 +711,11 @@ fn put_record() {
                 );
                 assert_eq!(swarms[0].behaviour_mut().queries.size(), 0);
                 for k in records.keys() {
-                    swarms[0].behaviour_mut().store.remove(k);
+                    swarms[0]
+                        .behaviour_mut()
+                        .store
+                        .remove(k)
+                        .expect("Valid response from MemoryStore.");
                 }
                 assert_eq!(swarms[0].behaviour_mut().store.records().count(), 0);
                 // All records have been republished, thus the test is complete.
--- a/protocols/kad/src/jobs.rs
+++ b/protocols/kad/src/jobs.rs
@@ -66,6 +66,7 @@ use futures::prelude::*;
 use futures_timer::Delay;
 use instant::Instant;
 use libp2p_core::PeerId;
+use log::error;
 use std::collections::HashSet;
 use std::pin::Pin;
 use std::task::{Context, Poll};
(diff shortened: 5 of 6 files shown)


protocols/kad/src/record/store/memory.rs
```diff
@@ -131,8 +131,10 @@ impl<'a> RecordStore<'a> for MemoryStore {
         Ok(())
     }
 
-    fn remove(&'a mut self, k: &Key) {
+    fn remove(&'a mut self, k: &Key) -> Result<()> {
```
thomaseizinger, line 134: Can you explain in more detail, what benefits you are expecting to see from this?

Removing an element should be an idempotent operation in my opinion. What failure scenarios would you like to express here?

shamil-gadelshin: The current `RecordStore` implies the `MemoryStore` implementation that could be considered infallible which is not the case for most other custom implementations. Custom stores implemented with DB or file layers are always fallible for multiple reasons.

Just an example of a possible error for the `remove` operation: "No write permissions for the database file". 

Also, the `remove` operation is just an example of the whole `RecordStore` API. Specifically, the most problems I have are with the `records` operation. My database returns a `Result` for such iterations and I needed to create a special empty iterator with some error logging which is not a good practice IMHO.

The original issue contains a similar concern from other users: https://github.com/libp2p/rust-libp2p/issues/3035#issuecomment-1283866786

thomaseizinger: I understand that DB or file system APIs are fallible. But what is the `NetworkBehaviour` supposed to do with the error? Log it? If there is no reasonable error handling strategy, we might as well not allow the implementation to return one.

thomaseizinger: An implementation of `RecordStore` that interacts with a DB could also just represent a `Handle` that sends commands to the actual implementation, similar to how SQL DBs usually write to a journal first before modifying data on disk.

shamil-gadelshin: > I understand that DB or file system APIs are fallible. But what is the `NetworkBehaviour` supposed to do with the error? Log it? If there is no reasonable error handling strategy, we might as well not allow the implementation to return one.

That depends on the use case and libp2p implementation (architecture). Some use cases are compatible with the `Result` already, for example, `remove_record` from `Behaviour` will return it to the caller. Some with the current code are hard to work with and logging seems the easiest change. I agree, that it will add complexity to the design, however, IMHO it makes the API more idiomatic and reflects a broader set of use cases. 

> An implementation of RecordStore that interacts with a DB could also just represent a Handle that sends commands to the actual implementation, similar to how SQL DBs usually write to a journal first before modifying data on disk.

I'm not sure I follow here. There are a lot of ways to implement the storage indeed and some will hide the errors.

shamil-gadelshin:
Overall, this PR is a practical answer to this comment from Max: https://github.com/libp2p/rust-libp2p/issues/3035#issuecomment-1297562743

protocols/kad/src/record/store/memory.rs
thomaseizinger: > > An implementation of RecordStore that interacts with a DB could also just represent a Handle that sends commands to the actual implementation, similar to how SQL DBs usually write to a journal first before modifying data on disk.
> 
> I'm not sure I follow here. There are a lot of ways to implement the storage indeed and some will hide the errors.

Sorry for me brevity. I'll try to expand:

In general, removing an entry from a set is an idempotent operation: If it is present, remove it. If it is not present, do nothing.
IO-failures are typically ephemeral. Permission errors can silently drop the "request" and not remove the element.

This means that you could implement your `RecordStore` as a channel where calling `RecordStore::remove` doesn't actually remove the record but sends an (asynchronous) `Remove` message to the actual implementation. Same thing for insert, etc.

As long as you retain the order of the messages, your actual store of records can apply those messages and should end up in the same state as if the `RecordStore` functions would have synchronously modified the data. The difference there however is that appending one of these messages is a non-fallible operation. It is then the responsibility of whoever reads these messages to apply them correctly.

The downside of such a design is increased complexity and eventual consistency, i.e. inserting an element might not directly allow you to retrieve it again but only after the `Insert` message has been processed.

Such a design would probably keep a cache of all records in memory and update it from its actual source as the messages are processed.

I am not convinced it makes sense to make these functions fallible, given that they are called in various places where we cannot apply a useful error handling strategy.

shamil-gadelshin: This design could be implemented and has useful features (e.g.: you mentioned "handling ephemeral errors") but it also has its limitations - what if the error persists ("Disk is full.")? In this case, if we use an unbounded buffer - we will get memory overflow and if we have a limited buffer we will get the same situation with error handling  (or we will need to delete some requests). 

I understand that you see the architecture as optimized for infallible usage. My initial request was because I encountered an issue with the implementation when wrapping the DB as storage. 
Probably, if we could ask @rklaehn about his experience https://github.com/libp2p/rust-libp2p/issues/3035#issuecomment-1283866786 we could add additional arguments to our discussion.

thomaseizinger: > but it also has its limitations - what if the error persists ("Disk is full.")?

Yes that is a fair point.

I see two solutions:

1. The entire `RecordStore` could live outside of `rust-libp2p` and the `NetworkBehaviour` emits an event per incoming request that needs to be answered.
2. We provide a fallible, and potentially async `RecordStore` interface which we directly integrate into the `ConnectionHandler`. This would be similar to the ideas in #2852. If we integrate the `RecordStore` directly into the `ConnectionHandler`, having the functions fallible makes a lot more sense because we could just fail the entire stream and return an error back directly in the context of the other party's request. We are also in an `async` context there so the implementation of `RecordStore` could even query a database or other things.

(2) is quite a paradigm shift but I think it would enable a lot of usecases and cut down the codesize of `libp2p-kad` a lot.

rkuhn: One important property I’d like to uphold is that NetworkBehaviour is about networking stuff, connection handlers care about handling connections, and other things live outside these abstractions — at least in terms of code module structure. Kademlia might require an implementation of RecordStore in its constructor, and it might use the provided functions from the NetworkBehaviour or from the ConnectionHandler, but the actual RecordStore should live outside these two pieces — also regarding resource usage.

One somewhat similar case is bitswap, where our store also is structured like this, but it has a fully synchronous API and is spawned inside a dedicated thread, with communication over async channels. Failures will need to be bubbled up to the network peer, plus logging (with careful tuning of levels — only things like “disk full” warrant a WARN or ERROR, everything else should stay below INFO, or preferably below DEBUG for per-interaction events).

shamil-gadelshin: I would vote for option 2 (direct integration into the ConnectionHandler). However, it seems like a major change. So, my question is: whether there is any benefit from this PR as the preparation for that refactoring (that unlikely will make it in the nearest working plan)? Otherwise, feel free to reject it.

shamil-gadelshin: > Kademlia might require an implementation of RecordStore in its constructor

Did you mean this one? https://github.com/libp2p/rust-libp2p/blob/5b4eab7bafe6cba8fea6c806027ea680662b7cc6/protocols/kad/src/behaviour.rs#L405

mxinden: > One important property I’d like to uphold is that NetworkBehaviour is about networking stuff, connection handlers care about handling connections, and other things live outside these abstractions — at least in terms of code module structure.

Agreed.

thomaseizinger: > One important property I’d like to uphold is that NetworkBehaviour is about networking stuff, connection handlers care about handling connections, and other things live outside these abstractions — at least in terms of code module structure. Kademlia might require an implementation of RecordStore in its constructor, and it might use the provided functions from the NetworkBehaviour or from the ConnectionHandler, but the actual RecordStore should live outside these two pieces — also regarding resource usage.

The reason I am challenging it is because we pay a very big price for this separation and the gains are hard to quantify for me:

- There is a lot of message passing between `ConnectionHandler`s and `NetworkBehaviour`s, only because we can't access a shared resource directly from the handlers.
- Because of the message passing, we can't use async-await to describe our protocols, despite them being mostly request-response.

The interface to `RecordStore` could still look something like this:

```rust
trait RecordStore: Clone {
	fn get_value(&self, key: Key) -> GetValueFuture;
}
```

This would force implementations to use channels or other things to communicate with the centralised store (self receiver is not mutable and future doesn't have a lifetime). Thus, we would still uphold the "share memory by communicating" mantra and the connections themselves would only wait for a channel wake-up but not perform any other work.

At the same time though, this `RecordStore` can be referenced within each `ConnectionHandler`, the actual network protocol can be expressed in a few LoC and even adding things like timeouts for retrieving something from the store could be added trivially.

shamil-gadelshin:
I'm closing this PR as obsolete.

PR #3082, opened 2022-11-04
Use `Description` for commit message

thomaseizinger:
# Description

Whilst using the new mergify configuration, I've found that I almost always want to have the PR description as the commit message body.

Additionally, this will incentivize us to edit the description to something nice before merging which helps future folks who find the PR via the changelog for example.

Quite often, I also find myself to wanting to make "notes" or remarks about the PR itself, for example, that patch-by-patch review is recommended or why I am requesting a review from certain people. Those would be unnecessary in the commit message so I am also introducing a `Notes` section.

## Notes

Requesting a review from all maintainers primarily for visibility and because it is somewhat a change in process.

## Change checklist


jxs (approved):

.github/pull_request_template.md
```diff
@@ -1,12 +1,16 @@
-# Description
+## Description
 
 <!-- Please write a summary of your changes and why you made them.-->
```
elenaf9, line 3: Will these comments be visible in the commit message body?

thomaseizinger: Nope. See https://docs.mergify.com/configuration/#template:

> By default, the HTML comments are stripped from body. To get the full body, you can use the body_raw attribute.

mxinden (approved):

PR #3083, opened 2022-11-04
chore: Enforce conventional commit PR title

### LINKED ISSUE #3054: Decision required: Change commit messages to "conventional commits"
I'd like to propose to change our commit message convention to "conventional commits": https://www.conventionalcommits.org/en/v1.0.0/

The main driver for this is automation in the form of automated changelogs and automated version bumps. See https://github.com/libp2p/rust-libp2p/issues/2902.

In particular, tools like https://github.com/googleapis/release-please can do this but there are many others and I'd like to make this decision independent of which tool we are going to use.

At the moment, the convention for our commit messages is:

- A prefix, which modules the PR touches (PR == commit message for us because we squash-merge PRs)
- A short description on what the PR does

With conventional commits, we would change this to:

- A prefix indicating the _kind_ of change, e.g. `feat`, `fix`, `chore`, `test` etc
- A short description on what the PR does

See some examples here: https://www.conventionalcommits.org/en/v1.0.0/#examples

I benefits I see are pretty much the ones mentioned in the specification: https://www.conventionalcommits.org/en/v1.0.0/#why-use-conventional-commits

Something that works really well with our current workflow is the squash-merging. It means we don't have to educate users in how to write commit messages but just edit the titles of their PRs to match our conventions. That is not very hard and something one of us can do as we review it. The only important bit is that we need to review that title (and whether the change is breaking) before merging. I'd suggest we use the `!` instead of `BREAKING CHANGE` to identify those.

There are also some downsides to conventional commits:

There is an argument that changelogs should be written manually and not auto-generated. The argument is mostly that changelogs are for humans and thus should be written with priority and clarity in mind, something that is hard to do in an automated fashion. There is more criticism here: https://common-changelog.org/#42-conventional-commits

I think the provided criticism is somewhat valid but can also be dealt with:

1. Our mapping of changelog entry to PR is already 1-to-1.
2. `release-please` differentiates between e.g. between `chore` and `feat` commits. `chore` PRs will [not show up](https://github.com/googleapis/release-please#release-please-bot-does-not-create-a-release-pr-why) in the changelog. This avoids the "noise" problem.
3. We can always manually sort the changelog after a release by priority.
4. We already do and should continue to write PR titles really descriptive and in imperative form.

@mxinden @jxs @elenaf9 Please voice your opinion on this topic here :)

thomaseizinger:
## Description

Forces PR titles to follow the "conventional commit" specification.

Currently, mergify will set the PR title as the commit message when squashing: https://github.com/libp2p/rust-libp2p/blob/f9b4af3d9d5b12b20756e496349b0866baa862da/.github/mergify.yml#L6

Based on the commit message, we can use tools like `release-please` to automatically generate changelogs and bump versions. Enforcing those rules brings us one step closer to that.

I choose the following set of prefixes:

- `fix`: A bugfix, resulting in a semver patch bump
- `feat`: A feature, resulting in a semver minor bump
- `deps`: Dependabot updates

Only these represent ["releasable units"](https://github.com/googleapis/release-please#release-please-bot-does-not-create-a-release-pr-why) for `release-please`. 

- `refactor`: Useful to indicate that a PR is a pure refactoring and should not change functionality
- `docs`: Documentation update only
- `chore`: Anything else

Fixes #3054.

## Change checklist


mxinden (approved):
Codifying instead of documenting this is perfect!

.github/mergify.yml
```diff
@@ -23,6 +23,24 @@ pull_request_rules:
...
       queue:
 
+  - name: Conventional Commit PR title
+    conditions:
+      - base=master
+    actions:
+      post_check:
+        success_conditions:
+          - "title~=^(fix|feat|docs|refactor|chore|deps)(?:\\(.+\\))?:"
+        title: |
```
elenaf9:
```suggestion
          - "title~=^(fix|feat|docs|refactor|chore|deps)(?:\\(.+\\))?!?:"
        title: |
```
I think we want to allow `!` to mark breaking changes, right?

thomaseizinger: Ah you are right! Thanks!

elenaf9 (approved):
Nice!

.github/mergify.yml
```diff
@@ -23,6 +23,24 @@ pull_request_rules:
     actions:
       queue:
 
+  - name: Conventional Commit PR title
+    conditions:
+      - base=master
+    actions:
+      post_check:
+        success_conditions:
+          - "title~=^(fix|feat|docs|refactor|chore|deps)(?:\\(.+\\))?!?:"
```
thomaseizinger:
```suggestion
          - "title~=^(fix|feat|docs|refactor|chore|deps)(\\(.+\\))?!?:"
```

The `:` at the start of the scope is wrong IMO.

thomaseizinger: I tested the new one here: https://www.regexpal.com/

In case you want to reproduce it, make sure to remove the escaping of `\` which is only necessary in this config file because the expression is within a string. (I think?)

PR #3096, opened 2022-11-08
mdns: update if-watch to 3.0.0

jxs:
and rename TokioMdns to Mdns living in its own module, and move Mdns to async_io::Mdns.

# Description

updates to `if-watch` `3.0`. This was simplest I found to not touch the signature of `GenMdns::new` wdyt @thomaseizinger? 
@mxinden turns out `IfWatcher::new` returns `Result`  so implementing `Default` for it wouldn't solve the problem, so I had to create a new `TryDefault` trait to be able to abstract over it.

We could also try put  `GenMdns` methods and trait impls on a macro and generate it for `tokio` and `async_io` code would be probably more complicated, but the doc would be cleaner and `GenMdns` would no longer be required.

CC @rkuhn 
superseeds #3095

## Change checklist


rkuhn (commented):
Looks good to me for mDNS! Shall I do the TCP part or do you want to take it?

jxs:
> Looks good to me for mDNS! Shall I do the TCP part or do you want to take it?

yeah, feel free to :) I didn't go for it yet as I wanted to clear the approach first, so if Thomas approves the approach feel free to implement for `tcp` Roland :)

protocols/mdns/CHANGELOG.md
```diff
@@ -4,6 +4,9 @@
 
 - Update to `libp2p-swarm` `v0.41.0`.
 
+- Update to `if-watch` `0.3.0` and both rename `TokioMdns` to `Mdns` living in `tokio::Mdns`,
```
```suggestion
- Update to `if-watch` `3.0.0` and both rename `TokioMdns` to `Mdns` living in `tokio::Mdns`,
```

protocols/mdns/src/lib.rs
```diff
@@ -42,10 +42,10 @@ mod behaviour;
 pub use crate::behaviour::{GenMdns, MdnsEvent};
 
 #[cfg(feature = "async-io")]
-pub use crate::behaviour::Mdns;
+pub use crate::behaviour::async_io;
 
 #[cfg(feature = "tokio")]
-pub use crate::behaviour::TokioMdns;
+pub use crate::behaviour::tokio;
```
thomaseizinger, line 67: If we already do this, should we just go all the way and rename `Mdns` to `Behaviour`, following #2217?

protocols/mdns/src/behaviour.rs
```diff
@@ -36,29 +36,53 @@ use smallvec::SmallVec;
...
+    use crate::behaviour::{socket::tokio::TokioUdpSocket, timer::tokio::TokioTimer};
+    use if_watch::tokio::IfWatcher;
+
+    pub type Mdns = GenMdns<TokioUdpSocket, TokioTimer, IfWatcher>;
+
+    impl TryDefault for IfWatcher {
+        fn try_default() -> Result<Self, std::io::Error> {
+            Self::new()
+        }
+    }
```
thomaseizinger: An alternative way of doing this which does not require a trait would be to say:

```rust
impl GenMdns<TokioUdpSocket, TokioTimer, IfWatcher> {
    /// Builds a new `Mdns` behaviour.
    pub fn new(config: MdnsConfig) -> io::Result<Self> {
        Ok(Self {
            config,
            if_watch: IfWatcher::new(),
            iface_states: Default::default(),
            discovered_nodes: Default::default(),
            closest_expiration: Default::default(),
        })
	}
}
```

thomaseizinger: Untested but I think this should work!

thomaseizinger (commented):
Thanks for tackling this @jxs!

protocols/mdns/src/behaviour.rs
jxs: Yeah, that works if we remove `GenMdns::new` otherwise compiler complains regarding duplication. The disadvantage I see with that is that then methods (and its docs) are scattered between `{tokio, async_io}::Mdns` and `GenMdns`, would you prefer that approach? Tbh ideally I think it would be great if we could deprecate/hide `GenMdns` and just present `{tokio, async_io}::Mdns` to the end user.

protocols/mdns/CHANGELOG.md
jxs: Thanks Thomas, updated.

protocols/mdns/src/lib.rs
jxs: yeah, agree! :)

protocols/mdns/src/behaviour.rs
thomaseizinger: We had a similar discussion in `libp2p-tcp`: https://github.com/libp2p/rust-libp2p/pull/2961#pullrequestreview-1129815307

From what I remember, the functions don't show up for the type alias at all so probably best to not do this at all.

The need for another trait is a bit annoying though.

protocols/mdns/src/behaviour.rs
```diff
@@ -36,29 +36,53 @@ use smallvec::SmallVec;
...
 /// them to the topology.
 #[derive(Debug)]
-pub struct GenMdns<S, T> {
+pub struct GenMdns<S, T, W> {
     /// InterfaceState config.
     config: MdnsConfig,
 
     /// Iface watcher.
-    if_watch: IfWatcher,
+    if_watch: W,
```
thomaseizinger: Can we do `T::IfWatcher` instead, similar to what @rkuhn did for TCP?

protocols/mdns/CHANGELOG.md
```diff
@@ -4,6 +4,9 @@
 
 - Update to `libp2p-swarm` `v0.41.0`.
 
+- Update to `if-watch` `3.0.0` and both rename `TokioMdns` to `Mdns` living in `tokio::Mdns`,
+and move `Mdns` to `async_io::Mdns`. see [PR 3096]
```
```suggestion
and move `Mdns` to `async_io::Mdns`. See [PR 3096].
```

mxinden (commented):
Looks good to me minus the outstanding comments.

protocols/mdns/CHANGELOG.md
jxs: Thanks Thomas, updated!

protocols/mdns/src/behaviour.rs
jxs: Yeah, great suggestion! Ptal if that's what you were expecting Thomas.

protocols/mdns/src/lib.rs
jxs: done :)

protocols/mdns/CHANGELOG.md
```diff
@@ -4,6 +4,16 @@
 
 - Update to `libp2p-swarm` `v0.41.0`.
 
+- Update to `if-watch` `3.0.0` and both rename `TokioMdns` to `Behaviour` living in `tokio::Behaviour`,
+and move and rename `Mdns` to `async_io::Behaviour`. See [PR 3096]
```
```suggestion
and move and rename `Mdns` to `async_io::Behaviour`. See [PR 3096].
```

protocols/mdns/src/behaviour.rs
```diff
@@ -36,29 +36,63 @@ use smallvec::SmallVec;
...
+            IfWatcher::new()
+        }
+    }
+}
 
 /// A `NetworkBehaviour` for mDNS. Automatically discovers peers on the local network and adds
 /// them to the topology.
 #[derive(Debug)]
-pub struct GenMdns<S, T> {
+pub struct Behaviour<S, T>
```
thomaseizinger: Would it make sense to combine all type parameters here to a single one, i.e. `P` and have a `Provider` trait?

```rust
trait Provider {
	type Socket;
	type Timer;
	type IfWatcher;
}
```

protocols/mdns/src/lib.rs
```diff
@@ -38,14 +38,26 @@
...
 mod behaviour;
-pub use crate::behaviour::{GenMdns, MdnsEvent};
+pub use crate::behaviour::{Behaviour, Event};
 
 #[cfg(feature = "async-io")]
-pub use crate::behaviour::Mdns;
+pub use crate::behaviour::async_io;
 
 #[cfg(feature = "tokio")]
-pub use crate::behaviour::TokioMdns;
```
thomaseizinger, line 48: If you'd add type aliases for this too then users would only be presented with deprecation warnings instead of compile errors.

thomaseizinger (commented):
Thanks @jxs! A few comments.

protocols/mdns/CHANGELOG.md
jxs: Thanks, addressed.

protocols/mdns/src/behaviour.rs
jxs: indeed it does! Thanks Thomas. :+1:  
We now have more than one `Provider` trait among the protocols (`tcp` and `mdns`) but as they are namedspaced by the crates I don't think it's a problem, it can be considered akin to the `tokio` and `async_io` namespaces.

protocols/mdns/src/lib.rs
jxs: yup, makes sense. When putting the types was thinking that since this is a breaking change putting the types would be redundant, but if we also add for `Mdns` then it doesn't become a breaking change. Thanks Thomas :)

protocols/mdns/CHANGELOG.md
```diff
@@ -4,8 +4,17 @@
 
 - Update to `libp2p-swarm` `v0.41.0`.
 
+- Update to `if-watch` `3.0.0` and both rename `TokioMdns` to `Behaviour` living in `tokio::Behaviour`,
+and move and rename `Mdns` to `async_io::Behaviour`. See [PR 3096].
+
+- Remove the remaning `Mdns` prefixes from types as per [discussion 2174].
+  I.e the `Mdns` prefix has been removed from various types like `MdnsEvent`.
+  Users should prefer importing the identify protocol as a module (`use libp2p::mdns;`),
```
```suggestion
  Users should prefer importing the mdns protocol as a module (`use libp2p::mdns;`),
```

protocols/mdns/CHANGELOG.md
```diff
@@ -22,6 +31,7 @@
 
 - Update to `if-watch` `v2.0.0` and thus the `async` method `Mdns::new` and `TokioMdns::new` becomes synchronous. See [PR 2978].
 
+[PR 3096]: https://github.com/libp2p/rust-libp2p/pull/3096
```
thomaseizinger: This link is in the wrong section, should be under `# 0.42.0`.

protocols/mdns/src/behaviour.rs
```diff
@@ -36,32 +36,80 @@ use smallvec::SmallVec;
 use std::collections::hash_map::{Entry, HashMap};
 use std::{cmp, fmt, io, net::IpAddr, pin::Pin, task::Context, task::Poll, time::Instant};
 
-#[cfg(feature = "async-io")]
-use crate::behaviour::{socket::asio::AsyncUdpSocket, timer::asio::AsyncTimer};
+/// The interface for non-blocking TCP I/O providers.
```
thomaseizinger:
```suggestion
/// An abstraction to allow for compatibility with various async runtimes.
```

protocols/mdns/src/behaviour.rs
```diff
@@ -72,19 +120,19 @@ pub struct GenMdns<S, T> {
...
-    closest_expiration: Option<T>,
+    closest_expiration: Option<P::Timer>,
 }
 
-impl<S, T> GenMdns<S, T>
+impl<P> Behaviour<P>
 where
-    T: Builder,
+    P: Provider,
+    P::Timer: Builder,
```
thomaseizinger: I think these could be specified on the trait itself, then you wouldn't have to repeat them here.

thomaseizinger (approved):
Nice!

protocols/mdns/src/behaviour.rs
jxs: Yup, you are right Thomas, thanks. Removed the redundancy and moved both here and in the `impl` of the trait to the trait definition ptal.

protocols/mdns/src/behaviour.rs
jxs: updated, thanks.

protocols/mdns/CHANGELOG.md
jxs: Right, thanks Thomas, moved it.

protocols/mdns/CHANGELOG.md
jxs: omg.. :man_facepalming: :see_no_evil: Thanks Thomas fixed it!

protocols/mdns/CHANGELOG.md
```diff
@@ -4,11 +4,23 @@
...
+- Update to `if-watch` `3.0.0` and both rename `TokioMdns` to `Behaviour` living in `tokio::Behaviour`,
+and move and rename `Mdns` to `async_io::Behaviour`. See [PR 3096].
+
+- Remove the remaning `Mdns` prefixes from types as per [discussion 2174].
+  I.e the `Mdns` prefix has been removed from various types like `MdnsEvent`.
+  Users should prefer importing the mdns protocol as a module (`use libp2p::mdns;`),
+  and refer to its types via `mdns::`. For example: `mdns::Behaviour` or `mdns::Event`.
+
+- Use `trust-dns-proto` to parse DNS messages. See [PR 3102].
+
```
```suggestion
```

Duplicate entry :)

thomaseizinger (approved):
Very nice & clean!

Good to go from my end :)

Issue #2230, opened 2021-09-13
tokio-based tcp transports panic when dialing due to threadpools with unset runtimes.


mental32:
I've spent the last couple of hours trying to figure this out but the best I can boil it down into is:

* When using a tokio-based transport such as `libp2p::tokio_development_transport`
* A call to `libp2p::swarm::Swarm::new` ends up using the `Default` constructed fields and calls `SwarmBuilder::build`
* And using a `Default` constructed `NetworkConfig` which sets the executor to `None` a call is made to `or_with_executor(f)` to construct a futures::executor, ThreadPool as the backing executor. 
* Any further calls to that use the executor that also interact with network machinery (like `Swarm::dial_addr`) will eventually lead to a panic because:
  *  it'll try to crate a tokio `TcpStream` which at some point calls `tokio::io::PollEvented::new_with_interest`
  * Which then calls `tokio::Handle::current` and that uses `tokio::runtime::context::io_handle()`
 * At which point we fail with the missing runtime message: `"there is no reactor running, must be called from the context of a Tokio 1.x runtime"`

Here are my deps:

```toml
[dependencies]
futures = "0.3.17"
libp2p = { version = "0.39.1", default-features = false, features = [
    "deflate",
    "gossipsub",
    "mdns",
    "tcp-tokio",
    "dns-tokio",
    "websocket",
    "noise",
    "mplex",
    "yamux",
] }

tokio = { version = "1.11.0", features = ["full"] }
```

Potential solutions I can think of are:

* Ask the transport for an executor during `SwarmBuilder::build` so tokio transports can produce a well-configured executor and the default would just be the futures ThreadPool executor closure that is present now.

* Document that tokio transports must set their own executor manually using `SwarmBuilder`

## Expected behavior

The swarm can dial the supplied Multiaddr.

## Actual behavior

Tokio panics because of an unset thread local deep in the io machinery, the context required to access a runtime.

## Reproducible example

Directory layout:

```
/tmp/asdasfa
 ├──Cargo.lock
 ├──Cargo.toml
 └──src
    └──main.rs
```

`src/main.rs`

<details>

```rust

use std::time::Duration;

use libp2p::core::identity::Keypair;
use libp2p::gossipsub::{self, Gossipsub, GossipsubEvent, MessageAuthenticity, ValidationMode};
use libp2p::mdns::{Mdns, MdnsConfig, MdnsEvent};
use libp2p::swarm::NetworkBehaviourEventProcess;
[...]

PR #3097, opened 2022-11-08
feat(swarm): Make executor for connection tasks explicit

### LINKED ISSUE #3068: swarm: Make API for creating a new `Swarm` executor aware
## Description

This issue is extracted out of #2173 and proposes a solution to issues like https://github.com/libp2p/rust-libp2p/issues/2230.

A `Swarm` needs to execute background tasks. Currently, each physical connection (i.e. a TCP connection) is executed on its own background task. This improves latency because it means the main event loop of the `Swarm` which also calls `NetworkBehaviour::poll` does not get blocked by IO. Also see https://github.com/libp2p/rust-libp2p/issues/2885.

Today, the way a `Swarm` achieves this is very subtle.

By default, `Swarm::new` will try to create a `futures::executor::Threadpool` to execute these connection background tasks. For async-IO types from the `tokio` runtime, this does not work because they require a tokio reactor to run on the same thread which by default is only active on a worker thread of a tokio `Runtime`. To achieve this, a user has to use the `SwarmBuilder` instead and call `executor(Box::new(|f| { tokio::spawn(f) }))`. See for example here:

https://github.com/libp2p/rust-libp2p/blob/b42f28630e2a7a04642c95aa8e7f4deea823f901/examples/chat-tokio.rs#L109-L115

In case creation of the `futures::executor::Threadpool` fails, the connection tasks are polled on the current thread instead:

https://github.com/libp2p/rust-libp2p/blob/b42f28630e2a7a04642c95aa8e7f4deea823f901/swarm/src/connection/pool.rs#L398-L404

https://github.com/libp2p/rust-libp2p/blob/b42f28630e2a7a04642c95aa8e7f4deea823f901/swarm/src/connection/pool.rs#L101-L103

I am proposed to change the API for `Swarm` and `SwarmBuilder` to force the user to make a decision on which executor they'd like their tasks to be executed on. There are several options and it is not yet clear, what the best one is:

- Introduce a type parameter on `Swarm` that has a trait-bound for spawning new tasks. This would allow us to re-export type-aliases like `libp2p::swarm::tokio::Swarm` which would point to something like `libp2p::swarm::Swarm<TokioRuntime>`. A type-parameter-only solution is likely the easiest to migrate too because it doesn't require any signature changes. We can deprecate the current type in favor of `libp2p::swarm::futures_threadpool::Swarm`.
- Introduce a new parameter to `libp2p::swarm::Swarm::new` that requires to pass an executor-specific type:
   ```rust
   impl Swarm {
       fn new<T>(..., executor: T) -> Self { }
   }

   Swarm::new(..., TokioExecutor);
   ```

There are likely other solutions too. Whatever we choose, it must play nicely with cargo features, i.e. be completely additive in terms of APIs and not change any behaviour.

## Motivation

- Make it easier for users to use `rust-libp2p` correctly.
- Less boilerplate to configure a custom executor.
- Treat all executors equally.

Maybe.

umgefahren:
## Description

Previously, the executor for connection tasks silently defaulted to a `futures::executor::ThreadPool`. This causes issues such as https://github.com/libp2p/rust-libp2p/issues/2230.

With this patch, we force the user to choose, which executor they want to run the connection tasks on which results in overall simpler API with less footguns.

Closes #3068.

## Open Questions

- [x]  Is the API satisfactory?
- [x] Should the implementation of the executor live in the core, swarm, or somewhere else?
- [x] What is the correct location for the Swarm aliases

## Change checklist

I will add a changelog entry once it's clear how broad the changes are.


swarm/src/connection/pool.rs
```diff
@@ -93,14 +128,9 @@ where
...
-    /// the tasks are kept in `local_spawns` instead and polled on the
-    /// current thread when the [`Pool`] is polled for new events.
-    executor: Option<Box<dyn Executor + Send>>,
-
-    /// If no `executor` is configured, tasks are kept in this set and
-    /// polled on the current thread when the [`Pool`] is polled for new events.
-    local_spawns: FuturesUnordered<Pin<Box<dyn Future<Output = ()> + Send>>>,
+    /// The executor to use for running the background tasks. Can either be an advanced executor
+    /// or a local queue.
+    executor: ExecSwitch<TExecutor>,
```
thomaseizinger:
```suggestion
    /// The executor to use for running the background tasks. Can either be a global executor
    /// or a local queue.
    executor: ExecSwitch<TExecutor>,
```

swarm/src/connection/pool.rs
```diff
@@ -54,10 +55,44 @@ use void::Void;
 mod concurrent_dial;
 mod task;
 
+enum ExecSwitch<TExecutor>
+where
+    TExecutor: Executor,
+{
+    Executor(TExecutor),
```
thomaseizinger: I would be in favor of using a trait object here. That will get rid of the type parameter which should make the signature changes less invasive!

swarm/src/connection/pool.rs
```diff
@@ -1050,9 +1083,9 @@ impl ConnectionLimits {
 ///
 /// The default configuration specifies no dedicated task executor, a
 /// task event buffer size of 32, and a task command buffer size of 7.
-pub struct PoolConfig {
+pub struct PoolConfig<TExecutor = ThreadPool> {
     /// Executor to use to spawn tasks.
-    pub executor: Option<Box<dyn Executor + Send>>,
+    pub executor: Option<TExecutor>,
```
thomaseizinger: Same here, lets stick to the trait object!

core/src/lib.rs
```diff
@@ -102,3 +103,42 @@ impl<F: Fn(Pin<Box<dyn Future<Output = ()> + Send>>)> Executor for F {
...
+        self.spawn_ok(future)
+    }
+}
+
+#[cfg(feature = "tokio")]
+#[derive(Clone, Copy, Debug, Default)]
+pub enum TokioExecutor<'a> {
+    #[default]
+    Empty,
+    Given(&'a tokio::runtime::Runtime),
```
thomaseizinger: I would be in favor of not supporting providing a runtime to start with unless someone comes and asks for it!

Always spawning into the global runtime should be good enough for now.

swarm/src/lib.rs
```diff
@@ -1286,15 +1326,36 @@ where
...
+{
     /// Configures the `Executor` to use for spawning background tasks.
     ///
     /// By default, unless another executor has been configured,
     /// [`SwarmBuilder::build`] will try to set up a
     /// [`ThreadPool`](futures::executor::ThreadPool).
-    pub fn executor(mut self, e: Box<dyn Executor + Send>) -> Self {
-        self.pool_config = self.pool_config.with_executor(e);
-        self
+    pub fn executor<NExecutor>(self, executor: NExecutor) -> SwarmBuilder<TBehaviour, NExecutor>
```
thomaseizinger: I think `SwarmBuilder::new` should take an `Option<TExecutor>` as a parameter. `None` will result in local spawns and `Some` will set the executor.

swarm/src/lib.rs
```diff
@@ -300,11 +301,24 @@ where
...
+}
+
+impl<TBehaviour, TExecutor> Swarm<TBehaviour, TExecutor>
+where
+    TBehaviour: NetworkBehaviour,
+    TExecutor: Executor,
+    PoolConfig<TExecutor>: Default,
 {
     /// Builds a new `Swarm`.
     pub fn new(
```
thomaseizinger, line 337: I am leaning towards deprecating this, defaulting to `Threadpool` in here (if construction fails, falling back to local spawns, same as now) and adding the following functions:

```rust
let swarm = Swarm::with_executor(transport, behaviour, local_peer_id, executor);
let swarm = Swarm::with_tokio_executor(transport, behaviour, local_peer_id);
let swarm = Swarm::with_async_std_executor(transport, behaviour, local_peer_id);
let swarm = Swarm::with_threadpool_executor(transport, behaviour, local_peer_id);
let swarm = Swarm::without_executor(transport, behaviour, local_peer_id);
```

thomaseizinger (commented):
Thanks for tackling this!

Please see inline comments :)

umgefahren:
I can implement these changes, but I have a question: If we go back to trait objects here, which would be fine by me, that means I will remove all the genetics from the APIs it that's what you want. I'm asking because I read your issue a bit differently.

swarm/src/lib.rs
umgefahren: I would avoid the `without_executor`. It might be ambiguous.

core/src/lib.rs
umgefahren: Alright

umgefahren:
l implemented the suggested changes. It results in a lot of warnings, but I will remove them if your happy with the api.

thomaseizinger:
> I'm asking because I read your issue a bit differently. 

Sorry if that was worded poorly! I tried to list multiple options there because I wasn't sure how things would turn out.

Breaking type inference is definitely a no-no in regards to ergonomics so adding feature-gated functions seems like a better approach!

core/Cargo.toml
```diff
@@ -56,9 +56,9 @@ serde_json = "1.0"
 prost-build = "0.11"
 
 [features]
-secp256k1 = [ "libsecp256k1" ]
-ecdsa = [ "p256" ]
-rsa = [ "dep:ring" ]
+secp256k1 = ["libsecp256k1"]
+ecdsa = ["p256"]
+rsa = ["dep:ring"]
```
thomaseizinger: No unrelated reformatting please :)

core/src/lib.rs
```diff
@@ -102,3 +103,9 @@ impl<F: Fn(Pin<Box<dyn Future<Output = ()> + Send>>)> Executor for F {
         self(f)
     }
 }
+
+impl Executor for ThreadPool {
+    fn exec(&self, future: Pin<Box<dyn Future<Output = ()> + Send>>) {
+        self.spawn_ok(future)
+    }
+}
```
thomaseizinger: I am leaning towards removing this in a future version. I don't see why we need it.

It can be deprecated in a follow-up though.

I'll open an issue.

examples/chat.rs
```diff
@@ -115,7 +115,7 @@ async fn main() -> Result<(), Box<dyn Error>> {
         };
 
         behaviour.floodsub.subscribe(floodsub_topic.clone());
-        Swarm::new(transport, behaviour, local_peer_id)
+        Swarm::with_async_std_executor(transport, behaviour, local_peer_id)
```
thomaseizinger: This should be futures-threadpool if we want to retain the exact same behaviour.

protocols/dcutr/examples/dcutr.rs
```diff
@@ -155,7 +156,12 @@ fn main() -> Result<(), Box<dyn Error>> {
         dcutr: dcutr::behaviour::Behaviour::new(),
     };
 
-    let mut swarm = SwarmBuilder::new(transport, behaviour, local_peer_id)
+    let executor: Option<Box<dyn Executor + Send>> = match ThreadPool::new() {
+        Ok(tp) => Some(Box::new(tp)),
+        Err(_) => None,
+    };
+
+    let mut swarm = SwarmBuilder::new(transport, behaviour, local_peer_id, executor)
```
thomaseizinger: I'd like this to be more ergonomic, ideally setting an executor here is a one-liner.

swarm/src/connection/pool.rs
```diff
@@ -93,14 +123,9 @@ where
...
-    /// The executor to use for running the background tasks. If `None`,
-    /// the tasks are kept in `local_spawns` instead and polled on the
-    /// current thread when the [`Pool`] is polled for new events.
-    executor: Option<Box<dyn Executor + Send>>,
-
-    /// If no `executor` is configured, tasks are kept in this set and
-    /// polled on the current thread when the [`Pool`] is polled for new events.
-    local_spawns: FuturesUnordered<Pin<Box<dyn Future<Output = ()> + Send>>>,
+    /// The executor to use for running the background tasks. Can either be a global executor
+    /// or a local queue.
```
thomaseizinger:
```suggestion
    /// The executor to use for running connection tasks. Can either be a global executor
    /// or a local queue.
```

swarm/src/connection/pool.rs
```diff
@@ -1075,11 +1099,17 @@ pub struct PoolConfig {
 
 impl Default for PoolConfig {
     fn default() -> Self {
```
thomaseizinger: I think we should delete this in the presence of our `new` function.

swarm/src/connection/pool.rs
```diff
@@ -1149,6 +1180,32 @@ impl PoolConfig {
         self.max_negotiating_inbound_streams = v;
         self
     }
+
+    /// Configures the executor to use for spawning connection background tasks,
+    /// only if no executor has already been configured.
+    #[allow(dead_code)] // TODO: Can we just remove this?
```
thomaseizinger: Probably.

swarm/src/lib.rs
```diff
@@ -307,12 +308,79 @@ where
...
+        behaviour: TBehaviour,
+        local_peer_id: PeerId,
+    ) -> Self {
+        SwarmBuilder::new(
+            transport,
+            behaviour,
+            local_peer_id,
+            Some(Box::new(crate::connection::pool::executor::TokioExecutor)),
+        )
+        .build()
```
```suggestion
        SwarmBuilder::with_executor(
            transport,
            behaviour,
            local_peer_id,
            Box::new(crate::connection::pool::executor::TokioExecutor),
        )
```

swarm/src/lib.rs
```diff
@@ -1277,12 +1345,13 @@ where
         transport: transport::Boxed<(PeerId, StreamMuxerBox)>,
         behaviour: TBehaviour,
         local_peer_id: PeerId,
+        executor: Option<Box<dyn Executor + Send>>,
```
thomaseizinger: I think we should deprecate this function too in favor of two new ones:

- `with_executor`
- `without_executor`

swarm/src/lib.rs
```diff
@@ -1292,8 +1361,8 @@ where
     /// By default, unless another executor has been configured,
     /// [`SwarmBuilder::build`] will try to set up a
     /// [`ThreadPool`](futures::executor::ThreadPool).
-    pub fn executor(mut self, e: Box<dyn Executor + Send>) -> Self {
-        self.pool_config = self.pool_config.with_executor(e);
+    pub fn executor(mut self, executor: Box<dyn Executor + Send>) -> Self {
+        self.pool_config = self.pool_config.with_executor(executor);
```
thomaseizinger: We should remove this, now that we pass the executor in on the ctor.

thomaseizinger (commented):
Thanks! Some inline comments :)

core/src/lib.rs
umgefahren: I'm not quite sure what you mean. Removing ThreadPool?

core/Cargo.toml
umgefahren: I wasn't ware I did this. Maybe `cargo fmt`?

protocols/dcutr/examples/dcutr.rs
umgefahren: Yes I'm aware, this is not ideal.

Ideally it would look like this:
```rust
let executor: Option<Box<dyn Executor + Send>> = ThreadPool::new().ok().map(|e| Box::new(e));
```
I'm pretty sure it is equivalent to the code above:
```rust
  let executor: Option<Box<dyn Executor + Send>> = match ThreadPool::new() {
      Ok(tp) => Some(Box::new(tp)),
      Err(_) => None,
  };
```
Correct me I might be wrong here, but it might be one of these classic trait object problems that still bug rust and why I try to avoid them.

core/src/lib.rs
thomaseizinger: Removing the possibility of using `ThreadPool` for the connection tasks. Any async Rust program will use some kind of executor so we can just spawn the tasks into those worker threads instead of spinning up some threads on our own.

core/Cargo.toml
thomaseizinger: Unlikely, `cargo fmt` doesn't yet do `Cargo.toml` files!

protocols/dcutr/examples/dcutr.rs
thomaseizinger: 1. I think `SwarmBuilder` should have two functions: `with_executor` and `without_executor`.
2. We can just change this example to use `AsyncStdExecutor`.

It is not fully equivalent but I don't think there is a reason to show off the `ThreadPoolExecutor` here.

swarm/src/lib.rs
thomaseizinger: How do you create a `Swarm` then that runs its connections on the current task?

core/src/lib.rs
umgefahren: Yes, you are right. But let's do that separately.

> I'll open an issue.

It would be so great if we could add special comments to PRs that open issues once the PR is merged...

umgefahren:
While we are on it: In the interest of #2166 we should add support for wasm bindgen futures.

core/Cargo.toml
umgefahren: Oh, then I have no idea.

protocols/dcutr/examples/dcutr.rs
umgefahren: Yes, that sounds good. How should I handle `SwarmBuilder::new`?

thomaseizinger: I suggest we deprecate it and have this snippet with the match-statement in there, calling through to either one or the other function.

core/src/lib.rs
thomaseizinger: Done: https://github.com/libp2p/rust-libp2p/issues/3107

thomaseizinger:
> While we are on it: In the interest of #2166 we should add support for wasm bindgen futures.

Oh that would be amazing actually!

core/Cargo.toml
```diff
@@ -59,7 +59,7 @@ prost-build = "0.11"
 secp256k1 = [ "libsecp256k1" ]
 ecdsa = [ "p256" ]
 rsa = [ "dep:ring" ]
-serde = ["multihash/serde-codec", "dep:serde"]
+serde = [ "multihash/serde-codec", "dep:serde" ]
```
thomaseizinger:
```suggestion
serde = ["multihash/serde-codec", "dep:serde"]
```

swarm/src/connection/pool.rs
```diff
@@ -1073,34 +1096,21 @@ pub struct PoolConfig {
...
-    /// only if no executor has already been configured.
-    pub fn or_else_with_executor<F>(mut self, f: F) -> Self
-    where
-        F: FnOnce() -> Option<Box<dyn Executor + Send>>,
-    {
-        self.executor = self.executor.or_else(f);
+    pub fn with_executor(mut self, executor: Box<dyn Executor + Send>) -> Self {
+        self.executor = Some(executor);
         self
     }
```
thomaseizinger, line 1115: Do we still need this function at all?

swarm/src/connection/pool.rs
```diff
@@ -52,12 +52,41 @@ use std::{
 use void::Void;
 
 mod concurrent_dial;
+#[cfg(any(feature = "tokio", feature = "async-std"))]
+pub mod executor;
 mod task;
 
+enum ExecSwitch {
+    Executor(Box<dyn Executor + Send>),
```
thomaseizinger, line 58: I only just realised that `Executor` still lives in `libp2p-core`, despite being only used in `libp2p-swarm`!

Can you move it to a top-level `executor` within `libp2p-swarm` please?

The various implementations can then just go next to it (but please make sure they remain crate-private).

swarm/src/lib.rs
thomaseizinger: Actually, lets keep this fn but deprecate it!

swarm/src/connection/pool.rs
umgefahren: I can move them to libp2p-swarm, but I would like to propose an alternative. I don't quite see why we are using our own custom Executor trait. Why not use [`futures::task::Spawn`](https://docs.rs/futures/0.3.25/futures/task/trait.Spawn.html#tymethod.spawn_obj)? With that we could simplify here and also get rid of every responsibility regarding ThreadPool.

Another advantage: With the status method we could check if we can actually spawn tasks.

umgefahren:
Let me know as soon as your fine with the changes made. I will then start the process of replacing all usages of the deprecated API.

swarm/src/lib.rs
```diff
@@ -86,7 +87,7 @@ use connection::pool::{EstablishedConnection, Pool, PoolConfig, PoolEvent};
 use connection::IncomingInfo;
 use dial_opts::{DialOpts, PeerCondition};
 use either::Either;
-use futures::{executor::ThreadPoolBuilder, prelude::*, stream::FusedStream};
+use futures::{prelude::*, stream::FusedStream};
```
thomaseizinger: Nit: Unnecessary diff.

swarm/src/lib.rs
```diff
@@ -1273,16 +1341,55 @@ where
...
     /// via [`SwarmBuilder::build`].
+    #[deprecated(
+        since = "0.50.0",
+        note = "Unnecessary since replaced by `SwarmBuilder::with_executor` and `SwarmBuilder::without_executor`."
+    )]
     pub fn new(
         transport: transport::Boxed<(PeerId, StreamMuxerBox)>,
         behaviour: TBehaviour,
         local_peer_id: PeerId,
+        executor: Option<Box<dyn Executor + Send>>,
```
thomaseizinger: Can you revert this line please? That is a breaking change.

swarm/src/lib.rs
```diff
@@ -1273,16 +1341,55 @@ where
...
+            local_peer_id,
+            transport,
+            behaviour,
+            pool_config: PoolConfig::new(Some(executor)),
+            connection_limits: Default::default(),
+        }
+    }
+
+    /// Creates a new `SwarmBuilder` from the given transport, behaviour and local peer ID. The
+    /// `Swarm` with its underlying `Network` is obtained via [`SwarmBuilder::build`].
```
thomaseizinger: We should say here that all connections will be polled on the current task, together with `NetworkBehaviour::poll` which will likely have quite bad performance characteristics.

When possible, users should use an executor!

swarm/src/lib.rs
```diff
@@ -1292,8 +1399,12 @@ where
     /// By default, unless another executor has been configured,
     /// [`SwarmBuilder::build`] will try to set up a
     /// [`ThreadPool`](futures::executor::ThreadPool).
-    pub fn executor(mut self, e: Box<dyn Executor + Send>) -> Self {
-        self.pool_config = self.pool_config.with_executor(e);
+    #[deprecated(
+        since = "0.50.0",
+        note = "Deprecated since executor creation now happens in `SwarmBuilder::with_executor`."
```
```suggestion
        note = "Use `SwarmBuilder::with_executor` instead."
```

swarm/src/lib.rs
```diff
@@ -1381,6 +1492,8 @@ where
     }
 
     /// Builds a `Swarm` with the current configuration.
+    ///
+    /// If no executor was given, no executor will be set.
```
```suggestion
```

This seems pretty redundant to me :)

swarm/src/lib.rs
```diff
@@ -1273,16 +1341,55 @@ where
     /// Creates a new `SwarmBuilder` from the given transport, behaviour and
     /// local peer ID. The `Swarm` with its underlying `Network` is obtained
     /// via [`SwarmBuilder::build`].
+    #[deprecated(
+        since = "0.50.0",
+        note = "Unnecessary since replaced by `SwarmBuilder::with_executor` and `SwarmBuilder::without_executor`."
```
```suggestion
        note = "Use `SwarmBuilder::with_executor` or `SwarmBuilder::without_executor` instead."
```

thomaseizinger (commented):
API looks good to me apart from one breaking change :)

swarm/src/lib.rs
```diff
@@ -1273,16 +1341,55 @@ where
...
         behaviour: TBehaviour,
         local_peer_id: PeerId,
+        executor: Option<Box<dyn Executor + Send>>,
     ) -> Self {
         SwarmBuilder {
             local_peer_id,
             transport,
             behaviour,
-            pool_config: Default::default(),
+            pool_config: PoolConfig::new(executor),
```
thomaseizinger, line 1397: We need to recreate the behaviour that happened in `build` before here!

swarm/src/lib.rs
umgefahren: Good question. Maybe with `Swarm::with_current_thread_executor`?

umgefahren: Decided otherwise, I hope you like my solution.

swarm/src/lib.rs
```diff
@@ -1381,6 +1515,7 @@ where
     }
 
     /// Builds a `Swarm` with the current configuration.
+    ///
```
```suggestion
```

swarm/src/lib.rs
```diff
@@ -1273,16 +1355,64 @@ where
...
+        transport: transport::Boxed<(PeerId, StreamMuxerBox)>,
+        behaviour: TBehaviour,
+        local_peer_id: PeerId,
+        executor: Box<dyn Executor + Send>,
+    ) -> Self {
+        Self {
+            local_peer_id,
+            transport,
+            behaviour,
+            pool_config: PoolConfig::new(Some(executor)),
```
thomaseizinger: We can probably make this slightly more ergonomic by boxing for the user:


```suggestion
        executor: impl Executor + Send,
    ) -> Self {
        Self {
            local_peer_id,
            transport,
            behaviour,
            pool_config: PoolConfig::new(Some(Box::new(executor))),
```

thomaseizinger: We can probably make this slightly more ergonomic by boxing for the user:


```suggestion
        executor: impl Executor + Send,
    ) -> Self {
        Self {
            local_peer_id,
            transport,
            behaviour,
            pool_config: PoolConfig::new(Some(Box::new(executor))),
```

swarm/src/lib.rs
```diff
@@ -307,12 +307,94 @@ where
     TBehaviour: NetworkBehaviour,
 {
     /// Builds a new `Swarm`.
+    #[deprecated(
+        since = "0.50.0",
+        note = "This constructor is considered ambiguous regarding the executor. Use `Swarm::with_threadpool_executor` for the same behaviour."
```
```suggestion
        note = "This constructor is considered ambiguous regarding the executor. Use one of the new, executor-specific constructors or `Swarm::with_threadpool_executor` for the same behaviour."
```

swarm/src/lib.rs
```diff
@@ -307,12 +307,94 @@ where
...
-        SwarmBuilder::new(transport, behaviour, local_peer_id).build()
+        Self::with_threadpool_executor(transport, behaviour, local_peer_id)
+    }
+
+    /// Builds a new `Swarm` with a provided executor.
+    pub fn with_executor(
+        transport: transport::Boxed<(PeerId, StreamMuxerBox)>,
+        behaviour: TBehaviour,
+        local_peer_id: PeerId,
+        executor: Box<dyn Executor + Send>,
```
thomaseizinger: Here too:

```suggestion
        executor: impl Executor + Send,
```

swarm/src/lib.rs
```diff
@@ -307,12 +307,94 @@ where
...
+        behaviour: TBehaviour,
+        local_peer_id: PeerId,
+    ) -> Self {
+        SwarmBuilder::with_executor(
+            transport,
+            behaviour,
+            local_peer_id,
+            Box::new(crate::connection::pool::executor::TokioExecutor),
+        )
+        .build()
```
thomaseizinger: This could use `Self`:


```suggestion
        Self::with_executor(
            transport,
            behaviour,
            local_peer_id,
            Box::new(crate::connection::pool::executor::TokioExecutor),
        )
```

swarm/src/connection/pool.rs
thomaseizinger: > I can move them to libp2p-swarm, but I would like to propose an alternative. I don't quite see why we are using our own custom Executor trait. Why not use [`futures::task::Spawn`](https://docs.rs/futures/0.3.25/futures/task/trait.Spawn.html#tymethod.spawn_obj)? With that we could simplify here and also get rid of every responsibility regarding ThreadPool.

Can we? You still need to poll the `FuturesUnordered` so you can't completely erase the type.

I don't see how it makes things simpler but feel free to have a try!

> Another advantage: With the status method we could check if we can actually spawn tasks.

And if we can't spawn it then what? I don't see a reasonable error-handling strategy. Also, the interface guarantees are pretty weak so I don't see what benefit we are getting from "the executor is likely to accept the future".

thomaseizinger: My gut feeling is that `Spawn` is not what we want but I am happy to be convinced otherwise :)

thomaseizinger (commented):
Thanks!

swarm/src/lib.rs
umgefahren: I like your idea, however, if the executor is `Box<dyn Executor + Send>` it will become `Box<Box<dyn Executor + Send>>`.  What about: `impl Into<Box<dyn Executor + Send>>`?

umgefahren:
Yes, you are right. Spawn won't change anything. I just thought we shouldn't do double work. However, I tried to redesign the implementations on foreign traits.

Doint this I discovered a very interesting problem:
```rust
impl<T> Executor for Box<T>
where T: Executor + ?Sized,
{
    fn exec(&self, future: Pin<Box<dyn Future<Output = ()> + Send>>) {
        (**self).exec(future)
    }
}

impl<T> Executor for &T
where T: Executor + ?Sized,
{
    fn exec(&self, future: Pin<Box<dyn Future<Output = ()> + Send>>) {
       T::exec(self, future) 
    }
}

impl<T> Executor for std::rc::Rc<T>
where T: Executor + ?Sized,
{
    fn exec(&self, future: Pin<Box<dyn Future<Output = ()> + Send>>) {
        (**self).exec(future)
    }
}

impl<T> Executor for std::sync::Arc<T>
where T: Executor + ?Sized,
{
    fn exec(&self, future: Pin<Box<dyn Future<Output = ()> + Send>>) {
        (**self).exec(future)
    }
}

impl<T> Executor for &mut T
where T: Executor + ?Sized,
{
    fn exec(&self, future: Pin<Box<dyn Future<Output = ()> + Send>>) {
        T::exec(self, future)
    }
}
```

This compiles fine. Unless I add:
```rust
impl<F> Executor for F
where F: Fn(Pin<Box<dyn Future<Output = ()> + Send>>),
{
    fn exec(&self, future: Pin<Box<dyn Future<Output = ()> + Send>>) {
        self(future)
    }
}
```
which results in a redefinition error. If i leave the impl on the closure out, the following doesn't compile anymore:
```rust
struct Dummy;

impl Executor for Dummy {
    fn exec(&self, _: Pin<Box<dyn Future<Output = ()> + Send>>) {}
}

  // TODO: This test has to be redesigned.
#[test]
fn set_executor() {
    PoolConfig::new(None)
        .with_executor(Box::new(Dummy))
        .with_executor(Box::new(|f| {
            async_std::task::spawn(f);
        }));
  }
```
the error is that the trait-bound Executor is not implemented for the closure. Why? This kinda violates the image I had of all of this. For now, I removed the extra impls but I'm irritated nevertheless.

umgefahren:
From my point of view, I'm pleased with the changes. There is only one downside: Since we require Send on the trait object one would need `Arc` if one ever wants to use a dedicated runtime, taken by reference.

core/CHANGELOG.md
```diff
@@ -4,8 +4,11 @@
 
 - Hide `prost::Error` from public API in `FromEnvelopeError::InvalidPeerRecord` and `signed_envelope::DecodingError`. See [PR 3058].
 
+- Moved `Executor` to `libp2p-swarm`. See [PR 3097].
```
```suggestion
- Move `Executor` to `libp2p-swarm`. See [PR 3097].
```

swarm/CHANGELOG.md
```diff
@@ -5,9 +5,70 @@
 - Export `NetworkBehaviour` derive as `libp2p_swarm::NetworkBehaviour`.
   This follows the convention of other popular libraries. `serde` for example exports the `Serialize` trait and macro as
   `serde::Serialize`. See [PR 3055].
+
 - Feature-gate `NetworkBehaviour` macro behind `macros` feature flag. See [PR 3055].
 
+- Made Swarm executor aware. See [PR 3097] for details.
```
```suggestion
- Make executor in Swarm constructor explicit. See [PR 3097].
```

swarm/CHANGELOG.md
```diff
@@ -5,9 +5,70 @@
...
+    let swarm = Swarm::new(transport, behaviour, peer_id);
+    ```
+
+    Now
+    ```
+    let swarm = Swarm::with_threadpool_executor(transport, behaviour, peer_id);
+    ```
+  - Without
+    
+    Spawns the tasks on the current task, this results in bad performance so try to use an executor if possible. Previously this was just a fallback for ThreadPool.
```
```suggestion
    Spawns the tasks on the current task, this may result in bad performance so try to use an executor where possible. Previously this was just a fallback when no executor was specified and constructing a `ThreadPool` failed.
```

swarm/CHANGELOG.md
```diff
@@ -5,9 +5,70 @@
...
+    
+    In most cases ThreadPool can be replaced by executors or spawning on the local task.
+
+    Previously
+    ```rust
+    let swarm = Swarm::new(transport, behaviour, peer_id);
+    ```
+
+    Now
+    ```
```
```suggestion
    ```rust
```

swarm/src/connection/pool.rs
```diff
@@ -54,10 +54,37 @@ use void::Void;
 mod concurrent_dial;
 mod task;
 
+enum ExecSwitch {
+    Executor(Box<dyn Executor + Send>),
+    LocalSpawn(FuturesUnordered<Pin<Box<dyn Future<Output = ()> + Send>>>),
+}
+
+impl ExecSwitch {
+    // advance the local queue
```
thomaseizinger:
```suggestion
```

swarm/src/connection/pool.rs
```diff
@@ -1175,9 +1183,10 @@ mod tests {
         fn exec(&self, _: Pin<Box<dyn Future<Output = ()> + Send>>) {}
     }
 
+    // TODO: This test has to be redesigned.
```
thomaseizinger: I think we can delete it now.

swarm/src/executor.rs
```diff
@@ -0,0 +1,48 @@
...
+}
+
+impl Executor for ThreadPool {
+    fn exec(&self, future: Pin<Box<dyn Future<Output = ()> + Send>>) {
+        self.spawn_ok(future)
+    }
+}
+
+#[cfg(feature = "tokio")]
+#[derive(Default, Debug, Clone, Copy, Hash, PartialEq, Eq, PartialOrd, Ord)]
```
thomaseizinger: Do we need these? 😁

swarm/src/lib.rs
```diff
@@ -64,6 +64,7 @@ mod upgrade;
 pub mod behaviour;
 pub mod dial_opts;
 pub mod dummy;
+pub mod executor;
```
```suggestion
mod executor;
```

Lets do this instead but re-export the trait.

swarm/CHANGELOG.md
```diff
@@ -5,9 +5,70 @@
...
+    let swarm = SwarmBuilder::new(transport, behaviour, peer_id)
+        .executor(Box::new(|fut| {
+                tokio::spawn(fut);
+        }))
+        .build();
+    ```
+    Now
+    ```rust
+    let swarm = Swarm::with_tokio_executor(transport, behaviour, peer_id);
+    ```
```
thomaseizinger, line 27: Would this look better if we used a markdown table with two columns?

thomaseizinger (approved):
Awesome! This looks really good, thank you!

I am very excited to ship this ❤️

A few more minor comments, plus I'd be great if @mxinden could glance over this aswell.

swarm/src/executor.rs
umgefahren: No, not really. But I like it and it doesn't hurt, as it's only instantiated when used (at least I read that somewhere I think). It only increases file size by a very, very small amount and probably takes some ns during compilation. If it's all the same to you, I'm in favor of just leaving it.

swarm/CHANGELOG.md
umgefahren: Maybe, but it doesn't really work with code highlighting and `<br>`. So we probably have to leave it like this.

swarm/src/connection/pool.rs
umgefahren: I deleted it.

umgefahren:
Unless you want me to remove `#[derive(Default, Debug, Clone, Copy, Hash, PartialEq, Eq, PartialOrd, Ord)]` I think there is nothing left to do for me here.

swarm/src/connection/pool.rs
```diff
@@ -54,10 +54,36 @@ use void::Void;
 mod concurrent_dial;
 mod task;
 
+enum ExecSwitch {
+    Executor(Box<dyn Executor + Send>),
+    LocalSpawn(FuturesUnordered<Pin<Box<dyn Future<Output = ()> + Send>>>),
+}
+
+impl ExecSwitch {
+    #[inline]
```
mxinden: Why inline these? Is there any prove that this is a performance optimization?

swarm/src/connection/pool.rs
```diff
@@ -821,7 +841,7 @@ where
         }
 
         // Advance the tasks in `local_spawns`.
```
mxinden: Outdated comment.

swarm/src/connection/pool.rs
```diff
@@ -1073,34 +1093,21 @@ pub struct PoolConfig {
...
-    }
-
-    /// Configures the executor to use for spawning connection background tasks,
-    /// only if no executor has already been configured.
-    pub fn or_else_with_executor<F>(mut self, f: F) -> Self
-    where
-        F: FnOnce() -> Option<Box<dyn Executor + Send>>,
-    {
-        self.executor = self.executor.or_else(f);
+    pub fn with_executor(mut self, executor: Box<dyn Executor + Send>) -> Self {
```
mxinden, line 1106: Providing the an `Option` to an `Executor` in the `PoolConfig::new` method and then also offering a `with_executor` builder method seems not intuitive to me. What do you think of dropping the `executor` parameter in `PoolConfig::new`?

swarm/src/executor.rs
```diff
@@ -0,0 +1,48 @@
+use futures::executor::ThreadPool;
+use std::{future::Future, pin::Pin};
+
+/// Implemented on objects that can run a `Future` in the background.
+///
+/// > **Note**: While it may be tempting to implement this trait on types such as
+/// >           [`futures::stream::FuturesUnordered`], please note that passing an `Executor` is
+/// >           optional, and that `FuturesUnordered` (or a similar struct) will automatically
+/// >           be used as fallback by libp2p. The `Executor` trait should therefore only be
+/// >           about running `Future`s in the background.
```
mxinden:
```suggestion
/// >           about running `Future`s on a separate task.
```

Not a strong opinion.

swarm/src/executor.rs
mxinden: I suggest dropping `Hash`, `PartialEq`, `Eq`, `PartialOrd` and `Ord`. Not because it would hurt (compile time) performance, but because it is misleading. In which case would you hash a `TokioExecutor`?

swarm/src/executor.rs
```diff
@@ -0,0 +1,48 @@
...
+
+#[cfg(feature = "tokio")]
+impl Executor for TokioExecutor {
+    fn exec(&self, future: Pin<Box<dyn Future<Output = ()> + Send>>) {
+        let _ = tokio::spawn(future);
+    }
+}
+
+#[cfg(feature = "async-std")]
+#[derive(Default, Debug, Clone, Copy, Hash, PartialEq, Eq, PartialOrd, Ord)]
```
mxinden: Same here.

swarm/src/lib.rs
```diff
@@ -328,12 +330,89 @@ where
...
+        local_peer_id: PeerId,
+    ) -> Self {
+        let builder = match ThreadPoolBuilder::new()
+            .name_prefix("libp2p-swarm-task-")
+            .create()
+        {
+            Ok(tp) => SwarmBuilder::with_executor(transport, behaviour, local_peer_id, tp),
+            Err(err) => {
+                log::warn!("Failed to create executor thread pool: {:?}", err);
+                SwarmBuilder::without_executor(transport, behaviour, local_peer_id)
```
mxinden, line 398: What do you think of returning an error instead? Users may still fall back to `without_executor` in case that is a valid option in their use-case.

swarm/src/lib.rs
```diff
@@ -1294,16 +1373,67 @@ where
...
             local_peer_id,
             transport,
             behaviour,
-            pool_config: Default::default(),
+            pool_config: PoolConfig::new(executor),
+            connection_limits: Default::default(),
+        }
+    }
+
+    /// Creates a new `SwarmBuilder` from the given transport, behaviour, local peer ID and
```
```suggestion
    /// Creates a new [`SwarmBuilder`] from the given transport, behaviour, local peer ID and
```

swarm/src/lib.rs
```diff
@@ -1294,16 +1373,67 @@ where
...
+        Self {
+            local_peer_id,
+            transport,
+            behaviour,
+            pool_config: PoolConfig::new(Some(Box::new(executor))),
+            connection_limits: Default::default(),
+        }
+    }
+
+    /// Creates a new `SwarmBuilder` from the given transport, behaviour and local peer ID. The
```
```suggestion
    /// Creates a new [`SwarmBuilder`] from the given transport, behaviour and local peer ID. The
```

mxinden (commented):
Great work. Happy for this footgun to soon be gone.

@umgefahren could you please also create a pull request removing the deprecated methods? Once this is released we can then merge that second pull request.

swarm/src/connection/pool.rs
thomaseizinger: It is the `with_executor` function that will go away eventually :)

At the moment it is used to support old, now deprecated APIs!

swarm/src/connection/pool.rs
umgefahren: Yes, I think it will have a performance benefit. Since `advance local` and `spawn` are called in the same function, the `match` statements can be reduced to one by the compiler and the inline doesn't increase binary size. `#[inline]` is more of a nudge to inline anyway.

swarm/src/lib.rs
thomaseizinger: At the moment, this is mostly there to allow easy migration and provide backwards-compatibility.

I want to get rid of this entirely: https://github.com/libp2p/rust-libp2p/issues/3107

swarm/src/connection/pool.rs
umgefahren: Yes, I will remove it once we deprecate these old functions.

swarm/src/executor.rs
umgefahren: Yes you are right.

swarm/src/lib.rs
umgefahren: I agree with Thomas here. It would mean a lot of hassle to return an error here and it's probably not worth the effort when we throw it out anyway.

swarm/CHANGELOG.md
thomaseizinger: I see. That is a shame. Thanks for trying!

umgefahren:
> This pull request has merge conflicts. Could you please resolve them @umgefahren? 🙏

That's probably the happiest merge conflict I ever had to resolve. Finally QUIC!!

swarm/src/connection/pool.rs
mxinden: My intuition tells me that a match statement is not expensive and thus, unless in a hot-path, not worth optimizing away.

For the `spawn` case, given that we only spawn one task per connection and given that establishing a connection has many operations magnitude larger than a `match`, I don't think the `#[inline]` has any impact.

For `advance_local`, this triggers operations again orders of magnitude more expensive. Again I don't think the `#[inline]` has any impact.

Unless we have a benchmark that shows a performance improvement, I don't think we should be adding an `#[inline]` anywhere.

See also https://github.com/libp2p/rust-libp2p/issues/897 for past discussions on this topic.

swarm/CHANGELOG.md
```diff
@@ -5,9 +5,70 @@
 - Export `NetworkBehaviour` derive as `libp2p_swarm::NetworkBehaviour`.
   This follows the convention of other popular libraries. `serde` for example exports the `Serialize` trait and macro as
   `serde::Serialize`. See [PR 3055].
+
 - Feature-gate `NetworkBehaviour` macro behind `macros` feature flag. See [PR 3055].
 
+- Make executor in Swarm constructor explicit. See [PR 3097].
```
mxinden, line 11: Thank you for the great changelog entry.

mxinden (commented):
Follow up on `inline` discussion. Otherwise looks good to me.

swarm/src/connection/pool.rs
umgefahren: I see. You are right, I will remove it. I inserted it because I tried to make it a habit to inline everywhere a compiler might. I see why it's not appropriate.

swarm/CHANGELOG.md
umgefahren: Thanks :)

mxinden (approved):

mxinden:
Again, thanks @umgefahren for the help!

mxinden:
(I am sure many future users not falling into this trap will be very happy (unknowingly).)

swarm/src/executor.rs
```diff
@@ -0,0 +1,48 @@
...
+
+impl Executor for ThreadPool {
+    fn exec(&self, future: Pin<Box<dyn Future<Output = ()> + Send>>) {
+        self.spawn_ok(future)
+    }
+}
+
+#[cfg(feature = "tokio")]
+#[derive(Default, Debug, Clone, Copy)]
+pub(crate) struct TokioExecutor;
```
mxinden, line 30: Why not expose `TokioExecutor` and `AsyncStdExecutor`? Folks that want to specify configs, e.g. `dial_concurrency_factor` need to use the `SwarmBuilder`. Instead of implementing `Executor` on their custom type wrapping e.g. the `async-std` executor, they could just reuse our implementation. Thoughts @umgefahren and @thomaseizinger?

umgefahren: Yes, I think you are right. If Thomas agrees I can open a PR real quick changing that and adding some documentation.

thomaseizinger: Alternatively, we could add the same set of `with_` APIs.

I am okay either way. (I thought we had two layers of `with_` APIs)

thomaseizinger: Can we just get rid of `SwarmBuilder` instead maybe? Is there a reason why any of the config values could not be set on `Swarm` itself?

dignifiedquire: Trying to upgrade iroh to laster `master` and blocked on this :/

dignifiedquire: (I can work around it by reimplementing, but not great)

umgefahren: Could you elaborate?

dignifiedquire: I nedd the SwarmBuilder and have to reimplement the Executor trait by copy pasting the one in here, to use tokio.

thomaseizinger: > Can we just get rid of `SwarmBuilder` instead maybe? Is there a reason why any of the config values could not be set on `Swarm` itself?

@umgefahren Would you mind:

- sending a hotfix PR that makes these types public
- attempting to remove `SwarmBuilder` altogether by moving the current "builder" functions directly onto `Swarm`

LMK if you are not keen for the latter, then I'll try myself :)

umgefahren: Yes, I can do that (although I'm done for today)

umgefahren: Done #3155

Issue #3046, opened 2022-10-19
Removing the `handler` field from `FromSwarm::ConnectionClosed` in favor of `ConnectionHandler::poll_close`


thomaseizinger:
Currently, `FromSwarm::ConnectionClosed` contains a `handler` field which makes handling the event klunky in many ways:

- It cannot implement `Clone`
- It adds a type-parameter
- We cannot unify `SwarmEvent` and `FromSwarm` despite them containing the same data

The usecase for returning the handler is to allow the behaviour to extract data from a closed connection which would other get lost as connections can close at any point. Nothing is currently using this functionality but it makes sense in general so we want to keep it.

We can solve this by adding a `poll_close` function as described by @mxinden below: https://github.com/libp2p/rust-libp2p/issues/3046#issuecomment-1552365682

```patch
modified   swarm/src/handler.rs
@@ -158,6 +158,15 @@ pub trait ConnectionHandler: Send + 'static {
         >,
     >;
 
+    /// Gracefully close the [`ConnectionHandler`].
+    ///
+    /// Implementations may transfer one or more events to their [`NetworkBehaviour`] implementation
+    /// by emitting [`Poll::Ready`] with [`Self::ToBehaviour`]. Implementations should eventually
+    /// return [`Poll::Ready(None)`] to signal successful closing.
+    fn poll_close(&mut self, cx: &mut Context<'_>) -> Poll<Option<Self::ToBehaviour>> {
+        Poll::Ready(None)
+    }
+
     /// Adds a closure that turns the input event into something else.
     fn map_in_event<TNewIn, TMap>(self, map: TMap) -> MapInEvent<Self, TNewIn, TMap>
     where
```

Upon closing a connection, `poll_close` would be called until it returns `Ready(None)` and all returned events would be delivered to the behaviour.

thomaseizinger:
Instead of passing the `Handler` along, how about we add a callback to `ConnectionHandler` where it can produce a final message that is then passed to the behaviour:

```rust
trait ConnectionHandler {
	fn on_close(self) -> Option<Self::OutEvent>;
}
```

thomaseizinger:
> Instead of passing the `Handler` along, how about we add a callback to `ConnectionHandler` where it can produce a final message that is then passed to the behaviour:
> 
> ```rust
> trait ConnectionHandler {
> 	fn on_close(self) -> Option<Self::OutEvent>;
> }
> ```

[...]

PR #3099, opened 2022-11-09
feat(swarm)!: Allow `NetworkBehaviour`s to manage incoming connections

### LINKED ISSUE #2824: swarm/: Support generic connection management

thomaseizinger:
## Description

Previously, `ConnectionHandler` employed the prototype pattern via the `IntoConnectionHandler` abstraction. This allowed the `Swarm` to construct an instance of `IntoConnectionHandler` before the connection was established. This abstraction is however unnecessary. We can model all existing usecases by delaying the call to `NetworkBehavour::new_handler` until the connection is established. Not only does this delete a lot of code, it also makes several APIs simpler:

- `NetworkBehaviour`s can track state for future connections within themselves (indexed by `PeerId`) and pass it into the `ConnectionHandler` upon construction. This removes the need for passing along a handler in `NetworkBehaviourAction::Dial`.
- Removing `Handler` from `NetworkBehaviourAction::Dial` also avoids the need to pass the handler back into the behaviour in `inject_dial_failure`.
- By making `NetworkBehaviour::new_handler` fallible, `NetworkBehaviour`s can implement almost arbitrary connection management policies by denying the construction of a `ConnectionHandler` for a newly established connection.

Resolves #2824.

## Change checklist


protocols/dcutr/src/behaviour.rs
```diff
@@ -86,11 +87,26 @@ impl Behaviour {
 impl NetworkBehaviour for Behaviour {
     type ConnectionHandler = handler::Prototype;
     type OutEvent = Event;
+    type DialPayload = handler::Prototype;
```
thomaseizinger: This is a hack for now, the idea would be not pass the handler here but just a boring data structure and then construct the handler inside the callback.

thomaseizinger:
Unfortunately, https://github.com/rust-lang/rust/issues/29661 seems stalled which would be really nice to have here as most users probably don't care about the `DialPayload`.

thomaseizinger:
> The reason DialPayload is an Option in new_outbound_handler is because dials can also be issued via Swarm::dial in which case we don't have a payload.

This is actually quite annoying.

There is obviously value in having a `Swarm::dial` function but at the same time, we pay for this convenience by having to introduce an `Option` in the `new_outbound_handler` API.

There are multiple ways forward:

1. Introduce the `Option` in the API. This is bad because users need to understand that for dials issued by their own behaviour, this `Option` is always `Some` whereas for dials issued through `Swarm::dial`, it will be `None`. We could fix this by building our own `Option` type with some documentation attached.
2. Remove `Swarm::dial` function. This function does provide convenience, so I am not sure we can go this route. Personally, I've mostly used it in tests though. We could replace it by offering a `Control` `NetworkBehaviour` that exposes all actions (dial peer, close connection, etc) as functions.
3. Force users to pass a `DialPayload` into `Swarm::dial`. This won't play particularly well with `derive(NetworkBehaviour)` because the type will be some complex nesting of `Either`.

I think (2) would be the cleanest. It does come with a mild usability hit because users need to compose another `NetworkBehaviour` to get access to some of these functions. On the flipside, I think it is a step in the same direction as the generic connection management issue: We strip down the capabilities of `Swarm` and instead delegate to implementations of `NetworkBehaviour`.

rkuhn:
The answer also depends on what `inject_dial_failure` shall mean, and to whom. Currently it is called upon any failure for all composed behaviours, but that won’t work anymore if every behaviour then expects its own payload — the dial was caused by one behaviour, with only one specific payload. So the composed behaviour needs to dispatch the failure to only exactly that sub-behaviour which requested the dial. While this makes some amount of sense, it is also weird, because dial failures are not specific to a particular sub-behaviour, their scope is the target peer.

rkuhn:
BTW: `Swarm::dial` could also be kept in this case, and it could carry its own payload that is returned to the swarm poller upon dial failure — and none of the composed behaviours are informed in this case.

thomaseizinger:
> The answer also depends on what `inject_dial_failure` shall mean, and to whom. Currently it is called upon any failure for all composed behaviours, but that won’t work anymore if every behaviour then expects its own payload — the dial was caused by one behaviour, with only one specific payload. So the composed behaviour needs to dispatch the failure to only exactly that sub-behaviour which requested the dial. While this makes some amount of sense, it is also weird, because dial failures are not specific to a particular sub-behaviour, their scope is the target peer.

That is right! This is more tricky than I thought. If I wanted to inject a `DialPayload` on every `on_new_outbound` call, I'd have to make that one up for all other behaviours that are composed because I need a handler for every behaviour.

thomaseizinger:
The question is, does a new outbound connection initiated by a behaviour even need policing? Surely that behaviour would have just not initiated the dial in the first place then.

Following that logic, we should probably only try to remove the `IntoConnectionHandler` abstraction.

We need a level of indirection though because we can't construct the handler directly within `NetworkBehaviourAction::Dial`. I'd consider that an edgecase though so maybe we should only retain the abstraction for that and only change the `new_handler` function to accept the `PeerId` parameter?

thomaseizinger:
We could introduce a `DialSource` enum:

```rust
enum DialSource<P> {
    Swarm,
    OtherBehaviour,
    ThisBehaviour {
        payload: P
    }
}
```

This would allow us to still notify each behaviour about a failed dial but only carry the data along for the one behaviour that issued it.

rkuhn:
My primary worry right now is not the structure of IntoConnectionHandler, so I don’t have good inputs there, but the proposed DialSource enum sounds very useful to me!

thomaseizinger:
I just had another idea which I think is much better:

The reason for passing a `handler` along with `Dial` was to enable users to directly inject a message into the new connection.

Previously, that required really complex tracking of state in the behaviour because of the `IntoConnectionHandler` abstraction and the delayed access to `PeerId`.

If we remove this abstraction, it is really easy for `NetworkBehaviour`s to track initial state for a new connection: Just make a HashMap indexed by `PeerId` and pick out the data whenever `new_outbound_handler` gets called!

That is functionally equivalent to what we have today and makes several APIs a lot simpler.

thomaseizinger:
> We could introduce a `DialSource` enum:
> 
> ```rust
> enum DialSource<P> {
>     Swarm,
>     OtherBehaviour,
>     ThisBehaviour {
>         payload: P
>     }
> }
> ```
> 
> This would allow us to still notify each behaviour about a failed dial but only carry the data along for the one behaviour that issued it.

I tried this and couldn't make it work unfortunately.

swarm/src/lib.rs
```diff
@@ -1401,14 +1361,16 @@ where
...
+        let supported_protocols = todo!();
+
+        // let supported_protocols = self
+        //     .behaviour
+        //     .new_handler()
+        //     .inbound_protocol()
+        //     .protocol_info()
+        //     .into_iter()
+        //     .map(|info| info.protocol_name().to_vec())
+        //     .collect();
```
thomaseizinger: This turns out to be difficult.

@mxinden I remember you saying that `PollParameters` was always something that didn't appeal to you. Did you ever think about alternative design? This might be a good time to bring them up :)

thomaseizinger: I am now initialising these to an empty vector and they are updated every time a new connection is made.

swarm/src/behaviour.rs
```diff
@@ -288,188 +288,12 @@ pub enum NetworkBehaviourAction<TOutEvent, TInEvent> {
...
+    /// of the newly established connection.
     ///
-    /// Note that the provided handler is returned to the [`NetworkBehaviour`] on connection failure
-    /// and connection closing. Thus it can be used to carry state, which otherwise would have to be
-    /// tracked in the [`NetworkBehaviour`] itself. E.g. a message destined to an unconnected peer
-    /// can be included in the handler, and thus directly send on connection success or extracted by
-    /// the [`NetworkBehaviour`] on connection failure.
+    /// Once the handler is constructed, [`NetworkBehaviour::inject_connection_established`] will be called.
     ///
-    /// # Example carrying state in the handler
```
thomaseizinger, line 442: Should we replace this example? Now that `new_handler` gives you a `PeerId`, I think it is quite easy to realize that you can store data for a future connection in the behaviour and just pass it into the handler here.

swarm/src/behaviour.rs
```diff
@@ -270,7 +260,7 @@ pub trait PollParameters {
     ///
     /// The iterator's elements are the ASCII names as reported on the wire.
     ///
-    /// Note that the list is computed once at initialization and never refreshed.
+    /// This list refreshed every time a new connection is established.
```
```suggestion
    /// This list is refreshed every time a new connection is established.
```

mxinden:
> * `NetworkBehaviour`s can track state for future connections within themselves (indexed by `PeerId`) and pass it into the `ConnectionHandler` upon construction. This removes the need for passing along a handler in `NetworkBehaviourAction::Dial`.

Say there are two `NetworkBehaviour` implementations, each requesting a new connection to a new peer. In such case, when `NetworkBehaviour::inject_connection_established` is called, neither of them knows whether this new connection corresponds to their dial request.

I am not sure whether we need to design for the above race condition. Your proposal might be just fine. What do folks think?

swarm/src/lib.rs
mxinden: > @mxinden I remember you saying that `PollParameters` was always something that didn't appeal to you.

Yes, very much. I would like to not have `PollParameters`.

> I am now initialising these to an empty vector and they are updated every time a new connection is made.

In my eyes that is the better behavior.

mxinden (commented):
I am **very** much in favor of this refactoring. I still remember in my early days that it took me quite a bit to understand the `IntoConnectionHandler` abstraction and why we need it in the first place.

swarm/src/behaviour.rs
mxinden: > * `NetworkBehaviour`s can track state for future connections within themselves (indexed by `PeerId`) and pass it into the `ConnectionHandler` upon construction. This removes the need for passing along a handler in `NetworkBehaviourAction::Dial`.

Say there are two `NetworkBehaviour` implementations, each requesting a new connection to a new peer. In such case, when `NetworkBehaviour::inject_connection_established` is called, neither of them knows whether this new connection corresponds to their dial request.

We could still allow the user to provide user data via `DialOpts`. For the case of `Swarm::dial` and `NetworkBehaviour::inject_dial_failure` we could wrap this in an `Option`.

I am not sure whether we need to design for the above race condition. Your proposal might be just fine. What do folks think?

rkuhn: In `ipfs-embed` I validate advertised peer addresses by dialling them — and closing such connections upon success. This is only reliably possible if I can detect whether the connection resulted from my own dialling attempt. OTOH, another cause may have resulted in dialling that same peer with that same address for other reasons, which would then possibly have said “nah, attempt is already underway”. But all of this would happen _in addition to existing connections_, so it might not be a huge problem.

thomaseizinger:
This is a fairly invasive breaking change that will hit most users. I think it may be worth it try and make it a non-breaking one. Let me try something ...

thomaseizinger:
> This is a fairly invasive breaking change that will hit most users. I think it may be worth it try and make it a non-breaking one. Let me try something ...

Actually, nevermind. I don't think it is possible. I was thinking along the lines of adding:

```rust
	#[deprecated]
    fn new_handler(&mut self) -> Self::ConnectionHandler;

    fn new_connection_handler(&mut self, peer: &PeerId, connected_point: &ConnectedPoint) -> Result<<Self::ConnectionHandler as IntoConnectionHandler>::Handler, ConnectionDenied> {
        Ok(self.new_handler().into_handler(peer, connected_point))
    }
```

This part would work but the problem is `NetworkBehaviourAction::Dial`. We would have to deprecate that one too and add a new one that doesn't take a `Handler`.

Not sure if worth the effort?

thomaseizinger:
My suggestion would be:

1. Make a release of all the existing changes, esp. the deprecation of the `inject` functions.
1. Merge this afterwards

I've made the `new_handler` function fallible right away too because I didn't want to have users go through two iterations of breaking changes. `ConnectionDenied` is an empty, non-exhaustive enum for now so we can extend it.

It would be good to get feedback on the design here. Using an enum gives us type-safety but it doesn't play well with the idea of having generic connection management. Ideally, a behaviour can define its own error type for why it wants to reject a connection. This however would add another associated type to the trait, complicating the interface.

cc @divagant-martian @mxinden

thomaseizinger:
> Ideally, a behaviour can define its own error type for why it wants to reject a connection. This however would add another associated type to the trait, complicating the interface.

On the other hand, this would make it also very clear, which behaviour rejected a connection.

thomaseizinger:
The last commit adds an associated type to the `NetworkBehaviour` that describes the error for why construction of the `ConnectionHandler` failed. This can be used for almost arbitrary connection management policies. See https://github.com/libp2p/rust-libp2p/pull/3156 as an example.

cc @divagant-martian

swarm/src/behaviour.rs
thomaseizinger: Do you need to actively close them? Assuming reasonable `keep_alive` implementations of your `ConnectionHandler`s, the connection should get closed automatically if it is not in use.

> OTOH, another cause may have resulted in dialling that same peer with that same address for other reasons, which would then possibly have said “nah, attempt is already underway”.

Can this be expressed with [`PeerCondition::NotDialing`](https://docs.rs/libp2p/0.49.0/libp2p/swarm/dial_opts/enum.PeerCondition.html#variant.NotDialing)?

Note that `new_handler` also gives you access to `ConnectedPoint`. So in addition to storing data in your behaviour based on `PeerId`, you could also index it by the dialed address. `new_handler` being called for an address you wanted to validate IS the validation that a connection was made to this address. If you have a dedicated `AddressValidationBehaviour`, that behaviour could then straight up deny that connection which instantly closes it again.

swarm/src/lib.rs
thomaseizinger: > > @mxinden I remember you saying that `PollParameters` was always something that didn't appeal to you.
> 
> Yes, very much. I would like to not have `PollParameters`.

Tracking this here: https://github.com/libp2p/rust-libp2p/issues/3124

thomaseizinger:
> The last commit adds an associated type to the `NetworkBehaviour` that describes the error for why construction of the `ConnectionHandler` failed. This can be used for almost arbitrary connection management policies. See #3156 as an example.

The downside of this design is that we need to auto-generate an enum within `derive(NetworkBehaviour)`. Generating the out-event enum is already enough magic I find. Adding one more enum to that doesn't feel great. This however just gave me an idea.

What if we define the following type:

```rust
pub enum OutEvent<A = Void, B = Void, C = Void, D = Void, E = Void, F = Void, G = Void, H = Void, I = Void> {
	A(A),
	B(B),
	C(C),
	D(D),
	E(E),
	F(F),
	G(G),
	H(H),
	I(I)
}
```

And have the derive macro fills in the types based on the number of fields? The names are not pretty but:
- it presents the same amount of type safety
- IDEs would allow you to jump to the definition
- you could abstract over the outevent of different `NetworkBehaviour`s 
- we could use the same type for the connection denied associated type too

thomaseizinger:
TODO:

- [x] Always box denied reason to avoid generating an enum
- [ ] Figure out how to manage pending connections

thomaseizinger:
Here is a proposal regarding pending connection management:

- Remove `addresses_of_peer` function
- Add:
	- 	```rust
		fn new_outgoing_connection(Option<PeerId>, &[Multiaddr]) -> Result<Vec<Multiaddr>, Box<dyn std::error::Error + Send + 'static>>
		```
 	- 	```rust
 		fn new_incoming_connection(local_addr: Multiaddr, send_back_addr: Multiaddr) -> Result<(), Box<dyn std::error::Error + Send + 'static>>
		```

The `new_outgoing_connection` replaces the `addresses_of_peer` function. It is given the `PeerId` - if known - and the addresses we already have that we will try to dial. It can decline the dial by returning an error or provide even more addresses to use.

The `new_incoming_connection` is given the local and remote's address. It doesn't return anything usually as there is no information needed to upgrade a connection.

I am reasonably happy with this design. We are only extending the `NetworkBehaviour` by one function, all three `new_` functions follow a similar naming pattern and function signature.

Thoughts @mxinden?

swarm/src/behaviour.rs
```diff
@@ -907,11 +609,11 @@ pub struct ConnectionEstablished<'a> {
 /// This event is always paired with an earlier
 /// [`FromSwarm::ConnectionEstablished`] with the same peer ID, connection ID
 /// and endpoint.
-pub struct ConnectionClosed<'a, Handler: IntoConnectionHandler> {
+pub struct ConnectionClosed<'a, Handler> {
     pub peer_id: PeerId,
     pub connection_id: ConnectionId,
     pub endpoint: &'a ConnectedPoint,
-    pub handler: <Handler as IntoConnectionHandler>::Handler,
+    pub handler: Handler,
```
elenaf9, line 616: Out of scope for this PR, but I do wonder if we really need the `handler` field on `ConnectionClosed`. 
The `handler` property was added in #2191 together with adding it to `Dial`, `DialFailure` and `ListenFailure`. If I understand it correctly, the main motivation behind it was:

> it allows one to attach state to a dial request, thus not having to keep track of it within a NetworkBehaviour implementation 

Per https://github.com/libp2p/rust-libp2p/pull/3099#issuecomment-1313284181 attaching a state to a dial is not needed anymore. What's the use-case of still returning the handler in `ConnectionClosed`? As far as I can tell, none of our behaviours ever use it.

Motivation behind the question is that it would be great if we could get rid of the duplication between `SwarmEvent` and `FromSwarm`.
Until now the main difference was that `FromSwarm` also contains the handler, which should not be exposed in the `SwarmEvent`[^1]: https://github.com/libp2p/rust-libp2p/discussions/2423#discussioncomment-1941073. If we could remove it here, we might be able to do something like:
```rust
enum SwarmEvent<TBehaviourOutEvent> {
    Behaviour(TBehaviourOutEvent),
    FromSwarm(FromSwarmEvent)
}
```

[^1]: There are also some other differences; `FromSwarm` contains `ConnectionId`s, and the `NewExternalAddr` and `ExpiredExternalAddr` variants. But I don't see see a drawback in exposing them to the user.

elenaf9 (commented):
Very much in favor of this! Did not do a full review yet.

swarm/src/behaviour.rs
thomaseizinger: Related: https://github.com/libp2p/rust-libp2p/issues/3046

thomaseizinger:
Any ideas on how to make this less breaking would be gladly appreciated! :)

thomaseizinger:
> Here is a proposal regarding pending connection management:
> 
> - Remove `addresses_of_peer` function
> - Add:
> 	- 	```rust
> 		fn new_outgoing_connection(Option<PeerId>, &[Multiaddr]) -> Result<Vec<Multiaddr>, Box<dyn std::error::Error + Send + 'static>>
> 		```
>  	- 	```rust
>  		fn new_incoming_connection(local_addr: Multiaddr, send_back_addr: Multiaddr) -> Result<(), Box<dyn std::error::Error + Send + 'static>>
> 		```
> 
> The `new_outgoing_connection` replaces the `addresses_of_peer` function. It is given the `PeerId` - if known - and the addresses we already have that we will try to dial. It can decline the dial by returning an error or provide even more addresses to use.
> 
> The `new_incoming_connection` is given the local and remote's address. It doesn't return anything usually as there is no information needed to upgrade a connection.
> 
> I am reasonably happy with this design. We are only extending the `NetworkBehaviour` by one function, all three `new_` functions follow a similar naming pattern and function signature.
> 
> Thoughts @mxinden?

One slight variation on this: `new_handler` could now be called `new_established_connection` to be more consistent with the two others. We could also split it into `new_established_inbound_connection` and `new_established_outbound_connection`. It is a bit odd at the moment that for pending connections, the function is split by inbound and outbound but for the established one it is not.

thomaseizinger:
I think we can limit the breaking changes of this PR to removing the handler from `Dial`. Everything else should be possible to do in a non-breaking way.

swarm/CHANGELOG.md
```diff
@@ -96,6 +97,10 @@
 
 - Update `rust-version` to reflect the actual MSRV: 1.62.0. See [PR 3090].
 
+- Remove `IntoConnectionHandler` abstraction and change the signature of `NetworkBehaviour::new_handler` to accept `PeerId` and `ConnectedPoint`.
```
mxinden, line 100: What do you think of a small before and after code snippet to help folks upgrade?

swarm/src/behaviour.rs
```diff
@@ -141,7 +139,11 @@ pub trait NetworkBehaviour: 'static {
     ///
     /// Note that the handler is returned to the [`NetworkBehaviour`] on connection failure and
     /// connection closing.
-    fn new_handler(&mut self) -> Self::ConnectionHandler;
+    fn new_handler(
+        &mut self,
+        peer: &PeerId,
```
mxinden, line 144:
```suggestion
        peer: PeerId,
```

Why provide a reference to a `Copy` type?

mxinden:
> One slight variation on this: `new_handler` could now be called `new_established_connection` to be more consistent with the two others.

Sounds good to me.

> We could also split it into `new_established_inbound_connection` and `new_established_outbound_connection`. It is a bit odd at the moment that for pending connections, the function is split by inbound and outbound but for the established one it is not.

Do I understand correctly that both methods would have the same signature and that we don't foresee the two signatures to diverge any time soon? If so, I suggest not splitting it. While nice for the sake of consistency, I don't think it justifies the additional complexity.

protocols/relay/src/v2/client.rs
```diff
@@ -96,6 +99,8 @@ pub struct Client {
     /// connection.
     directly_connected_peers: HashMap<PeerId, Vec<ConnectionId>>,
 
+    initial_events: HashMap<PeerId, handler::In>,
```
mxinden, line 102:
```suggestion
    /// [`handler::In`] event to be provided to a handler in `NetworkBehaviour::new_handler` once the corresponding connection is established.
    initial_events: HashMap<PeerId, handler::In>,
```

What do you think?

protocols/relay/src/v2/client.rs
```diff
@@ -146,11 +152,40 @@ impl Client {
...
+            let mut handler = Handler::new(
+                self.local_peer_id,
+                *peer,
+                connected_point.get_remote_address().clone(),
+            );
+
+            if let Some(event) = self.initial_events.remove(peer) {
+                #[allow(deprecated)]
+                handler.inject_event(event)
+            }
```
mxinden, line 185: Injecting events through two ways, i.e. (1) at creation and (2) regularly during the lifetime of the handler via `Swarm`, seems inconsistent to me. That said, I can not think of a better way. I suggest keeping as is.

protocols/relay/src/v2/client.rs
```diff
@@ -146,11 +152,40 @@ impl Client {
...
+        peer: &PeerId,
+        connected_point: &ConnectedPoint,
+    ) -> Result<Self::ConnectionHandler, Box<dyn std::error::Error + Send + 'static>> {
+        if connected_point.is_relayed() {
+            if let Some(event) = self.initial_events.remove(peer) {
+                log::debug!(
+                    "Established relayed instead of direct connection to {:?}, \
+                     dropping initial in event {:?}.",
+                    peer,
+                    event
```
mxinden, line 169: How do we know that there is no second **direct** connection in progress of establishment to the given peer? How do we know that this event was predetermined for this connection and not another connection still being established?

This is rather complex. I rewrote this comment 4 times. I might be missing something.

protocols/relay/src/v2/client.rs
```diff
@@ -269,16 +305,14 @@ impl NetworkBehaviour for Client {
...
-                            self.local_peer_id,
-                            Some(handler::In::Reserve { to_listener }),
-                        );
+                        self.initial_events
+                            .insert(relay_peer_id, handler::In::Reserve { to_listener });
+
                         NetworkBehaviourAction::Dial {
                             opts: DialOpts::peer_id(relay_peer_id)
                                 .addresses(vec![relay_addr])
                                 .extend_addresses_through_behaviour()
```
mxinden, line 314: Note that some `NetworkBehaviour` might add a relayed address here.

The relay protocol should never be spoken on top of a relayed connection. We should never do nested relaying as otherwise one could build circular relayed connections that could be misused for a DOS attack. Thus returning a `dummy::ConnectionHandler` on relayed connections is critical.

That said, we might be establishing both a relayed and a direct connection to a given peer in parallel. How do we know which one the `initial_events` event was destined for?

Not sure how to prevent this. The solution you do above, namely to drop the `initial_events` event in case it turns out to be a relayed connection covers the case where there is only a single connection attempt in progress. What about the case where both (relayed and direct) are in progress?

thomaseizinger:
> > We could also split it into `new_established_inbound_connection` and `new_established_outbound_connection`. It is a bit odd at the moment that for pending connections, the function is split by inbound and outbound but for the established one it is not.
> 
> Do I understand correctly that both methods would have the same signature and that we don't foresee the two signatures to diverge any time soon? If so, I suggest not splitting it. While nice for the sake of consistency, I don't think it justifies the additional complexity.

If split, I would inline the variants of the `ConnectedPoint` enum: https://github.com/libp2p/rust-libp2p/blob/f8f19baad02e204506dedb9680a039121fdab831/core/src/connection.rs#L116

So no, the signatures would not be the same.

divagant-martian:
Sorry for the delay. As far as lighthouse is concerned, this would work really well in particular having the `ConnectedPoint`, since most of the time we decide based on percentage of inbound and outbound connections. Regarding the possibly generic error, I see why it would be a nice to have, but at least in our case that can be directly handled when creating the handler and wouldn't really need the error reported by the swarm.

Also of course, thanks for all these efforts!

protocols/relay/src/v2/client.rs
thomaseizinger: I only 1-to-1 ported the code that already existed, as far as I know, no functionality should currently change in this PR.

protocols/relay/src/v2/client.rs
thomaseizinger: We can index the `HashMap` by the dialed address instead of the peer ID. That should fix things I guess?

protocols/relay/src/v2/client.rs
thomaseizinger: We could hold the event back here and instead listen for the `ConnectionEstablished` event and then emit an event later?

thomaseizinger:
@mxinden: Instead of a new `DialId` or `DialToken`, we could also create the `ConnectionId` as part of the dial attempt. The `Pool` would later allocate one anyway so we could actually directly use that instead. This gives us one identifier for the entire connection from the moment a dial is issued up until the connection gets closed.

thomaseizinger:
> @mxinden: Instead of a new `DialId` or `DialToken`, we could also create the `ConnectionId` as part of the dial attempt. The `Pool` would later allocate one anyway so we could actually directly use that instead. This gives us one identifier for the entire connection from the moment a dial is issued up until the connection gets closed.

I think with that idea, we might be able to ship the removal of `Dial::Handler` as a separate patch. That would be nice.

thomaseizinger:
> > @mxinden: Instead of a new `DialId` or `DialToken`, we could also create the `ConnectionId` as part of the dial attempt. The `Pool` would later allocate one anyway so we could actually directly use that instead. This gives us one identifier for the entire connection from the moment a dial is issued up until the connection gets closed.
> 
> I think with that idea, we might be able to ship the removal of `Dial::Handler` as a separate patch. That would be nice.

Turns out we can't but overall the version in https://github.com/libp2p/rust-libp2p/pull/3254 is a lot less breaking which is nice :)

PR #3101, opened 2022-11-09
transports/tcp: Update if-watch to `v3.0.0`

rkuhn:
## Description

Update to `if-watch` version 3.0.0 and pass through features, such that `libp2p-tcp/async-io` selects `if-watch/smol` and `libp2p-tcp/tokio` brings in `if-watch/tokio`.
The mDNS part is already done in #3096.

supersedes #3095

## Change checklist

## Commit message body

Release 3.0.0 introduces executor-specific features which we now activate accordingly.


transports/tcp/src/provider/async_io.rs
```diff
@@ -55,6 +55,15 @@ pub enum Tcp {}
...
     type Stream = Async<net::TcpStream>;
     type Listener = Async<net::TcpListener>;
+    type IfWatcher = if_watch::smol::IfWatcher;
+
+    fn new_if_watcher() -> io::Result<Self::IfWatcher> {
+        Self::IfWatcher::new()
+    }
+
+    fn addrs(if_watcher: &Self::IfWatcher) -> Vec<if_watch::IpNet> {
+        if_watcher.iter().copied().collect()
```
rkuhn, line 65: Returning the iterator would be possible using a HRTB, but given that this is only called in a rare error path I opted for the simpler approach.

transports/tcp/Cargo.toml
```diff
@@ -11,26 +11,26 @@ keywords = ["peer-to-peer", "libp2p", "networking"]
...
 log = "0.4.11"
 socket2 = { version = "0.4.0", features = ["all"] }
-tokio-crate = { package = "tokio", version = "1.19.0", default-features = false, features = ["net"], optional = true }
+tokio = { version = "1.19.0", default-features = false, features = ["net"], optional = true }
 
 [features]
-tokio = ["tokio-crate"]
-async-io = ["async-io-crate"]
+tokio = ["dep:tokio", "if-watch/tokio"]
+async-io = ["dep:async-io", "if-watch/smol"]
```
rkuhn, line 26: switching to `dep:tokio` style as was done in other crates — let me know if this one was deliberately left unchanged

rkuhn (commented):
/cc @thomaseizinger @mxinden @jxs

transports/tcp/src/provider/tokio.rs
```diff
@@ -39,7 +39,7 @@ use std::task::{Context, Poll};
 /// # use libp2p_core::Transport;
 /// # use futures::future;
 /// # use std::pin::Pin;
-/// # use tokio_crate as tokio;
+/// # use tokio as tokio;
```
jxs:
```suggestion
/// # use tokio;
```

transports/tcp/CHANGELOG.md
```diff
@@ -1,11 +1,14 @@
 # 0.38.0 [unreleased]
 
+- Update `if-watch` to version 3.0.0 and pass through `tokio` and `async-io` features. See [PR 3101].
```
jxs:
```suggestion
- Update to `if-watch`  `v3.0.0` and pass through `tokio` and `async-io` features. See [PR 3101].
```

jxs (approved):
LGTM Roland!

transports/tcp/src/provider/tokio.rs
rkuhn: ah, good catch, I think this can be removed completely

transports/tcp/src/provider.rs
```diff
@@ -46,6 +48,14 @@ pub trait Provider: Clone + Send + 'static {
     type Stream: AsyncRead + AsyncWrite + Send + Unpin + fmt::Debug;
     /// The type of TCP listeners obtained from [`Provider::new_listener`].
     type Listener: Send + Unpin;
+    /// The type of IfWatcher obtained from [`Provider::new_if_watcher`].
+    type IfWatcher: Stream<Item = io::Result<IfEvent>> + Send + Unpin;
```
thomaseizinger, line 52:
```suggestion
    type IfWatcher: Stream<Item = io::Result<IfEvent>> + IntoIterator<Item = IpNet> + Send + Unpin;
```

Is this legal Rust? My brain compiler is unsure but it could avoid the need for the `addrs` function I think.

thomaseizinger (approved):
One idea, otherwise LGTM!

transports/tcp/src/provider.rs
rkuhn: It is legal syntax, but neither IfWatcher implementation implements this trait, plus `into_iter` consumes `self`, so we wouldn’t be able to use this here.

thomaseizinger: Ah damn!

The naming of `iter` suggested that we actually implement the `Iterator` trait but I guess the consuming nature of `IntoIterator` makes this unworkable anyway.

Thanks for checking!

thomaseizinger (approved):

rkuhn:
anything still needed from my side to merge this?

thomaseizinger:
> anything still needed from my side to merge this?

No. For some reason mergify doesn't like this PR. I'll have to look into that.

thomaseizinger:
I also can't update the branch manually. Getting the same error as shown in the mergify check.

thomaseizinger:
@rkuhn Can you push a merge-commit with master manually?

rkuhn:
all done :-)

Issue #3100, opened 2022-11-09
The chat-tokio example is broken since v0.45.0


Ma233:
## Summary

I tried chat-tokio example on `v0.44.0`, `v0.45.0` and `v0.49.0`. Only worked on `v0.44.0`.
It seems to be something wrong on mdns as this issue #2676 mentioned.

FYI, this is my system info:
```
❯ uname -a
Darwin XXXXX 21.6.0 Darwin Kernel Version 21.6.0: Mon Aug 22 20:19:52 PDT 2022; root:xnu-8020.140.49~2/RELEASE_ARM64_T6000 arm64
```

## Expected behaviour

First setup some chat nodes by `cargo run --features="tcp-tokio mdns" --example chat-tokio`.
Then when you input some words in a node, the other nodes should receive it and print it out.

## Actual behaviour

The other nodes do not print anything after input.
And I logged the events, the connection event was not send after nodes setup.

<details><summary>Debug Output</summary>
<p>

```
Local peer id: PeerId("12D3KooWPwtV3juE79L1d3Y2Xc1UNFw8BTVsuwkXKv2qgwgeJtgd")
[2022-11-09T08:53:17Z DEBUG libp2p_tcp] listening on 0.0.0.0:0
[2022-11-09T08:53:17Z INFO  libp2p_mdns::behaviour::iface] creating instance on iface 10.206.53.188
[2022-11-09T08:53:17Z DEBUG libp2p_tcp] New listen address: /ip4/10.206.53.188/tcp/57067
[2022-11-09T08:53:17Z DEBUG libp2p_swarm] Listener ListenerId(1); New address: "/ip4/10.206.53.188/tcp/57067"
Listening on "/ip4/10.206.53.188/tcp/57067"
[2022-11-09T08:53:17Z DEBUG libp2p_tcp] New listen address: /ip4/127.0.0.1/tcp/57067
[2022-11-09T08:53:17Z DEBUG libp2p_swarm] Listener ListenerId(1); New address: "/ip4/127.0.0.1/tcp/57067"
Listening on "/ip4/127.0.0.1/tcp/57067"
[2022-11-09T08:53:17Z DEBUG libp2p_mdns::behaviour::iface::query] Parsing mdns packet failed: LabelIsNotAscii
[2022-11-09T08:53:17Z DEBUG libp2p_mdns::behaviour::iface::query] Parsing mdns packet failed: LabelIsNotAscii
[...]

Issue #2676, opened 2022-05-28
mDNS: LabelIsNotAscii


thearossman:
## Summary

mDNS discovery sometimes fails with "parsing MDNS packet failed" message; may be an old `mdns_parser` issue in general? 

## Expected behaviour

I'd *assume* that all mDNS packets sent/received should be correctly parsed? And that I shouldn't see this error message? 

## Actual behaviour

mDNS sometimes works, and sometimes prints the debug message: `DEBUG libp2p_mdns::behaviour::iface::query] Parsing mdns packet failed: LabelIsNotAscii`. 

After a threshold of errors, mDNS seems to begin behaving inconsistently -- e.g., expiring peers and erroneously disconnecting. (So, for example, if I'm using mDNS to do peer discovery for a GossipSub network, the network will eventually fail w/ insufficient peers or uptime will suffer.)

This first showed up on a separate project, but I was able to observe the same behavior running the `chat-tokio` example, which uses mDNS. I've reproduced this on multiple libp2p versions, including the most recent (0.44.0) and older (e.g., 0.39.1).

```
[2022-05-28T01:49:34Z DEBUG libp2p_mdns::behaviour::iface::query] Parsing mdns packet failed: LabelIsNotAscii
[2022-05-28T01:49:35Z DEBUG libp2p_mdns::behaviour::iface::query] Parsing mdns packet failed: LabelIsNotAscii
[2022-05-28T01:49:36Z DEBUG libp2p_mdns::behaviour::iface::query] Parsing mdns packet failed: LabelIsNotAscii
```

## Possible Solution

This is an error from dns_parser [here](https://github.com/tailhook/dns-parser/blob/191266705b30feabe9fbf9289b4cb4da0ea2be31/src/error.rs). Seems like others have encountered it [before](https://github.com/dylanmckay/mdns/issues/13).

I've been working around with a high TTL and query interval. 

## Version
0.44.0, 0.39.1 (I've been able to reproduce this on multiple versions)

mxinden:
Do these mdns packets come from libp2p hosts? If no, this should not affect `libp2p-mdns`.

thearossman:
I believe they come from libp2p hosts -- we're not using anything else, and we were able to get the same error just by running the chat-tokio example with nothing else (https://github.com/libp2p/rust-libp2p/blob/master/examples/chat-tokio.rs).

mxinden:
@thearossman can you use wireshark to post a dump of the original packet here?

[...]

PR #3102, opened 2022-11-09
refactor(mdns): Parse messages using `trust-dns-proto` instead of `dns-parse`

jxs:
# Description

While researching #3100 and #2676, I found that [`dns-parser`](https://github.com/tailhook/dns-parser) the dependency we are using to parse the DNS messages [only supports ASCII names](https://github.com/tailhook/dns-parser/blob/dfe8bc3e9272f559f2b960ff7f5d71dbb60faf4f/src/name.rs#L75),  [IDN](https://en.wikipedia.org/wiki/Internationalized_domain_name)'s seem to [support UTF8](https://www.rfc-editor.org/rfc/rfc6055#section-3). 

`dns-parser` looks unmaintained since 2018 and `trust-dns` [supports UTF8](https://github.com/bluejekyll/trust-dns/blob/main/crates/proto/src/rr/domain/name.rs#L1122-L1126), so while I am not sure this PR will solve the issues above mentioned, it will maybe help address the flakiness mentioned on #2676 (I myself have had cases where the `mdns` tests fail  sometimes) since `trust-dns` is more popular and probably used. We will also stop having `DEBUG` logs for UTF8 DNS messages.

#3100 and #2676 

## Change checklist


protocols/mdns/src/lib.rs
```diff
@@ -49,8 +49,12 @@ pub use crate::behaviour::TokioMdns;
 
 /// The DNS service name for all libp2p peers used to query for addresses.
 const SERVICE_NAME: &[u8] = b"_p2p._udp.local";
+/// `SERVICE_NAME` as a Fully Qualified Domain Name.
+const SERVICE_NAME_FQDN: &str = "_p2p._udp.local.";
```
jxs, line 53: `trust-dns` parses `Messages` as [`FQDN`](https://en.wikipedia.org/wiki/Fully_qualified_domain_name) therefore the trailing dot.

thomaseizinger: This might be an issue in the spec: https://github.com/libp2p/specs/blob/master/discovery/mdns.md

jxs: Thanks for referring this Thomas, the PR is only using `trust-dns` to parse incoming messages, dns messages sent are still constructed with [`build_query`](https://github.com/libp2p/rust-libp2p/blob/313ed588fec84c4307998f04282b708af70d9012/protocols/mdns/src/behaviour/iface/dns.rs#L72) [`build_query_response`](https://github.com/libp2p/rust-libp2p/blob/313ed588fec84c4307998f04282b708af70d9012/protocols/mdns/src/behaviour/iface/dns.rs#L106), and [`build_service_discovery_response`](https://github.com/libp2p/rust-libp2p/blob/313ed588fec84c4307998f04282b708af70d9012/protocols/mdns/src/behaviour/iface/dns.rs#L168) which use the non `FQDN` names, so we should be ok right?

thomaseizinger: But if `trust-dns` requires a dot at the end, doesn't this mean we are now not spec-compliant and can't parse the DNS messages from other implementations?

jxs: `trust-dns` [follows the spec](http://www.dns-sd.org/trailingdotsindomainnames.html), it doesn't require a dot at the end, it puts it on constructed [`Message`](https://docs.rs/trust-dns-proto/0.22.0/trust_dns_proto/op/message/struct.Message.html)'s cause they are FQDN's. 
If you try constructing a `Message` from a name with a trailing dot it will not be valid, see [here](https://github.com/jxs/trust-dns-fqdns/blob/main/src/main.rs#L72). We could also [`Name::set_fqdn`](https://docs.rs/trust-dns-proto/0.22.0/trust_dns_proto/rr/domain/struct.Name.html#method.set_fqdn) as `false`, and it will lose the trailing dot.

thomaseizinger: Got you now, thanks for explaining!

thomaseizinger (approved):
LGTM, thanks!

protocols/mdns/src/lib.rs
jxs: no worries Thomas! 😊

Ma233:
Hi, is this pull request ready to be merged and released?

thomaseizinger:
> Hi, is this pull request ready to be merged and released?

I think this is safe to merge, yes.

thomaseizinger:
@jxs: This PR actually had a broken job: https://github.com/libp2p/rust-libp2p/actions/runs/3465658904/jobs/5788668836

thomaseizinger:
> @jxs: This PR actually had a broken job: [libp2p/rust-libp2p/actions/runs/3465658904/jobs/5788668836](https://github.com/libp2p/rust-libp2p/actions/runs/3465658904/jobs/5788668836)

I opened an issue here: https://github.com/bluejekyll/trust-dns/issues/1830

thomaseizinger:
I'd like to avoid dragging in the entirety of `tokio` in so I am tempted to revert this to get a clean master branch back.

We can then patch `trust-dns-proto`, wait for their next release and bring this change back.

Thoughts?

mxinden:
Reverting works for me. Thanks for fixing this upstream @jxs.

jxs:
> > @jxs: This PR actually had a broken job: [libp2p/rust-libp2p/actions/runs/3465658904/jobs/5788668836](https://github.com/libp2p/rust-libp2p/actions/runs/3465658904/jobs/5788668836)
> 
> I opened an issue here: [bluejekyll/trust-dns#1830](https://github.com/bluejekyll/trust-dns/issues/1830)

Thanks Thomas, yeah makes sense to revert. Meanwhile I submitted  https://github.com/bluejekyll/trust-dns/pull/1831.

thomaseizinger:
> > > @jxs: This PR actually had a broken job: [libp2p/rust-libp2p/actions/runs/3465658904/jobs/5788668836](https://github.com/libp2p/rust-libp2p/actions/runs/3465658904/jobs/5788668836)
> > 
> > I opened an issue here: [bluejekyll/trust-dns#1830](https://github.com/bluejekyll/trust-dns/issues/1830)
> 
> Thanks Thomas, yeah makes sense to revert. Meanwhile I submitted  https://github.com/bluejekyll/trust-dns/pull/1831.

Thanks for submitting the PR!

PR #3108, opened 2022-11-10
test: Remove `dialer_can_receive` stream muxer test

thomaseizinger:
## Description

Various muxer implementations struggle to fulfill this test. In practice, it doesn't matter much because we always run `multistream-select` on top of a newly negotiated stream so we never end up actually reading from a stream that we have never written to.

Relevant discussion: https://github.com/kpp/rust-libp2p/pull/27#discussion_r1012128418

## Change checklist


mxinden (approved):
I am fine with keeping it. I.e. I am fine with keeping a test that is only green on mplex.

That said, I don't think mplex is particularly relevant for rust-libp2p and thus I am fine removing it. Your call.

thomaseizinger:
> I am fine with keeping it. I.e. I am fine with keeping a test that is only green on mplex.
> 
> That said, I don't think mplex is particularly relevant for rust-libp2p and thus I am fine removing it. Your call.

My goal with this test suite is to act as a first step towards standardising, how streams in (rust-)libp2p should behave. A test that is only green on some implementations doesn't help with that.

Issue #2617, opened 2022-04-13
WASM support


mriise:
This issue is an attempt to gather information relating to WASM support for rust-libp2p as well as serve as somewhat of a tracking / discussion issue.

## Transports

The general approach that we've been taking for transports it to use the `web-sys` bindings and code against those directly. This allows us to stay in Rust land without having to write any JS. Additionally, the acceptance criteria for any further transport is that they are integrated in the interop-test suite.

## Testing

We have additional tests that go beyond the interop suite in the `wasm-tests` directory: https://github.com/libp2p/rust-libp2p/tree/master/wasm-tests

---

```[tasklist]
- [x] https://github.com/libp2p/rust-libp2p/pull/4015
- [x] https://github.com/libp2p/rust-libp2p/pull/4248
- [ ] https://github.com/libp2p/rust-libp2p/issues/3611
- [ ] https://github.com/libp2p/rust-libp2p/issues/4389
- [ ] Remove `libp2p-wasm-ext` in the next breaking release
- [ ] https://github.com/libp2p/rust-libp2p/issues/4509
```

## Previous Implementations of WASM libp2p

- https://github.com/paritytech/smoldot
- https://github.com/iotaledger/identity.rs
- https://github.com/sireliah/wasm-p2p-experiment

mriise:
Pinging people that have previously worked on WASM here (that I know of) @wngr @tomaka @expenses 

Also pinging @mxinden

Again, feel free to make suggestions or LMK if I missed anything that should be included!

olanod:
A different kind of WASM support I'd like to bring to the table and something I'd like to experiment in the coming months(ideas and help will be welcomed) is the definition of a WASI interface that can be proposed later as `wasi-p2p` or something like that. Is a slightly different kind of WASM integration as the current direction is to be able to compile rust-libp2p to WASM expecting the host to provide calls that abstract the transport like using WebRTC with a set of custom host calls or perhaps the more standardized approach of targeting WASI that will have the sockets abstraction(could/will be polyfilled in the browser).

The proposal I mention is not so much for the browser support or environments that usually run a single bigger WASM blob where libp2p is included, my use case is embedded hardware where I want to allow a very constrained device to run a lightweight WASM interpreter like [WASM3](https://github.com/wasm3/wasm3) and run many  plugins that need to be as lightweight as possible, in systems where 1Mb is a lot and all there is I want to squeeze as many of those plugins as possible.  
[...]

PR #3115, opened 2022-11-13
feat(swarm): Add `wasm-bindgen` executor

umgefahren:
# Description

Add a `wasm-bindgen` executor to perform concurrent connections on browser or node background tasks.

Result of: https://github.com/libp2p/rust-libp2p/pull/3097#issuecomment-1310923167

## Open Questions

None

## Change checklist


swarm/src/executor.rs
```diff
@@ -0,0 +1,59 @@
...
+}
+
+#[cfg(feature = "wasm-bindgen")]
+#[derive(Default, Debug, Clone, Copy, Hash, PartialEq, Eq, PartialOrd, Ord)]
+pub(crate) struct WasmBindgenExecutor;
+
+#[cfg(feature = "wasm-bindgen")]
+impl Executor for WasmBindgenExecutor {
+    fn exec(&self, future: Pin<Box<dyn Future<Output = ()> + Send>>) {
+        wasm_bindgen_futures::spawn_local(future)
```
thomaseizinger, line 70: Does this end up being any different from using our local queue?

It would be cool if this could use the web-worker API for example but I am not sure how easy this is.

umgefahren: >  Does this end up being any different from using our local queue?

Good question. I think yes, although I'm not an expert here. As far as I understand it, spawn_local submits a task to the browser's native background task queue. Which might enable the browser to schedule more cleverly than we can do. We would probably have to try it or have an expert comment. Maybe @melekes?

> It would be cool if this could use the web-worker API for example but I am not sure how easy this is.

Yes, it would be, but it's really hard as I see it. There are some attempts to build something like this for Rust and although possible the main issue stays: When starting a Web Worker one has to give a path to a JS file which then creates the worker. Secondly sending to and from Web Workers is really difficult. I think it requires serialization. Maybe I'm wrong here. I would be happy to be corrected.

Of course, it would be better if WASM could use threads directly. That used to be possible but got really difficult because of the changes made to [SharedArrayBuffer](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/SharedArrayBuffer#security_requirements). Also: You can bring the Rust compiler to allow for multithreading but you have to compile the standard library yourself: [Parallel Raytracing Example](https://rustwasm.github.io/docs/wasm-bindgen/examples/raytrace.html).

Further reading: https://github.com/tokio-rs/tokio/issues/4827

thomaseizinger: > > Does this end up being any different from using our local queue?
> 
> Good question. I think yes, although I'm not an expert here. As far as I understand it, spawn_local submits a task to the browser's native background task queue. Which might enable the browser to schedule more cleverly than we can do. We would probably have to try it or have an expert comment. Maybe @melekes?
>
> 
> > It would be cool if this could use the web-worker API for example but I am not sure how easy this is.
> 
> Yes, it would be, but it's really hard as I see it. There are some attempts to build something like this for Rust and although possible the main issue stays: When starting a Web Worker one has to give a path to a JS file which then creates the worker. Secondly sending to and from Web Workers is really difficult. I think it requires serialization. Maybe I'm wrong here. I would be happy to be corrected.
> 
> Of course, it would be better if WASM could use threads directly. That used to be possible but got really difficult because of the changes made to [SharedArrayBuffer](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/SharedArrayBuffer#security_requirements). Also: You can bring the Rust compiler to allow for multithreading but you have to compile the standard library yourself: [Parallel Raytracing Example](https://rustwasm.github.io/docs/wasm-bindgen/examples/raytrace.html).
> 
> Further reading: [tokio-rs/tokio#4827](https://github.com/tokio-rs/tokio/issues/4827)

Thanks for linking all this. I am happy to merge this executor.

mxinden (approved):
Thanks @umgefahren for the work here!

From my minimal WASM knowledge, this looks good to me.

@vincev curious what your thoughts are here given your example https://github.com/libp2p/rust-libp2p/issues/2617#issuecomment-1304813987.

vincev:
> Thanks @umgefahren for the work here!
> 
> From my minimal WASM knowledge, this looks good to me.
> 
> @vincev curious what your thoughts are here given your example [#2617 (comment)](https://github.com/libp2p/rust-libp2p/issues/2617#issuecomment-1304813987).

@mxinden looks nice it makes it easier to create the `Swarm` I will update the example to use `with_wasm_executor`.

thomaseizinger (approved):
Rustfmt is not happy, otherwise this LGTM!

swarm/src/executor.rs
```diff
@@ -58,3 +58,14 @@ impl Executor for AsyncStdExecutor {
         let _ = async_std::task::spawn(future);
     }
 }
+
+#[cfg(feature = "wasm-bindgen")]
+#[derive(Default, Debug, Clone, Copy, Hash, PartialEq, Eq, PartialOrd, Ord)]
+pub(crate) struct WasmBindgenExecutor;
```
thomaseizinger:
```suggestion
pub struct WasmBindgenExecutor;
```

thomaseizinger (approved):

mxinden:
After merging https://github.com/libp2p/rust-libp2p/pull/3083/ which touched the mergify configuration, Mergify is complaining that "The current Mergify configuration is invalid", see also https://github.com/libp2p/rust-libp2p/pull/3115/checks?check_run_id=9673900404.

> warning The [subscription](https://dashboard.mergify.com/github/libp2p/subscription) needs to be updated to enable this feature.

Unfortunately I don't have access to the libp2p Mergify subscription page, thus I am not sure what it entails to "update the subscribtion".

@galargh do you have access? Can you check what we would need to do to "update the subscription"?

//CC @thomaseizinger

thomaseizinger:
Oh my god :(

https://docs.mergify.com/actions/post_check/#enforcing-conventional-commits

post_check's are a premium feature! I completely overlooked that! Sorry about that.

I suggest we revert #3083 for now.

thomaseizinger:
> Unfortunately I don't have access to the libp2p Mergify subscription page

You do have access to the dashboard though, right?

mxinden:
> > Unfortunately I don't have access to the libp2p Mergify subscription page
> 
> You do have access to the dashboard though, right?

I do. Is there some way that is helpful?

mxinden:
> I suggest we revert https://github.com/libp2p/rust-libp2p/pull/3083 for now.

:+1: done via https://github.com/libp2p/rust-libp2p/pull/3165.

thomaseizinger:
> > > Unfortunately I don't have access to the libp2p Mergify subscription page
> > 
> > You do have access to the dashboard though, right?
> 
> I do. Is there some way that is helpful?

Only to inspect the current merge queue. I just wanted to make sure this isn't tied to my account only for some reason.

Issue #2831, opened 2022-08-20
misc/multistream-select: Treat protocol as `String` and not `[u8]`


mxinden:
## Description

Today we treat protocols negotiated via `multistream-select` as a `[u8]`. Instead we should treat it as a `String`.

## Motivation

- I am not aware of a project using non-string protocol identifiers.
- The [specification treats them as strings](https://github.com/libp2p/specs/blob/master/connections/README.md#protocol-negotiation).
- `libp2p-identify` treats protocols as strings. See [spec](https://github.com/libp2p/specs/tree/master/identify#the-identify-message).
- The mismatch is unnecessary.
- See this confusion in the wild https://github.com/libp2p/rust-libp2p/pull/2798#issuecomment-1209026177
- The [Go implementation](https://github.com/multiformats/go-multistream/blob/46ef6cb8dd97068d755228615c74fbd3527fc2d6/client.go#L34) treats protocols as strings. 

## Current Implementation

https://github.com/libp2p/rust-libp2p/blob/3d3666e1a69c82afa9c966d668c32aae36c87599/misc/multistream-select/src/protocol.rs#L69-L77

## Are you planning to do it yourself in a pull request?

No

thomaseizinger:
Should we remove the `ProtocolName` abstraction as part of this or just change it to `&str` for now?

mxinden:
For the sake of type safety, what do you think of a _new type_ wrapping a `String` or `&str`?

efarg:
I'd like to work on this.

```
impl AsRef<str> for Protocol {
    fn as_ref(&self) -> &str {
	std::str::from_utf8(&self.0.as_ref()).unwrap()
    }
}
```

Are we thinking something along these line? I am using `&str` here, because, `String` requires new memory. With `AsRef<str>` in place, we can replace `AsRef<[u8]>` with `AsRef<str>` in `dialer_select.rs` and `listener_select.rs`, right?

But doing so, rippled its effect into places where dialers and listeners are used, as in `core/src/upgrade` where NameWrap, UpgradeInfo, ProtocolName want AsRef<str>. When I made changes to accommodate this, more places depending on the mentioned three started emitting errors. This keeps going on...

thomaseizinger:
It is a fairly invasive change yes. Happy to mentor you along if you want to pick it up!

See some of the communication in https://github.com/libp2p/rust-libp2p/pull/2798.

[...]

PR #3128, opened 2022-11-16
feat(core): Log protocol upgrades on trace

rkuhn:
## Description

Previously, every inbound or outbound upgrade generated a log at `debug` level, without information about the upgrade.
This commit changes it such that successful upgrades are logged at `trace` level (due to ubiquitous use of OneShot handlers) and that the negotiated protocol name is included in the message.

## Change checklist


core/src/upgrade.rs
```diff
@@ -134,6 +135,26 @@ impl<T: AsRef<[u8]>> ProtocolName for T {
...
+        for byte in self.0.protocol_name() {
+            if (b' '..=b'~').contains(byte) {
+                f.write_char(char::from(*byte))?;
+            } else {
+                write!(f, "<{:02X}>", byte)?;
+            }
+        }
+        Ok(())
+    }
+}
```
thomaseizinger:
```suggestion
impl<N: ProtocolName> fmt::Display for DisplayProtocolName<N> {
    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
        for byte in self.0.protocol_name() {
            if (b' '..=b'~').contains(byte) {
                f.write_char(char::from(*byte))?;
            } else {
                write!(f, "<{:02X}>", byte)?;
            }
        }
        Ok(())
    }
}
```

I typically prefer this way of referring to the types because you only end up needing `use std::fmt;`

core/src/upgrade.rs
```diff
@@ -134,6 +135,26 @@ impl<T: AsRef<[u8]>> ProtocolName for T {
...
+
+#[test]
+fn protocol_name_display() {
+    assert_eq!(DisplayProtocolName("hello").to_string(), "hello");
+    assert_eq!(DisplayProtocolName("hellö/").to_string(), "hell<C3><B6>/");
+}
```
thomaseizinger: Can we put this with the other tests at the bottom please in a `mod tests`?

core/src/upgrade/apply.rs
```diff
@@ -137,22 +141,30 @@ where
...
                     match Future::poll(Pin::new(&mut future), cx) {
                         Poll::Pending => {
-                            self.inner = InboundUpgradeApplyState::Upgrade { future };
+                            self.inner = InboundUpgradeApplyState::Upgrade { future, name };
                             return Poll::Pending;
                         }
                         Poll::Ready(Ok(x)) => {
-                            debug!("Successfully applied negotiated protocol");
+                            log::trace!(
+                                "Successfully applied negotiated inbound protocol {}",
```
thomaseizinger:
```suggestion
                                "Upgraded inbound stream to {}",
```

Perhaps we can shorted this a bit?

core/src/upgrade/apply.rs
```diff
@@ -137,22 +141,30 @@ where
...
+                                DisplayProtocolName(name)
+                            );
                             return Poll::Ready(Ok(x));
                         }
                         Poll::Ready(Err(e)) => {
-                            debug!("Failed to apply negotiated protocol");
+                            debug!(
+                                "Failed to apply negotiated inbound protocol {}",
```
thomaseizinger:
```suggestion
                                "Failed to upgrade inbound stream to {}",
```

thomaseizinger (commented):
Thanks!

A few minor suggestions.

core/src/upgrade.rs
rkuhn: okay, will do!

core/src/upgrade.rs
rkuhn: ah, I looked at some `tests.rs` but found it irrelevant, so placed the test next to the thing; recently I began doubting the dogma of “`mod tests` at the bottom” for such things because they serve to clearly document the intended behaviour right where people read the code

thomaseizinger: I don't mind much where the tests are but I want them to be feature-gated on `cfg(test)` so we don't compile them into binaries of our users.

rkuhn: ah, good point: we have a deal :-)

rkuhn: Wow, that was interesting: I checked what was needed, and indeed `#[test]` alone already hides the function in normal builds. This is the expansion:
```rust
    extern crate test;
    #[cfg(test)]
    #[rustc_test_marker = "upgrade::protocol_name_display"]
    pub const protocol_name_display: test::TestDescAndFn = test::TestDescAndFn {
        desc: test::TestDesc {
            name: test::StaticTestName("upgrade::protocol_name_display"),
            ignore: false,
            ignore_message: ::core::option::Option::None,
            compile_fail: false,
            no_run: false,
            should_panic: test::ShouldPanic::No,
            test_type: test::TestType::UnitTest,
        },
        testfn: test::StaticTestFn(|| test::assert_test_result(protocol_name_display())),
    };
    fn protocol_name_display() {
        ...
    }
```
So thanks to our exchange I learnt that `#[cfg(test)]` is only needed for hiding test-related support functions and infrastructure.
(EDIT: `fn protocol_name_display` is elided from the .rlib as dead code)

thomaseizinger: Are you sure `extern crate test` is not being dragged in?

rkuhn: Hmm, after lunch I followed this rabbit hole for a little while longer and compared the liblibp2p_core.rlib for three cases:

1. without the test (commented out)
2. test with only the `#[test]` attribute (120 bytes size increase, caused by a 16 byte increase and other changes in `lib.rmeta`)
3. with additional `#[cfg(test)]` attribute (106 bytes size increase, caused by a 104 byte increase in `lib.rmeta`)

As the metadata format is somewhat obscure, I decided to stop here. To my knowledge, `extern crate test` only brings a declared dependency explicitly into scope, which has been unnecessary since the 2015 edition (IIRC). Judging by the absence of the string `protocol_name_display` in the rlib, I’d leave the code as is, unless you (or others) have a strong opinion.

thomaseizinger: I am leaning towards adding a `mod tests` block:

1. Based on your investigation (thanks for doing that!), it seems to reduce code size.
2. It is consistent across the codebase.

I don't have a strong opinion on where the `mod tests` block sits although we tend to put it at the bottom of the file.

The argument that it should be next to the function is definitely valid, however I think the more common case is that you want to read the production code, not the test code.

If you want to actually co-locate it, you could make it a rustdoc test :)

rkuhn: That is a brilliant idea, in particular, because it found a usability bug (no public constructor).

rkuhn:
The failures in earlier commits are related to semver, which doesn’t seem to be caused by this PR, so I consider this ready now.

core/src/upgrade.rs
```diff
@@ -134,6 +135,28 @@ impl<T: AsRef<[u8]>> ProtocolName for T {
...
+/// ```
+pub struct DisplayProtocolName<N>(pub N);
+impl<N: ProtocolName> fmt::Display for DisplayProtocolName<N> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        use fmt::Write;
+        for byte in self.0.protocol_name() {
+            if (b' '..=b'~').contains(byte) {
+                f.write_char(char::from(*byte))?;
+            } else {
+                write!(f, "<{:02X}>", byte)?;
```
thomaseizinger:
```suggestion
                write!(f, "0x{:02X}", byte)?;
```

Thoughts?

core/src/upgrade.rs
```diff
@@ -134,6 +135,28 @@ impl<T: AsRef<[u8]>> ProtocolName for T {
     }
 }
 
+/// Wrapper for printing a [`ProtocolName`] that is expected to be mostly ASCII
+///
+/// ```rust
+/// # use libp2p_core::upgrade::DisplayProtocolName;
+/// assert_eq!(DisplayProtocolName("hello").to_string(), "hello");
```
thomaseizinger:
```suggestion
/// assert_eq!(DisplayProtocolName(b"hello").to_string(), "hello");
```

The tests should probably use byte-strings. Plus a protocol-like string would be better IMO, something like `/tls/1.0.0`.

core/src/upgrade.rs
```diff
@@ -134,6 +135,28 @@ impl<T: AsRef<[u8]>> ProtocolName for T {
...
+/// assert_eq!(DisplayProtocolName("hellö/").to_string(), "hell<C3><B6>/");
+/// ```
+pub struct DisplayProtocolName<N>(pub N);
```
mxinden:
```suggestion
pub struct DisplayProtocolName<N>(pub N);

```

core/src/upgrade/apply.rs
```diff
@@ -105,6 +108,7 @@ where
     },
     Upgrade {
         future: Pin<Box<U::Future>>,
+        name: SmallVec<[u8; 32]>,
```
mxinden, line 110: A stream upgrade involves various I/O operations on many layers. I don't think a heap allocated 32 byte vector have a significant performance impact.

Thus, unless the performance improvement is proven through a benchmark, I suggest allocating on the heap here.

```suggestion
        name: Vec<u8>,
```

core/src/upgrade.rs
mxinden: I am fine introducing this layer. My preference though would be to fix this the right way, see https://github.com/libp2p/rust-libp2p/issues/2831.

mxinden (commented):
Oh wonderful. Just 3 days ago I was debugging an issue and wished I had this patch in place. Thanks @rkuhn for the improvement.

core/src/upgrade.rs
rkuhn: Well, “layer” is too big a word — it is a newtype to tweak a trait implementation. Once protocol names are strings, it should go, of course.

core/src/upgrade/apply.rs
rkuhn: Given that all protocol names I’m aware of are quite small, I see no reason to unconditionally allocate on the heap — why should that be the default that needs no reason? I could reduce the inline capacity to keep `size_of` equal to that of a `Vec<u8>`, if you want.

core/src/upgrade.rs
rkuhn: We should probably use something that is established already. I went with what I know from `less`, we could also use `\xAB` like many programming languages use. Either way, making the output unambiguously parseable back to the bytes is a different story; I don’t think that would be appreciated enough to be worth it.

core/src/upgrade/apply.rs
thomaseizinger: I am with @rkuhn here. I think it is good to avoid allocations for the majority of the cases. I don't see the use of a different type as too much complexity.

core/src/upgrade.rs
thomaseizinger: I am fine to keep it as is :)

core/src/upgrade/apply.rs
mxinden: > I see no reason to unconditionally allocate on the heap — why should that be the default that needs no reason?

Reason for me to not use `SmallVec` by default is that it adds noise, additional compile time and increases the attack vector in our dependency supply chain. Though the last one might be a lost fight in the first place.

I wrote a benchmark to try to replicate a performance benefit using `SmallVec` here. I have not been able thus far. Please take a look at https://github.com/Actyx/rust-libp2p/pull/95. Maybe I can nerd-snipe either of you into proving me wrong @rkuhn and @thomaseizinger.

I am sorry in case this comes across as nit-picky. I am not trying to start an argument. I am genuinely interested in performance optimizations in rust-libp2p. I am curious in which cases we could optimize through stack allocation. I don't think this is one of those cases.

On a more general note, again out of curiosity, is either of you aware of a case where heap allocation **is** faster than stack allocation? I can not come up with a real world case where the heap would be faster than the stack.

core/src/upgrade.rs
mxinden: Why not use [`String::from_utf8_lossy`](https://doc.rust-lang.org/std/string/struct.String.html#method.from_utf8_lossy)?

core/src/upgrade.rs
mxinden: :+1: let's keep as is.

core/src/upgrade.rs
rkuhn: My intention was to convey the contents without losses, since this is a debugging tool.

core/src/upgrade.rs
rkuhn: It doesn’t really matter which bytes we use; I could make the test exhaustive by just using all 256 possible values.

core/src/upgrade/apply.rs
rkuhn: To me this is foremost a question of using the most appropriate data structure for the job. SmallVec was already in the dependencies, and it is better suited than Vec due to usage characteristics, so I used it.

---

Regarding the benchmark: memory management cannot be benchmarked in this fashion. In general, microbenchmarks like yours are very rarely meaningful and even then they require careful interpretation of the result. The reason is that most of the time you are looking at a huge machine doing a tiny task, with all sorts of optimisations kicking in. And when you run a relevant task instead, different optimisations will completely change the picture. It is quite obvious that a stack allocation takes less than one CPU cycle and that a heap allocation can in general not be this fast. If your microbenchmark finds them to be equally fast, then your heap usage allows the `malloc` implementation to optimise down to a single pointer subtraction — which is a very likely thread-local optimisation to make for `malloc()`.

But all of this theory is irrelevant in my opinion because we’re not optimising for the last percent of performance (as far as I’m aware we’re nowhere close to getting there at this point). So I just propose usage of the most appropriate data structure at hand.

core/src/upgrade.rs
thomaseizinger: It doesn't but I think it is good to use test data that is reasonably close to the real world.

core/src/upgrade/apply.rs
thomaseizinger: I think it is reasonable to trust in CS knowledge like "stack allocations are faster than heap", even if you can't produce a micro benchmark that proves it.

Soon, this will be replaced by `Cow` so I am happy either way. If @mxinden feels strongly, let's go with `Vec`.

rkuhn: BTW: this discussion uses “heap” and “stack”, but “stack” doesn’t actually enter the picture in the code in question since we’re talking about a memory slot in a generator, which will end up spawned onto a thread pool, so allocated on the heap in any case.

thomaseizinger: Right, what would be the right terminology then (out of curiosity?). "local variable"?

rkuhn: That sounds appropriate to me — it’ll end up being stored in a compiler-generated `enum` variant if it lives across an `.await` point.

core/src/upgrade.rs
```diff
@@ -134,6 +135,36 @@ impl<T: AsRef<[u8]>> ProtocolName for T {
...
+/// assert_eq!(DisplayProtocolName("/hellö/").to_string(), "/hell<C3><B6>/");
+/// # assert_eq!(
+/// #     DisplayProtocolName((0u8..=255).collect::<Vec<_>>()).to_string(),
+/// #     (0..32).map(|c| format!("<{:02X}>", c))
+/// #         .chain((32..127).map(|c| format!("{}", char::from_u32(c).unwrap())))
+/// #         .chain((127..256).map(|c| format!("<{:02X}>", c)))
+/// #         .collect::<String>()
+/// # );
+/// ```
+pub struct DisplayProtocolName<N>(pub N);
```
thomaseizinger: It would be nice if we could make this `pub(crate)`, otherwise it will be a breaking change to remove it later, despite it only being used internally.

However, I think this means we wouldn't be able to use the rustdoc test anymore.

Do you mind going back to a normal test?

thomaseizinger: For that matter, it should probably live in `upgrade/apply.rs`.

thomaseizinger (commented):
One more thought! I'll let @mxinden decide regarding `Vec` vs `SmallVec`.

core/src/upgrade.rs
rkuhn: sure thing!

thomaseizinger (dismissed):
LGTM, thanks!

mxinden (dismissed):
Thanks @rkuhn!

core/src/upgrade/apply.rs
mxinden: Thanks for expanding here. I think the points above are valid. That said, I haven't yet formed a good guiding principle for rust-libp2p.

As said above, the concrete instance is not relevant. Let's get this merged. Looking forward to the easier debugging of protocol upgrades on streams.

mxinden (approved):

PR #3132, opened 2022-11-17
docs(swarm): Reword `Swarm` top-level doc string

mikayelgr:


thomaseizinger (approved):
Slightly better IMO, thanks!

PR #3090, opened 2022-11-07
.github/workflows: Refactor CI jobs

thomaseizinger:
## Description

We refactor our continuous integration workflow with the following goals in mind:

- Run as few jobs as possible
- Have the jobs finish as fast as possible
- Have the jobs redo as little work as possible

There are only so many jobs that GitHub Actions will run in parallel.
Thus, it makes sense to not create massive matrices but instead group
things together meaningfully.

The new `test` job will:

- Run once for each crate
- Ensure that the crate compiles on its specified MSRV
- Ensure that the tests pass
- Ensure that there are no semver violations

This is an improvement to before because we are running all of these
in parallel which speeds up execution and highlights more errors at
once. Previously, tests run later in the pipeline would not get run
at all until you make sure the "first" one passes.

We also previously did not verify the MSRV of each crate, making the
setting in the `Cargo.toml` rather pointless.

The new `cross` job supersedes the existing `wasm` job.

This is an improvement because we now also compile the crate for
windows and MacOS. Something that wasn't checked before.
We assume that checking MSRV and the tests under Linux is good enough.
Hence, this job only checks for compile-errors.

The new `feature_matrix` ensures we compile correctly with certain feature combinations.

`libp2p` exposes a fair few feature-flags. Some of the combinations
are worth checking independently. For the moment, this concerns only
the executor related transports together with the executor flags but
this list can easily be extended.

The new `clippy` job runs for `stable` and `beta` rust.

Clippy gets continuously extended with new lints. Up until now, we would only
learn about those as soon as a new version of Rust is released and CI would
run the new lints. This leads to unrelated failures in CI. Running clippy on with `beta`
Rust gives us a heads-up of 6 weeks before these lints land on stable.

Fixes #2951.

## Links to any relevant issues

Depends-On: https://github.com/libp2p/rust-libp2p/pull/3131

## Open Questions

We could merge `cross` and `feature_matrix` together by adding another matrix variable that is either set to `--all-features` or `--features="mdns ..."`. Should we?

## Change checklist

- [ ] I have performed a self-review of my own code
- [ ] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] A changelog entry has been made in the appropriate crates

PR #3142, opened 2022-11-18
fix(webrtc): Don't emit addresses from other interfaces

thomaseizinger:
## Description

Previously, we would always run `IfWatcher`, even if we were only listening on a specific interface. This patch fixes this behaviour and aligns it with how `libp2p-quic` operates.

## Notes

cc @melekes 

## Change checklist

### DIFF
--- a/transports/webrtc/src/tokio/transport.rs
+++ b/transports/webrtc/src/tokio/transport.rs
@@ -19,7 +19,7 @@
 // DEALINGS IN THE SOFTWARE.
 
 use futures::{future::BoxFuture, prelude::*, ready, stream::SelectAll, stream::Stream};
-use if_watch::{IfEvent, IfWatcher};
+use if_watch::{tokio::IfWatcher, IfEvent};
 use libp2p_core::{
     identity,
     multiaddr::{Multiaddr, Protocol},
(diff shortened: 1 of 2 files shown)


transports/webrtc/src/tokio/transport.rs
```diff
@@ -232,7 +246,12 @@ impl ListenStream {
     }
 
     fn poll_if_watcher(&mut self, cx: &mut Context<'_>) -> Poll<<Self as Stream>::Item> {
-        while let Poll::Ready(event) = self.if_watcher.poll_if_event(cx) {
+        let if_watcher = match self.if_watcher.as_mut() {
+            Some(w) => w,
+            None => return Poll::Pending,
+        };
```
rkuhn, line 252: once we move to Rust 1.65 this can become nicer :-)
```rust
let Some(if_watcher) = self.if_watcher.as_mut() else { return Poll::Pending };
```

rkuhn (approved):
nice catch and thanks for the upgrade to version 3!

transports/webrtc/src/tokio/transport.rs
thomaseizinger: Yep, looking forward to this, I just didn't want to bump it yet. At least not until we have #3090 where we will start to properly verify our MSRV.

melekes (approved):
👍

mxinden (approved):
Thanks @thomaseizinger!

Issue #3133, opened 2022-11-17
quic: Add support for `quic` codepoint , interpreted as QUIC version draft-29


elenaf9:
## Description

The QUIC transport currently only supports the `quic-v1` codepoint in multiaddresses. This codepoint was recently added to represent QUIC version 1 / RFC 9000, whereas the previous codepoint `quic` should be interpreted as QUIC version draft-29. See multiformats/multiaddr#145.

`quinn` supports the versions draft-29..34 and version 1. On the server side this means that our users could already advertise each `quic-v1` address also as `quic` address. 
For dialing however we have to set one version, right now we use the default v1. We can add support for `quic(-draft-29)` by checking the multiaddress on dials and set the current version in the `quinn_proto::ClientConfig`.

## Motivation

`go-libp2p` does not support `quic-v1` yet, although there is WIP to add support in the very near future: libp2p/go-libp2p#1841. However, even when support is added we won't be able to communicate via QUIC with any go-libp2p nodes that did not upgrade to the newest version.  cc @dignifiedquire.

## Requirements

1. Allow dialing `quic` multiaddresses
2. Use QUIC draft-29 version when dialing `quic` addresses, use version 1 for `quic-v1`
3. To be discussed: support `quic` codepoint in listening addresses 

## Open questions

go-libp2p is planning a transition period (6 months?) before removing support for `quic`.  @marten-seemann raised a valid point on whether it even makes sense for rust-libp2p to add support, if ideally eventually the `quic` codepoint dies out because all new nodes run on `v1`.

I am in favor of at least adding support for dialing `quic` addresses so that we can connect to old go-libp2p nodes. I don't see any drawbacks in that.
I would propose to not add support for `quic` listening addresses. However, since in practice we still support draft-29 as servers, users are still free to announce a `quic-v1` address also as `quic` if they need it. They can already do this right now anyway unless we explicitly remove support for draft-29 in our config.
Wdyt @marten-seemann @mxinden?

## Are you planning to do it yourself in a pull request?

Yes (if folks agree with this).

marten-seemann:
> They can already do this right now anyway unless we explicitly remove support for draft-29 in our config.

I must have missed that in code review, I would've expected that draft-29 is disabled by default.

> go-libp2p is planning a transition period (6 months?) before removing support for quic. @marten-seemann raised a valid point on whether it even makes sense for rust-libp2p to add support, if ideally eventually the quic codepoint dies out because all new nodes run on v1.

[...]

PR #3151, opened 2022-11-21
feat(quic): Add draft-29 support

elenaf9:
## Description

Add support for QUIC draft-29 / the `quic` codepoint.  This enables both dialing and listening on `quic` addresses.
The motivation for adding support is to allow users to connect to old go-libp2p nodes that don't support the `quic-v1` codepoint.
**Per default support is disabled** (cc @marten-seemann).

See discussion in #3133.

## Open Questions

We use the same codepoint when reporting our listening addresses as address had that the user called `listen_on` on. 
If users want to report listening addresses for both codepoints they either have to manually duplicate and map each listening address, or create two listeners. 
The alternative would be to directly report listening addresses for both codepoints if draft-29 is enabled. I don't have a strong opinion, but I am leaning towards the current implementation to avoid reporting draft-29 addresses unless it is explicitly wanted by the user.

## Change checklist


thomaseizinger (commented):
Not too invasive so this looks okay to add. We may want to document that this will be removed in the future.

elenaf9:
> Not too invasive so this looks okay to add. We may want to document that this will be removed in the future.

Another option would be to deprecate `Protocol::Quic` in https://github.com/multiformats/rust-multiaddr.

elenaf9:
Would love to get this into #3163 if it gets a review in time and there are no major change requests, but we don't have to block the release on it.

mxinden:
~~@elenaf9 this is a non-breaking change, right? Mind if we release this as `libp2p-quic` `v0.1.1-alpha` then?~~

Never mind. It adds a `pub` field to  a `pub` struct.

elenaf9:
> ~@elenaf9 this is a non-breaking change, right? Mind if we release this as `libp2p-quic` `v0.1.1-alpha` then?~
> 
> Never mind. It adds a `pub` field to a `pub` struct.

I am not 100% sure, but is it really a breaking change adding a pub field to a pub struct if that struct also has other private fields anyway (in our case `server_tls_config` and `client_tls_config`)? Deconstructions like `let quic::Config { handshake_timeout, .. }  = config` are non-exhaustive anyway due to inaccessible fields I think.

transports/quic/src/endpoint.rs
```diff
@@ -280,6 +295,8 @@ pub enum ToEndpoint {
     Dial {
         /// UDP address to connect to.
         addr: SocketAddr,
+        /// Whether to dial the remote on QUIC version draft-29.
+        is_draft_29: bool,
```
mxinden: Instead of using a `bool`, I would prefer a new enum, e.g.:

``` rust
enum QuicVersion {
  V1,
  Draft29,
}
```

What do you think? Not a strong opinion.

transports/quic/src/transport.rs
```diff
@@ -49,12 +49,23 @@ use std::{
...
 /// Implementation of the [`Transport`] trait for QUIC.
+///
+/// Per default only QUIC Version 1 (RFC 9000) is supported. In the [`Multiaddr`] this maps to
+/// [`libp2p_core::multiaddr::Protocol::QuicV1`].
+/// The [`libp2p_core::multiaddr::Protocol::Quic`] codepoint is interpreted as QUIC version
+/// draft-29 and only supported if [`Config::support_draft_29`] is set to `true`.
+/// Note that in that case servers support both version an all QUIC listening addresses.
+///
+/// Version draft-29 should only be used to connect to nodes from other libp2p implementations
+/// that do not support `QuicV1` yet. Support for it will be removed long-term.
```
mxinden, line 60: How about linking to https://github.com/multiformats/multiaddr/issues/145 here?

mxinden (approved):
This looks good to me. Two suggestions, none of them blocking.

mxinden:
> > ~@elenaf9 this is a non-breaking change, right? Mind if we release this as `libp2p-quic` `v0.1.1-alpha` then?~
> > Never mind. It adds a `pub` field to a `pub` struct.
> 
> I am not 100% sure, but is it really a breaking change adding a pub field to a pub struct if that struct also has other private fields anyway (in our case `server_tls_config` and `client_tls_config`)? Deconstructions like `let quic::Config { handshake_timeout, .. } = config` are non-exhaustive anyway due to inaccessible fields I think.

Good point. I suggest following whatever [cargo semver-checks](https://github.com/obi1kenobi/cargo-semver-checks) on our CI says.

As far as I can tell, this is only time critical for [hole punching month](https://github.com/dennis-tra/punchr) in December. I am fine with either of:

- Releasing this as a breaking change.
- Releasing this as a non-breaking change (see cargo semver-checks comment above).
- Using an unreleased version in the Rust punchr client for hole punching month.

transports/quic/src/endpoint.rs
elenaf9: :+1: Good idea. added in 1fc87da14029b3fc338d8a717b97854924cd90fc

transports/quic/src/transport.rs
elenaf9: Done in fa39fc651ff49f1a173ced0c2c7cc05a57e27242.

transports/quic/src/transport.rs
```diff
@@ -49,12 +49,24 @@ use std::{
 };
 
 /// Implementation of the [`Transport`] trait for QUIC.
+///
+/// Per default only QUIC Version 1 (RFC 9000) is supported. In the [`Multiaddr`] this maps to
```
thomaseizinger:
```suggestion
/// By default only QUIC Version 1 (RFC 9000) is supported. In the [`Multiaddr`] this maps to
```

Not a native speaker but "per" sounds odd to me.

transports/quic/tests/smoke.rs
```diff
@@ -183,6 +183,98 @@ fn concurrent_connections_and_streams_tokio() {
...
+            }
+        })
+        .collect();
+    let dial = b_transport.dial(d_quic_addr_mapped).unwrap();
+    let drive_transports = poll_fn::<(), _>(|cx| {
+        let _ = b_transport.poll_next_unpin(cx);
+        let _ = d_transport.poll_next_unpin(cx);
+        Poll::Pending
+    });
+    select! {
```
thomaseizinger, line 234: @mxinden converted me to prefer the `select` fn. Can we convert you too? :)

transports/quic/tests/smoke.rs
```diff
@@ -183,6 +183,98 @@ fn concurrent_connections_and_streams_tokio() {
...
+        .clone()
+        .into_iter()
+        .map(|p| {
+            if matches!(p, Protocol::QuicV1) {
+                Protocol::Quic
+            } else {
+                p
+            }
+        })
+        .collect();
```
thomaseizinger: This snippet is repeated 3 times in the test. Can we make a little `swap_protocol` utility so these are one-liners?

transports/quic/src/transport.rs
```diff
@@ -486,6 +511,12 @@ impl<P: Provider> Drop for Listener<P> {
     }
 }
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum QuicVersion {
```
thomaseizinger:
```suggestion
pub enum ProtocolVersion {
```

transports/quic/src/transport.rs
```diff
@@ -486,6 +511,12 @@ impl<P: Provider> Drop for Listener<P> {
...
+    V1,
```
thomaseizinger:
```suggestion
    V1, // i.e. RFC9000
```

thomaseizinger (approved):
LGTM

Only a few nits!

thomaseizinger:
> > ~@elenaf9 this is a non-breaking change, right? Mind if we release this as `libp2p-quic` `v0.1.1-alpha` then?~
> > 
> > Never mind. It adds a `pub` field to a `pub` struct.
> 
> I am not 100% sure, but is it really a breaking change adding a pub field to a pub struct if that struct also has other private fields anyway (in our case `server_tls_config` and `client_tls_config`)? Deconstructions like `let quic::Config { handshake_timeout, .. }  = config` are non-exhaustive anyway due to inaccessible fields I think.

We are currently on an "alpha" version. Anything is allowed there, i.e. a user doesn't get any guarantees (which is probably why we shouldn't export it by default from the root `libp2p`).

transports/quic/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "libp2p-quic"
-version = "0.7.0-alpha"
+version = "0.7.1-alpha"
```
thomaseizinger: I think the appropriate bump would be:


```suggestion
version = "0.7.0-alpha.2"
```

`0.7.1-alpha` suggests that a `0.7.0` exists and we are pre-releasing a patch version (although the non-breaking aspect of said patch version is not guaranteed, hence the alpha).

We haven't even released `0.7.0` so we may as well push more alpha's out until we are happy with it.

thomaseizinger:
> > > ~@elenaf9 this is a non-breaking change, right? Mind if we release this as `libp2p-quic` `v0.1.1-alpha` then?~
> > > 
> > > Never mind. It adds a `pub` field to a `pub` struct.
> > 
> > I am not 100% sure, but is it really a breaking change adding a pub field to a pub struct if that struct also has other private fields anyway (in our case `server_tls_config` and `client_tls_config`)? Deconstructions like `let quic::Config { handshake_timeout, .. }  = config` are non-exhaustive anyway due to inaccessible fields I think.
> 
> We are currently on an "alpha" version. Anything is allowed there, i.e. a user doesn't get any guarantees (which is probably why we shouldn't export it by default from the root `libp2p`).

Note that the breaking/non-breaking aspect still matters for the re-exported interface in `libp2p`.

transports/quic/tests/smoke.rs
elenaf9: Can you point me to the discussion about the reasoning behind this? Right now I have a slight preference for the macro because of the readability.

transports/quic/tests/smoke.rs
elenaf9: Wdyt of 4cd494d?

transports/quic/src/transport.rs
elenaf9: Done in 8f05e09.

transports/quic/Cargo.toml
elenaf9: Ah yes you are right! Fixed in 2a83d3c. This also means that we don't need to bump the version in the `libp2p` crate.

transports/quic/tests/smoke.rs
thomaseizinger: Basically the rationale is:

- Plays better with IDEs
- Much easier to understand
- Clear type-safety in how to handle the various outcomes
- No fuse needed
- No magic syntax to learn

I don't know where exactly we had this discussion, so can't link you to it unfortunately.

transports/quic/Cargo.toml
thomaseizinger: > Ah yes you are right! Fixed in 2a83d3c. This also means that we don't need to bump the version in the `libp2p` crate.

I think you will have to, I don't think alpha versions update automatically. AFAIK, they are basically major version bumps but with an "unstable" co-notation.

thomaseizinger: I could be wrong though, please test it :)

elenaf9:
The semver check in our CI passes.
Also, the Rust API evolution RFC only defines adding a public field as major (= breaking) change if no private fields exist on the struct: https://github.com/rust-lang/rfcs/blob/master/text/1105-api-evolution.md#major-change-adding-a-public-field-when-no-private-field-exists.
Thus I think we can release this as a new alpha release `0.7.0-alpha.2`

mxinden (approved):
Changes and version handling looks good to me.

transports/quic/tests/smoke.rs
mxinden: While I very much believe in the cause @thomaseizinger, I suggest we delay further select-evangelizing towards @elenaf9 and others to future pull request. In other words, I don't think we need to block here. (Also gives us more time to print t-shirts and stickers.)

mxinden:
Interpreting @thomaseizinger approval as still valid.

mxinden:
`libp2p-dcutr` and `libp2p-relay` failures are unrelated and handled in https://github.com/libp2p/rust-libp2p/pull/2647#issuecomment-1335189994. Moving forward here.

Issue #3124, opened 2022-11-15
swarm: Remove `PollParameters`


thomaseizinger:
## Description

Remove the `PollParameters` trait.

- `local_peer_id` can be removed in favor of users having to pass it into their behaviour in the constructor.
- `external_addresses` can be built up from the `inject_new_external_addr` and `inject_expired_external_addr` callbacks.
- `listened_addresses` can be built up from the `inject_new_listen_addr` and `inject_expired_listen_addr` callbacks

The only function that does not have a replacement at the moment is `supported_protocols`.

Any ideas for how we could model that one differently?

As per usual, we should first deprecate the existing functions, give users a release or two to upgrade and then remove the deprecated code.

## Motivation

`rust-libp2p` is primarily designed around an async message-passing system. Accessing information synchronously as part of the `poll` function is not inline with this philosophy.

## Are you planning to do it yourself in a pull request?

Maybe.

mxinden:
I am in favor of removing `PollParameters`.

> The only function that does not have a replacement at the moment is `supported_protocols`.

I suggest similar to how we would communicate supported protocols of remote peers via https://github.com/libp2p/rust-libp2p/issues/2680, we can do the same for supported protocols of the local peer.

As an aside, the set of supported protocols can change at runtime. E.g. `libp2p-kad` with https://github.com/libp2p/rust-libp2p/issues/2032 will  only offer the Kademlia protocol for incoming streams when not behind a firewall or NAT. Thus the current mechanism is flawed.

thomaseizinger:
> I am in favor of removing `PollParameters`.
> 
> > The only function that does not have a replacement at the moment is `supported_protocols`.
> 
> I suggest similar to how we would communicate supported protocols of remote peers via #2680, we can do the same for supported protocols of the local peer.
> 
> As an aside, the set of supported protocols can change at runtime. E.g. `libp2p-kad` with #2032 will only offer the Kademlia protocol for incoming streams when not behind a firewall or NAT. Thus the current mechanism is flawed.

In theory, it could be that one connection of a node is behind firewall or NAT and the other one isn't. For example, lets say I am listening on TCP and WebRTC connections but only the WebRTC one is port-forwarded.

This thought experiment suggests that we'd have to tell our behaviour the supported protocols PER connection. In a way, this would fit in well with https://github.com/libp2p/rust-libp2p/pull/3099 where we only learn our supported protocols after we established the connection to another peer.

Thoughts?

[...]

PR #3153, opened 2022-11-22
refactor(swarm)!: deprecate `PollParameters` where possible

thomaseizinger:
## Description

This patch deprecates 3 out of 4 functions on `PollParameters`:

- `local_peer_id`
- `listened_addresses`
- `external_addresses`

The addresses can be obtained by inspecting the `FromSwarm` event. To make this easier, we introduce two utility structs in `libp2p-swarm`:

- `ExternalAddresses`
- `ListenAddresses`

A node's `PeerId` is always known to the caller, thus we can require them to pass it in.

Related: #3124.

## Notes

I am opening this PR to get an initial concept ACK.

It is still missing changelog entries and we might want to consider waiting for the introduction of `FromSwarm` to be merged before merging this. That would allow us to then extend `FromSwarm::NewExternalAddress` with the `AddressScore` that is tracked in the `Swarm` which makes `ExternaAddresses` actually feature-equal with the current `PollParameters::external_addresses`.

## Change checklist


protocols/kad/src/behaviour.rs
```diff
@@ -440,15 +443,17 @@ where
...
             queries: QueryPool::new(config.query_config),
             connected_peers: Default::default(),
             add_provider_job,
             put_record_job,
             record_ttl: config.record_ttl,
             provider_record_ttl: config.provider_record_ttl,
             connection_idle_timeout: config.connection_idle_timeout,
-            local_addrs: HashSet::new(),
             caching: config.caching,
+            external_addresses: ExternalAddresses::with_limit(MAX_LOCAL_EXTERNAL_ADDRS),
```
mxinden: Should we make `MAX_LOCAL_EXTERNAL_ADDRS` the default for `ExternalAddresses`? I don't see how `MAX_LOCAL_EXTERNAL_ADDRS` is specific to Kademlia.

swarm/src/behaviour/external_addresses.rs
```diff
@@ -0,0 +1,43 @@
+use crate::behaviour::{ExpiredExternalAddr, FromSwarm, NewExternalAddr};
+use crate::IntoConnectionHandler;
+use libp2p_core::Multiaddr;
+use std::collections::HashSet;
+
+/// Utility struct for tracking the external addresses of a [`Swarm`](crate::Swarm).
+#[derive(Debug, Default, Clone)]
+pub struct ExternalAddresses {
```
mxinden, line 13:
```suggestion
pub struct ExternalAddressTracker {
```

Not a strong opinion, though this would imply that one has to actively interact / pass-events with/to it.

mxinden (commented):
Two comments. Otherwise this looks great to me.

swarm/src/behaviour/listen_addresses.rs
```diff
@@ -0,0 +1,33 @@
...
+}
+
+impl ListenAddresses {
+    /// Returns an [`Iterator`] over all listen addresses.
+    pub fn iter(&self) -> impl ExactSizeIterator<Item = &Multiaddr> {
+        self.addresses.iter()
+    }
+
+    /// Feed a [`SwarmEvent`] to this struct.
+    pub fn on_event<THandler>(&mut self, event: &FromSwarm<THandler>)
```
jxs: Should we here also:
```suggestion
    pub fn on_swarm_event<THandler>(&mut self, event: &FromSwarm<THandler>)
```
to match `NetworkBehaviour::on_swarm_event` ? (and the same for `ExternalAddresses`)

swarm/src/behaviour/external_addresses.rs
thomaseizinger: I am not convinced that this is a better name. Can you elaborate what you don't like about the current one? :)

mxinden: The name `ExternalAddresses` does not imply to me as a user, that I need to pass the `FromSwarm` events into it. That said, I think `ExternalAddresses` is just fine.

mxinden (approved):
Solid work. Thanks!

swarm/src/behaviour/external_addresses.rs
thomaseizinger: I would argue that this is only something a user needs to learn once and quite obvious from the implementation.

Most of the time, this name should convey that it holds the external addresses and only every so often, you need to understand how it works so I optimised for the common bit :)

jxs (approved):
LGTM

PR #3156, opened 2022-11-22
feat(swarm): Add `deny_list` behaviour

thomaseizinger:
## Notes

To actually land this, we would obviously do it in a backwards-compatible way. This PR is just meant to serve as a showcase to see what it would look like to implement various capabilities on top of the new, generic connection management.

## Change checklist

### DIFF
--- a/swarm/src/behaviour/deny_list.rs
+++ b/swarm/src/behaviour/deny_list.rs
@@ -1,0 +1,60 @@
+use crate::behaviour::THandlerInEvent;
+use crate::{dummy, CloseConnection, NetworkBehaviour, NetworkBehaviourAction, PollParameters};
+use libp2p_core::{ConnectedPoint, PeerId};
+use std::collections::{HashSet, VecDeque};
+use std::task::{Context, Poll};
+
+#[derive(Default)]
+pub struct Behaviour {
+    list: HashSet<PeerId>,
+    to_disconnect: VecDeque<PeerId>,
+}
+
+impl Behaviour {
+    pub fn add_peer(&mut self, peer: PeerId) {
+        self.list.insert(peer);
+        self.to_disconnect.push_back(peer);
+    }
+
+    pub fn remove_peer(&mut self, peer: PeerId) {
+        self.list.remove(&peer);
+    }
+}
+
+#[derive(Debug, thiserror::Error)]
+#[error("peer is on a deny list")]
+pub struct Denied;
+
+impl NetworkBehaviour for Behaviour {
+    type ConnectionHandler = dummy::ConnectionHandler;
+    type ConnectionDenied = Denied;
+    type OutEvent = ();
+
+    fn new_handler(
+        &mut self,
+        peer: &PeerId,
+        _: &ConnectedPoint,
+    ) -> Result<Self::ConnectionHandler, Self::ConnectionDenied> {
+        if self.list.contains(peer) {
+            return Err(Denied);
+        }
+
+        Ok(dummy::ConnectionHandler)
+    }
+
+    fn poll(
+        &mut self,
+        _: &mut Context<'_>,
+        _: &mut impl PollParameters,
+    ) -> Poll<NetworkBehaviourAction<Self::OutEvent, THandlerInEvent<Self::ConnectionHandler>>>
+    {
+        if let Some(peer_id) = self.to_disconnect.pop_front() {
+            return Poll::Ready(NetworkBehaviourAction::CloseConnection {
+                peer_id,
+                connection: CloseConnection::All,
+            });
+        }
+
+        Poll::Pending
+    }
+}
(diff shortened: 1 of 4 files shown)


swarm/src/lib.rs
```diff
@@ -524,14 +521,6 @@ where
                         return Err(DialError::DialPeerConditionFalse(condition));
                     }
 
-                    // Check if peer is banned.
-                    if self.banned_peers.contains(&peer_id) {
-                        let error = DialError::Banned;
-                        #[allow(deprecated)]
-                        self.behaviour.inject_dial_failure(Some(peer_id), &error);
-                        return Err(error);
-                    }
```
thomaseizinger, line 533: We lose this early check here, i.e. we will initiate a dial and then immediately drop the connection instead of aborting the connection early. I guess that is the price for implementing this kind of functionality in a generic way.

swarm/src/behaviour/deny_list.rs
```diff
@@ -0,0 +1,60 @@
...
+#[derive(Default)]
+pub struct Behaviour {
+    list: HashSet<PeerId>,
+    to_disconnect: VecDeque<PeerId>,
+}
+
+impl Behaviour {
+    pub fn add_peer(&mut self, peer: PeerId) {
+        self.list.insert(peer);
+        self.to_disconnect.push_back(peer);
```
rkuhn, line 16: This should also lead to waking up the swarm. Perhaps using an async channel is the easiest and most readable solution? (instead of rebuilding such functionality with a `VecDeque` and an `Option<Waker>`)

rkuhn (commented):
Yes, looks quite clean to me!

swarm/src/behaviour/deny_list.rs
thomaseizinger: @mxinden was talking about building a `WakingVec` abstraction because we are coming across this very often actually.

thomaseizinger: I had a quick search on crates.io but couldn't find anything that does what we need. It might be possible that I used the wrong keywords!

thomaseizinger: I made a PoC here: https://github.com/libp2p/rust-libp2p/pull/3160

rkuhn: While that may have independent merit (not yet reviewed), I meant simply using [`async-channel`](https://docs.rs/async-channel) instead of the VecDeque.

thomaseizinger: > While that may have independent merit (not yet reviewed), I meant simply using [`async-channel`](https://docs.rs/async-channel) instead of the VecDeque.

I always find it confusing when the same data structure holds both ends of a channel.

But I agree with you that it would solve the problem.

mxinden (commented):
Wow, this is surprisingly simple. I am very much in favor of the direction this is going.

swarm/src/behaviour/deny_list.rs
```diff
@@ -0,0 +1,60 @@
...
+
+#[derive(Default)]
+pub struct Behaviour {
+    list: HashSet<PeerId>,
+    to_disconnect: VecDeque<PeerId>,
+}
+
+impl Behaviour {
+    pub fn add_peer(&mut self, peer: PeerId) {
+        self.list.insert(peer);
```
elenaf9, line 15:
```suggestion
pub struct Behaviour<T = ()> {
    list: HashMap<PeerId, T>,
    to_disconnect: VecDeque<PeerId>,
}

impl Behaviour<()> {
    pub fn add_peer(&mut self, peer: PeerId) {
        self.list.insert(peer, ());
        self.to_disconnect.push_back(peer);
    }
}

impl<T> Behaviour<T> {
    pub fn add_peer_because(&mut self, peer: PeerId, reason: T) {
        self.list.insert(peer, reason);
```

We could allow an optional `reason` parameter that we'd then also include in the `Denied` struct.  Might be useful if there are multiple reasons why a peer can be banned.
If we'd additionally expose what peers are currently banned (and for what reason) it would allow a more sophisticated management of banned peers. E.g. if a peer was only banned because an Application-Layer limit was reached.

But not sure if it's worth the extra complexity. Could wait until there is actually a use-case for this.

swarm/src/behaviour/deny_list.rs
```diff
@@ -0,0 +1,60 @@
+use crate::behaviour::THandlerInEvent;
+use crate::{dummy, CloseConnection, NetworkBehaviour, NetworkBehaviourAction, PollParameters};
+use libp2p_core::{ConnectedPoint, PeerId};
+use std::collections::{HashSet, VecDeque};
+use std::task::{Context, Poll};
+
+#[derive(Default)]
```
elenaf9, line 7:
```suggestion
#[derive(Default, Debug, Clone)]
```

thomaseizinger:
Closing for now as I am working on a new version of the connection management idea here: https://github.com/libp2p/rust-libp2p/pull/3254

PR #3159, opened 2022-11-22
refactor(request-response): revise public API to follow naming convention

jxs:
## Notes

Continues addressing #2217.
Per commit review is suggested :)

## Change checklist


protocols/request-response/src/lib.rs
```diff
@@ -61,7 +61,7 @@
 pub mod codec;
 pub mod handler;
 
-pub use codec::{ProtocolName, RequestResponseCodec};
```
jxs, line 64: https://github.com/rust-lang/rust/issues/41517 is not stable, so `RequestResponseCodec` to `Codec` rename is probably breaking.

protocols/request-response/src/handler.rs
```diff
@@ -191,9 +195,15 @@ where
...
+#[deprecated(
+    since = "0.23.0",
+    note = "Use re-exports that omit `RequestResponse` prefix, i.e. `libp2p::request_response::handler::HandlerEvent`"
+)]
+pub type RequestResponseHandlerEvent<TCodec> = HandlerEvent<TCodec>;
+
+/// The events emitted by the [`Handler`].
 #[doc(hidden)]
-pub enum RequestResponseHandlerEvent<TCodec>
+pub enum HandlerEvent<TCodec>
```
thomaseizinger: Would just `Event` be pushing it too far? :D

thomaseizinger: In case it isn't I think the `handler` module should be private so this `Event` can't clash with the other one.

protocols/request-response/src/handler.rs
```diff
@@ -54,8 +54,7 @@ use std::{
 )]
 pub type RequestResponseHandler<TCodec> = Handler<TCodec>;
 
-/// A connection handler of a `Behaviour` protocol.
-#[doc(hidden)]
+/// A connection handler of a [`Behaviour`](super::Behaviour) protocol.
```
thomaseizinger: This comment needs updating I'd say.

thomaseizinger (commented):
Nice work! Thanks :)

protocols/request-response/src/lib.rs
thomaseizinger: Damn. Will this work though?

1. Create a new trait `Codec`
1. Create a blanket impl of `Codec` for everything that is `RequestReponseCodec`
1. Deprecate `RequestResponseCodec`
1. Use `Codec` everywhere in `libp2p-request-response`

protocols/request-response/src/handler.rs
mxinden: Intuitively `libp2p::request_response::handler::Event` reads better than `libp2p::request_response::handler::HandlerEvent` and would be inline with the convention this pull request is enforcing.

mxinden (commented):
Other than the comments by Thomas above, this looks good to me.

Thanks @jxs.

protocols/request-response/src/handler.rs
jxs: updating the wording, ptal and see if it's what you were expecting Thomas.

protocols/request-response/src/handler.rs
jxs: yeah makes sense to me, thanks for the suggestion,
So, there is `request_response::Event` and `request_response::handler::Event` now. Can you expand on `Event`'s clashing Thomas?  
There will still be `request_response::RequestResponseHandlerEvent` which is deprecated and suggests `request_response::handler::Event`, if we make `handler` private how can we export both?

protocols/request-response/src/lib.rs
jxs: ehe nice! Updated, yeah it works at least to give a warning to users using `RequestResponseCodec` pointing how to update. But it will be still breaking cause `RequestResponse::new` expects `Codec` now

thomaseizinger: Damn. It would be nice to make it completely non-breaking.

protocols/request-response/CHANGELOG.md
```diff
@@ -12,6 +12,13 @@
 
 - Update `rust-version` to reflect the actual MSRV: 1.62.0. See [PR 3090].
 
+- Rename types as per [discussion 2174].
```
mxinden: We released `llibp2p-request-response` `v0.23.0` with `libp2p` `v0.50.0`.

If I understand correctly, this pull request does introduce a breaking change. Thus, would you mind bumping the version in the `libp2p-request-reponse` `Cargo.toml` and all its dependents and adding a changelog entry?

jxs: Thanks Max, updated

protocols/request-response/CHANGELOG.md
```diff
@@ -1,3 +1,15 @@
+# 0.24.0 [unreleased]
+
+- Rename types as per [discussion 2174].
+  `RequestResponse` has been renamed to `Behaviour`.
+  The `RequestResponse` prefix has been removed from various types like `RequestResponseEvent`.
+  Users should prefer importing the request_response protocol as a module (`use libp2p::request_response;`),
+  and refer to its types via `request_response::`. For example: `request_response::Behaviour` or `request_response::Event`.
+  See [PR 3159].
```
mxinden, line 8: :rocket:

protocols/request-response/src/codec.rs
```diff
@@ -25,8 +25,12 @@ use futures::prelude::*;
 use std::io;
 
 /// A `RequestResponseCodec` defines the request and response types
-/// for a [`RequestResponse`](crate::RequestResponse) protocol or
+/// for a [`Behaviour`](crate::Behaviour) protocol or
 /// protocol family and how they are encoded / decoded on an I/O stream.
+#[deprecated(
+    since = "0.23.0",
```
mxinden:
```suggestion
    since = "0.24.0",
```

protocols/request-response/src/handler.rs
```diff
@@ -48,11 +48,16 @@ use std::{
     time::Duration,
 };
 
-/// A connection handler of a `RequestResponse` protocol.
-#[doc(hidden)]
-pub struct RequestResponseHandler<TCodec>
+#[deprecated(
+    since = "0.23.0",
```
mxinden:
```suggestion
    since = "0.24.0",
```

protocols/request-response/src/handler.rs
```diff
@@ -191,11 +193,16 @@ where
     }
 }
 
-/// The events emitted by the [`RequestResponseHandler`].
-#[doc(hidden)]
-pub enum RequestResponseHandlerEvent<TCodec>
+#[deprecated(
+    since = "0.23.0",
```
mxinden:
```suggestion
    since = "0.24.0",
```

mxinden (commented):
This is missing a version bump in the root `Cargo.toml`, right?

protocols/request-response/src/handler.rs
jxs: Thanks Max, addressed.

protocols/request-response/src/handler.rs
jxs: Thanks Max, addressed.

protocols/request-response/src/handler.rs
thomaseizinger: Yeah if we make `handler` private, we can no longer export it. What I meant with clashing is that from the outside, there is really only one event that is interesting for users and it is the `NetworkBehaviour`'s `OutEvent`.

Given that we are already making this a breaking change, we might as well make `handler` private. In my opinion, no one should be depending on just the `ConnectionHandler`.

protocols/autonat/src/behaviour.rs
```diff
@@ -542,12 +542,12 @@ type Action = NetworkBehaviourAction<
     <Behaviour as NetworkBehaviour>::ConnectionHandler,
 >;
 
-// Trait implemented for `AsClient` as `AsServer` to handle events from the inner [`RequestResponse`] Protocol.
+// Trait implemented for `AsClient` as `AsServer` to handle events from the inner [`request_response::Behaviour`] Protocol.
```
thomaseizinger:
```suggestion
// Trait implemented for `AsClient` and `AsServer` to handle events from the inner [`request_response::Behaviour`] Protocol.
```

protocols/request-response/src/codec.rs
```diff
@@ -25,8 +25,12 @@ use futures::prelude::*;
 use std::io;
 
 /// A `RequestResponseCodec` defines the request and response types
-/// for a [`RequestResponse`](crate::RequestResponse) protocol or
+/// for a [`Behaviour`](crate::Behaviour) protocol or
```
thomaseizinger: A lot of these uses in the doc strings previously said `RequestResponse` protocol which no longer makes any sense now if we just verbatim replace `RequestResponse` with `Behaviour` :/

protocols/request-response/src/codec.rs
```diff
@@ -78,3 +82,58 @@ pub trait RequestResponseCodec {
     where
         T: AsyncWrite + Unpin + Send;
 }
+
+/// A `Codec` defines the request and response types
+/// for a [`Behaviour`](crate::Behaviour) protocol or
```
thomaseizinger: Same here.

protocols/request-response/src/codec.rs
```diff
@@ -78,3 +82,58 @@ pub trait RequestResponseCodec {
...
+/// protocol family and how they are encoded / decoded on an I/O stream.
+#[async_trait]
+pub trait Codec {
```
thomaseizinger, line 90: Don't we need a blanket impl of this for anything that implements `RequestResponseCodec` in order for it to be even worth it defining a new one?

protocols/request-response/src/handler.rs
```diff
@@ -225,58 +232,58 @@ where
...
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         match self {
-            RequestResponseHandlerEvent::Request {
+            Event::Request {
                 request_id,
                 request: _,
                 sender: _,
             } => f
-                .debug_struct("RequestResponseHandlerEvent::Request")
+                .debug_struct("HandlerEvent::Request")
```
thomaseizinger: This still says `HandlerEvent`.

protocols/request-response/src/handler.rs
```diff
@@ -309,7 +316,7 @@ where
         };
 
         // The handler waits for the request to come in. It then emits
-        // `RequestResponseHandlerEvent::Request` together with a
+        // `HandlerEvent::Request` together with a
```
thomaseizinger: Here too.

protocols/request-response/src/lib.rs
```diff
@@ -22,34 +22,34 @@
 //!
 //! ## General Usage
 //!
-//! [`RequestResponse`] is a `NetworkBehaviour` that implements a generic
+//! The [`Behaviour`] struct is a `NetworkBehaviour` that implements a generic
```
thomaseizinger:
```suggestion
//! The [`Behaviour`] struct is a [`NetworkBehaviour`] that implements a generic
```

protocols/request-response/src/lib.rs
```diff
@@ -93,30 +128,30 @@ pub enum RequestResponseMessage<TRequest, TResponse, TChannelResponse = TRespons
...
-        /// See [`RequestResponse::send_request`].
+        /// See [`Behaviour::send_request`].
         request_id: RequestId,
         /// The response message.
         response: TResponse,
     },
 }
 
-/// The events emitted by a [`RequestResponse`] protocol.
+/// The events emitted by a [`Behaviour`] protocol.
```
thomaseizinger: This reads rather odd.

protocols/request-response/src/lib.rs
```diff
@@ -838,30 +872,30 @@ where
...
                             peer,
                             request_id,
                             error: OutboundFailure::Timeout,
                         },
                     ));
             }
-            RequestResponseHandlerEvent::InboundTimeout(request_id) => {
-                // Note: `RequestResponseHandlerEvent::InboundTimeout` is emitted both for timing
+            handler::Event::InboundTimeout(request_id) => {
+                // Note: `HandlerEvent::InboundTimeout` is emitted both for timing
```
thomaseizinger:
```suggestion
                // Note: `handler::Event::InboundTimeout` is emitted both for timing
```

protocols/request-response/src/lib.rs
```diff
@@ -870,20 +904,20 @@ where
...
                             request_id,
                             error: OutboundFailure::UnsupportedProtocols,
                         },
                     ));
             }
-            RequestResponseHandlerEvent::InboundUnsupportedProtocols(request_id) => {
+            handler::Event::InboundUnsupportedProtocols(request_id) => {
                 // Note: No need to call `self.remove_pending_outbound_response`,
-                // `RequestResponseHandlerEvent::Request` was never emitted for this request and
+                // `HandlerEvent::Request` was never emitted for this request and
```
thomaseizinger:
```suggestion
                // `handler::Event::Request` was never emitted for this request and
```

thomaseizinger (commented):
Thanks for doing this!

A few minor comments :)

protocols/request-response/src/handler.rs
jxs: Thanks for the explanation Thomas. `Handler` and `Event` can't be private because `NetworkBehaviour::ConnectionBehaviour` associated type needs to be public, and by consequence `ConnectionHandler`'s  also

protocols/autonat/src/behaviour.rs
jxs: Thanks Thomas, updated.

protocols/request-response/src/codec.rs
jxs: do you think "for a [`request_response::Behaviour`](crate::Behaviour) protocol" makes more sense?

protocols/request-response/src/handler.rs
jxs: Thanks Thomas, addressed

protocols/request-response/src/handler.rs
jxs: Thanks, addressed.

protocols/request-response/src/codec.rs
jxs: yeah, helps in the transition as well, thanks Thomas,

protocols/request-response/src/lib.rs
jxs: Thanks, updated!

protocols/request-response/src/lib.rs
jxs: Thanks, updated!

protocols/request-response/src/lib.rs
jxs: Thanks, updated :)

protocols/request-response/src/codec.rs
thomaseizinger: Uhm, I think it would read better if we just not link to `Behaviour`!

protocols/request-response/src/codec.rs
```diff
@@ -78,3 +82,152 @@ pub trait RequestResponseCodec {
...
+        T: AsyncWrite + Unpin + Send,
+        T: 'async_trait,
+        'life0: 'async_trait,
+        'life1: 'async_trait,
+        'life2: 'async_trait,
+        Self: 'async_trait,
+    {
+        self.write_response(protocol, io, res)
+    }
+}
```
thomaseizinger: Why can't we use the `async_trait` macro for this?

jxs: we can yeah, but we need to upgrade `U`'s trait  to be `: RequestResponseCodec + Send` and `U::Protocol` to be`: Sync` due to how `#[async_trait]` expands.  
Don't know if it will break `RequestResponseCodec` impl's but updated Thomas.

thomaseizinger: Hmm okay. I'd expect most codecs to be ZST so this shouldn't matter really. Let's see if someone complains :)

protocols/request-response/src/lib.rs
jxs: yeah agreed, It's what you referred above,  verbatim replacing `RequestResponse` with `Behaviour` feels odd in a lot of cases. Tried to improve by adding the `request-response` context before. See if if you think it improves Thomas.

protocols/request-response/src/codec.rs
jxs: thanks, addressed.

protocols/request-response/src/codec.rs
jxs: ok done thanks :)

Cargo.toml
```diff
@@ -106,7 +106,7 @@ libp2p-plaintext = { version = "0.38.0", path = "transports/plaintext", optional
 libp2p-pnet = { version = "0.22.2", path = "transports/pnet", optional = true }
 libp2p-relay = { version = "0.14.0", path = "protocols/relay", optional = true }
 libp2p-rendezvous = { version = "0.11.0", path = "protocols/rendezvous", optional = true }
-libp2p-request-response = { version = "0.23.0", path = "protocols/request-response", optional = true }
+libp2p-request-response = { version = "0.24.0", path = "protocols/request-response", optional = true }
```
mxinden: Given that this is a breaking change for `libp2p`, this pull request needs to bump the `libp2p` version to `v0.51.0`. Can you please also add a changelog entry to the root `CHANGELOG.md` file. See previous entries for reference.

mxinden (commented):
Changes look good to me. No objections to merging other than the missing version change and changelog entry.

@thomaseizinger any remaining comments?

thomaseizinger:
> @thomaseizinger any remaining comments?

Nope, LGTM!

thomaseizinger (approved):

protocols/request-response/src/handler.rs
thomaseizinger: > Thanks for the explanation Thomas. `Handler` and `Event` can't be private because `NetworkBehaviour::ConnectionBehaviour` associated type needs to be public, and by consequence `ConnectionHandler`'s also

Hmm that is news to me! We have a lot of protocols where the `ConnectionHandler` implementation is completely private to the crate. For example `ping`: https://github.com/libp2p/rust-libp2p/blob/5755942d7742f9d0432c346318e25da504b65370/protocols/ping/src/lib.rs#L45

thomaseizinger:
> > @thomaseizinger any remaining comments?
> 
> Nope, LGTM!

Actually, I do want to finish one discussion: https://github.com/libp2p/rust-libp2p/pull/3159#discussion_r1041759415

protocols/request-response/src/lib.rs
thomaseizinger: > ehe nice! Updated, yeah it works at least to give a warning to users using `RequestResponseCodec` pointing how to update. But it will be still breaking cause `RequestResponse::new` expects `Codec` now

Is this actually breaking, now that we have the blanket impl where we implement `Codec` for everything that implements `RequestResponseCodec`?

protocols/request-response/src/handler.rs
thomaseizinger: I just tested it locally and things compile if I make the `handler` module private. However, I'd suggest that we don't do this for now. I think together with https://github.com/libp2p/rust-libp2p/pull/3159#discussion_r1041760501, we can make this completely non-breaking?

jxs: Thanks for digging this deeper Thomas! Did you understand what is it that with `request-response` that was making the compiler complain?

protocols/request-response/src/lib.rs
jxs: yeah good point Thomas, it probably isn't yeah AFAICT. Should we then make this a patch version?

thomaseizinger: > yeah good point Thomas, it probably isn't yeah AFAICT. Should we then make this a patch version?

I'd much prefer so, yes!

protocols/request-response/src/handler.rs
thomaseizinger: I think the issue is fairly subtle:

If you have a public type that implements a trait, the compiler enforces that the associated types also need to be public. It does however not check whether the type is within a public or private module in regards to the crate.

In our usecase, this means we can have a `pub struct Handler` that implements `ConnectionHandler` but define it within a crate-private module `handler`. This may or may not be a bug in Rust but I think one can argue that it makes sense:

The API contract promised to the user is that `Behaviour` implements `NetworkBehaviour`. `NetworkBehaviour` enforces that `NetworkBehaviour::ConnectionHandler` implements `IntoConnectionHandler`. The user can refer to this type as `<Behaviour::ConnectionHandler as IntoConnectionHandler>` and call all public APIs on that. They don't need to know the concrete type that implements the functionality. This way, only the minimal promised API is actually exposed.

protocols/request-response/src/lib.rs
jxs: Ok, updated!

thomaseizinger (approved):
Lovely, thanks so much!

I am thrilled that this is backwards-compatible now :)

mxinden (approved):

jxs:
fixed the merge conflicts, and updated the deprecation message to `0.24`

thomaseizinger (approved):

PR #3160, opened 2022-11-23
feat: Add `async-bounded-queue` to `misc/`

thomaseizinger:
## Notes

This is a PoC for what @mxinden termed "WakingVec" in one of our recent community calls. Feedback welcome.

## Change checklist

### DIFF
--- a/misc/async-bounded-queue/Cargo.toml
+++ b/misc/async-bounded-queue/Cargo.toml
@@ -1,0 +1,8 @@
+[package]
+name = "async-bounded-queue"
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
--- a/misc/async-bounded-queue/src/lib.rs
+++ b/misc/async-bounded-queue/src/lib.rs
@@ -1,0 +1,88 @@
+use std::collections::VecDeque;
+use std::task::{Context, Poll, Waker};
+
+/// A bounded queue for use in asynchronous contexts.
+pub struct Queue<T> {
+    storage: VecDeque<T>,
+
+    queue_full_waker: Option<Waker>,
+    queue_empty_waker: Option<Waker>,
+}
+
+impl<T> Queue<T> {
+    /// Construct a queue with the given capacity.
+    pub fn new(capacity: usize) -> Self {
+        Self {
+            storage: VecDeque::with_capacity(capacity),
+            queue_full_waker: None,
+            queue_empty_waker: None,
+        }
+    }
+
+    /// Check if we can push an item to the queue.
+    ///
+    /// In case the queue is full, [`Poll::Pending`] is returned and a waker is registered that will call the current task once there is a slot in the queue.
+    pub fn poll_push_ready(&mut self, cx: &mut Context<'_>) -> Poll<()> {
+        if !self.is_full() {
+            return Poll::Ready(());
+        }
+
+        self.queue_full_waker = Some(cx.waker().clone());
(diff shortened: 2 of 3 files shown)


misc/async-bounded-queue/Cargo.toml
```diff
@@ -0,0 +1,8 @@
+[package]
+name = "async-bounded-queue"
+version = "0.1.0"
+edition = "2021"
```
mxinden, line 4: Missing required fields.

misc/async-bounded-queue/Cargo.toml
```diff
@@ -0,0 +1,8 @@
...
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
```
mxinden, line 8: This is very nice!

misc/async-bounded-queue/src/lib.rs
```diff
@@ -0,0 +1,88 @@
...
+
+    queue_full_waker: Option<Waker>,
+    queue_empty_waker: Option<Waker>,
+}
+
+impl<T> Queue<T> {
+    /// Construct a queue with the given capacity.
+    pub fn new(capacity: usize) -> Self {
+        Self {
+            storage: VecDeque::with_capacity(capacity),
```
mxinden, line 16: Just a thought: Say that I am using `Queue` with a large non-boxed `T` type. Say that I want a bound of some large number, e.g. 1_000. Say that in the usual case, I only ever have 10 elements here. In such case, pre-allocating the whole vector would be a waste of memory.

Does the above make sense? Do we even need to care about this edge-case? I am happy to delay this decision.

misc/async-bounded-queue/src/lib.rs
```diff
@@ -0,0 +1,88 @@
...
+            queue_full_waker: None,
+            queue_empty_waker: None,
+        }
+    }
+
+    /// Check if we can push an item to the queue.
+    ///
+    /// In case the queue is full, [`Poll::Pending`] is returned and a waker is registered that will call the current task once there is a slot in the queue.
```
mxinden, line 24: Would you mind wrapping this line?

mxinden (commented):
A couple of minor comments. This looks great to me. 

I suggest holding off merging here until we have a concrete use-case. Would you agree?

misc/async-bounded-queue/src/lib.rs
rkuhn: I concur that this might be an issue — perhaps it is better to limit the allocation of the initial capacity to some (configurable?) value. Or perhaps more intuitive, add a second parameter for the initial capacity.

OTOH: if there are such knobs on the facility, people would probably expect the capacity to shrink back to its “desired value” after peak usage. A fixed buffer will always continue to work after initial allocation, but a variable buffer (or rather: thousands thereof) can cause more erratic runtime problems.

elenaf9 (commented):
Nice!
 
Wdyt about adding a `IntoIterator` implementation?
We could additionally consider implementing `Stream` and `Sink` as well, although those would require the `futures` dependency.

misc/async-bounded-queue/src/lib.rs
thomaseizinger: I'd like to avoid multiple configuration parameters if possible. What about no allocating initial capacity and calling `shrink_to_fit` as part of `poll_pop`?

thomaseizinger: That would keep the memory usage as low as possible.

thomaseizinger:
Closing as stale.

PR #3161, opened 2022-11-23
feat: Allow `BandwidthLogging` to be used with more transports

### LINKED ISSUE #3157: [transports/webrtc] Record total number of bytes sent / received in UDPMuxNewAddr
## Description

so the last thing which stops me from releasing webrtc in Substrate (blockchain framework) is bandwidth calculation. The way it's commonly done for other libp2p transports is wrapping Transport with https://docs.rs/libp2p/latest/libp2p/bandwidth/struct.BandwidthLogging.html, which basically records number of bytes written / read across all streams. 

BUT the problem with webrtc Transport is that a call to `DataChannel::write` reports "invalid" number of bytes.
by "invalid" I mean a call write([1,2,3]) will return n=3, but the actual number of bytes will be higher since [1,2,3] is wrapped into `ChunkPayloadData`

the correct amount is recorded by `Association` (https://github.com/webrtc-rs/webrtc/blob/cbfd3033c7605eff97988eef7257694fc3dcd717/sctp/src/association/mod.rs#L471)
but it's neither exposed nor present in stats.

One possible solution is to record total bytes sent / received in UDPMuxNewAddr, where the actual writing / reading from UDP socket happens. Then expose this info in `ListenStream`.

## Motivation

Being able to accurately measure WebRTC connection's bandwidth.

## Open questions

- [x] check that it will be possible to return a trait instead of `BandwidthSinks` struct in Substrate

Yes

thomaseizinger:
## Description

Previously, `BandwidthLogging` implied that `Transport::Output` implements `AsyncRead + AsyncWrite`. For transports like WebRTC and QUIC, this is not true as those transports directly provide a multiplexed and authenticated connection. Thus, `BandwidthLogging` could not be used with those transports.

In the spirit of "all problems in computer science can be solved by an additional layer of indirection", we introduce a layer of indirection in how we introduce `BandwidthSinks` into the wrapped transport. We enforce all `Transport::Output`s that we want to log to implement `WithBandwidthSinks`. This allows us to define a specific strategy for each technology. Unfortunately, we cannot implement generically over all `AsyncRead + AsyncWrite` types without specialization, meaning we now have to list all types we want it implemented on.

For WebRTC, `WithBandwidthSinks` constructs a `Connection` wrapper that implements `StreamMuxer` but also updates the `BandwidthSinks` upon each `StreamMuxer::poll` call. 

Fixes #3157.

## Notes

@melekes I did some quick local testing and things seem to compile with this abstraction. It needs some more filling in:

1. Implementations of `WithBandwidthSinks` for all our streams of our transports.
2. Actual tracking of bytes sent and received in `libp2p-webrtc`.

Can I take you up on your "offer" in https://github.com/libp2p/rust-libp2p/issues/3157 and you continue implementing things from here? :)

## Change checklist

### DIFF
--- a/src/bandwidth.rs
+++ b/src/bandwidth.rs
@@ -72,35 +72,35 @@ impl<TInner> BandwidthLogging<TInner> {
     }
 }
 
-pub trait AttachBandwidthLogger {
+pub trait WithBandwidthSinks {
     type Output;
 
-    fn attach(self, sinks: Arc<BandwidthSinks>) -> Self::Output;
+    fn with_bandwidth_sinks(self, sinks: Arc<BandwidthSinks>) -> Self::Output;
 }
 
 #[cfg(all(feature = "tcp", feature = "async-std"))]
-impl AttachBandwidthLogger for libp2p_tcp::async_io::Stream {
+impl WithBandwidthSinks for libp2p_tcp::async_io::Stream {
     type Output = BandwidthConnecLogging<Self>;
 
-    fn attach(self, sinks: Arc<BandwidthSinks>) -> Self::Output {
+    fn with_bandwidth_sinks(self, sinks: Arc<BandwidthSinks>) -> Self::Output {
         BandwidthConnecLogging { inner: self, sinks }
     }
 }
 
 #[cfg(all(feature = "tcp", feature = "tokio"))]
-impl AttachBandwidthLogger for libp2p_tcp::tokio::Stream {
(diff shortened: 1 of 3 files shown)


melekes:
> Can I take you up on your "offer" in https://github.com/libp2p/rust-libp2p/issues/3157 and you continue implementing things from here? :)

awesome 😎 I was also toying with the code, but this feels like a good abstraction (minus specialisation). Yeah, I can take this from here.

thomaseizinger:
Closing because outdated.

PR #3162, opened 2022-11-23
webrtc: track number of bytes sent / received by all substreams

melekes:
Continuation of https://github.com/libp2p/rust-libp2p/pull/3161

### DIFF
--- a/src/bandwidth.rs
+++ b/src/bandwidth.rs
@@ -162,10 +162,9 @@ pub trait UpdateBandwidthSinks {
 #[cfg(feature = "webrtc")]
 impl UpdateBandwidthSinks for libp2p_webrtc::tokio::Connection {
     fn update_bandwidth_sinks(self: Pin<&mut Self>, sinks: &BandwidthSinks) {
-        sinks.inbound.store(self.total_inbound(), Ordering::Relaxed);
-        sinks
-            .outbound
-            .store(self.total_outbound(), Ordering::Relaxed);
+        let (inbound, outbound) = self.fetch_current_bandwidth();
+        sinks.inbound.fetch_add(inbound, Ordering::Relaxed);
+        sinks.outbound.fetch_add(outbound, Ordering::Relaxed);
     }
 }
 
--- a/src/bandwidth.rs
+++ b/src/bandwidth.rs
@@ -161,8 +161,11 @@ pub trait UpdateBandwidthSinks {
 
 #[cfg(feature = "webrtc")]
 impl UpdateBandwidthSinks for libp2p_webrtc::tokio::Connection {
-    fn update_bandwidth_sinks(self: Pin<&mut Self>, _: &BandwidthSinks) {
-        // TODO: Fetch data from connection via public functions and update sinks
+    fn update_bandwidth_sinks(self: Pin<&mut Self>, sinks: &BandwidthSinks) {
+        sinks.inbound.store(self.total_inbound(), Ordering::Relaxed);
+        sinks
+            .outbound
+            .store(self.total_outbound(), Ordering::Relaxed);
     }
 }
 
--- a/src/bandwidth.rs
+++ b/src/bandwidth.rs
@@ -111,6 +111,140 @@ impl WithBandwidthSinks for (PeerId, libp2p_webrtc::tokio::Connection) {
     }
 }
 
+#[cfg(feature = "deflate")]
+impl<S> WithBandwidthSinks for libp2p_deflate::DeflateOutput<S> {
+    type Output = BandwidthConnecLogging<Self>;
+
+    fn with_bandwidth_sinks(self, sinks: Arc<BandwidthSinks>) -> Self::Output {
+        BandwidthConnecLogging { inner: self, sinks }
+    }
+}
+
+#[cfg(feature = "noise")]
+impl<T, C> WithBandwidthSinks
+    for (
+        libp2p_noise::RemoteIdentity<C>,
+        libp2p_noise::NoiseOutput<T>,
+    )
+{
+    type Output = (
+        libp2p_noise::RemoteIdentity<C>,
+        BandwidthConnecLogging<libp2p_noise::NoiseOutput<T>>,
+    );
+
+    fn with_bandwidth_sinks(self, sinks: Arc<BandwidthSinks>) -> Self::Output {
+        (
+            self.0,
+            BandwidthConnecLogging {
+                inner: self.1,
+                sinks,
+            },
+        )
+    }
+}
+
+#[cfg(feature = "plaintext")]
+impl<C> WithBandwidthSinks for (libp2p_core::PeerId, libp2p_plaintext::PlainTextOutput<C>)
+where
+    C: AsyncRead + AsyncWrite + Unpin,
+{
+    type Output = (
+        libp2p_core::PeerId,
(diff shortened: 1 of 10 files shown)


transports/webrtc/src/tokio/connection.rs
```diff
@@ -158,7 +181,7 @@ impl StreamMuxer for Connection {
             Some(detached) => {
                 log::trace!("Incoming substream {}", detached.stream_identifier());
 
-                let (substream, drop_listener) = Substream::new(detached);
+                let (substream, drop_listener) = Substream::new(detached, self.bandwidth.clone());
```
thomaseizinger: Interesting, I thought you wanted to count the bytes on the UDP socket level? Measuring them here will not include the DTLS framing IIUC.

melekes: I did, but we don't have access to UDP socket here.

melekes: Oh wait, we can pass it when creating connection. I will change the code.

melekes: Nah, doing it involves passing udp_mux to connections, which we're trying to avoid altogether with udp_mux_handle. Probably not worth it.

transports/webrtc/src/tokio/connection.rs
```diff
@@ -144,6 +151,22 @@ impl Connection {
             })
         }));
     }
+
+    /// Returns the total number of bytes received by all connections.
```
melekes, line 155: don't like that API exposes a function which don't belong here ("all connections").

melekes:
Okay, now measurement is more accurate, but I'm slightly worried about added complexity comparing to just logging bytes without webrtc headers & DTLS (https://github.com/libp2p/rust-libp2p/pull/3162/commits/49da84f2db4c82b99ab8aed298279670cee1c9c0).

@tomaka do you have an opinion here?

src/bandwidth.rs
```diff
@@ -161,8 +161,11 @@ pub trait UpdateBandwidthSinks {
 
 #[cfg(feature = "webrtc")]
 impl UpdateBandwidthSinks for libp2p_webrtc::tokio::Connection {
-    fn update_bandwidth_sinks(self: Pin<&mut Self>, _: &BandwidthSinks) {
-        // TODO: Fetch data from connection via public functions and update sinks
+    fn update_bandwidth_sinks(self: Pin<&mut Self>, sinks: &BandwidthSinks) {
+        sinks.inbound.store(self.total_inbound(), Ordering::Relaxed);
```
melekes: @thomaseizinger `sinks` is per connection, right? i.e. `store` is fine

thomaseizinger: I believe `sinks` will be across the entire transport. At least I think that is the case in the current code.

thomaseizinger: > @thomaseizinger `sinks` is per connection, right? i.e. `store` is fine

I don't think `store` is not going to be fine. We need to report the total amount of bytes written here. Exposing that from `Connection` is a bit hacky though.

Alternatively, we could only report the bytes written since we last called `total_inbound` (should have a different name then I guess). That would allow us to add the numbers up here.

melekes: `store` is fine if `connection::total_inbound` returns a total inbound across all connections, which seems completely wrong from API design perspective 😄 

> Alternatively, we could only report the bytes written since we last called total_inbound (should have a different name then I guess). That would allow us to add the numbers up here.

Yep 👍

transports/webrtc/src/tokio/bandwidth.rs
```diff
@@ -0,0 +1,97 @@
...
+    pub(crate) fn new() -> Self {
+        Self {
+            conns: Mutex::new(FnvHashMap::default()),
+        }
+    }
+
+    /// Increases the number of bytes received for a given address.
+    pub(crate) fn add_inbound(&self, addr: &SocketAddr, n: u64) {
+        let mut lock = self.conns.lock().unwrap();
+        let conn_bandwidth = lock.entry(*addr).or_insert(ConnBandwidth {
```
thomaseizinger: `or_default` should do as well.

tomaka:
Just logging without the headers would have the advantage of being a generic solution, as it could be plugged on top of any `StreamMuxer`.
From the perspective of a user of libp2p I don't care either way.

transports/webrtc/src/tokio/connection.rs
```diff
@@ -144,6 +151,13 @@ impl Connection {
             })
         }));
     }
+
+    /// Fetches the current bandwidth (for all connections). Counters are reset after each call.
```
melekes, line 155: I know this is not perfect from API design, but it's the trade-off. You can browse the commit history.

src/bandwidth.rs
```diff
@@ -111,6 +111,140 @@ impl WithBandwidthSinks for (PeerId, libp2p_webrtc::tokio::Connection) {
...
+            BandwidthConnecLogging {
+                inner: self.1,
+                sinks,
+            },
+        )
+    }
+}
+
+// #[cfg(all(feature = "uds", feature = "async-std"))]
+// impl WithBandwidthSinks for async_std::os::unix::net::UnixStream {
```
melekes, line 210: not sure what to do with transports who don't have their own connection type. is it okay to import async_std/tokio here? or I should reexport these types?

transports/websocket/src/lib.rs
```diff
@@ -35,7 +35,7 @@ use libp2p_core::{
     transport::{map::MapFuture, ListenerId, TransportError, TransportEvent},
     Transport,
 };
-use rw_stream_sink::RwStreamSink;
+pub use rw_stream_sink::RwStreamSink;
```
melekes, line 38: is it okay to reexport to avoid importing `rw_stream_sink` in `src/bandwidth.rs`?

thomaseizinger: I'd suggest make a type alias for the stream. That is why I did for TCP.

thomaseizinger:
> Just logging without the headers would have the advantage of being a generic solution, as it could be plugged on top of any `StreamMuxer`.
> From the perspective of a user of libp2p I don't care either way.

Thanks for commenting! I thought I had read a comment from you somewhere that you'd prefer if we measured as far down the stack as possible.

I'd very much be in favor of a generic solution instead of what we have here at the moment. Ideally, maybe even segregated by protocol, e.g. `/ipfs/ping/1.0.0` for example. That would need first-class support in `libp2p-swarm` though but I don't think that is necessarily a bad thing.

However, we can only do that if the trade-offs involved are acceptable for people.

melekes:
> Thanks for commenting! I thought I had read a comment from you somewhere that you'd prefer if we measured as far down the stack as possible.

I was under the same impression as well.

> I'd very much be in favor of a generic solution instead of what we have here at the moment. Ideally, maybe even segregated by protocol, e.g. /ipfs/ping/1.0.0 for example. That would need first-class support in libp2p-swarm though but I don't think that is necessarily a bad thing.

Same. @mxinden proposed counting only the goodput here: https://github.com/libp2p/rust-libp2p/issues/3157#issuecomment-1326126621

melekes:
Let's continue the discussion [here](https://github.com/libp2p/rust-libp2p/issues/3157)

melekes:
Replaced by https://github.com/libp2p/rust-libp2p/pull/3180

PR #3152, opened 2022-11-22
fix(kad): Don't wait for behaviour after receiving `AddProvider`

thomaseizinger:
## Description

Previously, we would erroneously always go into the `WaitingUser` (now called `WaitingBehaviour`) state after receiving a message on an inbound stream. However, the `AddProvider` message does not warrant a "response" from the behaviour. Thus, any incoming `AddProvider` message would result in a stale substream that would eventually be dropped as soon as more than 32 inbound streams have been opened.

With this patch, we inline the message handling into the upper match block and perform the correct state transition upon each message. For `AddProvider`, we go back into `WaitingMessage` because the spec mandates that we need to be ready to receive more messages on an inbound stream.

Fixes #3048.

## Change checklist

- [x] I have performed a self-review of my own code
- [x] I have made corresponding changes to the documentation
- [ ] ~I have added tests that prove my fix is effective or that my feature works~
- [x] A changelog entry has been made in the appropriate crates

PR #3164, opened 2022-11-24
fix(quic): allow closing stream after remote dropped it

### LINKED ISSUE #3144: transports/quic: `STOP_SENDING` fails `poll_close`
## Summary

Receiving a `STOP_SENDING` makes `poll_close` fail. This is problematic in `libp2p-identify` where a node expects to be able to close its write side before reading.

## Expected behaviour

`libp2p-identify` should successfully exchange identify information.

## Actual behaviour

Say that node A and B connected. A opens a `libp2p-identify` stream to B. A expects B to send its identify information.

After having negotiated the `libp2p-identify` protocol via `multistream-select`:

- B sends its identify information and closes its write side.
    https://github.com/libp2p/rust-libp2p/blob/0c85839dabf06beeb853602c0e69f4753b25b751/protocols/identify/src/protocol.rs#L202-L203
  B then drops the stream. Dropping the stream results in `libp2p-quic` sending a `STOP_SENDING`.
  https://github.com/libp2p/rust-libp2p/blob/0c85839dabf06beeb853602c0e69f4753b25b751/transports/quic/src/connection/substream.rs#L203-L207

- In parallel A closes its write side.
    https://github.com/libp2p/rust-libp2p/blob/0c85839dabf06beeb853602c0e69f4753b25b751/protocols/identify/src/protocol.rs#L212
  This results in `Substream::poll_close` being called in `libp2p-quic`.
  https://github.com/libp2p/rust-libp2p/blob/0c85839dabf06beeb853602c0e69f4753b25b751/transports/quic/src/connection/substream.rs#L177-L192
  In case the `STOP_SENDING` from B already arived, `poll_close` fails and thus the identify handshake fails.

## Possible Solution
1. Don't send a `STOP_SENDING` on `Drop`.
    - That would be unfortunate, the remote might send data which will never be received.
2. Don't send a `STOP_SENDING` on `DROP` IFF the remote already closed the write side.
    - That would be in-line with what upstream `quinn` does. https://github.com/quinn-rs/quinn/blob/1d390af2facdf424c7feb470909fffc29a1dca6c/quinn/src/recv_stream.rs#L396-L400
    - I am not sure it would entirely prevent this issue. E.g. A's close might not make it to B before B sends the `STOP_SENDING`.
3. Ignore the `STOP_SENDING` error in `poll_close`.
    - That would give a false sense to upstream users. While they think closing was successful and thus assume (one can never be sure) that ones data made it to the remote, there is actually a high likelihood that remote never received the data.
4. Don't close the write side in `libp2p-identify` before reading the remote's identify information.
    - I would expect `libp2p-identify` to be an edge-case here, i.e. I would expect other protocols not to run into this issue. E.g. in a standard request response style protocol A would send a request, then close, then B would read, then B writes, then B closes and only then B drops.

I am tending towards (2) AND (4).

## Version

- libp2p version (version number, commit, or branch):

I am using 055636eef on https://github.com/libp2p/rust-libp2p/pull/2712 but any commit on `master` should work. E.g.  https://github.com/libp2p/rust-libp2p/commit/0c85839dabf06beeb853602c0e69f4753b25b751.

Maybe. Let's find consensus on a good solution first.

//CC @elenaf9

elenaf9:
## Description

- Only send `STOP_SENDING` on a stream when dropping it if the remote did not finish the stream yet.
- Only call `quinn_proto::SendStream::finish` once. (A second call to it will always fail. Though I don't think this was the issue in #3144.)
- Add tests for reading and writing to streams after the remote dropped. Also adds a smoke test for backpressure.

Fixes #3144.

Fixes #3144 following proposed idea nr.2 from @mxinden.
> Don't send a STOP_SENDING on DROP IFF the remote already closed the write side.
>
> -  That would be in-line with what upstream quinn does. https://github.com/quinn-rs/quinn/blob/1d390af2facdf424c7feb470909fffc29a1dca6c/quinn/src/recv_stream.rs#L396-L400

I implemented it slightly different compared to quinn, since quinn only sets the `all_data_read` flag when the stream finishes while reading. In identify however we might receive a FIN for a stream on which we are not reading anymore, thus what we do now is in the `drop` implementation we manually check the receive stream again if it has finished.

## Change checklist


transports/quic/src/connection/substream.rs
```diff
@@ -204,7 +217,21 @@ impl Drop for Substream {
     fn drop(&mut self) {
         let mut state = self.state.lock();
         state.substreams.remove(&self.id);
-        let _ = state.connection.recv_stream(self.id).stop(0u32.into());
+        // Send `STOP_STREAM` if the remote did not finish the stream yet.
+        // We have to manually check the read stream since we might may have
```
mxinden:
```suggestion
        // We have to manually check the read stream since we might have
```

mxinden (commented):
Thanks @elenaf9. I will test this out in a bit.

mxinden:
I deployed this to kademlia-exporter.max-inden.de/ and **successfully** connected to it via a patched https://github.com/mxinden/libp2p-lookup.

elenaf9:
> I deployed this to kademlia-exporter.max-inden.de/ and **successfully** connected to it via a patched https://github.com/mxinden/libp2p-lookup.

Nice! Do I understand it correctly that before it always failed due to the issue in identify? Or was it a flaky problem and we need to do some more testing before we can confirm #3144 is fixed?

mxinden:
> Nice! Do I understand it correctly that before it always failed due to the issue in identify?

Correct. It was consistently failing.

transports/quic/src/connection/substream.rs
```diff
@@ -40,7 +40,10 @@ pub struct SubstreamState {
     /// Waker to wake if the substream becomes closed or stopped.
     pub finished_waker: Option<Waker>,
 
-    /// `true` if the stream finished, i.e. the writing side closed.
+    /// `true` if the writing side of the stream is closing, i.e. `AsyncWrite::poll_close`
+    /// was called.
+    pub is_finishing: bool,
```
mxinden:
```suggestion
    pub is_write_closing: bool,
```

Would that not be more intuitive / consistent with `is_write_closed`?

mxinden: An alternative would be to introduce an enum:

``` rust
write_state: WriteState,

// ...

enum WriteState {
  Open,
  Closing,
  Closed,
}
```

I don't have a strong opinion.

transports/quic/tests/smoke.rs
```diff
@@ -182,6 +183,66 @@ fn concurrent_connections_and_streams_tokio() {
...
+    let (mut stream_a, mut stream_b) = build_streams::<quic::async_std::Provider>().await;
+
+    let data = vec![0; max_stream_data as usize - 1];
+
+    stream_a.write_all(&data).await.unwrap();
+
+    let too_much_data = vec![0; 10];
+
+    assert!(
+        async_std::future::timeout(Duration::from_secs(1), stream_a.write_all(&too_much_data))
```
mxinden: Usage of a timeout in a unit test is unfortunate. That said, I can not think of an alternative, and I think it is worth it. Thus I suggest let's keep as is.

transports/quic/tests/smoke.rs
```diff
@@ -182,6 +183,66 @@ fn concurrent_connections_and_streams_tokio() {
         .quickcheck(prop::<quic::tokio::Provider> as fn(_, _) -> _);
 }
 
+#[cfg(feature = "async-std")]
+#[async_std::test]
+async fn backpressure() {
```
mxinden, line 188: :rocket:

mxinden (commented):
Small suggestion. Otherwise looks good to me. Thanks for the patch and the unit tests!

transports/quic/tests/smoke.rs
```diff
@@ -182,6 +183,66 @@ fn concurrent_connections_and_streams_tokio() {
...
+async fn write_after_peer_dropped_stream() {
+    let _ = env_logger::try_init();
+    let (stream_a, mut stream_b) = build_streams::<quic::async_std::Provider>().await;
+    drop(stream_a);
+    futures_timer::Delay::new(Duration::from_millis(1)).await;
+
+    let data = vec![0; 10];
+    stream_b.write_all(&data).await.expect("Write failed.");
+    stream_b.close().await.expect("Close failed.");
+}
```
thomaseizinger, line 241: Can we move these to the muxer test harness? :)

transports/quic/tests/smoke.rs
thomaseizinger: You could assert that it returns `Poll::Pending` instead and `Poll::Ready` once you read from the other side again.

transports/quic/tests/smoke.rs
```diff
@@ -182,6 +183,66 @@ fn concurrent_connections_and_streams_tokio() {
...
+#[should_panic]
+async fn write_after_peer_dropped_stream() {
+    let _ = env_logger::try_init();
+    let (stream_a, mut stream_b) = build_streams::<quic::async_std::Provider>().await;
+    drop(stream_a);
+    futures_timer::Delay::new(Duration::from_millis(1)).await;
+
+    let data = vec![0; 10];
+    stream_b.write_all(&data).await.expect("Write failed.");
+    stream_b.close().await.expect("Close failed.");
```
thomaseizinger, line 240: Is this the line that will panic? I'd like to see an `unwrap_err` then.

transports/quic/src/connection/substream.rs
thomaseizinger: `finish` is QUIC terminology right?

thomaseizinger (approved):
Nice! A few suggestions!

thomaseizinger:
A few suggestions regarding the PR description/title:

1. I'd suggest for scope to not be paths but modules, i.e. just quic is good enough and shortens the title (under 50 characters total is a good convention for commit messages)
2. "fix" is currently repeated, maybe a better title would be: "fix(quic): allow to read from a closed & dropped stream". i.e. describe what is possible now but failed before
3. Instead of bullet points in the commit message body (i.e. PR description), make it one or more paragraphs.
4. Put the `Fixes #XYZ` as the last line of the commit message.

I also often reference this blog post: https://cbea.ms/git-commit/

If we follow these guidelines, all squash commits will be nice commits ☺️

transports/quic/tests/smoke.rs
elenaf9: I am in favor of moving them to the test harness. 
However, contrary to the other tests in the harness these tests test a specific sequential order in which alice and bob do their calls, rather than the things running in parallel. So I think it would require some more "infrastructure" code and ideally we'd also add it for the other muxers. Since I don't want to block this PR on this I would propose to do it in a follow up PR.

transports/quic/tests/smoke.rs
elenaf9: Unfortunately this is not deterministic; it may either panic on the `write_all` or on the `close`, depending on whether the `STOP_SENDING` frame made it to B before B starts writing or not.

transports/quic/src/connection/substream.rs
elenaf9: I like `WriteState`. Added in c6256e1.

> `finish` is QUIC terminology right?

Not really QUIC terminology, but used in `quinn`. I stuck with it to avoid confusion on inner events like `StreamEvent::Finished`. But thinking about it again I think using the `AsyncWrite` terminology of `close` is more intuitive.

transports/quic/tests/smoke.rs
elenaf9: > You could assert that it returns `Poll::Pending` instead and `Poll::Ready` once you read from the other side again.

:+1:. Done in 9f2a7ee.

elenaf9:
> A few suggestions regarding the PR description/title:
> 
>  1. I'd suggest for scope to not be paths but modules, i.e. just quic is good enough and shortens the title (under 50 characters total is a good convention for commit messages)
> 
>  2. "fix" is currently repeated, maybe a better title would be: "fix(quic): allow to read from a closed & dropped stream". i.e. describe what is possible now but failed before
> 
>  3. Instead of bullet points in the commit message body (i.e. PR description), make it one or more paragraphs.
> 
>  4. Put the `Fixes #XYZ` as the last line of the commit message.
> 
> 
> I also often reference this blog post: https://cbea.ms/git-commit/
> 
> If we follow these guidelines, all squash commits will be nice commits relaxed

Agree with all but nr 3. I personally prefer bullet points to avoid unnecessary filler words, at least if no additional explanation is needed :slightly_smiling_face:

Edit: Ah you might be referring to the formatting and not the phrasing itself. I don't feel strongly about it, but why not use bullet points?

mxinden (approved):

mxinden:
Adding `send-it` here to include it in `libp2p` `v0.50.0`. As far as I can tell, only outstanding discussion is the move to the muxer test harness (https://github.com/libp2p/rust-libp2p/pull/3164#discussion_r1031805526). @elenaf9 could you follow up on that?

Thanks for prioritizing this @elenaf9 :pray:

transports/quic/tests/smoke.rs
thomaseizinger: Right, that makes sense. I didn't consider that :)

Pushing the boundaries of the harness here 😁

thomaseizinger:
> > A few suggestions regarding the PR description/title:
> > 
> >  1. I'd suggest for scope to not be paths but modules, i.e. just quic is good enough and shortens the title (under 50 characters total is a good convention for commit messages)
> > 
> >  2. "fix" is currently repeated, maybe a better title would be: "fix(quic): allow to read from a closed & dropped stream". i.e. describe what is possible now but failed before
> > 
> >  3. Instead of bullet points in the commit message body (i.e. PR description), make it one or more paragraphs.
> > 
> >  4. Put the `Fixes #XYZ` as the last line of the commit message.
> > 
> > 
> > I also often reference this blog post: https://cbea.ms/git-commit/
> > 
> > If we follow these guidelines, all squash commits will be nice commits relaxed
> 
> Agree with all but nr 3. I personally prefer bullet points to avoid unnecessary filler words, at least if no additional explanation is needed :slightly_smiling_face:
> 
> Edit: Ah you might be referring to the formatting and not the phrasing itself. I don't feel strongly about it, but why not use bullet points?

I totally get what you are saying and I agree with the gist of it. The issue with bullet points can be that they are often too concise. It is sometimes nice to read a little paragraph that summarizes what is happening and why. Often, I start the commit message with:

"Previously, our behaviour was XYZ. That caused problem ABC. We fix this by doing 1, 2 and 3."

That could or could not involve bulletpoints but it makes it clear what is the description of the original behaviour, what is the problem and what is the fix. I don't think there are too many unnecessary filler words in there but YMMV :)

Here is a recent example: https://github.com/libp2p/rust-libp2p/pull/3152

I don't feel too strongly about it, just noticed it and wanted to start a discussion to see where everyone is at!

transports/quic/tests/smoke.rs
elenaf9: Tracked the idea in issue #3211.

PR #3180, opened 2022-11-30
feat: log bandwidth on substream instead of socket level

### LINKED ISSUE #3157: [transports/webrtc] Record total number of bytes sent / received in UDPMuxNewAddr

melekes:
## Description

Previously, the `with_bandwidth_logging` extension to `Transport` would track the bytes sent and received on a socket level. This however only works in conjunction with `Transport` upgrades where a separate multiplexer is negotiated on top of a regular stream.

With QUIC and WebRTC landing, this no longer works as those `Transport`s bring their own multiplexing stack. To still allow for tracking of bandwidth, we refactor the `with_bandwidth_logging` extension to count the bytes send on all substreams opened through a `StreamMuxer`. This works, regardless of the underlying transport technology. It does omit certain layers. However, there isn't necessarily a "correct" layer to count bandwidth on because you can always go down another layer (IP, Ethernet, etc).

Closes #3157.

## Change checklist


swarm/src/lib.rs
```diff
@@ -457,6 +459,24 @@ where
...
+        let sinks_copy = sinks.clone();
+        self.transport = Transport::map(self.transport, |(peer_id, stream_muxer_box), _| {
+            (
+                peer_id,
+                StreamMuxerBox::new(BandwidthLogging::new(stream_muxer_box, sinks_copy)),
+            )
+        })
+        .boxed();
+        sinks
+    }
```
thomaseizinger: I don't think this should live at the `Swarm` level. Instead, I think we can have this in `libp2p` same as before but implement `TransportExt` for every `Transport` that outputs `(PeerId, S)` with `S: StreamMuxer`. Does that compile?

melekes: something like this: https://github.com/libp2p/rust-libp2p/pull/3180/commits/fac29fc95ed100f5fd7efedca8a1ac57e752dd83 ?

CHANGELOG.md
```diff
@@ -45,6 +45,10 @@
 
 # `libp2p` facade crate
 
+# Unreleased
```
```suggestion
# 0.51.0 [unreleased]
```

CHANGELOG.md
```diff
@@ -45,6 +45,10 @@
...
+
+- Count bandwidth at the application level ([PR 3180](https://github.com/libp2p/rust-libp2p/pull/3180)).
```
```suggestion
- Count bandwidth at the application level. Previously `BandwidthLogging` would implement `Transport` and now implements `StreamMuxer` ([PR 3180](https://github.com/libp2p/rust-libp2p/pull/3180)).
```

src/bandwidth.rs
```diff
@@ -43,120 +36,98 @@ use std::{
     task::{Context, Poll},
 };
 
-/// Wraps around a `Transport` and counts the number of bytes that go through all the opened
-/// connections.
+/// Wraps around a `StreamMuxer` and counts the number of bytes that go through all the opened
```
mxinden:
```suggestion
/// Wraps around a [`StreamMuxer`] and counts the number of bytes that go through all the opened
```

src/bandwidth.rs
```diff
@@ -177,13 +148,13 @@ impl BandwidthSinks {
 
 /// Wraps around an `AsyncRead + AsyncWrite` and logs the bandwidth that goes through it.
 #[pin_project::pin_project]
-pub struct BandwidthConnecLogging<TInner> {
+pub struct BandwidthConnecLogging<SMInner> {
```
mxinden:
```suggestion
pub struct InstrumentedStream<SMInner> {
```

This is no longer a connection, thus I would suggest a rename. How about `InstrumentedStream`?

src/bandwidth.rs
```diff
@@ -177,13 +148,13 @@ impl BandwidthSinks {
 
 /// Wraps around an `AsyncRead + AsyncWrite` and logs the bandwidth that goes through it.
```
mxinden:
```suggestion
/// Wraps around an [`AsyncRead`] + [`AsyncWrite`] and logs the bandwidth that goes through it.
```

While we are at it, if you don't mind.

src/bandwidth.rs
```diff
@@ -252,3 +223,13 @@ impl<TInner: AsyncWrite> AsyncWrite for BandwidthConnecLogging<TInner> {
         this.inner.poll_close(cx)
     }
 }
+
+/// Wraps around a `Future` that produces a connection. Wraps the connection around a bandwidth
+/// counter.
+#[deprecated(since = "0.50.1")]
+#[pin_project::pin_project]
+pub struct BandwidthFuture<TInner> {
```
mxinden: Changing `BandwidthLogging` to implement `StreamMuxer` instead of `Transport` is a breaking change. Given that you are proposing a breaking change anyways, why deprecate here?

mxinden (commented):
Direction looks good to me. A couple of comments.

src/bandwidth.rs
melekes: But that would be a breaking API change meaning we won't be able to release it in 0.50.1 and will have to wait 0.51, no?

src/bandwidth.rs
melekes: Oh, right. Will remove

src/transport_ext.rs
```diff
@@ -31,11 +39,31 @@ pub trait TransportExt: Transport {
     ///
```
mxinden, line 39: Comment above needs updating. We are no longer logging at the socket layer.

src/transport_ext.rs
```diff
@@ -31,11 +39,31 @@ pub trait TransportExt: Transport {
...
     /// This method returns an `Arc<BandwidthSinks>` that can be used to retreive the total number
```
mxinden:
```suggestion
    /// This method returns an `Arc<BandwidthSinks>` that can be used to retrieve the total number
```

src/transport_ext.rs
```diff
@@ -31,11 +39,31 @@ pub trait TransportExt: Transport {
...
     /// of bytes transferred through the sockets.
-    fn with_bandwidth_logging(self) -> (BandwidthLogging<Self>, Arc<BandwidthSinks>)
+    fn with_bandwidth_logging<S>(self) -> (Boxed<(PeerId, StreamMuxerBox)>, Arc<BandwidthSinks>)
```
mxinden, line 69: Can you either add a small test, or add a doc example to the `with_bandwidth_logging` method (preferred)?

src/transport_ext.rs
```diff
@@ -31,11 +39,31 @@ pub trait TransportExt: Transport {
...
+        S::Error: Send + Sync + 'static,
     {
-        BandwidthLogging::new(self)
+        let sinks = BandwidthSinks::new();
+        let sinks_copy = sinks.clone();
+        let transport = Transport::map(self, |output, _| {
+            let (peer_id, stream_muxer_box) = output.into();
+            (
+                peer_id,
+                StreamMuxerBox::new(BandwidthLogging::new(
```
thomaseizinger: If we use `StreamMuxerBox` here, then lets make `BandwidthLogging` and `InstrumentStream` `pub(crate)`!

thomaseizinger (approved):
Nice work, thank you!

One suggestion, otherwise happy with this modulo @mxinden's comments!

swarm/src/lib.rs
thomaseizinger: Yep, thanks!

thomaseizinger:
I've updated the PR description to prepare this for merging so we have a decent commit message.

src/transport_ext.rs
melekes: Added a doc test (ignored because it requires certain features)

mxinden: Thanks for adding the test.

In my eyes a doc test that is `ignore`d is not helpful and destined to be outdated. Is there some way to enable the features for testing only? //CC @thomaseizinger

thomaseizinger: Doc tests have access to all `dev-dependencies` so you can simply add whatever you need there and it will be available in the test. You will have to change the imports to though.

Cargo.toml
```diff
@@ -136,6 +136,10 @@ env_logger = "0.10.0"
 clap = { version = "4.0.13", features = ["derive"] }
 tokio = { version = "1.15", features = ["io-util", "io-std", "macros", "rt", "rt-multi-thread"] }
```
thomaseizinger, line 138: Why split these with a newline?

thomaseizinger (dismissed):
LGTM, thanks!

thomaseizinger (approved):

Cargo.toml
```diff
@@ -136,6 +136,10 @@ env_logger = "0.10.0"
...
 
+libp2p-mplex = { version = "0.38.0", path = "muxers/mplex" }
+libp2p-noise = { version = "0.41.0", path = "transports/noise" }
+libp2p-tcp = { version = "0.38.0", path = "transports/tcp", features = ["tokio"] }
```
mxinden, line 141: Is it a deliberate choice to link these with a `version`?

mxinden: Would be fixed with https://github.com/libp2p/rust-libp2p/pull/3261.

Issue #3186, opened 2022-12-02
swarm: Deprecate `Swarm::with_XYZ` and enforce creation via `SwarmBuilder`


thomaseizinger:
## Description

With the introduction of the `Swarm::with_executor` APIs (#3068), we have a fair bit of duplication in APIs on how to create a `Swarm`. Additionally, the current `SwarmBuilder` and `Swarm` APIs don't work well with the QUIC and WebRTC transports as those don't go through the typical builder pattern on `Transport` and hereby don't construct a `StreamMuxerBox` (see https://github.com/libp2p/rust-libp2p/issues/3179).

We should streamline the creation of a `Swarm` to work with any `Transport` and be easy to discover. @mxinden [suggested](https://github.com/libp2p/rust-libp2p/issues/3186#issuecomment-1342845258) to achieve this by forcing users to always go through `SwarmBuilder`.

## Motivation

- Less API surface and layers of indirection.

## Are you planning to do it yourself in a pull request?

Maybe.

mxinden:
Using the builder pattern to construct a `Swarm` enforces at compile time that a user can not reconfigure the below values at runtime after `Swarm` creation. I would like to uphold this enforcement at compile time, or explicitly allow it for a subset of the values below. The latter would require more thought, though at this point I don't see a use-case where a user would need to reconfigure these after creation during runtime.

> * `notify_handler_buffer_size`
> 
> * `connection_event_buffer_size`
> 
> * `dial_concurrency_factor`
> 
> * `connection_limits`
> 
> * `substream_upgrade_protocol_override`
> 
> * `max_negotiating_inbound_streams`

With that in mind I am advocating against removing `SwarmBuilder`.

> * Less API surface

In case we want to improve here, I would suggest going the opposite route, namely to enforce the creation of a `Swarm` through the `SwarmBuilder`, thus reducing the paths to creating a `Swarm` to a single one, namely through the `SwarmBuilder`.

thomaseizinger:
> > * Less API surface
> 
> In case we want to improve here, I would suggest going the opposite route, namely to enforce the creation of a `Swarm` through the `SwarmBuilder`, thus reducing the paths to creating a `Swarm` to a single one, namely through the `SwarmBuilder`.

That sounds reasonable to me!

thomaseizinger:
[...]

PR #3187, opened 2022-12-02
refactor(swarm): don't use `task_event_buffer_size` for pending connections

thomaseizinger:
## Description

The task for a pending connection only ever sends one event into this channel: Either a success or a failure. Cloning a sender adds one slot to the capacity of the channel. Hence, we can start this capacity at 0 and have the `cloning` of the `Sender` take care of properly increasing the capacity.

## Notes

~This is a first step towards https://github.com/libp2p/rust-libp2p/issues/3186 by removing one use of this parameter.~

## Change checklist


mxinden (approved):
Change looks good to me. We could even go as far as using a `oneshot` here.

thomaseizinger:
> Change looks good to me. We could even go as far as using a `oneshot` here.

That would make it more difficult to listen for the events although a `SelectAll<Receiver>` could do the trick.

PR #3189, opened 2022-12-02
feat(swarm): deprecate `SwarmBuilder` in favor of configuring `Swarm`

### LINKED ISSUE #3186: swarm: Deprecate `Swarm::with_XYZ` and enforce creation via `SwarmBuilder`

thomaseizinger:
## Description

We expose various fields on `Pool` as `pub(crate)` so we can modify them after a `Pool` has been constructed. This allows us to deprecate `SwarmBuilder` and add configuration options to `Swarm` directly.

Resolves #3186.

## Notes

Depends-On: https://github.com/libp2p/rust-libp2p/pull/3188

## Change checklist


mxinden:
As mentioned in https://github.com/libp2p/rust-libp2p/issues/3186#issuecomment-1342845258 I would prefer making `SwarmBuilder` the only mechanism through which one can obtain a `Swarm`. @thomaseizinger are we on the same page or am I missing something?

thomaseizinger:
> As mentioned in https://github.com/libp2p/rust-libp2p/issues/3186#issuecomment-1342845258 I would prefer making `SwarmBuilder` the only mechanism through which one can obtain a `Swarm`. @thomaseizinger are we on the same page or am I missing something? 

We are on the same page but this PR is older than your comment :)

Issue #2885, opened 2022-09-12
protocols/identify: Move I/O to ConnectionHandler


mxinden:
## Description

I/O should not happen in the `NetworkBehaviour` implementation. It should happen in the `ConnectionHandler` implementation.

## Motivation

All `NetworkBehaviour` implementations are driven by a single future task. Opposite to that we spawn a new future task for every connection. Doing as much as possible, especially I/O, on the connection tasks and not on the main task allows logic to potentially run in parallel.

## Current Implementation

Today `libp2p-identify` drives the future returning a response to a remote in the `NetworkBehaviour::poll` implementation. 
https://github.com/libp2p/rust-libp2p/blob/66c275520d8cf56152cce9a1c24bb4e2c42e0833/protocols/identify/src/identify.rs#L53-L54

## Are you planning to do it yourself in a pull request?

No

thomaseizinger:
The problem with today's implementation of identify is that we only know the observed address in the `NetworkBehaviour`. To fix this, we will have to implement `IntoConnectionHandler` for the identify protocol and grab the observed address out of the `ConnectedPoint`. Going further, this information can be passed into the `InboundUpgrade` implementation via the `OpenInboundInfo` type in `ConnectedHandler`.

jxs:
@thomaseizinger could you shed some light on how you envisioned passing the observed address from `Behaviour` to `Handler` ? 

I can't really see where to do that as `Behaviour::new_handler()` doesn't have access to the `MultiAddress`.  
Meanwhile I started implementing this using `NetworkBehaviourAction::NotifyHandler` to pass the `MultiAddress` and the `Substream` to the `Handler`, but it doesn't seem semantically correct passing the `Substream` from the Handler to the Behaviour to then get it back from the `Behaviour` again, I feel this should be encapsulated in the Handler.

Thanks :)

thomaseizinger:
You need to implement `IntoConnectionHandler` instead of `ConnectionHandler`. That will give you access to the `ConnectedPoint`.

jxs:
right, thanks Thomas. What about `listen_addresses` and `supported_protocols` needed to create the Identify `Info`?  
We only get them from `PollParameters` on `NetworkBehaviour::poll` not `ConnectionHandler:poll` and we need  .
This maybe also affects the development https://github.com/libp2p/rust-libp2p/pull/3153.

thomaseizinger:
[...]

PR #3208, opened 2022-12-07
feat: move Identify I/O from NetworkBehaviour to ConnectionHandler

jxs:
## Description
Addresses #2885 

## Notes
Following the discussion on https://github.com/libp2p/rust-libp2p/issues/2885 this was the simplest way I found to move the I/O to `Handler`. 
Per commit review is suggested. 
## Change checklist


protocols/identify/src/behaviour.rs
```diff
@@ -63,18 +61,10 @@ pub struct Behaviour {
...
-        io: ReplySubstream<NegotiatedSubstream>,
-        observed: Multiaddr,
-    },
-    /// The reply is being sent.
-    Sending {
-        peer: PeerId,
-        io: Pin<Box<dyn Future<Output = Result<(), UpgradeError>> + Send>>,
-    },
+struct Request {
+    peer: PeerId,
```
thomaseizinger: The handler could learn this via `IntoConnectionHandler`. It will never change across its lifespan.

protocols/identify/src/behaviour.rs
```diff
@@ -63,18 +61,10 @@ pub struct Behaviour {
...
+    io: ReplySubstream<NegotiatedSubstream>,
```
thomaseizinger: It would be nice to not pass this back and forth.

protocols/identify/src/behaviour.rs
```diff
@@ -63,18 +61,10 @@ pub struct Behaviour {
...
+    observed: Multiaddr,
```
thomaseizinger: Same here, a `Handler` can learn this via `IntoConnectionHandler` and it won't change[^1] across its lifetime.

[^1]: `AddressChange` is not emitted at the moment but once it will, it is directly given to the handler.

thomaseizinger (commented):
Thanks!

It would be nice if we changed a few more things:

- Initialise the `ConnectionHandler` with `PeerId` and observed address directly
- Don't send the substream back and forth
- Upon receiving a new substream, send an event to the behaviour asking for more information like protocols, upon answer, grab the next substream in line and send it back

I think this would be an overall easier design because it clearly shows, what information is needed from the behaviour that can't be accessed by the handler.

We can even pass stuff like the public key and the agent version directly to the handler upon initialisation.

protocols/identify/src/handler.rs
```diff
@@ -255,6 +302,46 @@ impl ConnectionHandler for Handler {
             }
         }
 
+        // Check for pending replies to send.
+        if let Some(ref info) = self.info {
```
jxs: NB: I thought `let_chains` had landed already been stabilized, but[ it was reverted](https://github.com/rust-lang/rust/pull/100538).

protocols/identify/src/behaviour.rs
jxs: Yeah agreed thanks, updated!

protocols/identify/src/behaviour.rs
jxs: but we only get it on `connection_established` which seems to be called after `new_handler`.

jxs (commented):
Thanks for the fast review and guidance Thomas!

>I think this would be an overall easier design because it clearly shows, what information is needed from the behaviour that can't be accessed by the handler.
>
>We can even pass stuff like the public key and the agent version directly to the handler upon initialisation.

the downside of that is that information comes from multiple different places instead of coming only from a single `Info` struct. I updated to code to no longer send back and forth the substream, if you still feel that we should pass first the `PeerId` on constructor and then all the other details when requested, to prepare for the changes discussed on https://github.com/libp2p/rust-libp2p/issues/2885 no worries I'll update the PR :)

protocols/identify/src/handler.rs
thomaseizinger: Really? Did you try with Rust 1.65?

protocols/identify/src/behaviour.rs
thomaseizinger: You have to implement `IntoConnectionHandler` additionally on a struct that you return from `new_handler`, said struct can then capture the `ConnectedPoint` and initialise the `ConnectionHandler` with it.

thomaseizinger:
> Thanks for the fast review and guidance Thomas!
> 
> >I think this would be an overall easier design because it clearly shows, what information is needed from the behaviour that can't be accessed by the handler.
> >
> >We can even pass stuff like the public key and the agent version directly to the handler upon initialisation.
> 
> the downside of that is that information comes from multiple different places instead of coming only from a single `Info` struct.

Information already comes from different places! I think it is cleaner to not hide this fact but explicitly show, what information we request from the behaviour.

> all the other details when requested

Everything that doesn't change over the lifetime of the handler should come in via the constructor in my opinion. This accurately models in the type system, what data is known when and how it flows through the system! In other words, only the supported protocols and external addresses should come in via the event.

protocols/identify/src/handler.rs
jxs: yeah, compiler points me to https://github.com/rust-lang/rust/issues/53667, see https://github.com/rust-lang/rust/issues/53667#issuecomment-1297933589

protocols/identify/src/behaviour.rs
jxs: thanks Thomas, updated!

protocols/identify/src/behaviour.rs
jxs: updated!

protocols/identify/src/behaviour.rs
```diff
@@ -51,8 +48,8 @@ pub struct Behaviour {
     config: Config,
     /// For each peer we're connected to, the observed address to send back to it.
     connected: HashMap<PeerId, HashMap<ConnectionId, Multiaddr>>,
-    /// Pending replies to send.
-    pending_replies: VecDeque<Reply>,
+    /// Information requests from the handlers to be fullfiled.
+    requests: VecDeque<PeerId>,
```
thomaseizinger: I think you'll need to track the connection ID here so you can respond to the correct handler!

protocols/identify/src/behaviour.rs
```diff
@@ -346,55 +327,17 @@ impl NetworkBehaviour for Behaviour {
...
-            }
+        // Check for information requests from the handlers.
+        if let Some(peer) = self.requests.pop_front() {
+            let info = BehaviourInfo {
+                listen_addrs: listen_addrs(params),
+                protocols: supported_protocols(params),
+            };
+            return Poll::Ready(NetworkBehaviourAction::NotifyHandler {
+                peer_id: peer,
+                handler: NotifyHandler::Any,
```
thomaseizinger, line 319: Here is the curlpit! This will be buggy with > 1 connection per peer.

protocols/identify/src/handler.rs
```diff
@@ -90,36 +130,84 @@ pub struct Handler {
...
+
+    /// Name and version of the peer, similar to the `User-Agent` header in
+    /// the HTTP protocol.
+    agent_version: String,
+
+    /// Address observed by or for the remote.
+    observed_addr: Multiaddr,
+
+    /// Information provided by the `Behaviour` upon requesting.
+    behaviour_info: Option<BehaviourInfo>,
```
thomaseizinger: Instead of tracking this here, I think it would be cleaner to:

1. Store pending substreams in a queue
2. Request a new `BehaviourInfo` every time a new substream comes in
3. Construct a `Sending` every time you receive a `BehaviourInfo` from the behaviour by popping the oldest substream from the queue in (1)

That will work off substreams in a FIFO fashion and feed them the most up-to-date information.

thomaseizinger: Alternatively, we could initialise this in the constructor as well. I think we can assume that `new_handler` in the behaviour only gets called after it got polled at least once!

protocols/identify/src/handler.rs
```diff
@@ -90,36 +130,84 @@ pub struct Handler {
...
+}
+
+/// Information provided by the `Behaviour` upon requesting.
+#[derive(Debug)]
+pub struct BehaviourInfo {
```
thomaseizinger: Not the best name but should do for now I guess :)

thomaseizinger (commented):
Thanks! A few more ideas!

protocols/identify/src/behaviour.rs
jxs: Thanks Thomas, addressed.

protocols/identify/src/handler.rs
jxs: > Instead of tracking this here, I think it would be cleaner to:
> 
> 1. Store pending substreams in a queue
> 2. Request a new `BehaviourInfo` every time a new substream comes in
> 3. Construct a `Sending` every time you receive a `BehaviourInfo` from the behaviour by popping the oldest substream from the queue in (1)
> 
> That will work off substreams in a FIFO fashion and feed them the most up-to-date information.

Thanks, makes sense to me. But used two queues instead of one with different states for a less confusing management of them ptal Thomas.

protocols/identify/src/handler.rs
```diff
@@ -207,14 +299,37 @@ impl ConnectionHandler for Handler {
...
+                    agent_version: self.agent_version.clone(),
+                    listen_addrs: behaviour_info.listen_addrs,
+                    protocols: behaviour_info.protocols,
+                    observed_addr: self.observed_addr.clone(),
+                };
+                let substream = self
+                    .reply_streams
+                    .pop_front()
+                    .expect("A BehaviourInfo reply should have a matching substream.");
+                let io = Box::pin(substream.send(info));
```
thomaseizinger:
```suggestion
                let io = Box::pin(async move { substream.send(info).await?; Ok(self.remote_peer_id) });
```

If you construct the future this way, you don't need the `Sending` struct and can you `FuturesUnordered` instead :)

protocols/identify/src/handler.rs
```diff
@@ -90,36 +128,81 @@ pub struct Handler {
...
 
-/// Identifying information of the local node that is pushed to a remote.
 #[derive(Debug)]
-pub struct Push(pub Info);
+#[allow(clippy::large_enum_variant)]
+pub enum InEvent {
+    /// Identifying information of the local node that is pushed to a remote.
+    Push(Info),
+    /// Identifying information requested from this node.
+    Identify(BehaviourInfo),
```
thomaseizinger: How different are these two really? Isn't exactly the same, just with different semantics?

protocols/identify/src/handler.rs
```diff
@@ -82,6 +114,12 @@ pub struct Handler {
         >; 4],
     >,
 
+    /// Streams awaiting `BehaviourInfo` to then send identify requests.
+    reply_streams: VecDeque<ReplySubstream<NegotiatedSubstream>>,
```
thomaseizinger: Is the abstraction of a `ReplySubstream` useful here? Would it be easier to just buffer the streams themselves?

thomaseizinger (approved):
Thanks! This looks good to me now, one more idea but happy to merge anytime!

protocols/identify/src/handler.rs
jxs: thanks for recalling this again
```
large size difference between variants
   --> protocols/identify/src/handler.rs:174:1
    |
174 | / pub enum InEvent {
175 | |     /// Identifying information of the local node that is pushed to a remote.
176 | |     Push(Info),
    | |     ---------- the largest variant contains at least 304 bytes
177 | |     /// Identifying information requested from this node.
178 | |     Identify(BehaviourInfo),
    | |     ----------------------- the second-largest variant contains at least 48 bytes
179 | | }
    | |_^ the entire enum is at least 304 bytes
    |
    = note: `-D clippy::large-enum-variant` implied by `-D warnings`
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#large_enum_variant
help: consider boxing the large fields to reduce the total size of the enum
    |
176 |     Push(Box<Info>),
    |          ~~~~~~~~~
```

protocols/identify/src/handler.rs
jxs: ah! How nice, thanks Thomas! :)

protocols/identify/src/handler.rs
jxs: Sorry Thomas, not following. What is the alternative? But the `send` method is implemented for `ReplySubstream`.

thomaseizinger: The alternative is inlining `send` to wherever it is called which I'd guess is only one place?

protocols/identify/src/handler.rs
thomaseizinger: I wasn't talking about the size difference actually.

Semantically, even for an identify push, all the handler needs to know are the updated protocols and listen addresses, which is represented as the `BehaviourInfo` struct. Can we model it as the same type?

protocols/identify/src/handler.rs
thomaseizinger: We don't really do this wrapping of a substream anywhere else in the codebase so it would be nice to align it here.

Constructing a `Framed` is what we typically do.

jxs: ah, right! Thanks! I was able to remove `ReplySubstream` but `send` is still being used in more places. 
We can go further and deprecate `Info` and just use the protobuf `Identify`, but probably do that on a subsequent PR?

protocols/identify/src/handler.rs
jxs: ah right! So I followed along and merge `Behaviour`'s `pending_push` into `requests` for better balance when fullfiling them. Previously `Push` requests would always have priority over `Identify` ptal Thomas.

thomaseizinger (approved):
Very nice, this actually went beyond what I thought could be simplified :)

protocols/identify/src/handler.rs
thomaseizinger: > ah, right! Thanks! I was able to remove `ReplySubstream` but `send` is still being used in more places. 
> We can go further and deprecate `Info` and just use the protobuf `Identify`, but probably do that on a subsequent PR?

Nah, I think that is fine. We usually don't use the protobuf struct directly which is good.

mxinden (commented):
Great simplifications along the main goal of the pull request. Thanks.

Needs a changelog entry. Otherwise good to go from my end.

jxs:
Updated with CHANGELOG.md entry

mxinden (approved):
:pray:

PR #3215, opened 2022-12-09
ci(caching): make caching more effective

thomaseizinger:
## Description

Currently, we create a new cache for each workflow run for each crate. That ends up blowing the maximum allowed cache size of 10GB and GitHub deletes the least-recently used cache again. Effectively, this means we don't have any caching.

This patch introduces a cache factory workflow that only runs on master and always _saves_ a new cache. The CI workflow run for pull-requests on the other hand only restore these caches but don't save them.

## Notes

Here is an example of a cache factory run: https://github.com/thomaseizinger/rust-libp2p/actions/runs/3672495214

With the new configuration, the `libp2p-core` job finishes in about 5 minutes: https://github.com/thomaseizinger/rust-libp2p/actions/runs/3672680646/jobs/6209084632.
With the old configuration, the same job takes about 13 minutes: https://github.com/libp2p/rust-libp2p/actions/runs/3651158690/jobs/6168039236

A good chunk of those 5 minutes is still used up by cargo semver-checks which [will hopefully improve further](https://github.com/obi1kenobi/cargo-semver-checks/issues/214) but overall, these are already massive savings in computing time :)

## Change checklist


.github/workflows/ci.yml
```diff
@@ -34,34 +34,33 @@ jobs:
           echo "version=${RUST_VERSION}" >> $GITHUB_OUTPUT
 
       - name: Install Rust ${{ steps.parse-msrv.outputs.version }} for MSRV check
-        uses: actions-rs/toolchain@16499b5e05bf2e26879000db0c1d13f7e13fa3af # v1.0.7
+        uses: dtolnay/rust-toolchain@e645b0cf01249a964ec099494d38d2da0f0b349f # master
```
mxinden: Why is the change to `dtolnay/rust-toolchain` needed?

mxinden (approved):
Thanks. Looking forward to the improved CI runtime.

.github/workflows/ci.yml
thomaseizinger: I experimented with different things, can change it back if you want. The `actions-rs` actions are somewhat unmaintained but by pinning the hash, we are not taking full advantage of the ergonomics of this action so happy to go back to the old one.

PR #3222, opened 2022-12-09
chore: Format code in doc comments

umgefahren:
## Description

I discovered there is an unstable option to format code in doc comments. Although the option is unstable the resulting changes seem very reasonable.

To reproduce:
1. Add file named `rustfmt.toml` to project root
2. Fill it with the following:
    ```toml
    unstable_features = true
    format_code_in_doc_comments = true
    ```
3. Run: `cargo fmt --all`

## Notes

As soon as this is stabilized, we should add it to our CI. Linking the PR here: rust-lang/rustfmt#3348

None.

## Open Questions

None.

## Change checklist


thomaseizinger:
> 1. Add file named `rustfmt.toml` to project root
> 2. Fill it with the following:
>     ```toml
>     unstable_features = true
>     format_code_in_doc_comments = true
>     ```

Unstable options are unavailable on stable Rust, right? You can hack around that by passing the options via the commandline last time I checked :)

We could do the same thing as we do for clippy and make a cargo alias for formatting.

See here for example: https://github.com/Restioson/xtra/blob/master/.cargo/config.toml

umgefahren:
> > 1. Add file named `rustfmt.toml` to project root
> > 2. Fill it with the following:
> >    ```toml
> >    unstable_features = true
> >    format_code_in_doc_comments = true
> >    ```
> 
> Unstable options are unavailable on stable Rust, right? You can hack around that by passing the options via the commandline last time I checked :)
> 
> We could do the same thing as we do for clippy and make a cargo alias for formatting.
> 
> See here for example: https://github.com/Restioson/xtra/blob/master/.cargo/config.toml

That's a good idea. I was looking for passing with options and probably overlooked it. I will add a custom cargo fmt command alias. Although it should be noted that this feature is not bug-free. Last time I looked it could case an ICE on empty doc comment code blocks.

.cargo/config.toml
```diff
@@ -1,3 +1,4 @@
 [alias]
 # Temporary solution to have clippy config in a single place until https://github.com/rust-lang/rust-clippy/blob/master/doc/roadmap-2021.md#lintstoml-configuration is shipped.
 custom-clippy = "clippy --workspace --all-features --all-targets -- -A clippy::type_complexity -A clippy::pedantic -D warnings"
+custom-fmt = "fmt -- --config format_code_in_doc_comments=true"
```
thomaseizinger, line 4: If we do this, I'd like to enforce it in CI as well.

thomaseizinger:
> That's a good idea. I was looking for passing with options and probably overlooked it. I will add a custom cargo fmt command alias. Although it should be noted that this feature is not bug-free. Last time I looked it could case an ICE on empty doc comment code blocks.

That is not ideal but I don't think it is a big issue. We can go back to regular rustfmt if it ever causes problems.

thomaseizinger:
It would be good to get some more opinions on this though. cc @libp2p/rust-libp2p-maintainers

jxs:
> fmt -- --config format_code_in_doc_comments=true

I think it makes sense! Can we also stage the `rustfmt.toml` so when we `cargo fmt` it runs with such options?

umgefahren:
> > fmt -- --config format_code_in_doc_comments=true
> 
> 
> 
> I think it makes sense! Can we also stage the `rustfmt.toml` so when we `cargo fmt` it runs with such options?

We could, but it would be a bit inconsistent with how we do things in this repo. Since we could could configure clippy via `clippy.toml` as well. But not a strong opinion.

thomaseizinger:
> > > fmt -- --config format_code_in_doc_comments=true
> > 
> > 
> > 
> > I think it makes sense! Can we also stage the `rustfmt.toml` so when we `cargo fmt` it runs with such options?
> 
> We could, but it would be a bit inconsistent with how we do things in this repo. Since we could could configure clippy via `clippy.toml` as well. But not a strong opinion. 

Configuring clippy lints via a config file is [unimplemented](https://github.com/rust-lang/rust-clippy/issues/6625) so we can't use it.

Activating unstable fmt options on stable Rust also doesn't work so we can't use it :)

PR #3224, opened 2022-12-10
feat(floodsub): make use of `prost-codec`

nloadholtes:
## Description

This patch addresses #2500 for the `libp2p-floodsub` crate.

For this PR the existing code was upgraded to use `Framed` with the `prost_codec::Codec` as the standard codec for handling the RPC message format serialization/deserialization.

## Notes

  - There are no pre-existing tests, so it is difficult to check if this is working correct. There were no errors observed locally.
  - The protocol name and message size were extracted into constants.
  - I'm still continuing my rust journey so suggestions/comments are welcomed.

#2500 

## Change checklist


nloadholtes:
I'd like to request a review to ensure the functionality is what is being looked for.

Once we are all on the same page I will update the change logs, Cargo.toml, etc.

protocols/floodsub/src/protocol.rs
```diff
@@ -135,13 +149,22 @@ where
...
-            socket.close().await?;
-
+            let mut framed = Framed::new(
+                socket,
+                prost_codec::Codec::<std::vec::Vec<u8>>::new(MAX_MESSAGE_LEN_BYTES),
+            );
+            framed
+                .send(bytes)
+                .await
+                .map_err(|e| io::Error::new(std::io::ErrorKind::ConnectionRefused, e))?;
```
thomaseizinger: What type is `e` here?

thomaseizinger (commented):
Looks good overall! Just one question, I am not quite happy with the error mapping there.

protocols/floodsub/src/protocol.rs
nloadholtes: Great question, and I'm not 100% sure. Its some kind of socket error, and I was wanting to use as general of a an error where as I could get away with. But I couldn't figure out how to get the compiler to accept a more "general" thing. VSCode kept offering suggestions of ConnectionRefused and other similar/likely errors (peer disconnected, etc.). I picked ConnectionRefused (and ConnectionReset below) knowing that we might be talking about them. ;)

I did try refactoring the `FloodsubDecodeError` to see if that could be expanded into more of a general "FloodsubError", but it really started to feel like scope creep on this ticket, at least in my opinion. I'd really like to do that approach (just feels more architecturally "correct" IMO), but I also want to keep this PR focused on the problem at hand.

I'm open to suggestions on all of this. (I also want to put some tests here, but first I need to crawl, then walk, then run :laughing: )

protocols/floodsub/src/protocol.rs
```diff
@@ -135,13 +149,22 @@ where
...
+    fn upgrade_outbound(self, socket: TSocket, _: Self::Info) -> Self::Future {
         Box::pin(async move {
             let bytes = self.into_bytes();
 
-            upgrade::write_length_prefixed(&mut socket, bytes).await?;
-            socket.close().await?;
-
+            let mut framed = Framed::new(
+                socket,
+                prost_codec::Codec::<std::vec::Vec<u8>>::new(MAX_MESSAGE_LEN_BYTES),
```
thomaseizinger: Feeding bytes into this somewhat defeats the purpose of using `Framed` and `Codec`.

Instead, you'd want to change `self.into_bytes` to return you the `rpc_proto::Rpc` type and feed that into `framed.send`.

protocols/floodsub/src/protocol.rs
thomaseizinger: I'd suggest we:

1. Create a type called `EncodeError` analogous to `DecodeError`.
2. Set `type Error = EncodeError` on line 149.
3. Remove these `map_err` calls.

protocols/floodsub/src/protocol.rs
thomaseizinger: You can then also remove this type hint, the compiler will figure it out!

```suggestion
                prost_codec::Codec::new(MAX_MESSAGE_LEN_BYTES),
```

thomaseizinger (commented):
Getting there! A few more comments :)

protocols/floodsub/src/protocol.rs
nloadholtes: Hmmm, I kept getting errors about the type after I removed it. I looked elsewhere in the code and [saw in indentify ](https://github.com/libp2p/rust-libp2p/blob/c1e68b75c0729102ee43e2ec3cedc643b4d1f324/protocols/identify/src/protocol.rs#L181-L190)there was a section of code using a similar pattern of adding a type hint when created its `FramedWrite`.

Also in that code is the calls to the `send()` and `write()` functions, but for some reason that code does not generate warning about unsued returns like my code does :thinking: I did see that `protocols/gossipsub/src/config.rs` uses this pattern, but there it makes sense: the code is just setting fields and doesn't care about the result. For my code, I don't know why the compiler suddenly cares about the return of those 2 calls.

protocols/floodsub/CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 - Update to `libp2p-swarm` `v0.42.0`.
 
+- Update to `prost-codec` `v0.3.0`.
+
+- Refactoring Floodsub to use common protobuf Codec. See [PR 3224].
```
thomaseizinger:
```suggestion
- Read and write protocols messages via `prost-codec`. See [PR 3224].
```

protocols/floodsub/src/protocol.rs
```diff
@@ -132,25 +149,26 @@ where
...
-
-            upgrade::write_length_prefixed(&mut socket, bytes).await?;
-            socket.close().await?;
-
+            let mut framed = Framed::new(
+                socket,
+                prost_codec::Codec::<rpc_proto::Rpc>::new(MAX_MESSAGE_LEN_BYTES),
+            );
+            let _ = framed.send(self.into_rpc()).await;
+            let _ = framed.close().await;
```
thomaseizinger:
```suggestion
            framed.send(self.into_rpc()).await?;
            framed.close().await?;
```

We should not ignore those errors :)

protocols/floodsub/src/protocol.rs
```diff
@@ -56,10 +61,18 @@ where
...
 
-    fn upgrade_inbound(self, mut socket: TSocket, _: Self::Info) -> Self::Future {
+    fn upgrade_inbound(self, socket: TSocket, _: Self::Info) -> Self::Future {
         Box::pin(async move {
-            let packet = upgrade::read_length_prefixed(&mut socket, 2048).await?;
-            let rpc = rpc_proto::Rpc::decode(&packet[..]).map_err(DecodeError)?;
+            let mut framed = Framed::<TSocket, prost_codec::Codec<rpc_proto::Rpc>>::new(
+                socket,
+                prost_codec::Codec::new(MAX_MESSAGE_LEN_BYTES),
+            );
```
thomaseizinger:
```suggestion
            let mut framed = Frame::new(
                socket,
                prost_codec::Codec::<rpc_proto::Rpc>::new(MAX_MESSAGE_LEN_BYTES),
            );
```

This should be sufficient.

thomaseizinger (changes requested):
Looks good, almost there :)

protocols/floodsub/src/protocol.rs
thomaseizinger: Right, my bad! It is needed :)

protocols/floodsub/src/protocol.rs
nloadholtes: Ahh! This just solved a mystery I ran into with the error structs/enums. I am refactoring some stuff :smile:

nloadholtes:
Ok, I untangled the error issue. 
* I renamed `FloodsubDecodeError` to the more general `FloodsubError`
* I added `prost_codec::Error` that was missing from the enum
* Re-ordered the fields to be alphabetical

Doing this allows the same enum to be used as the `type Error =` in both impl locations. This in turn lets the `let _ =` be removed and allows both the inbound and outbound to look a lot more similar and also match what is being done in other protocols. ;)

protocols/floodsub/src/protocol.rs
```diff
@@ -93,21 +106,24 @@ where
...
+    ProtobufError(#[from] EncodeDecodeError),
+    /// Error when reading the packet from the socket.
+    #[error("Failed to read from socket")]
+    ReadError(#[from] io::Error),
 }
 
 #[derive(thiserror::Error, Debug)]
 #[error(transparent)]
-pub struct DecodeError(prost::DecodeError);
+pub struct EncodeDecodeError(prost_codec::Error);
```
thomaseizinger:
```suggestion
pub struct CodecError(prost_codec::Error);
```

thomaseizinger (approved):
Just one naming suggestion, otherwise LGTM!

protocols/floodsub/src/protocol.rs
nloadholtes: My confusion: (In case I look at this PR again in the future)  the error message that adding the `?` generated was not clear to me. What it was asking was for the `Error` type to handle the output of the `?` and what I had there originally (just the struct) did not handle the prost_code:Error error types. So of course that error message now makes sense. The `type Error` needs be aware of any errors that could come out of the impl.

protocols/floodsub/src/protocol.rs
```diff
@@ -93,21 +106,24 @@ where
...
-    /// Error when reading the packet from the socket.
-    #[error("Failed to read from socket")]
-    ReadError(#[from] io::Error),
-    /// Error when decoding the raw buffer into a protobuf.
-    #[error("Failed to decode protobuf")]
-    ProtobufError(#[from] DecodeError),
+pub enum FloodsubError {
+    /// Error from prost_codec::Codec
+    #[error(transparent)]
+    Codec(#[from] prost_codec::Error),
```
thomaseizinger: This one should not exist! We want to hide `prost_codec::Error` from the public API of this crate!

protocols/floodsub/src/protocol.rs
nloadholtes: haha! I put that name down first, but changed my mind about 5 times before settling. I agree and will change it.

thomaseizinger (changes requested):

protocols/floodsub/src/protocol.rs
thomaseizinger: That is the reason for `CodecError` existing :)

nloadholtes: umm, I just got side tracked on this. While trying to fix it I caused an error ` the compiler unexpectedly panicked. this is a bug.` I'm going to go report that and then pick this up again tomorrow. Thanks for all your help on this @thomaseizinger!

thomaseizinger: > umm, I just got side tracked on this. While trying to fix it I caused an error ` the compiler unexpectedly panicked. this is a bug.` I'm going to go report that and then pick this up again tomorrow. Thanks for all your help on this @thomaseizinger!

`cargo clean` usually fixes these. I hit them too occasionally, often in combination with proc-macros.

nloadholtes: That did the trick! I cleaned up the commit message and pushed what I think is the corrected code.

thomaseizinger (dismissed):
Thanks!

thomaseizinger (dismissed):

thomaseizinger (approved):

mxinden (approved):
:rocket: Thanks @nloadholtes!

PR #3227, opened 2022-12-12
feat(noise): deprecate non-compliant implementation

thomaseizinger:
## Description

For the last [two years](https://github.com/libp2p/rust-libp2p/blob/master/CHANGELOG.md#version-0190-2020-05-18), we have been carrying around a non-compliant implementation of the noise protocol for libp2p. It is time to follow through on the announcement in that changelog entry and deprecate it. Users should use the spec compliant implementation instead.

1. This will reduce the maintenance effort of our codebase.
2. We will improve compile-times because we no longer need to depend on `libsodium` to test cryptography that is only part of the non-compliant implementation.
3. It will simplify usage of `rust-libp2p` because users cannot accidentally choose the wrong implementation.

## Change checklist


transports/noise/CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+# 0.41.1 [unreleased]
+
+- Deprecate non-compliant noise implementation. We intend to remove it in a future release without replacement. See [PR XXXX].
+
+[PR XXXX]: https://github.com/libp2p/rust-libp2p/pull/XXXX
```
mxinden:
```suggestion
- Deprecate non-compliant noise implementation. We intend to remove it in a future release without replacement. See [PR 3227].

[PR 3227]: https://github.com/libp2p/rust-libp2p/pull/3227
```

mxinden (approved):
Thanks @thomaseizinger. Good idea.

What do you think of deprecating `LegacyConfig` as well?

https://github.com/libp2p/rust-libp2p/blob/76abab9e20c29d03f787f39565de38dff42b0f80/transports/noise/src/lib.rs#L572-L585

mxinden:
> https://github.com/libp2p/rust-libp2p/blob/master/CHANGELOG.md#version-0190-2020-05-18

Great changelog spelunking!

thomaseizinger:
> Thanks @thomaseizinger. Good idea.
> 
> What do you think of deprecating `LegacyConfig` as well?
> 
> https://github.com/libp2p/rust-libp2p/blob/76abab9e20c29d03f787f39565de38dff42b0f80/transports/noise/src/lib.rs#L572-L585

Yep, that is a good one to remove too! I'll send a PR.

PR #3230, opened 2022-12-12
feat(kad): update republish interval and expiration time defaults

### LINKED ISSUE #3229: kad: Update republish and expiration interval to new spec
Update the `libp2p-kad` default values to the Kademlia specification revision 2 https://github.com/libp2p/specs/pull/451/, namely the republish interval and the expiration interval.

https://github.com/libp2p/rust-libp2p/blob/4fe518de86ecb0805421cca6e291db168c10eac1/protocols/kad/src/behaviour.rs#L182-L196

lidel:
## Description

This patch applies changes from https://github.com/libp2p/specs/pull/451. In particular, the new defaults are:

- Record Expiration: 48h
- Record Republish Interval: 22h

Closes #3229.

- https://github.com/libp2p/specs/pull/451
- #3229

## Change checklist


thomaseizinger (approved):

protocols/kad/src/behaviour.rs
```diff
@@ -185,12 +185,12 @@ impl Default for KademliaConfig {
             kbucket_pending_timeout: Duration::from_secs(60),
             query_config: QueryConfig::default(),
             protocol_config: Default::default(),
-            record_ttl: Some(Duration::from_secs(36 * 60 * 60)),
+            record_ttl: Some(Duration::from_secs(48 * 60 * 60)),
```
mxinden: Note @melekes and @kpp that this invalidates the assumptions in Substrate's Authority Discovery.

https://github.com/paritytech/substrate/blob/8d11cec9713a93b9f6bee1b29fde840da5ba56e0/client/authority-discovery/src/lib.rs#L88-L109

Either you go with the new defaults and update Authority Discovery, or you override the defaults of libp2p.

mxinden (commented):
@lidel thank you for tackling this. Very much appreciate you making sure implementations are consistent.

Can you add a changelog entry? Otherwise this is good to go from my end.

mxinden:
@thomaseizinger I added `!` to the title, given that this is a breaking change. Let me know in case I misunderstand the conventions of conventional commits.

thomaseizinger:
> @thomaseizinger I added `!` to the title, given that this is a breaking change. Let me know in case I misunderstand the conventions of conventional commits.

I guess this is semantically breaking yeah.

protocols/kad/src/behaviour.rs
melekes: Right. Thanks for mentioning 👍

mxinden:
@lidel friendly ping. Can you add a changelog entry? Let us know if you want us to take over.

guillaumemichel:
My first contribution here (:

Just added the changelog entry, let me know if anything else is required.

protocols/kad/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 0.46.0 - unreleased
```
thomaseizinger, line 1: Should this really be a breaking change? It is changing behaviour but arguably to better defaults!

thomaseizinger: If we want to release this as a breaking change, it will have to wait until we decide to cut another breaking release.

guillaumemichel: Increasing the default record TTL and reducing the reprovide interval won't really _break_ anything. It implies that there will be more records stored on the remote peers.

I don't known if this is _breaking enough_, but as the change has been pending for a year, it is fine to wait for the next breaking release 👍🏻

thomaseizinger: We could add a separate method that constructs a new instance of `Config` with the updated defaults? That way people could at least use the new ones but the defaults would not be broken.

guillaumemichel: I think the goal here is to update the default to match libp2p [specs](https://github.com/libp2p/specs/tree/master/kad-dht#content-provider-advertisement-and-discovery), that are backed by network [measurements](https://github.com/probe-lab/network-measurements/blob/master/results/rfm17-provider-record-liveness.md), rather than keeping these magic numbers.

So IMO we want to break the defaults. But this won't break anything in practice, because we increase the record TTL and decrease the reprovide interval. It would be more concerning if we were to increase the reprovide interval beyond the record TTL, or to decrease the record TTL below reprovide interval.

lidel: In my mind, this is a glorified increase in cache TTL :-)
Does not break anything, if anything it saves some work for clients, who don't have to reprovide that often.

thomaseizinger: What about Max's [comment](https://github.com/libp2p/rust-libp2p/pull/3230#discussion_r1051976550) above? He was of the opinion that it is a breaking change for substrate.

I'd rather not invalidate anybody's assumptions on these defaults. We can always remove the current defaults in the next breaking release thus I think just adding a new constructor with the updated defaults is much safer for now and we can instantly release that.

guillaumemichel: IMO the role of default is to _just work_. On that assumption, a network currently using the default and not caring further about the value as long as it _just works_ will not be broken, because these new values are better.

If a network is implementing custom values in its protocol, then it shouldn't depend on the default, and thus won't be broken either.

Besides, adding a new constructor would give the wrong information. Users will continue to use the default, whereas they should use the new constructor with better default values. As long as rust-libp2p doesn't change the default values, it won't implement the libp2p specs (which may be confusing).

> What about Max's [comment](https://github.com/libp2p/rust-libp2p/pull/3230#discussion_r1051976550) above? He was of the opinion that it is a breaking change for substrate.

The change has been [acknowledged](https://github.com/libp2p/rust-libp2p/pull/3230#discussion_r1055177364) by @melekes.

This change doesn't need to happen _now_, it can wait for the next breaking release if deemed necessary. However, I think that we eventually _need_ to implement the new libp2p specs defaults, even if it means updating the value in use in the networks depending on the defaults.

thomaseizinger: > This change doesn't need to happen _now_

It is already scheduled for the next breaking release but we don't have an ETA on that. I was simply suggesting a way of merging this sooner so it is at least available to users that would like to use the improved defaults today.

jxs (approved):

PR #3231, opened 2022-12-13
ci(mergify): dismiss approvals on push once `send-it` is applied

thomaseizinger:
## Description

Currently, a user can push code after it has been approved and the `send-it` label applied. We only want to merge code that we actually looked at. Use mergify to dismiss approvals in such circumstances.

## Change checklist


mxinden (approved):
:pray: thanks for the quick fix!

jxs (approved):

PR #3234, opened 2022-12-13
fix(kademlia): downgrade reusable stream dropping warning to debug log

nazar-pc:
## Description

There is nothing wrong about being near the edge of the concurrency Kademlia allows. If there was an older stream about to be reused, it doesn't mean there was anything wrong to warn about.

## Notes

We're trying to maximize concurrency allowed/configured and this generates a lot of unnecessary warnings.

## Change checklist

### DIFF
--- a/protocols/kad/src/handler.rs
+++ b/protocols/kad/src/handler.rs
@@ -580,7 +580,7 @@ where
                 )
             }) {
                 *s = InboundSubstreamState::Cancelled;
-                log::warn!(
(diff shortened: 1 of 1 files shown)


thomaseizinger (dismissed):
Thanks! I think this can merge without a changelog entry.

thomaseizinger (approved):

mxinden (approved):
Thanks @nazar-pc. Agreed.

PR #3214, opened 2022-12-08
refactor(dcutr): reshape public API to follow naming guidelines

thomaseizinger:
## Description

With this patch, the naming of types follows the guidelines discussed in #2217.

## Change checklist

- [ ] I have performed a self-review of my own code
- [ ] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] A changelog entry has been made in the appropriate crates

PR #3238, opened 2022-12-13
refactor(relay): revise public API to follow naming convention

jxs:
## Description

Continues addressing https://github.com/libp2p/rust-libp2p/issues/2217.

## Notes

Per commit review is suggested :)

## Change checklist


protocols/relay/CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # 0.15.0 [unreleased]
 
+- Rename types as per [discussion 2174].
+  `Relay` has been renamed to `Behaviour`.
+  The `Relay`, and `Client` prefixes have been removed from various types like `ClientTransport`.
+  Users should prefer importing the relay protocol as a module (`use libp2p::relay::v2::relay;`),
```
thomaseizinger: That is a bit weird, isn't it? Do we need to keep `v2` now that we don't even have `v1` anymore?

protocols/relay/CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # 0.15.0 [unreleased]
 
+- Rename types as per [discussion 2174].
```
thomaseizinger, line 3: This link is missing I think :)

thomaseizinger (commented):
Thanks, just one question :)

protocols/relay/CHANGELOG.md
jxs: indeed, thanks Thomas!

protocols/relay/CHANGELOG.md
jxs: now that you say so it indeed makes sense :smile:  what do you suggest Thomas?

thomaseizinger: We could just drop the entire `v2` module, i.e. move it all to the top layer of the crate so we get:

- `libp2p::relay::Behaviour`
- `libp2p::relay::client::Behaviour`

jxs: great suggestion, I think it's much better now and in line with the other protocols, ptal Thomas.

protocols/relay/src/client.rs
```diff
@@ -322,10 +324,10 @@ impl NetworkBehaviour for Client {
                 }
             }
             None => unreachable!(
-                "`Relay` `NetworkBehaviour` polled after channel from \
+                "`relay::Behaviour` polled after channel from \
                      `RelayTransport` has been closed. Unreachable under \
```
thomaseizinger: `RelayTransport` mentioned here.

protocols/relay/src/lib.rs
```diff
@@ -23,9 +23,37 @@
 
 #![cfg_attr(docsrs, feature(doc_cfg, doc_auto_cfg))]
 
+pub mod behaviour;
```
thomaseizinger:
```suggestion
mod behaviour;
```

protocols/relay/src/v2.rs
```diff
@@ -21,30 +21,149 @@
 //! Implementation of the [libp2p circuit relay v2
 //! specification](https://github.com/libp2p/specs/blob/master/relay/circuit-v2.md).
 
-#[allow(clippy::derive_partial_eq_without_eq)]
-mod message_proto {
-    include!(concat!(env!("OUT_DIR"), "/message_v2.pb.rs"));
+pub mod client {
```
thomaseizinger, line 24: All backwards-compatible 🔥

protocols/relay/src/lib.rs
```diff
@@ -23,9 +23,37 @@
...
+pub mod client;
```
thomaseizinger: Having entire modules public has types easily slip into the public API without notice. What do you think of making this private, naming it something else and making an inline module that reexports the types we want in the public API?

thomaseizinger (commented):
Very nice! Thank you!

protocols/relay/src/lib.rs
jxs: thanks, in the process I made `CircuitId` pub on `relay::CircuitId` as it was already `pub`

protocols/relay/src/client.rs
jxs: yeah nice catch, updated!

protocols/relay/src/lib.rs
jxs: yeah I was thinking about it during the afternoon while updating this, `pub` modules with `pub` types vs private modules with `pub` types then explicitly re-exported, acknowledging that the former is more slippery. Let's give it a try, updated! Ptal Thomas.

protocols/relay/src/lib.rs
```diff
@@ -23,9 +23,48 @@
...
+mod message_proto {
+    include!(concat!(env!("OUT_DIR"), "/message_v2.pb.rs"));
+}
+
+pub use behaviour::{Behaviour, CircuitId, Config, Event};
+pub use protocol::{
+    inbound_hop::FatalUpgradeError as InboundHopFatalUpgradeError,
+    inbound_stop::FatalUpgradeError as InboundStopFatalUpgradeError,
+    outbound_hop::FatalUpgradeError as OutboundHopFatalUpgradeError,
+    outbound_stop::FatalUpgradeError as OutboundStopFatalUpgradeError, HOP_PROTOCOL_NAME,
```
thomaseizinger: These will cause problems with `cargo semver-checks` unfortunately. See https://github.com/libp2p/rust-libp2p/pull/3213/files for details!

My recommendation would be to do something like https://github.com/libp2p/rust-libp2p/pull/3214 where we create dedicated `inbound_hop` etc modules.

protocols/relay/src/lib.rs
```diff
@@ -23,9 +23,48 @@
...
+    STOP_PROTOCOL_NAME,
+};
+
+// explicitly pub'ed client types.
+pub mod client {
+    pub use crate::priv_client::{Behaviour, Event, RelayedConnection, Transport};
+
+    pub mod transport {
+        pub use crate::priv_client::transport::{
+            Error, Listener, Reservation, ToListenerMsg, TransportToBehaviourMsg,
```
thomaseizinger: Do all of these have to be public for the user to interact with `libp2p-relay`?

protocols/relay/src/lib.rs
```diff
@@ -23,9 +23,48 @@
...
+pub use behaviour::{Behaviour, CircuitId, Config, Event};
+pub use protocol::{
+    inbound_hop::FatalUpgradeError as InboundHopFatalUpgradeError,
+    inbound_stop::FatalUpgradeError as InboundStopFatalUpgradeError,
+    outbound_hop::FatalUpgradeError as OutboundHopFatalUpgradeError,
+    outbound_stop::FatalUpgradeError as OutboundStopFatalUpgradeError, HOP_PROTOCOL_NAME,
+    STOP_PROTOCOL_NAME,
+};
+
+// explicitly pub'ed client types.
```
thomaseizinger: Let's upgrade this to a rustdoc comment, then it is also useful in the documentation :)

Something like:

```suggestion
/// Everything related to the relay protocol from a client's perspective.
```

protocols/relay/src/priv_client.rs
```diff
@@ -100,12 +104,10 @@ pub struct Client {
...
 }
 
-impl Client {
-    pub fn new_transport_and_behaviour(
-        local_peer_id: PeerId,
-    ) -> (transport::ClientTransport, Self) {
-        let (transport, from_transport) = transport::ClientTransport::new();
-        let behaviour = Client {
+impl Behaviour {
+    pub fn new_transport_and_behaviour(local_peer_id: PeerId) -> (transport::Transport, Self) {
```
thomaseizinger, line 120: Constructors that return a tuple are always a bit odd to interact with. If we look at the ecosystem (channel libraries f.e.), we can see that these are often handled slightly differently. In particular, creating channels happens via free-functions. For example:

```rust
let (sender, receiver) = futures::channel::mpsc::bounded(1);
```

We could do the same thing here:

```rust
let (behaviour, transport) = libp2p::relay::client::new(peer_id);
```

thomaseizinger (commented):
Thanks for pushing this forward, some more thoughts!

protocols/relay/src/lib.rs
thomaseizinger: Nice, I feel much better about this. Should make our own life easier going forward!

protocols/relay/src/priv_client.rs
jxs: yeah, agreed. Updated!

protocols/relay/src/lib.rs
jxs: Thanks, updated.

protocols/relay/src/lib.rs
jxs: Good question Thomas, I am not yet familiar with the `relay` and `dcutr` protocols. But taking into consideration that `dcutr` uses `relay` and doesn't use any of the `transport::` types, I'd wonder if they are required to be public at all? 
What if we deprecate them and see if anyone complains? CC @mxinden

thomaseizinger: - When these protocols were merged, `Listener` was still in the public API of `Transport` but it no longer is.
- I wonder if the messages show up in some other public API.
- I think it makes sense for the error to be exposed.

Deprecating them with a notice that they will be private in the future sounds good to me!

jxs: ok nice, updated! Ptal Thomas if you agree with the wording.

protocols/relay/src/v2.rs
```diff
@@ -42,48 +42,53 @@ pub mod client {
 
         #[deprecated(
             since = "0.15.0",
-            note = "Use libp2p_relay::client::Transport instead."
+            note = "ClientTransport has been renamed to Transport
+            and is now private as it shouldn't be required by end users."
```
thomaseizinger: I think they still need to interact with it right? We are returning it from `libp2p::relay::client::new`. Users should be allowed to name that one if they want.

We _could_ return an `impl Transport` actually but that is a bit more constraining. Esp. when users compose those transports together, they would lose the ability to name the type unless they use `.boxed`. In reality though, that is probably what they should do. Naming the entire chain of `.or`'d transports is super clunky.

Returning `impl Transport` would be efficient (no boxing) and would hide the type. I think that might be a good idea but I'd like to hear @mxinden's opinion on it!

protocols/relay/src/v2.rs
```diff
@@ -42,48 +42,53 @@ pub mod client {
...
         )]
-        pub type ClientTransport = crate::client::Transport;
+        pub type ClientTransport = crate::priv_client::transport::Transport;
 
         #[deprecated(
             since = "0.15.0",
-            note = "Use libp2p_relay::client::transport::Listener instead."
+            note = "RelayListener has been renamed to Listener
+            and is now private as it shouldn't be required by end users."
```
thomaseizinger:
```suggestion
            note = "RelayListener will become crate-private in the future as it shouldn't be required by end users."
```

I don't think we need to tell them about the rename :)

protocols/relay/src/v2.rs
```diff
@@ -42,48 +42,53 @@ pub mod client {
...
         )]
-        pub type RelayListener = crate::client::transport::Listener;
+        pub type RelayListener = crate::priv_client::transport::Listener;
 
         #[deprecated(
             since = "0.15.0",
-            note = "Use libp2p_relay::transport::RelayedDial instead."
+            note = "RelayedDial is now private as it shouldn't be required by end users."
```
thomaseizinger: `RelayedDial` is part of the public API in any case because of `Transport::Dial`. However, it is itself a type alias that we can simply remove (in the future). We should deprecate this and tell users to simply write out the type themselves.

protocols/relay/src/v2.rs
```diff
@@ -42,48 +42,53 @@ pub mod client {
...
             'static,
-            Result<crate::client::RelayedConnection, crate::client::transport::Error>,
+            Result<crate::client::RelayedConnection, crate::priv_client::transport::Error>,
         >;
 
         #[deprecated(
             since = "0.15.0",
-            note = "Use libp2p_relay::client::transport::Error instead."
+            note = "RelayError has been renamed to Error
+            and is now private as it shouldn't be required by end users."
```
thomaseizinger: I'd say the error needs to remain public because - in theory - someone polling the `Transport` manually needs to be able to name it.

protocols/relay/src/v2.rs
```diff
@@ -42,48 +42,53 @@ pub mod client {
...
         )]
-        pub type RelayError = crate::client::transport::Error;
+        pub type RelayError = crate::priv_client::transport::Error;
 
         #[deprecated(
             since = "0.15.0",
-            note = "Use libp2p_relay::client::transport::TransportToBehaviourMsg instead."
+            note = "TransportToBehaviourMsg is now private
+            as it shouldn't be required by end users."
```
thomaseizinger:
```suggestion
            note = "TransportToBehaviourMsg will become crate-private in the future as it shouldn't be required by end users."
```

protocols/relay/src/v2.rs
```diff
@@ -42,48 +42,53 @@ pub mod client {
...
         )]
-        pub type TransportToBehaviourMsg = crate::client::transport::TransportToBehaviourMsg;
+        pub type TransportToBehaviourMsg = crate::priv_client::transport::TransportToBehaviourMsg;
 
         #[deprecated(
             since = "0.15.0",
-            note = "Use libp2p_relay::client::transport::ToListenerMsg instead."
+            note = "ToListenerMsg is now private
+            as it shouldn't be required by end users."
```
thomaseizinger:
```suggestion
            note = "ToListenerMsg will become crate-private in the future as it shouldn't be required by end users."
```

protocols/relay/src/v2.rs
```diff
@@ -42,48 +42,53 @@ pub mod client {
...
         )]
-        pub type ToListenerMsg = crate::client::transport::ToListenerMsg;
+        pub type ToListenerMsg = crate::priv_client::transport::ToListenerMsg;
 
         #[deprecated(
             since = "0.15.0",
-            note = "Use libp2p_relay::client::transport::Reservation instead."
+            note = "Reservation is now private as it shouldn't be required by end users."
```
thomaseizinger:
```suggestion
            note = "Reservation will become crate-private in the future as it shouldn't be required by end users."
```

protocols/relay/src/v2.rs
thomaseizinger: Changing to `impl Trait` now would be a breaking change so let's not do it and stick to doing just the rename!

thomaseizinger (approved):
Thanks! A few minor suggestions in regards to the deprecation naming, otherwise ready :)

protocols/dcutr/tests/lib.rs
```diff
@@ -116,7 +116,7 @@ fn build_client() -> Swarm<Client> {
     let local_public_key = local_key.public();
     let local_peer_id = local_public_key.to_peer_id();
 
-    let (relay_transport, behaviour) = client::Client::new_transport_and_behaviour(local_peer_id);
+    let (relay_transport, behaviour) = client::new(local_peer_id);
```
thomaseizinger:
```suggestion
    let (relay_transport, behaviour) = relay::client::new(local_peer_id);
```

Maybe? :)

protocols/dcutr/tests/lib.rs
```diff
@@ -153,7 +153,7 @@ where
     prelude = "libp2p_swarm::derive_prelude"
 )]
 struct Client {
-    relay: client::Client,
+    relay: client::Behaviour,
```
thomaseizinger:
```suggestion
    relay: relay::client::Behaviour,
```

Here too?

protocols/relay/src/lib.rs
```diff
@@ -23,9 +23,56 @@
...
+        pub use crate::protocol::outbound_hop::FatalUpgradeError;
+    }
+    pub mod stop {
+        pub use crate::protocol::outbound_stop::FatalUpgradeError;
+    }
+}
+
+/// Everything related to the relay protocol from a client's perspective.
+pub mod client {
+    pub use crate::priv_client::{new, Behaviour, Event, RelayedConnection, Transport};
```
thomaseizinger: Uh I just had a closer look at `RelayedConnection` and saw that this one is a public enum with fairly complex variants! (An absolute semver hazard!) I think we should encapsulate that one into an opaque type that only exposes `AsyncRead` and `AsyncWrite` as its interface, i.e.:

- Rename the type to `RelayedConnectionState` & make it private
- Create a struct `RelayedConnection` (or maybe just `Connection` to stick with the naming convention?)
- Move the `AsyncRead` and `AsyncWrite` implementations from the enum to the struct

This will allow us in the future to change the layout of this enum without making a breaking change.

This is a breaking change now though but I'll happily take that for the benefit of sealing the API.

thomaseizinger: This could happen in a separate PR though, perhaps in this one, we just rename it to `Connection` using a type-alias?

protocols/relay/src/v2.rs
jxs: yeah makes sense, updated!

protocols/relay/src/v2.rs
jxs: updated!

protocols/relay/src/v2.rs
jxs: updated!

protocols/relay/src/v2.rs
jxs: updated!

protocols/relay/src/v2.rs
jxs: right, makes sense. This is interesting, and kinda relates to https://github.com/libp2p/rust-libp2p/pull/3159#discussion_r1043895628 how come `Transport` is public, implements `libp2p_core::Transport` and compiler doesn't complain that `Error` needs to be public?

protocols/relay/src/v2.rs
jxs: yeah, makes sense. Thanks!

protocols/relay/src/lib.rs
jxs: Yeah makes sense, created https://github.com/libp2p/rust-libp2p/issues/3255 Btw, there are more `Relayed` prefixed types, should we also address them?

protocols/dcutr/tests/lib.rs
jxs: yeah makes sense, updated!

protocols/dcutr/tests/lib.rs
jxs: yeah I think it makes sense across the whole file tbh. Updated!

protocols/relay/src/v2.rs
jxs: yeah right, I had forgotten that, thanks Thomas! Thou we probably can use the opaque type described on other cases.

protocols/relay/src/v2.rs
thomaseizinger: I think you'll find that making the type declaration itself private is not going to work but making the entire module private and exporting select items from it does work!

protocols/relay/src/lib.rs
thomaseizinger: > Yeah makes sense, created https://github.com/libp2p/rust-libp2p/issues/3255 Btw, there are more `Relayed` prefixed types, should we also address them?

Where it makes sense, sure. I think there is a `RelayedMultiaddr`, not sure if we can rename that to something better. As long as it is private, I don't really mind :)

jxs: yeah apart from that only `RelayedDial` was left, renamed it to `Dial`.

thomaseizinger (approved):

protocols/relay/CHANGELOG.md
```diff
@@ -1,9 +1,19 @@
...
+  the `v2` namespace has also been removed, users should prefer importing the relay protocol as a module (`use libp2p::relay;`),
+  and refer to its types via `relay::`. For example: `relay::Behaviour` or `relay::client::Behaviour`.
+  See [PR 3235].
+
 - Update to `libp2p-core` `v0.39.0`.
 
 - Update to `libp2p-swarm` `v0.42.0`.
 
+[PR 3235]: https://github.com/libp2p/rust-libp2p/pull/3235
+[discussion 2174]: https://github.com/libp2p/rust-libp2p/issues/2174
```
thomaseizinger:
```suggestion
  See [PR 3238].

- Update to `libp2p-core` `v0.39.0`.

- Update to `libp2p-swarm` `v0.42.0`.

[PR 3238]: https://github.com/libp2p/rust-libp2p/pull/3238
[discussion 2174]: https://github.com/libp2p/rust-libp2p/issues/2174
```

protocols/dcutr/examples/dcutr.rs
```diff
@@ -32,7 +32,7 @@ use libp2p_dns::DnsConfig;
 use libp2p_identify as identify;
 use libp2p_noise as noise;
 use libp2p_ping as ping;
-use libp2p_relay::v2::client::{self, Client};
+use libp2p_relay::client;
```
thomaseizinger:
```suggestion
use libp2p_relay as relay;
```

protocols/relay/src/priv_client.rs
```diff
@@ -46,7 +46,9 @@ use std::ops::DerefMut;
 use std::pin::Pin;
 use std::task::{Context, Poll};
 
-/// The events produced by the [`Client`] behaviour.
+pub use transport::Transport;
```
thomaseizinger: We can remove the `pub` here in favor of accessing the `Transport` directly in `lib.rs` for the re-export.

protocols/relay/src/priv_client.rs
```diff
@@ -100,18 +104,22 @@ pub struct Client {
...
-        let behaviour = Client {
-            local_peer_id,
-            from_transport,
-            directly_connected_peers: Default::default(),
-            queued_actions: Default::default(),
-        };
-        (transport, behaviour)
+/// Create a new client relay [`Behaviour`] with it's corresponding [`Transport`].
+pub fn new(local_peer_id: PeerId) -> (transport::Transport, Behaviour) {
+    let (transport, from_transport) = transport::Transport::new();
```
thomaseizinger:
```suggestion
pub fn new(local_peer_id: PeerId) -> (Transport, Behaviour) {
    let (transport, from_transport) = Transport::new();
```

Don't think the qualification is necessary! :)

protocols/relay/src/priv_client/transport.rs
```diff
@@ -50,9 +50,9 @@ use thiserror::Error;
 ///    # use libp2p_core::{Multiaddr, multiaddr::{Protocol}, Transport, PeerId};
 ///    # use libp2p_core::transport::memory::MemoryTransport;
 ///    # use libp2p_core::transport::choice::OrTransport;
-///    # use libp2p_relay::v2::client;
+///    # use libp2p_relay::client;
 ///    let actual_transport = MemoryTransport::default();
-///    let (relay_transport, behaviour) = client::Client::new_transport_and_behaviour(
+///    let (relay_transport, behaviour) = client::Behaviour::new_transport_and_behaviour(
```
thomaseizinger: This is a deprecated API now right? Let's not use deprecated APIs in examples.

protocols/relay/src/priv_client/transport.rs
```diff
@@ -50,9 +50,9 @@ use thiserror::Error;
 ///    # use libp2p_core::{Multiaddr, multiaddr::{Protocol}, Transport, PeerId};
 ///    # use libp2p_core::transport::memory::MemoryTransport;
 ///    # use libp2p_core::transport::choice::OrTransport;
-///    # use libp2p_relay::v2::client;
+///    # use libp2p_relay::client;
```
thomaseizinger:
```suggestion
///    # use libp2p_relay as relay;
```

Let's be consistent with our other examples.

protocols/relay/src/priv_client/transport.rs
```diff
@@ -114,7 +114,7 @@ impl ClientTransport {
     /// ```
```
thomaseizinger: Whilst we are cleaning up, let's remove this example. It is a crate-private function so users are not going to see it in the docs, see the current ones here: https://libp2p.github.io/rust-libp2p/libp2p_relay/v2/client/transport/struct.ClientTransport.html

protocols/relay/src/priv_client/transport.rs
```diff
@@ -398,17 +396,17 @@ impl Stream for RelayListener {
...
                 }
-                ToListenerMsg::Reservation(Err(())) => self.close(Err(RelayError::Reservation)),
+                ToListenerMsg::Reservation(Err(())) => self.close(Err(Error::Reservation)),
             };
         }
     }
 }
 
-pub type RelayedDial = BoxFuture<'static, Result<RelayedConnection, RelayError>>;
+pub type Dial = BoxFuture<'static, Result<Connection, Error>>;
```
thomaseizinger: We are already telling users to create their own type-alias if needed in the now deprecated `v2 module. Let's not make this public then.

Actually, we might just get rid of it entirely or does it have much value within our codebase?

protocols/relay/tests/lib.rs
```diff
@@ -31,8 +31,8 @@ use libp2p_core::PublicKey;
 use libp2p_core::{identity, upgrade, PeerId};
 use libp2p_ping as ping;
 use libp2p_plaintext::PlainText2Config;
-use libp2p_relay::v2::client;
-use libp2p_relay::v2::relay;
+use libp2p_relay as relay;
+use libp2p_relay::client;
```
thomaseizinger:
```suggestion
```

Let's not do this and refer to all types via the `relay` module, i.e. `relay::client`.

thomaseizinger (commented):
Sorry to add some more comments, I've had another thorough look and I think we can polish this a bit more :)

protocols/relay/tests/lib.rs
jxs: yeah makes sense! Updated

protocols/relay/src/priv_client/transport.rs
jxs: no, it's only once on L130. Removed the type alias.

protocols/relay/src/priv_client/transport.rs
jxs: yup agree, updated!

protocols/relay/src/priv_client/transport.rs
jxs: Thanks Thomas :bow: (I was expecting the compiler to catch the ones like this though :sweat_smile:)

protocols/relay/src/priv_client/transport.rs
jxs: yeah agreed! Updated :)

protocols/relay/src/priv_client.rs
jxs: and you are right, thanks! Updated

protocols/relay/src/priv_client.rs
jxs: yeah, this one was left dangling. Thanks!

protocols/relay/CHANGELOG.md
jxs: omg wth :facepalm: Thanks Thomas for catching this one! :bow:

protocols/dcutr/examples/dcutr.rs
jxs: addressed!

jxs:
> Sorry to add some more comments, I've had another thorough look and I think we can polish this a bit more :)

no, thank you for doing it so we caught things like the wrong PR link in the changelog!

protocols/relay/src/priv_client/transport.rs
thomaseizinger: Haven't tested it but I think you need to run rustdoc for that and check the output for warnings! Perhaps we should enforce no warnings in general when generating docs?

protocols/relay/src/lib.rs
```diff
@@ -23,9 +23,60 @@
...
+
+/// The ID of an outgoing / incoming, relay / destination request.
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Hash)]
+pub struct RequestId(u64);
+
+impl RequestId {
+    fn new() -> RequestId {
+        RequestId(rand::random())
+    }
+}
```
thomaseizinger: Just checked and this doesn't actually need to be public :)

In fact, this type isn't used for anything! The only place where the fields of `TransportToBehaviourMsg::DialReq` are read is here: https://github.com/jxs/rust-libp2p/blob/7f653b3676d719040cd8ffc39a1db47c9021b1e6/protocols/relay/src/priv_client.rs#L293-L299

and that instance ignores the `RequestId` :)

Just tried to delete it and everything still works :D

Here is my suggestion:

- Don't move the type from the v2 module but leave it there and deprecate it entirely with the message that it will be removed from the public API. I doubt that anyone will ever see that message but let's be correct. Next to type, make a TODO that we should delete this once we resolve the deprecations!

protocols/relay/src/priv_client/transport.rs
```diff
@@ -72,11 +72,11 @@ use thiserror::Error;
 ///    # use libp2p_core::{Multiaddr, multiaddr::{Protocol}, Transport, PeerId};
 ///    # use libp2p_core::transport::memory::MemoryTransport;
 ///    # use libp2p_core::transport::choice::OrTransport;
-///    # use libp2p_relay::v2::client;
+///    # use libp2p_relay::client;
 ///    # let relay_id = PeerId::random();
 ///    # let local_peer_id = PeerId::random();
 ///    let actual_transport = MemoryTransport::default();
-///    let (relay_transport, behaviour) = client::Client::new_transport_and_behaviour(
+///    let (relay_transport, behaviour) = client::Behaviour::new_transport_and_behaviour(
```
thomaseizinger: This still uses deprecated APIs :)

protocols/relay/tests/lib.rs
```diff
@@ -378,18 +377,18 @@ impl From<ping::Event> for RelayEvent {
...
     Ping(ping::Event),
 }
 
-impl From<client::Event> for ClientEvent {
-    fn from(event: client::Event) -> Self {
+impl From<relay::client::Event> for ClientEvent {
+    fn from(event: relay::client::Event) -> Self {
         ClientEvent::Relay(event)
     }
 }
```
thomaseizinger: If you are feeling like deleting a bit more code, we could turn on `event_process = true` (which is the default) and remove this. Can easily be done in a follow-up though.

thomaseizinger (commented):
Two more things! :)

protocols/relay/src/priv_client/transport.rs
jxs: yeah, that makes sense.

protocols/relay/src/lib.rs
mxinden: Thanks @jxs and @thomaseizinger. Deprecation sounds good to me.

protocols/relay/src/v2.rs
mxinden: Interesting idea on using `impl Transport`. Rename is fine by me (for now).

protocols/relay/tests/lib.rs
jxs: ehe yeah done. Also deleted the `ClientEvent` and `RelayEvent` that's what you were meaning right? And also removed from the examples.

protocols/relay/src/priv_client/transport.rs
jxs: pff thanks Thomas :bow: .

protocols/relay/src/lib.rs
jxs: great, addressed :)

protocols/relay/tests/lib.rs
thomaseizinger: Yes!

thomaseizinger (approved):
Great improvement :)

Can you open an issue to investigate how we can have the CI pipeline complain about deprecated APIs in docs?

jxs:
> Great improvement :)
> 
> Can you open an issue to investigate how we can have the CI pipeline complain about deprecated APIs in docs?

opened https://github.com/libp2p/rust-libp2p/issues/3282

mxinden (approved):
Appreciate the work here!

Issue #3240, opened 2022-12-14
Infer type parameters on `NetworkBehaviourAction`


thomaseizinger:
## Description

Currently, we expose multiple type parameters on `NetworkBehaviourAction` despite all of them being able to be inferred from the corresponding `NetworkBehaviour` type. However, we do also expose individual mapping functions for the in-event etc.

We could consolidate all those type parameters into one but that would require us to put a bound onto the struct definition. Trait bounds should ideally always only be applied to _functionality_ (i.e. impl blocks). Applying them to struct or enum definition easily creeps through into layers where you don't actually want them.

My suggestion would be to remove all trait bounds from `NetworkBehaviourAction` and expose the types as they are. For the common case where the types are unmapped, we can expose a type alias that infers all of these types from a type that implements `NetworkBehaviour`.

## Are you planning to do it yourself in a pull request?

Yes.

PR #3241, opened 2022-12-14
refactor(autonat): buffer entire actions instead of events

thomaseizinger:
## Description

Previously, we used to buffer events separately and emit actions directly. That is unnecessary. We can have a single place where we return from the `poll` loop and shove all actions into the same buffer.

## Notes

I originally created this because I thought it would fix an issue where I needed to supply type hints as part of implementing https://github.com/libp2p/rust-libp2p/issues/3240. That didn't turn out to be true but I still think that this change is a slight improvement. Feel free to close if you disagree :)

## Change checklist

### DIFF
--- a/protocols/autonat/src/behaviour.rs
+++ b/protocols/autonat/src/behaviour.rs
@@ -209,7 +209,12 @@ pub struct Behaviour {
 
     last_probe: Option<Instant>,
 
-    pending_out_events: VecDeque<<Self as NetworkBehaviour>::OutEvent>,
+    pending_actions: VecDeque<
+        NetworkBehaviourAction<
+            <Self as NetworkBehaviour>::OutEvent,
+            <Self as NetworkBehaviour>::ConnectionHandler,
+        >,
+    >,
 
     probe_id: ProbeId,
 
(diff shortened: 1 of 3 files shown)


thomaseizinger:
> I originally created this because I thought it would fix an issue where I needed to supply type hints as part of implementing #3240. That didn't turn out to be true but I still think that this change is a slight improvement. Feel free to close if you disagree :)

Turns out the issue I am running in might be a rustc bug: https://github.com/rust-lang/rust/issues/105680

protocols/autonat/src/behaviour.rs
```diff
@@ -431,14 +440,13 @@ impl NetworkBehaviour for Behaviour {
...
-                return Poll::Ready(NetworkBehaviourAction::GenerateEvent(event));
+            if let Some(event) = self.pending_actions.pop_front() {
+                return Poll::Ready(event);
             }
 
-            let mut is_inner_pending = false;
             match self.inner.poll(cx, params) {
                 Poll::Ready(NetworkBehaviourAction::GenerateEvent(event)) => {
-                    let (mut events, action) = match event {
+                    let (events, action) = match event {
```
elenaf9: Could we also change `HandleInnerEvent::handle_event` to directly return a `VecDeque<Action>` instead of this tuple?

thomaseizinger: Good idea, done!

elenaf9 (approved):
:+1:  LGTM

Issue #3244, opened 2022-12-14
identify/: Don't fail on unknown multiaddr protocol


mxinden:
## Summary

`libp2p-identify` currently discards the entire identify payload of a remote peer, if that payload contains a multiaddr it can not parse.

## Expected behaviour

When parsing an identify payload of a remote peer with an unknown multiaddr protocol (e.g. `quic-v1/` or `webrtc/`) log the parsing error, but don't discard the entire payload.

## Actual behaviour

In `libp2p-identify`'s parsing logic, it returns an error for the entire payload, in case it can not parse a multiaddr.

https://github.com/libp2p/rust-libp2p/blob/be3ec6c62b05b37f85fd60ce1a916f33235522af/protocols/identify/src/protocol.rs#L220-L226

This is especially relevant when rolling out a new protocol to a live network. Say that most nodes of a network run on an implementation version v1. Say that the `multiaddr` implementation is not aware of the `webrtc/` protocol. Say that a new version (v2) is rolled out to the network with support for the `webrtc/` protocol, listening via `webrtc/` by default. In such case all v1 nodes would discard all identify payloads of v2 nodes, given that the v2 identify payloads would contain the `webrtc/` protocol in their `listen_addr` addresses.

//CC @dignifiedquire for the role out of `quic-v1/` and `webtransport/` on IPFS
//CC @melekes for the role out of `webrtc/` and `quic-v1/` on Polkadot
//CC @divagant-martian and @AgeManning in case you experiment with `quic-v1/` on lighthouse

## Possible Solution

Release patch versions of `libp2p-identify` which log the parsing failure, but don't discard the entire identify payload.

## Version
`libp2p` `<=v0.50.0`

## Misc

- This was already discovered by @elenaf9 on https://github.com/libp2p/punchr/pull/64 before. I missed that this hard a larger impact beyond punchr.
- Ideally our Testground tests would catch issues like this in the future. Given that our current Testground tests run the `libp2p-ping` protocol only, they have not caught it.

## Would you like to work on fixing this bug?

Yes, unless someone else has capacity to do so in the next couple of days.

rkuhn:
In addition to this fix — which I fully agree with — I usually add a FutureCompat variant to all enums representing protocol elements that may be extended in the future (which is basically every enum that gets (de)serialised, unless Moses literally brought the protocol with him from Mt Sinaï). In this particular case, that variant could even contain the string value naming the locally unknown protocol.

[...]

PR #3246, opened 2022-12-14
fix(identify): Skip invalid multiaddr in listen_addrs

mxinden:
## Description

With this commit `libp2p-identify` no longer discards the whole identify payload in case a listen addr of the remote node is invalid, but instead logs the failure, skips the invalid multiaddr and parses the remaining identify payload.

This is especially relevant when rolling out a new protocol to a live network. Say that most nodes of a network run on an implementation version v1. Say that the `multiaddr` implementation is not aware of the `webrtc/` protocol. Say that a new version (v2) is rolled out to the network with support for the `webrtc/` protocol, listening via `webrtc/` by default. In such case all v1 nodes would discard all identify payloads of v2 nodes, given that the v2 identify payloads would contain the `webrtc/` protocol in their `listen_addr` addresses.

See https://github.com/libp2p/rust-libp2p/issues/3244 for details.

## Notes

- This pull request targets the `v0.50` branch such that we can release a patch version of `libp2p-identify`.
- I suggest we do the same for the `v0.49` release family.
- We can then backport the patch to `master` to be included in `v0.51`.

https://github.com/libp2p/rust-libp2p/issues/3244

## Change checklist

### DIFF
--- a/protocols/identify/src/protocol.rs
+++ b/protocols/identify/src/protocol.rs
@@ -26,7 +26,7 @@ use libp2p_core::{
     upgrade::{InboundUpgrade, OutboundUpgrade, UpgradeInfo},
     Multiaddr, PublicKey,
 };
-use log::trace;
+use log::{debug, trace};
 use std::convert::TryFrom;
 use std::{fmt, io, iter, pin::Pin};
 use thiserror::Error;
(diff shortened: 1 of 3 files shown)


protocols/identify/src/protocol.rs
```diff
@@ -236,14 +236,25 @@ impl TryFrom<structs_proto::Identify> for Info {
...
         };
 
         let public_key = PublicKey::from_protobuf_encoding(&msg.public_key.unwrap_or_default())?;
 
-        let observed_addr = parse_multiaddr(msg.observed_addr.unwrap_or_default())?;
+        let observed_addr = match parse_multiaddr(msg.observed_addr.unwrap_or_default()) {
+            Ok(a) => a,
+            Err(e) => {
+                debug!("Unable to parse multiaddr: {e:?}");
+                Multiaddr::empty()
```
mxinden, line 255: This is not ideal but in my eyes not worse than the status quo. Downstream code has to handle a `Multiaddr::empty` due to the `unwrap_or_default` anyways.

One might first think that a remote could never return an observed address, i.e. an address that it observes the local node as, that the local node doesn't know. One would expect that the local node supports the multiaddr to all all its supported transport protocols. One exception is a relayed address though. The relay between A and B might use a protocol to B that A does not support nor be able to parse.

thomaseizinger: > One might first think that a remote could never return an observed address, i.e. an address that it observes the local node as, that the local node doesn't know.

A remote can return all sorts of garbage :)

How big is the impact if we make this field an `Option`? That would be cleaner IMO.

thomaseizinger (approved):
LGTM as a bug fix. Long-term we might want to make better use of the type-system.

jxs (approved):
LGTM

protocols/identify/src/protocol.rs
mxinden: Agreed that `Option` would be the way to go here. That said, I think we should keep the patch release minimal, i.e. only do this on latest `master`.

mxinden:
[Tagged](https://github.com/libp2p/rust-libp2p/releases/tag/libp2p-identify-v0.41.1) and [published](https://crates.io/crates/libp2p-identify/0.41.1) `libp2p-identify` `v0.41.1`.

PR #3249, opened 2022-12-15
ci(cache): only save caches on master branch

thomaseizinger:
## Description

Previously, we would always save certain smaller caches like clippy etc on each PR. This still amounts to a lot of caches. With this patch, we conditionally only save them when the workflow is run on the master branch. This gives most pull requests a good base to work with to speed them up significantly without us keeping around a lot of caches and have GitHub invalidate them constantly.

On master, the condition evaluates to `true` which saves the cache: https://github.com/thomaseizinger/rust-libp2p/actions/runs/3699904381/jobs/6267774899#step:5:3
On pull-requests, the condition evaluates to `false` which skips saving it: https://github.com/thomaseizinger/rust-libp2p/actions/runs/3700055536/jobs/6268096357#step:4:3

## Notes

~Draft until I confirm this works as expected in my fork.~ Confirmed that it works.

## Change checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -115,6 +115,8 @@ jobs:
           target: ${{ matrix.target }}
 
       - uses: Swatinem/rust-cache@359a70e43a0bb8a13953b04a90f76428b4959bb6 # v2.2.0
+        with:
+          save-if: ${{ github.ref == 'refs/heads/master' }}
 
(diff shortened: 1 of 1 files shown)


jxs (approved):
Makes sense, thanks for this Thomas!

thomaseizinger:
Blocked from merging by https://github.com/libp2p/rust-libp2p/pull/3252.

PR #3253, opened 2022-12-16
refactor(swarm): express dial logic linearly

### LINKED ISSUE #3209: Concurrent dials to the same peer with duplicated multiaddrs
## Summary
Useless concurrent dials when a peer is found with two identical multiaddrs in the DHT: one with a peerid, and the other without a  peerid.

## Expected behaviour
When dialing a peer, identical multiaddrs (that only differ with the presence of the peerid) should be filtered out in order to prevent useless concurrent dials to the same peer.

## Actual behaviour
Two identical mutliaddrs filled by the DHT:

/ip4/10.10.0.1/tcp/80
and
/ip4/10.10.0.1/tcp/80/QmXXXXXXXX

are concurrently dialed twice.

## Possible Solution
[Maybe check uniqueness of multiaddrs without peerid](https://github.com/libp2p/rust-libp2p/blob/1c2712c1bc288dc608aaec5fc3458b0d07181feb/swarm/src/lib.rs#L590)

## Version

- libp2p version (version number, commit, or branch): 0.49

No

thomaseizinger:
## Description

Previously, the logic within `Swarm::dial` involved fairly convoluted `match` expressions. This patch refactors this function to use new utility functions introduced on `DialOpts` to handle one concern at a time.

This has the advantage that we are covering slightly more cases now. Because we are parsing the `PeerId` only once at the top, checks like banning will now also act on dials that specify the `PeerId` as part of the `/p2p` protocol.

## Notes

I am touching this function as part of https://github.com/libp2p/rust-libp2p/pull/3099 and found it hard to integrate the new callbacks into it.

## Change checklist

### DIFF
--- a/swarm/src/dial_opts.rs
+++ b/swarm/src/dial_opts.rs
@@ -20,6 +20,8 @@
 // DEALINGS IN THE SOFTWARE.
 
 use libp2p_core::connection::Endpoint;
+use libp2p_core::multiaddr::Protocol;
+use libp2p_core::multihash::Multihash;
 use libp2p_core::{Multiaddr, PeerId};
 use std::num::NonZeroU8;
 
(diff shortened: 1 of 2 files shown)


thomaseizinger:
With this refactoring, we are not executing certain checks more thoroughly. For example, we make all addresses to be dialed unique and filter out the ones we listen on. Previously, those was only true for the addresses coming from `WithPeerIdWithAddresses`. Was that on purpose @mxinden? (I think you introduced this builder).

I think the current version is more correct so I pushed a hack that fixes the test. Alternatively, we could also change the test (and perhaps remove the other code that checks for self dials?).

mxinden:
> For example, we make all addresses to be dialed unique and filter out the ones we listen on. Previously, those was only true for the addresses coming from `WithPeerIdWithAddresses`. Was that on purpose @mxinden? (I think you introduced this builder).

No longer dialing duplicate addresses would fix https://github.com/libp2p/rust-libp2p/issues/3209.

Not doing it everywhere but in `WithPeerIdWithAddresses` only was not a deliberate choice. I am in favor of doing it everywhere. Thanks.

mxinden (approved):
All in all a nice clean-up. Thanks @thomaseizinger. Happy my ugly deeply nested method is gone.

thomaseizinger:
> > For example, we make all addresses to be dialed unique and filter out the ones we listen on. Previously, those was only true for the addresses coming from `WithPeerIdWithAddresses`. Was that on purpose @mxinden? (I think you introduced this builder).
> 
> No longer dialing duplicate addresses would fix https://github.com/libp2p/rust-libp2p/issues/3209.

I don't think this patch fixed #3209 because we are doing the deduplication before we ensure the address has a `/p2p` ending.

Let's re-open #3209 for now and confirm with a test that it is actually fixed.

PR #3239, opened 2022-12-14
feat(kademlia)!: use GATs on `RecordStore` trait

thomaseizinger:
## Description

Previously, we applied a lifetime onto the entire `RecordStore` to workaround Rust not having GATs. With Rust 1.65.0 we now have GATs so we can remove this workaround.

Related https://github.com/libp2p/rust-libp2p/issues/3240. Without this change, we would have to specify HRTB in various places.

## Notes

Depends-On: https://github.com/libp2p/test-plans/pull/90

## Change checklist

- [ ] I have performed a self-review of my own code
- [ ] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] A changelog entry has been made in the appropriate crates

PR #3221, opened 2022-12-09
refactor!: Move `ConnectionId` and `PendingPoint` to `libp2p-swarm`

umgefahren:
## Description

Both of these are only needed as part of `libp2p-swarm`. Them residing in `libp2p-core` is a left-over from when `libp2p-core` still contained `Pool`.

Closes #3200.

## Notes

~As requested by @thomaseizinger I moved ConnectionId from core to swarm. I also changed every use of the struct from core to swarm.~

~I moved the `ConnectionId` struct and changed every relevant use. I also want to talk about the necessity of operator overloading.~

## Open Questions

None

## Change checklist

- [x] I have performed a self-review of my own code
- [ ] A changelog entry has been made in the appropriate crates

Issue #2680, opened 2022-05-30
swarm/: `NetworkBehaviour`/`ConnectionHandler` cross-communication


mxinden:
## Description

Today there is no way to send generic local and remote node information from one `NetworkBehaviour` to another `NetworkBehaviour`, apart from `NetworkBehaviour::inject_new_external_addr`.

Examples of such generic information could be:

- Supported protocols by the local peer.
- Supported protocols of a remote peer.
- Addresses of a remote peer.

## Motivation

One example use-case is the exchange of information between`libp2p-identify` and `libp2p-kad` to support Kademlia Client Mode (see https://github.com/libp2p/rust-libp2p/issues/2032).

1. `libp2p-kad` needs to know whether a remote peer advertises `/ipfs/kad/1.0.0` via the identify protocol to decide whether to include the remote peer in its routing table or not.
2. `libp2p-identify` needs to know whether the local node runs in client or in server mode, i.e. supports the `/ipfs/kad/1.0.0` or not.

See https://github.com/libp2p/rust-libp2p/pull/2521 which is currently blocked on this.

## Proposal

Below is an initial action plan proposal based on a meeting with @MarcoPolo.

1. Proof of concept 1 (communicating remote metadata across behaviours)
   - Extend `NetworkBehaviourAction` with `NetworkBehaviourAction::ReportRemoteMetaData`, containing a delta update.
   - Add `NetworkBehaviourAction::inject_remote_metadata`.
   - Write a `Select` `NetworkBehaviour` which takes two `NetworkBehaviour` implementations and passes `ReportRemoteMetaData` emitted by one behaviour to the other via `inject_remote_metadata`.
   - Showcase how `libp2p-kad` can learn supported protocols of remote peers from `libp2p-identify`.
2. Proof of concept 2 (communicating local metadata across behaviours)
   - Extend `NetworkBehaviourAction` with `NetworkBehaviourAction::ReportSupportedProtocol` which each `NetworkBehaviour` emits when it supports a protocol as a listener and which is passed to other `NetworkBehaviour`s via the `Select` `NetworkBehaviour`.
   - Showcase how remote peers can learn that `libp2p-kad` moved from client mode into server mode via `libp2p-identify`.
4. In case proof of concepts looks good, don't merge but consolidate `inject_*` methods into single `inject_event` method with a single `enum`. Motivation here is to not have yet another `inject_***` method on `NetworkBehaviour`.
[...]

PR #3254, opened 2022-12-16
feat(swarm)!: allow `NetworkBehaviour`s to manage connections

### LINKED ISSUE #2824: swarm/: Support generic connection management

thomaseizinger:
## Description

Previously, a `ConnectionHandler` was immediately requested from the `NetworkBehaviour` as soon as a new dial was initiated or a new incoming connection accepted.

With this patch, we delay the creation of the handler until the connection is actually established and fully upgraded, i.e authenticated and multiplexed.

As a consequence, `NetworkBehaviour::new_handler` is now deprecated in favor of a new set of callbacks:

- `NetworkBehaviour::handle_pending_inbound_connection`
- `NetworkBehaviour::handle_pending_outbound_connection`
- `NetworkBehaviour::handle_established_inbound_connection`
- `NetworkBehaviour::handle_established_outbound_connection`

All callbacks are fallible, allowing the `NetworkBehaviour` to abort the connection either immediately or after it is fully established. All callbacks also receive a `ConnectionId` parameter which uniquely identifies the connection. For example, in case a `NetworkBehaviour` issues a dial via `NetworkBehaviourAction::Dial`, it can unambiguously detect this dial in these lifecycle callbacks via the `ConnectionId`. 

Finally, `NetworkBehaviour::handle_pending_outbound_connection` also replaces `NetworkBehaviour::addresses_of_peer` by allowing the behaviour to return more addresses to be used for the dial.

Resolves #2824.

## Notes

This is a variant of https://github.com/libp2p/rust-libp2p/pull/3099 with less breaking changes and more deprecations.

- Depends-On: https://github.com/libp2p/rust-libp2p/pull/3243
- Depends-On: https://github.com/libp2p/rust-libp2p/pull/3239
- Depends-On: https://github.com/libp2p/rust-libp2p/pull/3253
- Depends-On: https://github.com/libp2p/rust-libp2p/pull/3272
- Depends-On: https://github.com/libp2p/rust-libp2p/pull/3264
- Depends-On: https://github.com/libp2p/rust-libp2p/pull/3346
- Depends-On: https://github.com/libp2p/rust-libp2p/pull/3334
- Depends-On: https://github.com/libp2p/rust-libp2p/pull/3327
- Depends-On: https://github.com/libp2p/rust-libp2p/pull/3335
- Depends-On: https://github.com/libp2p/rust-libp2p/pull/3372
- Depends-On: https://github.com/libp2p/rust-libp2p/pull/3328
- Depends-On: https://github.com/libp2p/rust-libp2p/pull/3368
- Depends-On: https://github.com/libp2p/rust-libp2p/pull/3373
- Depends-On: https://github.com/libp2p/rust-libp2p/pull/3374
- Depends-On: https://github.com/libp2p/rust-libp2p/pull/3375
- Depends-On: https://github.com/libp2p/rust-libp2p/pull/3381
- Depends-On: https://github.com/libp2p/rust-libp2p/pull/3384

## Open Questions

- ~The `ConnectionId` is currently generated through a random number generator. This is not ideal and could in theory lead to collisions:~
	- ~Should we use a UUID instead? I think those are less likely to produce a collision.~
	- ~Should we manage the connection IDs centrally, i.e. in the `Swarm`? This would require us to provide a facility to the `NetworkBehaviour` in how it can generate a new one to ensure it can track state for its own dial. We could use `PollParameters` for that. One downside of this is a slight ergonomic hit because the user can only fully construct a `NetworkBehaviourAction::Dial` within `NetworkBehaviour::poll` and not within the other callbacks. Some of our protocols already do that via an `ActionBuilder` pattern.~

## Change checklist


thomaseizinger:
> ##
> 
>     * The `ConnectionId` is currently generated through a random number generator. This is not ideal and could in theory lead to collisions:
>       
>       * Should we use a UUID instead? I think those are less likely to produce a collision.
>       * Should we manage the connection IDs centrally, i.e. in the `Swarm`? This would require us to provide a facility to the `NetworkBehaviour` in how it can generate a new one to ensure it can track state for its own dial. We could use `PollParameters` for that. One downside of this is a slight ergonomic hit because the user can only fully construct a `NetworkBehaviourAction::Dial` within `NetworkBehaviour::poll` and not within the other callbacks. Some of our protocols already do that via an `ActionBuilder` pattern. The interface could look something like:
> 
> 
> ```rust
> pub trait PollParameters {
>     fn new_dial(&mut self, opts: DialOpts) -> NetworkBehaviourAction<X, Y>;
> }
> ```

Use a static Atomic counter instead. That gives us the best of both worlds!

thomaseizinger:
@mxinden @jxs @elenaf9: I will continue to try and split things out of this PR into separate ones. You can see the list of "dependencies" in the pull request description. To push this forward, please have a look at those too! :)

dignifiedquire:
@thomaseizinger how can I know the connectionid when dialing from the behavior, with [NetworkBehaviourAction](https://docs.rs/libp2p/latest/libp2p/swarm/enum.NetworkBehaviourAction.html#)::Dial

thomaseizinger:
> @thomaseizinger how can I know the connectionid when dialing from the behavior, with [NetworkBehaviourAction](https://docs.rs/libp2p/latest/libp2p/swarm/enum.NetworkBehaviourAction.html#)::Dial

`ConnectionId` has a `Default` implementation now that will increment an atomic counter. You can construct the `Connectionid` first, clone it, pass it to `NetworkBehaviourAction::Dial` and then associate state with it in your `NetworkBehaviour`.

I am also intending to offer the following to make this more ergonomic!

```rust
impl NetworkBehaviourAction {
	pub fn dial(opts: DialOpts) -> (Self, ConnectionId) {
		let id = ConnectionId::default();
		let action = NetworkBehaviourAction::Dial {
			opts,
			id
		};

		(action, id)
	}
}
```

dignifiedquire:
Ah this is great 👍

mxinden:
> I will continue to try and split things out of this PR into separate ones. You can see the list of "dependencies" in the pull request description. To push this forward, please have a look at those too! :)

Thanks for that! Makes our life a lot easier. Reviewed all dependent PRs. Taking a look a this pull request next.

swarm/src/behaviour.rs
```diff
@@ -123,37 +116,133 @@ pub(crate) type THandlerOutEvent<THandler> =
 ///   }
 /// }
 /// ```
-pub trait NetworkBehaviour: 'static {
+pub trait NetworkBehaviour: Sized + 'static {
```
mxinden: How come this has to be explicitly `Sized` now?

swarm/src/behaviour.rs
```diff
@@ -123,37 +116,133 @@ pub(crate) type THandlerOutEvent<THandler> =
...
+            },
+        ))
+    }
+
+    /// Callback that is invoked for every outbound connection attempt.
+    ///
+    /// We have access to:
+    ///
+    /// - The [`PeerId`], if known. Remember that we can dial without a [`PeerId`].
+    /// - All addresses passes to [`DialOpts`] are passed in here too.
```
```suggestion
    /// - All addresses passed to [`DialOpts`] are passed in here too.
```

swarm/src/lib.rs
```diff
@@ -500,146 +511,100 @@ where
...
-                        Either::Left(addresses.into_iter()),
-                        dial_concurrency_factor_override,
-                        role_override,
-                    )
+            match self.behaviour.handle_pending_outbound_connection(
+                peer_id,
+                addresses_from_opts.as_slice(),
+                dial_opts.role_override(),
+                connection_id,
+            ) {
```
mxinden: Just calling this out to make sure I am not missing something. I don't think this is problematic.

Addresses collected through `handle_pending_outbound_connection` might not actually be used in the dial in case the dialer did not enable `extend_addresses_through_behaviour`.

swarm/src/behaviour.rs
```diff
@@ -123,37 +116,133 @@ pub(crate) type THandlerOutEvent<THandler> =
...
+        }
+    }
+
+    /// Callback that is invoked for every established outbound connection.
+    ///
+    /// This is invoked once we have successfully dialed a peer.
+    /// At this point, we have verified their [`PeerId`] and we know, which particular [`Multiaddr`] succeeded in the dial.
+    /// In order to actually use this connection, this function must return a [`ConnectionHandler`].
+    /// Returning an error will immediately close the connection.
+    fn handle_established_outbound_connection(
```
mxinden, line 227: Checking my understanding, we still need the `FromSwarm::ConnectionEstablished` event as any `NetworkBehaviour` within the hierarchy of behaviours might deny the connection.

If so, what do you think of elaborating this in the methods doc comment? Would that confuse the user more than it helps?

swarm/src/behaviour.rs
```diff
@@ -123,37 +116,133 @@ pub(crate) type THandlerOutEvent<THandler> =
...
+        ))
+    }
 
     /// Addresses that this behaviour is aware of for this specific peer, and that may allow
     /// reaching the peer.
     ///
     /// The addresses will be tried in the order returned by this function, which means that they
     /// should be ordered by decreasing likelihood of reachability. In other words, the first
     /// address should be the most likely to be reachable.
+    #[deprecated(note = "Use `NetworkBehaviour::handle_pending_outbound_connection` instead.")]
```
mxinden, line 250: I like that this is combined with the connection management method now.

swarm/src/behaviour.rs
```diff
@@ -270,14 +358,10 @@ pub trait NetworkBehaviour: 'static {
     fn inject_dial_failure(
         &mut self,
         peer_id: Option<PeerId>,
-        handler: Self::ConnectionHandler,
         error: &DialError,
+        id: ConnectionId,
```
mxinden: I am assuming that this might simplify a couple of `NetworkBehaviour` implementations. :+1:

mxinden (commented):
Well thought through. Thanks @thomaseizinger!

I reviewed `swarm/src/lib.rs` and `swarm/src/behaviour.rs` under the assumption that the other files are not yet ready for review. Please ping me in case I should look at more or take another look at the above.

swarm/src/lib.rs
thomaseizinger: Yep, I think this might warrant a `debug` log.

swarm/src/behaviour.rs
thomaseizinger: I don't remember, let me try remove it and see what breaks!

swarm/src/behaviour.rs
thomaseizinger: Yes, the event indicates that the connection was not denied and can actually be used.

swarm/src/behaviour.rs
thomaseizinger: To be honest, these events are quite a mess, esp. the divergence of `FromSwarm` and `SwarmEvent`. It would be nice if we can get rid of `SwarmEvent` apart from it emitting behaviour events.

swarm/src/behaviour.rs
thomaseizinger: Not needed!

thomaseizinger:
So it turns out deprecating trait functions doesn't trigger a deprecation warning ....

https://github.com/rust-lang/rust/pull/98991

protocols/relay/src/v2/client.rs
```diff
@@ -146,11 +152,51 @@ impl Client {
...
-    fn new_handler(&mut self) -> Self::ConnectionHandler {
-        handler::Prototype::new(self.local_peer_id, None)
+    fn handle_established_inbound_connection(
+        &mut self,
+        peer: PeerId,
+        connection_id: ConnectionId,
+        local_addr: &Multiaddr,
+        remote_addr: &Multiaddr,
+    ) -> Result<THandler<Self>, Box<dyn std::error::Error + Send + 'static>> {
+        let is_relayed = local_addr.iter().any(|p| p == Protocol::P2pCircuit); // TODO: Make this an extension on `Multiaddr`.
```
thomaseizinger: TODO: This should be in a separate PR.

protocols/ping/src/lib.rs
```diff
@@ -166,4 +160,33 @@ impl NetworkBehaviour for Behaviour {
             | FromSwarm::ExpiredExternalAddr(_) => {}
         }
     }
+
+    fn on_connection_handler_event(
+        &mut self,
+        peer: PeerId,
+        _: ConnectionId,
+        result: THandlerOutEvent<Self>,
```
thomaseizinger: TODO: Using these type aliases everywhere should be a separate PR.

protocols/relay/src/v2/client.rs
```diff
@@ -304,20 +349,21 @@ impl NetworkBehaviour for Client {
                         }),
                     },
                     None => {
-                        let handler = handler::Prototype::new(
-                            self.local_peer_id,
-                            Some(handler::In::EstablishCircuit {
+                        let (action, connection_id) = NetworkBehaviourAction::dial(
+                            DialOpts::peer_id(relay_peer_id)
+                                .addresses(vec![relay_addr])
+                                .extend_addresses_through_behaviour() // TODO: Why?
```
thomaseizinger: TODO: Fixing this should be a separate PR.

protocols/relay/src/v2/relay/handler.rs
```diff
@@ -432,6 +394,22 @@ pub struct Handler {
...
+            pending_error: Default::default(),
+            keep_alive: KeepAlive::Yes,
+            reservation_request_future: Default::default(),
+            active_reservation: Default::default(),
+            circuit_accept_futures: Default::default(),
+            circuit_deny_futures: Default::default(),
+            alive_lend_out_substreams: Default::default(),
+            circuits: Default::default(),
+        }
+    }
```
thomaseizinger: TODO: Introducing this can be done in a separate PR.

protocols/rendezvous/src/client.rs
```diff
@@ -173,26 +169,74 @@ impl NetworkBehaviour for Behaviour {
...
+
+        Ok(self
+            .discovered_peers
             .iter()
-            .filter_map(|((candidate, _), addresses)| (candidate == peer).then_some(addresses))
+            .filter_map(|((candidate, _), addresses)| (candidate == &peer).then_some(addresses))
             .flatten()
             .cloned()
-            .collect()
+            .collect())
```
thomaseizinger: TODO: Attempt to minimize diff here be storing addresses in separate variable.

protocols/rendezvous/src/client.rs
```diff
@@ -173,26 +169,74 @@ impl NetworkBehaviour for Behaviour {
...
+            | FromSwarm::DialFailure(_)
+            | FromSwarm::ListenFailure(_)
+            | FromSwarm::NewListener(_)
+            | FromSwarm::NewListenAddr(_)
+            | FromSwarm::ExpiredListenAddr(_)
+            | FromSwarm::ListenerError(_)
+            | FromSwarm::ListenerClosed(_)
+            | FromSwarm::NewExternalAddr(_)
+            | FromSwarm::ExpiredExternalAddr(_) => {}
+        }
```
thomaseizinger: TODO: Revert reordering this to minimize diff.

swarm/src/connection/pool.rs
```diff
@@ -285,16 +293,14 @@ pub enum PoolEvent<THandler: IntoConnectionHandler> {
...
-        handler: THandler,
     },
 
     /// A node has produced an event.
     ConnectionEvent {
         id: ConnectionId,
         peer_id: PeerId,
         /// The produced event.
-        event: THandlerOutEvent<THandler>,
+        event: THandler::OutEvent,
```
thomaseizinger, line 288: TODO: Revert this.

swarm/src/connection/pool.rs
```diff
@@ -310,11 +316,12 @@ pub enum PoolEvent<THandler: IntoConnectionHandler> {
...
 impl<THandler> Pool<THandler>
 where
-    THandler: IntoConnectionHandler,
+    THandler: ConnectionHandler,
 {
     /// Creates a new empty `Pool`.
     pub fn new(local_id: PeerId, config: PoolConfig, limits: ConnectionLimits) -> Self {
-        let (pending_connection_events_tx, pending_connection_events_rx) = mpsc::channel(0);
+        let (pending_connection_events_tx, pending_connection_events_rx) =
+            mpsc::channel(config.task_event_buffer_size);
```
thomaseizinger: TODO: Bad merge.

swarm/src/connection/pool.rs
```diff
@@ -348,7 +354,7 @@ where
     pub fn get_established(
         &mut self,
         id: ConnectionId,
-    ) -> Option<&mut EstablishedConnection<THandlerInEvent<THandler>>> {
+    ) -> Option<&mut EstablishedConnection<THandler::InEvent>> {
```
thomaseizinger, line 341: TODO: Revert this.

swarm/src/connection/pool.rs
```diff
@@ -774,23 +787,23 @@ where
...
+
                     return Poll::Ready(PoolEvent::ConnectionEstablished {
                         peer_id: obtained_peer_id,
                         endpoint,
                         id,
                         other_established_connection_ids,
                         concurrent_dial_errors,
-                        established_in,
+                        established_in: accepted_at.elapsed(),
+                        supported_protocols,
```
thomaseizinger: TODO: Minimize diff here.

swarm/src/connection/pool.rs
```diff
@@ -774,23 +787,23 @@ where
...
                     });
                 }
                 task::PendingConnectionEvent::PendingFailed { id, error } => {
                     if let Some(PendingConnection {
                         peer_id,
-                        handler,
                         endpoint,
                         abort_notifier: _,
-                        accepted_at: _, // Ignoring the time it took for the connection to fail.
+                        accepted_at: _,
```
thomaseizinger: TODO: Minimize diff here.

swarm/src/connection/pool.rs
```diff
@@ -1095,6 +1106,12 @@ impl PoolConfig {
         }
     }
 
+    /// Configures the executor to use for spawning connection background tasks.
+    pub fn with_executor(mut self, executor: Box<dyn Executor + Send>) -> Self {
+        self.executor = Some(executor);
+        self
+    }
```
thomaseizinger: TODO: Bad merge?

swarm/src/connection/pool/task.rs
```diff
@@ -66,7 +66,10 @@ pub enum PendingConnectionEvent {
 }
 
 #[derive(Debug)]
-pub enum EstablishedConnectionEvent<THandler: ConnectionHandler> {
+pub enum EstablishedConnectionEvent<THandler>
+where
+    THandler: ConnectionHandler,
+{
```
thomaseizinger: TODO: Revert or send separate PR.

swarm/src/dummy.rs
```diff
@@ -17,20 +22,24 @@ impl NetworkBehaviour for Behaviour {
...
-        _: &mut impl PollParameters,
-    ) -> Poll<NetworkBehaviourAction<Self::OutEvent, Self::ConnectionHandler>> {
-        Poll::Pending
+        _: PeerId,
+        _: &Multiaddr,
+        _: Endpoint,
+        _: ConnectionId,
+    ) -> Result<THandler<Self>, Box<dyn Error + Send + 'static>> {
+        Ok(ConnectionHandler)
     }
```
thomaseizinger: TODO: Revert reordering or send separate PR.

core/src/connection.rs
```diff
@@ -19,27 +19,37 @@
...
+    }
+}
 
-    fn add(self, other: usize) -> Self {
-        Self(self.0 + other)
+impl fmt::Display for ConnectionId {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{}", self.0)
     }
 }
```
thomaseizinger: TODO: Wait for https://github.com/libp2p/rust-libp2p/pull/3221 to be merged here.

protocols/floodsub/src/layer.rs
```diff
@@ -344,16 +336,31 @@ impl NetworkBehaviour for Floodsub {
...
+        Ok(Default::default())
+    }
+
+    fn handle_established_outbound_connection(
+        &mut self,
+        _: PeerId,
+        _: &Multiaddr,
+        _: Endpoint,
+        _: ConnectionId,
+    ) -> std::result::Result<THandler<Self>, Box<dyn std::error::Error + Send + 'static>> {
```
thomaseizinger: TODO: Audit PR for unnecessarily qualified `std::result::Result`.

protocols/relay/src/v2/client.rs
```diff
@@ -188,9 +234,9 @@ impl NetworkBehaviour for Client {
         &mut self,
         event_source: PeerId,
         _connection: ConnectionId,
-        handler_event: Either<handler::Event, void::Void>,
+        _event: THandlerOutEvent<Self>,
     ) {
-        let handler_event = match handler_event {
+        let handler_event = match _event {
```
thomaseizinger: TODO: Minimize diff.

protocols/relay/src/v2/client/handler.rs
```diff
@@ -213,6 +155,21 @@ pub struct Handler {
...
+            remote_addr,
+            pending_error: Default::default(),
+            keep_alive: KeepAlive::Yes,
+            queued_events: Default::default(),
+            reservation: Reservation::None,
+            alive_lend_out_substreams: Default::default(),
+            circuit_deny_futs: Default::default(),
+            send_error_futs: Default::default(),
+        }
+    }
```
thomaseizinger: TOOD: Can be in a separate PR to reduce diff.

protocols/rendezvous/src/client.rs
```diff
@@ -31,23 +31,19 @@ use instant::Duration;
...
 use std::iter::FromIterator;
 use std::task::{Context, Poll};
 
 pub struct Behaviour {
-    events: VecDeque<
-        NetworkBehaviourAction<
-            Event,
-            SubstreamConnectionHandler<void::Void, outbound::Stream, outbound::OpenInfo>,
-        >,
-    >,
```
thomaseizinger: TODO: Reduce diff here by first migrating everything to `THandler`.

thomaseizinger (commented):
A few remarks for myself.

swarm/src/lib.rs
```diff
@@ -917,25 +966,42 @@ where
...
+                match self.behaviour.handle_pending_inbound_connection(
+                    connection_id,
+                    &local_addr,
+                    &send_back_addr,
+                ) {
+                    Ok(()) => {}
+                    Err(error) => {
+                        // TODO: self.behaviour.on_swarm_event(FromSwarm::ListenFailure())
+
+                        return Some(todo!("Report connection error event, these are currently modeled in a weird way where we split them by incoming/outgoing on the `SwarmEvent` level but everywhere else by established/pending"));
```
thomaseizinger: @mxinden: Big question mark here on how we want to model this. Input welcome.

mxinden: If I understand correctly, we didn't return a `SwarmEvent` in this particular case at all, right? I am not opposed to returning a `SwarmEvent` in the case were we reject a pending inbound connection.

I don't have a strong opinion on which of the two namespace splits should take preference. Striving for consistency of the two sounds good to me, though seems to be a separate undertaking.

core/src/connection.rs
```diff
@@ -19,27 +19,37 @@
...
+    #[deprecated(note = "IDs must be unique and should not be constructed directly.")]
     pub fn new(id: usize) -> Self {
-        Self(id)
+        Self(id as u64)
     }
 }
 
-impl std::ops::Add<usize> for ConnectionId {
-    type Output = Self;
+impl Default for ConnectionId {
```
mxinden: Ignoring the fact that there has already been a `ConnectionId::new` and that that method takes a parameter.

My intuition for `Default` would be that two `Default::default()` would equal. Is that idiomatic Rust or only my intuition.

This is not particularly important. Fine with keeping it as is.

swarm/src/lib.rs
thomaseizinger: > If I understand correctly, we didn't return a `SwarmEvent` in this particular case at all, right?

Right, we didn't. I hadn't considered _not_ returning one, that might indeed unblock this bit, thanks!

> Striving for consistency of the two sounds good to me, though seems to be a separate undertaking.

It is a bit of a mess 😔

core/src/connection.rs
thomaseizinger: Good question. I think I agree with you, now that you are pointing it out!

We can easily replace it with a public `next_unused` constructor.

swarm/src/lib.rs
thomaseizinger: > > If I understand correctly, we didn't return a `SwarmEvent` in this particular case at all, right?
> 
> Right, we didn't. I hadn't considered _not_ returning one, that might indeed unblock this bit, thanks!
> 
> > Striving for consistency of the two sounds good to me, though seems to be a separate undertaking.
> 
> It is a bit of a mess pensive

Actually, this still proves to be a bit difficult. In follow-up PRs to this, I think we should re-implement some of the current functionality in regards to banning, connection limits etc on top of this new infrastructure. That will remove several, currently explicit error cases. I think if we don't model these error properly, it will be difficult for users to understand what is happening.

I am proposing to align these errors before we merge this PR.

thomaseizinger: > I am proposing to align these errors before we merge this PR.

It might have to be part of this, I'll see what I can do.

thomaseizinger:
> Well thought through. Thanks @thomaseizinger!
> 
> I reviewed `swarm/src/lib.rs` and `swarm/src/behaviour.rs` under the assumption that the other files are not yet ready for review. Please ping me in case I should look at more or take another look at the above.

Thanks @mxinden! Most tests should be passing now, go nuts :)

swarm/src/behaviour.rs
```diff
@@ -437,200 +524,12 @@ pub trait PollParameters {
...
-    /// #         >,
-    /// #     > {
-    /// #         todo!("If `Self::message.is_some()` send the message to the remote.")
-    /// #     }
-    /// # }
-    /// # #[derive(Debug, PartialEq, Eq)]
-    /// # struct PreciousMessage(String);
-    /// ```
-    Dial { opts: DialOpts, handler: THandler },
+    Dial { opts: DialOpts, id: ConnectionId },
```
mxinden: A user could create a `NetworkBehaviourAction::Dial` with an already-in-use `ConnectionId`. Should we prevent this at compile time? Should one only be able to create a `NetworkBehaviourAction::Dial` with a **new** `ConnectionId` provided?

mxinden (commented):
Needs changelog entries and doc comments. Otherwise this looks good to me.

Great work on making the upgrade path from `new_handler` and `addresses_of_peer` to the new methods easy! Double checked the flow through the various combinator `NetworkBehaviour` implementations.

swarm/src/lib.rs
mxinden: > I think we should re-implement some of the current functionality in regards to banning, connection limits etc on top of this new infrastructure.

Just to make sure we agree here, long term I would like all connection management to happen through `NetworkBehaviour` implementations. I.e. `libp2p-swarm` should not be aware of the concepts of connection limits nor peer banning.

mxinden: > I am proposing to align these errors before we merge this PR.

Fine with me.

swarm/src/lib.rs
elenaf9: Could we have `handle_pending_outbound_connection` return something like `impl Iterator<Item = &Multiaddr>` instead? That way we'd avoid unnecessary cloning of addresses if they end up not being used anyway.

thomaseizinger: > Could we have `handle_pending_outbound_connection` return something like `impl Iterator<Item = &Multiaddr>` instead? That way we'd avoid unnecessary cloning of addresses if they end up not being used anyway.

I'll explore it, good idea!

mxinden: > That way we'd avoid unnecessary cloning of addresses if they end up not being used anyway.

My initial comment was not aimed at the potential performance impact, but rather at the surprise for an implementer of the API, expecting their addresses to be used in the final dial. Instead of optimizing away the allocations, I suggest adding a comment to the trait's method to prevent implementers being surprised.

swarm/src/lib.rs
thomaseizinger: > > I think we should re-implement some of the current functionality in regards to banning, connection limits etc on top of this new infrastructure.
> 
> Just to make sure we agree here, long term I would like all connection management to happen through `NetworkBehaviour` implementations. I.e. `libp2p-swarm` should not be aware of the concepts of connection limits nor peer banning.

Yes. I assumed we would still provide a simple `Blocklist` behaviour that replaces the current banning functionality?

swarm/src/behaviour.rs
thomaseizinger: It is a valid concern. Do you have a design in mind already? We would have to either make the entire enum and opaque type or at least create a `Dial` type with private fields so we can encapsulate it properly.

swarm/src/lib.rs
mxinden: :+1: we are on the same page. I am fine with `Blocklist` to live within `libp2p-swarm`, as long as it simply implements `NetworkBehaviour`.

swarm/src/behaviour.rs
mxinden: > Do you have a design in mind already?

No design in mind. I don't see this as a blocker.

thomaseizinger:
> Great work on making the upgrade path from `new_handler` and `addresses_of_peer` to the new methods easy!

Unfortunately, the deprecation warnings don't show up in all places: https://github.com/rust-lang/rust/pull/98991.

Hopefully people read the changelogs.

swarm/src/behaviour.rs
thomaseizinger: Note: In https://github.com/libp2p/rust-libp2p/pull/3328, we now use a design that together with https://github.com/libp2p/rust-libp2p/pull/3346 will allow us to make the constructor private and thus prevent at compile-time that users can assign an existing `ConnectionId` to a new connection.

mxinden:
Let us know once this is ready for another review @thomaseizinger.

thomaseizinger:
> Let us know once this is ready for another review @thomaseizinger.

Will do! At least two more PRs need to merge before that. I'll see what else I can extract out of this one.

thomaseizinger:
> > Let us know once this is ready for another review @thomaseizinger.
> 
> Will do! At least two more PRs need to merge before that. I'll see what else I can extract out of this one.

Here are five more @jxs @mxinden:

- https://github.com/libp2p/rust-libp2p/pull/3328
- https://github.com/libp2p/rust-libp2p/pull/3368
- https://github.com/libp2p/rust-libp2p/pull/3373
- https://github.com/libp2p/rust-libp2p/pull/3374
- https://github.com/libp2p/rust-libp2p/pull/3375

protocols/ping/src/lib.rs
```diff
@@ -167,4 +159,33 @@ impl NetworkBehaviour for Behaviour {
...
+    ) {
+        self.events.push_front(Event { peer, result })
+    }
```
thomaseizinger: Note to self: Check that this diff gets better once we merge https://github.com/libp2p/rust-libp2p/pull/3368.

swarm/src/dummy.rs
```diff
@@ -51,6 +57,23 @@ impl NetworkBehaviour for Behaviour {
...
+        void::unreachable(event)
+    }
+
+    fn poll(
+        &mut self,
+        _: &mut Context<'_>,
+        _: &mut impl PollParameters,
+    ) -> Poll<NetworkBehaviourAction<Self::OutEvent, THandlerInEvent<Self>>> {
+        Poll::Pending
+    }
```
thomaseizinger: TODO: Check that this diff improves with https://github.com/libp2p/rust-libp2p/pull/3368.

protocols/dcutr/src/behaviour_impl.rs
```diff
@@ -235,18 +232,93 @@ impl Behaviour {
...
-    fn new_handler(&mut self) -> Self::ConnectionHandler {
-        handler::Prototype
+    fn handle_established_inbound_connection(
+        &mut self,
+        peer: PeerId,
+        connection_id: ConnectionId,
+        local_addr: &Multiaddr,
+        remote_addr: &Multiaddr,
+    ) -> Result<THandler<Self>, ConnectionDenied> {
+        let is_relayed = local_addr.iter().any(|p| p == Protocol::P2pCircuit); // TODO: Make this an extension on `Multiaddr`.
```
thomaseizinger: Any thoughts on how to handle this? With the introduction of the new APIs, we are no longer passing a `ConnectedPoint` which has the `is_relayed` utility function on it.

thomaseizinger:
This is now free of TODOs and should be fully working. I'll build a PR on top of this next that deprecated the `Swarm` built-in banning functionality and introduces a `Banlist` behaviour. Additionally, with this PR, we can remove the built-in connection limiting and build it as a `NetworkBehaviour`.

thomaseizinger:
> This is now free of TODOs and should be fully working. I'll build a PR on top of this next that deprecated the `Swarm` built-in banning functionality and introduces a `Banlist` behaviour. Additionally, with this PR, we can remove the built-in connection limiting and build it as a `NetworkBehaviour`.

Here is what connection limits look like when implemented using these new APIs: https://github.com/libp2p/rust-libp2p/pull/3386

thomaseizinger:
Actually, I think all of the breaking changes are in other pull requests and this by itself is not breaking! :astonished:

protocols/dcutr/src/behaviour_impl.rs
mxinden: I am fine with:

- Adding a method `Multiaddr::is_relayed`
- Adding a method `MultiaddrExt::is_relayed`
- Keeping as is, i.e. `iter().any`

protocols/relay/src/behaviour.rs
```diff
@@ -250,17 +252,61 @@ impl Behaviour {
...
-        handler::Prototype {
-            config: handler::Config {
+    fn handle_established_inbound_connection(
+        &mut self,
+        _: ConnectionId,
+        _: PeerId,
+        local_addr: &Multiaddr,
+        remote_addr: &Multiaddr,
+    ) -> Result<THandler<Self>, ConnectionDenied> {
+        let is_relayed = local_addr.iter().any(|p| p == Protocol::P2pCircuit); // TODO: Make this an extension on `Multiaddr`.
```
thomaseizinger: TODO: Resolve this in some form.

protocols/relay/src/priv_client.rs
```diff
@@ -156,11 +159,51 @@ impl Behaviour {
...
-    fn new_handler(&mut self) -> Self::ConnectionHandler {
-        handler::Prototype::new(self.local_peer_id, None)
+    fn handle_established_inbound_connection(
+        &mut self,
+        _connection_id: ConnectionId,
+        peer: PeerId,
+        local_addr: &Multiaddr,
+        remote_addr: &Multiaddr,
+    ) -> Result<THandler<Self>, ConnectionDenied> {
+        let is_relayed = local_addr.iter().any(|p| p == Protocol::P2pCircuit); // TODO: Make this an extension on `Multiaddr`.
```
thomaseizinger: Here too.

swarm/src/lib.rs
```diff
@@ -1608,6 +1716,9 @@ pub enum DialError {
         obtained: PeerId,
         endpoint: ConnectedPoint,
     },
+    Denied {
+        cause: ConnectionDenied,
+    },
```
thomaseizinger, line 1693: The addition of this variant makes this a breaking change.

swarm/src/lib.rs
```diff
@@ -1705,8 +1820,13 @@ pub enum ListenError {
...
     },
-    /// The peer identity obtained on the connection did not match the one that was expected.
-    LocalPeerId { endpoint: ConnectedPoint },
+    /// The connection was dropped because it resolved to our own [`PeerId`].
+    LocalPeerId {
+        endpoint: ConnectedPoint,
+    },
+    Denied {
+        cause: ConnectionDenied,
+    },
```
thomaseizinger, line 1801: Here too.

thomaseizinger:
> Actually, I think all of the breaking changes are in other pull requests and this by itself is not breaking! astonished

Not quite actually (see comments above) but we are pretty damn close :)

swarm/src/lib.rs
thomaseizinger: Are we happy with the log statement that I added?

swarm/src/connection/pool.rs
```diff
@@ -812,6 +819,48 @@ where
...
+    fn drop(&mut self) {
+        if let Some(connection) = self.connection.take() {
+            let _ = self
+                .drop_sender
+                .take()
+                .expect("`drop_sender` to always be `Some`")
+                .send(connection);
+        }
+    }
+}
```
thomaseizinger, line 862: @mxinden: I think this should solve your "always gracefully close connections" concern :)

thomaseizinger:
@libp2p/rust-libp2p-maintainers Exactly two months after initially opening this and after a series of "pave-the-way" pull requests (see PR description), this is now finally open for review.

@divagant-martian @rkuhn @melekes @dignifiedquire I am quite confident that this patch will greatly simplify connection management on your end. Nevertheless, feel free to experiment with it and please post any feedback you might have here. I've tried to keep the breaking changes to a minimum.

swarm/CHANGELOG.md
```diff
@@ -33,6 +33,27 @@
...
+- Allow `NetworkBehaviour`s to manage connections.
+  We deprecate `NetworkBehaviour::new_handler` and `NetworkBehaviour::addresses_of_peer` in favor of four new callbacks:
+
+  - `NetworkBehaviour::handle_pending_inbound_connection`
+  - `NetworkBehaviour::handle_pending_outbound_connection`
+  - `NetworkBehaviour::handle_established_inbound_connection`
+  - `NetworkBehaviour::handle_established_outbound_connection`
+
+  Please note that due to limitations in the Rust compiler, _implementations_ of `new_handler` and `addresses_of_peer` are not flagged as deprecated.
+  Nevertheless, they will be removed in the future.
```
mxinden: What do you think of linking to this from the main `CHANGELOG.md` file as a special callout for the release?

swarm/src/handler/either.rs
```diff
@@ -19,84 +19,23 @@
...
-use libp2p_core::{ConnectedPoint, PeerId};
+use libp2p_core::UpgradeError;
 use std::task::{Context, Poll};
 
-/// Auxiliary type to allow implementing [`IntoConnectionHandler`]. As [`IntoConnectionHandler`] is
-/// already implemented for T, we cannot implement it for Either<A, B>.
-pub enum IntoEitherHandler<L, R> {
-    Left(L),
-    Right(R),
-}
```
mxinden: Given that `IntoConnectionHandler` is only deprecated at this point, should this already be removed or only removed once we remove `IntoConnectionHandler`?

swarm/src/lib.rs
```diff
@@ -756,15 +784,76 @@ where
                 established_in,
             } => {
                 if self.banned_peers.contains(&peer_id) {
-                    self.pool.close_connection(connection);
```
mxinden, line 759: Following up on https://github.com/libp2p/rust-libp2p/pull/3328#discussion_r1101991956 and https://github.com/libp2p/rust-libp2p/pull/3254/#pullrequestreview-1292581602 it was not intuitive for me on a first read that `connection` on drop would close the underlying connection.

Thinking about it some more, I guess it is fine and significantly simplifies the error handling in this match arm.

swarm/src/lib.rs
```diff
@@ -802,6 +891,7 @@ where
                             other_established: other_established_connection_ids.len(),
                         },
                     ));
+                self.supported_protocols = supported_protocols;
```
mxinden, line 894: This is new behavior, right? Previously we would only retrieve the supported protocols once at start-up. Now we retrieve them on each new connection. Could this cause any side-effects for folks depending on the previous behavior? Would you mind introducing this change in a separate pull request?

swarm/src/lib.rs
```diff
@@ -1733,11 +1853,37 @@ impl error::Error for ListenError {
...
             ListenError::Aborted => None,
+            ListenError::Denied { cause } => Some(cause),
             ListenError::LocalPeerId { .. } => None,
         }
     }
 }
 
+#[derive(Debug)]
+pub struct ConnectionDenied {
+    inner: Box<dyn error::Error + Send + Sync + 'static>,
```
mxinden, line 1864: I am assuming that we require `Send` here to allow users to forward this error through a channel at a later point?

Why does the error need to be `Sync`?

swarm/src/lib.rs
```diff
@@ -1891,10 +2037,8 @@ mod tests {
     /// calls should be registered.
     #[test]
     fn test_connect_disconnect_ban() {
-        let _ = env_logger::try_init();
-
         // Since the test does not try to open any substreams, we can
-        // use keep alive protocols handler.
+        // use the dummy protocols handler.
```
mxinden: Merge conflict with https://github.com/libp2p/rust-libp2p/pull/3465.

swarm/src/lib.rs
```diff
@@ -1953,31 +2096,25 @@ mod tests {
                     {
                         // Setup to test that new connections of banned peers are not reported.
                         swarm1.dial(addr2.clone()).unwrap();
-                        s1_expected_conns += 1;
                         stage = Stage::BannedDial;
                     }
                 }
                 Stage::BannedDial => {
-                    if swarm1.behaviour.assert_disconnected(s1_expected_conns, 2) {
```
mxinden: Merge conflict with https://github.com/libp2p/rust-libp2p/pull/3465.

swarm/src/connection/pool.rs
mxinden: At first I thought this is too implicit for an internal component (i.e. `Pool`). Thinking about it some more, it does simplify error handling in `Swarm`. I am in favor of it. Thanks!

protocols/relay/src/priv_client/transport.rs
```diff
@@ -247,7 +248,7 @@ struct RelayedMultiaddr {
 
 /// Parse a [`Multiaddr`] containing a [`Protocol::P2pCircuit`].
 fn parse_relayed_multiaddr(addr: Multiaddr) -> Result<RelayedMultiaddr, TransportError<Error>> {
-    if !addr.iter().any(|p| matches!(p, Protocol::P2pCircuit)) {
+    if !addr.is_relayed() {
```
mxinden, line 251: :rocket:

protocols/kad/src/behaviour.rs
```diff
@@ -1978,21 +1980,66 @@ impl<TStore> NetworkBehaviour for Kademlia<TStore>
...
     TStore: RecordStore + Send + 'static,
 {
-    type ConnectionHandler = KademliaHandlerProto<QueryId>;
+    type ConnectionHandler = KademliaHandler<QueryId>;
     type OutEvent = KademliaEvent;
 
-    fn new_handler(&mut self) -> Self::ConnectionHandler {
-        KademliaHandlerProto::new(KademliaHandlerConfig {
-            protocol_config: self.protocol_config.clone(),
-            allow_listening: true,
```
mxinden, line 1987: Uuf, when did we miss this?

protocols/dcutr/src/behaviour_impl.rs
```diff
@@ -386,3 +460,7 @@ impl NetworkBehaviour for Behaviour {
         }
     }
 }
+
+fn is_relayed(addr: &Multiaddr) -> bool {
+    addr.iter().any(|p| p == Protocol::P2pCircuit)
+}
```
mxinden, line 466: No extension trait? Not a strong opinion.

protocols/dcutr/src/behaviour_impl.rs
```diff
@@ -237,11 +239,82 @@ impl Behaviour {
...
+        handler::relayed::Handler,
+        Either<handler::direct::Handler, dummy::ConnectionHandler>,
+    >;
     type OutEvent = Event;
 
-    fn new_handler(&mut self) -> Self::ConnectionHandler {
-        handler::Prototype
+    fn handle_established_inbound_connection(
+        &mut self,
+        _connection_id: ConnectionId,
```
mxinden: Why the `_`? It is used further below in the `remove` call.

protocols/dcutr/src/behaviour_impl.rs
```diff
@@ -237,11 +239,82 @@ impl Behaviour {
...
+        match self
+            .outgoing_direct_connection_attempts
+            .remove(&(_connection_id, peer))
+        {
+            None => {
+                let handler = if is_relayed(local_addr) {
+                    Either::Left(handler::relayed::Handler::new(ConnectedPoint::Listener {
+                        local_addr: local_addr.clone(),
+                        send_back_addr: remote_addr.clone(),
+                    })) // TODO: We could make two `handler::relayed::Handler` here, one inbound one outbound.
```
mxinden, line 264: Can you expand on this comment?

mxinden (commented):
Overall looking great. I have a couple of comments and questions. Nothing major.

swarm/src/lib.rs
thomaseizinger: I don't think we can do this in a separate PR. It definitely wasn't possible in earlier iterations. Will check.

The problem is that `new_handler` will panic by default now because you are not meant to call it. Thus, I cannot rely on it anywhere but have to call the new callbacks.

Those require connection-specific metadata that I only have after one is established.

swarm/src/lib.rs
thomaseizinger: I think it is because we pass references of the error to the behaviours via `FromSwarm`.

protocols/kad/src/behaviour.rs
thomaseizinger: What do you mean?

swarm/src/handler/either.rs
thomaseizinger: Yeah we can retain it!

protocols/dcutr/src/behaviour_impl.rs
thomaseizinger: Copy-paste error 🙃

protocols/dcutr/src/behaviour_impl.rs
thomaseizinger: "dcutr" cares about who the dialer and listener on a relayed connection is. Now that we have different callbacks for this, it is even more obvious that we could split this logic into two handlers instead of being conditional on the `ConnectedPoint` within the single handler. That would remove some `panic!`s in the implementation that are unreachable.

protocols/dcutr/src/behaviour_impl.rs
thomaseizinger: I opened https://github.com/libp2p/rust-libp2p/pull/3484 to prevent this in the future.

protocols/dcutr/src/behaviour_impl.rs
thomaseizinger: Didn't think it was worth it if the usage is only in a single file.

swarm/src/lib.rs
thomaseizinger: If the error is not `Sync`, users won't be able to use this with `anyhow`. See:

```
error[E0277]: `(dyn std::error::Error + std::marker::Send + 'static)` cannot be shared between threads safely
   --> interop-tests/src/bin/ping.rs:115:48
    |
115 |         swarm.dial(other.parse::<Multiaddr>()?)?;
    |                                                ^ `(dyn std::error::Error + std::marker::Send + 'static)` cannot be shared between threads safely
    |
```

swarm/src/lib.rs
thomaseizinger: Yes, I figured doing it on `Drop` is easier given that we would have several paths in here that require the connection to be closed properly. I see it as the responsibility of `Pool` to do this, meaning it should provide a hard-to-misuse API, even if it is just an internal one.

swarm/src/lib.rs
thomaseizinger: > Could this cause any side-effects for folks depending on the previous behavior? Would you mind introducing this change in a separate pull request?

Previously, we gathered the supported protocols on start-up of the swarm. Now, we are gathering it on the first connection. I think this can be observable if a `NetworkBehaviour` expects to read all supported protocols on the first call to `NetworkBehaviour::poll` via `PollParameters` when no connection is established yet.

I don't see a way around this change though unless we want to first completely redesign how `NetworkBehaviour`s signal which protocols they support.

We already agreed that `PollParameters` needs to go away (https://github.com/libp2p/rust-libp2p/issues/3124). I think the answer to this is that as part of https://github.com/libp2p/rust-libp2p/issues/2680, we scope supported protocols to a connection and thus remove this Swarm-wide state of "supported protocols".

thomaseizinger:
Before we merge this, are we happy with the `handle_` naming of the new callbacks? It is going to be a pain to rename these later so I want to make sure we are all happy with the naming.

mxinden:
> Before we merge this, are we happy with the handle_ naming of the new callbacks? It is going to be a pain to rename these later so I want to make sure we are all happy with the naming.

I am in favor of `handle_`. It matches with the return of a `ConnectionHandler`. By not using `on_` it is contrasted to the pure event handlers that in contrast don't return anything.

protocols/kad/src/behaviour.rs
mxinden: If I understand correctly `allow_listening` is always `true`, thus a dummy configuration option, thus fine to remove. I am assuming that this is a relict from a past refactoring. Not directly related to this pull request. Resolving the comment.

swarm/src/lib.rs
mxinden: Thanks for elaborating. Let's keep as is :+1:

swarm/src/lib.rs
mxinden: Happy!

mxinden (approved):
Ready for merge from my end. Please add the `send-it` label once you want this merged @thomaseizinger. Including it in v0.51.0 sounds good to me.

protocols/kad/src/behaviour.rs
thomaseizinger: It is not removed in this PR but I guess we should remove it, I'll open an issue!

jxs:
Sorry I missed this. 

> Before we merge this, are we happy with the `handle_` naming of the new callbacks? It is going to be a pain to rename these later so I want to make sure we are all happy with the naming.

Why not keep the `on_*` naming and have something like `on_upgrade_event` with `Enum` variants as the other `on_*` methods?

thomaseizinger:
> > Before we merge this, are we happy with the `handle_` naming of the new callbacks? It is going to be a pain to rename these later so I want to make sure we are all happy with the naming.
> 
> Why not keep the `on_*` naming and have something like `on_upgrade_event` with `Enum` variants as the other `on_*` methods?

In general, events typically don't have return values, they just state what happened. Event handlers are thus often named with `on_`. I'd find it weird to have an event handler that actually returns value.

melekes:
> Finally, NetworkBehaviour::handle_pending_outbound_connection also replaces NetworkBehaviour::addresses_of_peer by allowing the behaviour to return more addresses to be used for the dial.

Q: what's the preferred way of getting a list of known addresses for a peer? Before 0.51 we're using `addresses_of_peer` in our RPC to print a list of addresses (and other info) https://github.com/paritytech/substrate/blob/c69c0a61215724334a1b8a2ec4dfbf82bb4bb6e4/client/network/src/service.rs#L519. Using `handle_pending_outbound_connection` seems weird in that case since we're not going to open any outbound connections, we just need a list of addresses https://github.com/paritytech/substrate/blob/c69c0a61215724334a1b8a2ec4dfbf82bb4bb6e4/client/network/src/service.rs#L529.

thomaseizinger:
> > Finally, NetworkBehaviour::handle_pending_outbound_connection also replaces NetworkBehaviour::addresses_of_peer by allowing the behaviour to return more addresses to be used for the dial.
> 
> Q: what's the preferred way of getting a list of known addresses for a peer? Before 0.51 we're using `addresses_of_peer` in our RPC to print a list of addresses (and other info) [paritytech/substrate@`c69c0a6`/client/network/src/service.rs#L519](https://github.com/paritytech/substrate/blob/c69c0a61215724334a1b8a2ec4dfbf82bb4bb6e4/client/network/src/service.rs#L519). Using `handle_pending_outbound_connection` seems weird in that case since we're not going to open any outbound connections, we just need a list of addresses [paritytech/substrate@`c69c0a6`/client/network/src/service.rs#L529](https://github.com/paritytech/substrate/blob/c69c0a61215724334a1b8a2ec4dfbf82bb4bb6e4/client/network/src/service.rs#L529).

See also https://github.com/libp2p/rust-libp2p/issues/3633.

I consider all functions of `NetworkBehaviour` to be "internal" in the sense that they should not be called by anyone other than composing `NetworkBehaviour` implementations or `Swarm`.

At the moment, there is no replacement for your usecase. Which behaviours do you need to gather addresses from? I am open to add public functions to e.g. `mdns` or other "discovery" behaviours that expose this information. I don't think we should add a direct replacement function to `NetworkBehaviour`. Once the above functions are exposed, you should be able to create your own abstraction if you desire one.

melekes:
> I consider all functions of NetworkBehaviour to be "internal" in the sense that they should not be called by anyone other than composing NetworkBehaviour implementations or Swarm.

Makes sense.

> Which behaviours do you need to gather addresses from? I am open to add public functions to e.g. mdns or other "discovery" behaviours that expose this information.

mdns, kad.

> Once the above functions are exposed, you should be able to create your own abstraction if you desire one.

Sounds great 👍

thomaseizinger:
> > Which behaviours do you need to gather addresses from? I am open to add public functions to e.g. mdns or other "discovery" behaviours that expose this information.
> 
> mdns, kad.

Are you open to send a PR for mdns? I think kad already exposes the necessary functionality, see linked ticket above.

melekes:
> Are you open to send a PR for mdns?

sure

PR #3258, opened 2022-12-19
ci: install `protoc` from repositories where possible

thomaseizinger:
## Description

With the addition of more CI jobs, we are constantly running into API limits on setting up protoc. For all jobs that run on ubuntu, we can install it from `apt` instead.

On ubuntu 22.04, which is what `ubuntu-latest` points to, this installs `protoc v3.12.4`.

## Change checklist

### DIFF
--- a/.github/workflows/cache-factory.yml
+++ b/.github/workflows/cache-factory.yml
@@ -34,7 +34,7 @@ jobs:
         rust: ${{ fromJSON(needs.gather_msrv_versions.outputs.versions) }}
     steps:
       - name: Install Protoc
-        uses: arduino/setup-protoc@64c0c85d18e984422218383b81c52f8b077404d3 # v1.1.2
+        run: sudo apt-get install protobuf-compiler
 
       - uses: actions/checkout@v3
 
@@ -57,7 +57,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Install Protoc
-        uses: arduino/setup-protoc@64c0c85d18e984422218383b81c52f8b077404d3 # v1.1.2
+        run: sudo apt-get install protobuf-compiler
 
       - uses: actions/checkout@v3
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,9 +21,7 @@ jobs:
         crate: ${{ fromJSON(needs.gather_published_crates.outputs.members) }}
     steps:
       - name: Install Protoc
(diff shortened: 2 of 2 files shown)


MarcoPolo:
I’ve not looked closely, but let’s make sure this installs an up to date version. (Not something a year old)

thomaseizinger:
> I’ve not looked closely, but let’s make sure this installs an up to date version. (Not something a year old)

It is only meant to be a temporary fix until we ship https://github.com/libp2p/rust-libp2p/pull/3066.

I just ran the command in a `ubuntu:jammy` (22.04) docker image and got protoc 3.12.4 which is about 2.5 years old. What are your concerns in installing an older version?

MarcoPolo:
On my phone, but this has some context: https://github.com/libp2p/specs/issues/465#issuecomment-1272291524

thomaseizinger:
> On my phone, but this has some context: [libp2p/specs#465 (comment)](https://github.com/libp2p/specs/issues/465#issuecomment-1272291524)

Thanks for linking this. At the moment, I'd just like to ease us from the burden of having constantly red CI jobs, PRs being dequeued from the merge queue and having to baby-sit and restart them constantly until we finally get a window where we don't hit the API limit upon installing protoc.

Long-term ,we are moving away from requiring protoc. I'd assume that by the time this decision is made and enforced across all libp2p implementations, we will have long merged #3066.

@libp2p/rust-libp2p-maintainers Do you agree with this?

thomaseizinger:
Funnily enough, our interoperability tests install `protoc` through the exact same mechanism: https://github.com/libp2p/test-plans/blob/34b6eaeffd9f4a011e41ad8b2d8103229e750c12/ping/rust/Dockerfile#L4

I'd say that is an argument for doing it the same way here as well because it guarantees us that what we test interoperability on is actually the same thing as we write our unit tests etc for.

mxinden (approved):
I am fine moving forward here. Issues due to the protoc version should be detected at compile time, e.g. the proto3 optional issue would result in a compile time failure as our Rust code would expect an `Option`. https://github.com/libp2p/specs/issues/465#issuecomment-1272291524

mxinden:
@Mergifyio refresh

thomaseizinger:
I am backporting this to v0.49 because the actual backport is otherwise super painful.

thomaseizinger:
@mergifyio backport rust-v0490-testplan

PR #3085, opened 2022-11-04
swarm/handler: replace inject_* methods

jxs:
## Description

Previously, we had one callback for each kind of message that a `ConnectionHandler` would receive from either its `NetworkBehaviour` or the connection itself.

With this patch, we combine these functions, resulting in two callbacks:

- `on_behaviour_event`
- `on_connection_event`

Resolves #3080.

## Links to any relevant issues
`ConnectionHandler` counterpart to #3011 

## Change checklist

- [ ] I have performed a self-review of my own code
- [ ] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] A changelog entry has been made in the appropriate crates

PR #3264, opened 2022-12-20
refactor(swarm): remove deprecated inject calls

jxs:
## Description

Finishes work first started with https://github.com/libp2p/rust-libp2p/issues/2832

## Notes

review by commit is suggested :)

## Change checklist


protocols/rendezvous/src/substream_handler.rs
```diff
@@ -367,35 +366,47 @@ where
         TInboundSubstreamHandler::upgrade(())
     }
 
-    fn inject_fully_negotiated_inbound(
+    fn on_connection_event(
```
jxs, line 369: this was missed in the previous #3085 sorry, should I add a new changelog entry CC @mxinden? (there was already one in the previous version mentioning the replacement of the methods for `Handler`)

swarm/src/handler.rs
```diff
@@ -279,18 +183,17 @@ pub trait ConnectionHandler: Send + 'static {
     }
 
     /// Informs the handler about an event from the [`NetworkBehaviour`](super::NetworkBehaviour).
-    fn on_behaviour_event(&mut self, _event: Self::InEvent) {}
+    fn on_behaviour_event(&mut self, _event: Self::InEvent);
 
     fn on_connection_event(
```
thomaseizinger, line 188: This one is missing a doc comment!

swarm/src/handler/either.rs
```diff
@@ -178,102 +176,144 @@ where
                 protocol: output,
                 info,
             }) => match (self, output, info) {
-                (Either::Left(handler), EitherOutput::First(output), Either::Left(info)) =>
-                {
-                    #[allow(deprecated)]
-                    handler.inject_fully_negotiated_outbound(output, info)
+                (Either::Left(handler), EitherOutput::First(output), Either::Left(info)) => {
+                    handler.on_connection_event(ConnectionEvent::FullyNegotiatedOutbound(
```
thomaseizinger: It seems weird that we need to match on the event here (line 174). I think it would be better if we added a

```rust
impl ConnectionEvent<EitherOutput, EitherOutput, Either, Either> { }
```

block where we map the entire event into a `Either<ConnectionEvent, ConnectionEvent>` so you can directly dispatch the event to the left or the right handler.

swarm/src/handler/select.rs
```diff
@@ -165,60 +169,72 @@ where
         >,
     ) {
         match (info, error) {
```
thomaseizinger: Same here, I think we could make use of the same mapping function from above here.

thomaseizinger (commented):
Nice, thanks for following up on that!

I think we can simplify some of this a bit. Let me know if you need more input :)

swarm-derive/src/lib.rs
```diff
@@ -188,11 +188,23 @@ fn build_struct(ast: &DeriveInput, data_struct: &DataStruct) -> TokenStream {
             .enumerate()
             .map(|(field_n, field)| match field.ident {
                 Some(ref i) => quote! {
-                    #[allow(deprecated)]
-                    self.#i.inject_connection_established(&peer_id, &connection_id, endpoint, Some(&failed_addresses.into()), other_established);},
+                    self.#i.on_swarm_event(#from_swarm::ConnectionEstablished(#connection_established {
```
mxinden, line 191: Needs a minor (breaking) version bump in `libp2p-swarm-derive`.

swarm/src/behaviour.rs
```diff
@@ -159,7 +160,7 @@ pub trait NetworkBehaviour: 'static {
     }
 
     /// Informs the behaviour about an event from the [`Swarm`](crate::Swarm).
-    fn on_swarm_event(&mut self, _event: FromSwarm<Self::ConnectionHandler>) {}
+    fn on_swarm_event(&mut self, event: FromSwarm<Self::ConnectionHandler>);
```
mxinden, line 163: :+1:

mxinden (commented):
Thanks for tackling this.

swarm/src/handler/either.rs
jxs: yeah, that would be great. I tried:
```
impl<'a, R, L>
    ConnectionEvent<
        'a,
        EitherUpgrade<SendWrapper<L::InboundProtocol>, SendWrapper<R::InboundProtocol>>,
        EitherUpgrade<SendWrapper<L::OutboundProtocol>, SendWrapper<R::OutboundProtocol>>,
        Either<L::InboundOpenInfo, R::InboundOpenInfo>,
        Either<L::OutboundOpenInfo, R::OutboundOpenInfo>,
    >
where
    L: ConnectionHandler,
    R: ConnectionHandler,
{}
```
but get that `L` and `R` are not constrained.  I wonder if https://github.com/libp2p/rust-libp2p/pull/3085#discussion_r1016988595 would have changed that.

swarm-derive/src/lib.rs
jxs: thanks. updated!

swarm/src/handler.rs
jxs: thanks Thomas! Addressed.

swarm/src/handler/either.rs
thomaseizinger: I was thinking something like https://github.com/libp2p/rust-libp2p/commit/e3480d2db8916e7c9ba0115126880ea11d699f71. It doesn't compile yet but maybe you can make it work!

protocols/rendezvous/src/substream_handler.rs
mxinden: I don't think this change is breaking, right? Thus advocating for no changelog.

On a related note, shouldn't our CI have caught this? Shouldn't this have generated a deprecation warning and thus shouldn't CI have failed? Can you investigate this @jxs?

swarm-derive/CHANGELOG.md
```diff
@@ -1,15 +1,19 @@
-# 0.31.0
+# 0.32.0 [unreleased]
 
 - Replace `NetworkBehaviour` Derive macro deprecated `inject_*` method implementations
   with the new `on_swarm_event` and `on_connection_handler_event`.
```
mxinden, line 4: This entry was previously mistakenly in `v0.31.0`, right? I.e. we did not actually move to the `on_` methods in `swarm-derive` in `v0.31.0` as otherwise our non-breaking-change hack would not have worked, correct?

mxinden (commented):
Thanks for the follow-ups.

mxinden (approved):
Please trigger mergify once the above discussion with @thomaseizinger is resolved.

swarm/src/handler/either.rs
jxs: thanks for the lead Thomas, managed to make it work. Used `Either::Left` for `AddressChange` instead of the `Result` as it seemed more semantically accurate in the sense that it's not err, it just doesn't matter if it's either `Left` or `Right` for that variant, cause when matching on `on_connection_event` we still have the `Either::Right` . Ptal and see if you agree.

swarm/src/handler/select.rs
jxs: Addressed, see if that's what you were expecting Thomas

swarm-derive/CHANGELOG.md
jxs: yeah exactly, thanks for referring this Max.

protocols/rendezvous/src/substream_handler.rs
jxs: yeah this isn't a breaking change afaik, as they are only called by the `Behaviour` 

>On a related note, shouldn't our CI have caught this? Shouldn't this have generated a deprecation warning and thus shouldn't CI have failed?

good point, `cargo` doesn't trigger deprecation warning on the implementation, only on the invocation IRC, do we have any code triggering calls to the methods?

swarm/src/handler/either.rs
thomaseizinger: > thanks for the lead Thomas, managed to make it work. Used `Either::Left` for `AddressChange` instead of the `Result` as it seemed more semantically accurate in the sense that it's not err, it just doesn't matter if it's either `Left` or `Right` for that variant, cause when matching on `on_connection_event` we still have the `Either::Right` . Ptal and see if you agree.

Yeah I wasn't sure about that one. I think one could argue it is an error because we couldn't unambiguously transpose it. Can we make it work with `Result` or is moving to `Either::Left` essential?

Alternatively, we could wrap it in another `Either` but that is a bit weird. If possible, I'd prefer the `Result` return type and some documentation on why it is necessary.

swarm/src/handler/select.rs
```diff
@@ -98,6 +99,80 @@ impl<TProto1, TProto2> ConnectionHandlerSelect<TProto1, TProto2> {
...
+        EitherOutput<S1OOI, S2OOI>,
+        EitherUpgrade<SendWrapper<S1OP>, SendWrapper<S2OP>>,
+    >
+where
+    S1OP: OutboundUpgradeSend,
+    S2OP: OutboundUpgradeSend,
+    S1OOI: Send + 'static,
+    S2OOI: Send + 'static,
+{
+    fn transpose(self) -> Either<DialUpgradeError<S1OOI, S1OP>, DialUpgradeError<S2OOI, S2OP>> {
```
thomaseizinger, line 164: I like that you split this out. I actually think it makes sense to split this out for all structs of `ConnectionEvent`. Can we reuse this as part of implementing the bigger `transpose` function?

swarm/src/handler/either.rs
jxs: no you are right, documenting it informing that an `Err` can be because `AddressChange` isn't transposable makes sense to me, updated it :)

swarm/src/handler/select.rs
jxs: Ah, thanks :D but I did it because in some variants like `ListenUpgradeError` or `AddressChange` transpose doesn't win us anything as we have to call it for both `Select` variants. I updated the code implementing for the variants where it makes sense, ptal Thomas and see if it makes sense to you.

swarm/src/handler/either.rs
```diff
@@ -90,6 +92,207 @@ impl<L, R> IntoEitherHandler<L, R> {
...
+        >,
+        AddressChange<'a>,
+    > {
+        match self {
+            ConnectionEvent::FullyNegotiatedInbound(FullyNegotiatedInbound {
+                protocol: EitherOutput::First(protocol),
+                info: Either::Left(info),
+            }) => Ok(Either::Left(ConnectionEvent::FullyNegotiatedInbound(
+                FullyNegotiatedInbound { protocol, info },
+            ))),
```
thomaseizinger:
```suggestion
            ConnectionEvent::FullyNegotiatedInbound(inner) => Ok(inner.transpose()),
```

Does this work? Or do we have to call `.map` on it again or something?

It would be cool if we can express this `transpose` function somehow through the smaller ones.

jxs: It doesn't because `Select`'s implementation uses `SelectUpgrade` whereas `Either` uses `EitherUpgrade`, do you have any idea how can we combine them in `transpose`? :P

jxs:
@thomaseizinger submitted #3307 which started as fork from this one and also addresses some of the conversations here. Can we merge this one and continue there? I can also open an issue to address potential unresolved conversations.

swarm/CHANGELOG.md
```diff
@@ -7,12 +7,18 @@
 
 - Add `estblished_in` to `SwarmEvent::ConnectionEstablished`. See [PR 3134].
 
+- Remove deprecated `inject_*` methods from `NetworkBehaviour` and `ConnectionHandler`.
+  see [PR 3260].
```
```suggestion
  see [PR 3264].
```

Wrong issue reference :)

swarm/src/handler/either.rs
thomaseizinger: I don't think we can combine them for select and either because those are fundamentally different.

thomaseizinger: Also, one of them is using `EitherOutput` which we need because `Either` doesn't implement `AsyncRead` + `AsyncWrite`.

See https://github.com/libp2p/rust-libp2p/issues/2650.

swarm/src/handler/select.rs
```diff
@@ -430,19 +438,45 @@ where
...
-                self.on_dial_upgrade_error(dial_upgrade_error)
+                match dial_upgrade_error.transpose() {
+                    Either::Left(err) => self
+                        .proto1
+                        .on_connection_event(ConnectionEvent::DialUpgradeError(err)),
+                    Either::Right(err) => self
+                        .proto2
+                        .on_connection_event(ConnectionEvent::DialUpgradeError(err)),
+                }
             }
```
thomaseizinger, line 480: I wonder if it would be better to do it like this for the `either::Handler` as well? i.e. only transpose the inner type instead of the entire event and matching on the tuple.

thomaseizinger (approved):
A few more comments, overall looks good to me!

swarm/src/handler/select.rs
jxs: Yeah you are right Thomas, you had already mentioned before sorry it slipped me. It makes sense cause then we don't need `Err` for `AddressChange`. Updated ptal!

thomaseizinger (approved):
LGTM

Nice to see it being a negative diff!

PR #3265, opened 2022-12-20
feat(noise): deprecate `LegacyConfig` without replacement

thomaseizinger:
## Description

As the name implies, `LegacyConfig` allows users to interact with older versions of the noise protocol. These are not interoperable and we've been supporting them for a long time now. Hopefully, users have migrated away from it since. To not directly break them, we officially deprecate now without a replacement.

## Notes

Raised by @mxinden in https://github.com/libp2p/rust-libp2p/pull/3227#pullrequestreview-1222536273.

## Change checklist


jxs (approved):

mxinden (approved):
Deprecation first sounds ideal to me.

I suggest adding a short changelog entry never the less. What do you think @thomaseizinger?

thomaseizinger:
https://github.com/libp2p/rust-libp2p/commit/929cbb4670846dfec9439c4dafd0c7705376306c#diff-022f11eec986ac7c281f192c0e10064513ee5ef6ad2ab45d5dc0c3bbdc617b2f is another good example that our changelog management is flawed and prone to errors.

PR #3272, opened 2022-12-21
refactor(swarm)!: don't be generic over `Transport`

thomaseizinger:
## Description

Ever since we moved `Pool` into `libp2p-swarm`, we always use it with the same `Transport`: `Boxed`. It is thus unnecessary for us to be overly generic over what kind of `Transport` we are using. This allows us to remove a few type parameters from the implementation which overall simplifies things.

This is technically a breaking change because I am removing a type parameter from two exported type aliases:

- `PendingInboundConnectionError`
- `PendingOutboundConnectionError`

Those have always only be used with `std::io::Error` in our API but it is still a breaking change.

## Notes

Discovered as part of working on #2824.

## Change checklist

### DIFF
--- a/swarm/CHANGELOG.md
+++ b/swarm/CHANGELOG.md
@@ -7,9 +7,13 @@
 
 - Add `estblished_in` to `SwarmEvent::ConnectionEstablished`. See [PR 3134].
 
+- Remove type parameter from `PendingOutboundConnectionError` and `PendingInboundConnectionError`.
+  These two types are always used with `std::io::Error`. See [PR XXXX.
+
 [PR 3170]: https://github.com/libp2p/rust-libp2p/pull/3170
 [PR 3134]: https://github.com/libp2p/rust-libp2p/pull/3134
 [PR 3153]: https://github.com/libp2p/rust-libp2p/pull/3153
+[PR XXXX]: https://github.com/libp2p/rust-libp2p/pull/XXXX
 
 # 0.41.1
 
--- a/swarm/CHANGELOG.md
+++ b/swarm/CHANGELOG.md
@@ -8,12 +8,12 @@
 - Add `estblished_in` to `SwarmEvent::ConnectionEstablished`. See [PR 3134].
 
 - Remove type parameter from `PendingOutboundConnectionError` and `PendingInboundConnectionError`.
-  These two types are always used with `std::io::Error`. See [PR XXXX.
+  These two types are always used with `std::io::Error`. See [PR 3272].
 
 [PR 3170]: https://github.com/libp2p/rust-libp2p/pull/3170
 [PR 3134]: https://github.com/libp2p/rust-libp2p/pull/3134
 [PR 3153]: https://github.com/libp2p/rust-libp2p/pull/3153
-[PR XXXX]: https://github.com/libp2p/rust-libp2p/pull/XXXX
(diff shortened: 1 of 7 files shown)


swarm/CHANGELOG.md
```diff
@@ -7,9 +7,13 @@
 
 - Add `estblished_in` to `SwarmEvent::ConnectionEstablished`. See [PR 3134].
 
+- Remove type parameter from `PendingOutboundConnectionError` and `PendingInboundConnectionError`.
+  These two types are always used with `std::io::Error`. See [PR XXXX.
+
 [PR 3170]: https://github.com/libp2p/rust-libp2p/pull/3170
 [PR 3134]: https://github.com/libp2p/rust-libp2p/pull/3134
 [PR 3153]: https://github.com/libp2p/rust-libp2p/pull/3153
+[PR XXXX]: https://github.com/libp2p/rust-libp2p/pull/XXXX
```
```suggestion
  These two types are always used with `std::io::Error`. See [PR 3272].

[PR 3170]: https://github.com/libp2p/rust-libp2p/pull/3170
[PR 3134]: https://github.com/libp2p/rust-libp2p/pull/3134
[PR 3153]: https://github.com/libp2p/rust-libp2p/pull/3153
[PR 3272]: https://github.com/libp2p/rust-libp2p/pull/3272
```

swarm/src/connection/pool.rs
```diff
@@ -79,9 +79,8 @@ impl ExecSwitch {
 }
 
 /// A connection `Pool` manages a set of connections for each peer.
-pub struct Pool<THandler, TTrans>
```
jxs, line 82: what do you think of making this `pub (crate)` instead? I know that it doesn't change anything as `pool` is not public but it helps visually understand that the type is shared across the crate (as opposed to just being private, which it can't be). `pool` is also `pub (crate)`. 
(this suggestion also applies to `PoolEvent`)

jxs (approved):
LGTM seems like a nice simplification, thanks Thomas!

mxinden (commented):
I am in favor of this change.

mxinden (approved):
Looks good to me. Leaving this open for the above discussion by @jxs.

swarm/src/connection/pool.rs
thomaseizinger: In general I think I'd prefer to not use `pub(crate)` but I could be convinced otherwise.

It would be nice if we can assume that:

- All modules are private
- pub types therefore don't leak into the public API
- pub(crate) is therefore a workaround because rhe module isn't private for some reason

I am not fully convinced it is a good idea though, might cause a lot of complexity.

thomaseizinger: I wish Rust had a way to whitelist what is in the public API of a crate and fail to compile if something is being added.

thomaseizinger:
> Looks good to me. Leaving this open for the above discussion by @jxs.

In case you both don't mind, I am sending this to master so work on #3254 can continue. Let's keep the discussion going though! :)

swarm/src/connection/pool.rs
jxs: if we decide that every `pub` type follows the convention used in #3238 we could use `pub (crate)` only inside the the crate itself to distinguish from the actually `pub` types on the main `lib.rs` for example, one issue with the assumptions above for example is that we have `pub mod` but in the end no strong opinion, I can understand your POV as well and share the same frustrations.

PR #3273, opened 2022-12-21
feat(swarm): introduce `ToSwarm` type alias

### LINKED ISSUE #3123: swarm: Rename `NetworkBehaviourAction` to `ToSwarm`
## Description

Similar to https://github.com/libp2p/rust-libp2p/issues/2854, what do people think of renaming `NetworkBehaviourAction` to `ToSwarm`?

## Motivation

Consistent naming of messages between our "actors".

## Current Implementation

The current naming is very inconsistent which makes things hard to understand.

Maybe.

thomaseizinger:
## Description

The `NetworkBehaviourAction` type is often clunky to name because it has several type parameters. Usually, those can all be inferred from the corresponding `NetworkBehaviour`. However, that would require us to add bounds to the type definition which generally not a good idea because those bounds creep easily into upper layers where you don't want them.

This PR instead adds a type alias for `NetworkBehaviourAction` called `ToSwarm` which is just a `NetworkBehaviourAction` with all parameters inferred from a `NetworkBehaviour`.

Resolves https://github.com/libp2p/rust-libp2p/issues/3123.

## Notes

At the moment, this is just an experiment.

This might be easier after we merge https://github.com/libp2p/rust-libp2p/pull/3254 where we remove the `THandler` type parameter from `NetworkBehaviourAction`.

At the moment, this only compiles by adding type annotations everywhere which is kind of unacceptable. I opened an issue on rust-lang/rust for that: https://github.com/rust-lang/rust/issues/105680.

Depends-On: https://github.com/libp2p/rust-libp2p/pull/3239 to remove the build failures in `libp2p-kad`.

## Change checklist

### DIFF
--- a/protocols/autonat/src/behaviour.rs
+++ b/protocols/autonat/src/behaviour.rs
@@ -209,7 +209,12 @@ pub struct Behaviour {
 
     last_probe: Option<Instant>,
 
-    pending_out_events: VecDeque<<Self as NetworkBehaviour>::OutEvent>,
+    pending_actions: VecDeque<
+        NetworkBehaviourAction<
+            <Self as NetworkBehaviour>::OutEvent,
+            <Self as NetworkBehaviour>::ConnectionHandler,
+        >,
+    >,
 
     probe_id: ProbeId,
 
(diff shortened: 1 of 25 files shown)


thomaseizinger:
Too many merge conflicts, might pursue in the future.

Issue #3236, opened 2022-12-13
kademlia: Records republication exceeds inbound stream limit


shamil-gadelshin:
## Summary

We have a feature in Kademlia for provider records republication which is controlled by `provider_publication_interval` configuration parameter.

However, if we have multiple provider records (several hundred) that should be republished we could experience "inbound stream issues" because Kademlia has a hard-coded limit (`MAX_NUM_INBOUND_SUBSTREAMS = 32)`. 

Some of the messages relate to "reusing of the inbound streams" and [seem OK](https://github.com/libp2p/rust-libp2p/pull/3234). Other relate to dropped streams. 

Currently, we don't have means to limit rates of the "republication" requests because it's a built-in protocol feature. Am I wrong here? 

## Expected behaviour

I expect an internal Kademlia process to work within Kademlia's limits or have a setting to control its rate.

## Actual behaviour

Peers exceed other peers` inbound stream limits and produce warnings and actually drop republication messages.

<details><summary>Debug Output</summary>
<p>

```
2022-12-13T12:54:30.637379Z  WARN libp2p_kad::handler: New inbound substream to PeerId("12D3KooWFUrJybJKR4EkeXY3WPknWJ25RtZU4XoXivCTvQsqFHCw") exceeds inbound substream limit. No older substream waiting to be reused. Dropping new substream.    
2022-12-13T12:54:30.637470Z  WARN libp2p_kad::handler: New inbound substream to PeerId("12D3KooWFUrJybJKR4EkeXY3WPknWJ25RtZU4XoXivCTvQsqFHCw") exceeds inbound substream limit. No older substream waiting to be reused. Dropping new substream.    
2022-12-13T12:54:30.637557Z  WARN libp2p_kad::handler: New inbound substream to PeerId("12D3KooWFUrJybJKR4EkeXY3WPknWJ25RtZU4XoXivCTvQsqFHCw") exceeds inbound substream limit. No older substream waiting to be reused. Dropping new substream.    
2022-12-13T12:54:30.637644Z  WARN libp2p_kad::handler: New inbound substream to PeerId("12D3KooWFUrJybJKR4EkeXY3WPknWJ25RtZU4XoXivCTvQsqFHCw") exceeds inbound substream limit. No older substream waiting to be reused. Dropping new substream.    
[...]

PR #3285, opened 2022-12-27
feat(kad): Limit in-flight Kademlia queries

nazar-pc:
## Description

Limit number of in-flight Kademlia queries to not exceed configured number of substreams

## Notes

I went with a design where each query is first inserted into pending queue and `QueryPool` will maintain at most `32` queries active at any time, any queries that exceed the limit will remain in pending queue.

In order to guarantee insertion order of pending queries I have introduced an abstraction in second commit.

Since constant restricting inbound streams to `32` is now used for outbound too, I moved it up to the root of the crate and changed name to more generic.

https://github.com/libp2p/rust-libp2p/discussions/3235
https://github.com/libp2p/rust-libp2p/issues/3236

## Open Questions

Not entirely sure how to test higher-level behavior here.

## Change checklist

### DIFF
--- a/protocols/kad/CHANGELOG.md
+++ b/protocols/kad/CHANGELOG.md
@@ -6,9 +6,12 @@
 
 - Remove lifetime from `RecordStore` and use GATs instead. See [PR 3239].
 
+- Limit in-flight Kademlia queries. See [PR 3285].
+
 - Bump MSRV to 1.65.0.
 
 [PR 3239]: https://github.com/libp2p/rust-libp2p/pull/3239
+[PR 3285]: https://github.com/libp2p/rust-libp2p/pull/3285
 
 # 0.42.1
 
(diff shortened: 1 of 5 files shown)


nazar-pc:
I just noticed that performance is severely limited because limit is enforced on Kademlia globally, while it needs to be on per-connection basis, changing to draft for now.

dignifiedquire:
How does this deal with queries being canceled before they have been started? (this is a very relevant use case for my code, where I have run into limits before)

nazar-pc:
It will be removed as if it was cancelled before the first `.poll()`, so there are no fundamental mechanics changes here.

PR as is does achieve the goal, but with poor performance due to global limit. I'll dig into making limits per-connection instead tomorrow and will update PR once I'm there.

nazar-pc:
Replaced with https://github.com/libp2p/rust-libp2p/pull/3287 that works on handler level instead

PR #3288, opened 2022-12-28
fix(quic): Identify `/quic` as QUIC address

mxinden:
## Description

Identify multiaddress with `/quic` (draft 29) as QUIC address in case `support_draft_29` is `true`.

Without this patch the Rust punchr client would discard any QUIC addresses with `/quic` in its `Transport::address_translation`. Thus `/quic` based observed addresses from `libp2p-identify` would not be added to the local set of external addresses and thus QUIC would not be available as a transport for hole punching.

https://github.com/libp2p/rust-libp2p/pull/3151 introduced draft-29 support.

## Change checklist

- [ ] ~~A changelog entry has been made in the appropriate crates~~ Not necessary as it hasn't yet been released.


elenaf9 (approved):
Thanks @mxinden!

Issue #3268, opened 2022-12-20
swarm: Deprecate `ConnectionHandler::{In,Out}boundOpenInfo`


thomaseizinger:
Currently, a user can specify an arbitrary payload type that can be passed a long with the opening of an inbound or an outbound stream. In general, being able to pass along user data with an asynchronous operation can be helpful because one doesn't need to track this information locally. Furthermore, for some asynchronous operations (like connection dialing in the case of relay's and the dcutr protocol) it is very important that one can 1-to-1 associate the initiated action with the result of the asynchronous operation.

Streams however are completely interchangeable in libp2p once the protocol to be spoken on top of it has been negotiated. Thus, we should be able to remove `ConnectionHandler::InboundOpenInfo` and `ConnectionHandler::OutboundOpenInfo` completely in favor of `ConnectionHandler`s tracking the user data in a FIFO queue.

In fact, that is what we are doing within `swarm::Connection` anyway. Upon opening a new stream, we just associate it with the next upgrade from the list: https://github.com/libp2p/rust-libp2p/blob/ee775137f30c19dfe43598400b1dda9fcdd46f51/swarm/src/connection.rs#L289-L303

```[tasklist]
- [ ] https://github.com/libp2p/rust-libp2p/pull/3760
- [ ] https://github.com/libp2p/rust-libp2p/pull/3762
- [ ] https://github.com/libp2p/rust-libp2p/pull/3763
- [ ] https://github.com/libp2p/rust-libp2p/issues/4075
- [ ] https://github.com/libp2p/rust-libp2p/pull/3914
- [ ] https://github.com/libp2p/rust-libp2p/issues/4510
- [ ] https://github.com/libp2p/rust-libp2p/pull/4901
- [ ] #4900
- [ ] Write tutorial on how to write your own `ConnectionHandler`, documenting use of `futures-bounded` and oneshots
```

thomaseizinger:
With the :+1: from @mxinden and @elenaf9, I am removing the `decision-pending` label and putting this up for grabs.

thomaseizinger:
Linking the original discussion here: https://github.com/libp2p/rust-libp2p/issues/1709

[...]

PR #3290, opened 2022-12-30
docs(coding-guidelines): Add request response correlation

mxinden:
## Description

Allow correlating asynchronous responses to their requests.

Came up in discussions around https://github.com/libp2p/rust-libp2p/issues/2824.

## Change checklist

- [ ] ~~I have added tests that prove my fix is effective or that my feature works~~
- [ ] ~~A changelog entry has been made in the appropriate crates~~


jxs (approved):

thomaseizinger (changes requested):
Thanks for writing this up!

I think we should settle on one approach on how we do this in the codebase, otherwise it is not really a guideline.

My preference would be client-side generated IDs for commands:

They are a special case of "user data" just not being configurable, thus they can carry the same usecases but without the downsides of type parameters which make the code complex.

I am not sure about methods. For https://github.com/libp2p/rust-libp2p/pull/3292, we would need client-side generated IDs but we can't do that if we have a more low-level component that hands out IDs.

thomaseizinger:
We might also want to add that this only applies to certain things. Streams for example should not have an identity but be interchangable in my opinion.

See https://github.com/libp2p/rust-libp2p/issues/3268.

mxinden:
> I think we should settle on one approach on how we do this in the codebase, otherwise it is not really a guideline.

I am assuming that you want to decide on either userdata or user-provided-ids, correct?

I don't feel strongly that we should be doing either or. In other words, I am fine using both in the rust-libp2p codebase, depending on the use-case.

> I am not sure about methods. For #3292, we would need client-side generated IDs but we can't do that if we have a more low-level component that hands out IDs.

Thanks for bringing this up. for `ListenOn` we could later on return an event with both the listener ID and the user-provided-id. For `RemoveListener` the listener ID might suffice.

docs/coding-guidelines.md
```diff
@@ -299,3 +300,55 @@ stack potentially unboundedly. Instead use iteration e.g. via `loop` or `for`.
 - https://en.wikipedia.org/wiki/Tail_call
 - https://stackoverflow.com/questions/65948553/why-is-recursion-not-suggested-in-rust
 - https://stackoverflow.com/questions/59257543/when-is-tail-recursion-guaranteed-in-rust
+
+## Allow Correlating Asynchronous Responses to Their Requests
+
+In an asynchronous context, it is important to enable users to determine the correlation between a
+response and a previous request. For example, if a user requests two new connections to the same
+peer, they should be able to match each new connection to the corresponding previous connection
+request without having to guess.
```
elenaf9, line 309: I agree with the general guideline, but I am not sure if I agree with this example. Connections can still be differentiated by the `ConnectedPoint`. And if that's also the same, why does it matter what connection was initiated by what request?

thomaseizinger:
> > I think we should settle on one approach on how we do this in the codebase, otherwise it is not really a guideline.
> 
> I am assuming that you want to decide on either userdata or user-provided-ids, correct?

Yes, I'd very much appreciate consistency here. I think it will make it easier to understand the codebase and allow us to better settle on reusable components.

> > I am not sure about methods. For #3292, we would need client-side generated IDs but we can't do that if we have a more low-level component that hands out IDs.
> 
> Thanks for bringing this up. for `ListenOn` we could later on return an event with both the listener ID and the user-provided-id. For `RemoveListener` the listener ID might suffice.

Ugh, not sure about two IDs 🫤

I think I'd prefer if we require the user to pass a `ListenerId` into the `Transport`. Those APIs shouldn't really be used directly anyway so the impact would be minimal.

docs/coding-guidelines.md
```diff
@@ -299,3 +300,55 @@ stack potentially unboundedly. Instead use iteration e.g. via `loop` or `for`.
...
+``` rust
+fn my_method() -> Id {
+  // ...
+}
+```
+
+``` rust
+fn my_method() -> impl Future<Output = Response> {
+  // ...
+}
```
elenaf9: Are we doing this anywhere in rust-libp2p, apart from `Transport::dial`?

docs/coding-guidelines.md
```diff
@@ -299,3 +300,55 @@ stack potentially unboundedly. Instead use iteration e.g. via `loop` or `for`.
...
+```
+
+When accepting a **command** that eventually results in a response through an event, either require
+that command to contain a unique ID, which is later on contained in the asynchronous response event,
+or allow the user to provide an instance of a custom type which again is later on returned in the
+response event.
```
elenaf9: Could we briefly elaborate where such commands are used rust-libp2p? I would consider the events from `NetworkBehaviour` -> `Handler` as such commands, or did you have something different in mind?

docs/coding-guidelines.md
```diff
@@ -299,3 +300,55 @@ stack potentially unboundedly. Instead use iteration e.g. via `loop` or `for`.
...
+
+When providing a **method** where the response to that method is delivered asynchronously through an
+event, either synchronously return a request ID which is later on contained in the asynchronous
+response event, or synchronously return a `Future` that eventually resolves into the response.
+
+``` rust
+fn my_method() -> Id {
+  // ...
+}
+```
```
elenaf9: Compared to the `Command` below the ID here is not user generated but instead handed out to the user. When would it be recommended to use user-generated IDs and when should they be handed out?

elenaf9 (commented):
Thanks for writing this up @mxinden. 
Right now, this describes different approaches to solve more or less the same thing. I think it would be useful if there would be some additional notes on when it makes sense to use what.

docs/coding-guidelines.md
mxinden: Say that I have two `NetworkBehaviour`s `A` and `B`. Both dial a peer concurrently. `A` requests a dial to the peer with a set of addresses. `B` requests a dial to the peer with a set of addresses. `A` keeps some state during the time of the dial. `A` is notified that a connection request to the peer failed. Should `A` delete the state? Maybe this was the dial of `B` with the wrong set of addresses.

docs/coding-guidelines.md
mxinden: I don't have a good rule here, nor an opinion. Thus far I was simply deciding case-by-case.

docs/coding-guidelines.md
mxinden: `Tranport::dial` is the only one that comes to my mind. That said, I would like to experiment with it on the API surface of `NetworkBehaviour` implementations. E.g. a `NotBitswap` `NetworkBehaviour` implementation providing a `request_cid` method that returns a `Future`.

I don't have a strong preference, nor a rule for either of the two (return ID or return Future). Thus I suggest keeping both options open. Objections?

mxinden:
a951fbdf and 694a09ce remove the user-data option. Thanks for pushing for it @thomaseizinger and @elenaf9.

docs/coding-guidelines.md
mxinden: 694a09ce adds the example of `NetworkBehaviourAction` from `NetworkBehaviour` to `Swarm`. `NetworkBehaviour` -> `Handler` is an option as well, though in my eyes the less ideal example as it has the indirection through `Swarm`.

docs/coding-guidelines.md
```diff
@@ -299,3 +300,42 @@ stack potentially unboundedly. Instead use iteration e.g. via `loop` or `for`.
...
```
thomaseizinger: Would it make sense to remove this and be consistent with commands where the caller always specifies the ID? Only `Transport::dial` currently behaves like stated here and we already said we wanted to fix this! :)

I'd be in favor of removing this from the guidelines (or rather, say that we want to accept the ID) and opening an issue to fix `Transport::dial`.

mxinden: > with commands where the caller always specifies the ID?

I have to give this more thought. In other words, I am not sure this is the way to go in all situations.

Do you feel strongly about removing this @thomaseizinger? I would prefer proceeding here. I see this pull request as a net-positive at this point.

thomaseizinger: I think it makes sense for anything that is public API in our library. Internally, where we control all usages of an ID, it doesn't matter as much.

The reason for that is that the majority of our public API is command-based. Anything command-based needs to create the ID ahead of time. Any component underneath that creates its own IDs causes friction here because it doesn't compose well with commands.

thomaseizinger: I am a bit torn. There is certainly value in documenting this but if we document something, I'd prefer if we can back the written words 100%, e.g. have strong convictions that this is the way to go. Everything else is just confusing to the reader.

thomaseizinger: Would removing this section be a middle ground? i.e. only document that commands should come with a user-generated ID and not talk about functions at all?

mxinden: Done with 12e61b63.

mxinden:
Addressed latest comment. @thomaseizinger please add the `send-it` label if you like the pull request as is.

thomaseizinger (approved):

PR #3292, opened 2022-12-31
feat(swarm): allow `NetworkBehaviour`s to create and remove listeners

### LINKED ISSUE #3291: Extend NetworkBehaviourAction to add and remove listeners
## Description

Extend `NetworkBehaviourAction` with `NetworkBehaviourAction::ListenOn` and `NetworkBehaviourAction::RemoveListener` to allow adding or removing a listener within the `NetworkBehaviour`

## Motivation
 

One project that I am working on is an attempt at connecting to a relay automatically, which could benefit from this by emitting an event to listen on the relay and be able to remove it by using the listener id later if needed (eg relay becoming unreliable). Current workaround is emitting an event which is then seen when polling the swarm to then listen to  (and soon remove)  the address there. 

Maybe

dariusc93:
## Description

This extends `ToSwarm` to add `ToSwarm::ListenOn` and `ToSwarm::RemoveListener`, which allows creating and removing listeners from a `NetworkBehaviour`.

Resolves https://github.com/libp2p/rust-libp2p/issues/3291.

## Change checklist


swarm/src/behaviour.rs
```diff
@@ -630,7 +630,20 @@ pub enum NetworkBehaviourAction<
...
-    Dial { opts: DialOpts, handler: THandler },
+    Dial {
+        opts: DialOpts,
+        handler: THandler,
+    },
+
+    /// Instructs the `Swarm` to listen on the provided address
+    ListenOn {
+        address: Multiaddr,
+    },
```
thomaseizinger: With what we figured out in the discussion around #2824, we should provide a mechanism on how to associate a listening request with the later on produced events.

To follow what we will likely end up doing in https://github.com/libp2p/rust-libp2p/pull/3254, I think we should pass the `ListenerId` into here already. That is a bit tricky with the current `Transport` interface though.

thomaseizinger: Related: https://github.com/libp2p/rust-libp2p/pull/3290

thomaseizinger (changes requested):
Overall I am in support of this but we should probably wait until we have the connection management story figured out or at least found consensus on https://github.com/libp2p/rust-libp2p/pull/3290.

mxinden:
@dariusc93 note that the connection management work, namely https://github.com/libp2p/rust-libp2p/pull/3254, is merged, thus no longer blocking this pull request.

dariusc93:
> @dariusc93 note that the connection management work, namely #3254, is merged, thus no longer blocking this pull request.

Thank you for letting me know. I would have to double check, but was it ever decided on rather a `ListenerId` would be supplied to the `Transport` through the event or would the current changes in this PR suffice for the time being?

thomaseizinger:
> > @dariusc93 note that the connection management work, namely #3254, is merged, thus no longer blocking this pull request.
> 
> Thank you for letting me know. I would have to double check, but was it ever decided on rather a `ListenerId` would be supplied to the `Transport` through the event or would the current changes in this PR suffice for the time being? 

I think we will need to issue the `ListenerId` as part of the command. Not only will this make things more consistent, it is also the only way how a `NetworkBehaviour` can effectively manage listeners and not having to reverse-engineer them from the addresses.

dariusc93:
> > > @dariusc93 note that the connection management work, namely #3254, is merged, thus no longer blocking this pull request.
> > 
> > 
> > Thank you for letting me know. I would have to double check, but was it ever decided on rather a `ListenerId` would be supplied to the `Transport` through the event or would the current changes in this PR suffice for the time being?
> 
> I think we will need to issue the `ListenerId` as part of the command. Not only will this make things more consistent, it is also the only way how a `NetworkBehaviour` can effectively manage listeners and not having to reverse-engineer them from the addresses.

Providing the `ListenerId` should be simple enough in this case (and honestly would be better for tracking, than trying to sort it out, like you said). I would then take it that instead of using `Swarm::listen_on` in `Swarm::handle_behaviour_event` in my PR, I would use the transport itself with the provided id, while in `Swarm::listen_on`, for the sake of compatibility, to generate a `ListenerId` there and pass it onto the transport? 

This can be done outside of this PR too in its own separate PR 

EDIT: I made a PR at https://github.com/libp2p/rust-libp2p/pull/3567 just to get your thoughts on the change.

swarm-derive/src/lib.rs
```diff
@@ -699,12 +699,19 @@ fn build_struct(ast: &DeriveInput, data_struct: &DataStruct) -> TokenStream {
             }
         };
 
+
```
thomaseizinger:
```suggestion
```

swarm/src/behaviour.rs
```diff
@@ -278,6 +278,12 @@ pub enum ToSwarm<TOutEvent, TInEvent> {
     /// This allows a [`NetworkBehaviour`] to identify a connection that resulted out of its own dial request.
     Dial { opts: DialOpts },
 
+    /// Instructs the `Swarm` to listen on the provided address
```
thomaseizinger:
```suggestion
    /// Instructs the [`Swarm`] to listen on the provided address
```

swarm/src/behaviour.rs
```diff
@@ -278,6 +278,12 @@ pub enum ToSwarm<TOutEvent, TInEvent> {
...
+    ListenOn { id: ListenerId, address: Multiaddr },
+
+    /// Instructs the `Swarm` to remove the listener
```
thomaseizinger:
```suggestion
    /// Instructs the [`Swarm`] to remove the listener
```

swarm/src/lib.rs
```diff
@@ -1121,6 +1121,12 @@ where
                     }
                 }
             }
+            ToSwarm::ListenOn { id, address } => {
+                self.transport.listen_on(id, address).ok()?;
```
thomaseizinger: If this fails, the `NetworkBehaviour` will never learn about it. I think we need to capture this error and report it back to the `NetworkBehaviour` as well as issue a `ListenerError`.

The documentation on `ListenerError` talks about "non-fatal" errors but I think adding another variant is even more confusing.

dariusc93: I was thinking about this back during the initial PR, but I assumed it would be alright since `ToSwarm::Dial` ignores the error. We could do this instead 

```suggestion
               if let Err(e) = self.transport.listen_on(id, address) {
                    self.behaviour.on_swarm_event(FromSwarm::ListenerError(
                        behaviour::ListenerError {
                            listener_id: id,
                            err: &e,
                        },
                    ));
                }
```

So it would notify the behaviour of the error instead of ignoring it.

thomaseizinger: > I was thinking about this back during the initial PR, but I assumed it would be alright since `ToSwarm::Dial` ignores the error.

Are we? Probably also something that should be fixed then :)

> We could do this instead 
> 
> ```suggestion
>                if let Err(e) = self.transport.listen_on(id, address) {
>                     self.behaviour.on_swarm_event(FromSwarm::ListenerError(
>                         behaviour::ListenerError {
>                             listener_id: id,
>                             err: &e,
>                         },
>                     ));
>                 }
> ```
> 
> So it would notify the behaviour of the error instead of ignoring it. 

Yep, that looks good to me!

dariusc93: > > I was thinking about this back during the initial PR, but I assumed it would be alright since `ToSwarm::Dial` ignores the error.
> 
> Are we? Probably also something that should be fixed then :)

Looking at `Swarm::dial`, it does send the error to the behaviour internally before returning the error itself so it doesnt look to be necessary

EDIT: Overlooked the beginning, but it looks like it doesnt send anything for `dial_opts.get_or_parse_peer_id().map_err(DialError::InvalidPeerId)`, but not sure if this would need to be sent to the behaviour. If it does we could do this internally there in the function by doing

```rust
let peer_id = match dial_opts
            .get_or_parse_peer_id()
            .map_err(DialError::InvalidPeerId)
        {
            Ok(peer_id) => peer_id,
            Err(e) => {
                self.behaviour
                    .on_swarm_event(FromSwarm::DialFailure(DialFailure {
                        peer_id: None,
                        error: &e,
                        connection_id,
                    }));
                return Err(e);
            }
        };
```

thomaseizinger: That error will go away once we have https://github.com/multiformats/rust-multiaddr/issues/73.

swarm/src/behaviour.rs
```diff
@@ -275,6 +275,12 @@ pub enum ToSwarm<TOutEvent, TInEvent> {
     /// This allows a [`NetworkBehaviour`] to identify a connection that resulted out of its own dial request.
     Dial { opts: DialOpts },
 
+    /// Instructs the [`Swarm`] to listen on the provided address
+    ListenOn { id: ListenerId, address: Multiaddr },
+
+    /// Instructs the [`Swarm`] to remove the listener
+    RemoveListener { id: ListenerId },
```
thomaseizinger:
```suggestion
    /// Instructs the [`Swarm`] to listen on the provided address.
    ListenOn { id: ListenerId, address: Multiaddr },

    /// Instructs the [`Swarm`] to remove the listener.
    RemoveListener { id: ListenerId },
```

thomaseizinger (approved):
This looks good to me, thank you!

@mxinden on board with that?

thomaseizinger (changes requested):
Needs a changelog entry.

thomaseizinger (approved):

thomaseizinger:
CI is not yet happy: https://github.com/libp2p/rust-libp2p/actions/runs/4979900766/jobs/8912098351?pr=3292

dariusc93:
> CI is not yet happy: https://github.com/libp2p/rust-libp2p/actions/runs/4979900766/jobs/8912098351?pr=3292

I noticed that since the links were placed there but wasnt resolving to `Swarm` since it wasnt being imported there. May have to change it to `crate::Swarm` or maybe do `[Swarm](crate::Swarm)`.

mxinden (commented):
One small comment. Otherwise this looks good to me.

swarm/src/lib.rs
```diff
@@ -1016,6 +1016,19 @@ where
...
                 }
             }
+            ToSwarm::ListenOn { id, address } => {
+                if let Err(e) = self.transport.listen_on(id, address) {
+                    self.behaviour.on_swarm_event(FromSwarm::ListenerError(
+                        behaviour::ListenerError {
+                            listener_id: id,
+                            err: &e,
+                        },
+                    ));
```
mxinden: When a call to `Swarm::listen_on` fails, we return the `TransportError` to the user, but don't call `Behaviour::on_swarm_event`.

https://github.com/libp2p/rust-libp2p/blob/67b26cce4f5563e5b8e45d28731d24f24e468914/swarm/src/lib.rs#L368-L381

When we handle a `ToSwarm::ListenOn` and the call fails, we provide the error to `Behaviour::on_swarm_event` but not to the user.

Would it make sense to always provide the error both to the `NetworkBehaviour` and to the user? That would be in-line with our general goal of mirroring the behavi**o**r from the `Swarm` to the user and the `Swarm` to the `NetworkBehaviour`.

dariusc93: Yes I think it would make more sense to provide it to both `NetworkBehaviour` and the user.

swarm/src/lib.rs
```diff
@@ -372,7 +372,16 @@ where
...
+            self.behaviour
+                .on_swarm_event(FromSwarm::ListenerError(behaviour::ListenerError {
+                    listener_id: id,
+                    err: &e,
+                }));
+            return Err(e);
+        }
+
         self.behaviour
             .on_swarm_event(FromSwarm::NewListener(behaviour::NewListener {
```
thomaseizinger: This should happen as well for the `ToSwarm` case I guess?

Kind of the most important bit :D

This makes me think that:

1. We should extract this into a `add_listener` function.
2. We should add a regression test for this functionality: a `NetworkBehaviour` that creates a listener, we connect to it and then remove it again. Prob best to use the `MemoryTransport` for it.

thomaseizinger (changes requested):
Found what I think is a bug.

swarm/src/lib.rs
dariusc93: Yes it should. I was wondering about last night prior to pushing up the code. Originally, I was calling `Swarm::listen_on`, which would have emitted the event (which worked in prior test), but since I was calling it directly from the transport, it would never emit the event (didnt get the chance to test). Glad that was caught now than later. 

As for your thoughts:
1. This is for calling `NetworkBehaviour::on_swarm_event`, correct? Also, If we are to add a `add_listener` function, would this be apart of `Swarm` and accessible by the end user or a private function? I assume the latter.
2. Should this test be located in the current crate?

thomaseizinger: >     1. This is for calling `NetworkBehaviour::on_swarm_event`, correct? Also, If we are to add a `add_listener` function, would this be apart of `Swarm` and accessible by the end user or a private function? I assume the latter.

Yes a private function. We already have `remove_listener` so just a counter-part to that.

>     2. Should this test be located in the current crate?

I'd suggest we make it a test in `swarm/tests`. Where possible, I think our tests should use the public API and we can enforce that by putting them into the `tests` directory.

thomaseizinger: >     1. This is for calling `NetworkBehaviour::on_swarm_event`, correct? Also, If we are to add a `add_listener` function, would this be apart of `Swarm` and accessible by the end user or a private function? I assume the latter.

Yes a private function. We already have `remove_listener` so just a counter-part to that.

>     2. Should this test be located in the current crate?

I'd suggest we make it a test in `swarm/tests`. Where possible, I think our tests should use the public API and we can enforce that by putting them into the `tests` directory.

swarm/src/lib.rs
```diff
@@ -372,11 +372,17 @@ where
...
-            }));
+
+        if let Err(e) = self.transport.listen_on(id, addr) {
+            self.behaviour
+                .on_swarm_event(FromSwarm::ListenerError(behaviour::ListenerError {
+                    listener_id: id,
+                    err: &e,
+                }));
+            return Err(e);
+        }
```
thomaseizinger: Can this be part of `add_listener`?

mxinden:
@thomaseizinger do the latest commits address your concerns?

swarm/src/lib.rs
```diff
@@ -544,6 +550,14 @@ where
         self.confirmed_external_addr.iter()
     }
 
+    /// TBD
```
```suggestion
```

swarm/tests/listener.rs
```diff
@@ -0,0 +1,146 @@
...
+        if let Some(event) = self.events.pop_front() {
+            return Poll::Ready(event);
+        }
+
+        Poll::Pending
+    }
+}
+
+#[async_std::test]
+async fn behaviour_listener() {
```
thomaseizinger: Nit: Can we have the test at the top please?

swarm/src/behaviour.rs
```diff
@@ -250,6 +250,12 @@ pub enum ToSwarm<TOutEvent, TInEvent> {
     /// This allows a [`NetworkBehaviour`] to identify a connection that resulted out of its own dial request.
     Dial { opts: DialOpts },
 
+    /// Instructs the [`Swarm`](crate::Swarm) to listen on the provided address.
+    ListenOn { id: ListenerId, address: Multiaddr },
```
thomaseizinger: This isn't quite consistent with `Dial` at the moment. Should we create a `ListenOpts` struct that by default creates the `ListenerId` for you and the user just has a ctor to pass in the `Multiaddr`?

dariusc93: Either way would be fine to me, though If we did use `ListenOpts`, one might expect `Into<ListenOpts>` to be used for `Swarm::listen_on` as well, which would be fine to do, but would be redundant for it to return a `ListenerId` (and thus being a breaking change if its dropped from being returned).

thomaseizinger: The entire PR is a breaking change anyway because we are adding variants to a public enum.

I am not sure if consistency with the `Swarm::dial` API is important enough here. I don't see us adding more options to `ListenOpts` any time soon, so it will always just be a `Multiaddr`.

For the `ToSwarm` case, having `ListenOpts` would be convenient because it allows users to construct the variant in one go, without having to also construct `ListenerId`.

swarm/src/lib.rs
```diff
@@ -68,6 +68,7 @@ pub mod dial_opts;
 pub mod dummy;
 pub mod handler;
 pub mod keep_alive;
+pub mod listen_opts;
```
thomaseizinger: Minor thing: Can we make this private?

swarm/src/lib.rs
```diff
@@ -121,6 +122,7 @@ pub use handler::{
 };
 #[cfg(feature = "macros")]
 pub use libp2p_swarm_derive::NetworkBehaviour;
+use listen_opts::ListenOpts;
```
thomaseizinger: And this a `pub use`?

swarm/src/lib.rs
```diff
@@ -544,6 +543,24 @@ where
...
+    fn add_listener(&mut self, opts: ListenOpts) {
+        let addr = opts.address();
+        let listener_id = opts.listener_id();
+
+        if let Err(e) = self.transport.listen_on(listener_id, addr) {
+            self.behaviour
+                .on_swarm_event(FromSwarm::ListenerError(behaviour::ListenerError {
+                    listener_id,
+                    err: &e,
+                }));
```
thomaseizinger, line 553: After this, we need to `return Err(e)`, otherwise we will always emit a `NewListener` event.

swarm/src/lib.rs
```diff
@@ -371,12 +373,9 @@ where
...
     pub fn listen_on(&mut self, addr: Multiaddr) -> Result<ListenerId, TransportError<io::Error>> {
-        let id = ListenerId::next();
-        self.transport.listen_on(id, addr)?;
-        self.behaviour
-            .on_swarm_event(FromSwarm::NewListener(behaviour::NewListener {
-                listener_id: id,
-            }));
+        let opts = ListenOpts::new(addr);
+        let id = opts.listener_id();
+        self.add_listener(opts);
```
```suggestion
        self.add_listener(opts)?;
```

Once we return the error.

swarm/src/lib.rs
```diff
@@ -1016,6 +1033,12 @@ where
                     });
                 }
             }
+            ToSwarm::ListenOn { opts } => {
+                self.add_listener(opts);
```
```suggestion
                // Error is dispatched internally, safe to ignore.
                let _ = self.add_listener(opts);
```

swarm/src/listen_opts.rs
```diff
@@ -0,0 +1,33 @@
...
+
+    /// Get the [`ListenerId`] of this listen attempt
+    pub fn listener_id(&self) -> ListenerId {
+        self.id
+    }
+
+    /// Get the [`Multiaddr`] that is being listened on
+    pub fn address(&self) -> Multiaddr {
+        self.address.clone()
+    }
```
thomaseizinger:
```suggestion
    pub fn address(&self) -> &Multiaddr {
        &self.address
    }
```

The user can always clone themselves but never undo the clone.

thomaseizinger (commented):
A few minor things but overall, happy to merge this design. Thank you for iterating on this with me @dariusc93 !

swarm/src/lib.rs
dariusc93: This was overlooked on my side.

thomaseizinger (approved):
Thanks!

mxinden (approved):
@dariusc93 thanks for being patient with all the feedback. Your work on this is valued.

PR #3293, opened 2023-01-01
refactor(core): Improve `MapErr` for `transport`; improve docs and arrange code

yjhmelody:
## Description

- Format code and improve docs.
- Remove some unused fields.
- Rename some modules to keep the same style.
- Define `MapErrFuture` to keep the same code style.

## Change checklist


jxs (commented):
Hi, and thanks for your interest! Overall looks good, can we maintain the function order for less diffs?

core/src/transport.rs
```diff
@@ -35,11 +35,11 @@ use std::{
 };
 
 pub mod and_then;
-pub mod choice;
 pub mod dummy;
 pub mod map;
 pub mod map_err;
 pub mod memory;
+pub mod or;
```
thomaseizinger, line 42: This is a breaking change. I'd like to have this non-breaking please. Should be easy with another module and type aliases.

core/src/transport/and_then.rs
```diff
@@ -78,7 +78,6 @@ where
                     role_override: Endpoint::Dialer,
                 },
             )),
-            _marker: PhantomPinned,
```
thomaseizinger, line 81: Is this safe? I am sure this wasn't added without reason.

@mxinden Do you know why we phantom pin this type?

core/src/transport/map_err.rs
```diff
@@ -76,73 +77,49 @@ where
...
+    fn address_translation(&self, server: &Multiaddr, observed: &Multiaddr) -> Option<Multiaddr> {
+        self.transport.address_translation(server, observed)
     }
 }
 
-/// Dialing future for `MapErr`.
+/// Future for `MapErr`.
 #[pin_project::pin_project]
-pub struct MapErrDial<T: Transport, F> {
+pub struct MapErrFuture<T, Fut, F> {
```
thomaseizinger, line 112: This is also breaking. Please add a type alias to make it non-breaking with a deprecation attribute.

thomaseizinger (changes requested):
Thanks!

There are several unrelated things happening here. We squash merge PRs so I'd appreciate it if we can split those i.e. do a separate PR for reordering the functions etc

thomaseizinger:
Closing due to inactivity.

PR #3280, opened 2022-12-23
fix(kad): Skip invalid multiaddr

mxinden:
## Description

With this commit `libp2p-kad` no longer discards the whole peer payload in case an addr is invalid, but instead logs the failure, skips the invalid multiaddr and parses the remaining payload.

See https://github.com/libp2p/rust-libp2p/issues/3244 for details.

## Notes

Targets the `v0.50` branch.

## Change checklist

- [x] I have performed a self-review of my own code
- [ ] I have made corresponding changes to the documentation
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] A changelog entry has been made in the appropriate crates

PR #3300, opened 2023-01-02
fix(dcutr): Skip unparsable multiaddr (#3280)

mxinden:
## Description

With this commit `libp2p-dcutr` no longer discards the whole remote payload in case an addr is unparsable, but instead logs the failure and skips the unparsable multiaddr.

See https://github.com/libp2p/rust-libp2p/issues/3244 for details.

## Change checklist

### DIFF
--- a/protocols/dcutr/src/protocol/inbound.rs
+++ b/protocols/dcutr/src/protocol/inbound.rs
@@ -58,14 +58,23 @@ impl upgrade::InboundUpgrade<NegotiatedSubstream> for Upgrade {
             } else {
                 obs_addrs
                     .into_iter()
-                    .map(Multiaddr::try_from)
-                    // Filter out relayed addresses.
-                    .filter(|a| match a {
-                        Ok(a) => !a.iter().any(|p| p == Protocol::P2pCircuit),
-                        Err(_) => true,
+                    .filter_map(|a| match Multiaddr::try_from(a) {
+                        Ok(a) => Some(a),
+                        Err(e) => {
+                            log::debug!("Unable to parse multiaddr: {e}");
+                            None
+                        }
                     })
-                    .collect::<Result<Vec<Multiaddr>, _>>()
-                    .map_err(|_| UpgradeError::InvalidAddrs)?
+                    // Filter out relayed addresses.
+                    .filter(|a| {
+                        if a.iter().any(|p| p == Protocol::P2pCircuit) {
+                            log::debug!("Dropping relayed address {a}");
+                            false
+                        } else {
+                            true
+                        }
+                    })
+                    .collect::<Vec<Multiaddr>>()
             };
 
             let r#type = hole_punch::Type::from_i32(r#type).ok_or(UpgradeError::ParseTypeField)?;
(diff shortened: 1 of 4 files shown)


elenaf9 (commented):
Good catch :+1: 
Unless I am missing something, the same also needs to be fixed in `dcutr/protocol/outbound.rs`?

mxinden:
> Unless I am missing something, the same also needs to be fixed in `dcutr/protocol/outbound.rs`?

:man_facepalming: Thanks! Mind taking another look @elenaf9?

protocols/dcutr/src/protocol/inbound.rs
```diff
@@ -58,14 +58,23 @@ impl upgrade::InboundUpgrade<NegotiatedSubstream> for Upgrade {
...
-                        Ok(a) => !a.iter().any(|p| p == Protocol::P2pCircuit),
-                        Err(_) => true,
+                    .filter(|a| {
+                        if a.iter().any(|p| p == Protocol::P2pCircuit) {
+                            log::debug!("Dropping relayed address {a}");
+                            false
+                        } else {
+                            true
+                        }
                     })
```
elenaf9, line 76: We could merged these two filter blocks:

```rust 
                    .filter_map(|a| match Multiaddr::try_from(a) {
                        Ok(a) if !a.iter().any(|p| p == Protocol::P2pCircuit) => Some(a),
                        Ok(a) => {
                            log::debug!("Dropping relayed address {a}");
                            None
                        }
                        Err(e) => {
                            log::debug!("Unable to parse multiaddr: {e}");
                            None
                        }
                    })

```
(here and below)
No strong opinion.

elenaf9 (approved):
Looks good to me

protocols/dcutr/src/protocol/inbound.rs
mxinden: What would be the benefit of merging the two?

Structure wise, I find the separation to ease understanding.

elenaf9: The benefit would be to have a single place where addresses are filtered out. I actually first oversaw that there was a second filter call. 
But that's subjective and nit-picky; feel free to ignore :slightly_smiling_face:.

thomaseizinger: Alternative idea: Keep using `.map(Multiaddr::try_from)` and use `.filter` twice for consistency.

thomaseizinger: > Alternative idea: Keep using `.map(Multiaddr::try_from)` and use `.filter` twice for consistency.

Don't mind me, that doesn't work because we need to unpack the result somewhere.

thomaseizinger (approved):
One way to reduce the code duplication here would be to create our own type that implements `FromIterator` and filters unparseable and relayed addresses.

mxinden:
> One way to reduce the code duplication here would be to create our own type that implements `FromIterator` and filters unparseable and relayed addresses.

Neat trick. That said, is it worth the abstraction given that it is only used twice?

---

I will move forward here, releasing this patch right away.

mxinden:
[Tagged](https://github.com/libp2p/rust-libp2p/releases/tag/libp2p-dcutr-v0.8.1) and [published](https://crates.io/crates/libp2p-dcutr/0.8.1).

thomaseizinger:
> > One way to reduce the code duplication here would be to create our own type that implements `FromIterator` and filters unparseable and relayed addresses.
> 
> Neat trick. That said, is it worth the abstraction given that it is only used twice?

Probably not :)

PR #3303, opened 2023-01-02
refactor(gossipsub): revise symbol naming to follow conventions

StemCll:
## Description
Changes regarding the  #2217

## Change checklist

### DIFF
--- a/examples/gossipsub-chat.rs
+++ b/examples/gossipsub-chat.rs
@@ -49,7 +49,7 @@ use async_std::io;
 use futures::{prelude::*, select};
 use libp2p::gossipsub::MessageId;
 use libp2p::gossipsub::{
-    Gossipsub, GossipsubEvent, IdentTopic as Topic, Message, MessageAuthenticity, ValidationMode,
+    Event, Gossipsub, IdentTopic as Topic, Message, MessageAuthenticity, ValidationMode,
 };
 use libp2p::{
     gossipsub, identity, mdns, swarm::NetworkBehaviour, swarm::SwarmEvent, PeerId, Swarm,
@@ -139,7 +139,7 @@ async fn main() -> Result<(), Box<dyn Error>> {
                         swarm.behaviour_mut().gossipsub.remove_explicit_peer(&peer_id);
                     }
                 },
-                SwarmEvent::Behaviour(MyBehaviourEvent::Gossipsub(GossipsubEvent::ProtobufMessage {
+                SwarmEvent::Behaviour(MyBehaviourEvent::Gossipsub(Event::ProtobufMessage {
                     propagation_source: peer_id,
                     message_id: id,
                     message,
--- a/examples/gossipsub-chat.rs
+++ b/examples/gossipsub-chat.rs
@@ -47,11 +47,6 @@
 
 use async_std::io;
 use futures::{prelude::*, select};
-use libp2p::gossipsub::MessageId;
-use libp2p::gossipsub::{
-    Behaviour as Gossipsub, Event, IdentTopic as Topic, Message, MessageAuthenticity,
-    ValidationMode,
-};
 use libp2p::{
     gossipsub, identity, mdns, swarm::NetworkBehaviour, swarm::SwarmEvent, PeerId, Swarm,
 };
@@ -73,31 +68,34 @@ async fn main() -> Result<(), Box<dyn Error>> {
     // We create a custom network behaviour that combines Gossipsub and Mdns.
     #[derive(NetworkBehaviour)]
     struct MyBehaviour {
-        gossipsub: Gossipsub,
+        gossipsub: gossipsub::Behaviour,
         mdns: mdns::async_io::Behaviour,
     }
 
     // To content-address message, we can take the hash of message and use it as an ID.
-    let message_id_fn = |message: &Message| {
+    let message_id_fn = |message: &gossipsub::Message| {
         let mut s = DefaultHasher::new();
         message.data.hash(&mut s);
-        MessageId::from(s.finish().to_string())
+        gossipsub::MessageId::from(s.finish().to_string())
     };
 
     // Set a custom gossipsub configuration
     let gossipsub_config = gossipsub::ConfigBuilder::default()
         .heartbeat_interval(Duration::from_secs(10)) // This is set to aid debugging by not cluttering the log space
-        .validation_mode(ValidationMode::Strict) // This sets the kind of message validation. The default is Strict (enforce message signing)
+        .validation_mode(gossipsub::ValidationMode::Strict) // This sets the kind of message validation. The default is Strict (enforce message signing)
         .message_id_fn(message_id_fn) // content-address messages. No two messages of the same content will be propagated.
         .build()
         .expect("Valid config");
 
     // build a gossipsub network behaviour
-    let mut gossipsub = Gossipsub::new(MessageAuthenticity::Signed(local_key), gossipsub_config)
-        .expect("Correct configuration");
+    let mut gossipsub = gossipsub::Behaviour::new(
+        gossipsub::MessageAuthenticity::Signed(local_key),
+        gossipsub_config,
+    )
+    .expect("Correct configuration");
 
     // Create a Gossipsub topic
-    let topic = Topic::new("test-net");
+    let topic = gossipsub::IdentTopic::new("test-net");
 
     // subscribes to our topic
     gossipsub.subscribe(&topic)?;
@@ -140,7 +138,7 @@ async fn main() -> Result<(), Box<dyn Error>> {
                         swarm.behaviour_mut().gossipsub.remove_explicit_peer(&peer_id);
                     }
                 },
-                SwarmEvent::Behaviour(MyBehaviourEvent::Gossipsub(Event::ProtobufMessage {
+                SwarmEvent::Behaviour(MyBehaviourEvent::Gossipsub(gossipsub::Event::ProtobufMessage {
                     propagation_source: peer_id,
                     message_id: id,
                     message,
--- a/examples/gossipsub-chat.rs
+++ b/examples/gossipsub-chat.rs
@@ -80,9 +80,9 @@ async fn main() -> Result<(), Box<dyn Error>> {
     };
 
     // Set a custom gossipsub configuration
-    let gossipsub_config = gossipsub::ConfigBuilder::default()
+    let gossipsub_config = gossipsub::config::Builder::default()
         .heartbeat_interval(Duration::from_secs(10)) // This is set to aid debugging by not cluttering the log space
-        .validation_mode(gossipsub::ValidationMode::Strict) // This sets the kind of message validation. The default is Strict (enforce message signing)
+        .validation_mode(gossipsub::config::ValidationMode::Strict) // This sets the kind of message validation. The default is Strict (enforce message signing)
         .message_id_fn(message_id_fn) // content-address messages. No two messages of the same content will be propagated.
         .build()
         .expect("Valid config");
--- a/examples/gossipsub-chat.rs
+++ b/examples/gossipsub-chat.rs
@@ -49,7 +49,8 @@ use async_std::io;
 use futures::{prelude::*, select};
 use libp2p::gossipsub::MessageId;
 use libp2p::gossipsub::{
-    Event, Gossipsub, IdentTopic as Topic, Message, MessageAuthenticity, ValidationMode,
+    Behaviour as Gossipsub, Event, IdentTopic as Topic, Message, MessageAuthenticity,
+    ValidationMode,
 };
 use libp2p::{
     gossipsub, identity, mdns, swarm::NetworkBehaviour, swarm::SwarmEvent, PeerId, Swarm,
@@ -84,7 +85,7 @@ async fn main() -> Result<(), Box<dyn Error>> {
     };
 
     // Set a custom gossipsub configuration
-    let gossipsub_config = gossipsub::GossipsubConfigBuilder::default()
+    let gossipsub_config = gossipsub::ConfigBuilder::default()
         .heartbeat_interval(Duration::from_secs(10)) // This is set to aid debugging by not cluttering the log space
         .validation_mode(ValidationMode::Strict) // This sets the kind of message validation. The default is Strict (enforce message signing)
         .message_id_fn(message_id_fn) // content-address messages. No two messages of the same content will be propagated.
--- a/examples/gossipsub-chat.rs
+++ b/examples/gossipsub-chat.rs
@@ -80,7 +80,7 @@ async fn main() -> Result<(), Box<dyn Error>> {
     };
 
     // Set a custom gossipsub configuration
-    let gossipsub_config = gossipsub::Builder::default()
+    let gossipsub_config = gossipsub::ConfigBuilder::default()
         .heartbeat_interval(Duration::from_secs(10)) // This is set to aid debugging by not cluttering the log space
         .validation_mode(gossipsub::ValidationMode::Strict) // This sets the kind of message validation. The default is Strict (enforce message signing)
         .message_id_fn(message_id_fn) // content-address messages. No two messages of the same content will be propagated.
--- a/examples/gossipsub-chat.rs
+++ b/examples/gossipsub-chat.rs
@@ -49,8 +49,7 @@ use async_std::io;
 use futures::{prelude::*, select};
 use libp2p::gossipsub::MessageId;
 use libp2p::gossipsub::{
-    Gossipsub, GossipsubEvent, GossipsubMessage, IdentTopic as Topic, MessageAuthenticity,
-    ValidationMode,
+    Gossipsub, GossipsubEvent, IdentTopic as Topic, Message, MessageAuthenticity, ValidationMode,
 };
 use libp2p::{
     gossipsub, identity, mdns, swarm::NetworkBehaviour, swarm::SwarmEvent, PeerId, Swarm,
@@ -78,7 +77,7 @@ async fn main() -> Result<(), Box<dyn Error>> {
     }
 
     // To content-address message, we can take the hash of message and use it as an ID.
-    let message_id_fn = |message: &GossipsubMessage| {
+    let message_id_fn = |message: &Message| {
         let mut s = DefaultHasher::new();
         message.data.hash(&mut s);
         MessageId::from(s.finish().to_string())
@@ -140,7 +139,7 @@ async fn main() -> Result<(), Box<dyn Error>> {
                         swarm.behaviour_mut().gossipsub.remove_explicit_peer(&peer_id);
                     }
                 },
-                SwarmEvent::Behaviour(MyBehaviourEvent::Gossipsub(GossipsubEvent::Message {
+                SwarmEvent::Behaviour(MyBehaviourEvent::Gossipsub(GossipsubEvent::ProtobufMessage {
                     propagation_source: peer_id,
                     message_id: id,
                     message,
--- a/examples/gossipsub-chat.rs
+++ b/examples/gossipsub-chat.rs
@@ -138,7 +138,7 @@ async fn main() -> Result<(), Box<dyn Error>> {
                         swarm.behaviour_mut().gossipsub.remove_explicit_peer(&peer_id);
                     }
                 },
-                SwarmEvent::Behaviour(MyBehaviourEvent::Gossipsub(gossipsub::Event::ProtobufMessage {
+                SwarmEvent::Behaviour(MyBehaviourEvent::Gossipsub(gossipsub::Event::Message {
                     propagation_source: peer_id,
                     message_id: id,
                     message,
--- a/examples/gossipsub-chat.rs
+++ b/examples/gossipsub-chat.rs
@@ -80,9 +80,9 @@ async fn main() -> Result<(), Box<dyn Error>> {
     };
 
     // Set a custom gossipsub configuration
-    let gossipsub_config = gossipsub::config::Builder::default()
+    let gossipsub_config = gossipsub::Builder::default()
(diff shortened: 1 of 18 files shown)


StemCll:
I have to test the changes from backwards compatibility correctness, however I think it's ready for review cc @thomaseizinger :pray:

examples/gossipsub-chat.rs
```diff
@@ -140,7 +138,7 @@ async fn main() -> Result<(), Box<dyn Error>> {
                         swarm.behaviour_mut().gossipsub.remove_explicit_peer(&peer_id);
                     }
                 },
-                SwarmEvent::Behaviour(MyBehaviourEvent::Gossipsub(GossipsubEvent::Message {
+                SwarmEvent::Behaviour(MyBehaviourEvent::Gossipsub(gossipsub::Event::ProtobufMessage {
```
StemCll: Should it be `ProtobufMesssage` or maybe revert it to `Message`? I think that externally developer shouldn't care if the message is protobuf or other type, but not sure

protocols/gossipsub/src/lib.rs
```diff
@@ -142,7 +142,7 @@ pub mod protocol;
 
 mod backoff;
 mod behaviour;
-mod config;
+pub mod config;
```
StemCll: not sure if that should be public...

examples/gossipsub-chat.rs
thomaseizinger: We can't easily deprecate enum variant names so I'd say revert it to `Message` so we remain non-breaking!

protocols/gossipsub/src/lib.rs
thomaseizinger: I'd rather not, what prompted the change?

protocols/gossipsub/src/lib.rs
```diff
@@ -156,18 +156,51 @@ mod types;
...
+    since = "0.44.0",
+    note = "Use re-exports that omit `Gossipsub` infix, i.e. `libp2p::gossipsub::RawMessage"
+)]
+pub type RawGossipsubMessage = RawMessage;
+
+#[deprecated(
+    since = "0.44.0",
+    note = "Use re-exports that omit `Gossipsub` prefix, i.e. `libp2p::gossipsub::config::Builder"
+)]
+pub type GossipsubBuilder = config::Builder;
```
thomaseizinger: Ah yes. That is a tricky one.

I somewhat agree that `config::Builder` is more accurate. However, we didn't call it `GossipsubConfigBuilder` before so if we stick to `gossipsub::Builder` now we technically aren't making the situation any worse :)

For the sake of not bike-shedding this and moving on, I'd suggest to just stick to that and keep `config` private.

If somebody cares enough, they can think of a solution to improve this assuming balanced trade-offs (private/public modules, length of type names/paths, expressiveness, etc).

protocols/gossipsub/src/lib.rs
```diff
@@ -156,18 +156,51 @@ mod types;
...
+
+#[deprecated(
+    since = "0.44.0",
+    note = "Use re-exports that omit `Gossipsub` prefix, i.e. `libp2p::gossipsub::config::Version"
+)]
+pub type GossipsubVersion = config::Version;
```
thomaseizinger: Here I'd say that `gossipsub::Version` is actually preferable.

thomaseizinger (commented):
Thanks for doing this :)

I'd spot checked it and looks alright. I'll give it a more thorough look once we have the comments resolved!

examples/gossipsub-chat.rs
StemCll: done :heavy_check_mark:

protocols/gossipsub/src/lib.rs
StemCll: If I remember correctly, the `::config::` prefix. I've reverted to private.

protocols/gossipsub/src/lib.rs
StemCll: :heavy_check_mark:

protocols/gossipsub/src/lib.rs
StemCll: :heavy_check_mark:

protocols/gossipsub/src/types.rs
```diff
@@ -23,7 +23,7 @@ use crate::rpc_proto;
 use crate::TopicHash;
 use libp2p_core::{connection::ConnectionId, PeerId};
 use prometheus_client::encoding::EncodeLabelValue;
-use prost::Message;
+use prost::Message as ProtobufMessage;
```
thomaseizinger: If you just need to the trait to be in scope to call functions on it you can do this:

```suggestion
use prost::Message as _;
```

protocols/gossipsub/src/behaviour.rs
```diff
@@ -670,7 +667,7 @@ where
                     }
                 }
 
-                // Gossipsub peers
+                // Behaviour peers
```
thomaseizinger: This rename doesn't make much sense IMO.

protocols/gossipsub/src/behaviour.rs
```diff
@@ -919,7 +916,7 @@ where
         }
     }
 
-    /// Gossipsub JOIN(topic) - adds topic peers to mesh and sends them GRAFT messages.
+    /// Behaviour JOIN(topic) - adds topic peers to mesh and sends them GRAFT messages.
```
thomaseizinger: Same here. Can you revert the ones in the comments please?

protocols/gossipsub/src/error.rs
```diff
@@ -88,7 +88,7 @@ impl From<SigningError> for PublishError {
 
 /// Errors that can occur in the protocols handler.
 #[derive(Debug, Error)]
-pub enum GossipsubHandlerError {
+pub enum HandlerError {
```
thomaseizinger, line 97: This module is public (unfortunately), meaning this rename needs a deprecation type alias.

thomaseizinger: While we are at it, I think it would make sense to re-export all of these from the root and add deprecation type aliases that they shouldn't be imported from the `error` module so we can make it private in a future release.

protocols/gossipsub/src/lib.rs
```diff
@@ -156,18 +156,53 @@ mod types;
 
 mod rpc_proto;
 
-pub use self::behaviour::{Gossipsub, GossipsubEvent, MessageAuthenticity};
```
thomaseizinger, line 159: Deprecation type aliases for `Gossipsub` and `GossipsubEvent` are missing.

protocols/gossipsub/src/lib.rs
```diff
@@ -156,18 +156,53 @@ mod types;
...
+    since = "0.44.0",
+    note = "Use re-exports that omit `Gossipsub` infix, i.e. `libp2p::gossipsub::RawMessage"
+)]
+pub type RawGossipsubMessage = RawMessage;
+
+#[deprecated(
+    since = "0.44.0",
+    note = "Use re-exports that omit `Gossipsub` prefix, i.e. `libp2p::gossipsub::Builder"
+)]
+pub type GossipsubBuilder = config::Builder;
```
thomaseizinger:
```suggestion
#[deprecated(
    since = "0.44.0",
    note = "Use re-exports that omit `Gossipsub` prefix, i.e. `libp2p::gossipsub::Builder"
)]
pub type GossipsubConfigBuilder = config::Builder;
```

That is what the type used to be called.

protocols/gossipsub/src/lib.rs
```diff
@@ -156,18 +156,53 @@ mod types;
...
+
+#[deprecated(
+    since = "0.44.0",
+    note = "Use re-exports that omit `Gossipsub` infix, i.e. `libp2p::gossipsub::RawMessage"
+)]
+pub type RawGossipsubMessage = RawMessage;
+
+#[deprecated(
+    since = "0.44.0",
+    note = "Use re-exports that omit `Gossipsub` prefix, i.e. `libp2p::gossipsub::Builder"
```
thomaseizinger: Can we name `Builder` `ConfigBuilder` instead? I think that would make more sense from the public API PoV and the rename is consistent with just dropping the `Gossipsub` prefix.

protocols/gossipsub/src/lib.rs
```diff
@@ -156,18 +156,53 @@ mod types;
...
-};
+pub use self::types::{FastMessageId, Message, MessageAcceptance, MessageId, RawMessage, Rpc};
+
+pub use self::config::{Builder, Config, ValidationMode, Version};
+
+#[deprecated(
+    since = "0.44.0",
+    note = "Use re-exports that omit `Gossipsub` prefix, i.e. `libp2p::gossipsub::Config"
+)]
+pub type GossipsubConfig = config::Config;
```
thomaseizinger:
```suggestion
pub type GossipsubConfig = Config;
```

Unnecessary qualification.

protocols/gossipsub/src/lib.rs
thomaseizinger:
```suggestion
pub type GossipsubBuilder = Builder;
```

Same here.

protocols/gossipsub/src/lib.rs
```diff
@@ -156,18 +156,53 @@ mod types;
...
+    since = "0.44.0",
+    note = "Use re-exports that omit `Gossipsub` prefix, i.e. `libp2p::gossipsub::Builder"
+)]
+pub type GossipsubBuilder = config::Builder;
+
+#[deprecated(
+    since = "0.44.0",
+    note = "Use re-exports that omit `Gossipsub` prefix, i.e. `libp2p::gossipsub::Version"
+)]
+pub type GossipsubVersion = config::Version;
```
thomaseizinger:
```suggestion
pub type GossipsubVersion = Version;
```

And here.

protocols/gossipsub/src/behaviour.rs
```diff
@@ -32,7 +32,7 @@ use std::{
 use futures::StreamExt;
 use log::{debug, error, trace, warn};
 use prometheus_client::registry::Registry;
-use prost::Message;
+use prost::Message as ProtobufMessage;
```
thomaseizinger: This might also do with an underscore instead of an actual name.

protocols/gossipsub/src/config.rs
```diff
@@ -101,22 +100,22 @@ pub struct GossipsubConfig {
...
 
     /// The protocol id to negotiate this protocol. By default, the resulting protocol id has the form
-    /// `/<prefix>/<supported-versions>`, but can optionally be changed to a literal form by providing some GossipsubVersion as custom_id_version.
+    /// `/<prefix>/<supported-versions>`, but can optionally be changed to a literal form by providing some Version as custom_id_version.
     /// As gossipsub supports version 1.0 and 1.1, there are two suffixes supported for the resulting protocol id.
     ///
-    /// Calling `GossipsubConfigBuilder::protocol_id_prefix` will set a new prefix and retain the prefix logic.
-    /// Calling `GossipsubConfigBuilder::protocol_id` will set a custom `protocol_id` and disable the prefix logic.
+    /// Calling `Builder::protocol_id_prefix` will set a new prefix and retain the prefix logic.
+    /// Calling `Builder::protocol_id` will set a custom `protocol_id` and disable the prefix logic.
```
thomaseizinger:
```suggestion
    /// Calling [`Builder::protocol_id_prefix`] will set a new prefix and retain the prefix logic.
    /// Calling [`Builder::protocol_id`] will set a custom `protocol_id` and disable the prefix logic.
```

Should be a proper type reference.

thomaseizinger (changes requested):
Thanks! This looks pretty good already.

I've made some inline comments, please check for yourself that we are only renaming actual types and not comments.

Another thing we've noticed in previous PRs for renaming types is that the rustdoc comments might suddenly read a bit odd if it f.e. just says `Behaviour` now instead of `Gossipsub`. I've not through all docs in this review yet, can you please check them?

Some of the deprecation type aliases are still missing. I can recommend to temporarily undo all changes in `examples/` and `tests/` and see if everything still compiles. If not, you are missing a deprecation type alias :)

protocols/gossipsub/src/lib.rs
thomaseizinger: I was actually mistaken with what I wrote in https://github.com/libp2p/rust-libp2p/pull/3303#discussion_r1064236220, sorry to send you on a wild goose chase here.

StemCll:
Got a question regarding force pushing - why isn't it suitable in this repo. Couldn't find anything in `Contributing` guide :thinking: 

Personally, I force push when conflicts occur - I'm rebasing trunk and then force pushing to remote. What do you prefer in these situations?

thomaseizinger:
> Got a question regarding force pushing - why isn't it suitable in this repo. Couldn't find anything in `Contributing` guide :thinking: 

We squash-merge all PRs so the commit history within a PR is not that important and thus regular merges are fine. They also play nicely with GitHub's review interface, allowing you to review only the new commits.

protocols/gossipsub/src/types.rs
StemCll: Fixed all occurrences in `gossipsub`

protocols/gossipsub/src/behaviour.rs
StemCll: :heavy_check_mark:

protocols/gossipsub/src/behaviour.rs
StemCll: :heavy_check_mark:

protocols/gossipsub/src/lib.rs
StemCll: :heavy_check_mark:

protocols/gossipsub/src/lib.rs
StemCll: No worries at all :smile:

protocols/gossipsub/src/lib.rs
StemCll: Sure :heavy_check_mark:

protocols/gossipsub/src/config.rs
StemCll: :heavy_check_mark:

StemCll:
I've pushed some changes, but will double-check all your comments. Will mention you when ready for re-review :smile:

protocols/gossipsub/src/error.rs
StemCll: I had to add type alias in [error.rs](https://github.com/libp2p/rust-libp2p/pull/3303/files#diff-ca068b1f00890e595a99fcf98ce12d293f47fdc3cd1dd9c2fb89004f12035755R93) too. Not sure if that's a neat solution here

StemCll:
@thomaseizinger I've checked all your comments - right now the PR should be ready for review. Sorry that it took very long...

protocols/gossipsub/src/lib.rs
```diff
@@ -156,18 +156,73 @@ mod types;
...
+
+pub use self::config::{Config, ConfigBuilder, ValidationMode, Version};
+
+pub use self::error::{HandlerError, PublishError, SubscriptionError, ValidationError};
+
+#[deprecated(
+    since = "0.44.0",
+    note = "Use re-exports that omit `Gossipsub` prefix, i.e. `libp2p::gossipsub::HandlerError"
+)]
+pub type GossipsubHandlerError = HandlerError;
```
thomaseizinger: We previously didn't export this error on the root level right? I think this type alias is unnecessary.

protocols/gossipsub/src/lib.rs
```diff
@@ -156,18 +156,73 @@ mod types;
...
 };
 pub use self::topic::{Hasher, Topic, TopicHash};
-pub use self::types::{
-    FastMessageId, GossipsubMessage, GossipsubRpc, MessageAcceptance, MessageId,
-    RawGossipsubMessage,
-};
+pub use self::types::{FastMessageId, Message, MessageAcceptance, MessageId, RawMessage, Rpc};
+
+pub use self::config::{Config, ConfigBuilder, ValidationMode, Version};
+
```
thomaseizinger:
```suggestion
pub use self::config::{Config, ConfigBuilder, ValidationMode, Version};
```

Let's not use more whitespace than necessary :)

protocols/gossipsub/src/lib.rs
```diff
@@ -156,18 +156,73 @@ mod types;
 
 mod rpc_proto;
 
-pub use self::behaviour::{Gossipsub, GossipsubEvent, MessageAuthenticity};
+pub use self::behaviour::{Behaviour, Event, MessageAuthenticity};
 pub use self::transform::{DataTransform, IdentityTransform};
```
thomaseizinger:
```suggestion
```

protocols/gossipsub/src/lib.rs
thomaseizinger: It is slightly related but it would be really great if we can get rid of the public `error` module while we are at it. I'd prefer to not have any public modules actually. They are an easy semver hazard because types that shouldnt' be part of the public API easily slip into it.

To achieve this in a backwards-compatible way, we need to:

1. Make the `error` module private and rename it to `error_priv`
2. Make a new, inline `error` module that contains type aliases with deprecation notice, re-exporting the errors from `error_priv`
3. Export all errors in the root module (you are already doing this on line 171)

This allows us to migrate users towards importing the errors from the root module and we can remove the public `error` module in a future release!

See what we did in this PR for example: https://github.com/libp2p/rust-libp2p/pull/3238/files#diff-acd49d828b2b4f41554331034dd696f7946d8c1becc7635d118435454c262cde

Everything used to be under a `v2` module that now only contains deprecated type-aliases that re-export symbols from their new location. Some of the new locations are _crate-private_ however. In other words, we will be making certain types inaccessible in a future release. That is because they should have never been part of the public API.

I see some opportunities for this in gossipsub too:

- `protocol` module
- `metrics` (contains many public types and functions that a user doesn't need to interact with)
- `subscription_filter` (can be completely flattened, i.e. re-export from root to shorten import paths)
- `time_cache` (not sure why that is public at all?)

This is quite an undertaking by itself so it is probably best done in a separate PR. Let me know if you are up for it :)

thomaseizinger (commented):
Well done, only a few more comments! :)

thomaseizinger:
> @thomaseizinger I've checked all your comments - right now the PR should be ready for review. Sorry that it took very long...

No worries at all, thank you for pushing this forward!

protocols/gossipsub/src/lib.rs
StemCll: Perfectly understandable. I like this direction. I'm not sure about making `V2` APIs in these kind of changes, but let's talk about it in the next PR ;)

protocols/gossipsub/src/lib.rs
StemCll: :heavy_check_mark:

protocols/gossipsub/src/lib.rs
StemCll: :heavy_check_mark:

protocols/gossipsub/src/lib.rs
StemCll: I've created a new issue to track these changes -> https://github.com/libp2p/rust-libp2p/issues/3392

protocols/gossipsub/CHANGELOG.md
```diff
@@ -8,8 +8,16 @@
 
 - Initialize `ProtocolConfig` via `GossipsubConfig`. See [PR 3381].
 
+- Rename types as per [discussion 2174].
+	`Gossipsub` has been renamed to `Behaviour`.
+	The `Gossipsub` prefix has been removed from various types like `GossipsubConfig` or `GossipsubMessage`.
+	It is preferred to import the gossipsub protocol as a module (`use libp2p::gossipsub;`),
+	and refer to its types via `gossipsub::`. For example: `gossipsub::Behaviour` or `gossipsub::RawMessage`.
```
thomaseizinger:
```suggestion
- Rename types as per [discussion 2174].
	`Gossipsub` has been renamed to `Behaviour`.
	The `Gossipsub` prefix has been removed from various types like `GossipsubConfig` or `GossipsubMessage`.
	It is preferred to import the gossipsub protocol as a module (`use libp2p::gossipsub;`),
	and refer to its types via `gossipsub::`. For example: `gossipsub::Behaviour` or `gossipsub::RawMessage`. See [PR 3303].
```

With mentioning it somewhere, the link reference doesn't do anything :)

thomaseizinger (dismissed):
Thanks! This looks good. I double checked our exports again, they all match the old types :)

protocols/gossipsub/CHANGELOG.md
```diff
@@ -8,8 +8,16 @@
 
 - Initialize `ProtocolConfig` via `GossipsubConfig`. See [PR 3381].
 
+- Rename types as per [discussion 2174].
+	`Gossipsub` has been renamed to `Behaviour`.
+	The `Gossipsub` prefix has been removed from various types like `GossipsubConfig` or `GossipsubMessage`.
+	It is preferred to import the gossipsub protocol as a module (`use libp2p::gossipsub;`),
+	and refer to its types via `gossipsub::`. For example: `gossipsub::Behaviour` or `gossipsub::RawMessage`. See [PR 3303].
```
thomaseizinger:
```suggestion
- Rename types as per [discussion 2174].
  `Gossipsub` has been renamed to `Behaviour`.
  The `Gossipsub` prefix has been removed from various types like `GossipsubConfig` or `GossipsubMessage`.
  It is preferred to import the gossipsub protocol as a module (`use libp2p::gossipsub;`), and refer to its types via `gossipsub::`.
  For example: `gossipsub::Behaviour` or `gossipsub::RawMessage`. See [PR 3303].
```

thomaseizinger (dismissed):

thomaseizinger (approved):

Issue #3299, opened 2023-01-02
dcutr: Remove `ActionBuilder`


mxinden:
In `libp2p-dcutr` we introduced `ActionBuilder` as events from the `ConnectionHandler` could not be enriched with the local node's external addresses from the `on_*` event handlers. Instead, one had to wait for the `NetworkBehaviour::poll` method to access the addresses.

https://github.com/libp2p/rust-libp2p/blob/68d0f882bd7fc0b4a83a11d0d034ae9f085f3e62/protocols/dcutr/src/behaviour_impl.rs#L349-L363

Now that https://github.com/libp2p/rust-libp2p/pull/3153 is merged, event handlers have access to the external addresses right away, thus the `ActionBuilder` abstraction is no longer needed. Let's remove it.

thomaseizinger:
Done in https://github.com/libp2p/rust-libp2p/pull/3304.

PR #3304, opened 2023-01-02
refactor(dcutr): remove ActionBuilder.

jxs:
## Description

addresses #3299

## Open Questions

We are now getting the `observed_addresses` at different places, i.e. previously we had `PollParameters` on `poll` with what I assume would be the list of updated `external_addresses` that would be then used to calculate the list of `observed_addresses`. Now we get the list of `observed_addresses` from the `ExternalAddrs` which is updated on `NewExternalAddr`, but will it make a difference that it might be updated when an `Event` has already been added to `queued_events`?

## Change checklist

### DIFF
--- a/protocols/dcutr/CHANGELOG.md
+++ b/protocols/dcutr/CHANGELOG.md
@@ -1,5 +1,7 @@
 # 0.9.0 [unreleased]
 
+- Remove no longer required `ActionBuilder` from `Behaviour` implementation. See [PR 3304].
+
 - Update to `libp2p-core` `v0.39.0`.
 
 - Update to `libp2p-swarm` `v0.42.0`.
@@ -15,6 +17,7 @@
 [PR 3153]: https://github.com/libp2p/rust-libp2p/pull/3153
 [issue 2217]: https://github.com/libp2p/rust-libp2p/issues/2217
 [PR 3214]: https://github.com/libp2p/rust-libp2p/pull/3214
+[PR 3304]: https://github.com/libp2p/rust-libp2p/pull/3304
 
 # 0.8.0
 
--- a/protocols/dcutr/CHANGELOG.md
+++ b/protocols/dcutr/CHANGELOG.md
@@ -1,7 +1,5 @@
 # 0.9.0 [unreleased]
 
-- Remove no longer required `ActionBuilder` from `Behaviour` implementation. See [PR 3304].
-
 - Update to `libp2p-core` `v0.39.0`.
 
 - Update to `libp2p-swarm` `v0.42.0`.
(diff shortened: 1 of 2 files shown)


mxinden:
> We are now getting the `observed_addresses` at different places, i.e. previously we had `PollParameters` on `poll` with what I assume would be the list of updated `external_addresses` that would be then used to calculate the list of `observed_addresses`. Now we get the list of `observed_addresses` from the `ExternalAddrs` which is updated on `NewExternalAddr`, but will it make a difference that it might be updated when an `Event` has already been added to `queued_events`?

Sharp thinking!

I would expect `NewExternalAddr` to be triggered somewhere on the order of magnitude of <100 over the lifetime of a node, with a "burst" at start-up time.

I would expect the time between an `on_connection_event` (receiving the external address request from the `ConnectionHandler`) and a `poll` (returning the external addresses via the `Swarm` to the `ConnectionHandler`) to be below the order of ms.

With the above in mind, I would deem the probability that we are missing an external address very low, where missing an address does not even mean a failed direct-connection-upgrade in the first place. Also note that a failed direct-connection-upgrade is retried 2 times.

Does the above reasoning make sense?

jxs:
> > We are now getting the `observed_addresses` at different places, i.e. previously we had `PollParameters` on `poll` with what I assume would be the list of updated `external_addresses` that would be then used to calculate the list of `observed_addresses`. Now we get the list of `observed_addresses` from the `ExternalAddrs` which is updated on `NewExternalAddr`, but will it make a difference that it might be updated when an `Event` has already been added to `queued_events`?
> 
> Sharp thinking!
> 
> I would expect `NewExternalAddr` to be triggered somewhere on the order of magnitude of <100 over the lifetime of a node, with a "burst" at start-up time.
> 
> I would expect the time between an `on_connection_event` (receiving the external address request from the `ConnectionHandler`) and a `poll` (returning the external addresses via the `Swarm` to the `ConnectionHandler`) to be below the order of ms.
> 
> With the above in mind, I would deem the probability that we are missing an external address very low, where missing an address does not even mean a failed direct-connection-upgrade in the first place. Also note that a failed direct-connection-upgrade is retried 2 times.
> 
> Does the above reasoning make sense?

yes! Thanks for the explanation Max. :+1:

protocols/dcutr/CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 # 0.9.0 [unreleased]
 
+- Remove no longer required `ActionBuilder` from `Behaviour` implementation. See [PR 3304].
```
mxinden: I suggest not referencing this, given that it is not a user-facing change. Agreed?
```suggestion
```

mxinden (approved):
Small suggestion. Otherwise this looks good to me.

protocols/dcutr/CHANGELOG.md
jxs: Thanks Max, yeah I was not so sure either. Updated!

thomaseizinger (commented):
Nice to see this simplification, well spotted @mxinden !

Gives me confidence that #3153 was a good call!

thomaseizinger (approved):
Whoops, wrong button :)

Issue #3298, opened 2023-01-02
identify: Exchange fails with go-libp2p over QUIC when initialy closing


mxinden:
To receive a remote's identify information one opens a stream and waits for the remote to send the identify payload in that stream.

When connecting to a go-libp2p node over a QUIC connection, identify exchanges fail.

https://github.com/libp2p/rust-libp2p/blob/68d0f882bd7fc0b4a83a11d0d034ae9f085f3e62/protocols/identify/src/protocol.rs#L192-L196

Changing the above to flush instead of close resolves the issue. Note that flushing instead of no action at all is needed in case multistream-select [`V1Lazy`](https://docs.rs/libp2p/latest/libp2p/core/upgrade/enum.Version.html#variant.V1Lazy) is used where the multistream-select message is only send on the stream once userdata is send or it is flushed.

Occurs with the rust-libp2p punchr client when connecting to go-libp2p IPFS nodes.

Opening with little details for now to have a central place to track. @elenaf9 in case you need more details, let me know.

mxinden:
> Note that flushing instead of no action at all is needed in case multistream-select [`V1Lazy`](https://docs.rs/libp2p/latest/libp2p/core/upgrade/enum.Version.html#variant.V1Lazy) is used where the multistream-select message is only send on the stream once userdata is send or it is flushed.

This is likely incorrect. That is, a read should be enough according to the below:

https://github.com/libp2p/rust-libp2p/blob/68d0f882bd7fc0b4a83a11d0d034ae9f085f3e62/misc/multistream-select/src/negotiated.rs#L45-L46

dignifiedquire:
Might this be the reason I am not able to get a single hole punch to work with the punchr setup + rust-libp2p-client?

**Edit**: Almost no connections, there is like 1 per hour or so at most that succeeds (been running for a couple of days now)

mxinden:
@dignifiedquire small update on rust-libp2p punchr. With a couple of upcoming patches I am able to achieve ~50% success rate.

- This patch, i.e. `flush` instead of `close`, or even none of either.
- https://github.com/libp2p/rust-libp2p/pull/3288/
- rust-libp2p punchr to use `keep_alive::Behaviour` to await identify.
- rust-libp2p punchr to prioritize QUIC connections to bootstrap nodes to learn QUIC external addresses.

While 50% is still behind go-libp2p, I would consider this a good step forward.

PR #3306, opened 2023-01-05
fix(quic): Trigger Quic as Transport wakeup on dial

mxinden:
## Description

Scenario: rust-libp2p node A dials rust-libp2p node B. B listens on a QUIC address. A dials B via the `libp2p-quic` `Transport` wrapped in a `libp2p-dns` `Transport`.

Note that `libp2p-dns` in itself is not relevant here. Only the fact that `libp2p-dns` delays a dial is relevant, i.e. that it first does other async stuff (DNS lookup) before creating the QUIC dial. In fact, dialing an IP address through the DNS `Transport` where no DNS resolution is needed triggers the below just fine.

1. A calls `Swarm::dial` which creates a `libp2p-dns` dial.
2. That dial is spawned onto the connection `Pool`, thus starting the DNS resolution.
3. A continuously calls `Swarm::poll`.
4. `libp2p-quic` `Transport::poll` is called, finding no dialers in `self.dialer` given that the spawned dial is still only resolving the DNS address.
5. On the spawned connection task:
    1. The DNS resolution finishes.
    2. Thus calling `Transport::dial` on `libp1p-quic` (note that the DNS dial has a clone of the QUIC `Transport` wrapped in an `Arc<Mutex<_>>`).
    3. That adds a dialer to `self.dialer`. Note that there are no listeners, i.e. `Swarm::listen_on` was never called.
    4. `DialerState::new_dial` is called which adds a message to `self.pending_dials` and wakes `self.waker`. Given that on the last `Transport::poll` there was no `self.dialer`, that waker is empty.

Result: The message is stuck in the `DialerState::pending_dials`. The message is never send to the endpoint driver. The dial never succeeds.

This commit fixes the above, waking the `<Quic as Transport>:poll` method.

## Notes

Steps to reproduce:

- Server

    - Checkout https://github.com/mxinden/rust-libp2p-server/.
    - Run via `cargo run`

- Client
    - Checkout https://github.com/mxinden/libp2p-lookup/pull/60
    - Point to patched rust-libp2p fixing https://github.com/libp2p/rust-libp2p/issues/3298
    - Run `RUST_LOG=debug cargo run -- direct --address /ip4/127.0.0.1/udp/4001/quic`

## Change checklist

### DIFF
--- a/transports/quic/tests/smoke.rs
+++ b/transports/quic/tests/smoke.rs
@@ -1,12 +1,15 @@
 #![cfg(any(feature = "async-std", feature = "tokio"))]
 
 use futures::channel::{mpsc, oneshot};
+use futures::future::BoxFuture;
 use futures::future::{poll_fn, Either};
 use futures::stream::StreamExt;
 use futures::{future, AsyncReadExt, AsyncWriteExt, FutureExt, SinkExt};
+use futures_timer::Delay;
 use libp2p_core::either::EitherOutput;
 use libp2p_core::muxing::{StreamMuxerBox, StreamMuxerExt, SubstreamBox};
 use libp2p_core::transport::{Boxed, OrTransport, TransportEvent};
+use libp2p_core::transport::{ListenerId, TransportError};
 use libp2p_core::{multiaddr::Protocol, upgrade, Multiaddr, PeerId, Transport};
 use libp2p_noise as noise;
 use libp2p_quic as quic;
@@ -19,6 +22,10 @@ use std::io;
 use std::num::NonZeroU8;
 use std::task::Poll;
 use std::time::Duration;
+use std::{
+    pin::Pin,
+    sync::{Arc, Mutex},
+};
 
 #[cfg(feature = "tokio")]
 #[tokio::test]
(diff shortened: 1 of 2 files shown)


elenaf9 (commented):
Thanks for the detailed bug description @mxinden, very much appreciated!

Looking through the code again, we basically always want to wake our `Transport::poll` if a new dial happens. 
I think we can remove the `waker` from `DialerState` and instead  in `GenTransport::dial` simply always call `dial_waker.wake()`, independently of whether we then use a listener/ existing dialer or create a new dialer. `GenTransport::poll` will poll all of them again anyway.

Only if long-term we switch to a more fine-grained  polling logic (i.e. only poll something again if this structure triggered the waker) this would have to be reconsidered. But for now I think having a single waker is cleaner.

mxinden:
> Looking through the code again, we basically always want to wake our `Transport::poll` if a new dial happens. I think we can remove the `waker` from `DialerState` and instead in `GenTransport::dial` simply always call `dial_waker.wake()`, independently of whether we then use a listener/ existing dialer or create a new dialer. `GenTransport::poll` will poll all of them again anyway.

Sounds good to me. See recent changes.

@elenaf9 thanks for the review. Would you mind taking another look?

transports/quic/tests/smoke.rs
```diff
@@ -90,6 +97,113 @@ async fn ipv4_dial_ipv6() {
     assert_eq!(b_connected, a_peer_id);
 }
 
+/// Tests that a [`Transport::dial`] wakes up the task previously polling [`Transport::poll`].
+///
+/// See https://github.com/libp2p/rust-libp2p/pull/3306 for context.
+#[cfg(feature = "async-std")]
+#[async_std::test]
+async fn wrapped_with_dns() {
```
elenaf9, line 105: :pray:

transports/quic/tests/smoke.rs
```diff
@@ -90,6 +97,113 @@ async fn ipv4_dial_ipv6() {
...
+    });
+
+    // Spawn b
+    //
+    // Note that the dial is spawned on a different task than the transport allowing the transport
+    // task to poll the transport once and then suspend, waiting for the wakeup from the dial.
+    let dial = async_std::task::spawn({
+        let dial = b_transport.dial(a_addr).unwrap();
+        async { dial.await.unwrap().0 }
+    });
```
elenaf9, line 198: Nit: isn't it sufficient that a new task is already spawned for polling the transport future below? 
We could just create the dial-future here, and then drive it in the `future::join` below. If we already spawn a task for the dial here, couldn't it happen that the dial resolves before the executor starts polling the transport task below (since with QUIC, the endpoint runs in a separate background task and thus things "run" even if the transport is not being polled)?

elenaf9 (approved):
One more comment, and clippy is complaining. Otherwise this looks good to me.
Thanks @mxinden!

transports/quic/tests/smoke.rs
```diff
@@ -90,6 +97,113 @@ async fn ipv4_dial_ipv6() {
...
 
+/// Tests that a [`Transport::dial`] wakes up the task previously polling [`Transport::poll`].
+///
+/// See https://github.com/libp2p/rust-libp2p/pull/3306 for context.
+#[cfg(feature = "async-std")]
+#[async_std::test]
+async fn wrapped_with_dns() {
+    let _ = env_logger::try_init();
+
+    struct FakeDns(Arc<Mutex<Boxed<(PeerId, StreamMuxerBox)>>>);
```
thomaseizinger: Call this `DialDelay` maybe?

transports/quic/tests/smoke.rs
```diff
@@ -90,6 +97,113 @@ async fn ipv4_dial_ipv6() {
...
+        /// Delayed dial, i.e. calling [`Transport::dial`] on the inner [`Transport`] not within the
+        /// synchronous [`Transport::dial`] method, but within the [`Future`] returned by the outer
+        /// [`Transport::dial`].
+        fn dial(&mut self, addr: Multiaddr) -> Result<Self::Dial, TransportError<Self::Error>> {
+            let t = self.0.clone();
+            Ok(async move {
+                // Simulate DNS lookup. Giving the `Transport::poll` the chance to return
+                // `Poll::Pending` and thus suspending its task, waiting for a wakeup from the dial
+                // on the inner transport below.
+                Delay::new(Duration::from_millis(1)).await;
```
thomaseizinger: This is a bit short. Setting this to 100 wouldn't increase the test run time much because they all run in parallel and would make me feel easier that this actually makes a difference :)

thomaseizinger (approved):

transports/quic/tests/smoke.rs
mxinden: > If we already spawn a task for the dial here, couldn't it happen that the dial resolves before the executor starts polling the transport task below (since with QUIC, the endpoint runs in a separate background task and thus things "run" even if the transport is not being polled)?

Increased the `Delay` to `100ms`, thus I don't think the above is probable. Will keep it as is. @elenaf9 please comment in case you feel strongly about this.

thomaseizinger:
Ever since this was merged, the `libp2p-quic` job hangs forever: https://github.com/libp2p/rust-libp2p/actions/runs/3893847657/jobs/6647025747

thomaseizinger:
I'd recommend we do two things:

1. Revert this PR to unblock others
2. Add `libp2p-quic` to the list of required jobs so this doesn't happen again: https://github.com/libp2p/github-mgmt/blob/573764714d656f03f35e665f6afd157614486e26/github/libp2p.yml#L4292C2-L4338

thomaseizinger:
Revert PR submitted here: https://github.com/libp2p/rust-libp2p/pull/3322

thomaseizinger:
Adding `libp2p-quic` and `libp2p-webrtc` to required jobs here: https://github.com/libp2p/github-mgmt/pull/104

elenaf9:
Sorry, this was my bad. The issue is in the above solution I proposed i.e. to only have a single waker for the transport instead of having the waker in the `DialState`. I wrongly assumed to following:

> Only if long-term we switch to a more fine-grained polling logic (i.e. only poll something again if this structure triggered the waker) this would have to be reconsidered. But for now I think having a single waker is cleaner.

It's true that we ourselves don't implement fine-grained polling. What I missed however is that we use a `SelectAll` for polling our Listeners. And the `SelectAll` does such fine-grained polling so that it only polls a listener if the listener registered a waker. 
With this PR's change the waker was not registered from within the `Listener`'s `Stream` implementation anymore, causing the listener to not be polled again.

Will do a PR with the fix. I'll try to think of an alternative solution to handle our wakers in a nice way, or else fall back to the original fix from @mxinden.

Issue #2894, opened 2022-09-12
swarm/handler: Split `inject_listen_upgrade_error` into two


mxinden:
## Description

Split the `ConectionHandler::inject_listen_upgrade_error` into two methods:

1. `inject_listen_negotiation_error` for errors during the multistream select protocol execution. Not specifically related to the protocol of this `ConnectionHandler`.
2. `inject_listen_upgrade_error` for errors during the actualy `InboundUpgrade`. Specific to the protocol of this `ConnectionHandler`.

## Motivation

Using the same method for both use-cases is error prone. E.g. see `libp2p-relay` where we print an error for a potentially unrelated issue:

https://github.com/libp2p/rust-libp2p/blob/66c275520d8cf56152cce9a1c24bb4e2c42e0833/protocols/relay/src/v2/client/handler.rs#L379-L383

## Current Implementation

Today `ConnectionHandler::inject_listen_upgrade_error` is called for both multistream select negotiation errors and protocol specific upgrade errors.

## Are you planning to do it yourself in a pull request?

No

//CC @dignifiedquire as we ran into this during our recent debugging session.

Still remember discussing this a long time ago with @romanb.

thomaseizinger:
I am in favor of this. The error nesting is quite annoying to deal with!

maschad:
I can take a look into it this week.

mxinden:
> 1. `inject_listen_negotiation_error` for errors during the multistream select protocol execution. Not specifically related to the protocol of this `ConnectionHandler`.

In case neither us nor any user ever makes use of this method, we should consider not introducing it in the first place.

mxinden:
> Thanks for the update @jxs I am bit preoccupied with some issues on js-libp2p so I haven't had a chance to work on this. Perhaps it may be best to assign this to someone else?

https://github.com/libp2p/rust-libp2p/pull/3098#issuecomment-1348526516

@jxs would you be interested in tackling this one?

jxs:
no worries Max, I'll take it!

maschad:
> no worries Max, I'll take it!

Thanks @jxs !

jxs:
update: waiting on https://github.com/libp2p/rust-libp2p/pull/3264 to be merged to start tackling this.

jxs:
@mxinden there's an issue with the separation of errors, 
`DialUpgradeError` error can also be a `Timeout` see
[...]

PR #3307, opened 2023-01-05
refactor(swarm)!: refactor ListenUpgradeError and DialUpgradeError

jxs:
## Description

Addresses #2894

## Notes

In the protocols, moved the `TimedOut` variants from nested `UpgradeError`'s to `Event` variants themselves. Like `identify` `dcutr` and `relay`. 
**Per commit review is suggested.**

## Change checklist

### DIFF
--- a/protocols/dcutr/src/behaviour_impl.rs
+++ b/protocols/dcutr/src/behaviour_impl.rs
@@ -25,12 +25,13 @@ use crate::protocol;
 use either::Either;
 use libp2p_core::connection::{ConnectedPoint, ConnectionId};
 use libp2p_core::multiaddr::Protocol;
+use libp2p_core::UpgradeError;
 use libp2p_core::{Multiaddr, PeerId};
 use libp2p_swarm::behaviour::{ConnectionClosed, ConnectionEstablished, DialFailure, FromSwarm};
 use libp2p_swarm::dial_opts::{self, DialOpts};
 use libp2p_swarm::{
-    ConnectionHandler, ConnectionHandlerUpgrErr, ExternalAddresses, IntoConnectionHandler,
-    NetworkBehaviour, NetworkBehaviourAction, NotifyHandler, PollParameters,
+    ConnectionHandler, ExternalAddresses, IntoConnectionHandler, NetworkBehaviour,
+    NetworkBehaviourAction, NotifyHandler, PollParameters,
 };
 use std::collections::{HashMap, HashSet, VecDeque};
 use std::task::{Context, Poll};
@@ -56,6 +57,9 @@ pub enum Event {
         remote_peer_id: PeerId,
         error: Error,
     },
+    DirectConnectionUpgradeTimedout {
+        remote_peer_id: PeerId,
+    },
 }
 
 #[derive(Debug, Error)]
@@ -63,7 +67,7 @@ pub enum Error {
     #[error("Failed to dial peer.")]
     Dial,
     #[error("Failed to establish substream: {0}.")]
-    Handler(ConnectionHandlerUpgrErr<void::Void>),
+    Handler(UpgradeError<void::Void>),
 }
 
 pub struct Behaviour {
@@ -237,6 +241,14 @@ impl NetworkBehaviour for Behaviour {
                     NetworkBehaviourAction::GenerateEvent(Event::DirectConnectionUpgradeFailed {
                         remote_peer_id: event_source,
                         error: Error::Handler(error),
+                    })
+                    .into(),
+                );
+            }
+            Either::Left(handler::relayed::Event::InboundNegotiationTimedout) => {
+                self.queued_actions.push_back(
+                    NetworkBehaviourAction::GenerateEvent(Event::DirectConnectionUpgradeTimedout {
+                        remote_peer_id: event_source,
                     })
                     .into(),
                 );
--- a/protocols/dcutr/src/behaviour_impl.rs
+++ b/protocols/dcutr/src/behaviour_impl.rs
@@ -57,9 +57,6 @@ pub enum Event {
         remote_peer_id: PeerId,
         error: Error,
     },
-    DirectConnectionUpgradeTimedout {
-        remote_peer_id: PeerId,
-    },
 }
 
 #[derive(Debug, Error)]
@@ -68,6 +65,8 @@ pub enum Error {
     Dial,
     #[error("Failed to establish substream: {0}.")]
     Handler(UpgradeError<void::Void>),
+    #[error("Timeout expired upgrading the direct connection.")]
+    Timeout,
 }
 
 pub struct Behaviour {
@@ -247,8 +246,9 @@ impl NetworkBehaviour for Behaviour {
             }
             Either::Left(handler::relayed::Event::InboundNegotiationTimedout) => {
                 self.queued_actions.push_back(
-                    NetworkBehaviourAction::GenerateEvent(Event::DirectConnectionUpgradeTimedout {
+                    NetworkBehaviourAction::GenerateEvent(Event::DirectConnectionUpgradeFailed {
                         remote_peer_id: event_source,
+                        error: Error::Timeout,
                     })
                     .into(),
                 );
--- a/protocols/identify/src/behaviour.rs
+++ b/protocols/identify/src/behaviour.rs
@@ -25,9 +25,9 @@ use libp2p_core::{
 };
 use libp2p_swarm::behaviour::{ConnectionClosed, ConnectionEstablished, DialFailure, FromSwarm};
 use libp2p_swarm::{
-    dial_opts::DialOpts, AddressScore, ConnectionHandler, ConnectionHandlerUpgrErr, DialError,
-    ExternalAddresses, IntoConnectionHandler, ListenAddresses, NetworkBehaviour,
-    NetworkBehaviourAction, NotifyHandler, PollParameters,
+    dial_opts::DialOpts, AddressScore, ConnectionHandler, DialError, ExternalAddresses,
+    IntoConnectionHandler, ListenAddresses, NetworkBehaviour, NetworkBehaviourAction,
+    NotifyHandler, PollParameters,
 };
 use lru::LruCache;
 use std::num::NonZeroUsize;
@@ -303,6 +303,9 @@ impl NetworkBehaviour for Behaviour {
                         error,
                     }));
             }
+            handler::Event::IdentificationTimedout => self
+                .events
+                .push_back(NetworkBehaviourAction::GenerateEvent(Event::Timeout)),
         }
     }
 
@@ -459,8 +462,11 @@ pub enum Event {
         /// The peer with whom the error originated.
         peer_id: PeerId,
         /// The error that occurred.
-        error: ConnectionHandlerUpgrErr<UpgradeError>,
+        error: libp2p_core::UpgradeError<UpgradeError>,
     },
+
+    /// Timeout expired while attempting to identify the remote.
+    Timeout,
 }
 
 fn supported_protocols(params: &impl PollParameters) -> Vec<String> {
--- a/protocols/identify/src/behaviour.rs
+++ b/protocols/identify/src/behaviour.rs
@@ -38,6 +38,7 @@ use std::{
     task::Poll,
     time::Duration,
 };
+use thiserror::Error;
 
 /// Network behaviour that automatically identifies nodes periodically, returns information
 /// about them, and answers identify queries from other nodes.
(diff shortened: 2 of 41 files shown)


swarm/CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 # 0.42.0 [unreleased]
 
 - Remove uncontructed variant `Timer` from `ConnectionHandlerUpgrErr`.
+  Make `ListenUpgradeError::error` an `UpgradeError` instead of `ConnectionHandlerUpgrErr`, in case of timeout expiration we
+  just log the error as we don't know which ConnectionHandler should handle the error.
```
thomaseizinger: These need a reference to this PR.

swarm/src/connection.rs
```diff
@@ -239,12 +239,17 @@ where
...
-                }
+                Poll::Ready(Some((info, Err(error)))) => match error {
+                    ConnectionHandlerUpgrErr::Upgrade(error) => {
+                        handler.on_connection_event(ConnectionEvent::ListenUpgradeError(
+                            ListenUpgradeError { info, error },
+                        ));
+                        continue;
+                    }
+                    ConnectionHandlerUpgrErr::Timeout => {
+                        log::debug!("Timeout expired during an inbound substream negotiation")
```
thomaseizinger, line 253:
```suggestion
                        log::debug!("Failed to negotiate protocol on inbound stream within {seconds}s")
```

That would be more useful IMO.

swarm/src/handler.rs
```diff
@@ -207,6 +207,8 @@ pub enum ConnectionEvent<'a, IP: InboundUpgradeSend, OP: OutboundUpgradeSend, IO
     AddressChange(AddressChange<'a>),
     /// Informs the handler that upgrading an outbound substream to the given protocol has failed.
     DialUpgradeError(DialUpgradeError<OOI, OP>),
+    /// Informs the handler that upgrading an outbound substream to the given protocol has expired its timeout.
+    DialTimeout(DialTimeout<OOI>),
```
thomaseizinger, line 211: Would it make sense for `DialUpgradeError` to be an enum with a `Timeout` and `Upgrade` variant?

thomaseizinger: https://github.com/libp2p/rust-libp2p/issues/3268 is related to this. If we agree to remove the `OpenInfo` from `ConnectionHandler` a lot of this code would be simpler because we wouldn't need to track `info` here.

swarm/src/lib.rs
```diff
@@ -113,10 +113,13 @@ pub use connection::{
...
-    ConnectionHandler, ConnectionHandlerEvent, ConnectionHandlerSelect, ConnectionHandlerUpgrErr,
-    IntoConnectionHandler, IntoConnectionHandlerSelect, KeepAlive, OneShotHandler,
-    OneShotHandlerConfig, SubstreamProtocol,
+    ConnectionHandler, ConnectionHandlerEvent, ConnectionHandlerSelect, IntoConnectionHandler,
+    IntoConnectionHandlerSelect, KeepAlive, OneShotHandler, OneShotHandlerConfig,
+    SubstreamProtocol,
 };
+
+#[deprecated(since = "0.42.0", note = "Shouldn't be required by end users")]
+pub type ConnectionHandlerUpgrErr<TUpgrErr> = handler::ConnectionHandlerUpgrErr<TUpgrErr>;
```
thomaseizinger, line 122: :clap:

protocols/dcutr/src/behaviour_impl.rs
```diff
@@ -56,14 +57,17 @@ pub enum Event {
...
+    },
 }
 
 #[derive(Debug, Error)]
 pub enum Error {
     #[error("Failed to dial peer.")]
     Dial,
     #[error("Failed to establish substream: {0}.")]
-    Handler(ConnectionHandlerUpgrErr<void::Void>),
+    Handler(UpgradeError<void::Void>),
```
thomaseizinger, line 67: Previously, we emitted an error (and thus closed the connection) when we ran into a timeout. Why are we changing this behaviour now?

protocols/identify/src/behaviour.rs
```diff
@@ -459,8 +462,11 @@ pub enum Event {
         /// The peer with whom the error originated.
         peer_id: PeerId,
         /// The error that occurred.
-        error: ConnectionHandlerUpgrErr<UpgradeError>,
+        error: libp2p_core::UpgradeError<UpgradeError>,
     },
+
+    /// Timeout expired while attempting to identify the remote.
+    Timeout,
```
thomaseizinger: A timeout is an error so it seems odd that we have an `Error` variant and then this. I'd suggest we either log this internally and don't report it or create a dedicated error enum ourselves.

thomaseizinger (commented):
I love the simplifications coming from this!

I don't quite agree with reporting a kind of error (timeout) next to other errors. I think that should be unified or not reported at all. I am tending towards the latter for a lot of protocols where we can make a decision on what that timeout means.

swarm/src/connection.rs
jxs: `seconds` doesn't seem to exist in this scope, we probably can make `ConnectionHandlerUpgrErr::Timeout` have it . I will implement after clarifying https://github.com/libp2p/rust-libp2p/pull/3307#discussion_r1064264754

swarm/src/handler.rs
jxs: At first sight, making `DialUpgradeError` an enum with a `Timeout` and `Upgrade` variant sounded similar to what we currently have, where `ConnectionHandlerUpgrErr` then has the variants : https://github.com/libp2p/rust-libp2p/blob/5b8c3d290187b888b2afc8288c8e0b5843ef4964/swarm/src/handler.rs#L340-L343

But what you are suggesting seems to be in fact renaming `ConnectionHandlerUpgradeError` to `DialUpgradeError` and using it in `ConnectionEvent::DialUpgradeError` right? 
https://github.com/libp2p/rust-libp2p/blob/021502ccd20b67a20d99f1af9c2c53ac659e170c/swarm/src/handler.rs#L442-L448

 In fact it probably makes sense (we are in fact reducing one indirection layer) even more sense than having `Timeout` as root variants where I agree with you that they are errors by themselves. Do you also agree @mxinden?

protocols/identify/src/behaviour.rs
jxs: Since we are already reporting it up I'd go as we do in `dcutr` and expose it as an `Error`, updated to do that! Does it make sense to you to also do the same for `relay`?

protocols/dcutr/src/behaviour_impl.rs
jxs: Isn't the behaviour the same? In the sense that the connection was closed because of a timeout and we are bubbling it up? But after reading and replying on https://github.com/libp2p/rust-libp2p/pull/3307#discussion_r1064264754 I now see that Timeout makes sense being a variant on the `Error` itself of `DirectConnectionUpgradeFailed` rather than a variant of Event.

swarm/src/handler.rs
mxinden: Off the top of my head this sounds good to me. I will have to give this more thought with a clear mind to give a good answer.

swarm/CHANGELOG.md
jxs: yeah, it was done on 11db523b7d094f6ed1a73d32d332d534525c95db

swarm/src/connection.rs
thomaseizinger: > `seconds` doesn't seem to exist in this scope, we probably can make `ConnectionHandlerUpgrErr::Timeout` have it . I will implement after clarifying https://github.com/libp2p/rust-libp2p/pull/3307#discussion_r1064264754

Yes, it was only meant to be illustrative on what I'd like to have printed!

protocols/dcutr/src/behaviour_impl.rs
thomaseizinger: > Isn't the behaviour the same? In the sense that the connection was closed because of a timeout and we are bubbling it up?

Unless I am mistaken, the timeout refers to a stream upgrade. The connection can/should still be intact at this stage. Something like `PendingUpgrade` should trigger this timeout on the remote but the connection underneath works perfectly fine.

It does raise the question whether we should close an entire connection at all just because one protocol failed to upgrade a stream. Just disabling that protocol (and handler) and letting the collaborative keep-alive eventually close the connection seems like a better idea at first sight.

We may want to do this in a separate PR though, either before or after we merge this.

protocols/identify/src/behaviour.rs
thomaseizinger: > Does it make sense to you to also do the same for `relay`?

Yes I think we should apply that thinking to all protocols.

jxs: thanks done, ptal! Also removed what seems to be unrequired `inbound_hop::UpgradeError` and shortcircuit to `inbound_hop::FatalUpgradeError`  see c2eda37b0997e2cf7bc306db2e63796e3c9a069a and see if you agree.

thomaseizinger:
Sorry for only mentioning this only now but it just came to me that once we have https://github.com/libp2p/rust-libp2p/issues/2863, all of these errors will go away because there won't be any "upgrade" mechanism anymore for protocols defined through `ConnectionHandler`s.

Is it worth doing this refactoring here at all or should we just work towards https://github.com/libp2p/rust-libp2p/issues/2863 instead? Personally, that would be my preference but it is also more work.

Thoughts?

jxs:
> Sorry for only mentioning this only now but it just came to me that once we have #2863, all of these errors will go away because there won't be any "upgrade" mechanism anymore for protocols defined through `ConnectionHandler`s.
> 
> Is it worth doing this refactoring here at all or should we just work towards #2863 instead? Personally, that would be my preference but it is also more work.
> 
> Thoughts?

No worries Thomas. Good question, no strong opinion though I think it probably doesn't makes sense to introduce this and release it, make libp2p users upgrade to this to then upgrade again with #2863 right? So I'd rather just jump to #2863 CC @mxinden wdyt?

thomaseizinger:
> I think it probably doesn't makes sense to introduce this and release it, make libp2p users upgrade to this to then upgrade again with #2863 right?

Yeah I thought so too which is why I brought it up. I updated the issue with a few more thoughts on how we could move forward there.

thomaseizinger:
Whilst hacking on https://github.com/libp2p/rust-libp2p/pull/3201, I got very confused in how the error handling for connection and upgrade errors is implemented across our protocols.

At the highest level, we have `ConnectionHandlerUpgrErr` which can be emitted for two different cases:

- Inbound streams
- Outbound streams

Confusingly enough, we call those `ListenUpgradeError` and `DialUpgradeError`. Usually we only use the listen and dial terminology for connections but not for streams.

`ConnectionHandlerUpgrErr` has three variants:

https://github.com/libp2p/rust-libp2p/blob/e3c7023350ecfb1a932fe8c87423886068b9bf06/swarm/src/handler.rs#L433-L442

As you @jxs discovered in here, the `Timer` variant is not actually used.

The `Timeout` variant will go away once we ship https://github.com/libp2p/rust-libp2p/issues/2863 because we won't be upgrading streams anymore.

What is left here is `UpgradeError`:

https://github.com/libp2p/rust-libp2p/blob/e3c7023350ecfb1a932fe8c87423886068b9bf06/core/src/upgrade/error.rs#L24-L31

Once we ship https://github.com/libp2p/rust-libp2p/issues/2863, the `Apply` variant here should also go away because we will directly hand the stream to the user and not perform any processing on it.

Meaning the only error that can actually happen after #2863 is `NegotiationError`. This one is a also a bit odd because it combines the `Failed` variant with `ProtocolError`.

- `ProtocolError` is fatal, it means some really bad error happened during our handshake
- `Failed` means we couldn't agree on _any_ protocol. This one is not necessarily fatal, it just means the remote does not support the protocol we requested.

I am drawing three conclusions from this:

- We should prioritize https://github.com/libp2p/rust-libp2p/issues/2863 as it will make the entire error handling situation much clearer.
- We are currently mixing fatal and non-fatal errors. Most of our protocols close the entire _connection_ on any error. That is not very nice. A protocol should simply disable itself if the remote doesn't support it. I think the best way to achieve this is by splitting the "not supported" part into a separate error that users can just ignore for the most part.
- Fatal errors should perhaps close the connection unconditionally? If we fail to negotiate protocols with an IO error or something similar, there isn't much a user can do about that. Why report it to the handlers?

mxinden:
Agreed with everything in the latest comment from @thomaseizinger. Thanks for looking so deeply into this.

> I think it probably doesn't makes sense to introduce this and release it, make libp2p users upgrade to this to then upgrade again with #2863 right? So I'd rather just jump to #2863 CC @mxinden wdyt?

I don't have an opinion on how to proceed, e.g. whether to first finish here, do https://github.com/libp2p/rust-libp2p/issues/2863, or clean-up fatal and non-fatal errors. Up to you @jxs and @thomaseizinger.

thomaseizinger:
> > I think it probably doesn't makes sense to introduce this and release it, make libp2p users upgrade to this to then upgrade again with #2863 right? So I'd rather just jump to #2863 CC @mxinden wdyt?
> 
> I don't have an opinion on how to proceed, e.g. whether to first finish here, do https://github.com/libp2p/rust-libp2p/issues/2863, or clean-up fatal and non-fatal errors. Up to you @jxs and @thomaseizinger.

Any work that pushes #2863 forward would be much appreciated from my end!

PR #3311, opened 2023-01-06
remove usage of Code::Identity in core/src/peer_id.rs

### LINKED ISSUE #3276: Don't depend on the multihash identity feature
This feature is broken and will be removed eventually. To avoid deprecation warnings for our users that they can't fix, let's remove it here first.

See:

- https://github.com/multiformats/rust-multihash/pull/260
- https://github.com/multiformats/rust-multihash/issues/194

jmmaloney4:
## Description

Remove usage of Code::Identity in core/src/peer_id.rs, to fix #3276.

 #3276

## Open Questions

I'm unsure how to test this change thoroughly. I couldn't get `cargo test` to compile, even on the master branch.

## Change checklist

### DIFF
--- a/core/src/peer_id.rs
+++ b/core/src/peer_id.rs
@@ -28,10 +28,6 @@ use thiserror::Error;
 #[cfg(feature = "serde")]
 use serde::{Deserialize, Serialize};
 
-/// Public keys with byte-lengths smaller than `MAX_INLINE_KEY_LENGTH` will be
-/// automatically used as the peer id using an identity multihash.
-const MAX_INLINE_KEY_LENGTH: usize = 42;
-
 /// Identifier of a peer of the network.
 ///
 /// The data is a CIDv0 compatible multihash of the protobuf encoded public key of the peer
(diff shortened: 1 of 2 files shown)


core/src/peer_id.rs
```diff
@@ -56,17 +52,9 @@ impl fmt::Display for PeerId {
...
-            Code::Identity
-        } else {
-            Code::Sha2_256
-        };
-
-        let multihash = hash_algorithm.digest(&key_enc);
-
-        PeerId { multihash }
+        PeerId {
+            multihash: Code::Sha2_256.digest(&key.to_protobuf_encoding()),
```
rkuhn, line 56: This will break lots of code without alerting maintainers through compile errors, i.e. this is really bad: Actyx for example only supports ed25519 and then makes use of the fact that the public key can be recovered from the PeerId.

Changing the bit format of PeerId (for the same key) also is a network breaking change, so such a proposal should come with a thorough analysis of the intended and unintended effects as well as a migration plan. And if the only benefit is to avoid a deprecated internal function, then I don’t think this is worth it.

thomaseizinger: Thanks for commenting @rkuhn! There was a misunderstanding on the original issue that I've hopefully been able to clarify! :)

rkuhn: Yes, thanks!

thomaseizinger:
Closing in favor of https://github.com/libp2p/rust-libp2p/pull/3514.

PR #3312, opened 2023-01-07
feat: migrate to quick-protobuf 

### LINKED ISSUE #3024: Migrate from `protoc` to `quick-protobuf` and remove the buildscripts

kckeiks:
## Description

Instead of relying on `protoc` and buildscripts, we generate the bindings using `pb-rs` and version them within our codebase. This makes for a better IDE integration, a faster build and an easier use of `rust-libp2p` because we don't force the `protoc` dependency onto them.

Resolves #3024.

## Change checklist


.github/workflows/ci.yml
```diff
@@ -305,4 +305,22 @@ jobs:
...
+
+      - uses: actions-rs/toolchain@16499b5e05bf2e26879000db0c1d13f7e13fa3af # v1.0.7
+        with:
+          profile: minimal
+          toolchain: stable
+          override: true
+
+      - name: Generated proto files and diff
+        run: |
+          cargo install pb-rs
```
thomaseizinger: Two things: Use `--locked` and if you add the cache action before this, we don't have to reinstall it every time.

.github/workflows/ci.yml
```diff
@@ -305,4 +305,22 @@ jobs:
...
+        with:
+          profile: minimal
+          toolchain: stable
+          override: true
+
+      - name: Generated proto files and diff
+        run: |
+          cargo install pb-rs 
+          pb-rs "**/generated/**.proto"
+          git diff --exit-code ${{ github.ref }}
```
thomaseizinger: Isn't the `github.ref` part redundant here? Git should automatically diff to that commit right?

thomaseizinger (approved):
Nice, this is what I was looking for, thanks!

Once we delete the other proto files, those should hopefully show up as moved rather than added.

thomaseizinger:
Dang, just saw CI doesn't actually like this. I am no bash expert but maybe try globbing into a variable first?

core/src/generated/mod.rs
```diff
@@ -0,0 +1,4 @@
+// Automatically generated mod.rs
```
mxinden, line 1: Note that we don't use `mod.rs` files anywhere else within this repository. Is this necessary to function or would a `generated.rs` file one level above work as well?

mxinden (commented):
Thanks to both of you for the work here!

core/src/generated/mod.rs
thomaseizinger:
```
❯ pb-rs --help
pb-rs 0.10.0
Johann Tuffe <tafia973@gmail.com>
A converter from proto files into quick-protobuf compatible Rust module

USAGE:
    pb-rs [FLAGS] [OPTIONS] [INPUT]...

FLAGS:
        --add-deprecated-fields    Add deprecated fields and mark them as #[deprecated]
    -e, --error-cycle              Error out if recursive messages do not have optional fields
    -D, --dont_use_cow             Don't use Cow for String and Byte types
        --gen-info                 Generate MessageInfo implementations
        --hashrown                 Use hashrown for HashMap implementation
        --nostd                    Generate no_std compliant code
    -H, --no-headers               Do not add module comments and module attributes in generated file
    -n, --no-output                Show enums and messages in this .proto file, including those imported. No code
                                   generated
        --owned                    Generate Owned structs when the proto stuct has a lifetime
    -s, --single-mod               Omit generation of modules for each package when there is only one package
    -h, --help                     Prints help information
    -V, --version                  Prints version information

OPTIONS:
    -R, --custom_repr <CUSTOM_REPR>                      The value to use for the optional #[repr(...)] for every struct
    -C, --custom_struct_derive <CUSTOM_STRUCT_DERIVE>
            The comma separated values to add to #[derive(...)] for every struct

    -I, --include <INCLUDE_PATH>                         Path to search for imported protobufs
    -o, --output <OUTPUT>
            Generated file name, defaults to INPUT with 'rs' extension, cannot be used with --output_directory

    -d, --output_directory <OUTPUT_DIR>
            Output directory of generated code, cannot be used with --output


ARGS:
    <INPUT>...    The .proto files used to generate quick-protobuf code
```

There doesn't seem to be an option to not generate a `mod.rs` file.

rkuhn: FWIW: I find `mod.rs` more consistent than the libp2p approach, since it keeps all files of a module in the same place.

kckeiks:
> Dang, just saw CI doesn't actually like this. I am no bash expert but maybe try globbing into a variable first?

I'll try that. I'm no expert on that either so I will probably have to push quite a bit to test it so I apologize in advance. Also, will update this PR soon.

kckeiks:
Looks like it works https://github.com/libp2p/rust-libp2p/actions/runs/3898831952/jobs/6657940569

kckeiks:
Looks like the diff is not working though. This run should have failed. https://github.com/libp2p/rust-libp2p/actions/runs/3898847742/jobs/6657969252

thomaseizinger:
> Looks like the diff is not working though. This run should have failed. [libp2p/rust-libp2p/actions/runs/3898847742/jobs/6657969252](https://github.com/libp2p/rust-libp2p/actions/runs/3898847742/jobs/6657969252)

Why did you expect it to fail? Did you modify any of the generated code in a commit?

kckeiks:
> > Looks like the diff is not working though. This run should have failed. [libp2p/rust-libp2p/actions/runs/3898847742/jobs/6657969252](https://github.com/libp2p/rust-libp2p/actions/runs/3898847742/jobs/6657969252)
> 
> Why did you expect it to fail? Did you modify any of the generated code in a commit?

I modified the glob so every *.proto file should be getting compiled including the ones outside /generated.

thomaseizinger:
> > > Looks like the diff is not working though. This run should have failed. [libp2p/rust-libp2p/actions/runs/3898847742/jobs/6657969252](https://github.com/libp2p/rust-libp2p/actions/runs/3898847742/jobs/6657969252)
> > 
> > 
> > Why did you expect it to fail? Did you modify any of the generated code in a commit?
> 
> I modified the glob so every *.proto file should be getting compiled including the ones outside /generated.

The problem is that those are untracked files so they don't seem to "fail" the `git diff` command. I tried by modifying an existing generated file, committing it and `git diff` fails properly on those.

kckeiks:
> > > > Looks like the diff is not working though. This run should have failed. [libp2p/rust-libp2p/actions/runs/3898847742/jobs/6657969252](https://github.com/libp2p/rust-libp2p/actions/runs/3898847742/jobs/6657969252)
> > > 
> > > 
> > > Why did you expect it to fail? Did you modify any of the generated code in a commit?
> > 
> > 
> > I modified the glob so every *.proto file should be getting compiled including the ones outside /generated.
> 
> The problem is that those are untracked files so they don't seem to "fail" the `git diff` command. I tried by modifying an existing generated file, committing it and `git diff` fails properly on those.

Thanks for the help! 

- Are we good with the CI then? 
- What is the plan with `prost-codec`? Will it be removed by another PR? https://github.com/libp2p/rust-libp2p/pull/3066#issuecomment-1378475251

kckeiks:
I'll start working on removing prost from the crates now then.

thomaseizinger:
> I'll start working on removing prost from the crates now then.

Sounds good! Thank you :)

thomaseizinger:
> > > > > Looks like the diff is not working though. This run should have failed. [libp2p/rust-libp2p/actions/runs/3898847742/jobs/6657969252](https://github.com/libp2p/rust-libp2p/actions/runs/3898847742/jobs/6657969252)
> > > > 
> > > > 
> > > > Why did you expect it to fail? Did you modify any of the generated code in a commit?
> > > 
> > > 
> > > I modified the glob so every *.proto file should be getting compiled including the ones outside /generated.
> > 
> > 
> > The problem is that those are untracked files so they don't seem to "fail" the `git diff` command. I tried by modifying an existing generated file, committing it and `git diff` fails properly on those.
> 
> Thanks for the help!
> 
>  Are we good with the CI then?

Yes I think this is all sorted now!

thomaseizinger:
@kckeiks Is this ready for a new review?

kckeiks:
> @kckeiks Is this ready for a new review?

Yeah, go for it. I think I'm going to start working on [asynchronous-codec](https://github.com/mxinden/asynchronous-codec) in the meantime.

core/src/lib.rs
```diff
@@ -37,18 +37,12 @@
...
-#[allow(clippy::derive_partial_eq_without_eq)]
-mod peer_record_proto {
-    include!(concat!(env!("OUT_DIR"), "/peer_record_proto.rs"));
+mod proto {
+    include!("generated/mod.rs");
+    pub use self::{
+        envelope_proto::*, keys_proto::*, peer_record_proto::mod_PeerRecord::*,
+        peer_record_proto::PeerRecord,
+    };
 }
```
thomaseizinger, line 46: :+1:

core/src/lib.rs
```diff
@@ -37,18 +37,12 @@
...
-
-mod envelope_proto {
-    include!(concat!(env!("OUT_DIR"), "/envelope_proto.rs"));
-}
-
-#[allow(clippy::derive_partial_eq_without_eq)]
-mod peer_record_proto {
-    include!(concat!(env!("OUT_DIR"), "/peer_record_proto.rs"));
+mod proto {
+    include!("generated/mod.rs");
```
thomaseizinger, line 41: Why do we need an `include` here? Wouldn't it be simpler to just declare the `generated` module as a normal module on the root level and then just re-export?

thomaseizinger: Perhaps the `--single-mod` option helps us removing some of this boilerplate?

protocols/autonat/src/protocol.rs
```diff
@@ -150,28 +154,30 @@ impl DialRequest {
...
+        let msg = proto::Message {
+            type_pb: Some(proto::MessageType::DIAL),
+            dial: Some(proto::Dial {
+                peer: Some(proto::PeerInfo {
+                    id: Some(Cow::from(peer_id)),
                     addrs,
                 }),
             }),
-            dial_response: None,
+            dialResponse: None,
```
thomaseizinger, line 175: This is odd. Does `pb-rs` not generate idiomatic Rust identifiers?

thomaseizinger (commented):
Stellar work, very much looking forward to this! Does this encompass all crates already that depend on `prost` currently? (Modulo the ones using `prost-codec`?

core/src/lib.rs
kckeiks: My concern is devs importing from root or proto inconsistently. I can add it to root though, no problem.

protocols/autonat/src/protocol.rs
kckeiks: No it doesn't :/

kckeiks:
> Stellar work, very much looking forward to this! Does this encompass all crates already that depend on `prost` currently? (Modulo the ones using `prost-codec`?

Thanks! There is still a few more to go.

core/src/lib.rs
thomaseizinger: Ah that is a valid concern! Happy to stick to `include` then! I am quite surprised this works though. This just imports the content of the file right? Why doesn't rustc expect those modules to be within a `proto` directory when the modules are declared within a `proto` module?

kckeiks: I think it's because include! parses the file as an expression.

kckeiks:
Update: Still working on this. I'm running into some issues with lifetimes and the generated code. Atm I have this struct in `dcutr`

```rust
pub struct PendingConnect {
    substream: Framed<NegotiatedSubstream, quick_protobuf_codec::Codec<proto::HolePunch<'static>>>,
    remote_obs_addrs: Vec<Multiaddr>,
}
```

Im trying to figure out why the bounds are unsatisfied.

```rust
        self.substream.send(msg).await?;
        let proto::HolePunch { type_pb, .. } = self
            .substream
            .next()
            .await
            .ok_or(UpgradeError::StreamClosed)??;
```

```
error[E0599]: the method `next` exists for struct `Framed<multistream_select::negotiated::Negotiated<SubstreamBox>, Codec<HolePunch<'static>>>`, but its trait bounds were not satisfied
   --> protocols/dcutr/src/protocol/inbound.rs:114:14
    |
114 |               .next()
    |                ^^^^ method cannot be called on `Framed<multistream_select::negotiated::Negotiated<SubstreamBox>, Codec<HolePunch<'static>>>` due to unsatisfied trait bounds
    |
   ::: /Users/acadia/Repo/rust-libp2p/misc/multistream-select/src/negotiated.rs:49:1
    |
49  |   pub struct Negotiated<TInner> {
    |   -----------------------------
    |   |
    |   doesn't satisfy `_: StreamExt`
    |   doesn't satisfy `_: Stream`
    |
   ::: /Users/acadia/.cargo/registry/src/github.com-1ecc6299db9ec823/asynchronous-codec-0.6.1/src/framed.rs:15:1
    |
15  | / pin_project! {
16  | |     /// A unified `Stream` and `Sink` interface to an underlying I/O object,
17  | |     /// using the `Encoder` and `Decoder` traits to encode and decode frames.
18  | |     ///
...   |
44  | |     }
45  | | }
    | | -
    | | |
    | |_doesn't satisfy `_: StreamExt`
    |   doesn't satisfy `_: Stream`
    |
    = note: the following trait bounds were not satisfied:
            `Framed<multistream_select::negotiated::Negotiated<SubstreamBox>, Codec<HolePunch<'static>>>: Stream`
            which is required by `Framed<multistream_select::negotiated::Negotiated<SubstreamBox>, Codec<HolePunch<'static>>>: StreamExt`
            `multistream_select::negotiated::Negotiated<SubstreamBox>: Stream`
            which is required by `multistream_select::negotiated::Negotiated<SubstreamBox>: StreamExt`

For more information about this error, try `rustc --explain E0599`.
error: could not compile `libp2p-dcutr` due to previous error
```

misc/quick-protobuf-codec/src/lib.rs
```diff
@@ -31,21 +30,22 @@ impl<In, Out> Codec<In, Out> {
...
+        item.write_message(&mut writer)
+            .expect("Encoding to succeed");
         self.uvi.encode(encoded_msg.freeze(), dst)?;
 
         Ok(())
     }
 }
 
-impl<In, Out: Message + Default> Decoder for Codec<In, Out> {
+impl<In, Out: for<'a> MessageRead<'a> + Default> Decoder for Codec<In, Out> {
```
thomaseizinger: Have you tried to fix this to `'static`?

thomaseizinger (commented):
One idea, I'll play around with this some time tomorrow.

misc/quick-protobuf-codec/src/lib.rs
kckeiks: That doesn't work because `msg` in `Self::Item::from_reader(&mut reader, &msg)` does not live long enough.

kckeiks:
`dcutr` compiles when changing the lifetime to static.

```rust
impl<In, Out: MessageRead<'static> + Default> Decoder for Codec<In, Out> {
    type Item = Out;
    type Error = Error;

    fn decode(&mut self, src: &mut BytesMut) -> Result<Option<Self::Item>, Self::Error> {
        Ok(None)
    }
}
```
But that is too constraining for what we want to do in that method...hmmmm

kckeiks:
I think these might be related:
- https://github.com/rust-lang/rust/issues/89196
- https://github.com/rust-lang/rust/issues/56556.

Similar misleading error messages too.

thomaseizinger:
I think we might have to give up on `Cow`. The issue is that our `Decoder` implementation creates a local variable for the slice of bytes that it read from the `BytesMut`. Those bytes are "moved" i.e. the local function has ownership of them. But the `MessageRead` trait doesn't take ownership of the bytes. Instead, it decodes the message while referencing those bytes. If you use `Cow`, the implementation will retain this lifetime and point to the slice of bytes that you gave it. In order to make this work for us, we'd have to create a `Decoder` trait that can return items that point to initially read buffer (via GATs probably).

I don't think this is worth doing at this stage. If we use the `-D` option of `pb-rs`, we will get structs without `Cow`. I pushed a commit that does that. It doesn't fully compile as there are a few more crates to convert from `prost` to `quick-protobuf` but `libp2p-relay` compiles without errors now.

kckeiks:
> I think we might have to give up on `Cow`. The issue is that our `Decoder` implementation creates a local variable for the slice of bytes that it read from the `BytesMut`. Those bytes are "moved" i.e. the local function has ownership of them. But the `MessageRead` trait doesn't take ownership of the bytes. Instead, it decodes the message while referencing those bytes. If you use `Cow`, the implementation will retain this lifetime and point to the slice of bytes that you gave it. In order to make this work for us, we'd have to create a `Decoder` trait that can return items that point to initially read buffer (via GATs probably).

That makes sense. For my own learning, how did you figure that the implementation retained that lifetime and that that was causing the error? 

> I don't think this is worth doing at this stage. 

Do you mean it's not worth using Cow or migrating to quick-protobuf?

thomaseizinger:
> > I think we might have to give up on `Cow`. The issue is that our `Decoder` implementation creates a local variable for the slice of bytes that it read from the `BytesMut`. Those bytes are "moved" i.e. the local function has ownership of them. But the `MessageRead` trait doesn't take ownership of the bytes. Instead, it decodes the message while referencing those bytes. If you use `Cow`, the implementation will retain this lifetime and point to the slice of bytes that you gave it. In order to make this work for us, we'd have to create a `Decoder` trait that can return items that point to initially read buffer (via GATs probably).
> 
> That makes sense. For my own learning, how did you figure that the implementation retained that lifetime and that that was causing the error?

The thought-process was something like:

1. `MessageRead` has a lifetime in its definition on the type-level, not the function-level
2. The only reason to put a lifetime on a trait instead of one of its functions is if you want to use the same lifetime in the struct that implements the trait
3. I looked at some of the generated code, for example https://github.com/libp2p/rust-libp2p/blob/651b9c92aa841dc4d95a861164f6a5369ff771e2/protocols/gossipsub/src/generated/gossipsub/pb.rs#L100-L128
4. Here we can see that the lifetime `'a` is bound to the incoming bytes and also used in the struct definition, meaning the struct can only live as long as the reference to the passed bytes is valid.

> > I don't think this is worth doing at this stage.
> 
> Do you mean it's not worth using Cow or migrating to quick-protobuf?

I meant it is not worth using `Cow`! I certainly want to continue with `quick-protobuf` :)

thomaseizinger:
Based on the experiments [here](https://github.com/MarcoPolo/proto3-and-2-compat-experiments/pull/1) we definitely need to wait for https://github.com/tafia/quick-protobuf/pull/243 before we can ship this, otherwise any `proto3` code might not work.

kckeiks:
> Based on the experiments [here](https://github.com/MarcoPolo/proto3-and-2-compat-experiments/pull/1) we definitely need to wait for [tafia/quick-protobuf#243](https://github.com/tafia/quick-protobuf/pull/243) before we can ship this, otherwise any `proto3` code might not work.

https://github.com/MarcoPolo/proto3-and-2-compat-experiments/pull/1#issuecomment-1413101275 All good then?

thomaseizinger:
> > Based on the experiments [here](https://github.com/MarcoPolo/proto3-and-2-compat-experiments/pull/1) we definitely need to wait for [tafia/quick-protobuf#243](https://github.com/tafia/quick-protobuf/pull/243) before we can ship this, otherwise any `proto3` code might not work.
> 
> https://github.com/MarcoPolo/proto3-and-2-compat-experiments/pull/1#issuecomment-1413101275 All good then? 

All good yes, assuming https://github.com/tafia/quick-protobuf/pull/243 lands roughly in its current form, at least feature-wise.

thomaseizinger:
Are we using `proto3` anywhere at the moment? Because if not, then we can even merge this before we that `quick-protobuf` PR lands.

kckeiks:
> Are we using `proto3` anywhere at the moment? Because if not, then we can even merge this before we that `quick-protobuf` PR lands.

I think we're using it in `kad` and `noise`.

thomaseizinger:
> > Are we using `proto3` anywhere at the moment? Because if not, then we can even merge this before we that `quick-protobuf` PR lands.
> 
> I think we're using it in `kad` and `noise`.

Hmm okay. As far as I understand, `quick-protobuf` can handle proto3 correctly but it is just a bit awkward the way field presence is implemented (before https://github.com/tafia/quick-protobuf/pull/243). Let me know if you run into any issues.

kckeiks:
3 more crates to go

kckeiks:
@thomaseizinger This is ready for review now. `prost` has been removed from every crate.

thomaseizinger:
Amazing! I'll have a look later. I am guessing you saw that CI still has some failures?

kckeiks:
> Amazing! I'll have a look later. I am guessing you saw that CI still has some failures?

I think the CI is fixed now. I'll check back tomorrow.

kckeiks:
I'm not sure why`Interoperability Testing` is failing

> Error: Credentials could not be loaded, please check your action inputs: Could not load credentials from any providers

thomaseizinger:
> I'm not sure why `Interoperability Testing` is failing
> 
> > Error: Credentials could not be loaded, please check your action inputs: Could not load credentials from any providers

Interesting, definitely not your fault!

I'll review the PR nevertheless, I think the credentials issue should be sorted once we update to the latest version of our test-suite.

thomaseizinger:
Can you update to latest master please? That should fix the test!

core/src/identity.rs
```diff
@@ -162,35 +163,37 @@ impl Keypair {
             Self::Ecdsa(_) => return Err(DecodingError::encoding_unsupported("ECDSA")),
         };
 
-        Ok(pk.encode_to_vec())
+        let mut buf = Vec::with_capacity(pk.get_size());
+        let mut writer = Writer::new(&mut buf);
+        pk.write_message(&mut writer).expect("Encoding to succeed");
```
thomaseizinger, line 168: Can we use `serialize_into_vec` here?
 https://docs.rs/quick-protobuf/0.8.1/quick_protobuf/writer/fn.serialize_into_vec.html

thomaseizinger: And in other places?

misc/quick-protobuf-codec/CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+# 0.1.0 [unreleased]
+
+- Migrate from `prost` to `quick-protobuf`. This removes `protoc` dependency. See [PR 3312].
+
+[PR 3312]: https://github.com/libp2p/rust-libp2p/pull/3312
```
thomaseizinger, line 5: Can you delete the old changelog entries below?

.github/workflows/ci.yml
```diff
@@ -292,3 +292,30 @@ jobs:
...
+
+      - name: Glob match
+        uses: tj-actions/glob@v16
+        id: glob
+        with:
+          files: |
+            **/generated/*.proto
+
+      - name: Generate proto files
+        run: pb-rs -D ${{ steps.glob.outputs.paths }}
```
thomaseizinger: Can we use the long option name here please? That would be self-descriptive then.

thomaseizinger: An optimization idea for the future: We could use `Cow` in several places actually, just not all of them.

We would have to find a way how we can declare here, which `.proto` files should be generated with `-D` and which ones can omit it.

protocols/autonat/src/protocol.rs
```diff
@@ -184,29 +190,27 @@ pub enum ResponseError {
     InternalError,
 }
 
-impl From<ResponseError> for i32 {
+impl From<ResponseError> for proto::ResponseStatus {
```
thomaseizinger, line 193: Is `ResponseError` exposed in the public API? If not, we could omit it now for the generated type.

protocols/dcutr/CHANGELOG.md
```diff
@@ -11,10 +11,14 @@
...
 - Rename types in public API to follow naming conventions defined in [issue 2217]. See [PR 3214].
 
+- Migrate from `prost` to `quick-protobuf`. This removes `protoc` dependency. See [PR 3312].
+
 [PR 3213]: https://github.com/libp2p/rust-libp2p/pull/3213
 [PR 3153]: https://github.com/libp2p/rust-libp2p/pull/3153
 [issue 2217]: https://github.com/libp2p/rust-libp2p/issues/2217
 [PR 3214]: https://github.com/libp2p/rust-libp2p/pull/3214
+[PR 3312]: https://github.com/libp2p/rust-libp2p/pull/3312
+
```
thomaseizinger, line 21:
```suggestion
```

No need for this newline.

.github/workflows/ci.yml
```diff
@@ -292,3 +292,30 @@ jobs:
...
+
+  check-proto-files:
+    name: Check for changes in proto files
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+
+      - uses: Swatinem/rust-cache@359a70e43a0bb8a13953b04a90f76428b4959bb6 # v2.2.0
+
+      - run: cargo install pb-rs --locked
```
thomaseizinger: Can we pin the version here please to avoid new releases breaking our CI?

protocols/dcutr/src/protocol/inbound.rs
```diff
@@ -128,7 +125,7 @@ impl PendingConnect {
 #[derive(Debug, Error)]
 pub enum UpgradeError {
     #[error(transparent)]
-    Codec(#[from] prost_codec::Error),
+    Codec(#[from] quick_protobuf_codec::Error),
```
thomaseizinger, line 128: Is `UpgradeError` part of the public API? If yes, can you create a `CodecError` newtype please like we do in `libp2p-floodsub` to hide `quick-protobuf-codec` from the public API?

protocols/floodsub/CHANGELOG.md
```diff
@@ -6,7 +6,10 @@
 
 - Read and write protocols messages via `prost-codec`. See [PR 3224].
 
+- Migrate from `prost` to `quick-protobuf`. This removes `protoc` dependency. See [PR 3312].
+
 [pr 3224]: https://github.com/libp2p/rust-libp2p/pull/3224
+[PR 3312]: https://github.com/libp2p/rust-libp2p/pull/3312
```
thomaseizinger: As we can see here, we only migrated to `prost-codec` recently which hasn't been released yet.

Thus, the previous changelog entry is meaningless to our users.

Can you merge the two entries please?

thomaseizinger: The same may apply to other crates, please check all changelogs!

thomaseizinger (commented):
This looks really good, thanks for this work!

I've left a few minor comments but I am excited to merge this soon :)

core/src/identity.rs
kckeiks: I thought we agreed not to use `serialize_into_vec/slice`? https://github.com/libp2p/rust-libp2p/pull/3066#discussion_r1049193826

protocols/autonat/src/protocol.rs
kckeiks: Yes

protocols/dcutr/src/protocol/inbound.rs
kckeiks: Looks like it is https://github.com/libp2p/rust-libp2p/pull/3214/files#diff-1139510a64e6bc16c649cc904d0c3c799663ee17f7828f2dea6839cfb3f5c283R39.

I see that there are several places where this pattern has been implemented. I suggest we split this into a separate task because it's a chunk of work and I fear I'm gonna have limited bandwidth soon.

kckeiks:
Ready for another round of review.

kckeiks:
Not sure why that Interoperability Testing test is failing 🤔

thomaseizinger:
> Not sure why that Interoperability Testing test is failing thinking

That is likely a flake. We are still waiting for https://github.com/webrtc-rs/webrtc/pull/405 to be released.

core/src/identity.rs
thomaseizinger: Ah right, I forgot that this writes the length as well.

protocols/gossipsub/src/behaviour.rs
```diff
@@ -178,8 +178,8 @@ impl From<MessageAuthenticity> for PublishConfig {
                 let public_key = keypair.public();
                 let key_enc = public_key.to_protobuf_encoding();
                 let key = if key_enc.len() <= 42 {
-                    // The public key can be inlined in [`rpc_proto::Message::from`], so we don't include it
-                    // specifically in the [`rpc_proto::Message::key`] field.
+                    // The public key can be inlined in [`rpc_proto::proto::::Message::from`], so we don't include it
+                    // specifically in the [`rpc_proto::proto::Message::key`] field.
```
thomaseizinger, line 183:
```suggestion
                    // The public key can be inlined in [`proto::Message::from`], so we don't include it
                    // specifically in the [`proto::Message::key`] field.
```

thomaseizinger (dismissed):
Nice work!

I think this is ready to merge. I'll give @mxinden a few days to have a look too!

thomaseizinger:
One more thing: Can you remove the installation of protoc from all CI steps and also from the Dockerfile in `interop-tests/Dockerfile`?

thomaseizinger:
I opened https://github.com/libp2p/github-mgmt/pull/128, it needs to be merged just before we decide to merge this one to update the required checks for GitHub.

core/src/identity.rs
```diff
@@ -162,35 +163,37 @@ impl Keypair {
...
     }
 }
 
-impl zeroize::Zeroize for keys_proto::PrivateKey {
+impl zeroize::Zeroize for proto::PrivateKey {
     fn zeroize(&mut self) {
-        self.r#type.zeroize();
-        self.data.zeroize();
+        self.Type = proto::KeyType::default();
+        self.Data = Vec::default();
```
mxinden: I don't think this properly zeroizes the underlying memory region, right? Can you not call `zeroize` on `Type` and `Data`?

.github/workflows/ci.yml
```diff
@@ -292,3 +292,30 @@ jobs:
...
+
+      - name: Glob match
+        uses: tj-actions/glob@v16
+        id: glob
+        with:
+          files: |
+            **/generated/*.proto
+
+      - name: Generate proto files
+        run: pb-rs --dont_use_cow ${{ steps.glob.outputs.paths }}
```
mxinden, line 315: What is the reasoning for not using `Cow`? I am assuming code simplicity? I find it a neat feature, though I would assume that the performance gain is not worth the additional complexity in most cases. Can we easily use `Cow` on a case-by-case basis in the future?

protocols/autonat/src/protocol.rs
```diff
@@ -269,35 +270,38 @@ impl DialResponse {
...
+            dialResponse: Some(dial_response),
         };
 
-        let mut bytes = Vec::with_capacity(msg.encoded_len());
-        msg.encode(&mut bytes)
-            .expect("Vec<u8> provides capacity as needed");
-        bytes
+        let mut buf = Vec::with_capacity(msg.get_size());
+        let mut writer = Writer::new(&mut buf);
+        msg.write_message(&mut writer).expect("Encoding to succeed");
```
mxinden, line 296: Nit pick, I would prefer the wording `.expect("Vec<u8> provides capacity as needed")`. Yes, we expect the encoding to succeed, but why do we do so? Because we only expect errors on memory allocation and given that we are using a `Vec` which can dynamically allocate, this is not an issue. Or am I misunderstanding the reasoning?

mxinden (commented):
Thanks for all the work here. 3 questions. Otherwise good to go from my end.

.github/workflows/ci.yml
thomaseizinger: The problem / feature is that when using `Cow`, the deserialised data structure retains a reference to the byte buffer.

This doesn't play well with `asynchronous-codec` where `Encoder` passes you a byte buffer and expects you to create an owned type, without references to the original buffer. 

I think `Encoder` would have to undergo some serious refactoring to make this possible. Perhaps now that GATs are stable, this is actually a viable interface. For the time being, we opted to not go down this route.

protocols/autonat/src/protocol.rs
thomaseizinger: You are completely on point. It think it would be useful to upstream the following to `quick-protobuf`:


```rust
/// A trait to handle deserialization based on parsed `Field`s
pub trait MessageWrite: Sized {

	// Other functions as currently defined ...

	fn write_to_vec(&self) -> Vec<u8> {
		let mut vec = Vec::with_capacity(self.get_size());
		let mut writer = Writer::new(&mut vec);

		self.write_message(&mut writer).expect("Vec<u8> provides enough capacity");

		vec
	}
}
```

On this note, I think there is some clean-up to be done in terms of APIs. It is quite confusing that there is `MessageWrite::write_message` which writes to a `Writer` and there is also `Writer::write_message` which writes the message using `MessageWriter::write_message` **with a length-prefix**!

thomaseizinger: Opened an issue here: https://github.com/tafia/quick-protobuf/issues/246

core/src/identity.rs
kckeiks: `self.r#type` is i32 but `self.Type` is an enum and it doesn't satisfy the `DefaultIsZeroes` bound necessary to be zeroizable. I'll update `Data` though.

protocols/autonat/src/protocol.rs
kckeiks: > Nit pick, I would prefer the wording .expect("Vec<u8> provides capacity as needed"). Yes, we expect the encoding to succeed, but why do we do so? Because we only expect errors on memory allocation and given that we are using a Vec which can dynamically allocate, this is not an issue. Or am I misunderstanding the reasoning?

It wasn't clear from the documentation and code in what cases we should expect errors https://github.com/libp2p/rust-libp2p/pull/3066#discussion_r1006239290 so that is why the message is so vague. Do we want this changed in this PR or wait for changes upstream?

kckeiks:
> One more thing: Can you remove the installation of protoc from all CI steps and also from the Dockerfile in `interop-tests/Dockerfile`?

Done. I'll check tomorrow to see if I broke something in the CI.

thomaseizinger:
> > One more thing: Can you remove the installation of protoc from all CI steps and also from the Dockerfile in `interop-tests/Dockerfile`?
> 
> Done. I'll check tomorrow to see if I broke something in the CI.

It seems we still need `protoc` for our semver checking because the baseline version still depends on `protoc` so we will have to revert 66a2af5 unfortunately.

protocols/autonat/src/protocol.rs
mxinden: I am fine merging as is.

thomaseizinger:
Damn, it seems the newtypes don't help in making this a non-breaking change.

Can we look into why the error type of `quick-protobuf` is not `UnwindSafe`?

kckeiks:
> Damn, it seems the newtypes don't help in making this a non-breaking change.
> 
> Can we look into why the error type of `quick-protobuf` is not `UnwindSafe`?

I did not see this failure yesterday when tests were passing, is this test new?

kckeiks:
We have this similar code in plaintext but there is no failure for it
```
#[derive(Debug)]
pub struct DecodeError(quick_protobuf::Error);
```
The places where we're getting errors have `#[error(transparent)]`.

thomaseizinger:
> > Damn, it seems the newtypes don't help in making this a non-breaking change.
> > 
> > Can we look into why the error type of `quick-protobuf` is not `UnwindSafe`?
> 
> I did not see this failure yesterday when tests were passing, is this test new?

It is because we released a new major version yesterday. Thus, until yesterday, breaking changes were fine. Today, they are flagged as required.

thomaseizinger:
> We have this similar code in plaintext but there is no failure for it
> ```
> #[derive(Debug)]
> pub struct DecodeError(quick_protobuf::Error);
> ```
> The places where we're getting errors have `#[error(transparent)]`.

The culprit seems to be that `std::io::Error` is not `UnwindSafe`: https://doc.rust-lang.org/nightly/std/io/struct.Error.html#synthetic-implementations

This is quite unfortunate. `prost::Error` was `UnwindSafe` so this is now a breaking change. I am tempted to say I don't care about this breaking change but as of today, I don't think we have a capability to selectively ignore something like that.

cc @obi1kenobi Any input on this?

.github/workflows/ci.yml
thomaseizinger: > What is the reasoning for not using `Cow`? I am assuming code simplicity? I find it a neat feature, though I would assume that the performance gain is not worth the additional complexity in most cases. Can we easily use `Cow` on a case-by-case basis in the future?

We would be able to use `Cow` on a case-by-case basis if we find a way to run this CI check on all proto files, without having to enumerate them all so we don't forget adding new ones but also have a list of files that we want to enable `Cow` for.

kckeiks:
> > We have this similar code in plaintext but there is no failure for it
> > ```
> > #[derive(Debug)]
> > pub struct DecodeError(quick_protobuf::Error);
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The places where we're getting errors have `#[error(transparent)]`.
> 
> The culprit seems to be that `std::io::Error` is not `UnwindSafe`: https://doc.rust-lang.org/nightly/std/io/struct.Error.html#synthetic-implementations
> 
> This is quite unfortunate. `prost::Error` was `UnwindSafe` so this is now a breaking change. I am tempted to say I don't care about this breaking change but as of today, I don't think we have a capability to selectively ignore something like that.
> 
> cc @obi1kenobi Any input on this?

Yeah I don't see any clean way of getting by this. We could define another type to replace `quick_protobuf::Error`,

```
pub struct DecodeError(QuickProtobufError);
```

```
enum QuickProtobufError {
    Io(String),
....
}
``` 

```
impl From<quick_protobug::Error> for QuickProtobufError {
....
}
```

and open an issue to change this once the issue with `UnwindSafe` is solved.

thomaseizinger:
Can we box it as `Box<dyn Error>`?

kckeiks:
> Can we box it as `Box<dyn Error>`?

Trying it out now but I just noticed this https://doc.rust-lang.org/std/boxed/struct.Box.html#impl-UnwindSafe-for-Box%3CT%2C%20A%3E.

obi1kenobi:
TL;DR: I don't see a way out of the major bump here, unless you want to break semver. `clap` also got bitten by `UnwindSafe` [a while ago](https://github.com/clap-rs/clap/issues/3876). If you do decide to break semver on this, `cargo-semver-checks` will stop complaining about this as soon as you release the next version, since the baseline will move forward. There's a way to tell `cargo-semver-checks` to not skip checks, I'll answer that in the linked issue since it'll be more discoverable for future users.

This one of the reasons auto-traits are tricky: they silently become part of your API and then you are stuck with them. They propagate across newtypes and can even cause editing a completely private bit of code to be semver-major: https://predr.ag/blog/toward-fearless-cargo-update/#breaking-semver-with-auto-traits

> Can we box it as `Box<dyn Error>`?

Not exactly sure what you had in mind. Boxing it won't make the type `UnwindSafe` so you won't be able to avoid the breaking change here. But `dyn Error` isn't `UnwindSafe` even if the non-erased underlying error was `UnwindSafe`, so that can prevent the type from ever becoming `UnwindSafe` again going forward.

Here's an example showing that `Box<dyn Error>` isn't `UnwindSafe`: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=e3ec97f9694f1f2fc1a57a6269434e5d

***

Tangent: This is a good example of a new lint category I'd love to implement in `cargo-semver-checks`. The name needs bikeshedding, I've been calling it "major-if-reverted." These are `warn`-level lints for changes that are difficult to undo:
- removing `#[non_exhaustive]` from a struct, enum, or variant
- existing type starts implementing an auto-trait ("did you really mean to make it `Send / Sync / UnwindSafe`?")
- adding a new pub type without making it `#[non_exhaustive]`
- etc.

I'd love to make `cargo-semver-checks` into a holistic API evolution tool, not just a "did I do semver right" linter.

core/src/lib.rs
```diff
@@ -75,6 +74,44 @@ pub use translation::address_translation;
 pub use transport::Transport;
 pub use upgrade::{InboundUpgrade, OutboundUpgrade, ProtocolName, UpgradeError, UpgradeInfo};
 
-#[derive(thiserror::Error, Debug)]
-#[error(transparent)]
-pub struct DecodeError(Box<dyn Error + Send + Sync + 'static>);
```
thomaseizinger: I am inclined to leave this as, accept that we are removing the `UnwindSafe` auto trait and just do another major version bump. I don't think complicating the API with an additional enum is worth it at this stage.

@mxinden What do you think?

thomaseizinger: Actually, lets do the following:

- Make the enum private and wrap it in a new type that doesn't expose anything.
- Record an issue to replace the enum with `quick_protobuf::Error`
- Track the issue in the `0.52` milestone
- Optional: Prepare a PR that fixes the above issue by removing the enum again and add that to `0.52` as well.

That way, we can merge this as a non-breaking change and merge in the breaking change as soon as we decide to do another breaking change. I think very likely, nobody actually depends on this error being `UnwindSafe`. I don't actually care about removing this auto trait but I do care about not making as many major version bumps.

kckeiks: > Make the enum private and wrap it in a new type that doesn't expose anything.

Since the new type doesn't expose anything, I removed the enum.

I created https://github.com/libp2p/rust-libp2p/issues/3534 but can't add issues to milestones.

thomaseizinger: Thank you, I've added it to the milestone!

thomaseizinger:
@mxinden This is ready to merge from my end. You will need to merge https://github.com/libp2p/github-mgmt/pull/128 before you can merge here to update the required jobs.

mxinden (dismissed):

thomaseizinger:
I think some of the changelog diffs have been applied to the wrong version as of the last release now.

mxinden:
Good catch. @thomaseizinger can you send a pull request in the next couple of days?

mxinden:
Thanks @kckeiks and @thomaseizinger for all the work here!

thomaseizinger:
> Good catch. @thomaseizinger can you send a pull request in the next couple of days?

My mind is engineering a `semverlog` tool as we speak to avoid this.

I can send a PR although it might take a while, have a busy weekend ahead and flying to Europe on Tuesday morning.

thomaseizinger:
Sent it here: https://github.com/libp2p/rust-libp2p/pull/3541

PR #3322, opened 2023-01-13
fix(ci): revert "Trigger Quic as Transport wakeup on dial"

thomaseizinger:
### Description

This patch broke CI and made the job hang forever. That did not get caught during the review process because the job is not yet required for merging into `master`.


elenaf9 (approved):
Sorry that I didn't catch this in review!

PR #3325, opened 2023-01-15
feat(metrics)!: expose identify metrics for connected peers only

mxinden:
## Description

Previously we would increase a counter / gauge / histogram on each received identify information. These metrics are missleading, as e.g. they depend on the identify interval and don't represent the set of currently connected peers.

With this commit, identify information is tracked for the currently connected peers only. Instead of an increase on each received identify information, metrics represent the status quo (Gauge).

Example:

```
\# HELP libp2p_libp2p_identify_remote_protocols Number of connected nodes supporting a specific protocol, with "unrecognized" for each peer supporting one or more unrecognized protocols...
\# TYPE libp2p_libp2p_identify_remote_protocols gauge
libp2p_libp2p_identify_remote_protocols_total{protocol="/ipfs/id/push/1.0.0"} 1
libp2p_libp2p_identify_remote_protocols_total{protocol="/ipfs/id/1.0.0"} 1
libp2p_libp2p_identify_remote_protocols_total{protocol="/ipfs/ping/1.0.0"} 1
libp2p_libp2p_identify_remote_protocols_total{protocol="unrecognized"} 1
\# HELP libp2p_libp2p_identify_remote_listen_addresses Number of connected nodes advertising a specific listen address...
\# TYPE libp2p_libp2p_identify_remote_listen_addresses gauge
libp2p_libp2p_identify_remote_listen_addresses_total{listen_address="/ip4/tcp"} 1
libp2p_libp2p_identify_remote_listen_addresses_total{listen_address="/ip4/udp/quic"} 1
\# HELP libp2p_libp2p_identify_local_observed_addresses Number of connected nodes observing the local node at a specific address...
\# TYPE libp2p_libp2p_identify_local_observed_addresses gauge
libp2p_libp2p_identify_local_observed_addresses_total{observed_address="/ip4/tcp"} 1
```

## Notes

Depends on new release of https://github.com/prometheus/client_rust that includes https://github.com/prometheus/client_rust/pull/82.

## Change checklist


thomaseizinger:
> Depends on new release of [prometheus/client_rust](https://github.com/prometheus/client_rust) that includes [prometheus/client_rust#82](https://github.com/prometheus/client_rust/pull/82).

Setting to draft until ready for review.

mxinden:
@thomaseizinger can you give this pull request a review? It builds on top of https://github.com/prometheus/client_rust/pull/82. https://github.com/prometheus/client_rust/pull/82 has not yet been released. Would be good to get some input on a usage of https://github.com/prometheus/client_rust/pull/82 before I release https://github.com/prometheus/client_rust/pull/82.

thomaseizinger:
Looks fine to me. I am not fond of the name "collector". What about "static" or "const" metrics?

mxinden:
> Looks fine to me. I am not fond of the name "collector". What about "static" or "const" metrics?

The name is used across all Prometheus client libraries and well documented. I agree that it is not intuitive. That said, I think staying consistent is the better approach.

misc/metrics/src/identify.rs
```diff
@@ -20,39 +20,79 @@
...
+static PROTOCOLS_DESCRIPTOR: Lazy<Descriptor> = Lazy::new(|| {
+    Descriptor::new(
+        "remote_protocols",
+        "Number of connected nodes supporting a specific protocol, with \
+         \"unrecognized\" for each peer supporting one or more unrecognized \
+         protocols",
+        None,
+        None,
+        vec![],
+    )
```
thomaseizinger: Just an idea, could we make `new` a `const fn`?

misc/metrics/src/identify.rs
```diff
@@ -213,55 +175,104 @@ struct AddressLabels {
...
-                        let count = acc.entry(protocol.to_string()).or_default();
-                        *count += 1;
-                    }
-                    acc
-                },
-            );
+impl Collector for Peers {
+    fn collect<'a>(
+        &'a self,
+    ) -> Box<dyn Iterator<Item = (Cow<'a, Descriptor>, MaybeOwned<'a, Box<dyn LocalMetric>>)> + 'a>
```
thomaseizinger, line 194: When do we finally get RPITIT?

misc/metrics/src/identify.rs
```diff
@@ -213,55 +175,104 @@ struct AddressLabels {
...
+                    .map(|p| {
+                        if ALLOWED_PROTOCOLS.contains(&p.as_bytes()) {
+                            p.to_string()
+                        } else {
+                            "unrecognized".to_string()
+                        }
+                    })
+                    .collect();
+                protocols.sort();
+                protocols.dedup();
```
thomaseizinger: Could we extract this into a helper and thus make the loop a bit easier to read?

misc/metrics/src/identify.rs
```diff
@@ -213,55 +175,104 @@ struct AddressLabels {
...
+                *count += 1;
+            }
         }
 
-        Ok(())
-    }
+        let count_by_protocols: Box<dyn LocalMetric> =
+            Box::new(RefCell::new(count_by_protocols.into_iter().map(
+                |(protocol, count)| ([("protocol", protocol)], ConstGauge::new(count)),
+            )));
```
thomaseizinger: See https://github.com/prometheus/client_rust/pull/82/files#r1158205762. I've found it very hard to discover what happens here.

mxinden: This pull request is now updated to the proposal in https://github.com/prometheus/client_rust/pull/138.

misc/metrics/src/identify.rs
mxinden: Oh, I have not thought about that. Only blocker documented in https://github.com/prometheus/client_rust/issues/139.

mxinden:
Note that this pull request now updates `libp2p-metrics`, `libp2p-gossipsub` and `metrics-example` to `prometheus-client` `v0.21.0`.

mxinden:
This pull request is ready to merge from my end. @thomaseizinger would you mind taking another look?

misc/metrics/CHANGELOG.md
```diff
@@ -1,9 +1,29 @@
 ## 0.13.0 - unreleased
 
+- Previously `libp2p-metrics::identify` would increase a counter / gauge / histogram on each
+  received identify information. These metrics are missleading, as e.g. they depend on the identify
```
thomaseizinger:
```suggestion
  received identify information. These metrics are misleading, as e.g. they depend on the identify
```

misc/metrics/src/identify.rs
```diff
@@ -21,39 +21,79 @@
...
+use prometheus_client::MaybeOwned;
+use std::borrow::Cow;
 use std::collections::HashMap;
-use std::iter;
 use std::sync::{Arc, Mutex};
 
+static PROTOCOLS_DESCRIPTOR: Lazy<Descriptor> = Lazy::new(|| {
+    Descriptor::new(
+        "remote_protocols",
+        "Number of connected nodes supporting a specific protocol, with \
```
thomaseizinger: Use an `r#""#` string to avoid the escaping?

misc/metrics/src/identify.rs
```diff
@@ -21,39 +21,79 @@
...
-        let protocols = Protocols::default();
-        sub_registry.register(
-            "protocols",
-            "Number of connected nodes supporting a specific protocol, with \
-             \"unrecognized\" for each peer supporting one or more unrecognized \
-             protocols",
-            protocols.clone(),
-        );
+        let peers = Peers::default();
+        sub_registry.register_collector(Box::new(peers.clone()));
```
thomaseizinger, line 95: I think initializing this `Metrics` struct would almost be cleaner if we would just say `Metrics::default()` and then have a separate `&self` method that takes the registry and registers each metric.

thomaseizinger: Idea for your `prometheus-client` crate: Implementing a constructor that does this could be implemented as a derive macro that goes through all fields of the struct and registers them by:

- using the field name as a metric name
- using the rustdoc string as the metric help
- collectors would have to be tagged with an attribute to allow the macro to call the different function

thomaseizinger (approved):
LGTM.

misc/metrics/src/identify.rs
mxinden: > I think initializing this `Metrics` struct would almost be cleaner if we would just say `Metrics::default()` and then have a separate `&self` method that takes the registry and registers each metric.

Ideally I would like to prevent at the type level that someone creates a `Metrics` struct but then does not register it with a `Registry`.


> Idea for your `prometheus-client` crate:

Good idea! Tracked on https://github.com/prometheus/client_rust/issues/140.

PR #3318, opened 2023-01-11
fix(ci): properly escape PR title

thomaseizinger:
## Description

Within double quoted strings, bash tries to evaluate everything within backticks as a command. The GitHub security guide recommends to use an intermediary environment variable instead: https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-an-intermediate-environment-variable

See https://github.com/libp2p/rust-libp2p/actions/runs/3889880383/jobs/6638520274#step:3:11.

## Change checklist

- [ ] I have performed a self-review of my own code
- [ ] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] A changelog entry has been made in the appropriate crates

PR #3327, opened 2023-01-16
feat(core)!: make `ConnectionId`s globally unique

thomaseizinger:
## Description

Instead of offering a public constructor, users are now no longer able to construct `ConnectionId`s at all. They only public API exposed are the derived traits. Internally, `ConnectionId`s are monotonically incremented using a static atomic counter, thus no two connections will ever get assigned the same ID.

## Notes

Extracted out of https://github.com/libp2p/rust-libp2p/pull/3254.

## Change checklist


core/src/connection.rs
```diff
@@ -30,16 +33,14 @@ impl ConnectionId {
     /// This is primarily useful for creating connection IDs
     /// in test environments. There is in general no guarantee
     /// that all connection IDs are based on non-negative integers.
+    #[deprecated(since = "0.39.0", note = "Use `ConnectionId::next` instead.")]
     pub fn new(id: usize) -> Self {
```
mxinden: Are we aware of any users of `ConnectionId::new` outside of this repository?

Providing both `new` and `next` removes the guarantee of `next`, namely to create unique `ConnectionId`s.

In case there are no users, how about removing `new` in favor of `next` directly?

core/src/connection.rs
```diff
@@ -19,6 +19,9 @@
 // DEALINGS IN THE SOFTWARE.
 
 use crate::multiaddr::{Multiaddr, Protocol};
+use std::sync::atomic::{AtomicUsize, Ordering};
+
+static NEXT_CONNECTION_ID: AtomicUsize = AtomicUsize::new(0);
 
 /// Connection identifier.
 #[derive(Debug, Copy, Clone, Hash, PartialEq, Eq, PartialOrd, Ord)]
```
mxinden:
```suggestion
#[derive(Debug, Copy, Clone, Hash, PartialEq, Eq)]
```

I suggest the only guarantee that `ConnectionId` provides is uniqueness, not monotonic increasing. Thus there is no need for ordering. Is any existing functionality depending on this? If not, how about removing it right away?

core/src/connection.rs
thomaseizinger: I was going to deprecate it first because it doesn't matter until we merge https://github.com/libp2p/rust-libp2p/pull/3328.

core/src/connection.rs
thomaseizinger: Yeah happy to remove it.

thomaseizinger:
Converting to draft until https://github.com/libp2p/rust-libp2p/pull/3318 is merged to prevent accidental merges.

core/src/connection.rs
thomaseizinger: Let's merge https://github.com/libp2p/rust-libp2p/pull/3221 first anyway.

thomaseizinger:
This is ready for review despite being draft. Only #3318 needs to merge first.

swarm/src/connection.rs
```diff
@@ -45,30 +45,30 @@ use libp2p_core::upgrade::{InboundUpgradeApply, OutboundUpgradeApply};
...
-    type Output = Self;
-
-    fn add(self, other: usize) -> Self {
-        Self(self.0 + other)
+    /// Really, you should not use this, not even for testing but it is here if you need it.
+    #[deprecated(
+        since = "0.42.0",
+        note = "Don't use this, it will be removed at a later stage again."
+    )]
+    pub const DUMMY: ConnectionId = ConnectionId(0);
```
thomaseizinger, line 67: I am not very happy about this but I propose to accept this technical debt to be able to move forward with the connection management PR.

Long-term, we would need to re-write the gossipsub and kademlia tests to not manually call lifecycle functions and thus only have `Swarm` construct `ConnectionId`s.

thomaseizinger: @libp2p/rust-libp2p-maintainers Please voice your opinion on this.

swarm/src/connection.rs
```diff
@@ -45,30 +45,30 @@ use libp2p_core::upgrade::{InboundUpgradeApply, OutboundUpgradeApply};
 use libp2p_core::{upgrade, UpgradeError};
 use libp2p_core::{Endpoint, PeerId};
 use std::future::Future;
+use std::sync::atomic::{AtomicUsize, Ordering};
 use std::task::Waker;
 use std::time::Duration;
 use std::{fmt, io, mem, pin::Pin, task::Context, task::Poll};
 
+static NEXT_CONNECTION_ID: AtomicUsize = AtomicUsize::new(1);
```
thomaseizinger, line 53: We start IDs at `1` so they can never clash with the `DUMMY` one below.

swarm/src/connection.rs
jxs: Yeah it would be nice if [this ](https://users.rust-lang.org/t/pub-only-for-tests/28153) was possible. Other than that I don't see a better alternative for here.

swarm/src/connection.rs
```diff
@@ -45,30 +45,30 @@ use libp2p_core::upgrade::{InboundUpgradeApply, OutboundUpgradeApply};
...
+    /// Really, you should not use this, not even for testing but it is here if you need it.
+    #[deprecated(
+        since = "0.42.0",
+        note = "Don't use this, it will be removed at a later stage again."
+    )]
+    pub const DUMMY: ConnectionId = ConnectionId(0);
+
+    /// Returns the next available [`ConnectionId`].
+    pub(crate) fn next() -> Self {
+        Self(NEXT_CONNECTION_ID.fetch_add(1, Ordering::SeqCst))
```
jxs, line 71: this is very elegant.

jxs (approved):
LGTM

swarm/src/connection.rs
thomaseizinger: Credits go to @rkuhn !

thomaseizinger:
Thanks for the review @jxs! @mxinden I'd like you to have a look at this too before I merge it.

mxinden (approved):
I am fine with the technical dept. I suggest moving forward here.

mxinden:
rust-libp2p-head x rust-libp2p-head is failing. Error seems to be:

```
   rust-libp2p-headxrust-libp2p-headwebrtc-dialer-1    | [2023-01-23T12:07:47Z WARN  webrtc::peer_connection::peer_connection_internal] failed to open data channel: there already exists a stream with identifier
  rust-libp2p-headxrust-libp2p-headwebrtc-listener-1  | Error: Failed to wait for dialer conclusion
```

thomaseizinger:
> rust-libp2p-head x rust-libp2p-head is failing. Error seems to be:
> 
> ```
>    rust-libp2p-headxrust-libp2p-headwebrtc-dialer-1    | [2023-01-23T12:07:47Z WARN  webrtc::peer_connection::peer_connection_internal] failed to open data channel: there already exists a stream with identifier
>   rust-libp2p-headxrust-libp2p-headwebrtc-listener-1  | Error: Failed to wait for dialer conclusion
> ```

Isn't that the flake that came up a couple of times now?

jxs:
> > rust-libp2p-head x rust-libp2p-head is failing. Error seems to be:
> > ```
> >    rust-libp2p-headxrust-libp2p-headwebrtc-dialer-1    | [2023-01-23T12:07:47Z WARN  webrtc::peer_connection::peer_connection_internal] failed to open data channel: there already exists a stream with identifier
> >   rust-libp2p-headxrust-libp2p-headwebrtc-listener-1  | Error: Failed to wait for dialer conclusion
> > ```
> 
> Isn't that the flake that came up a couple of times now?

yup exactly

thomaseizinger:
I am tracking removal of the tech-debt in https://github.com/libp2p/rust-libp2p/issues/3371.

melekes:
A bit late to the party, but we (@paritytech) rely on being able to construct `ConnectionId` in our tests - https://github.com/paritytech/substrate/blob/master/client/network/src/protocol/notifications/behaviour.rs#L3391

Would you be willing to change your opinion on making `ConnectionId` constructor private?

thomaseizinger:
> A bit late to the party, but we (@paritytech) rely on being able to construct `ConnectionId` in our tests - [paritytech/substrate@`master`/client/network/src/protocol/notifications/behaviour.rs#L3391](https://github.com/paritytech/substrate/blob/master/client/network/src/protocol/notifications/behaviour.rs?rgh-link-date=2023-03-15T11%3A56%3A13Z#L3391)
> 
> Would you be willing to change your opinion on making `ConnectionId` constructor private?

We do too in some tests but my current position is that you shouldn't call those lifecycle functions yourself but instead construct two or more `Swarm`s and establish a connection between them, possible using the new `libp2p-swarm-test`: https://github.com/libp2p/rust-libp2p/tree/master/swarm-test

Does that work for you?

@mxinden I think we should publish `libp2p-swarm-test` to crates.io.

melekes:
> @mxinden I think we should publish libp2p-swarm-test to crates.io.

please do 🙏 

> We do too in some tests but my current position is that you shouldn't call those lifecycle functions yourself but instead construct two or more Swarms and establish a connection between them, possible using the new libp2p-swarm-test: https://github.com/libp2p/rust-libp2p/tree/master/swarm-test
> Does that work for you?

There are over 2000 lines of tests, but I think it may be possible to rewrite them using `Swarm`s.

melekes:
In some tests we need to inject custom handshake and I'm worried it won't be possible with `libp2p-swarm-test` API

```
notif.on_connection_handler_event(
			peer,
			conn1,
			conn_yielder.open_substream(peer, 0, connected.clone(), vec![1, 2, 3, 4]),
		);
```

Overall this change is quite upsetting because we just want to test our state machine built on top of libp2p so spawning libp2p Swarms under it is overcomplicating the issue

thomaseizinger:
> Overall this change is quite upsetting because we just want to test our state machine built on top of libp2p so spawning libp2p Swarms under it is overcomplicating the issue

Thank you for sharing.

We can re-introduce a public constructor for `ConnectionId`.

However, I think these whitebox tests are not a good idea. For example, recently we introduced callbacks for connection management that also involves the behaviour upon pending connections. Is your home-grown test framework calling these correctly?

Going through `Swarm` for your tests is going to simulate a real network much better than calling these lifecycle functions yourself.

melekes:
> Is your home-grown test framework calling these correctly?

Probably not, which is why I agree that we should use `Swarm`s as much as possible. BUT at the same time 

a) we want to ship 0.51.1 libp2p, which is currently blocked on this 
b) rewriting 2000 lines of tests will take some time 
c) libp2p-swarm-test is not even published.

I'd humbly propose to reintroduce `ConnectionId::new` and mark it as deprecated (it was not marked as deprecated before?). This will give us some time to rewrite the tests. Thank you 🙏

thomaseizinger:
> c) libp2p-swarm-test is not even published.

We are working on that:

- https://github.com/libp2p/rust-libp2p/pull/3650
- https://github.com/libp2p/rust-libp2p/pull/3629

> I'd humbly propose to reintroduce `ConnectionId::new` and mark it as deprecated (it was not marked as deprecated before?). This will give us some time to rewrite the tests. Thank you pray

Yep, we can do that. I'll send a PR.

thomaseizinger:
> > I'd humbly propose to reintroduce `ConnectionId::new` and mark it as deprecated (it was not marked as deprecated before?). This will give us some time to rewrite the tests. Thank you pray
> 
> Yep, we can do that. I'll send a PR.

Actually, with how we have now modeled `DialOpts`, there isn't any risk in users creating an already used `ConnectionId` as part of dialing a peer.

PR #3328, opened 2023-01-16
refactor(swarm)!: remove `handler` from `NetworkBehaviourAction::Dial`

thomaseizinger:
## Description

We create the `ConnectionId` for the new connection as part of `DialOpts`. This allows `NetworkBehaviour`s to accurately track state regarding their own dial attempts.

This patch is the main enabler of https://github.com/libp2p/rust-libp2p/pull/3254. Removing the `handler` field will allow us to deprecate the `NetworkBehaviour::new_handler` function in favor of four new ones that give more control over the connection lifecycle.

## Notes

- Extracted out of https://github.com/libp2p/rust-libp2p/pull/3254.
- Depends-On: https://github.com/libp2p/rust-libp2p/pull/3327.
- Depends-On: https://github.com/libp2p/rust-libp2p/pull/3335
- Depends-On: https://github.com/libp2p/rust-libp2p/pull/3372

Please review this design critically. I believe it has value on its own but at the end of the day, we only do this for https://github.com/libp2p/rust-libp2p/pull/3254.

## Change checklist


swarm/src/behaviour.rs
```diff
@@ -248,177 +233,13 @@ pub enum NetworkBehaviourAction<
...
-    /// # #[derive(Debug, PartialEq, Eq)]
-    /// # struct PreciousMessage(String);
-    /// ```
-    Dial { opts: DialOpts, handler: THandler },
+    /// The provided [`ConnectionId`] will be used throughout the connection's lifecycle to associate
+    /// events with it. This allows a [`NetworkBehaviour`] to identify a connection that resulted out
+    /// of its own dial request.
+    Dial {
+        opts: DialOpts,
+        connection_id: ConnectionId,
```
mxinden: Why not make `connection_id` part of `DialOpts`? For those that don't need the `connection_id`, they don't need to bother creating one, as the `DialOpts` builder would create one for them.

Do you favor being explicit, i.e. requiring the user to create the `ConnectionId`, thus being aware of it?

thomaseizinger: We could move it into `DialOpts` I guess, that would make things slightly more ergonomic. Together with https://github.com/libp2p/rust-libp2p/pull/3221, we could then make the constructor `pub(crate)` and thus prevent that users assign existing connection IDs to new connections.

thomaseizinger:
Converting to draft until https://github.com/libp2p/rust-libp2p/pull/3327 is merged to prevent accidental merges.

swarm/src/behaviour.rs
thomaseizinger: Addressed. I also created https://github.com/libp2p/rust-libp2p/pull/3335 which I temporarily merged into this PR to keep going here.

thomaseizinger:
This is ready for review, just keep in mind that it also contains https://github.com/libp2p/rust-libp2p/pull/3335.

PR description is up-to-date.

protocols/dcutr/src/handler.rs
```diff
@@ -20,61 +20,27 @@
...
-                assert!(
-                    !endpoint.is_relayed(),
-                    "`Prototype::DirectConnection` is never created for relayed connection."
-                );
-                Either::Right(Either::Left(direct::Handler::new(relayed_connection_id)))
-            }
+        if endpoint.is_relayed() {
+            Either::Left(relayed::Handler::new(endpoint.clone()))
+        } else {
+            Either::Right(direct::Handler::default()) // This is a direct connection. What we don't know is whether it is the one we created or another one that happened accidentally.
```
thomaseizinger, line 39: This is a bit of a hack that will go away once https://github.com/libp2p/rust-libp2p/pull/3254 merges. That will allow us to create the correct handler directly in the behaviour where we have the knowledge, which dials we initiated ourselves.

protocols/dcutr/src/behaviour_impl.rs
```diff
@@ -215,24 +239,34 @@ impl NetworkBehaviour for Behaviour {
...
-        handler_event: <<Self::ConnectionHandler as IntoConnectionHandler>::Handler as
-            ConnectionHandler>::OutEvent,
+        connection_id: ConnectionId,
+        handler_event: <<Self::ConnectionHandler as IntoConnectionHandler>::Handler as ConnectionHandler>::OutEvent,
     ) {
+        let relayed_connection_id = match handler_event.as_ref() {
+            Either::Left(_) => connection_id,
+            Either::Right(_) => match self.direct_to_relayed_connections.get(&connection_id) {
+                None => {
+                    // If the connection ID is unknown to us, it means we didn't create it so ignore any event coming from it.
```
thomaseizinger, line 257: Same here. This will be handled by the type-system once we have https://github.com/libp2p/rust-libp2p/pull/3254. Then we can return a `dummy::ConnectionHandler` for connections that we didn't initiate ourselves and thus every event we get here is always from a connection that we initiated.

swarm/src/behaviour.rs
```diff
@@ -229,16 +222,8 @@ pub trait PollParameters {
...
-// Note: `TInEvent` is needed to be able to implement
-// [`NetworkBehaviourAction::map_in`], mapping the handler `InEvent` leaving the
-// handler itself untouched.
 #[derive(Debug)]
-pub enum NetworkBehaviourAction<
-    TOutEvent,
-    THandler: IntoConnectionHandler,
-    TInEvent = THandlerInEvent<THandler>,
-> {
+pub enum NetworkBehaviourAction<TOutEvent, TInEvent> {
```
thomaseizinger, line 226: No more trait bounds :tada:

thomaseizinger:
> This is ready for review, just keep in mind that it also contains #3335.

#3335 is merged, only depending on #3327 now.

thomaseizinger:
> > This is ready for review, just keep in mind that it also contains #3335.
> 
> #3335 is merged, only depending on #3327 now.

With #3327 merged, this is now ready for review.

swarm/src/connection/pool.rs
```diff
@@ -528,7 +517,11 @@ where
     }
 
     /// Polls the connection pool for events.
-    pub fn poll(&mut self, cx: &mut Context<'_>) -> Poll<PoolEvent<THandler>>
+    pub fn poll(
+        &mut self,
+        cx: &mut Context<'_>,
+        mut new_handler_fn: impl FnMut() -> THandler,
```
mxinden: Providing a function to create a handler on an as-needed basis is convenient, but using this pattern across the code-base can, in my eyes, make control flow difficult to understand. An alternative could be for the `Pool` to emit an event, and for the `Swarm` to respond by returning the event and the `Handler`. This approach is more verbose, but it does not introduce a new control flow and instead uses the existing one of passing events up and down. Do you have any other alternative in mind or do you have a strong opinion about the pattern being introduced @thomaseizinger?

swarm/src/behaviour.rs
```diff
@@ -743,10 +471,9 @@ pub struct DialFailure<'a, Handler> {
 /// This can include, for example, an error during the handshake of the encryption layer, or the
 /// connection unexpectedly closed.
 #[derive(Clone, Copy)]
-pub struct ListenFailure<'a, Handler> {
+pub struct ListenFailure<'a> {
     pub local_addr: &'a Multiaddr,
     pub send_back_addr: &'a Multiaddr,
-    pub handler: Handler,
```
mxinden, line 741: Should this not include the `ConnectionId` now?

protocols/relay/src/priv_client.rs
```diff
@@ -99,7 +100,9 @@ pub struct Behaviour {
     directly_connected_peers: HashMap<PeerId, Vec<ConnectionId>>,
 
     /// Queue of actions to return when polled.
-    queued_actions: VecDeque<NetworkBehaviourAction<Event, handler::Prototype>>,
+    queued_actions: VecDeque<NetworkBehaviourAction<Event, Either<handler::In, Void>>>,
+
+    pending_handler_commands: HashMap<ConnectionId, handler::In>,
```
mxinden, line 105: Entries are currently not cleaned up on `FromSwarm::DialFailure`, right?

protocols/dcutr/src/behaviour_impl.rs
```diff
@@ -68,14 +69,18 @@ pub enum Error {
...
     /// All direct (non-relayed) connections.
     direct_connections: HashMap<PeerId, HashSet<ConnectionId>>,
 
     external_addresses: ExternalAddresses,
 
     local_peer_id: PeerId,
+
+    direct_to_relayed_connections: HashMap<ConnectionId, ConnectionId>,
+
+    outgoing_direct_connection_attempts: HashMap<(ConnectionId, PeerId), u8>,
```
mxinden, line 85:
```suggestion
    /// Indexed by the [`ConnectionId`] of the relayed connection.
    outgoing_direct_connection_attempts: HashMap<(ConnectionId, PeerId), u8>,
```

mxinden (commented):
I think this is a much better approach than introduced with https://github.com/libp2p/rust-libp2p/pull/2191. I consider the mechanism to be significantly easier, both for the user and our internal implementation within `libp2p-swarm`.

I have a couple of questions below.

swarm/src/connection/pool.rs
thomaseizinger: Let me know what you think of [f5f37ab](https://github.com/libp2p/rust-libp2p/pull/3328/commits/f5f37ab3a2f78a25bf9ecc41cea92d2e41f629bb).

swarm/src/behaviour.rs
thomaseizinger: It does now!

protocols/relay/src/priv_client.rs
thomaseizinger: Done!

protocols/dcutr/src/behaviour_impl.rs
thomaseizinger: Thanks, this also made me discover a bug!

swarm/src/lib.rs
```diff
@@ -502,19 +503,13 @@ where
     /// swarm.dial("/ip6/::1/tcp/12345".parse::<Multiaddr>().unwrap());
     /// ```
     pub fn dial(&mut self, opts: impl Into<DialOpts>) -> Result<(), DialError> {
```
thomaseizinger, line 499: I opted not to return the `ConnectionId` here even though it might at first be counter-intuitive. The user already has access to it via `DialOpts` so they can get it if needed but it is a little bit more verbose.

swarm/src/connection/pool.rs
thomaseizinger: I had to adapt the `test_connect_disconnect_ban` test, please have a thorough look!

thomaseizinger:
Friendly ping @elenaf9 @jxs @mxinden. This is ready for review.

swarm/src/lib.rs
jxs: why not return it and let the user decide what to do with it? Since we have it already.

swarm/src/connection/pool.rs
```diff
@@ -518,13 +504,57 @@ where
...
                 endpoint: endpoint.into(),
                 abort_notifier: Some(abort_notifier),
                 accepted_at: Instant::now(),
             },
         );
-        Ok(connection_id)
+        Ok(())
+    }
+
+    pub fn spawn_connection(
```
jxs, line 504: yeah I think this fits here better, and being handled by the `Swarm` on `Swarm::handle_pool_event` after receiving `ConnectionEstablished` rather than on the `Pool` before returning it :rocket:

protocols/dcutr/src/behaviour_impl.rs
```diff
@@ -148,40 +157,57 @@ impl Behaviour {
...
     ) {
-        if let handler::Prototype::DirectConnection {
-            relayed_connection_id,
-            role: handler::Role::Initiator { attempt },
-        } = handler
+        let peer_id = if let Some(peer_id) = peer_id {
+            peer_id
+        } else {
+            return;
+        };
```
jxs, line 169: nit: I think we can do:
```suggestion
        let Some(peer_id) = peer_id {
            peer_id
        } else {
            return;
        };
```
if we bump `rust-version` on `Cargo.toml` to [`1.65`](https://blog.rust-lang.org/2022/11/03/Rust-1.65.0.html#let-else-statements) :stuck_out_tongue:

jxs (approved):
Overall looks good to me Thomas! I am not very familiar with `dctur` and `relay`yet so Max and Elena should also chime in.

protocols/dcutr/src/behaviour_impl.rs
thomaseizinger: I didn't want to bump MSRV just for that but I am certainly looking forward to `let-else` :)

swarm/src/lib.rs
thomaseizinger: If we return the ID, it gives the impression that we are creating it, which isn't true.

swarm/src/connection/pool.rs
thomaseizinger: Yeah it turned out surprisingly clean and fixes up some weird state tracking where we had to "filter" connection IDs that are banned. Now we just don't spawn the connection in the first place :)

swarm/src/connection/pool.rs
mxinden: Thanks. This looks good to me.

core/src/muxing/boxed.rs
```diff
@@ -13,6 +13,12 @@ pub struct StreamMuxerBox {
     inner: Pin<Box<dyn StreamMuxer<Substream = SubstreamBox, Error = io::Error> + Send>>,
 }
 
+impl fmt::Debug for StreamMuxerBox {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("StreamMuxerBox").finish_non_exhaustive()
```
mxinden, line 18: Fancy. Wasn't aware of `finish_non_exhaustive`.

swarm/src/lib.rs
mxinden: No strong opinion. Fine with either. Let's keep as is.

mxinden (approved):
Ready to go from my end.

swarm/src/lib.rs
```diff
@@ -762,74 +751,77 @@ where
...
-                other_established_connection_ids,
+                connection,
                 concurrent_dial_errors,
                 established_in,
             } => {
                 if self.banned_peers.contains(&peer_id) {
-                    // Mark the connection for the banned peer as banned, thus withholding any
-                    // future events from the connection to the behaviour.
-                    self.banned_peer_connections.insert(id);
                     self.pool.disconnect(peer_id);
```
mxinden: Given that we drop `connection` here, is the call to `pool.disconnect` still needed? There should be no connection to the peer in the pool. The previous `Swarm::ban_peer_id` should have already called `pool.disconnect`.

Also, if I am not mistaken, we would previously gracefully close connections to banned peers via the `pool.disconnect` call. Given that `connection` is now just dropped, this is no longer the case, right?

mxinden (commented):
This still needs further investigation:

swarm/src/lib.rs
thomaseizinger: > Given that we drop `connection` here, is the call to `pool.disconnect` still needed? There should be no connection to the peer in the pool. The previous `Swarm::ban_peer_id` should have already called `pool.disconnect`.

Originally I thought that we might have other connections to this peer but I guess that isn't actually possible because we would have disconnected those at the time of banning the peer.

> Also, if I am not mistaken, we would previously gracefully close connections to banned peers via the `pool.disconnect` call. Given that `connection` is now just dropped, this is no longer the case, right?

True. I didn't bother much because this code will go away anyway once we implement it via the new `NetworkBehaviour` callbacks but yeah, it is a change in behaviour so we should probably keep this behaviour.

This made me think though, we should probably gracefully close connections in https://github.com/libp2p/rust-libp2p/pull/3254 that have been rejected by a `NetworkBehaviour`.

thomaseizinger: I've pushed some more commits, let me know what you think!

mxinden (approved):

swarm/src/lib.rs
mxinden: Changes look good to me, thanks!

mxinden:
@thomaseizinger ready to merge from my end. Please trigger mergify once you are ready. Looking forward for this to land.

PR #3331, opened 2023-01-16
ci: update interop workflow

jxs:
## Description

Addresses https://github.com/libp2p/test-plans/pull/99

## Change checklist


mxinden:
https://github.com/libp2p/github-mgmt/pull/105/ is merged. Let me know once this is ready for review.

.github/workflows/interop-test.yml
```diff
@@ -1,32 +1,40 @@
...
+      - name: Upload ping versions info
+        uses: actions/upload-artifact@v3
+        with:
+          name: ping-versions
+          path: ./test-plans/ping-versions.json
+      - name: Upload image tar
+        uses: actions/upload-artifact@v3
+        with:
+          name: ping-image
+          path: ./test-plans/ping-image.tar
```
thomaseizinger, line 38: Now that we are using plain docker compose, can't we just build the image as part of the workflow here, upload it to GitHub's container registry and reference it here? That seems a lot simpler than storing it as an artifact.

cc @MarcoPolo

thomaseizinger: In more detail, why don't we have the following:

1. A workflow per repository that only runs on pushes to master that builds a docker container and uploads it to the container registry under the tag "master".
2. A workflow per repository that only runs on tag pushes and builds a docker container and uploads it to the container registry under the Git tag, i.e. `0.45.0`
3. A PR workflow that builds a container of the current commit, uploads it to the container registry under the tag of the PR number.

Once (3) finishes, we can then kick off a dependent workflow that runs a matrix of containers against each other.

test-plans/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+async-trait = "0.1.58"
+env_logger = "0.9.0"
+futures = "0.3.1"
+if-addrs = "0.7.0"
+log = "0.4"
+redis = { version = "0.22.1", features = ["tokio-native-tls-comp", "tokio-comp"] }
+tokio = { version = "1.24.1", features = ["full"] }
+
+libp2p = { git = "https://github.com/libp2p/rust-libp2p", branch = "master", default_features = false, features = ["websocket", "quic",  "mplex", "yamux", "tcp", "tokio", "ping", "noise", "tls", "dns", "rsa", "macros", "webrtc"] }
+testplan = { version = "0.2.0", git = "https://github.com/jxs/test-plans/", branch = "add-webrtc-transport-multidim-interop" }
```
thomaseizinger: I'd rather not have this dependency in here. Can we just copy the code over?

test-plans/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+anyhow = "1"
+async-trait = "0.1.58"
+env_logger = "0.9.0"
+futures = "0.3.1"
+if-addrs = "0.7.0"
+log = "0.4"
+redis = { version = "0.22.1", features = ["tokio-native-tls-comp", "tokio-comp"] }
+tokio = { version = "1.24.1", features = ["full"] }
+
+libp2p = { git = "https://github.com/libp2p/rust-libp2p", branch = "master", default_features = false, features = ["websocket", "quic",  "mplex", "yamux", "tcp", "tokio", "ping", "noise", "tls", "dns", "rsa", "macros", "webrtc"] }
```
thomaseizinger: I think this should be a path dependency and for building the correct binary, we should checkout `master` in the workflow.

test-plans/Cargo.toml
jxs: interesting! Why you'd prefer having redundancy instead of depending on the abstractions defined there? Max seems to also agree with you as well so I copied `lib.rs` here ptal :)

test-plans/Cargo.toml
jxs: Thanks Thomas, I was thorn on that and opted not to touch that part initially, but agree it makes sense, updated the `Dockerfile` accordingly please take a look.

.github/workflows/interop-test.yml
MarcoPolo: You can upload to a container registry if you want. If the containerImageID is a tag, then docker will automatically try to fetch that tag. You don't have to upload image artifacts if you don't want to.

That said, I think the approach of building the container from the commit is better. For one it's guaranteed to be reproducible. You don't have multiple commits from the same PR mutating the tag number. Two, it's the same workflow on master, PRs, or tags. Three, when you introduce a container registry, now you need to deal with authentication, limits, and garbage collecting from that container registry.

The one benefit I see to container registries is them acting as a cache for faster tests. But we can cache layers already using GitHub Actions cache. 

I'm not sure the extra work involved in setting up and maintaining a registry is worth it. But of course, if you want to feel free. I'm just sharing my thoughts from thinking about this as well.

test-plans/Cargo.toml
MarcoPolo: Agree with copying the code. I'll add that in the longer term, all the ping implementations are in the implementation repo. The `test-plans` repo is simply a source of truth of what versions are supported and what they support. But the building/code for the impl should be in the implementation repo.

test-plans/Dockerfile
```diff
@@ -0,0 +1,19 @@
+FROM rust:1.65-bullseye as builder
+WORKDIR /usr/src/libp2p
+
+# TODO fix this, it breaks reproducibility
+RUN apt-get update && apt-get install -y cmake protobuf-compiler
```
MarcoPolo: Do we still need the protobuf compiler? I thought we fixed this by inlining generated content?

test-plans/Makefile
```diff
@@ -0,0 +1,10 @@
+all: ping-image.tar
+
+ping-image.tar: Dockerfile Cargo.toml src
```
MarcoPolo: Does this properly depend on all files in src? I thought you needed `wildcard`, but not sure.

.github/workflows/interop-test.yml
thomaseizinger: Thanks for sharing your thoughts. Perhaps a mix of both would be the best solution?

- The `master` tag would only ever produce one container so no GC needed, i.e. the container for master would be updated every time the master branch gets updated. That hurts reproducibility a bit but it is not a big deal IMO. We already accept that merging one PR affects other PRs.
- The containers for releases are useful for other implementations too. I don't think we can utilize the GH Actions Cache here? i.e. `go-libp2p` testing against releases of `rust-libp2p`
- Not uploading the container of the current commit/PR is a good idea, those would just be garbage we need to collect.

If we can remove container builds of master and tags from this workflow, then the Dockerfile also gets simpler because we always just build for the current source, no overrides needed.

test-plans/Dockerfile
thomaseizinger: That isn't merged yet :)

test-plans/Cargo.toml
thomaseizinger: > Agree with copying the code. I'll add that in the longer term, all the ping implementations are in the implementation repo. The `test-plans` repo is simply a source of truth of what versions are supported and what they support. But the building/code for the impl should be in the implementation repo.

What he said 😄

.github/workflows/interop-test.yml
thomaseizinger: > I'm not sure the extra work involved in setting up and maintaining a registry is worth it. But of course, if you want to feel free. I'm just sharing my thoughts from thinking about this as well.

GitHub makes authenticating and interacting with its own registry really easy so that wouldn't be much work in my experience.

test-plans/Dockerfile
mxinden: For completeness, this is tracked in https://github.com/libp2p/rust-libp2p/issues/3024.

test-plans/Cargo.toml
```diff
@@ -0,0 +1,21 @@
+[package]
+edition = "2021"
+name = "testcases"
+version = "0.1.0"
+
+# Required to avoid conflicts with parent workspace.
+[workspace]
```
thomaseizinger: What conflicts are we talking? Can this crate not be part of our other workspace so we make sure it compiles at all times?

test-plans/Cargo.toml
```diff
@@ -0,0 +1,21 @@
+[package]
+edition = "2021"
+name = "testcases"
```
thomaseizinger: Perhaps rename this directory and crate to `interop-tests`?

test-plans/Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+
+COPY ./ ./
+
+RUN cd ./test-plans/ && cargo build  # Initial build acts as a cache.
```
thomaseizinger: I would prefer if we can write this `Dockerfile` such that `docker build .` _inside_ the `test-plans` directory works. In other words, I think it is preferable to run the build within this directory instead of running `docker build -f ./test-plans/Dockerfile`.

test-plans/Dockerfile
```diff
@@ -0,0 +1,19 @@
+FROM rust:1.65-bullseye as builder
+WORKDIR /usr/src/libp2p
+
+# TODO fix this, it breaks reproducibility
+RUN apt-get update && apt-get install -y cmake protobuf-compiler
+
+COPY ./ ./
```
thomaseizinger: Copying everything will invalidate this layer every time you modify anything. You need to copy files in two steps:

1. Copy `Cargo.toml` + `Cargo.lock` + `src/main.rs`
2. Run `cargo build`
3. Copy `src/lib.rs` and `src/bin/ping.rs`

Because the `main.rs` is a dummy file, the docker layer created in step (2) does not get invalidated when you work on `src/bin/ping.rs` or `src/lib.rs`.

test-plans/src/lib.rs
```diff
@@ -0,0 +1,123 @@
...
+
+    async fn loop_on_next(&mut self);
+
+    fn local_peer_id(&self) -> String;
+}
+
+/// Run a ping interop test. Based on `is_dialer`, either dial the address
+/// retrieved via `listenAddr` key over the redis connection. Or wait to be pinged and have
+/// `dialerDone` key ready on the redis connection.
+pub async fn run_ping<S>(
```
thomaseizinger: Now that we only depend on one `libp2p` version, we can completely bin ( :wastebasket: ) this abstraction and have all the code in one place and properly reference types like `PeerId` etc!

.github/workflows/interop-test.yml
```diff
@@ -1,32 +1,37 @@
...
+  run-multidim-interop:
+    name: Run multidimensional interoperability tests
+    needs: build-ping-container
+    uses: "libp2p/test-plans/.github/workflows/run-testplans.yml@master"
```
thomaseizinger, line 42: I think we could save some complexity here if this were just a [composite action](https://docs.github.com/en/actions/creating-actions/creating-a-composite-action) that we call after building the images rather than a separate job.

That would allow the action within `test-plans` to access local state like built docker images, configuration files without us having to up and download artifacts.

@MarcoPolo What was the reasoning for making this a workflow rather than a composite action?

.github/workflows/interop-test.yml
```diff
@@ -1,32 +1,37 @@
...
     with:
-      composition_file: "ping/_compositions/all-interop-latest.toml"
-      custom_git_target: github.com/${{ github.event.pull_request.head.repo.full_name || github.event.repository.full_name }}
-      custom_git_reference: ${{ github.event.pull_request.head.sha || github.sha }}
-      custom_interop_target: rust
+      dir: "multidim-interop"
+      extra-versions: ping-versions
```
thomaseizinger, line 45: What does `extra-versions` here mean? @MarcoPolo

We seem to be referencing the same image in there that we are also tarring up.

.github/workflows/interop-test.yml
thomaseizinger: For example, within the same job, we wouldn't need to tar up the container because it would still be accessible by tag.

thomaseizinger (commented):
Some more thoughts!

This is looking really good already :)

test-plans/Makefile
jxs: Yeah good point Marco, it seems to only depend on the directory creation/modification itself. Updated to use `wildcard`

test-plans/Cargo.toml
jxs: yeah good point, thanks Thomas

test-plans/Cargo.toml
jxs: as this is a subdir to a workspace, without that declaration cargo outputs this:
```
error: current package believes it's in a workspace when it's not:
current:   /home/jxs/dev/oss/rust-libp2p/test-plans/Cargo.toml
workspace: /home/jxs/dev/oss/rust-libp2p/Cargo.toml
```

Can we have it as a subdir without belonging to the other workspace?

test-plans/Dockerfile
thomaseizinger: It is maybe a bit hacky but what if we just build the binary outside of the container and copy it in afterwards? That would allow us to use regular Github Action caching to make it fast and we don't have to much around with docker's layers.

If we don't optimize this step, we will have to wait for 8-10 minutes on every run for this to be built.

thomaseizinger: I wonder if we can reuse some layers from the image we are going to build on the master branch. Something like: https://github.com/dtinth/github-actions-docker-layer-caching-poc/blob/84481ef34a8f6039fe27743b4ffe0d38c7e7301f/.github/workflows/dockerimage.yml#L61-L68

One of the problems might be that our dependencies include a path dependency (obviously) so even that "built all dependencies" layer will be invalid on each new commit.

thomaseizinger: > It is maybe a bit hacky but what if we just build the binary outside of the container and copy it in afterwards? That would allow us to use regular Github Action caching to make it fast and we don't have to much around with docker's layers.
> 
> If we don't optimize this step, we will have to wait for 8-10 minutes on every run for this to be built.

If we go down this path, the crate can just be part of the regular workspace and share the same `Cargo.lock` file which makes that part also simpler. Our CI already runs on ubuntu, so we can just take one of the ubuntu-slim images, shove the binary into it and done.

The GitHub actions caches are a lot more resilient towards changes, both in the lock file and in the source code so we should get speedy builds most of the time.

test-plans/src/lib.rs
jxs: yeah exactly, thanks, updated!

.github/workflows/interop-test.yml
MarcoPolo: I think the ability to checkout an arbitrary commit of test-plans and be able to run the exact those exact tests is a very good feature (reproduicibility). As long as we hold that, I don’t mind using tagged images in a registry.

But what is the benefit of tagging images in a registry? To me, the only benefit is faster builds. But you can get this with caching as well. Maybe standard GHA cache isn’t good enough (since it caches per branch I believe), but we could cache the layers from the latest test-plans master builds as well, and I think that would cover all the use cases of tagged images.

> The master tag would only ever produce one container so no GC needed, i.e. the container for master would be updated every time the master branch gets updated.

I don’t see the benefit here since you are building this every time. When do you get to reuse this tag?

> If we can remove container builds of master and tags from this workflow, then the Dockerfile also gets simpler because we always just build for the current source, no overrides needed.

What overrides are used? I imagine you are just building from the current source.

One idea for released versions is that inside the test-plans repo we specify the way to build the interop image is to checkout this repo at this commit, then cd into test-plans and run `make`.

test-plans/Dockerfile
MarcoPolo: Would be great to be able to simply copy the binary in! I’m not sure if there are shared libs you are depending on that you would also need in the container.

A tangent, one idea is to use Nix to build the binary and then you can copy the binary + all shared libraries into a scratch image. This lets you use [Cachix](https://www.cachix.org/) to speed up the Nix build, and the container image builds will always be fast because you are simply `cp`ing files. A bit more details here: https://marcopolo.io/code/nix-and-small-containers/

jxs: yeah I started by trying that, the problem is that [docker doesn't allow to copy dirs from parent dir](https://stackoverflow.com/questions/24537340/docker-adding-a-file-from-a-parent-directory/34300129#34300129) .  
 That's why I also eliminated the caching you had established and mentioned again on https://github.com/libp2p/rust-libp2p/pull/3331#discussion_r1073535617. I should have mentioned this before sorry, but I bet you have a nice solution for this :innocent:

test-plans/Dockerfile
mxinden: See https://github.com/libp2p/test-plans/blob/master/ping/rust/Dockerfile as an example @jxs.

test-plans/Dockerfile
mxinden: I don't have an opinion on the above. That said, are folks fine with delaying these decisions to future pull requests?

We are currently holding off merging pull requests into rust-libp2p master, waiting for them to be tested with the interoperability setup first. Doesn't make this urgent, but still would allow more parallelization on this repository.

test-plans/Cargo.toml
mxinden: @jxs did you add it here?

https://github.com/libp2p/rust-libp2p/blob/73cbbe29679f5d56d81d5477d3796e82712f9ac2/Cargo.toml#L144-L145

mxinden (commented):
Thanks for the work here @jxs! Direction looks good to me. Still many outstanding comments. I suggest let's try to delay some of the work to follow-up pull requests to unblock other pull requests on libp2p/rust-libp2p.

.github/workflows/interop-test.yml
thomaseizinger: > I think the ability to checkout an arbitrary commit of test-plans and be able to run the exact those exact tests is a very good feature (reproduicibility). As long as we hold that, I don’t mind using tagged images in a registry.
> 
> But what is the benefit of tagging images in a registry? To me, the only benefit is faster builds.

The benefit is faster builds but it also just feels wrong. The source code for a test for e.g. libp2p 0.45 should _never_ change. The version is out there, our users are using it. If we want to take interoperability seriously, I think it is good to constrain ourselves such that changing the test for a particular released version is _hard_.

Building the binary every time doesn't sit well for me on that front.

> Maybe standard GHA cache isn’t good enough (since it caches per branch I believe), but we could cache the layers from the latest test-plans master builds as well, and I think that would cover all the use cases of tagged images.

It caches per user-defined key which can be completely arbitrary and is not restricted to branches or anything. Bear in mind that each repo only has 10GB here and at least in `rust-libp2p` we already have to jump through hoops to not blow this cache limit for our regular jobs :(

> > The master tag would only ever produce one container so no GC needed, i.e. the container for master would be updated every time the master branch gets updated.
> 
> I don’t see the benefit here since you are building this every time. When do you get to reuse this tag?

It gets reused across all pushes on all open PRs until the next one merges. I myself have usually anywhere from 2-10 PRs open that I work on in parallel i.e. I switch to another PR once I incorporated feedback and wait for the next round of reviews.

> > If we can remove container builds of master and tags from this workflow, then the Dockerfile also gets simpler because we always just build for the current source, no overrides needed.
> 
> What overrides are used? I imagine you are just building from the current source.

My bad, I wrongly assumed that we would still need the "sed"-ing inside the Dockerfile. That should be gone now in any case.

> One idea for released versions is that inside the test-plans repo we specify the way to build the interop image is to checkout this repo at this commit, then cd into test-plans and run `make`.

This exact step could have happened when the tag was pushed. It is like pre-built binaries of any other application here on GitHub.

test-plans/Dockerfile
thomaseizinger: Happy to merge without optimisations for now as long as the CI job is stable.

thomaseizinger (approved):
Great stuff to see this working again, thank you @MarcoPolo and @jxs !

Follow-ups I'd like to see (happy to take some of these over):

- Use a composite action instead of a workflow: https://github.com/libp2p/rust-libp2p/pull/3331#discussion_r1073550520. We should be able to add this backwards-compatible so repositories can transition without breaking CI.
- Pre-build images for tagged versions.
- Optimise building of the Dockerfile in general to be as quick as possible. Copying the binary would be my suggestion but I am open to other solutions. 
- Ideally pre-build image for `master` branch. If we have a good caching strategy for building the Dockerfile, it may not be worth prebuilding the image for master. **Edit**: This may be moot, see https://github.com/libp2p/rust-libp2p/pull/3331#discussion_r1080611886.

.github/workflows/interop-test.yml
MarcoPolo: Thanks. I think we're starting to understand each other's points and we agree on the fundamentals.

Just a couple of questions for me:

> It gets reused across all pushes on all open PRs until the next one merges.

Maybe there's a misunderstanding somewhere? An open rust-libp2p PR wouldn't do an interop test against rust-libp2p master (we could do that, I'm just not sure it's useful). It runs interop tests against every other released version of libp2p.

>  The source code for a test for e.g. libp2p 0.45 should never change. The version is out there, our users are using it. If we want to take interoperability seriously, I think it is good to constrain ourselves such that changing the test for a particular released version is hard.

I agree with this. I don't think building from source prevents this. If we constrain ourselves to never mutate an image tag, then I think we agree except on the little details. To me a sha256 hash is by definition an immutable tag of an image.

I don't have strong opinions on how rust-libp2p wants to handle building their image. If the project wants to simply submit a container tag that will be used in the interop tests, I only ask that the tag is immutable.

.github/workflows/interop-test.yml
MarcoPolo: The source of truth for the capabilities of a released libp2p version are defined in [`version.ts`](https://github.com/libp2p/test-plans/blob/master/multidim-interop/versions.ts). The extra versions is more version data that is added to that `versions.ts` array. This lets you add a "rust-libp2p-head" version to the defined versions and run an interop test between all defined versions on all capabilites. This combined with the ability to filter tests lets you run the interop tests that only include "rust-libp2p-head".

.github/workflows/interop-test.yml
MarcoPolo: Good point! I didn't realize I could make a composite action. My knowledge here was out of date. Thanks!

I'll try this out soon. I agree it'll be nice not to have to upload the image. And the upload artifacts are always directories, so it'll be nice to avoid that.

.github/workflows/interop-test.yml
thomaseizinger: > > It gets reused across all pushes on all open PRs until the next one merges.
> 
> Maybe there's a misunderstanding somewhere? An open rust-libp2p PR wouldn't do an interop test against rust-libp2p master (we could do that, I'm just not sure it's useful). It runs interop tests against every other released version of libp2p.

Interesting, I thought we also ran interop tests against our own master branch. @mxinden do I remember this incorrectly?

If not then I agree that it is not useful to have a moving `master` tag.

> >  The source code for a test for e.g. libp2p 0.45 should never change. The version is out there, our users are using it. If we want to take interoperability seriously, I think it is good to constrain ourselves such that changing the test for a particular released version is hard.
> 
> I agree with this. I don't think building from source prevents this. If we constrain ourselves to never mutate an image tag, then I think we agree except on the little details. To me a sha256 hash is by definition an immutable tag of an image.
> 
> I don't have strong opinions on how rust-libp2p wants to handle building their image. If the project wants to simply submit a container tag that will be used in the interop tests, I only ask that the tag is immutable. 

The tag should definitely be immutable. As far as I know, one can append the sha256 hash of a docker image tag to the overall ID and thus guarantee that anyone pulling this image from the registry will get exactly the image with this hash. That is how I would specify those images in test-plans.

Example:

`docker.io/library/hello-world@sha256:faa03e786c97f07ef34423fccceeec2398ec8a5759259f94d99078f264e9d7a`

test-plans/Cargo.toml
jxs: no no, cause I assumed what Thomas said, that it would be better for it not to be there, would you prefer that?

test-plans/Dockerfile
jxs: yeah I was researching the Docker in Docker option raised by Thomas on https://github.com/libp2p/rust-libp2p/pull/3331#discussion_r1073532935 cause that was the reason for having this. As we are depending on `rust-libp2p` from the parent dir, it needs to be copied along `transports` and `protocols`, I take it you prefer to copy all that instead of having this right?

thomaseizinger: I think we are better off with:

1. making this crate part of the main workspace
2. adding our usual caching layer in github actions (with `save-if: <master branch>`)
3. building the binary outside of docker
4. have the dockerfile copy in the binary from the root target directory

The cache is reusable across all our jobs (master, PRs and tags) so we get maximum cache hits and the configuration is very simple. The only downside is that MacOS and Windows users cannot build this docker image locally. I am okay with that:

- We have no MacOS or Windows users among the core maintainers
- The Docker container isn't a strict requirement for the tests, the binary also works standalone

test-plans/Cargo.toml
thomaseizinger: Based on https://github.com/libp2p/rust-libp2p/pull/3331#discussion_r1080732134, yes I'd say let's make it part of the workspace. (Doesn't have to be in this PR)

.github/workflows/interop-test.yml
thomaseizinger: Thanks for explaining, this makes sense now.

Is there a usecase where we extend `versions.ts` with more than one image and/or don't filter by the one we just added? It looks like the configuration API of this action could be simpler, i.e. it just needs the contents of one entry in `versions.ts` and we always want to filter the tests for this one version.

thomaseizinger: In particular, this seems to be sufficient and perhaps a bit easier to understand:

```yaml
  run-multidim-interop:
    name: Run multidimensional interoperability tests
    needs: build-ping-container
    uses: "libp2p/test-plans/.github/workflows/run-testplans.yml@master"
    with:
      imageId: "rust-libp2p-head"
      transports: ["ws", "tcp", "quic-v1", "webrtc"]
      secureChannels: ["tls", "noise"]
      muxers: ["mplex", "yamux"]
```

- The name of the version can be implied from the image ID (by convention we should make this a tag and not just a hash).
- With the composite action suggested in https://github.com/libp2p/rust-libp2p/pull/3331/files#r1073550520, we don't need to pass the artifact name of the tar'd image.
- By directly specifying the transports, secure channels and muxers, we can avoid a configuration file in our repository.
- Not sure I understand why `dir: "multidim-interop"` needs to be specified, can't this be implied by the GitHub action?

.github/workflows/interop-test.yml
```diff
@@ -1,32 +1,46 @@
...
-    uses: "libp2p/test-plans/.github/workflows/run-composition.yml@master"
+  build-ping-container:
+    name: Build Ping interop container
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: Swatinem/rust-cache@359a70e43a0bb8a13953b04a90f76428b4959bb6 # v2.2.0
+        with:
+          shared-key: interop-tests
+          save-if: <master branch>
```
thomaseizinger: This was meant to be a placeholder, you'll have to put the correct expression here :)

See the CI workflow.

.github/workflows/interop-test.yml
```diff
@@ -1,32 +1,46 @@
...
+      - name: Install Protoc
+        run: sudo apt-get install protobuf-compiler
+      - name: build test plans
+        run: cargo build --release -p interop-tests
```
thomaseizinger: Should this be part of the makefile maybe?

test-plans/Makefile
```diff
@@ -0,0 +1,10 @@
+all: ping-image.tar
+
+ping-image.tar: Dockerfile Cargo.toml $(wildcard src/*)
+	cd .. && docker build -t rust-libp2p-head -f test-plans/Dockerfile .
```
thomaseizinger: This is just an intuition but I think this will send the entire directory to the Docker daemon as "context" so the actual build takes long because it copies the entire target directory to somewhere but it only ends up picking one binary out of it.

It might be worthwhile to move the binary out of the target dir into the `test-plans` dir and run `docker build` from within there instead (and remove the `cd ..`)

test-plans/Makefile
```diff
@@ -0,0 +1,10 @@
+all: ping-image.tar
+
+ping-image.tar: Dockerfile Cargo.toml $(wildcard src/*)
```
thomaseizinger: This should mention the binary now instead of the source code I believe.

.github/workflows/interop-test.yml
```diff
@@ -1,32 +1,46 @@
...
+      - uses: actions/checkout@v3
+        with:
+          submodules: recursive
```
thomaseizinger: Is this 2nd checkout necessary?

.github/workflows/interop-test.yml
```diff
@@ -1,32 +1,46 @@
...
-
 on:
   pull_request:
   push:
     branches:
-      - master
-
-concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}
-  cancel-in-progress: true
```
thomaseizinger: I think we should keep that bit.

thomaseizinger (commented):
Few more thoughts :)

.github/workflows/interop-test.yml
jxs: thanks, added back!

.github/workflows/interop-test.yml
jxs: no yeah, removed it after noticing. Thanks Thomas

test-plans/Makefile
jxs: since we are now compiling the code in the makefile it should be `src/bin/ping.rs` right?

test-plans/Makefile
jxs: yup, agreed and updated. Thanks Thomas

.github/workflows/interop-test.yml
jxs: yup, makes sense, updated!

jxs:
(sorry for having force pushed @thomaseizinger , didn't know you were already reviewing it :pray: )
Addressed all the last remarks by Thomas, Binary is now being built outside of the container and cached with our caching action. I  think this is now ready to be merged CC @mxinden.

.github/workflows/interop-test.yml
mxinden: > > > It gets reused across all pushes on all open PRs until the next one merges.
> > 
> > 
> > Maybe there's a misunderstanding somewhere? An open rust-libp2p PR wouldn't do an interop test against rust-libp2p master (we could do that, I'm just not sure it's useful). It runs interop tests against every other released version of libp2p.
> 
> Interesting, I thought we also ran interop tests against our own master branch. @mxinden do I remember this incorrectly?

If I recall correctly, we did. I was planning to have each pull request tested against `master` with the wo-testground tests, though based on intuition, without specific reasoning. Thinking about it some more, I don't see what kind of bugs this would catch, thus I suggest not investing into PR-to-master testing for now. Thoughts?

test-plans/README.md
```diff
@@ -0,0 +1,34 @@
+# test-plans test implementation
+
+This folder defines the implementation for the test-plans interop tests.
+
+# Running this test locally
```
mxinden, line 5: :rocket: worked like a charm.

mxinden (approved):

mxinden:
> Addressed all the last remarks by Thomas, Binary is now being built outside of the container and cached with our caching action. I think this is now ready to be merged CC @mxinden.

Great work @jxs @MarcoPolo @thomaseizinger. Thank you!

Please add the send-it label @jxs once you want to merge.

Unless I am missing something, the below is the (potential) outstanding follow-up work.

> * Use a composite action instead of a workflow: [#3331 (comment)](https://github.com/libp2p/rust-libp2p/pull/3331#discussion_r1073550520). We should be able to add this backwards-compatible so repositories can transition without breaking CI.
> 
> * Pre-build images for tagged versions.
> 
> * Optimise building of the Dockerfile in general to be as quick as possible. Copying the binary would be my suggestion but I am open to other solutions.
> 
> * Ideally pre-build image for `master` branch. If we have a good caching strategy for building the Dockerfile, it may not be worth prebuilding the image for master. **Edit**: This may be moot, see [#3331 (comment)](https://github.com/libp2p/rust-libp2p/pull/3331#discussion_r1080611886).

https://github.com/libp2p/rust-libp2p/pull/3331#pullrequestreview-1260847859

Potential format change to `extra-versions` https://github.com/libp2p/rust-libp2p/pull/3331#discussion_r1080736900

.github/workflows/interop-test.yml
mxinden: > But what is the benefit of tagging images in a registry?

Another benefit would be for someone to easily test their libp2p-xxx against libp2p-yyy v1.2.3 locally without having checkout a repository, build a Docker image or install libp2p-yyy toolchain, simply pulling and running a pre-build image from a public registry. Just mentioning it here. I am fine with either and fine with Go doing ad-hoc builds and Rust pushing Docker images.

mxinden:
Adding the below to the list of follow-up tasks:

- Make "Interoperability Testing / Run multidimensional interoperability tests / Run testplans test" a required task via https://github.com/libp2p/github-mgmt.

jxs:
> Adding the below to the list of follow-up tasks:
> 
> * Make "Interoperability Testing / Run multidimensional interoperability tests / Run testplans test" a required task via https://github.com/libp2p/github-mgmt.

https://github.com/libp2p/github-mgmt/pull/107 for reference

.github/workflows/interop-test.yml
MarcoPolo: > Not sure I understand why dir: "multidim-interop" needs to be specified, can't this be implied by the GitHub action?

This should be changed. Originally the "run test-plans" was generic enough that it would work for any `npm test` like thing. But there's little value for this, and it would be nicer to be clearer here.

> By directly specifying the transports, secure channels and muxers, we can avoid a configuration file in our repository.

It's true. With the tradeoff of dealing not being able to run this locally as easily. I think it's nice to be able to locally run: `npm test -- --extra-versions=$RUST_LIBP2P/test-plans/version.json --name-filter="rust-libp2p-head"`. Inputs are also only strings, so you'd have to stringify the given example.

I agree with all the other points :)

thomaseizinger: > > By directly specifying the transports, secure channels and muxers, we can avoid a configuration file in our repository.
> 
> It's true. With the tradeoff of dealing not being able to run this locally as easily. I think it's nice to be able to locally run: `npm test -- --extra-versions=$RUST_LIBP2P/test-plans/version.json --name-filter="rust-libp2p-head"`. Inputs are also only strings, so you'd have to stringify the given example.

How often do we need that? At the moment, this config file just sits there, not co-located with its consumer and it requires a fair amount of knowledge to figure out what it is for.

.github/workflows/interop-test.yml
```diff
@@ -1,32 +1,47 @@
...
+        with:
+          submodules: recursive
+      - uses: Swatinem/rust-cache@359a70e43a0bb8a13953b04a90f76428b4959bb6 # v2.2.0
+        with:
+          shared-key: interop-tests
+          save-if: ${{ github.ref == 'refs/heads/master' }}
+      - name: Install Protoc
+        run: sudo apt-get install protobuf-compiler
+      - name: Build image
+        working-directory: ./test-plans
```
thomaseizinger, line 27: Sometimes we call things test-plans, sometimes we call them interop-tests.

I'd be in favor of calling all of this `interop-tests`. `test-plans` is too generic IMO and doesn't tell me anything about what they are testing.

test-plans/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+[dependencies]
+anyhow = "1"
+async-trait = "0.1.58"
+env_logger = "0.9.0"
+futures = "0.3.1"
+if-addrs = "0.7.0"
+log = "0.4"
+redis = { version = "0.22.1", features = ["tokio-native-tls-comp", "tokio-comp"] }
+tokio = { version = "1.24.1", features = ["full"] }
+
```
thomaseizinger, line 16: Any reason for this newline?

test-plans/Makefile
```diff
@@ -0,0 +1,14 @@
+all: ping-image.tar
+
+ping-image.tar: Dockerfile Cargo.toml src/bin/ping.rs
```
thomaseizinger, line 3: Technically, this also needs to rebuild on every change in libp2p but I am not sure we can or should express this here.

I am not a massive fan of Makefiles altogether. They are sometimes nice to bundle a few commands together but you run into limitations very quickly. At that point, you then have to have extra knowledge again which kind of defeats the point of trying to encapsulate it.

Given that we will be supplying images to the other repos for our release, nobody apart from us has to build this so we could get rid of this file and just inline it into our CI job.

Thoughts @libp2p/rust-libp2p-maintainers ?

test-plans/Dockerfile
```diff
@@ -0,0 +1,4 @@
+FROM ubuntu:latest
```
thomaseizinger, line 1: This needs to align with the ubuntu version of our CI job. Perhaps it is better to use a fixed tag in both with a comment pointing to each other.

.github/workflows/interop-test.yml
thomaseizinger: > It's true. With the tradeoff of dealing not being able to run this locally as easily. I think it's nice to be able to locally run: `npm test -- --extra-versions=$RUST_LIBP2P/test-plans/version.json --name-filter="rust-libp2p-head"`. Inputs are also only strings, so you'd have to stringify the given example.

Would it make sense to unify the API of the test runner to just accept a series of JSON files with one version each? i.e. it builds the matrix of tests out of all files that are passed to it. Each JSON file would contain exactly one entry.

We could then pass the JSON as a string into the workflow and it writes it to a file. If you really need to run things locally, you just have to quickly copy paste that and make the file.

To me, that seems like an acceptable trade-off for having one less configuration file around per repository :)

test-plans/Dockerfile
jxs: sorry forgot to comment on why the change, it's due to the libc mismatch. I could not run the CI compiled ubuntu-latest bin on the Debian bullseye as it's version is older.

thomaseizinger: Yeah that might happen again if either of these rolling tags update so I'd like us to point to 22.10 or whatever latest is both here and in the github workflow (`runs-on:`).

.github/workflows/interop-test.yml
MarcoPolo: > Would it make sense to unify the API of the test runner to just accept a series of JSON files with one version each? i.e. it builds the matrix of tests out of all files that are passed to it. Each JSON file would contain exactly one entry.

Yeah, this seems reasonable 👍 .

> We could then pass the JSON as a string into the workflow and it writes it to a file. If you really need to run things locally, you just have to quickly copy paste that and make the file.

The less logic that is in the github actions yaml the better. At least for me, I find this stuff hard to work with since it only runs in CI. The simpler the yaml file is the better.

I don’t see why having this configuration file is bad. I actually think it’s a good thing because:
1. It defines what the capabilities of the current rust-libp2p are.
2. If a new capability is added (e.g. webtransport), then you modify this json file, not a stringified version in a github action yaml file.
3. When a release is made, you simply have to copy this JSON file to `test-plans/multidim-interop/version.ts` so other implementions test against it.

thomaseizinger: > > Would it make sense to unify the API of the test runner to just accept a series of JSON files with one version each? i.e. it builds the matrix of tests out of all files that are passed to it. Each JSON file would contain exactly one entry.
> 
> Yeah, this seems reasonable 👍 .
> 

Nice, I'll open an issue on the repo.

> I don’t see why having this configuration file is bad. I actually think it’s a good thing because:
> 1. It defines what the capabilities of the current rust-libp2p are.
> 2. If a new capability is added (e.g. webtransport), then you modify this json file, not a stringified version in a github action yaml file.
> 3. When a release is made, you simply have to copy this JSON file to `test-plans/multidim-interop/version.ts` so other implementions test against it.

You are mentioning some good points. Happy to keep the file, thanks for getting to the bottom of this :)

PR #3341, opened 2023-01-17
refactor(core)!: remove `EitherOutput`

thomaseizinger:
## Description

The trick with this one is to use `futures::Either` everywhere where we may wrap something that implements any of the `futures` traits. This includes the output of `EitherFuture` itself. We also need to implement `StreamMuxer` on `future::Either` because `StreamMuxer`s may be the the `Output` of `InboundUpgrade`.

## Notes

~Draft because it doesn't compile yet.~

This should unblock the compile errors I've been getting while resolving the merge conflicts in https://github.com/libp2p/rust-libp2p/pull/3254.

## Change checklist


swarm/src/handler/either.rs
```diff
@@ -101,110 +101,18 @@ where
...
-                error: ConnectionHandlerUpgrErr::Timeout,
-                info: Either::Right(info),
-            } => Either::Right(DialUpgradeError {
-                error: ConnectionHandlerUpgrErr::Timeout,
-                info,
-            }),
-            _ => unreachable!(),
-        }
-    }
-}
```
thomaseizinger, line 206: These can now go away because we unified the either types and there is already an implementation for it.

thomaseizinger:
This is ready for review @mxinden @jxs.

core/src/either.rs
```diff
@@ -194,32 +57,54 @@ where
...
+            future::Either::Right(inner) => inner.poll(cx).map_err(Either::Right),
+        }
+    }
+}
+
+/// Convert `Pin<&mut Either<A, B>>` to `Either<Pin<&mut A>, Pin<&mut B>>`,
+/// pinned projections of the inner variants.
+///
+/// Local function until <https://github.com/rust-lang/futures-rs/pull/2691> is merged.
+fn as_pin_mut<A, B>(
```
jxs, line 94: Should we ask for [project](https://docs.rs/futures-util/0.3.25/src/futures_util/future/either.rs.html#36) to be public? As it seems the same purpose and would allow us to not have unsafe code here.

update:
sorry only now saw https://github.com/bluss/either/issues/76, can't we use [`as_pin_mut` from `either::Either`](https://docs.rs/either/latest/either/enum.Either.html#method.as_pin_mut)?

jxs (approved):
LGTM Thanks Thomas. The unsafe code seems fine, but if possible I'd rather have it upstream.

core/src/either.rs
thomaseizinger: > Should we ask for [project](https://docs.rs/futures-util/0.3.25/src/futures_util/future/either.rs.html#36) to be public? As it seems the same purpose and would allow us to not have unsafe code here.

See https://github.com/rust-lang/futures-rs/issues/2629#issuecomment-1196338344 :)

This is the `future::Either` type, we can't use `either::Either` here because it doesn't implement `AsyncRead` etc

thomaseizinger:
> LGTM Thanks Thomas. The unsafe code seems fine, but if possible I'd rather have it upstream.

Yeah me too. It will merge eventually and it is a private function so I don't think it is blocking :)

core/src/either.rs
jxs: yeah I saw after, thanks Thomas :)  
It it would be nice for `either::Either` to implement the `futures` traits as you asked on https://github.com/bluss/either/issues/79. But it seems you also submitted https://github.com/rust-lang/futures-rs/pull/2691 which will allow us to remove the `unsafe` code right? Ideally what would be really nice was for a single `Either` type :grinning: which will be possible when https://github.com/rust-lang/futures-rs/pull/2691 is merged right? Nonetheless thanks for your work simplifying this Thomas!

thomaseizinger: > yeah I saw after, thanks Thomas :)  
> It it would be nice for `either::Either` to implement the `futures` traits as you asked on https://github.com/bluss/either/issues/79. But it seems you also submitted https://github.com/rust-lang/futures-rs/pull/2691 which will allow us to remove the `unsafe` code right?

Correct!

> Ideally what would be really nice was for a single `Either` type :grinning: which will be possible when https://github.com/rust-lang/futures-rs/pull/2691 is merged right?

Not quite unfortunately, we also use `Either` in places where it needs to implement `std::error::Error` for example and `futures::Either` doesn't AFAIK. I agree that it would be nice but at least after these patches, we don't have a problem with there being two different `Either` types.

> Nonetheless thanks for your work simplifying this Thomas!

You are welcome ☺️

jxs: wdyt of then leaving a `TODO` on the unsafe code to be removed when a `futures` gets released with https://github.com/rust-lang/futures-rs/pull/2691 ?

thomaseizinger: > wdyt of then leaving a `TODO` on the unsafe code to be removed when a `futures` gets released with https://github.com/rust-lang/futures-rs/pull/2691 ? 

Do you mean like line 93? 🙈

jxs: ah! right LOL :joy:

thomaseizinger:
Merging here to move forward with #3254.

thomaseizinger:
@mergifyio requeue

thomaseizinger:
@mergifyio queue

PR #3342, opened 2023-01-17
feat(quic): Wake transport when adding a new dialer or listener

elenaf9:
## Description

Wake `quic::GenTransport` if a new dialer or listener is added.

## Notes

This re-adds #3306 with the solution initially drafted by @mxinden for waking the transport if a new dialer is added. 
I added to also wake the transport if a new listener was pushed. The `SelectAll` around the listeners itself doesn't wake the task if a new element is pushed. This is not a major problem since the user would call `poll` again after a `listen_on` call and thus registering a new waker, but it solves the edge cast of a user doing something like this:
```rust
match quic_transport.poll_next(cx) {
   Poll::Ready(event) => return event,
   Poll::Pending => {}
}
quic_transport.listen_on(addr);
return Poll::Pending
```

## Change checklist

- [ ] <s>.A changelog entry has been made in the appropriate crates</s>

### DIFF
--- a/transports/quic/src/transport.rs
+++ b/transports/quic/src/transport.rs
@@ -71,7 +71,8 @@ pub struct GenTransport<P: Provider> {
     listeners: SelectAll<Listener<P>>,
     /// Dialer for each socket family if no matching listener exists.
     dialer: HashMap<SocketFamily, Dialer>,
-    dialer_waker: Option<Waker>,
+    /// Waker to poll the transport again when a new dialer or listener is added.
+    waker: Option<Waker>,
 }
 
 impl<P: Provider> GenTransport<P> {
@@ -85,7 +86,7 @@ impl<P: Provider> GenTransport<P> {
             quinn_config,
             handshake_timeout,
             dialer: HashMap::new(),
-            dialer_waker: None,
+            waker: None,
             support_draft_29,
         }
     }
@@ -109,6 +110,10 @@ impl<P: Provider> Transport for GenTransport<P> {
             version,
         )?;
         self.listeners.push(listener);
+
+        if let Some(waker) = self.waker.take() {
+            waker.wake();
+        }
 
         // Remove dialer endpoint so that the endpoint is dropped once the last
         // connection that uses it is closed.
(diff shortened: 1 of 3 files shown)


transports/quic/src/transport.rs
```diff
@@ -202,6 +212,8 @@ impl<P: Provider> Transport for GenTransport<P> {
                 errored.push(*key);
             }
         }
+        self.waker = Some(cx.waker().clone());
```
thomaseizinger: Can we pair this with the `Poll::Pending` below?

transports/quic/tests/smoke.rs
```diff
@@ -90,6 +97,113 @@ async fn ipv4_dial_ipv6() {
...
+
+    // Spawn B
+    //
+    // Note that the dial is spawned on a different task than the transport allowing the transport
+    // task to poll the transport once and then suspend, waiting for the wakeup from the dial.
+    let dial = async_std::task::spawn({
+        let dial = b_transport.dial(a_addr).unwrap();
+        async { dial.await.unwrap().0 }
+    });
+    async_std::task::spawn(async move { b_transport.next().await });
```
thomaseizinger, line 198: Nit: Ideally this would be a loop but it doesn't matter really :)

thomaseizinger (approved):

transports/quic/src/transport.rs
mxinden: I am in favor of the above suggestion. Something along the lines of:


``` diff
-       self.waker = Some(cx.waker().clone());


        for key in errored {
            // Endpoint driver of dialer crashed.
            // Drop dialer and all pending dials so that the connection receiver is notified.
            self.dialer.remove(&key);
        }
-       match self.listeners.poll_next_unpin(cx) {
-           Poll::Ready(Some(ev)) => Poll::Ready(ev),
-            _ => Poll::Pending,
-       }
+        if let Poll::Ready(Some(ev)) = self.listeners.poll_next_unpin(cx) {
+           return Poll::Ready(ev);
+       }

+       self.waker = Some(cx.waker().clone());
+       Poll::Pending
```

mxinden (approved):
Change looks good to me, minus the small suggestion below.

Thanks for putting more thought into this @elenaf9!

mxinden:
Also can you add a short changelog entry? I forgot to add one in https://github.com/libp2p/rust-libp2p/pull/3306 unfortunately.

transports/quic/tests/smoke.rs
elenaf9: Since `b_transport` is neither listening nor receiving an inbound dial `next` will never return, will it?

transports/quic/src/transport.rs
elenaf9: Done in 7cdaee7.

transports/quic/tests/smoke.rs
thomaseizinger: > Since `b_transport` is neither listening nor receiving an inbound dial `next` will never return, will it?

Fair point!

Issue #3343, opened 2023-01-17
QUIC: handle the case of both peers half-closing a stream


elenaf9:
## Summary

If for a QUIC stream both peers agree at the same time that no more data should be send in one direction, the one side tries to close the stream (send a `FIN`), while the other sends a `STOP_SENDING`.  

Receiving a `STOP_SENDING` before / while closing a stream right now causes the `close` call to return an `Error::Stopped`.
However, in practice in the case that all of our write data was received and acknowledged by the remote, this should not be an error. Instead both peers simply reached the agreement that the stream is closed in this direction, and no data was lost.

The current behaviour is causing issues like #3298 and #3281.

## Expected behaviour

If all write data has been received (and ack'ed) by the remote peer closing a stream should not fail.

## Actual behaviour

`Substream::close` always returns an error if the stream is stopped by the peer before the `FIN` frame is acknowledged.

## Possible Solution

We have been discussion different solutions for this out of band in #3302.
One option would be to simply treat a stopped send-stream as successfully closed. Typically a remote would only send a `STOP_SENDING` if it read all of our data.
However, if something unexpected happened and the remote stopped the stream (e.g. because it dropped the stream) before reading our latest data, the error would be lost. Our user would in this case wrongly assume that the remote read the data, when it actually didn't.

Another option would be to check if we still have unacknowledged data on the stream, and depending on that return `Ok`/`Err`. The `quinn-proto` API currently doesn't give us this information, so we'd have to add an upstream patch.  
Side note: the high-level `quinn` implementation handles this case the same way as we do it, i.e. return an `Error` in the close call if the stream stopped. Given that we long-term consider switching to `quinn`, this is also something we should take into consideration.

I still need to think more about this. Input is appreciated.
There are easy fixes for #3298 #3281 on the application layer for which I'll do PRs, but long-term we should fix this within `libp2p-quic`. 

// cc @marten-seemann 

## Version

- libp2p-quic `0.7.0-alpha.2`

## Would you like to work on fixing this bug?

Yes

thomaseizinger:
> Another option would be to check if we still have unacknowledged data on the stream, and depending on that return `Ok`/`Err`. The `quinn-proto` API currently doesn't give us this information, so we'd have to add an upstream patch.

I think this is the way to go.

> Side note: the high-level `quinn` implementation handles this case the same way as we do it, i.e. return an `Error` in the close call if the stream stopped.

[...]

PR #3344, opened 2023-01-17
fix(identify): don't close stream in `protocol::recv`

elenaf9:
## Description

Don't close the stream `protocol::recv`.

This is a short-term fix for #3298. 

The issue behind this is a general one on the QUIC transport when closing streams, as described in #3343. This PR only circumvents the issue for identify. A proper solution for our QUIC transport still needs more thought.

- Issue was reported in #3298 
- Issue for the underlying problem in QUIC is #3343.

## Change checklist


protocols/identify/src/protocol.rs
```diff
@@ -193,7 +193,7 @@ async fn recv<T>(mut socket: T) -> Result<Info, UpgradeError>
 where
     T: AsyncRead + AsyncWrite + Unpin,
 {
-    socket.close().await?;
+    socket.flush().await?;
```
mxinden: Is flushing needed here? Would removing the whole line be enough?

See https://github.com/libp2p/rust-libp2p/issues/3298#issuecomment-1369055550 for more details.

mxinden (commented):
Thanks for the fix. Generally in favor even though it is only a hot-fix for an issue rooted somewhere else.

protocols/identify/src/protocol.rs
thomaseizinger: Does it work if we ignore the error on the close? Would that be a viable hot-fix too?

Plus, I'd like us to have a comment here linking to an issue on GitHub tracking the overall problem.

mxinden: > Does it work if we ignore the error on the close? Would that be a viable hot-fix too?

Preference for removing the close as it is not needed in the first place. Ignoring an error might hinder debugging in the future.

> Plus, I'd like us to have a comment here linking to an issue on GitHub tracking the overall problem.

Yes please. Good idea.

mxinden: @elenaf9 friendly ping.

elenaf9: After looking into this a bit more I think closing a stream here is / will remain problematic, independently of #3343. 
The reason is that we also use `recv` in the inbound upgrade on `Push`. If this is combined with using upgrade `Version::V1Lazy` I think the following can happen in theory:
1. The dialer opens a new outbound identify-push stream
2. With the `V1Lazy` it directly flushes the identify payload together with the protocol-id
3. It drops the stream once the write is done
4. The listener reads the protocol-id and echoes it back before starting the inbound upgrade
5. the remote never reads the echoed-back protocol-id, thus closing the stream fails

In go-libp2p this issue is solved by simply ignoring the error in both, the multi-stream-select upgrade[^1] and when reading the identify[^2] response.

I am not too familiar with multi-stream select in rust-libp2p so it could be that in the above case the dialer would not drop the stream until it received the echoed-back protocol-id. That said, even if it's implemented like that in rust-libp2p, I assume from the linked go-libp2p code that this is how they are doing it. So I think we also have to handle that case. // CC @MarcoPolo since we talked about this out-of-band - did I understand it correctly?

Am I missing something?

[^1]: https://github.com/multiformats/go-multistream/blob/5555a5c43be95c669ce5637723ccf690a6a30e23/multistream.go#L203-L205
[^2]: https://github.com/libp2p/go-libp2p/blob/6b9c11680e0bb412ef7515f266531c23245015dc/p2p/protocol/identify/id.go#L435

mxinden: > the remote never reads the echoed-back protocol-id, thus closing the stream fails

I am not sure I understand the last step here. Note that closing a stream is a local operation only in TCP + Yamux. It only flushes the stream buffers to the local connection buffers. It does not ensure that the closing of the stream is communicated to the remote. Communicating the closing to the remote is asynchronous to calling `close`. The connection task will make sure the closing of the stream is communicated to the remote.

https://github.com/libp2p/rust-yamux/blob/72ccd5734f71971fc1f02ac4a5f43b8eb4dff660/yamux/src/connection/stream.rs#L365-L387

> I am not too familiar with multi-stream select in rust-libp2p so it could be that in the above case the dialer would not drop the stream until it received the echoed-back protocol-id. 

It should await receiving the protocol-id-echo.

https://github.com/libp2p/rust-libp2p/blob/ab59af4d46e0f6835429204f622475adcb6b74df/misc/multistream-select/src/negotiated.rs#L308-L310

Never the less, I am advocating for removing the `close` in `recv`. Any objections @elenaf9?

elenaf9: >>  the remote never reads the echoed-back protocol-id, thus closing the stream fails
>
> I am not sure I understand the last step here. Note that closing a stream is a local operation only in TCP + Yamux. It only flushes the stream buffers to the local connection buffers. It does not ensure that the closing of the stream is communicated to the remote. Communicating the closing to the remote is asynchronous to calling close. The connection task will make sure the closing of the stream is communicated to the remote.

Sorry, I was probably not clear with what I meant; I was referring to @thomaseizinger suggestion to add a comment here that links to #3343. I was arguing that **because** in QUIC (contrary to TCP+Yamux) the close is not local only but instead fails if not all data gets acknowledged, even with #3343 resolved `close` could cause issues in the described scenario. 

I was mostly wondering if my understanding was correct. If it is, I would add a comment that describes that. 

> Never the less, I am advocating for removing the `close` in `recv`. Any objections @elenaf9?

No objections; I will remove it.

Edit: What do you think of cf21edba36c8faa87b6de09284fbc596d33c162f? (sorry for the broken commit message - forgot to escape the backticks in the message)

protocols/identify/CHANGELOG.md
```diff
@@ -7,7 +7,10 @@
 
 - Update to `libp2p-swarm` `v0.42.0`.
 
+- Flush stream in `protocol::recv` instead of closing it. See [PR 3344].
```
elenaf9:
```suggestion
- Don't close the stream when reading the identify info in `protocol::recv`. See [PR 3344].
```

protocols/identify/src/protocol.rs
elenaf9: > Edit: What do you think of https://github.com/libp2p/rust-libp2p/commit/cf21edba36c8faa87b6de09284fbc596d33c162f? 

Friendly ping @mxinden @thomaseizinger.

thomaseizinger: I am okay with the comment. At the end of the day, this is a temporary situation until we fix `quinn` (+ `quinn-proto`), right?

elenaf9: Not sure. That's what I meant above: even with the fix, it could cause errors in case of identify `Push` + `V1Lazy` if the remote doesn't read the echo-ed back protocol-id. Not an issue if both peers are rust-libp2p, but I am not sure how go-libp2p implements it.

mxinden (approved):
With the latest changes, this looks good to me.

Thanks for thinking this through @elenaf9. Thanks for documenting further steps with https://github.com/libp2p/rust-libp2p/issues/3343.

mxinden:
@thomaseizinger I am assuming that you are not opposed to this patch. In case you are, please comment and I will revert.

protocols/identify/src/protocol.rs
```diff
@@ -189,11 +189,14 @@ where
...
-async fn recv<T>(mut socket: T) -> Result<Info, UpgradeError>
+async fn recv<T>(socket: T) -> Result<Info, UpgradeError>
 where
     T: AsyncRead + AsyncWrite + Unpin,
 {
-    socket.close().await?;
+    // Even though we won't write to the stream anymore we don't close it here.
+    // The reason for this is that the `close` call on some transport's require the
+    // remote's ACK, but it could be that the remote already dropped the stream
+    // after finishing their write.
```
mxinden:
```suggestion
    // after finishing their write.
```

mxinden:
Interoperability tests failing due to https://github.com/libp2p/test-plans/pull/121#pullrequestreview-1292957819.

PR #3350, opened 2023-01-19
feat: introduce `libp2p-identity` crate

### LINKED ISSUE #3349: Extract `libp2p-identity` crate
This issue is part of https://github.com/libp2p/rust-libp2p/issues/3271.

- These modules are described in a single spec: https://github.com/libp2p/specs/blob/master/peer-ids/peer-ids.md
- It will be useful to improve the `multiaddr` crate: https://github.com/multiformats/rust-multiaddr/issues/73
- This crate will be a dependency of pretty much the entire ecosystem. It should be heavily feature-flagged and its API should try and avoid to leak any dependencies.
- We can consider moving `PeerRecord` and `SignedEnvelope` into this crate as well or extract them into their own. I am undecided here.

thomaseizinger:
## Description

This patch combines the `libp2p_core::identity` and `libp2p_core::peer_id` modules into a new crate: `libp2p-identity`.

Resolves https://github.com/libp2p/rust-libp2p/issues/3349.

## Notes

This initial patch only moves the code and adds feature flags. What is yet undecided is how we want to adjust the public API as part of this. The goal would be that the first release (0.1) is backwards compatible with what we currently ship. In a 2nd release, we then harden the public API such that breaking changes will be very unlikely in the future. As part of `0.2`, I also want to remove the dependency on `multiaddr` to enable https://github.com/multiformats/rust-multiaddr/issues/73 which requires `multiaddr` to depend on `libp2p-identity`.

What do people think about this plan? @mxinden @jxs @elenaf9 

## Further tasks

- [ ] Seal public API by making `Keypair` and `PublicKey` opaque structs. I think we should **not** have conditional enum variants. Those are a hazard. Someone could be exhaustively matching on them with one feature combination and adding a dependency that activates another suddenly causes the program to fail to compile. Even worse, this can happen in dependencies that don't add the necessary `cfg`s.

## Change checklist


jxs:
Makes sense to me Thomas :+1:

core/Cargo.toml
```diff
@@ -11,61 +11,44 @@ keywords = ["peer-to-peer", "libp2p", "networking"]
...
-rsa = [ "dep:ring" ]
-serde = ["multihash/serde-codec", "dep:serde"]
-
-[[bench]]
-name = "peer_id"
-harness = false
+secp256k1 = [ "libp2p-identity/secp256k1" ]
+ecdsa = [ "libp2p-identity/ecdsa" ]
+rsa = [ "libp2p-identity/rsa" ]
+ed25519 = [ "libp2p-identity/ed25519" ]
```
thomaseizinger: These are all deprecated and will be removed in the future.

identity/src/keypair.rs
```diff
@@ -67,42 +55,100 @@ use std::convert::{TryFrom, TryInto};
 #[allow(clippy::large_enum_variant)]
 pub enum Keypair {
     /// An Ed25519 keypair.
+    #[cfg(feature = "ed25519")]
+    #[deprecated(
+        since = "0.1.0",
+        note = "This enum will be made opaque in the future, use `Keypair::into_ed25519` instead."
+    )]
```
thomaseizinger, line 63: The plan with this is that once users have migrated to the `into_` functions, we can make `Keypair` a struct and have the enum as a private field with `cfg`'d variants. That moves all the code that touches the conditional variants under our control where we can assure that it works under all feature-flag combinations. It will also allow us to remove the `keypair_dummy.rs` file again.

thomaseizinger:
@mxinden Would appreciate your input on this.

core/Cargo.toml
mxinden: Just to make sure we don't miss it, this is a breaking change. Previously `libp2p-core` would offer `ed25519` by default.

mxinden (commented):
I am sorry for the delay here.

I am in favor of the direction this is going.

Do I understand correctly that the main motivation for this change is to make `multiaddr`'s `/p2p` type safe by using `PeerId` from `libp2p-identity?

The proposals around hardening the API sound good to me in words. That said, I would need to see them in code and thus see their implications to really judge.

In case we go for the breaking change in `libp2p-core`, I am happy to push the version bumps right before merging.

core/Cargo.toml
thomaseizinger: Right, I would like it to not be a breaking change so I'll set this as a default feature, nice catch!

thomaseizinger:
> In case we go for the breaking change in `libp2p-core`, I am happy to push the version bumps right before merging.

As it currently stands, this PR should be fully backwards-compatible because we re-export everything.

> The proposals around hardening the API sound good to me in words. That said, I would need to see them in code and thus see their implications to really judge.

Think of it like this:

```rust
pub struct Keypair(InnerKeypair);

enum InnerKeypair {
	Ed25519(...),
	Secp256k1(...),
	...
}
```

Basically, we are making it an opaque type. There shouldn't really be much of an implication for users. Previously, they also didn't have a static guarantee, what keypair they were given and it therefore required branching. The same is true after these changes, just that instead of matching on an enum, they have to call one of `into_xyz` functions now and branch on that.

> Do I understand correctly that the main motivation for this change is to make `multiaddr`'s `/p2p` type safe by using `PeerId` from `libp2p-identity?

Partly yes. That is the first place where we will be able to benefit from this.

The other motivation is splitting `libp2p-core` into more meaningful modules as discussed in https://github.com/libp2p/rust-libp2p/discussions/3072#discussioncomment-4394785. This should in the long-run stop the cascade of breaking changes. For example, a change in the transport API should not affect a protocol crate but if `Transport`and `PeerId` are in the same crate, it does.

core/Cargo.toml
thomaseizinger: I removed the feature from `libp2p-core` as it would be a breaking change either way because users could depend on it with `default-features = false`.

identity/src/peer_id.rs
```diff
@@ -94,6 +95,10 @@ impl PeerId {
     ///
     /// In case the given [`Multiaddr`] ends with `/p2p/<peer-id>`, this function
     /// will return the encapsulated [`PeerId`], otherwise it will return `None`.
+    #[deprecated(
+        since = "0.1.0",
+        note = "`Multiaddr` will become type-safe in a future release, making this function obsolete. See https://github.com/multiformats/rust-multiaddr/issues/73."
+    )]
```
thomaseizinger: Does it make sense to have this deprecation warning? At this stage, there isn't anything users can do about it because https://github.com/multiformats/rust-multiaddr/issues/73 isn't implemented yet. But at the same time, I want to tell users that this will go away. Thoughts?

core/src/lib.rs
```diff
@@ -55,25 +50,78 @@ mod peer_record_proto {
 pub use multiaddr;
 pub type Negotiated<T> = multistream_select::Negotiated<T>;
 
-mod peer_id;
+#[deprecated(since = "0.39.0", note = "Depend on `libp2p-identity` instead.")]
+pub mod identity {
+    pub use libp2p_identity::Keypair;
+    pub use libp2p_identity::PublicKey;
```
thomaseizinger, line 56: @obi1kenobi I think we might have hit another bug here! `cargo semver-checks` is reporting these two as missing. I've tried with type aliases as well and it doesn't make a difference which makes me think that I am doing something wrong here? It did work in my testing though, i.e. I got heaps of deprecation warnings but stuff was still working when I developed this feature.

obi1kenobi: This is a move to another crate + re-export, right? Semver-checking right now only generates the tested crate's rustdoc JSON, so it doesn't see cross-crate re-exports: https://github.com/obi1kenobi/cargo-semver-checks/issues/355

It's fixable issue with a clear path forward. Just a matter of engineering cycles, which are always in short supply.

Thankfully, once you publish the next version of the affected crate, the false-positive will go away since the baseline will shift to that new version.

identity/src/peer_id.rs
mxinden: I don't think it makes sense to advertise features through deprecation warnings before they are in place. I advocate for removing the deprecation warning.

core/Cargo.toml
mxinden: Sounds good to me. Thanks for the change.

mxinden (approved):
Change looks good to me minus the deprecation warning.

thomaseizinger:
> Change looks good to me minus the deprecation warning.

I've removed the deprecation warning.

Until we release these changes, we will have a broken CI because of https://github.com/libp2p/rust-libp2p/pull/3350#discussion_r1121018717. Can you exercise your admin rights to merge this and make patch releases of all crates?

This PR attracts a lot of merge conflicts so I'd rather not keep it updated for a long time.

thomaseizinger:
> > Change looks good to me minus the deprecation warning.
> 
> I've removed the deprecation warning.
> 
> Until we release these changes, we will have a broken CI because of [#3350 (comment)](https://github.com/libp2p/rust-libp2p/pull/3350#discussion_r1121018717). Can you exercise your admin rights to merge this and make patch releases of all crates?
> 
> This PR attracts a lot of merge conflicts so I'd rather not keep it updated for a long time.

Friendly ping @mxinden.

mxinden:
Update here. I won't get to releasing this today, thus not yet merging. Otherwise other pull requests are red in the meantime.

I hope to get to this Saturday or Monday.

@thomaseizinger no need to resolve merge conflicts in the meantime. Sorry for the delay.

mxinden:
As discussed, semver check failure for `libp2p-core` is a false positive. Thus merging here. Will start the release process next.

thomaseizinger:
Thank you for releasing this!

As part of the next breaking change, I want to remove the `multiaddr` dependency and clean up the API of `libp2p-identity`.

drHuangMHT:
This kind of breaks my own implementation lol. I tried to keep `PeerId` and `Keypair` in the same place to unify the use. But opaque `Keypair` seems to make it harder to export keys? Because you cannot know the algorithm used in the keypair at a glance and I have to try calling those methods to see what it is, which is not so ergonomic. If you don't know that the algorithm is, it will be harder to import keys.

thomaseizinger:
> This kind of breaks my own implementation lol. I tried to keep `PeerId` and `Keypair` in the same place to unify the use. But opaque `Keypair` seems to make it harder to export keys? Because you cannot know the algorithm used in the keypair at a glance and I have to try calling those methods to see what it is, which is not so ergonomic. If you don't know that the algorithm is, it will be harder to import keys.

I am happy to add features to make exporting easier to use. I am not sure why our `to_protobuf_encoding` is actually unsupported for anything other than ed25519 (cc @mxinden ).

Getting the conditional compilation code right is extremely tricky with the current situation and I don't want to put that burden on any of our users hence the deprecation of the direct access to the enum. If you get that wrong and your code happens to be another library, you can easily produce unresolvable compile-errors for the users of your library.

We could add a `KeypairKind` enum that allows you to branch accordingly. Would that be an acceptable solution?

thomaseizinger:
The [specs](https://github.com/libp2p/specs/blob/master/peer-ids/peer-ids.md) don't mention anything about key types being deprecated so I think we could add support for encoding the older key types via protobuf.

@mxinden may have more context here on why we don't currently support them.

What is your usecase for using non-ed25519 keys? Those should certainly be the preferred ones.

drHuangMHT:
> What is your usecase for using non-ed25519 keys? Those should certainly be the preferred ones.  

That actually doesn't matter. I'll open a discussion and see what I can do there.

mxinden:
> The [specs](https://github.com/libp2p/specs/blob/master/peer-ids/peer-ids.md) don't mention anything about key types being deprecated so I think we could add support for encoding the older key types via protobuf.
> 
> @mxinden may have more context here on why we don't currently support them.

I introduced the protobuf functionality in https://github.com/libp2p/rust-libp2p/pull/2090 in order to import IPFS configs in https://github.com/mxinden/rust-libp2p-server/. I only implemented ed25519 because I didn't need any other. I am not aware of any other reason why the others are not implemented. In other words, contributions welcome.

thomaseizinger:
I opened an issue here: https://github.com/libp2p/rust-libp2p/issues/3630

Issue #2719, opened 2022-06-21
swarm/src/behaviour: Add `AndBehaviour`


mxinden:
## Description

To combine two `NetworkBehaviour` implementations `libp2p-swarm` should provide a combinator. Contrary to [`NetworkBehaviour` implementation of `Either`](https://github.com/libp2p/rust-libp2p/blob/master/swarm/src/behaviour/either.rs) both `NetworkBehaviour` implementations would be running concurrently.

## Motivation

One use-case is when wanting to either enable two `NetworkBehaviour`s or disable both, but never enable one but disable the other.

`Toggle<And<BehaviourA, BehaviourB>>`

Came up here: https://github.com/n0-computer/iroh/pull/126#discussion_r901236064

## Open questions

- What is the ideal name? Off the top of my head it should be `And`.

## Are you planning to do it yourself in a pull request?

No. Though happy to support anyone interested in contributing this.

GlenDC:
Is it ok for me @mxinden to pick up this one, or do you have another one in mind that you want me to look at?

mxinden:
Please prioritize any WebRTC and Testground related work. That said, yes, help here is very much appreciated whenever you are blocked on WebRTC and Testground. Let us know in case you need any help @GlenDC.

GlenDC:
Understood. I am going to continue to give all my love and energy in the Protocol Labs context to Testground. Let's do it!

thomaseizinger:
I think we agreed to not do this in #3366.

PR #3366, opened 2023-01-21
feat(swarm): Add AndBehaviour

jxs:
## Description

Addresses #2719. 

## Notes
 I started this by basically following the code generated by a `NetworkBehaviour`  derive implementation for a struct similar to `And` (but without generic fields) but then it seemed a lot of similar code to the generated macro to maintain.

I then followed to the approach submitted on this PR, have the `And` struct  itself the `NetworkBehaviour` derived. While going for this approach I found a bug with the `NetworkBehaviour` derive implementation for generics, when `out_event` was not provided the enum generated missed the `NetworkBehaviour` impl constraint for the generic variants whilst using the generics for `<Generic>::OutEvent`. This is what a4570af94c34c8d9e17d3f56ecdbda80e0b04d44 addresses. 

The manual `Debug` implementation was required because the `Debug` derive is not smart enough, when we annotate it on a struct with generics it requires it's generics to impl `Debug` but on our case what we want is that only the `OutEvent` associated type of the generic variants implement `Debug`.  see similar [here](https://stackoverflow.com/questions/72392583/how-to-ignore-generic-argument-for-derivedebug) or [here](https://github.com/rust-lang/rust/issues/26925) 

The derive macro shouldn't be breaking, from what I tested the printed debug output is the same for non generic types. 
I updated the ping example to use `And`, if agreed will update the other examples.
If this seems to hacky I can rollback and submit instead the first approach.
## Change checklist


thomaseizinger:
Hmm, I think the fix is worth landing (maybe even in a separate PR?).

If `And` is this trivial to implement, I don't think we need to provide it? Not sure. It might be a nice utility to have. Perhaps we should be implementing `NetworkBehaviour` for tuples?

jxs:
Good question!
IIUC you are suggesting tupples so we can have more than two elements right? 
AFACT to implement `NetworkBehaviour` for tuples it seems we would need to do it manually, (see [here](https://doc.rust-lang.org/1.5.0/std/primitive.tuple.html#implementations)  or [here](https://github.com/rust-lang/rust/blob/master/library/core/src/tuple.rs) how std does it for its traits) instead of deriving.

So it seems a matter of choosing where to have more complexity, either in the code (manually implementation) or in the user API, `<And<A, And<B, C>>` which would mean `and.second_as_ref().first_as_ref()` to access it's behaviours, and `AndEvent(AndEvent(Behaviour)` vs `tuple.2` and `TupleEvent(Behaviour)`.
No strong opinion on my end Thomas.

thomaseizinger:
My vote would go against providing an `And` type. The implementation is so trivial, users can just create it themselves when they need it, esp. if they need to compose more than two behaviours.

swarm-derive/CHANGELOG.md
```diff
@@ -1,9 +1,14 @@
 # 0.32.0 [unreleased]
 
+- Fix `NetworkBehaviour` Derive macro for generic types when `out_event` was not provided. Previously The enum generated
```
mxinden:
```suggestion
- Fix `NetworkBehaviour` Derive macro for generic types when `out_event` was not provided. Previously the enum generated
```

swarm/CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # 0.42.0 [unreleased]
 
+- Add `And` struct that combines two `NetworkBehaviour` implementations. Helfpful for when one wants
```
```suggestion
- Add `And` struct that combines two `NetworkBehaviour` implementations. Helpful for when one wants
```

swarm/src/behaviour.rs
```diff
@@ -40,6 +41,47 @@ pub(crate) type THandlerInEvent<THandler> =
...
     <<THandler as IntoConnectionHandler>::Handler as ConnectionHandler>::OutEvent;
 
+/// Implementation of [`NetworkBehaviour`] that combines two underlying implementations.
+#[cfg(feature = "macros")]
+#[derive(Debug, Default, libp2p_swarm_derive::NetworkBehaviour)]
+#[behaviour(prelude = "crate::derive_prelude")]
+pub struct And<A, B> {
+    first: A,
+    second: B,
+}
```
mxinden: `NetworkBehaviour` combinators like `Toggle` and `Either` are in their own file. I would suggest doing the same for `And`.

mxinden (commented):
I am still in favor of adding an `And` `NetworkBehaviour`, even though it is very simple. Not a strong opinion.

Can you please extract the `swarm-derive` fix into a separate pull request @jxs?

thomaseizinger:
> I am still in favor of adding an `And` `NetworkBehaviour`, even though it is very simple. Not a strong opinion.
> 
> Can you please extract the `swarm-derive` fix into a separate pull request @jxs?

If we add it, I think the fields should just be public to make access easier.

swarm/src/behaviour.rs
```diff
@@ -40,6 +41,47 @@ pub(crate) type THandlerInEvent<THandler> =
 pub(crate) type THandlerOutEvent<THandler> =
     <<THandler as IntoConnectionHandler>::Handler as ConnectionHandler>::OutEvent;
 
+/// Implementation of [`NetworkBehaviour`] that combines two underlying implementations.
+#[cfg(feature = "macros")]
+#[derive(Debug, Default, libp2p_swarm_derive::NetworkBehaviour)]
```
elenaf9: Right now this will have the macro-generated `OutEvent` looking kinda like this: 
```rust
enum AndEvent<A, B> {
    First(A)
    Second(B)
}
```
What do you think about specifying `Either<A, B>` as out-event instead? Would also be more consistent with all of our other "combinators".

mxinden: Good point. Note @jxs that you can still achieve the above suggestion with the derive macro by using the `out_event` derive parameter, see https://docs.rs/libp2p-swarm/latest/libp2p_swarm/behaviour/trait.NetworkBehaviour.html#custom-networkbehaviour-with-the-derive-macro.

swarm-derive/CHANGELOG.md
jxs: Thanks Max, addressed in #3393

swarm/CHANGELOG.md
jxs: thanks! updated!

swarm/src/behaviour.rs
jxs: Thanks, Addressed!

swarm/src/behaviour.rs
jxs: Yeah I tried that while experimenting, but when you specify a custom `out_event` one has to impl `From` for each subvariant. In case of using `Either` the generated code is the following:
```Rust
    impl<A, B> crate::derive_prelude::NetworkBehaviour for And<A, B>
    where
        A: crate::derive_prelude::NetworkBehaviour,
        B: crate::derive_prelude::NetworkBehaviour,
        Either<A, B>: From<<A as crate::derive_prelude::NetworkBehaviour>::OutEvent>,
        Either<A, B>: From<<B as crate::derive_prelude::NetworkBehaviour>::OutEvent>,
```
which kinda doesn't make sense as it would require for us to know `A` and `B`, and means that specifying `out_event` for a derived `NetworkBehaviour` with generics is not supported now.  Thanks for referring that Elena. I am kind of out of ideas to allow `derive` to support custom `out_event` for generic structures. Do you ave any suggestion? Else I can add this to the doc.

jxs:
Opened https://github.com/libp2p/rust-libp2p/pull/3393 

> > I am still in favor of adding an `And` `NetworkBehaviour`, even though it is very simple. Not a strong opinion.
> > Can you please extract the `swarm-derive` fix into a separate pull request @jxs?
> 
> If we add it, I think the fields should just be public to make access easier.

``

> > I am still in favor of adding an `And` `NetworkBehaviour`, even though it is very simple. Not a strong opinion.
> > Can you please extract the `swarm-derive` fix into a separate pull request @jxs?
> 
> If we add it, I think the fields should just be public to make access easier.

Yeah, it's the dilemma regarding being allowed to make changes without breaking API easily vs easier to access API, funny enough I thought you'd go for the first :grin: Updated!

jxs:
meanwhile opened #3393 but forgot to open it from a libp2p/rust-libp2p repo so we could then make this one target it sorry. After that one gets merged this one will be updated.

swarm/src/behaviour.rs
elenaf9: Ahh yes you're right. Using `Either` as out-event would only be possible if we manually derive `NetworkBehaviour`. I don't feel strongly about it, thus I'm fine with sticking with the current solution.

> I am kind of out of ideas to allow derive to support custom out_event for generic structures. Do you ave any suggestion?

I don't think its possible. We will always run into the issue of conflicting implementations if we try to auto-implement `From`, since A and B theoretically may have the same type.

thomaseizinger: I think this is a point for not providing an `And` behaviour. If a user wants to group a set of `NetworkBehaviour`s together, they can do so themselves and provide a more meaningful name than `And`. Like, if they combine certain behaviours together than are only available to under a "legacy" feature-toggle, they can call this behaviour `Legacy` and the generated name will make more sense.

Plus, they can add as many behaviours as they want, not just two, they can decide on the visibility, they can decide on the names etc.

I really don't see the value in providing an `And` behaviour.

thomaseizinger (changes requested):
I am against adding this. See discussion: https://github.com/libp2p/rust-libp2p/pull/3366#discussion_r1091412910.

PR #3368, opened 2023-01-23
refactor: expose and use `THandlerOutEvent` type alias

thomaseizinger:
## Description

Previously, we used the full reference to the `OutEvent` of the `ConnectionHandler` in all implementations of `NetworkBehaviour`. Not only is this very verbose, it is also more brittle to changes. With the current implementation plan for #2824, we will be removing the `IntoConnectionHandler` abstraction. Using a type-alias to refer to the `OutEvent` makes the migration much easier.

## Notes

This will reduce the diff in https://github.com/libp2p/rust-libp2p/pull/3254.

## Change checklist


swarm/src/connection/pool.rs
```diff
@@ -297,7 +297,7 @@ pub enum PoolEvent<THandler: IntoConnectionHandler> {
         id: ConnectionId,
         peer_id: PeerId,
         /// The produced event.
-        event: THandlerOutEvent<THandler>,
+        event: <<THandler as IntoConnectionHandler>::Handler as ConnectionHandler>::OutEvent,
```
thomaseizinger, line 297: I temporarily inlined this one so that I can delete the 2nd type alias named `THandlerOutEvent` defined in `swarm::behaviour`.

This will be simplified to `ConnectionHandler::OutEvent` in #3254.

swarm/src/behaviour.rs
```diff
@@ -122,7 +119,7 @@ pub(crate) type THandlerOutEvent<THandler> =
 ///   }
 /// }
 /// ```
-pub trait NetworkBehaviour: 'static {
+pub trait NetworkBehaviour: Sized + 'static {
```
thomaseizinger: This requires a `Sized` bound now but I'd honestly not know why a `NetworkBehaviour`wouldn't be sized ...

mxinden: How is the requirement to be `Sized` related to this change?

thomaseizinger: I think it is because I am now using `Self` within a function signature via a type-alias. To be frank, I just followed the compiler :upside_down_face:

mxinden: I would prefer to avoid casually introducing trait bounds without careful consideration. Is there any way around introducing this trait bound? Is this pull request worth doing given the added complexity.

I would have to dig deeper here. Based on intuition, adding `Sized`, `Send` or `Sync` unintentionally has not proven to be good ideas in the past.

mxinden (commented):
Other than the introduction of `Sized`, this looks good to me.

swarm/src/behaviour.rs
thomaseizinger: Fair point! I never looked deeply into it but apparently, requiring `Sized` prohibits the use as a trait object: https://doc.rust-lang.org/std/marker/trait.Sized.html

I'll try to work around it.

thomaseizinger: Turns out it is not needed, lol :joy:

jxs (approved):
LGTM :rocket:

swarm/src/behaviour.rs
jxs: > Fair point! I never looked deeply into it but apparently, requiring `Sized` prohibits the use as a trait object:
> 
> I'll try to work around it.

yeah, that's why you can't have `Box<dyn Clone>` for example

mxinden: Thanks for investigating. Great to see it not being needed.

mxinden (approved):
Thanks for the follow-up.

PR #3372, opened 2023-01-24
refactor(relay): directly store actions instead of events

thomaseizinger:
## Description

Storing `NetworkBehaviourAction`s within the behaviour is more flexible than only storing `OutEvent`s. Additionally, I find expression-oriented code easier to reason about because it typically doesn't contain side-effects.

## Notes

This will reduce the diff in https://github.com/libp2p/rust-libp2p/pull/3328.

## Change checklist


jxs (approved):
Yeah agree with your reasoning, LGTM Thomas

mxinden (commented):
:+1:

PR #3375, opened 2023-01-24
feat(swarm)!: introduce `ListenError`

thomaseizinger:
## Description

In case an error happens for an outgoing connection, `Pool` reports an `OutgoingConnectionError`. This one is mapped to a `DialError` and reported via `SwarmEvent::OutgoingConnectionError` and `FromSwarm::DialFailure`.

For incoming connections, we didn't quite do the same thing. For one, `SwarmEvent::IncomingConnectionError` directly contained a `PendingInboundConnectionError`. Two, `FromSwarm::ListenFailure` did not include an error at all.

With this patch, we now introduce a `ListenError` enum which we use in `SwarmEvent::IncomingConnectionError` and we pass a reference to it along in `FromSwarm::ListenFailure`.

## Notes

In https://github.com/libp2p/rust-libp2p/pull/3254, I am going to introduce a `Denied` variant in both `DialError` and `ListenError` which contains the error that a `NetworkBehaviour` denied a connection with.

## Change checklist


swarm/src/lib.rs
```diff
@@ -1677,6 +1683,68 @@ impl error::Error for DialError {
     }
 }
 
+/// Possible errors when upgrading an inbound connection.
+#[derive(Debug)]
+pub enum ListenError {
```
mxinden, line 1692: Mentioning it here, though not directly related: I always confuse `LIstenError`, signaling a failure on upgrading an inbound connection, with the failure of listening on a new address. Anyone else?

mxinden (approved):
Change looks good to me.

swarm/src/lib.rs
thomaseizinger: A lot of naming could be improved here :)

I believe this is consistent with the naming of `DialError` but I agree that something like `InboundConnectionError` and `OutboundConnectionError` would probably be better.

I'll see to tidy up these when we remove the connection limits and banned variants.

PR #3376, opened 2023-01-24
fix(ci): apply shorter timeout to cache download

thomaseizinger:
## Description

Sometimes, GitHub Actions gets stuck in downloading the cache. My hypothesis is that because we have so many jobs, some of them get rate-limited and end up failing the download.

Typically, GitHub Actions downloads with a speed of 100MB/s. With the maximum allowed cache size of 10GB, any cache download should not take longer than 100 seconds. With this patch, we set the timeout to 120 seconds for the continuous integration workflow.

Upon failure, the job will continue with a cache-miss which is nice because it means the workflow will continue and not fail.

More information here: https://github.com/actions/cache/blob/main/tips-and-workarounds.md#cache-segment-restore-timeout

Example workflows:

- https://github.com/libp2p/rust-libp2p/actions/runs/3992478158/jobs/6848975863
- https://github.com/libp2p/rust-libp2p/actions/runs/3992492999/jobs/6848989008
- https://github.com/libp2p/rust-libp2p/actions/runs/3992564680/jobs/6849012006

## Change checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,6 +10,9 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
 
+env:
+  SEGMENT_DOWNLOAD_TIMEOUT_MINS: 2 # Fail cache download after 2 minutes.
+
 jobs:
   test:
     name: Test ${{ matrix.crate }}


jxs (approved):
LGTM

PR #3382, opened 2023-01-25
feat(inter-op): make `Dockerfile` generic over test binary

thomaseizinger:
## Description

We make the binary that should be copied into the container configurable via a build time `ARG`. This will allow us to reuse the same `Dockerfile` once we write more tests.
Additionally, this allows us to remove some of the directory switching and creation code. In order to not send the entire repository over to the docker daemon as a build context, we introduce a `.dockerignore` file that only allows select binary to be sent over.

The downside of this is that we need to extend this ignore file every time we add a new test. This shouldn't happen very often though and is easily discovered because the building of the docker container will fail.

## Change checklist

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,10 @@
+# This Docker ignore file aims to optimise the build process of the interop tests.
+# The Dockerfile in ./interop-tests/Dockerfile copies a locally built binary into the container and
+# thus does not need anything else to be present in the working directory.
+*
+
+# CI will build and copy over the release binary.
+target/release/ping
+
+# Locally, we might want to build it as `debug` binary to accelerate the build process.
+target/debug/ping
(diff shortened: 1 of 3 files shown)


jxs (approved):
I thought about this when first creating the workflow, but I'd rather have the directory creation and file copying on the workflow file (as it's all in the same place) than a `.dockerignore` that needs to be maintained in the main repo. Nonetheless not blocking.

thomaseizinger:
> I thought about this when first creating the workflow, but I'd rather have the directory creation and file copying on the workflow file (as it's all in the same place) than a `.dockerignore` that needs to be maintained in the main repo. Nonetheless not blocking.

Yeah I also thought along similar lines and then I started working on https://github.com/libp2p/rust-libp2p/pull/3383 which would have duplicated it all in another workflow.

Not sure if there is another solution?

jxs:
ah! I see, thanks for explaining Thomas! And feel free to merge :)

PR #3385, opened 2023-01-25
docs(websocket): showcase wrapping dns and tcp transport

### LINKED ISSUE #3330: Is secure WebSocket transport even working?
## Summary

So I'm currently debugging https://github.com/paritytech/substrate/issues/12024 and what I found is I don't really understand how secure WS transport (/wss) is supposed to work.

I give it an address like `/dns/polkadot-boot.dwellir.com/tcp/443/wss/p2p/<nodeID>` and expect it to connect. However, I get an error saying `invalid multi-address: /dns/polkadot-boot.dwellir.com/tcp/443/p2p/<nodeID>` address is invalid from inner TCP transport. Am I supposed to supply DNS as an inner transport if I want to support `/wss` addresses? Documentation is rather scarse.

## Expected behaviour

I'd expect for WSS transport to work kinda.

## Possible Solution

Use DNS transport as inner T when creating WSS transport.

## Version

0.50.0

Yes

melekes:
Closes #3330


melekes:
Re https://github.com/libp2p/rust-libp2p/issues/3330#issuecomment-1397799688 providing `Default` on the transport, we can do so for tokio, but how to do it for async-io since `dns::DnsConfig::system` returns future? Is there a async Default trait?

```rust
impl Default for WsConfig<T: Transport> {
    #[cfg(feature = "async-io")]
    fn default() -> Self {
        WsConfig {
            transport: dns::DnsConfig::system(
                tcp::async_io::Transport::new(tcp::Config::default()),
            )
            .await
            .unwrap(),
        }
    }

    #[cfg(feature = "tokio")]
    fn default() -> Self {
        WsConfig {
            transport: dns::TokioDnsConfig::system(
                tcp::tokio::Transport::new(tcp::Config::default()),
            )
        }
    }
}
```

cc @thomaseizinger

mxinden (dismissed):
Ready to merge from my end.

Thanks @melekes for making this easier for future users!

Will wait a bit to give @thomaseizinger a chance to reply as well.

thomaseizinger:
There is no async default so we can't do that unfortunately.

thomaseizinger (commented):
I would still be in favor of dedicated constructors that take the necessary inner configs:

```rust
impl WsConfig {
    async fn new_wss(dns_config: ..., tcp_config: ...) -> Self {}
    fn new_ws(tcp_config: ...) -> Self {}
}
```

The examples can then move to the docs of these functions.

Happy to merge this too as it is better than what we have today.

Thanks @melekes !

melekes:
> I would still be in favor of dedicated constructors that take the necessary inner configs:

It just feels too cumbersome when I add 4 new methods (2 fn and 2 features: tokio and async-io). I'd be in favor of just merging as is.

thomaseizinger:
> > I would still be in favor of dedicated constructors that take the necessary inner configs:
> 
> It just feels too cumbersome when I add 4 new methods (2 fn and 2 features: tokio and async-io). I'd be in favor of just merging as is.

Ah yes, I forgot that we'd have to do one per executor.

thomaseizinger (dismissed):

thomaseizinger (approved):

PR #3386, opened 2023-01-25
feat: introduce `libp2p-connection-limits` connection management module

thomaseizinger:
## Description

This patch deprecates the existing connection limits within `Swarm` and uses the new `NetworkBehaviour` APIs to implement it as a plugin instead.

Related #2824.

## Notes

~This is a WIP in that it still needs tests but overall, this is pretty much what it will look like.~

## Change checklist


swarm/src/connection_limits.rs
```diff
@@ -0,0 +1,196 @@
+use crate::{
```
mxinden: License header missing :angel:

mxinden (commented):
:sparkles: 

Ignoring the fact that this is still a draft pull request, change looks good to me. Just missing changelog entries.

swarm/src/connection_limits.rs
```diff
@@ -0,0 +1,422 @@
...
+    dummy, ConnectionClosed, ConnectionDenied, ConnectionId, ConnectionLimit, ConnectionLimits,
+    FromSwarm, NetworkBehaviour, NetworkBehaviourAction, PollParameters, THandler, THandlerInEvent,
+    THandlerOutEvent,
+};
+use libp2p_core::{Endpoint, Multiaddr, PeerId};
+use std::collections::{HashMap, HashSet};
+use std::task::{Context, Poll};
+use void::Void;
+
+pub struct Behaviour {
```
mxinden: The deprecation note on the old connection limiting will point to this type. What do you think of including a small example showcasing how one can use this `NetworkBehaviour`?

mxinden (commented):
:rocket: great to see the new feature in action. One comment, otherwise looks good to me.

swarm/src/connection_limits.rs
thomaseizinger: Done!

swarm/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+# 0.42.1 [unreleased]
```
mxinden, line 1:
```suggestion
# 0.42.1
```

I can cut a new release of `libp2p-swarm` directly once merged. What do you think?

mxinden (approved):
Ready to merge from my end. Let me know in case you want a new release of `libp2p-swarm` right away @thomaseizinger.

swarm/CHANGELOG.md
thomaseizinger: I'd want to replace the banning API as well do make #2824 "complete" for now.

thomaseizinger:
@mxinden: I deprecated the old `ConnectionLimit` type in favor of a new `Exceeded` type that has a clearer error message and smaller API. I also discovered that there were a few old, not yet deprecated APIs in the error types.

Let me know what you think.

thomaseizinger:
Use the new `libp2p-swarm-test` framework for this.

thomaseizinger:
> Use the new `libp2p-swarm-test` framework for this.

To actually do this, I had to move `connection_limits` to a separate crate because it didn't work to depend on `libp2p-swarm-test` from `libp2p-swarm`. I think this is a tad cleaner anyway because we don't unnecessarily grow the `libp2p-swarm` API then.

misc/connection-limits/src/lib.rs
```diff
@@ -0,0 +1,483 @@
...
+
+    /// Configures the maximum number of concurrent established connections per peer,
+    /// regardless of direction (incoming or outgoing).
+    pub fn with_max_established_per_peer(mut self, limit: Option<u32>) -> Self {
+        self.max_established_per_peer = limit;
+        self
+    }
+}
+
+impl NetworkBehaviour for Behaviour {
```
mxinden, line 202: We overload the string _protocol_ with many meanings. The main one for in relation with the `protocols/` folder being that all things in their implement `NetworkBehaviour`.

Given that `libp2p_connection_limits::Behaviour` implements `NetworkBehaviour` I think it should live in `protocols/`.

misc/connection-limits/CHANGELOG.md
```diff
@@ -0,0 +1,3 @@
+# 0.1.0 - unreleased
```
mxinden, line 1: Nit pick, we recommend the below. I don't care which syntax we use, as long as we stay consistent. Makes my life creating the `docs(CHANGELOG): Prepare xxx` a bit easier.

```suggestion
# 0.1.0 [unreleased]
```

https://github.com/libp2p/rust-libp2p/blob/master/docs/release.md#development-between-releases

mxinden (commented):
Looks good to me apart from suggestion to move to `protocols/`.

misc/connection-limits/CHANGELOG.md
thomaseizinger: `[unreleased]` is a link reference that doesn't point to anything so I stopped using that. Can update the docs if you agree.

misc/connection-limits/src/lib.rs
thomaseizinger: > The main one for in relation with the `protocols/` folder being that all things in their implement `NetworkBehaviour`.

We already have multiple utilities in `libp2p-swarm` that implement `NetworkBehaviour` without living under `protocols/`. To me, a libp2p "protocol" must have some form of network communication. This one here uses the `dummy::ConnectionHandler` so it will never perform any IO, hence I placed it in `misc/`.

misc/connection-limits/CHANGELOG.md
mxinden: Never thought about that. Yes, patch would be appreciated.

misc/connection-limits/src/lib.rs
thomaseizinger: As you know, I am not a massive fan of the directory structure anyway but putting this one into `protocols/` actually feels wrong.

Protocols should be present on the wire, this one here is just an internal `rust-libp2p` concept.

`misc/` is not a great name either because everything fits the description :)

With `libp2p-identity` merged, I'd like to push for the next step towards moving around the crate boundaries. Once we have all that in place, I think it makes sense to think about new directories.

Thoughts?

misc/connection-limits/CHANGELOG.md
thomaseizinger: PR here: https://github.com/libp2p/rust-libp2p/pull/3615

thomaseizinger:
I am removing the dependency on #3615. Based on https://github.com/libp2p/rust-libp2p/pull/3386#discussion_r1134619924, I am assuming that @mxinden is in favor of the new changelog syntax so I am using this here and we can iterate on the exact wording in #3615 after that.

PR #3389, opened 2023-01-26
fix: deal with new lints from beta clippy

thomaseizinger:
## Description

Most of this is trivial, apart from the rename of the `clippy::derive_hash_xor_eq` lint to `clippy::derived_hash_with_manual_eq`.

Instead of allowing that lint, we manually implement `PartialEq` and add a comment why the difference between the `PartialEq` and `Hash` implementations are okay.

## Change checklist


elenaf9:
I would prefer if we simply resolve the lint, i.e. manually implement `PartialEq` for `LocalStreamId` so that it respects the (imo reasonable) invariant `v1 == v2 <=> h(v1) == h(v2)`:

```rust 
impl PartialEq for LocalStreamId {
    fn eq(&self, other: &Self) -> bool {
        self.num == other.num
    }
}
```

Else if `PartialEq` should also compare the `role` field we can add that in the manual implementation, but then I think it deserves a comment stating why it is okay to do this (I assume it is because both sides agree on the role for each stream).

muxers/mplex/src/codec.rs
```diff
@@ -65,7 +65,12 @@ impl fmt::Display for LocalStreamId {
 }
 
 impl Hash for LocalStreamId {
-    #![allow(clippy::derive_hash_xor_eq)]
+    #![allow(
+        unknown_lints, // Prior to Rust 1.68, ignore the new `derived_hash_with_manual_eq` lint.
+        renamed_and_removed_lints, // After Rust 1.68, ignore usage of the old `derive_hash_xor_eq` lint.
+        clippy::derive_hash_xor_eq, // Prior to Rust 1.68, silence the lint.
+        clippy::derived_hash_with_manual_eq // After Rust 1.68, silence the lint.
```
elenaf9, line 72: That lint name is confusing. We are actually doing it the other way around here: we derive eq and manually implement hash.

thomaseizinger:
Of course, that is a much simpler way forward. Thanks @elenaf9!

elenaf9 (approved):

jxs (approved):

thomaseizinger:
@elenaf9 @jxs So I actually had to replicate the exact implementation of the derived `PartialEq` to make the tests pass again. For some reason, this even affected network compatibility (interop tests wuhu!).

thomaseizinger:
@mergifyio backport rust-v050-testplan

mxinden:
> @ Mergifyio backport rust-v050-testplan

:exploding_head:

thomaseizinger:
> > @ Mergifyio backport rust-v050-testplan
> 
> :exploding_head: 

😉

PR #3391, opened 2023-01-26
fix(ci): replace `actions-rs/toolchain` with `dtolnay/toolchain`

### LINKED ISSUE #3352: ci: migrate away from unmaintained `actions-rs/toolchain` action
## Description

https://github.com/actions-rs/toolchain is pretty much unmaintained. It had its last commit in 2020. Our CI generated a lot of warnings due to outdated NodeJS toolchains and deprecated GitHub actions commands. See https://github.com/libp2p/rust-libp2p/actions/runs/3963608745 for example.

We have until **June 2023** to migrate away from the deprecated GitHub actions commands: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

## Motivation

A working CI.

Maybe.

leviathanbeak:
## Description
This one seems it [got a blessing](https://github.com/rust-lang/regex/pull/883) from rust-lang.

Closes #3352.


.github/workflows/cache-factory.yml
```diff
@@ -38,7 +38,7 @@ jobs:
 
       - uses: actions/checkout@v3
 
-      - uses: actions-rs/toolchain@16499b5e05bf2e26879000db0c1d13f7e13fa3af # v1.0.7
+      - uses: dtolnay/rust-toolchain@v1
```
thomaseizinger: Thanks!

We have the policy of pinning third-party actions to a hash. See https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions.

Please use the hash and write the comment next to it. Dependabot can parse that and will bump it for is when there is a new release.

thomaseizinger (changes requested):
One request, otherwise LGTM!

.github/workflows/cache-factory.yml
leviathanbeak: I see, updated the PR

thomaseizinger (changes requested):
Please check the warnings of the CI pipeline: https://github.com/libp2p/rust-libp2p/actions/runs/4027747607

dtolnay's action seems to support a different set of inputs!

leviathanbeak:
@thomaseizinger as the warning says, dtolnay's action does not support `profile` and `override` and there does not seem to be an alternative for those inputs.

The only available [ones are](https://github.com/dtolnay/rust-toolchain#inputs):
- toolchain
- target(s)
- components

Let me know do you want me to just delete the _missing_ inputs or just abort this PR altogether because they are not supported.

thomaseizinger:
Given that CI is overall green, I think we should be able to just delete them. `override` seems to be implied: https://github.com/libp2p/rust-libp2p/actions/runs/4027747607/jobs/6923843196#step:5:78 and I don't think we need to specify the `profile`. That is probably a performance optimization that doesn't really pull its weight compared to the other steps in the workflow.

thomaseizinger (dismissed):

thomaseizinger (dismissed):

thomaseizinger (approved):

PR #3412, opened 2023-01-31
refactor(identify): use `ReadyUpgrade` for {In,Out}boundUpgrade

jxs:
## Description
Starts addressing https://github.com/libp2p/rust-libp2p/issues/2863
## Notes

I read the discussion Thomas, namely [your comment](https://github.com/libp2p/rust-libp2p/issues/2863#issuecomment-1377330094) to start with the `TimedFutures` implementation. But as I wasn't completely familiar with everything you wrote I started experimenting with `identify` and was able to use `ReadyUpgrade` for `{In,Out}boundUpgrade` therefore removing most of the `protocol` code.

## Open Questions

- I am not familiar with all the protocols yet, but wdyt of following the same approach hereby presented for `identify` and first replace all custom `{In,Out}boundUpgrade`'s with `ReadyUpgrade`'s then creating the `TimedFutures` and then introducing the breaking changes on the `ConnectionHandler`?
The reason I ask this, is because with `identify` for example this PR AFAICT maintains the same timeouts and backpressure right? Nonetheless if you prefer we can add `TimedFutures` now and introduce it already to `identify` no strong opinion.

- What do you think of joining all the `events` on the same `FuturesUnordered`? I feel it makes sense as they are all events, it's just some haven't been resolved but if prefered I can rollback to the `events` and `pending_replies` approach, though that will probably mean the need for another enum with the identify variants.  

## Change checklist

### DIFF
--- a/protocols/identify/src/handler.rs
+++ b/protocols/identify/src/handler.rs
@@ -18,15 +18,14 @@
 // FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
 // DEALINGS IN THE SOFTWARE.
 
-use crate::protocol::{
-    self, Identify, InboundPush, Info, OutboundPush, Protocol, Push, UpgradeError,
-};
+use crate::protocol::{self, InboundPush, Info, OutboundPush, Protocol, Push, UpgradeError};
+use crate::PROTOCOL_NAME;
 use either::Either;
 use futures::future::BoxFuture;
 use futures::prelude::*;
 use futures::stream::FuturesUnordered;
 use futures_timer::Delay;
-use libp2p_core::upgrade::SelectUpgrade;
+use libp2p_core::upgrade::{ReadyUpgrade, SelectUpgrade};
 use libp2p_core::{ConnectedPoint, Multiaddr, PeerId, PublicKey};
 use libp2p_swarm::handler::{
     ConnectionEvent, DialUpgradeError, FullyNegotiatedInbound, FullyNegotiatedOutbound,
@@ -87,7 +86,7 @@ impl IntoConnectionHandler for Proto {
     }
 
     fn inbound_protocol(&self) -> <Self::Handler as ConnectionHandler>::InboundProtocol {
-        SelectUpgrade::new(Identify, Push::inbound())
+        SelectUpgrade::new(ReadyUpgrade::new(PROTOCOL_NAME), Push::inbound())
     }
 }
 
@@ -101,14 +100,20 @@ pub struct Handler {
     inbound_identify_push: Option<BoxFuture<'static, Result<Info, UpgradeError>>>,
     /// Pending events to yield.
     events: SmallVec<
-        [ConnectionHandlerEvent<Either<Identify, Push<OutboundPush>>, (), Event, io::Error>; 4],
+        [ConnectionHandlerEvent<
+            Either<ReadyUpgrade<&'static [u8]>, Push<OutboundPush>>,
+            (),
+            Event,
+            io::Error,
+        >; 4],
     >,
 
     /// Streams awaiting `BehaviourInfo` to then send identify requests.
     reply_streams: VecDeque<NegotiatedSubstream>,
 
-    /// Pending identification replies, awaiting being sent.
-    pending_replies: FuturesUnordered<BoxFuture<'static, Result<PeerId, UpgradeError>>>,
+    /// Pending identification replies, awaiting being sent and received.
+    pending_replies:
+        FuturesUnordered<BoxFuture<'static, Result<Either<PeerId, Info>, UpgradeError>>>,
 
     /// Future that fires when we need to identify the node again.
     trigger_next_identify: Delay,
@@ -235,11 +240,11 @@ impl Handler {
     ) {
         match output {
             future::Either::Left(remote_info) => {
-                self.events
-                    .push(ConnectionHandlerEvent::Custom(Event::Identified(
-                        remote_info,
-                    )));
-                self.keep_alive = KeepAlive::No;
+                let fut = Box::pin(async move {
+                    let info = protocol::recv(remote_info).await?;
+                    Ok(Either::Right(info))
+                });
+                self.pending_replies.push(fut);
             }
             future::Either::Right(()) => self
                 .events
@@ -258,7 +263,7 @@ impl Handler {
 
         let err = err.map_upgrade_err(|e| match e {
             UpgradeError::Select(e) => UpgradeError::Select(e),
-            UpgradeError::Apply(Either::Left(ioe)) => UpgradeError::Apply(ioe),
+            UpgradeError::Apply(Either::Left(_ioe)) => unreachable!(),
             UpgradeError::Apply(Either::Right(ioe)) => UpgradeError::Apply(ioe),
         });
         self.events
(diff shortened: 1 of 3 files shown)


protocols/identify/src/handler.rs
```diff
@@ -227,60 +239,53 @@ impl Handler {
...
-                    .push(ConnectionHandlerEvent::Custom(Event::Identified(
-                        remote_info,
-                    )));
-                self.keep_alive = KeepAlive::No;
+            future::Either::Left(substream) => {
+                let fut = Box::pin(async move {
+                    let info = protocol::recv(substream).await?;
+                    Ok(ConnectionHandlerEvent::Custom(Event::Identified(info)))
+                });
+                self.pending_events.push(fut);
```
thomaseizinger: The difference in behaviour is that `swarm::Connection` would have raced the future returned from `OutboundUpgrade` against a timeout. See https://github.com/libp2p/rust-libp2p/blob/8f3b7e3876014aa232fa4c54cd67a145686fffc6/swarm/src/connection.rs#L461-L500.

Meaning, if successful, `output` on line 230 was the result of the future. On a timeout, `on_dial_upgrade_errror` would have been called.

With this patch, we receive the bare substream on line 242 and practically wait forever to receive a message on it.

This is what the `TimedFutures` abstraction is meant to solve: It is essentially a wrapped around `FuturesUnordered`, same as you have it here with the only addition that it will wrap each provided future in a timeout.

If you want, we wan can pair on creating this abstraction together :)

thomaseizinger:
> * I am not familiar with all the protocols yet, but wdyt of following the same approach hereby presented for `identify` and first replace all custom `{In,Out}boundUpgrade`'s with `ReadyUpgrade`'s then creating the `TimedFutures` and then introducing the breaking changes on the `ConnectionHandler`?
>       The reason I ask this, is because with `identify` for example this PR AFAICT maintains the same timeouts and backpressure right? Nonetheless if you prefer we can add `TimedFutures` now and introduce it already to `identify` no strong opinion.

As I laid out in the comment below, I don't think we will maintain the same timeouts unfortunately. What we can do is manually compose together `protocol::recv` and `futures_timer::Delay` with a `future::select`. It will get a bit clunky to name things but it is functionally equivalent to what I am planning to have with `TimedFutures`.

thomaseizinger:
> * What do you think of joining all the `events` on the same `FuturesUnordered`? I feel it makes sense as they are all events, it's just some haven't been resolved but if prefered I can rollback to the `events` and `pending_replies` approach, though that will probably mean the need for another enum with the identify variants.

I think having multiple sets (i.e. different fields) of tasks is better.

1. It allows us to prioritize tasks amongst each other, i.e. process outbound requests before inbound requests by choosing, which one to poll first in `ConnectionHandler::poll`.
2. It should result in a flatter type hierarchy as each set of tasks can have a separate output and we don't need to make an enum that handles all task outputs. Within `poll`, we just need to map to `ConnectionHandlerEvent` then.

protocols/identify/src/handler.rs
jxs: Thanks for the explanation Thomas, implemented a `TimedFutures` based on your description and `async-bounded-queue`: c3646562c201ed9a85ea23787b286fa74a9994db. You mentioned for it to be bounded and a waker to wake when it gets empty, but you meant full right? Another question is, how can we `poll_push_ready` if `on_connection_event` is not an async function? 

>If you want, we wan can pair on creating this abstraction together :)

totally! I went ahead and bootstrapped it as you were off but we can arrange for Tomorrow, or whenever you are available.

jxs:
> > * What do you think of joining all the `events` on the same `FuturesUnordered`? I feel it makes sense as they are all events, it's just some haven't been resolved but if prefered I can rollback to the `events` and `pending_replies` approach, though that will probably mean the need for another enum with the identify variants.
> 
> I think having multiple sets (i.e. different fields) of tasks is better.
> 
> 1. It allows us to prioritize tasks amongst each other, i.e. process outbound requests before inbound requests by choosing, which one to poll first in `ConnectionHandler::poll`.
> 2. It should result in a flatter type hierarchy as each set of tasks can have a separate output and we don't need to make an enum that handles all task outputs. Within `poll`, we just need to map to `ConnectionHandlerEvent` then.

yeah makes sense, reverted it ptal :)

protocols/identify/src/handler.rs
thomaseizinger: > Thanks for the explanation Thomas, implemented a `TimedFutures` based on your description and `async-bounded-queue`: c3646562c201ed9a85ea23787b286fa74a9994db. You mentioned for it to be bounded and a waker to wake when it gets empty, but you meant full right? 

We need both:

- A waker that wakes the `poll` task when it is empty and a new element gets pushed
- A waker that wakes the push task when it is full and a futures finishes.

thomaseizinger: > Another question is, how can we `poll_push_ready` if `on_connection_event` is not an async function?

At the moment we can't unfortunately but as part of implementing back-pressure, we need to continue this chain eventually up to the connection task :)

For now, we can construct a no-op waker and call the function. If it is pending, we should print a warning and drop the event.

thomaseizinger:
Originally, I envisioned `TimedFutures` to be used together with this: https://github.com/thomaseizinger/rust-async-response-stream/

But, I've since evolved this idea a bit and the latest draft is now here: https://github.com/mxinden/asynchronous-codec/pull/5

That implements `Stream` now instead of `Futures`. If we like this idea, we may need to adapt `TimedFutures` to be `TimedStreams` instead where we race a stream against a timeout.

Technically, this can also be done separately but I think it might be good to test compatibility of these ideas first and see how it pans out.

protocols/identify/src/handler.rs
jxs: >A waker that wakes the poll task when it is empty and a new element gets pushed

Since we are polling `FuturesUnordered` shouldn't it be enough to add to its list and poll it? Why do we need an extra `Waker` ?

thomaseizinger: > > A waker that wakes the poll task when it is empty and a new element gets pushed
> 
> Since we are polling `FuturesUnordered` shouldn't it be enough to add to its list and poll it? Why do we need an extra `Waker` ?

I think `FuturesUnordered` will return `Poll::Ready(None)` if it has no items and not `Poll::Pending`.

thomaseizinger:
> If we like this idea

The idea more concretely is that we have a `futures::stream::SelectAll` (or equivalent like `TimedStreams`) in our `ConnectionHandler`. Each new stream is wrapped in a "message pattern", e.g. in the [PR](https://github.com/mxinden/asynchronous-codec/pull/5), I implemented `RecvSend`. `RecvSend` implements `Stream`, yielding events. In the simplest case (closing stream after), we only ever yield one event per stream: The incoming message plus a placeholder that we can send the response into. The stream internally listens for this placeholder to be filled, writes it to the stream, closes it and then terminates. `SelectAll` will then clean up the stream.

This we would combine with `TimedStreams` which ensures that each stream finishes (i.e. returns `None`) in a given timeout. I believe this would yield in code that gives a lot of control over what is happening without much boilerplate. After all, we will be doing this in every `ConnectionHandler` hence being concise is important.

thomaseizinger:
In https://github.com/libp2p/rust-libp2p/pull/4275, we are now adding a `futures-bounded` crate which should be of good use here. But perhaps this should just be restarted from current `master`.

thomaseizinger:
https://github.com/libp2p/rust-libp2p/pull/4275 is now merged and `futures-bounded` is ready to use if you are still keen for making that change here @jxs :)

jxs:
Thanks for the ping Thomas, I won't have much bandwidth in the next week, @dgarus you wanna go ahead and take it?

dgarus:
> Thanks for the ping Thomas, I won't have much bandwidth in the next week, @dgarus you wanna go ahead and take it?

Hi colleagues!
I could address this issue.

thomaseizinger:
Closing in favor of https://github.com/libp2p/rust-libp2p/pull/4563.

PR #3409, opened 2023-01-31
refactor(inter-op): misc clean-up

thomaseizinger:
## Description

This patch contains misc cleanup that I did as part of reading through the code to understand how it works.

## Change checklist

- [ ] I have performed a self-review of my own code
- [ ] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] A changelog entry has been made in the appropriate crates

PR #3415, opened 2023-02-01
feat: Backport interop test to rust-libp2p v050

thomaseizinger:
## Change checklist


jxs (approved):
LGTM!

jxs:
sorry @thomaseizinger only now related the publishing of the images with https://github.com/libp2p/test-plans/pull/121. Since we are on it shouldn't we publish images conforming with it?

thomaseizinger:
> sorry @thomaseizinger only now related the publishing of the images with https://github.com/libp2p/test-plans/pull/121. Since we are on it shouldn't we publish images conforming with it?

Probably yes! We can always just publish a new version though!

But yeah, might as well wait until the test in master is updated and then update these branches!

jxs:
do you mind then if I go ahead, and update the the `master` here following the work you did on https://github.com/libp2p/rust-libp2p/pull/3409?

jxs:
submitted https://github.com/libp2p/rust-libp2p/pull/3423

interop-tests/Cargo.toml
```diff
@@ -11,7 +11,7 @@ either = "1.8.0"
 env_logger = "0.9.0"
 futures = "0.3.26"
 if-addrs = "0.8.0"
-libp2p = { path = "../", default_features = false, features = ["websocket", "quic",  "mplex", "yamux", "tcp", "tokio", "ping", "noise", "tls", "dns", "rsa", "macros", "webrtc"] }
+libp2p = { version = "0.50.0", default_features = false, features = ["websocket", "quic",  "mplex", "yamux", "tcp", "tokio", "ping", "noise", "tls", "dns", "rsa", "macros", "webrtc"] }
```
thomaseizinger: An alternative to this would be to branch off the `0.50.0` tag and copy the test in but I am not sure that buys us anything.

thomaseizinger:
We will still need this, even after the outcome of the discussion in https://github.com/libp2p/test-plans/pull/121#discussion_r1095146882 to not use the registry but these commits are what we will need to reference for the tests of those versions.

thomaseizinger:
@galargh If we define a branch protection for these branches, they cannot accidentally be deleted right?

thomaseizinger:
This is now based on the `v0.50.0` tag and targets a `release/v0.50.0` branch that I created. I think that makes conceptually more sense as we are "backporting" something to the `v0.50.0` release.

mxinden:
In general I am in favor of backporting the interoperability tests to the v0.50 family.

> This is now based on the `v0.50.0` tag and targets a `release/v0.50.0` branch that I created.

What is the benefit of the `release/v0.50.0` branch over the existing `v0.50` branch?

> I think that makes conceptually more sense as we are "backporting" something to the `v0.50.0` release.

Say there is ever a `v0.50.1` release, why would we still be testing against `v0.50.0`? In my eyes we should always be testing against the latest patch release of a release family.

galargh:
> @galargh If we define a branch protection for these branches, they cannot accidentally be deleted right?

Yes, that's correct ✅

thomaseizinger:
> In general I am in favor of backporting the interoperability tests to the v0.50 family.
> 
> > This is now based on the `v0.50.0` tag and targets a `release/v0.50.0` branch that I created.
> 
> What is the benefit of the `release/v0.50.0` branch over the existing `v0.50` branch?

Nothing, I did't know it existed or put differently, I "looked" like a tag when I went through the auto-complete options of my local Git installation.

I can target the `v0.50` branch instead.

> > I think that makes conceptually more sense as we are "backporting" something to the `v0.50.0` release.
> 
> Say there is ever a `v0.50.1` release, why would we still be testing against `v0.50.0`? In my eyes we should always be testing against the latest patch release of a release family.

I agree so too and now understand your naming strategy!

mxinden (approved):
Looks good to me. Thanks for changing the target branch @thomaseizinger.

PR #3416, opened 2023-02-01
feat: Backport interop test to rust-libp2p v049

thomaseizinger:
## Description

Version `0.49.0` does not support QUIC, WebRTC and TLS so I have removed those from the JSON file. What this effectively does is that we "pretend" as if the ping test is now "rust-libp2p-head". Meaning if the interop tests pass on this PR, we should be good to go with releasing this as a docker container and supplying the JSON file we see here to the test-plans repository.

## Change checklist


jxs (approved):
LGTM :)

Cargo.toml~
```diff
@@ -0,0 +1,218 @@
...
+    "transports/plaintext",
+    "transports/pnet",
+    "transports/tcp",
+    "transports/uds",
+    "transports/websocket",
+    "transports/wasm-ext",
+<<<<<<< HEAD
+=======
+    "transports/webrtc",
+>>>>>>> eeca244c (feat(interop-tests): build test binary inside container (#3441))
```
mxinden: Merge conflict? Also is this file (`Cargo.toml~`) supposed to be checked in?

thomaseizinger: It is not, not sure what went wrong there :thinking:

mxinden:
Great. Thank you!

PR #3417, opened 2023-02-01
feat: Backport interop test to rust-libp2p v048

thomaseizinger:
## Description

Same as for the backport to `v.049.0`, we remove the QUIC, WebRTC and TLS capabilities from the JSON file which should give us a green interop test result.

## Change checklist


jxs (approved):
LGTM!

MarcoPolo:
One downside is that this container will be an x86 image. Is it possible to publish both arm64 and x86 images?

thomaseizinger:
> One downside is that this container will be an x86 image. Is it possible to publish both arm64 and x86 images?

Yes, we just need to cross-compile the binary.

thomaseizinger:
> > One downside is that this container will be an x86 image. Is it possible to publish both arm64 and x86 images?
> 
> Yes, we just need to cross-compile the binary.

I've done this in the past. It is a bit involved so I'd like to defer it past the initial milestone of our interop tests.

Are you planning to run the tests on ARM as well? That is another variable in the matrix.

The way docker multi-arch images work is that the root image is just a manifest that points to one image per arch. Hence, you can `docker run` the same image on multiple arch's and it will always work.

Consequently, you need to publish 3[^1] images:

- one for x86
- one for arm64
- manifest image pointing to both

Doable, but a bit involved.

[^1]: `docker buildx` can help here but it is a convoluted setup so I wouldn't use it.

mxinden (approved):
I suggest targeting `v0.48` branch and not `release/v0.48.0`. `v0.48` branch does not yet exist, thus needs to be created off of the `v0.48.0` tag first.

Otherwise this looks good to me.

(I have not yet caught up on the discussion around checking in the `Cargo.lock` file, but that doesn't need to block this pull request.)

thomaseizinger:
> I suggest targeting `v0.48` branch and not `release/v0.48.0`. `v0.48` branch does not yet exist, thus needs to be created off of the `v0.48.0` tag first.

Will do!

> I suggest targeting `v0.48` branch and not `release/v0.48.0`. `v0.48` branch does not yet exist, thus needs to be created off of the `v0.48.0` tag first.
> 
> Otherwise this looks good to me.
> 
> (I have not yet caught up on the discussion around checking in the `Cargo.lock` file, but that doesn't need to block this pull request.)

Without a lockfile, the builds aren't necessarily reproducible so I think we definitely need it for these tests. Overall, I came around to the view that we should have it for the library as well. It helps us manage potential semver breaks in our dependencies better. Otherwise those can block our CI pipeline which is annoying.

mxinden (approved):
Good to go from my end. Thanks for the work!

thomaseizinger:
I ignored the clippy failures because it is too painful to backport them.

PR #3420, opened 2023-02-01
fix(quic): discard correct waker upon accepting inbound stream

tthebst:
## Notes

Set `poll_inbound_waker = None` instead of `poll_outbound_waker` when successfully accepting an inbound stream.


thomaseizinger (dismissed):
Thank you! Well spotted!

Did you observe a bug and as a result found this or was this just from reviewing the code? If that results in an observable bug, we should add a changelog entry I think.

Also cc @elenaf9.

tthebst:
I just noticed this when reviewing the code.

thomaseizinger:
@tthebst I can't seem to update the branch with latest master. Can you grant the necessary permissions please?

thomaseizinger (approved):

mxinden:
> Well spotted!

Indeed. Thank you @tthebst.

PR #3439, opened 2023-02-09
[DO NOT MERGE] v050 testplan backport

MarcoPolo:
## Description

Updates the test implementation to follow the new spec.

Updates the dockerfile so it can be built by anyone, not just linux folks

In draft to fiddle with CI.


MarcoPolo:
@thomaseizinger Can you take a look? If this looks good we should port the relevant changes to the master branch.

interop-tests/ping-version.json
```diff
@@ -0,0 +1,18 @@
...
+  ],
+  "secureChannels": [
+    "tls",
+    "noise"
+  ],
+  "muxers": [
+    "mplex",
+    "yamux"
+  ]
+}
```
thomaseizinger, line 18: We don't need this file here (for v050), right? (I know I added it)

interop-tests/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+# Setup protoc. TODO this breaks reproducibility and uses an old version of protoc.
+# In the future protobuf generated files will be checked in, so we can remove this
+WORKDIR /protoc-setup
+RUN apt-get update && apt-get install -y cmake protobuf-compiler
+
+# Run with access to the target cache to speed up builds
+WORKDIR /workspace
+ADD . .
+RUN --mount=type=cache,target=./target \
+    --mount=type=cache,target=/usr/local/cargo/registry \
```
thomaseizinger, line 13: What exactly is the cache referenced here? Old layers that don't match exactly?

interop-tests/Dockerfile
```diff
@@ -0,0 +1,20 @@
+# syntax=docker/dockerfile:1.5-labs
+FROM rust:1.67.0
```
thomaseizinger, line 2: Can this be a multi-stage dockerfile?

.github/workflows/interop-test.yml
```diff
@@ -1,32 +1,30 @@
...
-      custom_git_reference: ${{ github.event.pull_request.head.sha || github.sha }}
-      custom_interop_target: rust
+  run-multidim-interop:
+    name: Run multidimensional interoperability tests
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v3
+      - name: Set up Docker Buildx
+        id: buildx
+        uses: docker/setup-buildx-action@v2
```
thomaseizinger:
```suggestion
      - uses: docker/setup-buildx-action@v2
```

Do we need the id?

interop-tests/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+# In the future protobuf generated files will be checked in, so we can remove this
+WORKDIR /protoc-setup
+RUN apt-get update && apt-get install -y cmake protobuf-compiler
+
+# Run with access to the target cache to speed up builds
+WORKDIR /workspace
+ADD . .
+RUN --mount=type=cache,target=./target \
+    --mount=type=cache,target=/usr/local/cargo/registry \
+    cd interop-tests && cargo build --bin ping
```
thomaseizinger:
```suggestion
    cargo build --package interop-tests
```

thomaseizinger (commented):
Couple of questions, otherwise LGTM. I already have backport PRs open. Do you mind just opening a PR to master? We can merge that separately and I'll backport it to the releases later.

interop-tests/ping-version.json
MarcoPolo: Not for this test itself, but we need it in master in general. It does nothing here.

interop-tests/Dockerfile
MarcoPolo: This is a fairly new cool feature. It is a cache that is saved separate from layers. So even if the earlier layers get rebuilt we still persist our cargo and build cache. So builds will be just be fast without any tricks. https://docs.docker.com/engine/reference/builder/#run---mount

interop-tests/Dockerfile
MarcoPolo: Sure, but why? We aren't pushing this to a registry or uploading/downloading the image somewhere.

MarcoPolo:
>  Do you mind just opening a PR to master? We can merge that separately and I'll backport it to the releases later.

Sounds good

interop-tests/Dockerfile
thomaseizinger: Right and we need to pull all layers anyway so I guess it doesn't matter!

interop-tests/Dockerfile
thomaseizinger: This is amazing! Thank for your introducing me to this feature, that is indeed a godsend for Rust-based containers!

Issue #3438, opened 2023-02-08
protocols/relay: Upgrade to proto3


mxinden:
## Description

The circuit relay v2 specification migrated from proto2 to proto3. The rust-libp2p circuit relay v2 implementation should change accordingly.

https://github.com/libp2p/specs/issues/509

## Are you planning to do it yourself in a pull request?

No

thomaseizinger:
Untagging this because it is blocked by a release of `quick-protobuf`.

PR #3455, opened 2023-02-11
feat(relay): upgrade circuit relay v2 to proto3 according the spec

d-roak:
## Description
Changed the circuit v2 to proto3, also changed the relevant code because of the "new" optional fields.

- #3438 

## Change checklist


thomaseizinger:
This may have to wait for https://github.com/libp2p/rust-libp2p/pull/3312. The version of `protoc` that we are installing in CI seems to be too old for the features that we are using in these proto files. Whilst we could install a newer version with a different action, we explicitly moved away from that in https://github.com/libp2p/rust-libp2p/pull/3258 because it caused CI to be flaky.

https://github.com/libp2p/rust-libp2p/pull/3312 is very close to being merged so I suggest we wait here until that has landed.

One note of feedback: We cannot use use `.unwrap` here. That would allow a remote node to panic our code by sending us empty protobuf structures. Instead, we need to explicitly unpack these fields and gracefully fail if they are not present.

d-roak:
> This may have to wait for #3312. The version of `protoc` that we are installing in CI seems to be too old for the features that we are using in these proto files. Whilst we could install a newer version with a different action, we explicitly moved away from that in #3258 because it caused CI to be flaky.
> 
> #3312 is very close to being merged so I suggest we wait here until that has landed.

Seems good!

> One note of feedback: We cannot use use `.unwrap` here. That would allow a remote node to panic our code by sending us empty protobuf structures. Instead, we need to explicitly unpack these fields and gracefully fail if they are not present.

Thank you for the feedback. I thought that .ok_or() would prevent the panic if the .from_i32() received an empty value. What would you think should be the approach for unpacking? Something along the lines of the image below?
image

protocols/relay/src/protocol/inbound_hop.rs
```diff
@@ -66,7 +66,7 @@ impl upgrade::InboundUpgrade<NegotiatedSubstream> for Upgrade {
                 .ok_or(FatalUpgradeError::StreamClosed)??;
 
             let r#type =
-                hop_message::Type::from_i32(r#type).ok_or(FatalUpgradeError::ParseTypeField)?;
+                hop_message::Type::from_i32(r#type.unwrap()).ok_or(FatalUpgradeError::ParseTypeField)?;
```
thomaseizinger:
```suggestion
                hop_message::Type::from_i32(r#type.ok_or(FatalUpgradeError::MissingType)?).ok_or(FatalUpgradeError::ParseTypeField)?;
```
This will fail without panicking.

thomaseizinger (commented):
Left an idea for more graceful failure :)

d-roak:
Quick-protobuf does not seem to support proto3 optionals currently (https://github.com/tafia/quick-protobuf/issues/219)

thomaseizinger:
Yep, we will need to wait for https://github.com/tafia/quick-protobuf/pull/247 to be merged and released for this.

thomaseizinger:
@jpldcarvalho https://github.com/tafia/quick-protobuf/pull/247 is ready for review. To speed up the feedback loop there, would you mind testing that branch within this PR?

To do that, you would need to:

- Install the `pb-rs` binary from that branch: `cargo install --git https://github.com/snproj/quick-protobuf --branch field-presence pb-rs`
- Run `pb-rs --dont_use_cow **/generated/*.proto` on this branch
- Commit the generated sources
- Check the branch for compile errors and see whether the newly generated code is sufficient for our usecase

Much appreciated!

d-roak:
@thomaseizinger that branch fixes the proto3 optionals.

We still have a build error when the generated pb code tries to import the `PackedFixed` stuff from quick-protobuf. This is easily fixed by using the `field-presence` branch as a dependency. But it seems we will have some problems with the `quick_protobuf_codec`.

d-roak:
Regarding the last message, if you use that branch as a dependency in the `quick_protobuf_codec` and every package that uses the pb_codec (dcutr, flood/gossipsub, identify, relay, rendezvous and webrtc), you can build the workspace without errors.

thomaseizinger:
> @thomaseizinger that branch fixes the proto3 optionals.
> 
> We still have a build error when the generated pb code tries to import the `PackedFixed` stuff from quick-protobuf. This is easily fixed by using the `field-presence` branch as a dependency. But it seems we will have some problems with the `quick_protobuf_codec`.

Sweet, thanks for trying it out so quickly!

thomaseizinger:
CI still highlights a few things:

- MSRV needs to be bumped to 1.62 for some packages (which we treat as a breaking change)
- For `libp2p-relay`, the addition of the error type is a breaking change

Adding this to the `0.52` milestone as a result (i.e. we will merge this once we decide to cut a new breaking change release)

thomaseizinger:
> Adding this to the `0.52` milestone as a result (i.e. we will merge this once we decide to cut a new breaking change release)

And `quick-protobuf` is released obviously :)

d-roak:
> And `quick-protobuf` is released obviously :)

Yep, the whole `ac209d474c0162d9b21dff40e82f4b7348fb81b6` commit should be reverted after `quick-protobuf` is released!

mxinden:
Thanks for the work here and upstream @jpldcarvalho and @thomaseizinger!

thomaseizinger:
I am removing this from the milestone because I don't want to block the release on it and we are still waiting for a release of `quick-protobuf`.

PR #3457, opened 2023-02-12
refactor(gossipsub): make `error` module private

### LINKED ISSUE #3392: `gossipsub`: make modules private
## Description
As described in [this comment](https://github.com/libp2p/rust-libp2p/pull/3303#discussion_r1084698192), some of the internal gossipsub modules are publicly exported.

Yes, when [3033](https://github.com/libp2p/rust-libp2p/pull/3303) is merged

StemCll:
Add error_priv module; export error enums from new module; make error private

## Description

Resolves https://github.com/libp2p/rust-libp2p/issues/3392.

As per this comment - https://github.com/libp2p/rust-libp2p/pull/3303#discussion_r1084698192

## Change checklist

### DIFF
--- a/protocols/gossipsub/src/error.rs
+++ b/protocols/gossipsub/src/error.rs
@@ -20,13 +20,13 @@
 
 #[deprecated(
     since = "0.44.0",
-    note = "Use `libp2p::gossipsub::PublishError` instead, as the `error` will become crate-private in the future."
+    note = "Use `libp2p::gossipsub::PublishError` instead, as the `error` module will become crate-private in the future."
 )]
 pub type PublishError = crate::error_priv::PublishError;
 
 #[deprecated(
     since = "0.44.0",
-    note = "Use `libp2p::gossipsub::SubscriptionError` instead, as the `error` will become crate-private in the future."
+    note = "Use `libp2p::gossipsub::SubscriptionError` instead, as the `error` module will become crate-private in the future."
 )]
 pub type SubscriptionError = crate::error_priv::SubscriptionError;
 
@@ -38,12 +38,12 @@ pub type GossipsubHandlerError = crate::error_priv::HandlerError;
 
 #[deprecated(
     since = "0.44.0",
-    note = "Use `libp2p::gossipsub::HandlerError` instead, as the `error` will become crate-private in the future."
+    note = "Use `libp2p::gossipsub::HandlerError` instead, as the `error` module will become crate-private in the future."
 )]
 pub type HandlerError = crate::error_priv::HandlerError;
 
 #[deprecated(
     since = "0.44.0",
-    note = "Use `libp2p::gossipsub::ValidationError` instead, as the `error` will become crate-private in the future."
+    note = "Use `libp2p::gossipsub::ValidationError` instead, as the `error` module will become crate-private in the future."
 )]
 pub type ValidationError = crate::error_priv::ValidationError;
--- a/protocols/gossipsub/src/error.rs
+++ b/protocols/gossipsub/src/error.rs
@@ -22,7 +22,7 @@ pub type GossipsubHandlerError = crate::error_priv::HandlerError;
 
 #[deprecated(
     since = "0.44.0",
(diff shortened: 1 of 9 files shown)


protocols/gossipsub/src/error.rs
```diff
@@ -1,153 +1,33 @@
-// Copyright 2020 Sigma Prime Pty Ltd.
+// TODO: which copyright here?
```
StemCll: Protocol labs, or other?

thomaseizinger: Given that this all new content, please use the Protocol Labs copyright header.

thomaseizinger (approved):
Thanks!

protocols/gossipsub/src/error.rs
```diff
@@ -18,136 +18,32 @@
...
-    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
-        match self {
-            Self::PublishError(err) => Some(err),
-            _ => None,
-        }
-    }
-}
+#[deprecated(
+    since = "0.44.0",
+    note = "Use `libp2p::gossipsub::PublishError` instead, as the `error` will become crate-private in the future."
```
thomaseizinger:
```suggestion
    note = "Use `libp2p::gossipsub::PublishError` instead, as the `error` module will become crate-private in the future."
```

protocols/gossipsub/src/error_priv.rs
```diff
@@ -0,0 +1,147 @@
...
+    /// Message source existed when validation has been sent to
+    /// [`crate::behaviour::MessageAuthenticity::Anonymous`].
+    MessageSourcePresent,
+    /// The data transformation failed.
+    TransformFailed,
+}
+
+impl std::fmt::Display for ValidationError {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        write!(f, "{self:?}")
```
thomaseizinger, line 131: We should fix this to be a better text and not just debug. Can be in a follow-up though!

protocols/gossipsub/src/handler.rs
```diff
@@ -18,7 +18,7 @@
 // FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
 // DEALINGS IN THE SOFTWARE.
 
-use crate::error::{HandlerError, ValidationError};
+use crate::error_priv::{HandlerError, ValidationError};
```
thomaseizinger:
```suggestion
use crate::{HandlerError, ValidationError};
```

I'd prefer if we import from the crate-root, same as what external users will do!

thomaseizinger (approved):
Nice, thanks!

Two more suggestions otherwise good to to in :)

protocols/gossipsub/src/error.rs
StemCll: :heavy_check_mark:

protocols/gossipsub/src/handler.rs
StemCll: changed all occurrences in gossipsub :)

protocols/gossipsub/src/error_priv.rs
StemCll: I've put something more, but not sure if that's what you mean :thinking:

thomaseizinger: Not quite. The debug print of the error will just render its name. The doc strings on the variants itself are actually quite good.

I'd prefer it in a separate PR though as it is unrelated to the module being private now.

Would you mind reverting that commit? Everything else looks 💯

StemCll: done :heavy_check_mark:

thomaseizinger: If you have a look at the doc strings on the variants, those would provide a much better error message than their debug print which will just print the variants name.

You can write a test where you `to_string` a variant to explore the difference :)

PR #3467, opened 2023-02-15
docs(roadmap): Update done projects and reprioritize

mxinden:
## Description

See individual commits.


rkuhn:
Where can I read up on the HTTP work that supposedly makes “streaming response” not worth it?

mxinden:
@rkuhn please see https://github.com/libp2p/specs/pull/508. Removing it here might have sent the wrong signal. I am saying that we need to reassess whether it makes sense, thus removing it from the roadmap.

thomaseizinger:
Additionally, with the simplifications coming to `ConnectionHandler`, I am hoping that implementing your own is so simple that you don't need another abstraction.

thomaseizinger (approved):

PR #3469, opened 2023-02-15
feat(ci): cache registry baseline rustdoc JSON

thomaseizinger:
## Description

The latest release (v0.18) of `cargo semver-checks` includes a [feature](https://github.com/obi1kenobi/cargo-semver-checks/pull/339) where the rustdoc JSON files for registry baselines is cached in `target/semver-checks/cache`. The rustdoc JSON files for released crates never changes as the source code on crates.io cannot be changed once a version is published. Thus, it is unnecessary to generate that JSON on every CI run.

We extract a separate action that installs `cargo semver-checks` for us and also sets up a cache. The cache is scoped to released version of the crate, meaning it automatically invalidates once we publish a new version.

This speeds up the `cargo semver-checks` step by 50% which is ~ 1 minute per job.

## Notes

In total, this allows our entire CI workflow to finish in as little as 11 minutes!

See https://github.com/libp2p/rust-libp2p/actions/runs/4258774854 for example.

~Draft because I am still testing this on my fork.~

## Change checklist


thomaseizinger:
@obi1kenobi Upgrading to `v0.18` results in a failing semver check despite this crate not having received any significant changes. Could you look into this?

thomaseizinger:
> @obi1kenobi Upgrading to `v0.18` results in a failing semver check despite this crate not having received any significant changes. Could you look into this?

This is the only diff since the last release, produced with `git diff v0.50.0 -- transports/uds`:

```diff
diff --git a/transports/uds/Cargo.toml b/transports/uds/Cargo.toml
index 9dd1321a..476888c8 100644
--- a/transports/uds/Cargo.toml
+++ b/transports/uds/Cargo.toml
@@ -12,15 +12,15 @@ categories = ["network-programming", "asynchronous"]
 
 [target.'cfg(all(unix, not(target_os = "emscripten")))'.dependencies]
 async-std = { version = "1.6.2", optional = true }
-libp2p-core = { version = "0.38.0", path = "../../core" }
+libp2p-core = { version = "0.39.0", path = "../../core" }
 log = "0.4.1"
-futures = "0.3.1"
+futures = "0.3.26"
 tokio = { version = "1.15", default-features = false, features = ["net"], optional = true }
 
 [target.'cfg(all(unix, not(target_os = "emscripten")))'.dev-dependencies]
 tempfile = "3.0"
 
-# Passing arguments to the docsrs builder in order to properly document cfg's. 
+# Passing arguments to the docsrs builder in order to properly document cfg's.
 # More information: https://docs.rs/about/builds#cross-compiling
 [package.metadata.docs.rs]
 all-features = true
```

obi1kenobi:
Confirmed bug, sorry about that. The query itself is fine, but the input rustdoc JSON is legitimately missing those structs.

We seem to have run into _yet another_ edge case in rustdoc generation.

obi1kenobi:
On a lighter note, I believe this moves the `rust-libp2p` repo into first place on our ~~integration tests~~ scoreboard for "bugs found in `cargo-semver-checks`" 😅

thomaseizinger:
> On a lighter note, I believe this moves the `rust-libp2p` repo into first place on our ~~integration tests~~ scoreboard for "bugs found in `cargo-semver-checks`" 😅

I guess somebody has to lead that board 😁

obi1kenobi:
We're getting ready to roll out a mini-crater for semver-checking so we can spot issues like these early (before releasing), so if it all goes well, you'll be in first place for a long time because it'll get a lot more difficult to find issues like this :)

obi1kenobi:
Heads-up that I'm about to release a new patch version with a bunch of CLI bugfixes sometime in the next ~hour. It might be good to adopt it ASAP since it's possible some of the bugs were causing false-negatives.

.github/actions/setup-cargo-semver-checks/action.yml
```diff
@@ -0,0 +1,7 @@
+name: "Setup cargo semver-checks"
+description: "Installs the cargo semver-checks tool"
+runs:
+  using: "composite"
+  steps:
+    - run: wget -q -O- https://github.com/obi1kenobi/cargo-semver-checks/releases/download/v0.18.2/cargo-semver-checks-x86_64-unknown-linux-gnu.tar.gz | tar -xz -C ~/.cargo/bin
```
obi1kenobi:
```suggestion
    - run: wget -q -O- https://github.com/obi1kenobi/cargo-semver-checks/releases/download/v0.18.3/cargo-semver-checks-x86_64-unknown-linux-gnu.tar.gz | tar -xz -C ~/.cargo/bin
```
v0.18.3 is released and all prior 0.18.x are yanked due to a CLI parsing bug that could lead to false-negatives

thomaseizinger: Thank you!

.github/actions/cargo-semver-checks/action.yml
```diff
@@ -0,0 +1,24 @@
...
+    - run: wget -q -O- https://github.com/obi1kenobi/cargo-semver-checks/releases/download/v0.18.3/cargo-semver-checks-x86_64-unknown-linux-gnu.tar.gz | tar -xz -C ~/.cargo/bin
+      shell: bash
+
+    - uses: Swatinem/rust-cache@359a70e43a0bb8a13953b04a90f76428b4959bb6 # v2.2.0
+      with:
+        shared-key: semver-checks-cache
+        save-if: ${{ github.ref_name == 'master' }}
+        workspaces: |
+          .
+          target/semver-checks
```
obi1kenobi: As far as I can tell, this doesn't help at all right now — it ends up rebuilding everything for the "current" rustdoc, and you can confirm this by setting `CARGO_TERM_VERBOSE=true` and observing that compilations are happening instead of being shown as skipped due to `Fresh` build data.

I know that cargo uses mtimes to decide whether to rebuild or not. I tried setting a synthetic mtime in the future on the restored build artifacts, but even that didn't seem to help: https://github.com/obi1kenobi/cargo-semver-checks/pull/399

I don't think I have the cycles to continue digging into the mtimes issue right now, but I thought I'd give you a heads-up in case you're able to dig further.

.github/actions/cargo-semver-checks/action.yml
```diff
@@ -0,0 +1,24 @@
+name: "Run cargo semver-checks"
+description: "Install and run the cargo semver-checks tool"
+inputs:
+  args:
+    required: true
+    description: "The arguments to pass to `cargo semver-checks`"
+runs:
+  using: "composite"
+  steps:
+    - run: wget -q -O- https://github.com/obi1kenobi/cargo-semver-checks/releases/download/v0.18.3/cargo-semver-checks-x86_64-unknown-linux-gnu.tar.gz | tar -xz -C ~/.cargo/bin
```
obi1kenobi, line 10: I haven't (yet) used it myself, but a few projects seem to have started using [taiki-e/install-action](https://github.com/marketplace/actions/install-development-tools) to fetch the latest `cargo-semver-checks` prebuilt binary automatically, without having to always keep bumping the version number in the `wget`.

It's all the same to me, I just thought it might be worth a look in case you haven't already heard of it.

thomaseizinger: I'd rather bump the version to be honest. Our CI has _a lot_ of moving parts so I prefer explicit control over the different versions of things interacting with each other.

My overarching goal is to optimize for development velocity: I don't mind sending a few more PRs if the overall turn-around time of a PR is super short. That also gives us a clean track record of what has changed and issues that come up can (hopefully) be fixed quicker :)

obi1kenobi: Makes sense!

.github/actions/cargo-semver-checks/action.yml
thomaseizinger: Hmm, thanks for that! Yeah I did wonder why this isn't working as expected. I may resort to only caching the registry JSONs files for now.

It is a bit annoying because this job by itself takes almost an hour for us and would run on every PR merged into master with most of the compute being wasted if I can't cache the debug symbols for all the crates.

obi1kenobi: Yeah, I feel that :( I opened https://github.com/obi1kenobi/cargo-semver-checks/issues/401 for this.

I'm not sure how to pull this off while `cargo-semver-checks` is generating the "current" rustdoc itself, but if generating the current rustdoc separately is on the table then I wrote up an idea in that issue for how that might work. If you're open to exploring it together, I'd be happy to help.

thomaseizinger:
Ah damn, now I understand why my workflow wasn't working as expected. I was still working with `0.16` locally where there was still a `target` directory ...

thomaseizinger:
@obi1kenobi I am reasonably happy with the solution now. It uses almost no storage space in the GitHub cache, invalidates correctly on new releases and saves us 50% compute while running `cargo semver-checks`. Perhaps it is useful for v2 of your action!

.github/actions/cargo-semver-checks/action.yml
```diff
@@ -0,0 +1,28 @@
...
+      shell: bash
+      id: get-released-version
+      run: |
+        MAX_STABLE_VERSION=$(curl https://crates.io/api/v1/crates/${{ inputs.crate }} --silent | jq '.crate.max_stable_version')
+        echo "version=${MAX_STABLE_VERSION}" >> $GITHUB_OUTPUT
+
+    - uses: actions/cache@v3
+      with:
+        path: ${{ github.workspace }}/target/semver-checks/cache
+        key: semver-checks-cache-${{ inputs.crate }}-steps.get-released-version.outputs.version
```
obi1kenobi:
```suggestion
        key: semver-checks-cache-${{ inputs.crate }}-${{ steps.get-released-version.outputs.version }}
```
Did you mean to evaluate this, perhaps?

thomaseizinger: I definitely did haha

thomaseizinger:
> @obi1kenobi I am reasonably happy with the solution now. It uses almost no storage space in the GitHub cache, invalidates correctly on new releases and saves us 50% compute while running `cargo semver-checks`. Perhaps it is useful for v2 of your action!

I just had a dirty idea. I am trying to copy in the workspace's target directory and see if I can at least reuse some of the build artifacts from the dependencies.

obi1kenobi:
Setting `CARGO_TERM_VERBOSE=true` as an env var when running `cargo-semver-checks` will show whether cargo is rebuilding (`Running rustdoc ...`) or using existing artifacts (`Fresh <artifact-name>`).

Hoping it works! 🤞

thomaseizinger:
I tried to set that: https://github.com/libp2p/rust-libp2p/actions/runs/4259376112/jobs/7411549615#step:13:73

thomaseizinger:
Okay, this is ready for review again. For now, we are only caching the registry baseline JSON which already gives us a good speedup. Further improvements are tracked in `cargo semver-checks`: https://github.com/obi1kenobi/cargo-semver-checks/issues/401

obi1kenobi:
> I tried to set that: https://github.com/libp2p/rust-libp2p/actions/runs/4259376112/jobs/7411549615#step:13:73

Ah, sorry — you'll also need to pass `--verbose` to `cargo-semver-checks`, so it allows cargo build output to pass through. Otherwise it just swallows the verbose cargo output.

.github/actions/cargo-semver-checks/action.yml
```diff
@@ -0,0 +1,28 @@
...
+      run: |
+        MAX_STABLE_VERSION=$(curl https://crates.io/api/v1/crates/${{ inputs.crate }} --silent | jq '.crate.max_stable_version')
+        echo "version=${MAX_STABLE_VERSION}" >> $GITHUB_OUTPUT
+
+    - uses: actions/cache@v3
+      with:
+        path: ${{ github.workspace }}/target/semver-checks/cache
+        key: semver-checks-cache-${{ inputs.crate }}-${{ steps.get-released-version.outputs.version }}
+
+    - run: cargo semver-checks check-release --package ${{ inputs.crate }}
```
thomaseizinger:
```suggestion
    - run: cargo semver-checks check-release --package ${{ inputs.crate }} --verbose
```

thomaseizinger:
Merging because it is likely to be uncontroversial and speeds up our CI.

PR #3474, opened 2023-02-15
feat(kad): reuse outbound streams instead of closing them

nazar-pc:
## Description

This implements outgoing stream reuse in Kademlia by storing substream for future reuse after message was processed successfully instead of closing substream right away.

## Notes

This is, unfortunately not working properly and after spending a few hours on it I could use external help.

This will be useful for https://github.com/libp2p/rust-libp2p/pull/3468 (regardless of whatever version of it ends up landing).

## Open Questions

None

## Change checklist

### DIFF
--- a/protocols/kad/CHANGELOG.md
+++ b/protocols/kad/CHANGELOG.md
@@ -10,8 +10,11 @@
 
 - Bump MSRV to 1.65.0.
 
+- Outgoing stream reuse. See [PR 3474].
+
 [PR 3239]: https://github.com/libp2p/rust-libp2p/pull/3239
 [PR 3287]: https://github.com/libp2p/rust-libp2p/pull/3287
+[PR 3474]: https://github.com/libp2p/rust-libp2p/pull/3474
 
 # 0.42.1
 
--- a/protocols/kad/CHANGELOG.md
+++ b/protocols/kad/CHANGELOG.md
@@ -10,7 +10,7 @@
 
 - Bump MSRV to 1.65.0.
 
-- Outgoing stream reuse. See [PR 3474].
+- Reuse outbound streams instead of closing them after one message. See [PR 3474].
 
 [PR 3239]: https://github.com/libp2p/rust-libp2p/pull/3239
 [PR 3287]: https://github.com/libp2p/rust-libp2p/pull/3287
--- a/protocols/kad/src/handler.rs
+++ b/protocols/kad/src/handler.rs
@@ -854,7 +854,7 @@ where
             }
         }
 
-        return Err((msg, user_data));
+        Err((msg, user_data))
     }
 }
 
--- a/protocols/kad/src/handler.rs
+++ b/protocols/kad/src/handler.rs
@@ -768,30 +768,17 @@ where
             return Poll::Ready(event);
         }
 
-        'outer: while self.outbound_substreams.len() + self.num_requested_outbound_streams
+        while self.outbound_substreams.len() + self.num_requested_outbound_streams
             < MAX_NUM_SUBSTREAMS
         {
             if let Some((msg, user_data)) = self.pending_substream_requests.pop_front() {
-                // Search for outbound substream waiting to be reused first.
-                for outbound_substream in self.outbound_substreams.iter_mut() {
-                    match std::mem::replace(outbound_substream, OutboundSubstreamState::Poisoned) {
-                        OutboundSubstreamState::Idle(substream, waker) => {
-                            *outbound_substream =
-                                OutboundSubstreamState::PendingSend(substream, msg, user_data);
-                            waker.wake();
-                            continue 'outer;
-                        }
-                        other => {
-                            *outbound_substream = other;
-                        }
-                    }
+                if let Err(info) = self.try_reuse_outbound_stream(msg, user_data) {
+                    // If reuse wasn't possible, request new substream.
+                    self.num_requested_outbound_streams += 1;
+                    return Poll::Ready(ConnectionHandlerEvent::OutboundSubstreamRequest {
+                        protocol: SubstreamProtocol::new(self.config.protocol_config.clone(), info),
+                    });
                 }
-
-                // If not found request new substream.
-                self.num_requested_outbound_streams += 1;
-                let protocol =
-                    SubstreamProtocol::new(self.config.protocol_config.clone(), (msg, user_data));
-                return Poll::Ready(ConnectionHandlerEvent::OutboundSubstreamRequest { protocol });
             } else {
                 break;
             }
@@ -846,6 +833,28 @@ where
         }
 
         debug_assert!(false, "Cannot find inbound substream for {request_id:?}")
+    }
+
+    fn try_reuse_outbound_stream(
+        &mut self,
+        msg: KadRequestMsg,
+        user_data: Option<TUserData>,
+    ) -> Result<(), (KadRequestMsg, Option<TUserData>)> {
+        // Search for idle outbound substream waiting to be reused.
+        for state in self.outbound_substreams.iter_mut() {
+            match std::mem::replace(state, OutboundSubstreamState::Poisoned) {
+                OutboundSubstreamState::Idle(substream, waker) => {
+                    *state = OutboundSubstreamState::PendingSend(substream, msg, user_data);
+                    waker.wake();
+                    return Ok(());
+                }
+                other => {
+                    *state = other;
+                }
+            }
+        }
+
+        return Err((msg, user_data));
     }
 }
 
--- a/protocols/kad/src/handler.rs
+++ b/protocols/kad/src/handler.rs
@@ -97,10 +97,12 @@ pub struct KademliaHandler<TUserData> {
     /// Number of outbound streams being upgraded right now.
     num_requested_outbound_streams: usize,
 
-    /// List of outbound substreams that are waiting to become active next.
+    /// List of outbound substreams that are waiting to be created next.
     /// Contains the request we want to send, and the user data if we expect an answer.
-    requested_streams:
-        VecDeque<SubstreamProtocol<KademliaProtocolConfig, (KadRequestMsg, Option<TUserData>)>>,
+    pending_substream_requests: VecDeque<(KadRequestMsg, Option<TUserData>)>,
+
+    /// Outgoing substreams available for future reuse.
+    reusable_outgoing_substreams: Vec<KadOutStreamSink<NegotiatedSubstream>>,
 
     /// List of active inbound substreams with the state they are in.
     inbound_substreams: SelectAll<InboundSubstreamState<TUserData>>,
@@ -161,8 +163,8 @@ enum OutboundSubstreamState<TUserData> {
     WaitingAnswer(KadOutStreamSink<NegotiatedSubstream>, TUserData),
     /// An error happened on the substream and we should report the error to the user.
     ReportError(KademliaHandlerQueryErr, TUserData),
-    /// The substream is being closed.
-    Closing(KadOutStreamSink<NegotiatedSubstream>),
+    /// The substream is available for future reuse.
+    Available(KadOutStreamSink<NegotiatedSubstream>),
     /// The substream is complete and will not perform any more work.
     Done,
     Poisoned,
@@ -352,6 +354,12 @@ pub enum KademliaHandlerEvent<TUserData> {
         /// The user data passed to the `PutValue`.
         user_data: TUserData,
     },
+
+    /// The substream is available for future reuse.
+    Available {
+        /// Substream to be reused for outgoing requests.
+        substream: KadOutStreamSink<NegotiatedSubstream>,
+    },
 }
 
 /// Error that can happen when requesting an RPC query.
@@ -531,7 +539,8 @@ where
             inbound_substreams: Default::default(),
             outbound_substreams: Default::default(),
             num_requested_outbound_streams: 0,
-            requested_streams: Default::default(),
+            pending_substream_requests: Default::default(),
+            reusable_outgoing_substreams: Default::default(),
             keep_alive,
             protocol_status: ProtocolStatus::Unconfirmed,
         }
@@ -677,10 +686,8 @@ where
             }
             KademliaHandlerIn::FindNodeReq { key, user_data } => {
                 let msg = KadRequestMsg::FindNode { key };
-                self.requested_streams.push_back(SubstreamProtocol::new(
-                    self.config.protocol_config.clone(),
-                    (msg, Some(user_data)),
-                ));
+                self.pending_substream_requests
+                    .push_back((msg, Some(user_data)));
             }
             KademliaHandlerIn::FindNodeRes {
                 closer_peers,
@@ -688,10 +695,8 @@ where
             } => self.answer_pending_request(request_id, KadResponseMsg::FindNode { closer_peers }),
             KademliaHandlerIn::GetProvidersReq { key, user_data } => {
                 let msg = KadRequestMsg::GetProviders { key };
-                self.requested_streams.push_back(SubstreamProtocol::new(
-                    self.config.protocol_config.clone(),
-                    (msg, Some(user_data)),
-                ));
+                self.pending_substream_requests
+                    .push_back((msg, Some(user_data)));
             }
             KademliaHandlerIn::GetProvidersRes {
                 closer_peers,
@@ -706,24 +711,17 @@ where
             ),
             KademliaHandlerIn::AddProvider { key, provider } => {
                 let msg = KadRequestMsg::AddProvider { key, provider };
-                self.requested_streams.push_back(SubstreamProtocol::new(
-                    self.config.protocol_config.clone(),
-                    (msg, None),
-                ));
+                self.pending_substream_requests.push_back((msg, None));
             }
             KademliaHandlerIn::GetRecord { key, user_data } => {
                 let msg = KadRequestMsg::GetValue { key };
-                self.requested_streams.push_back(SubstreamProtocol::new(
-                    self.config.protocol_config.clone(),
-                    (msg, Some(user_data)),
-                ));
+                self.pending_substream_requests
+                    .push_back((msg, Some(user_data)));
             }
             KademliaHandlerIn::PutRecord { record, user_data } => {
                 let msg = KadRequestMsg::PutValue { record };
-                self.requested_streams.push_back(SubstreamProtocol::new(
-                    self.config.protocol_config.clone(),
-                    (msg, Some(user_data)),
-                ));
+                self.pending_substream_requests
+                    .push_back((msg, Some(user_data)));
             }
             KademliaHandlerIn::GetRecordRes {
                 record,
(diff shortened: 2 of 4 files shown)


nazar-pc:
I think this is ready to review, please take a look. Changes are relatively straightforward.

protocols/kad/src/handler.rs
```diff
@@ -970,13 +989,12 @@ where
...
-                OutboundSubstreamState::Closing(mut stream) => match stream.poll_close_unpin(cx) {
-                    Poll::Ready(Ok(())) | Poll::Ready(Err(_)) => return Poll::Ready(None),
-                    Poll::Pending => {
-                        *this = OutboundSubstreamState::Closing(stream);
-                        return Poll::Pending;
-                    }
-                },
+                OutboundSubstreamState::Available(substream) => {
+                    *this = OutboundSubstreamState::Done;
+                    let event = KademliaHandlerEvent::Available { substream };
```
thomaseizinger: Instead of returning the stream here, can we just remain in `Available` and search all streams for one of these before we open a new one (assuming we are still within the limit)?

That should avoid the `unreachable` case in the behaviour I think.

protocols/kad/src/protocol.rs
```diff
@@ -250,15 +251,55 @@ where
...
                     future::ready(proto_to_resp_msg(response))
                 }),
-        )
+        ))
     }
 }
 
 /// Sink of responses and stream of requests.
-pub type KadInStreamSink<S> = KadStreamSink<S, KadResponseMsg, KadRequestMsg>;
+pub struct KadInStreamSink<S>(KadStreamSink<S, KadResponseMsg, KadRequestMsg>);
```
thomaseizinger: Why do we need this wrapper type?

protocols/kad/src/handler.rs
```diff
@@ -161,8 +163,8 @@ enum OutboundSubstreamState<TUserData> {
     WaitingAnswer(KadOutStreamSink<NegotiatedSubstream>, TUserData),
     /// An error happened on the substream and we should report the error to the user.
     ReportError(KademliaHandlerQueryErr, TUserData),
-    /// The substream is being closed.
-    Closing(KadOutStreamSink<NegotiatedSubstream>),
+    /// The substream is available for future reuse.
+    Available(KadOutStreamSink<NegotiatedSubstream>),
```
thomaseizinger: I think I'd prefer `Idle` as the wording here.

protocols/kad/CHANGELOG.md
```diff
@@ -10,8 +10,11 @@
 
 - Bump MSRV to 1.65.0.
 
+- Outgoing stream reuse. See [PR 3474].
```
thomaseizinger:
```suggestion
- Reuse outbound streams instead of closing them after one message. See [PR 3474].
```

protocols/kad/src/handler.rs
```diff
@@ -772,20 +770,41 @@ where
...
-            self.outbound_substreams.len() + self.num_requested_outbound_streams;
-        if num_in_progress_outbound_substreams < MAX_NUM_SUBSTREAMS {
-            if let Some(protocol) = self.requested_streams.pop_front() {
-                self.num_requested_outbound_streams += 1;
-                return Poll::Ready(ConnectionHandlerEvent::OutboundSubstreamRequest { protocol });
+        while !self.pending_substream_requests.is_empty()
+            && self.outbound_substreams.len() + self.num_requested_outbound_streams
+                < MAX_NUM_SUBSTREAMS
+        {
+            if let Some((msg, user_data)) = self.pending_substream_requests.pop_front() {
```
thomaseizinger: You could omit `pending_substream_requests` from the first condition here and instead `break` from the loop in an `else` block of the inner `if` (or change it to a `match` with a `break` on `None`).

protocols/kad/src/handler.rs
```diff
@@ -772,20 +770,41 @@ where
...
+                } else {
+                    self.num_requested_outbound_streams += 1;
+                    let protocol = SubstreamProtocol::new(
+                        self.config.protocol_config.clone(),
+                        (msg, user_data),
+                    );
+                    return Poll::Ready(ConnectionHandlerEvent::OutboundSubstreamRequest {
+                        protocol,
+                    });
+                }
```
thomaseizinger: Like I mentioned in the other comment, I'd prefer if we just search `outbound_substreams` for one that is `Available` and have it mutate itself to `PendingSend` with the requested message.

We do something similar here somewhere already.

thomaseizinger: We can do that unconditionally of the number of streams as a first step in here and if there are messages left, we can open new ones up to 32.

Let me know what you think!

thomaseizinger (approved):

protocols/kad/src/handler.rs
nazar-pc: I considered that, you mentioned such approach in https://github.com/libp2p/rust-libp2p/discussions/3411. I didn't like it because it is a hot path and it should be slower to iterate streams in `SelectAll` one by one rather than taking one deterministically from a stored vector. I don't have a very strong opinion here though if that is what you'd like to do anyway.

protocols/kad/src/handler.rs
nazar-pc: > We do something similar here somewhere already.

Yes, there is such pattern for incoming streams reuse already, there is an iteration there searching for streams waiting for subsequent incoming messages. But it felt less efficient searching things in `SelectAll` as I commented in the other place.

protocols/kad/src/protocol.rs
nazar-pc: This is because `KadOutStreamSink` is now stored in `KademliaHandlerEvent::Available` which implements `Debug`, thus `KadOutStreamSink` should as well. `KadInStreamSink` was changed similarly just to be consistent.

protocols/kad/src/handler.rs
thomaseizinger: We are at most searching 32 elements here. I'd be surprised if we can measure this optimization. For example, pushing a new item into `SelectAll` [unconditionally allocates memory](https://docs.rs/futures-util/0.3.26/src/futures_util/stream/futures_unordered/mod.rs.html#167-176) which I think we avoid with the search and overwrite operation?

For consistency and simplicity (YMMV) I'd suggest to use the search strategy.

If you do feel strongly, then I'd like to at least avoid the `unreachable` case. For that, we'd need to refactor `<OutboundSubstreamState as Stream>::Item` to return a custom enum: either a stream that is now available or the event to be returned.

Regardless of the above, `<OutboundSubstreamState as Stream>::Item` can reference `KademliaHandlerEvent` directly as we currently return `ConnectionHandlerEvent::Custom` from all codepaths.

protocols/kad/src/protocol.rs
thomaseizinger: I'd like to remove that event variant in any case so this can be reverted then, correct?

nazar-pc:
Applied suggestions, diff is much smaller now.

While it is easy to read, it feels a bit wrong to swap something with poisoned value and replace back with itself in case it didn't match. Such pattern is used in other places, so I adopted it for consistency, but memory copies right and left during iteration just to simplify things might be a bit less efficient than checking for match first and only update value if necessary. On the other hand maybe compiler is smart enough to optimize it away since the code is quite obvious, I didn't benchmark to compare.

protocols/kad/src/handler.rs
```diff
@@ -780,12 +768,32 @@ where
...
+                            *outbound_substream =
+                                OutboundSubstreamState::PendingSend(substream, msg, user_data);
+                            waker.wake();
+                            continue 'outer;
+                        }
+                        other => {
+                            *outbound_substream = other;
+                        }
+                    }
+                }
```
thomaseizinger: For consistency, can we extract this into a fn like `try_answer_request`?

That should also allow us to get rid of the loop markers.

protocols/kad/src/handler.rs
```diff
@@ -780,12 +768,32 @@ where
             return Poll::Ready(event);
         }
 
-        let num_in_progress_outbound_substreams =
-            self.outbound_substreams.len() + self.num_requested_outbound_streams;
```
thomaseizinger, line 784: Let's retain this variable please.

thomaseizinger (commented):
One more request, otherwise LGTM.

I'd also like @mxinden to approve this. The gist is: By reusing substreams here as the spec allows, we can make use of yamux's backpressure because a stream's buffer will eventually fill up if the remote is slower in reading `AddProvider` messages than we are with writing them.

Together with a solution for backpressuring how fast we read new messages from the stream, this should allow us to saturate the network or at least get us closer to it.

protocols/kad/src/handler.rs
nazar-pc: Then we'll be able to only process a single outbound stream request per poll (`while` will have to change to `if`). Is that okay?

thomaseizinger:
> Applied suggestions, diff is much smaller now.
> 
> While it is easy to read, it feels a bit wrong to swap something with poisoned value and replace back with itself in case it didn't match. Such pattern is used in other places, so I adopted it for consistency, but memory copies right and left during iteration just to simplify things might be a bit less efficient than checking for match first and only update value if necessary. On the other hand maybe compiler is smart enough to optimize it away since the code is quite obvious, I didn't benchmark to compare.

I think we could avoid the copies by first doing a `find` on the list and only swapping the one that `matches!(s, OutboundSubstreamState::Idle)`. That also creates an `unreachable` path but that one would at least be fairly local.

Feel free to leave as is though.

protocols/kad/src/handler.rs
thomaseizinger: Ah right, my bad. Thanks for pointing this out! We should process all available ones, yes.

thomaseizinger (dismissed):
Clippy fails, otherwise LGTM.

thomaseizinger:
Waiting for @mxinden now.

mxinden (changes requested):
I am sorry for only reviewing this change now. Thanks for the proposal. @nazar-pc can you please expand on the benefit of this change?

> The gist is: By reusing substreams here as the spec allows, we can make use of yamux's backpressure because a stream's buffer will eventually fill up if the remote is slower in reading AddProvider messages than we are with writing them.

The default initial Yamux stream receive window is [256 KiB](https://github.com/libp2p/rust-yamux/blob/72ccd5734f71971fc1f02ac4a5f43b8eb4dff660/yamux/src/lib.rs#L99). With this change there are never more than one Kademlia request in-flight at once on a single stream. I don't think a single Kademlia request ever exceeds the stream receive window size (256 KiB). Thus  I don't think the Yamux backpressure mechanism will ever kick in.

> Together with a solution for backpressuring how fast we read new messages from the stream, this should allow us to saturate the network or at least get us closer to it.

Looking at this from a Kademlia perspective only, I would be surprised if one could saturate a link with Kademlia requests only. Though that is only an intuition.

I would like to go away from reusing streams across all libp2p protocols. Muxers should provide backpressure on new-stream creation (e.g. QUIC does). A consumer should backpressure on the number of inflight requests, which would equal the number of open streams, through the muxer implementation. For per-protocol backpressure see https://github.com/libp2p/specs/issues/520.

thomaseizinger:
> > The gist is: By reusing substreams here as the spec allows, we can make use of yamux's backpressure because a stream's buffer will eventually fill up if the remote is slower in reading AddProvider messages than we are with writing them.
> 
> The default initial Yamux stream receive window is [256 KiB](https://github.com/libp2p/rust-yamux/blob/72ccd5734f71971fc1f02ac4a5f43b8eb4dff660/yamux/src/lib.rs#L99). With this change there are never more than one Kademlia request in-flight at once on a single stream.

Incorrect, the opposite is the intention. `AddProvider` does not have a reponse so the state machine is complete after sending the message. With this change, we will instantly reuse the stream and eventually fill up the buffer if the other party is slower than us.

> I would like to go away from reusing streams across all libp2p protocols. Muxers should provide backpressure on new-stream creation (e.g. QUIC does). A consumer should backpressure on the number of inflight requests, which would equal the number of open streams, through the muxer implementation. For per-protocol backpressure see https://github.com/libp2p/specs/issues/520.

I would like to move away from it too. But it is going to take a while before we can design and ship backpressure via status codes.

I would argue that it doesn't matter how we ship the feature. We have a spec-compliant solution here. Once we have backpressure via error codes or something else working, we should be able to transparently (for the user) change the implementation.

This gives us the best of both worlds: A working solution today and buys us time for designing something more general.

nazar-pc:
> I would like to go away from reusing streams across all libp2p protocols. Muxers should provide backpressure on new-stream creation (e.g. QUIC does). A consumer should backpressure on the number of inflight requests, which would equal the number of open streams, through the muxer implementation. For per-protocol backpressure see https://github.com/libp2p/specs/issues/520.

I'm curious, what is the benefit of re-creating streams over and over again rather than creating once and using it for as long as necessary? Sounds less efficient and more tricky to manage.

mxinden:
> > > The gist is: By reusing substreams here as the spec allows, we can make use of yamux's backpressure because a stream's buffer will eventually fill up if the remote is slower in reading AddProvider messages than we are with writing them.
> > 
> > 
> > The default initial Yamux stream receive window is [256 KiB](https://github.com/libp2p/rust-yamux/blob/72ccd5734f71971fc1f02ac4a5f43b8eb4dff660/yamux/src/lib.rs#L99). With this change there are never more than one Kademlia request in-flight at once on a single stream.
> 
> Incorrect, the opposite is the intention. `AddProvider` does not have a reponse so the state machine is complete after sending the message. With this change, we will instantly reuse the stream and eventually fill up the buffer if the other party is slower than us.

Good point.

> > I would like to go away from reusing streams across all libp2p protocols. Muxers should provide backpressure on new-stream creation (e.g. QUIC does). A consumer should backpressure on the number of inflight requests, which would equal the number of open streams, through the muxer implementation. For per-protocol backpressure see [libp2p/specs#520](https://github.com/libp2p/specs/issues/520).
> 
> I would like to move away from it too. But it is going to take a while before we can design and ship backpressure via status codes.

I don't think we need status codes. `libp2p-quic` (i.e. quinn) gives us a way to limit the number of inbound streams, e.g. [I think the default is 100](https://docs.rs/quinn-proto/0.9.2/src/quinn_proto/config.rs.html#258). As long as we don't drop an inbound stream, the remote is eventually backpressured by that limit if we use a new stream per request.

> > I would like to go away from reusing streams across all libp2p protocols. Muxers should provide backpressure on new-stream creation (e.g. QUIC does). A consumer should backpressure on the number of inflight requests, which would equal the number of open streams, through the muxer implementation. For per-protocol backpressure see [libp2p/specs#520](https://github.com/libp2p/specs/issues/520).
> 
> I'm curious, what is the benefit of re-creating streams over and over again rather than creating once and using it for as long as necessary?

When reusing streams, how does a producer know how many concurrent requests it can send? In other words, how does it know how many streams it can open and keep open? I would prefer not to keep the magic number `32` that we have today.

When not reusing streams, the consumer provides a dynamic limit on the number of streams via the multiplexer stream-creation backpressure mechanism. I.e. by keeping streams alive on the inbound side, the outbound side is backpressured at creating new outbound streams, i.e. new outbound requests.

When reusing streams and sending requests without a response (here the `AddProvider` case), how do you prevent head-of-line blocking?

When not reusing streams, the underlying transport can do retransmits on per-stream basis, thus one request is not head-of-line blocked by another.

> Sounds less efficient

I would treat streams as free, at least from a networking perspective. What makes you think they are expensive?

nazar-pc:
> I would prefer not to keep the magic number 32 that we have today.

Fair, I'm certainly onboard with this.

> When reusing streams and sending requests without a response (here the AddProvider case), how do you prevent head-of-line blocking?
>
> When not reusing streams, the underlying transport can do retransmits on per-stream basis, thus one request is not head-of-line blocked by another.

I'm not as familiar with implementation, but I assumed that the whole point of streams within the same TCP connection is that one stream doesn't block another. Otherwise what is the point?

> I would treat streams as free, at least from a networking perspective. What makes you think they are expensive?

Just intuitively, if feels like some sort of negotiation is necessary, so when we're talking about tens or hundreds of concurrent requests on each peer for hours it adds up. Especially because messages are small, overhead will be percentage-wise larger. I might be completely wrong though, I'm not familiar with implementation details yet.

thomaseizinger:
> > > I would like to go away from reusing streams across all libp2p protocols. Muxers should provide backpressure on new-stream creation (e.g. QUIC does). A consumer should backpressure on the number of inflight requests, which would equal the number of open streams, through the muxer implementation. For per-protocol backpressure see [libp2p/specs#520](https://github.com/libp2p/specs/issues/520).
> > 
> > I would like to move away from it too. But it is going to take a while before we can design and ship backpressure via status codes.
> 
> I don't think we need status codes. `libp2p-quic` (i.e. quinn) gives us a way to limit the number of inbound streams, e.g. [I think the default is 100](https://docs.rs/quinn-proto/0.9.2/src/quinn_proto/config.rs.html#258). As long as we don't drop an inbound stream, the remote is eventually backpressured by that limit if we use a new stream per request.

But our quic implementation is in alpha status and thus not recommendes for production I guess?

In general, I don't think these efforts are incompatible with each other.

Until we have a back-pressure mechanism across all transports and muxers, we will need to keep the limit to mitigate DoS attacks. As long as we have the limit, we will drop new inbound streams upon bursts of requests where there isn't a response to requests because the client-side limit of 32 gets "out of sync" with the server side.

Why can't we provide backpressure through reuse in Kademlia and switch to backpressure through new streams at a later point? Both are spec compliant because streams CAN be reused but don't have to.

thomaseizinger:
> > When reusing streams and sending requests without a response (here the AddProvider case), how do you prevent head-of-line blocking?
> >
> > When not reusing streams, the underlying transport can do retransmits on per-stream basis, thus one request is not head-of-line blocked by another.
> 
> I'm not as familiar with implementation, but I assumed that the whole point of streams within the same TCP connection is that one stream doesn't block another. Otherwise what is the point?

Streams avoid HoL blocking between each other. Once you start to pipeline requests within a stream, you have HoL blocking again. In fact, the HoL blocking is desireable in this case because it creates a form of backpressure.

> > I would treat streams as free, at least from a networking perspective. What makes you think they are expensive?
> 
> Just intuitively, if feels like some sort of negotiation is necessary, so when we're talking about tens or hundreds of concurrent requests on each peer for hours it adds up. Especially because messages are small, overhead will be percentage-wise larger. I might be completely wrong though, I'm not familiar with implementation details yet.

No negotiation is necessary for new streams. QUIC operates on a locally-known budget. Yamux only sets a flag with a new payload frame to indicate a new stream. Thus, the only overhead is the framing of the muxer which applies to new and existing streams as well.

Together with `upgrade::Version::V1Lazy`, the `multistream-select` handshake for new streams is 0 RTT but has a few bytes overhead. That is a small price to pay for the benefits of dynamically scaling the use of resources.

mxinden:
Closing here as I think we reached consensus that reusing a stream is not the way forward. Please comment in case you disagree.

PR #3484, opened 2023-02-21
feat(ci): lint against usages of variables with underscore

thomaseizinger:
## Description

Prefixing a variable with an underscore (`_`) in Rust indicates that it is not used. Through refactorings, it can sometimes happen that we do end up using such a variable. In this case, the underscore should be removed.

Clippy can help us with this.

## Notes

Triggered by @mxinden's review in https://github.com/libp2p/rust-libp2p/pull/3254#discussion_r1111807888.

## Change checklist


mxinden (approved):
:pray: always thinking one step ahead.

thomaseizinger:
And it found more usages of variables with underscore!

PR #3495, opened 2023-02-22
refactor(gossipsub): make protocol module private

### LINKED ISSUE #3392: `gossipsub`: make modules private

StemCll:

Resolves [3392](https://github.com/libp2p/rust-libp2p/issues/3392)

## Change checklist


protocols/gossipsub/src/protocol.rs
```diff
@@ -18,640 +18,20 @@
...
+    since = "0.44.0",
+    note = "Use `libp2p::gossipsub::ProtocolId` instead, as the `protocol` module will become crate-private in the future."
+)]
+pub type ProtocolId = crate::protocol_priv::ProtocolId;
+
+#[deprecated(
+    since = "0.44.0",
+    note = "Use `libp2p::gossipsub::GossipsubCodec` instead, as the `protocol` module will become crate-private in the future."
+)]
+pub type GossipsubCodec = crate::protocol_priv::GossipsubCodec;
```
thomaseizinger: Do we need to publicly expose these at all? In other words, do they appear in the public API elsewhere and thus should be accessible?

I could imagine that all of these are actually implementation details and could be made private (eventually). In a first step, we'd obv deprecate them but without a replacement. That also gives users time to "complain" if they need them for whatever reason.

StemCll: > I could imagine that all of these are actually implementation details and could be made private (eventually).

+1. I'm not sure if these should be accessible publicly, but I've looked to [a documentation](https://docs.rs/libp2p/latest/libp2p/gossipsub/protocol/index.html) and they're listed. Thus I've exported them in the root. 


So if I understand correctly, should I revert the changes regarding `protocol_priv` and `protocol`, and just put the deprecation warns in `protocol` module?

thomaseizinger: > > I could imagine that all of these are actually implementation details and could be made private (eventually).
> 
> +1. I'm not sure if these should be accessible publicly, but I've looked to [a documentation](https://docs.rs/libp2p/latest/libp2p/gossipsub/protocol/index.html) and they're listed. Thus I've exported them in the root.

Yep, they currently are but I am questioning whether they should be!

My rationale (which applies to all protocols really) is that I want the public API of a given `libp2p` module to be as small as possible. The only API a user should interact with is the `NetworkBehaviour` implementation with its accompanying types. That is:

- The `NetworkBehaviour::OutEvent`
- Any type within public functions of `impl Behaviour {}`, for example `MessageAuthenticity` or `Config`

Everything else, like the `ConnectionHandler` or the messages from the `NetworkBehaviour` to `ConnectionHandler` don't need to be exposed. This will allow us to change more things without having to break the public API and cause a cascade of version bumps for downstream users.

My vision is that the public API is only the root module, with no other modules being accessible such that `use libp2p::gossipsub` is all you need to do and any relevant type can be referenced via `gossipsub::`, e.g. `gossipsub::Event`.

StemCll: Sure, perfectly understandable. So for now I'll add the deprecation messages and leave the mod public.

thomaseizinger: Thank you!

StemCll: I've pushed the changes, however I'm not a fan of using `allow` attribute in the lib code. I couldn't find a better solution though

thomaseizinger: I agree, this isn't very nice.

I think we can solve this by:

- Renaming the `protocol` module to `protocol_internal` or `protocol_priv` that is `pub(crate)`.
- Re-export its contents in the root module like: `#[deprecated] pub mod protocol { pub use crate::protocol_priv::*; }`
- Import all types within the crate from `protocol_priv`.

That will allow us to import the types without deprecation warnings within the crate but still have the warnings if users import them.

What do you think?

tcoratger:
@thomaseizinger I think this one is closed by #3777 or is there something specific left to do here? I think it is also valid for PR #3564, #3565, #3566, what do you think?

thomaseizinger:
That is correct, good catch, thank you!

PR #3504, opened 2023-02-24
fix(kad): remove `allow_listening` handler config

### LINKED ISSUE #3499: kad: remove `allow_listening` handler config
_Originally posted by @mxinden in https://github.com/libp2p/rust-libp2p/pull/3254#discussion_r1111806268_

The `allow_listening` config option in Kademlia does nothing as it is always set to `true`. We should remove it.

chirag-bgh:
## Description
The ```allow_listening``` config option in Kademlia does nothing as it is always set to ```true```. Thus a dummy configuration option and should be removed.

Closes #3499.

## Change checklist


protocols/kad/src/handler.rs
```diff
@@ -618,12 +612,7 @@ where
     type InboundOpenInfo = ();
 
     fn listen_protocol(&self) -> SubstreamProtocol<Self::InboundProtocol, Self::InboundOpenInfo> {
-        if self.config.allow_listening {
-            SubstreamProtocol::new(self.config.protocol_config.clone(), ())
-                .map_upgrade(Either::Left)
-        } else {
-            SubstreamProtocol::new(Either::Right(upgrade::DeniedUpgrade), ())
-        }
+        SubstreamProtocol::new(self.config.protocol_config.clone(), ()).map_upgrade(Either::Left)
```
thomaseizinger: You should be able to now simplify `InboundUpgrade` to no longer use `Either`.

chirag-bgh: Does it mean I just have to remove ```.map_upgrade(Either::Left)``` from here?

thomaseizinger: Yes and change the `InboundUpgrade` associated type of `KademliaHandler`.

protocols/kad/src/handler.rs
```diff
@@ -611,19 +605,13 @@ where
     type InEvent = KademliaHandlerIn<TUserData>;
     type OutEvent = KademliaHandlerEvent<TUserData>;
     type Error = io::Error; // TODO: better error type?
-    type InboundProtocol = Either<KademliaProtocolConfig, upgrade::DeniedUpgrade>;
```
thomaseizinger, line 614: That is not quite correct. You still need to set the associated type but just to `KademliaProtocolConfig` instead of the composed `Either`.

thomaseizinger:
CI is still failing. Did you manage to compile this locally? I typically run `cargo custom-clippy` for local development.

protocols/kad/src/handler.rs
```diff
@@ -534,13 +531,6 @@ where
             <Self as ConnectionHandler>::InboundOpenInfo,
         >,
     ) {
-        // If `self.allow_listening` is false, then we produced a `DeniedUpgrade` and `protocol`
-        // is a `Void`.
-        let protocol = match protocol {
-            future::Either::Left(p) => p,
-            future::Either::Right(p) => void::unreachable(p),
-        };
-
```
chirag-bgh, line 543: Is this correct? I removed this  since ```self.allow_listening``` is no more present and we don't have to produce a `void` protocol anymore.

thomaseizinger: Yep that is correct!

thomaseizinger:
There are two warnings about unused imports left.

Also, this is unfortunately a breaking change because the `handler` module is public. I would like to change that. I am going to hold off on merging this until we decide to make the next breaking change. Then we can include this. Making a breaking change for removing an obsolete, internal config option is not worth it in my eyes.

I hope you don't mind. The change is relatively isolated so there shouldn't be many merge conflicts.

I am going to open an issue about minimizing the public API of `libp2p-kad`.

thomaseizinger:
> I am going to hold off on merging this until we decide to make the next breaking change. 

Setting this to draft to prevent accidental merge.

thomaseizinger:
I wrote some more about this here: https://github.com/libp2p/rust-libp2p/discussions/3532

thomaseizinger:
> I am going to open an issue about minimizing the public API of `libp2p-kad`.

Done here: https://github.com/libp2p/rust-libp2p/issues/3533

mxinden:
> Also, this is unfortunately a breaking change because the `handler` module is public. I would like to change that. I am going to hold off on merging this until we decide to make the next breaking change. Then we can include this. Making a breaking change for removing an obsolete, internal config option is not worth it in my eyes.

I don't think we need to consider this a breaking change. I don't think anyone is using the `libp2p-kad` `ConnectionHandler` directly. @thomaseizinger do you feel strongly about being strict here?

thomaseizinger:
> > Also, this is unfortunately a breaking change because the `handler` module is public. I would like to change that. I am going to hold off on merging this until we decide to make the next breaking change. Then we can include this. Making a breaking change for removing an obsolete, internal config option is not worth it in my eyes.
> 
> I don't think we need to consider this a breaking change. I don't think anyone is using the `libp2p-kad` `ConnectionHandler` directly. @thomaseizinger do you feel strongly about being strict here?

You'll have to override CI and merge it manually then because `cargo semver-checks` flags it as a breaking change. To avoid a red CI, we'd have to immediately make a patch release.

Feel free to do so if you want. Is there a pressing need though?

mxinden:
> Feel free to do so if you want. Is there a pressing need though?

Good point. In the contrived example where someone is in need for this patch, I would consider ignoring the fact that it is a breaking change. Given that no one directly depends on the change, let's treat it as a breaking change.

thomaseizinger:
We are going to need this for client-mode right. Should we then really remove it? See https://github.com/libp2p/rust-libp2p/pull/3651.

mxinden:
For client mode it will need to be dynamic during the lifetime of a single `ConnectionHandler`, and not a configuration flag static during the entire lifetime of a `ConnectionHandler`. Does that make sense?

thomaseizinger:
> For client mode it will need to be dynamic during the lifetime of a single `ConnectionHandler`, and not a configuration flag static during the entire lifetime of a `ConnectionHandler`. Does that make sense?

That flag can also be modified by a message coming in from the behaviour. As soon as it is changed to `true`, the next inbound stream will also provide the kademlia protocol.

I think we will need the flag, or something equivalent. We just don't have the "change the flag" mechanism yet.

thomaseizinger:
This is obsolete with https://github.com/libp2p/rust-libp2p/pull/3877.

